PR #1964, opened 2021-06-09
OCPBUGSM-29819: Installation fails after cancel and reset of installation before node reboot

ori-amizur:
Bug 1966545:
This is an edge case that will probably not happen when using with UI.  It happens when cluster starts installing, then installation is cancelled (and possibly reset), and then starts installing again.
Since the cluster preparation for installation may be long, there is a possibility that two go-routines (one per each installation attempt) each preparing the same cluster will run in parallel.
Both of these go-routines are using the same work directory **prefix/cluster-id** for preparation artifacts and may collide with each other.
The solution for it is to let each of these go-routines to use different work directory, so they will not disturb one another.
The work dir is a temporary directory (ioutil.TempDir).

- [x] Cloud
- [ ] None

# How was this code tested?

- [ ] dev-scripts environment
- [x] Manual (Elaborate on how it was tested) - Simulated with bash script that starts installation, cancels installation, and starts installation again
- [ ] No tests needed

- Is this PR relying on CI for an e2e test run? No
- Should this PR be tested in a specific environment? No

# Assignees

Please, add one or two reviewers that could help review this PR.

/assign @empovit 

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/generator/generator.go
+++ b/pkg/generator/generator.go
@@ -2,6 +2,7 @@ package generator
 
 import (
 	"context"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
@@ -51,12 +52,15 @@ func New(log logrus.FieldLogger, s3Client s3wrapper.API, cfg Config, workDir str
 // GenerateInstallConfig creates install config and ignition files
 func (k *installGenerator) GenerateInstallConfig(ctx context.Context, cluster common.Cluster, cfg []byte, releaseImage string) error {
 	log := logutil.FromContext(ctx, k.log)
-	clusterWorkDir := filepath.Join(k.workDir, cluster.ID.String())
-	installerCacheDir := filepath.Join(k.workDir, "installercache")
-	err := os.MkdirAll(clusterWorkDir, 0755)
-	if err != nil && !os.IsExist(err) {
+	err := os.MkdirAll(k.workDir, 0o755)
+	if err != nil {
 		return err
 	}
+	clusterWorkDir, err := ioutil.TempDir(k.workDir, cluster.ID.String()+".")
+	if err != nil {
+		return err
+	}
+	installerCacheDir := filepath.Join(k.workDir, "installercache")
 	defer func() {
 		// keep results in case of failure so a human can debug
 		if err != nil {


osherdp:
@ori-amizur if you'll write "Bug 966545: ..." in the PR title it will sync BZ and this PR

tsorya:
don't we need it after installation?

ori-amizur:
> installation

It is erased after upload to S3

https://github.com/openshift/assisted-service/blob/13c2dfc4588e507637d9cb3eed6ce28d701a5690/pkg/generator/generator.go#L77

pkg/generator/generator.go
```diff
@@ -51,9 +52,16 @@ func New(log logrus.FieldLogger, s3Client s3wrapper.API, cfg Config, workDir str
 // GenerateInstallConfig creates install config and ignition files
 func (k *installGenerator) GenerateInstallConfig(ctx context.Context, cluster common.Cluster, cfg []byte, releaseImage string) error {
 	log := logutil.FromContext(ctx, k.log)
-	clusterWorkDir := filepath.Join(k.workDir, cluster.ID.String())
+	err := os.MkdirAll(k.workDir, 0o755)
```
tsorya, line 55: o?

pkg/generator/generator.go
```diff
@@ -51,9 +52,16 @@ func New(log logrus.FieldLogger, s3Client s3wrapper.API, cfg Config, workDir str
...
+	if err != nil {
+		return err
+	}
+	clusterWorkDir, err := ioutil.TempDir(k.workDir, cluster.ID.String()+".")
```
tsorya, line 59: cluster.ID.String()+"." ? /folder/cluster-id. ? maybe better timestamp or something as /folder/cluster-id_temp?

pkg/generator/generator.go
```diff
@@ -51,9 +52,16 @@ func New(log logrus.FieldLogger, s3Client s3wrapper.API, cfg Config, workDir str
...
+	if err != nil {
+		return err
+	}
 	installerCacheDir := filepath.Join(k.workDir, "installercache")
-	err := os.MkdirAll(clusterWorkDir, 0755)
+	err = os.MkdirAll(clusterWorkDir, 0755)
```
tsorya, line 64: i think it is not needed anymore

pkg/generator/generator.go
empovit, line 55: What's the meaning of the `o` in `0o755`?

pkg/generator/generator.go
empovit: @tsorya is right, you should have a `clusterWorkDir` by this time, `ioutil.TempDir(k.workDir, cluster.ID.String()+".") created it with the parent directory taken care of by `os.MkdirAll(k.workDir, 0o755)`.

ori-amizur: OK

pkg/generator/generator.go
ori-amizur: octal

PR #1970, opened 2021-06-10
NO-ISSUE: Added custom manifests to enable testing for OCS internal builds

priyanka19-98:
Signed off by: Priyanka Jiandani <pjiandan@redhat.com>


internal/operators/ocs/custom_manifest.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+// CatalogSource Manifest for testing internal builds
+func ocsCatalogSource() (string, error) {
+
+	image := "quay.io/rhceph-dev/ocs-registry:latest"
```
petr-balogh, line 42: Actually as engineering building also previous versions for z-streams there is no latest tag.
We have tags like:
`latest-stable-4.8`
`latest-stable-4.7`
So I would go with one of this as default one. Or we have also `latest-4.8` but this build was not tested and we don't know if it's even deployable, so the stable tag is better.

internal/operators/ocs/custom_manifest.go
```diff
@@ -0,0 +1,82 @@
...
+    base64data: ""
+    mediatype: ""
+  image: "{{.OPERATOR_IMAGE}}"
+  publisher: Red Hat
+  sourceType: grpc
+  # If the registry image still have the same tag (latest-stable-4.6, or for stage testing)
+  # we need to have this updateStrategy, otherwise we will not see new pushed content.
+  updateStrategy:
+    registryPoll:
+      interval: 15m`
```
petr-balogh, line 69: This part can be maybe deleted from here.
It has its pros and cons.

Pros: It keeps you cluster upgraded when new stable build is produced and you are using such tag like latest-stable-4.6.

But if you are in the middle of the run - you maybe don't want to have cluster upgraded in the middle - for this purpose user should rather use some static tag like:  4.8.0-414.ci

I think you can remove it from here probably. For manual re-load of the catalog, user can also delete the pod for ocs-catalogsource from openshift-marketplace.

internal/operators/ocs/custom_manifest.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	const ocsSubscription = `apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: "{{.OPERATOR_SUBSCRIPTION_NAME}}"
+  namespace: "{{.OPERATOR_NAMESPACE}}"
+spec:
+  installPlanApproval: Automatic
+  name: ocs-operator
```
petr-balogh, line 23: How the channel is picked up @priyanka19-98 ? Or if it's not specified it will take the default which is defined as the default channel in the packageManifest?  Or is there any other logic later on which is updating channel in the subscription?

priyanka19-98: default channel from the packageManifest is taken.

internal/operators/ocs/custom_manifest.go
priyanka19-98: updated

internal/operators/ocs/custom_manifest.go
priyanka19-98: updated!

internal/operators/ocs/custom_manifest.go
priyanka19-98: do we need a way to specify channel as well?

petr-balogh: We are always specifying in our automation so I am not sure how it works if not defined as I haven't tried.
Have you tried to verify?

internal/operators/ocs/custom_manifest.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+// CatalogSource Manifest for testing internal builds
+func ocsCatalogSource() (string, error) {
+
+	image := "quay.io/rhceph-dev/ocs-registry:latest-stable-4.8"
```
nbalacha, line 42: This will require the pull secret for this repo.

internal/operators/ocs/custom_manifest.go
priyanka19-98: yes, we don't specify a channel and it takes the latest channel always. But for testing I guess we would need to specify a channel. What should be the default channel?

internal/operators/ocs/custom_manifest.go
priyanka19-98: yes, it will be required to add the Pull Secret if using internal builds

internal/operators/ocs/custom_manifest.go
petr-balogh: Depends on version. It's stable-x.y where x.y is installed version of ocs. So if you will use image for 4.8 then stable-4.8 . If testing 4.7 build, then stable-4.7.

Makefile
```diff
@@ -69,6 +69,9 @@ STORAGE := $(or ${STORAGE},s3)
 GENERATE_CRD := $(or ${GENERATE_CRD},true)
 PERSISTENT_STORAGE := $(or ${PERSISTENT_STORAGE},True)
 IPV6_SUPPORT := $(or ${IPV6_SUPPORT}, True)
+OCS_TEST_INTERNAL_BUILD:=$(or ${OCS_TEST_INTERNAL_BUILD},False)
```
petr-balogh, line 72: I see that in the rest of the file it's used to have spaces around `:=` but it's not the case of those 3 new lines. Probably better to follow the same style in the file.

```suggestion
OCS_TEST_INTERNAL_BUILD := $(or ${OCS_TEST_INTERNAL_BUILD},False)
```

README.md
```diff
@@ -351,6 +351,13 @@ It's possible to modify the discovery ISO (via the API) to enable password login
 
 More information is available here: [Set discovery ISO user password example](docs/set-discovery-password.md)
 
+## Testing internal builds of OCS
+```shell
+export OCS_TEST_IMAGE=<custom build image>
```
petr-balogh, line 356: Can you please also add an example of whole url pointing to the build in quay.io? 
```
quay.io/rhceph-dev/ocs-registry:4.8.0-420.ci
```
This can be mentioned maybe after the shell block as example of: `<custom build image>`

priyanka19-98: done

petr-balogh (approved):

petr-balogh (approved):

priyanka19-98:
closing the PR as its no longer relevant.

PR #1992, opened 2021-06-14
Bug 1937696: Use hostname for BMH resource name

mkowalski:
# Description

Currently BMH resources in the clusters installed by AI are named in
the form of `openshift-{master|worker}-{0...}`. This PR changes the
convention so that a real hostname of the node will be used, or its UUID
as a fallback option.

For virtual environments this gives us 3 different naming schemas

- Node: libvirt domain name
- BareMetalHost: `openshift-{master|worker}-{0...}`
- Machine: `cluster_name-ID-{master|worker}-{0...}`

This PR unifies naming between Node and BareMetalHost, so that the
output of the 3 following commands is more consistent

- `oc get nodes`
- `oc -n openshift-machine-api get baremetalhosts`
- `oc -n openshift-machine-api get machines`

Closes: [OCPBUGSM-26086](https://issues.redhat.com/browse/OCPBUGSM-26086)

- [x] Cloud
- [ ] None

# How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Manual test

Blablabla

# Assignees

/assign @
/assign @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/installcfg/installcfg.go
+++ b/internal/installcfg/installcfg.go
@@ -6,7 +6,6 @@ import (
 	"encoding/pem"
 	"fmt"
 	"sort"
-	"strconv"
 	"strings"
 
 	"github.com/go-openapi/swag"
@@ -143,18 +142,6 @@ func (i *installConfigBuilder) countHostsByRole(cluster *common.Cluster, role mo
 	return count
 }
 
-func (i *installConfigBuilder) getBMHName(host *models.Host, masterIdx, workerIdx *int) string {
-	prefix := "openshift-master-"
-	index := masterIdx
-	if host.Role == models.HostRoleWorker {
-		prefix = "openshift-worker-"
-		index = workerIdx
-	}
-	name := prefix + strconv.Itoa(*index)
-	*index = *index + 1
-	return name
-}
-
 func (i *installConfigBuilder) getNetworkType(cluster *common.Cluster) string {
 	networkType := "OpenShiftSDN"
 	if network.IsIPv6CIDR(cluster.ClusterNetworkCidr) || network.IsIPv6CIDR(cluster.MachineNetworkCidr) || network.IsIPv6CIDR(cluster.ServiceNetworkCidr) {
@@ -277,8 +264,6 @@ func (i *installConfigBuilder) setBMPlatformInstallconfig(cluster *common.Cluste
 	hosts := make([]host, numWorkers+numMasters)
 
 	yamlHostIdx := 0
-	masterIdx := 0
-	workerIdx := 0
 	sortedHosts := make([]*models.Host, len(cluster.Hosts))
 	copy(sortedHosts, cluster.Hosts)
 	sort.Slice(sortedHosts, func(i, j int) bool {
@@ -292,8 +277,9 @@ func (i *installConfigBuilder) setBMPlatformInstallconfig(cluster *common.Cluste
 		if swag.StringValue(host.Status) == models.HostStatusDisabled {
 			continue
 		}
-		i.log.Infof("host name is %s", hostutil.GetHostnameForMsg(host))
-		hosts[yamlHostIdx].Name = i.getBMHName(host, &masterIdx, &workerIdx)
+		hostName := hostutil.GetHostnameForMsg(host)
+		i.log.Infof("host name is %s", hostName)
+		hosts[yamlHostIdx].Name = hostName
 		hosts[yamlHostIdx].Role = string(host.Role)
 
 		var inventory models.Inventory
--- a/internal/installcfg/installcfg_test.go
+++ b/internal/installcfg/installcfg_test.go
@@ -84,6 +84,19 @@ var _ = Describe("installcfg", func() {
 		Expect(len(result.Platform.Baremetal.Hosts)).Should(Equal(3))
 	})
 
+	It("create_configuration_with_hostnames", func() {
+		var result InstallerConfigBaremetal
+		mockMirrorRegistriesConfigBuilder.EXPECT().IsMirrorRegistriesConfigured().Return(false).Times(2)
+		data, err := installConfig.GetInstallConfig(&cluster, false, "")
+		Expect(err).ShouldNot(HaveOccurred())
+		err = yaml.Unmarshal(data, &result)
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(len(result.Platform.Baremetal.Hosts)).Should(Equal(3))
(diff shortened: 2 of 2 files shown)


mkowalski:
`install-config.yaml` generated during [CI run]

```
[...]
platform:
  baremetal:
    provisioningNetwork: Disabled
    apiVIP: 192.168.127.91
    ingressVIP: 192.168.127.23
    hosts:
    - name: test-infra-cluster-d63379ba-master-0
      role: master
      bootMACAddress: 02:00:00:46:6e:e7
      bootMode: legacy
    - name: test-infra-cluster-d63379ba-master-1
      role: master
      bootMACAddress: 02:00:00:20:c1:bd
      bootMode: legacy
    - name: test-infra-cluster-d63379ba-master-2
      role: master
      bootMACAddress: 02:00:00:ad:00:15
      bootMode: legacy
    - name: test-infra-cluster-d63379ba-worker-0
      role: worker
      bootMACAddress: 02:00:00:77:ad:e1
      bootMode: legacy
    - name: test-infra-cluster-d63379ba-worker-1
      role: worker
      bootMACAddress: 02:00:00:cd:34:03
      bootMode: legacy
[...]
```

[CI run]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-service/1992/pull-ci-openshift-assisted-service-master-system-test/1404367150545637376/artifacts/system-test/baremetalds-assisted-gather/artifacts/2021-06-14_10:21:30_9795b597-5f8f-44bb-8e2b-962103c44d70/cluster_files/install-config.yaml

internal/installcfg/installcfg.go
```diff
@@ -144,15 +143,7 @@ func (i *installConfigBuilder) countHostsByRole(cluster *common.Cluster, role mo
...
-	prefix := "openshift-master-"
-	index := masterIdx
-	if host.Role == models.HostRoleWorker {
-		prefix = "openshift-worker-"
-		index = workerIdx
-	}
-	name := prefix + strconv.Itoa(*index)
-	*index = *index + 1
-	return name
+	return hostutil.GetHostnameForMsg(host)
```
pawanpinjarkar, line 146: I think instead of this, you can directly do 
```
hostName := hostutil.GetHostnameForMsg(host)
i.log.Infof("host name is %s", hostName)
hosts[yamlHostIdx].Name = hostName
```
here https://github.com/openshift/assisted-service/pull/1992/files#diff-f9d8c8224b172499b34cb44479e6e3606f19036989fa80449d085d1f3780e6eeR286-R287
and remove `getBMHName` function alltogether.

mkowalski: Good catch, thanks! 👍🏻

mkowalski:
`quay.io/mkowalski/assisted-service:bz1937696-1` is a test image

mkowalski:
/retitle Bug 1937696: Use hostname for BMH resource name

Still needs full E2E test, but already the install-config.yaml has a correct content

mkowalski:
/hold

It fails every time on `ci/prow/system-test`. I suspect it may be a valid error and not a CI shakiness, but that needs a bit of investigation

mkowalski:
/test subsystem-aws
Failed on `[minimal-set][full-iso]create_and_get_image`. Not related to the PR

flaper87 (commented):
This looks good! Would love to get some sign-off from folks in cloud/core! 

/approve

internal/installcfg/installcfg.go
```diff
@@ -292,8 +277,9 @@ func (i *installConfigBuilder) setBMPlatformInstallconfig(cluster *common.Cluste
 		if swag.StringValue(host.Status) == models.HostStatusDisabled {
 			continue
 		}
-		i.log.Infof("host name is %s", hostutil.GetHostnameForMsg(host))
-		hosts[yamlHostIdx].Name = i.getBMHName(host, &masterIdx, &workerIdx)
+		hostName := hostutil.GetHostnameForMsg(host)
```
ori-amizur, line 280: Why are we using GetHostnameForMsg and not GetCurrentHostName ?

mkowalski: `GetHostnameForMsg` is calling `GetCurrentHostName` and adds additional handling in case the latter did not return a valid name. So I picked `GetHostnameForMsg` because it is a safer version of `GetCurrentHostName`

https://github.com/openshift/assisted-service/blob/master/internal/host/hostutil/host_utils.go#L32
https://github.com/openshift/assisted-service/blob/master/internal/host/hostutil/host_utils.go#L20

flaper87:
/hold 

Let's hold this until Monday to give the rest of the team some time to review it and make sure this doesn't break other platforms.

flaper87:
Now that it's held, I can:

/lgtm

internal/installcfg/installcfg.go
```diff
@@ -292,8 +277,9 @@ func (i *installConfigBuilder) setBMPlatformInstallconfig(cluster *common.Cluste
...
+		i.log.Infof("host name is %s", hostName)
+		hosts[yamlHostIdx].Name = hostName
```
eranco74, line 282: Just curious, is there any validation that checks that all hosts have different hostnames (I guess there isn't).
So far each BMH had a unique Name, what will happen if 2 (or more) hosts have the same hostname?

mkowalski: From my tests of this PR here, the new naming schema generates `test-infra-cluster-d63379ba-master-0` and so on, i.e. `clusterName-master|worker-0..n`

I couldn't backtrace exactly why it happens like this, but my best bet is that this is because at the moment when the `install-config.yaml` is generated not all the values are available for this function... In the `GetHostnameForMsg` we are getting in order the following: either `host.RequestedHostname`, `""` (which fails back to `host.ID`) or `inventory.Hostname`.

Having said this, the CR `Nodes` is already using the name coming directly from the target machine (so either OS visible hostname or libvirt domain name). If there were duplicated hostnames, `Nodes` CR would be the first one to fail and it would have been already failing in the past. By this line of reasoning I believe that even though we don't check explicitly for duplicates, this PR here is not introducing another failure path.

flaper87: This matches my understanding too!

flaper87 (commented):
Alright, we let this PR sit for a bit to give folks from other platforms time to review. Questions were asked here and in other means. It seems like the PR is good to go by the looks of it, so I will cancel the hold.

/hold cancel

PR #1988, opened 2021-06-14
OCPBUGSM-30853: Change default telemeter server of installed clusters to prod.

ybettan:
# Description

    Until now, in the service, we changed the default behavior of openshift,
    which is to always send cluster metrics to prod-telemeter-server unless
    changed by the user, and we:
        * left untouched clusters created by cloud prod env
        * redirected to telemeter-stage stage clusters
        * redirected to dummy-url all other clusters

    The issue with this approach is that all prod clusters that aren't
    created in the cloud, operator clusters for example,  will fail to
    deliver telemetry.

    Instead, we will now default to prod-telemeter instead of dummy-url
    unless we know better:
        * left untouched clusters created by ALL prod envs (not just cloud)
        * redirect to telemeter-stage stage clusters
        * redirect to dummy-url integration clusters

    Signed-off-by: Yoni Bettan <ybettan@redhat.com>

# What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [x] Cloud
- [x] Operator Managed Deployments
- [ ] None

# How was this code tested?

Please, select one or more if needed:

- [x] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

It's recommended to take a few extra minutes to provide more information about
how this code was tested. Here are some questions that may be worth answering:

- Should this PR be tested by the reviewer?
- Is this PR relying on CI for an e2e test run?
- Should this PR be tested in a specific environment?
- Any logs, screenshots, etc that can help with the review process?

**Manual system tests:**

I have run test-infra locally and we can see that no manifest was created to redirect the metrics (we defaulted to prod):
```
$ curl -s $(minikube service assisted-service -n assisted-installer --url)/api/assisted-install/v1/clusters/baa05cab-8b82-41e0-aef2-6771e350c2d2/manifests | jq '.'
[
  {
    "file_name": "50-masters-chrony-configuration.yaml",
    "folder": "openshift"
  },
  {
    "file_name": "50-workers-chrony-configuration.yaml",
    "folder": "openshift"
  }
]
```

Also, in the created cluster, we can see that the ConfigMap for redirection wasn't created:
```
$ oc --kubeconfig kubeconfig describe cm/cluster-monitoring-config -n openshift-monitoring
Error from server (NotFound): configmaps "cluster-monitoring-config" not found
```

# Assignees

[...]

PR #1993, opened 2021-06-14
Bug 1972096: Use .invalid domain instead of dummy.com

mkowalski:
# Description

As a convention, one of the following should be used when a dummy domain
name is required for testing or documentation purposes - `example.com`,
`.invalid`, `.example`, `.localhost` or `.test` as those are reserved by
IANA/IETF.

Using a domain from `.com` as a default value in our configuration may
cause undesired affects in the future if such a domain is not changed by
the operator and we use it for sending data out of the cluster.

Closes: [OCPBUGSM-30935](https://issues.redhat.com/browse/OCPBUGSM-30935)

- [x] Cloud
- [ ] None

# How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

# Assignees

/assign @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/network/manifests_generator.go
+++ b/internal/network/manifests_generator.go
@@ -315,7 +315,7 @@ data:
 	stageServiceBaseURL       = "https://api.stage.openshift.com"
 	integrationServiceBaseURL = "https://api.integration.openshift.com"
 	stageTelemeterURL         = "https://infogw.api.stage.openshift.com"
-	dummyURL                  = "https://dummy.com"
+	dummyURL                  = "https://dummy.invalid"
 )
 
 // Default Telemeter server is prod.
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -6291,7 +6291,7 @@ var _ = Describe("AMS subscriptions", func() {
 				subStrings: []string{"Updated status of cluster", clusterName},
 			}, gomock.Any())
 
-			dummyDNSDomain := "dummy.com"
+			dummyDNSDomain := "dummy.test"
 			reply = bm.UpdateCluster(ctx, installer.UpdateClusterParams{
 				ClusterID: *c.ID,
 				ClusterUpdateParams: &models.ClusterUpdateParams{
--- a/internal/cluster/cluster_test.go
+++ b/internal/cluster/cluster_test.go
@@ -2661,7 +2661,7 @@ var _ = Describe("Validation metrics and events", func() {
 
 	const (
 		openshiftVersion = "dummyVersion"
-		emailDomain      = "dummy.com"
+		emailDomain      = "dummy.test"
 	)
 
 	var (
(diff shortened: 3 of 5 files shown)


internal/network/manifests_generator.go
```diff
@@ -315,7 +315,7 @@ data:
 	prodServiceBaseURL  = "https://api.openshift.com"
 	stageServiceBaseURL = "https://api.stage.openshift.com"
 	stageTelemeterURL   = "https://infogw.api.stage.openshift.com"
-	dummyURL            = "https://dummy.com"
+	dummyURL            = "https://dummy.example"
```
carbonin, line 318: I liked the suggestion of `.invalid` for this.
Why change to `.example`?

mkowalski: I don't have a particular preference, but it's more of a convention - it's a default value present if the user does not set anything explicitly. In such a scenario it feels somehow not natural that we by default put an invalid value - I see it a bit like default "-1" in a field asking for my age.

But it's not a strong preference, we can go with `dummy.invalid`

carbonin: I suppose that depends on the case when we use this `dummyURL` is this something we expect users to update manually?
I thought it was something we explicitly expected to be unused.

If it's the former then, yeah, example makes more sense. @ybettan ?

mkowalski: Ok, changed to `.invalid`

ybettan:
@celebdor
Does this PR requires another PR to `ocm-2.3`.

I am not sure how it works but this PR is based on https://github.com/openshift/assisted-service/pull/1988 which does have a clone PR in `ocm-2.3`.

mkowalski:
I'd say it's small enough that there's no strong reason not to backport it...

/cherry-pick ocm-2.3

mkowalski:
/retitle Bug 1972096: Use .invalid domain instead of dummy.com

mbukatov:
Hmm, what is wrong with [`example.com`](https://www.iana.org/domains/reserved)? 

> As described in RFC 2606 and RFC 6761, a number of domains such as example.com and example.org are maintained for documentation purposes. These domains may be used as illustrative examples in documents without prior coordination with us. They are not available for registration or transfer.

carbonin:
@mbukatov see the discussion in https://github.com/openshift/assisted-service/pull/1993#discussion_r651003420

mkowalski:
@mbukatov `example.com` would have been perfectly okay but here the `dummy.com` was used and that one is registered by someone, not reserved by IANA

Another thing is that `example.com` has an existing backend behind whereas `.invalid` and others are not resolved by the DNS. This means that using `example.com` in the code that is not doing anything useful is causing a bit more of useless traffic going around. A bit here, a bit there and it amplifies the effect :)

flaper87 (approved):

PR #2011, opened 2021-06-15
Bug 1972262: Change default value for BMH_DETACHED_ANNOTATION

mkowalski:
Currently for the `baremetalhost.metal3.io/detached` annotation we are
using `"true"` as a value whenever we decide to create it. Given that,
based on the behaviour of the baremetal-operator, this annotation
accepts string and not a boolean, it may create a confusion if an
administrator decides to set the value to e.g. `"false"` or
`"disabled"`.

It is a design decision in the upstream that the annotation could
provide a context from whatever external system the annotation is coming
from and that the value itself is completely ignored.

Conflicts:
  - internal/controller/controllers/bmh_agent_controller_test.go

Cherry-picks: openshift#1974
Closes: [OCPBUGSM-30953](https://issues.redhat.com/browse/OCPBUGSM-30953)

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -316,7 +316,7 @@ func (r *BMACReconciler) addBMHDetachedAnnotationIfAgentHasStartedInstallation(c
 		bmh.ObjectMeta.Annotations = make(map[string]string)
 	}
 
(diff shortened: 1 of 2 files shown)


rwsu (approved):

flaper87 (approved):

mkowalski:
@flaper87, rebase was required, please lgtm again :)

flaper87 (approved):
/lgmt

PR #2012, opened 2021-06-15
MGMT-6942: Bump OCP versions 4.8.0-fc.8 -> 4.8.0-fc.9

oshercc:
@gamli75 @oshercc @YuviGold
4.8 has no available release notes


oshercc:
Running all tests

gamli75:
@oshercc can you verify why we delete the config/onprem-iso-config.ign instead of only updating it - the same as we did in https://github.com/openshift/assisted-service/commit/3cab52dfe9474282b7a7d71139e7447f7ae827c2 ?

gamli75 (approved):

PR #2042, opened 2021-06-20
MGMT-6908: Adding enhancement for installation-progress-bar.

ybettan:

- [x] Cloud
- [ ] None

# How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

# Assignees

this PR directly to someone.

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/installation-progress-bar.md
+++ b/docs/enhancements/installation-progress-bar.md
@@ -1,0 +1,164 @@
+# Cluster installation progress bar
+
+## Goals
+
+* Progress should be computed in the BE by design and not in the UI
+  * that will also make it usable from other APIs as kube-api
+* Installed clusters with failed non-essential workers or OLM-operators should still reach 100% progress.
+
+
+## Current implementation
+
+* Computed in UI
+* `progressPercent = hostsProgressPercent * 0.75 + operatorsProgressPercent * 0.25`
+  * `operatorsProgressPercent =(completedOperators / monitoredOperators) * 100`
+    * A completedOperator is one of the following:
+      * A built-in operator with `available` status
+      * An OLM operator with `available` or `failed` status
+  * `hostsProgressPercent = (ΣcompletedStages / ΣallStages) * 100`
+    * For each host the data is retrieved by using the following APIs
+
+```
+GET api/assisted-install/v1/clusters/{cluster_id} | jq '.hosts[0].progress'
+{
+  "current_stage": "Done",
+  "stage_started_at": "2021-05-31T13:37:26.224Z",
+  "stage_updated_at": "2021-05-31T13:37:26.224Z"
+}
+
+GET /api/assisted-install/v1/clusters/{cluster_id} | jq '.hosts[0].progress_stages'
+[
+  "Starting installation",
+  "Installing",
+  "Writing image to disk",
+  "Waiting for control plane",
+  "Rebooting",
+  "Configuring",
+  "Joined",
+  "Done"
+]
+```
+
+## New implementation
+
+* Will be computed in BE
+* `progres =ΣclusterStageProgress * Wi = preparingForInstallation * Wpi + installing * Wi + finalizing * Wf` while `ΣWi=1`
+  * We can manually compute an avg duration for the previous installations based on events to choose those weights and hardcode them.
+
+
+#### Preparing for installation
+
+This stage mostly generates manifests and installation configs, therefore, we will set  `preparingForInstallationProgress={1, 0}` depending on if it is done or not.
+
+
+#### Installing
+
+* This stage refers to hosts' installation
+* We will “assume” all host’s stages have a similar duration, therfore, they will be equally weighted.  
+* Hosts' progresses will be computed the same way we it is done in the current implementation, aka `hostsProgressPercent = ΣcompletedStages / ΣallStages`
+
+
+#### Finalizing
+
+* This stage refers to operators' installation
+* We will “assume” all operatos have a similar installation duration.
+* Operators' progresses will be computed the same way we it is done in the current implementation, aka `operatorsProgressPercent = completedOperators / monitoredOperators`
+
+Suggested APIs:
+* Note that those new percentage values should be reset on clusters' installation reset
+
+```
+diff --git a/swagger.yaml b/swagger.yaml
+index d53ddf59..ddbc7e9b 100644
+--- a/swagger.yaml
++++ b/swagger.yaml
+@@ -4171,6 +4171,9 @@ definitions:
+         format: date-time
+         x-go-custom-tag: gorm:"type:timestamp with time zone"
+         description: The last time that the cluster status was updated.
++      progress:
++        type: object
++        ref: '#/definitions/cluster-progress-info'
+       hosts:
+         x-go-custom-tag: gorm:"foreignkey:ClusterID;association_foreignkey:ID"
+         type: array
+@@ -4458,8 +4461,12 @@ definitions:
+   host-progress-info:
+     type: object
+     required:
++      - installation_percentage
+       - current_stage
+     properties:
++      installation_percentage:
++        type: integer
+       current_stage:
+         type: string
+         $ref: '#/definitions/host-stage'
+@@ -4477,6 +4484,21 @@ definitions:
+         x-go-custom-tag: gorm:"type:timestamp with time zone"
+         description: Time at which the current progress stage was last updated.
+
++  cluster-progress-info:
++    type: object
++    required:
++      - total_percentage
++    properties:
++      total_percentage:
++        type: integer
++      preparing_for_installation_stage_percentage:
++        type: integer
++      installing_stage_percentage:
++        type: integer
++      finalizing_stage_percentage:
++        type: integer
++
+   host-stage:
+     type: string
+     enum:
+```
+
+```
+GET /api/assisted-install/v1/clusters/{cluster_id}
+{
+  "id": "3a745598-dde4-46dd-bdb1-ce7e1dea119b",
+  "progress":                                    <-------- NEW
+  {
+    "total_percentage": 21,
+    "preparing_for_installation_stage_percentage": 100,
+    "installing_stage_percentage": 38,
+    "finalizing_stage_percentage": 0,
+  }
+  "hosts": [
+    {
+      "kind": "Host",
+      "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+      "role": "master",
+      "progress": {
+        "installation_percentage": 15,                  <-------- NEW (but coming from hosts API)
+        "current_stage": "Starting installation",
+      },
+    },
+    {
+      "kind": "Host",
+      "id": "3fa85f64-5717-4562-b3fc-2c963f66afa7",
+      "role": "worker",
+      "progress": {
+        "installation_percentage": 13,                  <-------- NEW (but coming from hosts API)
+        "current_stage": "Starting installation",
+      },
+    },
+  ]
+}
+```
+
+```
+GET /api/assisted-install/v1/clusters/{cluster_id}/hosts/{host_id}
+{
+  "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+  "progress": {
+    "installation_percentage": 15,                  <-------- NEW
+    "current_stage": "Starting installation",
+    "progress_info": "string",
+  },
+}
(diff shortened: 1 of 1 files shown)


docs/enhancements/installation-progress-bar.md
```diff
@@ -0,0 +1,116 @@
+# Cluster installation progress bar
+
+## Goals
+
+* Progress should be computed in the BE by design and not in the UI
+  * that will also make it usable from other APIs as kube-api
+* Failing workers or OLM-operators shouldn't fail the cluster installation
+  * installed clusters with failing hosts or failing OLM-operators should still reach 100% progress.
```
avishayt, line 8:
```suggestion
* Installed clusters with failed non-essential workers or OLM-operators should still reach 100% progress.
```

docs/enhancements/installation-progress-bar.md
```diff
@@ -0,0 +1,116 @@
...
+* We should align our design to other installation processes like OCM, ACM, etc.
+  * https://marvelapp.com/prototype/gfa4i5b/screen/80132841
```
avishayt, line 10: This is related to the UI, I would suggest it's not relevant here?

docs/enhancements/installation-progress-bar.md
```diff
@@ -0,0 +1,116 @@
...
+  "Configuring",
+  "Joined",
+  "Done"
+]
+```
+
+## New implementation
+
+* Will be computed in BE
+* `progres =ΣclusterStageProgress * Wi = preparingForInstallation * Wpi + installing * Wi + finalizing * Wf` while `ΣWi=1`
```
eranco74, line 48: Why not keep the current implementation logic and just move it to the backend?
Who will do all the progress calculations and when?
Will every host installation progress update trigger a calculation of the cluster progress? or it will get calculated and updated by some monitor?

ybettan: > Why not keep the current implementation logic and just move it to the backend?

Because the new implementation is more precise and more flexible.
By setting Wpi = 0, Wi = 0.75, Wf = 0.25 you get the current implementation anyway if that we want.

> Who will do all the progress calculations and when?

The BE, When is yet to be decided. Can be on API call, can be on refresh status etc.

> Will every host installation progress update trigger a calculation of the cluster progress? or it will get calculated and updated by some monitor?

Didn't decide yet, we can discuss it here. I think we can put it in the monitoring.

docs/enhancements/installation-progress-bar.md
ybettan: Done.

docs/enhancements/installation-progress-bar.md
ybettan: Removed.

docs/enhancements/installation-progress-bar.md
```diff
@@ -0,0 +1,113 @@
...
+
+Suggested APIs:
+
+```
+GET /api/assisted-install/v1/clusters/{cluster_id}
+{
+  "id": "3a745598-dde4-46dd-bdb1-ce7e1dea119b",
+  "progress":                                    <-------- NEW
+  {
+    "total_percent": "21",
```
jkilzi, line 75: I think we mentioned this a couple of times, please make sure these "*percent" field types are **not strings**.
Please provide numbers only; either floats between 0 and 1.0 or integers between 0 and 100.

docs/enhancements/installation-progress-bar.md
```diff
@@ -0,0 +1,113 @@
...
+    "finalizing_percent": "0",
+  }
+  "hosts": [
+    {
+      "kind": "Host",
+      "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
+      "role": "master",
+      "progress": {
+        "installation_percent": "15",                  <-------- NEW (but coming from hosts API)
+        "current_stage": "Starting installation",
```
jkilzi, line 87: I don't agree with the "current_stage" field semantics.  
As it is presented right now I can argue it is providing a description/explanation about the stage. If this is intentional, then its place is with the API documentation. 
If this field is meant to actually represent the host's current stage then its value must be constrained to be one of all possible stages a host can be at any moment during the installation. Please consider changing this to be an enum/iota value if this is the case.  
Right now this value is not something the UI can work with:
1. The content might need a microcopy in order to improve the UX
2. The casing of the words in the sentence might not follow UX best practices (this is a presentation layer concern)
3. Establishing the current stage in the FE code demands parsing a potentially large string.

Being that said, I understand that some consumers of the API will find value in having a description provided by the same API for the host's current stage. In this case, I suggest adding an additional field (called "description" maybe?) that can be optionally added to the API response by adding some URL params in the GET request, for example: `GET /api/assisted-install/v1/clusters/{cluster_id}?description=true`.

jkilzi (changes requested):

docs/enhancements/installation-progress-bar.md
ybettan: > If this field is meant to actually represent the host's current stage then its value must be constrained to be one of all possible stages a host can be at any moment during the installation

This is the case, please see https://github.com/openshift/assisted-service/blob/ad7260c5a8583fd945fd58096d617cb5ebfec669/swagger.yaml#L4483

Anyway, this is the current API (not part of the design), I just added it for "sanity", the new parts are marked by the `NEW` marker with `<----`.

I don't think we should address that in this design, if you think it should be changed, please consult with @gamli75  and if needed let us have a different task to change it.

docs/enhancements/installation-progress-bar.md
ybettan: Fixed.

docs/enhancements/installation-progress-bar.md
eranco74: So far the progress info didn't exist and we didn't need to keep it up to date.
I think that updating these new fields (host and cluster) will diminish the performance enhancements @tsorya and @ori-amizur recently added.

docs/enhancements/installation-progress-bar.md
jkilzi: > I don't think we should address that in this design, if you think it should be changed, please consult with
> @gamli75 and if needed let us have a different task to change it.
I agree with you on this we shouldn't tackle this issue here. I thought it was a new addition :)

docs/enhancements/installation-progress-bar.md
avishayt: Good point, but I think it should be OK:
1. During the actual installation the cluster resource is probably being modified anyway and we're not optimizing much if at all.
2. It would only need to be updated when a host finishes a stage, not constantly.

@tsorya and/or @ori-amizur should confirm.

docs/enhancements/installation-progress-bar.md
ybettan: Great :) 
Resolving the thread.

jkilzi (approved):

docs/enhancements/installation-progress-bar.md
ybettan: It may indeed.

@tsorya, @ori-amizur How bad do you guys think it will be to add those calculations to the monitoring processes?

What about computing it only on API call?
* On cluster API call:
  * trigger the hosts' progress computation
  * trigger the operators' progress computation
  * compute the total progress

* On Host API call trigger the hosts' computations only
* On Operator API call trigger the operators' computations only

tsorya: as i understand it will happened only during installation so shouldn't be a big deal

ronniel1 (commented):
Please add the swagger changes to the enhancement so we can all see what the new API will be

docs/enhancements/installation-progress-bar.md
```diff
@@ -0,0 +1,113 @@
...
+#### Installing
+
+* This stage refers to hosts' installation
+* We will “assume” all host’s stages have a similar duration, therfore, they will be equally weighted.  
+* Hosts' progresses will be computed the same way we it is done in the current implementation, aka `hostsProgressPercent = ΣcompletedStages / ΣallStages`
+
+
+#### Finalizing
+
+* This stage refers to operators' installation
```
gamli75, line 63: Please mention here that it's only for built-in operators and not OLM

docs/enhancements/installation-progress-bar.md
```diff
@@ -0,0 +1,113 @@
...
+* We will “assume” all operatos have a similar installation duration.
+* Operators' progresses will be computed the same way we it is done in the current implementation, aka `operatorsProgressPercent = completedOperators / monitoredOperators`
+
+Suggested APIs:
+
+```
+GET /api/assisted-install/v1/clusters/{cluster_id}
+{
+  "id": "3a745598-dde4-46dd-bdb1-ce7e1dea119b",
+  "progress":                                    <-------- NEW
```
gamli75, line 73: maybe we should call it installation_progress

gamli75:
Please also add a section about when we reset those fields - on cancel/ reset? what will happen if the preparing for installation will fail?

gamli75:
regarding the UI and the stepper presentation - from what I understand we will display 100% on the progress bar but the user will be on the OLM operators step - is it good UX? @nirfarkas

docs/enhancements/installation-progress-bar.md
ybettan: It is called `progress` in any other APIs (you can see it in the suggested API which is part of this PR).

docs/enhancements/installation-progress-bar.md
ybettan: It is not, finalizing refers to all operator's installation.

You can see it here https://github.com/ybettan/assisted-service/blob/e3418973cb0c2e08cd63fafab85afa0c3e0cddc8/internal/cluster/transition.go#L215-L216

docs/enhancements/installation-progress-bar.md
ori-amizur: I think adding the calculation to the monitor is not a good idea with regards to the performance improvements.  The logic of selecting the monitored entities might change soon, and we don't want progress change to trigger monitor on hosts or clusters. Computing it in API call seems more focused and consumes less resources.

ybettan:
I have added:
* swagger additions @ronniel1 
* comment on when those fields should be reset @gamli75 

Please review.

docs/enhancements/installation-progress-bar.md
ybettan: After talking to @ori-amizur this is how it will be handled:
* `preparing-for-installation` --> `installing` progress change will be handled by the monitoring process
  * The monitoring is the one responsible for changing that cluster stage.
* Hosts' stages progress will be updated by `PUT /clusters/{cluster_id}/hosts/{host_id}/progress` API
  * When this API is called to update the stage we will compute the new progress percentage as well.
* Operators' progress will be updated by the monitoring process
  * This is the only process that keeps track of the operators' statuses.

filanov: > On Host API call trigger the hosts' computations only
Shouldn't the cluster progress be calculated as well?

docs/enhancements/installation-progress-bar.md
```diff
@@ -0,0 +1,164 @@
...
+  {
+    "total_percentage": 21,
+    "preparing_for_installation_percentage": 100,
+    "installing_percentage": 38,
+    "finalizing_percentage": 0,
```
avishayt, line 129: Clarify that we are talking about stages here?  Otherwise installing_percentage vs. installation_percentage is confusing?

```suggestion
    "preparing_for_installation_stage_percentage": 100,
    "installing_stage_percentage": 38,
    "finalizing_stage_percentage": 0,
```

docs/enhancements/installation-progress-bar.md
ybettan: @filanov This is the updated solution decided https://github.com/openshift/assisted-service/pull/2042#discussion_r659296212.

Regarding the cluster progress I think the best is that each time an "internal" progress change, we compute the cluster total progress and store it as well.

This way when any of the API is called, we just need to fetch the data from the DB and return it to the user.

docs/enhancements/installation-progress-bar.md
ybettan: Fixed.

avishayt (approved):

PR #2061, opened 2021-06-22
OCPBUGSM-30767: AI Pods throws errors cannot access the media (ISO) - me…

michaellevy101:
# Description
When a media (ISO device) disconnection event has occurred, the host is unstable, and every command should fail.
The agent checks the device before every command and return a reply with the errorCode  "256" which indicates to the service that media disconnection event  has occurred.
The service changes the host status to "error", log the error message, and append the error message to the StatusInfo.
Our state machine keep send commands to the failed host and all those command failed (Virtual media disconnection requires a reboot) and the assist service logs are full with the message "cannot access the media (ISO)".
After this PR the service should log this message only once per host.
This is a short term solution and in the next version we should change the state machine and stop sending commands to an "unstable  hosts" or allow commands to run in this stage (for example we should collect the logs even after this failure) https://issues.redhat.com/browse/MGMT-7072

- [ ] Cloud
- [x] None

# How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

# Assignees

this PR directly to someone.

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -72,6 +72,7 @@ import (
 const DefaultUser = "kubeadmin"
 
 const WindowBetweenRequestsInSeconds = 10 * time.Second
+const mediaDisconnectionMessage = "Cannot read from the media (ISO) - media was likely disconnected"
 
 const (
 	MediaDisconnected int64 = 256
@@ -2857,7 +2858,7 @@ func (b *bareMetalInventory) PostStepReply(ctx context.Context, params installer
 			WithPayload(common.GenerateError(http.StatusNotFound, err))
 	}
 
-	logReplyReceived(params, log)
+	logReplyReceived(params, log, host)
 
 	if params.Reply.ExitCode != 0 {
 		handlingError := b.handleReplyError(params, ctx, log, &host.Host, params.Reply.ExitCode)
@@ -2917,8 +2918,7 @@ func (b *bareMetalInventory) handleReplyError(params installer.PostStepReplyPara
 }
 
 func (b *bareMetalInventory) handleMediaDisconnection(params installer.PostStepReplyParams, ctx context.Context, log logrus.FieldLogger, h *models.Host) error {
-	statusInfo := fmt.Sprintf("%s - %s", string(models.HostStageFailed),
-		"Cannot read from the media (ISO) - media was likely disconnected")
+	statusInfo := fmt.Sprintf("%s - %s", string(models.HostStageFailed), mediaDisconnectionMessage)
 
 	// Install command reports its status with a different API, directly from the assisted-installer.
 	// Just adding our diagnose to the existing error message.
@@ -3114,7 +3114,11 @@ func handleReplyByType(params installer.PostStepReplyParams, b *bareMetalInvento
 	return err
 }
 
-func logReplyReceived(params installer.PostStepReplyParams, log logrus.FieldLogger) {
+func logReplyReceived(params installer.PostStepReplyParams, log logrus.FieldLogger, host *common.Host) {
+	if !shouldStepReplyBeLogged(params, host) {
+		return
+	}
+
 	message := fmt.Sprintf("Received step reply <%s> from cluster <%s> host <%s> exit-code <%d> stderr <%s>",
 		params.Reply.StepID, params.ClusterID, params.HostID, params.Reply.ExitCode, params.Reply.Error)
 	messageWithOutput := fmt.Sprintf("%s stdout <%s>", message, params.Reply.Output)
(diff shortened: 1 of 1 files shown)


internal/bminventory/inventory.go
```diff
@@ -3131,6 +3135,11 @@ func logReplyReceived(params installer.PostStepReplyParams, log logrus.FieldLogg
 	}
 }
 
+func shouldBeLogged(params installer.PostStepReplyParams, host *common.Host) bool{
+	return !(params.Reply.ExitCode == MediaDisconnected && *host.Status == string(models.HostStageFailed) &&
+		strings.Contains(*host.StatusInfo, mediaDisconnectionMessage))
```
ronniel1, line 3140: Why do we need to check both the host.StatusInfo and the MediaDisconnected exit code?
If we got media disconnection, and we are in failed stage, than that is enough, no?

michaellevy101: If the installation has failed the assisted-installer itself reports an error (not indicative) and change the status  and afterward the agent reports the error (with the media disconnection indicative error) so i want it to be logged but only one time.

omertuc:
Please fill the PR description field in the template. The commit is not clear. The bug doesn't give enough context (and even if it did, don't rely on it - bugs usually contain a lot of irrelevant discussion with too much information). The commit message should summarize it in a sentence or two.

I have a lot of questions right now:
- When are those errors happening? What triggers them (reconciliation? cluster refresh?)
- Which kind of errors? Are they events? Logs? 
- Were the errors wrong? Why are we changing them? Or were there just too many?
- What was the old behavior with the errors?
- What is the new behavior? What changed? What is this commit trying to achieve that wasn't there before?

It sounds like a lot of questions but they can practically be answered with 1 or 2 sentences, and it would greatly help to both save time for reviewers and to allow for a much better review, it's hard to review without understanding the context of changes.

internal/bminventory/inventory.go
```diff
@@ -77,6 +77,7 @@ const (
 	MediaDisconnected int64 = 256
 	// 125 is the generic exit code for cases the error is in podman / docker and not the container we tried to run
 	ContainerAlreadyRunningExitCode = 125
+	mediaDisconnectionMessage       = "Cannot read from the media (ISO) - media was likely disconnected"
```
omertuc, line 80: Doesn't seem like this message should be grouped with those other constants under the same `()`

internal/bminventory/inventory.go
```diff
@@ -3131,6 +3135,11 @@ func logReplyReceived(params installer.PostStepReplyParams, log logrus.FieldLogg
 	}
 }
 
+func shouldBeLogged(params installer.PostStepReplyParams, host *common.Host) bool {
```
omertuc, line 3138: `shouldBeLogged` could use some documentation and a better name `shouldStepReplyBeLogged`, I think? Why don't we log under those conditions?

omertuc: Maybe put `params.Reply.ExitCode == MediaDisconnected && *host.Status == models.HostStatusError &&
		host.StatusInfo != nil)` inside a var with a name like `isStepReplyDisconnectionError`
		
and create another var `isHostInError := strings.Contains(*host.StatusInfo, mediaDisconnectionMessage)`
		
then the condition can be read more naturally as `return !(isStepRepliDisconnectionError && isHostInError)`

internal/bminventory/inventory.go
omertuc: Nevermind, I just realized this message match the exact message in the agent. Please write a comment explaining that this is an API and not simply an arbitrary message.

michaellevy101: no, This message does not match the message in the agent (the ErrorCode is matched)

internal/bminventory/inventory.go
michaellevy101: I added the doc and changed the method name.

internal/bminventory/inventory.go
michaellevy101: why?

omertuc: Oh, right, misunderstood. Then my [original comment](https://github.com/openshift/assisted-service/pull/2061/files#r657009527) still applies

omertuc:
Please update the PR description as requested.

> ## Checklist
> * [ ]  Title and description added to both, commit and PR.
> * [x]  Relevant issues have been associated (see [CONTRIBUTING](https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md?rgh-link-date=2021-06-22T11%3A25%3A05Z) guide)
> * [ ]  Reviewers have been listed
> * [x]  This change does not require a documentation update (docstring, `docs`, README, etc)
> * [ ]  Does this change include unit-tests (note that code changes require unit-tests)
> 
> ## Reviewers Checklist
> * [ ]  Are the title and description (in both PR and commit) meaningful and clear?
> * [x]  Is there a bug required (and linked) for this change?
> * [x]  Should this PR be backported?

PR #2067, opened 2021-06-22
OCPBUGSM-30214: add controller for AgentClusterInstall

danielerez:
# Description

Added AgentClusterInstallReconciler for watching the CRD
and set sync condition in case of a missing ClusterDeployment.

If the specified ClusterDeployment is missing the following message should be set for SpecSynced condition:
"The Spec could not be synced due to an input error: ClusterDeployment with name <name> in namespace <namespace> not found. Ensure that the CRD is already applied."

- [ ] Cloud
- [ ] None

# How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

* Apply an AgentClusterInstall CRD, without creating the correlated ClusterDeployment CRD.
* Status.Conditions should include the following condition:
```
     Message:               The Spec could not be synced due to an input error: ClusterDeployment with name <name> in namespace <namespace> not found. Ensure that the CRD is already applied.
     Reason:                InputError
     Status:                False
     Type:                  SpecSynced
```
* Other conditions should be in status Unknown

# Assignees

this PR directly to someone.

/cc @rollandf 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agentclusterinstall_controller.go
+++ b/internal/controller/controllers/agentclusterinstall_controller.go
@@ -1,0 +1,112 @@
+/*
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/openshift/assisted-service/internal/common"
+	hiveext "github.com/openshift/assisted-service/internal/controller/api/hiveextension/v1beta1"
+	logutil "github.com/openshift/assisted-service/pkg/log"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// AgentClusterInstallReconciler reconciles a AgentClusterInstall object
+type AgentClusterInstallReconciler struct {
+	client.Client
+	Log              logrus.FieldLogger
+	CRDEventsHandler CRDEventsHandler
+}
+
+// +kubebuilder:rbac:groups=extensions.hive.openshift.io,resources=agentclusterinstalls,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=extensions.hive.openshift.io,resources=agentclusterinstalls/status,verbs=get;update;patch
+
+func (r *AgentClusterInstallReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	ctx := addRequestIdIfNeeded(origCtx)
+	log := logutil.FromContext(ctx, r.Log).WithFields(
+		logrus.Fields{
+			"agent_cluster_install":           req.Name,
+			"agent_cluster_install_namespace": req.Namespace,
+		})
+
+	defer func() {
+		log.Info("AgentClusterInstall Reconcile ended")
+	}()
+
+	log.Info("AgentClusterInstall Reconcile started")
+
+	// Retrieve AgentClusterInstall
+	clusterInstall := &hiveext.AgentClusterInstall{}
+	if err := r.Get(ctx, req.NamespacedName, clusterInstall); err != nil {
+		log.WithError(err).Errorf("Failed to get resource %s", req.NamespacedName)
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// Retrieve ClusterDeployment
+	clusterDeploymentKubeKey := types.NamespacedName{
+		Namespace: clusterInstall.Namespace,
+		Name:      clusterInstall.Spec.ClusterDeploymentRef.Name,
+	}
+	clusterDeployment := &hivev1.ClusterDeployment{}
+	if err := r.Get(ctx, clusterDeploymentKubeKey, clusterDeployment); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			log.WithError(err).Error(fmt.Sprintf(
+				"failed to get ClusterDeployment with name '%s' in namespace '%s'",
+				clusterDeploymentKubeKey.Name, clusterDeploymentKubeKey.Namespace))
+			return ctrl.Result{RequeueAfter: defaultRequeueAfterOnError}, nil
+		}
+
+		apiErr := errors.New(fmt.Sprintf(
+			"ClusterDeployment with name '%s' in namespace '%s' not found. Ensure that the CRD is already applied.",
+			clusterDeploymentKubeKey.Name, clusterDeploymentKubeKey.Namespace))
+		err = common.NewApiError(http.StatusBadRequest, apiErr)
+		log.Error(err)
+
+		// Set conditions
+		clusterSpecSynced(clusterInstall, err)
+		setClusterConditionsUnknown(clusterInstall)
+
+		if updateErr := r.Status().Update(ctx, clusterInstall); updateErr != nil {
+			log.WithError(updateErr).Error("failed to update AgentClusterInstall Status")
+			return ctrl.Result{Requeue: true}, nil
+		}
+	}
+
+	log.Info("AgentClusterInstall Reconcile successfully retrieved the associated ClusterDeployment")
+	return ctrl.Result{}, nil
+}
+
+func (r *AgentClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&hiveext.AgentClusterInstall{}).
+		Watches(&source.Kind{Type: &hiveext.AgentClusterInstall{}},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Complete(r)
+}
--- a/subsystem/kubeapi_test.go
+++ b/subsystem/kubeapi_test.go
@@ -353,6 +353,16 @@ func checkAgentClusterInstallCondition(ctx context.Context, key types.Namespaced
 	}, "2m", "2s").Should(Equal(reason))
 }
 
+func checkAgentClusterInstallConditionConsistency(ctx context.Context, key types.NamespacedName, conditionType string, reason string) {
+	Consistently(func() string {
+		condition := controllers.FindStatusCondition(getAgentClusterInstallCRD(ctx, kubeClient, key).Status.Conditions, conditionType)
+		if condition != nil {
+			return condition.Reason
+		}
+		return ""
+	}, "15s", "5s").Should(Equal(reason))
+}
+
 func checkInfraEnvCondition(ctx context.Context, key types.NamespacedName, conditionType conditionsv1.ConditionType, message string) {
 	Eventually(func() string {
 		condition := conditionsv1.FindStatusCondition(getInfraEnvCRD(ctx, kubeClient, key).Status.Conditions, conditionType)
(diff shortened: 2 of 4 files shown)


filanov:
Please add a subsystem as well, it will help us to see if we have conflicts with cluster deployment controller.

internal/controller/controllers/agentclusterinstall_controller.go
```diff
@@ -0,0 +1,112 @@
...
+			err = common.NewApiError(http.StatusBadRequest, apiErr)
+			log.WithError(err).Error(errMsg)
+		} else {
+			errMsg = fmt.Sprintf("failed to get ClusterDeployment with name '%s' in namespace '%s'",
+				clusterDeploymentKubeKey.Name, clusterDeploymentKubeKey.Namespace)
+			log.WithError(err).Error(errMsg)
+		}
+
+		// Set conditions
+		clusterSpecSynced(clusterInstall, err)
```
filanov, line 91: I'm not sure if it's required to set sync status to error and the conditions to unknown in case of an internal error to get the resource. maybe you just need to requeue? 
@mhrivnak what do you think?

internal/controller/controllers/agentclusterinstall_controller.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *AgentClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&hiveext.AgentClusterInstall{}).
+		Watches(&source.Channel{Source: r.CRDEventsHandler.GetAgentUpdates()},
```
filanov, line 109: Not sure that this watch is required. 
This is usually notifications from the backend, not something that should affect the logic of this controller.
Unless i'm missing something?

nmagnezi (commented):
I wish to discuss the decision to add this controller:
From what I understand, the sole purpose (currently) of this controller is to take care of AgentClusterInstall conditions in an "orphan" AgentClusterInstall.

Do we need a dedicated controller for that?

The clusterDeployments controller already watches for AgentClusterInstall resources:

https://github.com/openshift/assisted-service/blob/568c24282c8b22588c1fb5fd6200812c4e05a993/internal/controller/controllers/clusterdeployments_controller.go#L1115

 and try to pair a watched resource to its referenced clusterDeployment:

https://github.com/openshift/assisted-service/blob/568c24282c8b22588c1fb5fd6200812c4e05a993/internal/controller/controllers/clusterdeployments_controller.go#L1088-L1109

Can't this logic be implemented in `mapClusterInstallToClusterDeployment` if it didn't find the matching clusterDeployment?

filanov:
> I wish to discuss the decision to add this controller:
> From what I understand, the sole purpose (currently) of this controller is to take care of AgentClusterInstall conditions in an "orphan" AgentClusterInstall.
> 
> Do we need a dedicated controller for that?
> 
> The clusterDeployments controller already watches for AgentClusterInstall resources:
> 
> https://github.com/openshift/assisted-service/blob/568c24282c8b22588c1fb5fd6200812c4e05a993/internal/controller/controllers/clusterdeployments_controller.go#L1115
> 
> and try to pair a watched resource to its referenced clusterDeployment:
> 
> https://github.com/openshift/assisted-service/blob/568c24282c8b22588c1fb5fd6200812c4e05a993/internal/controller/controllers/clusterdeployments_controller.go#L1088-L1109
> 
> Can't this logic be implemented in `mapClusterInstallToClusterDeployment` if it didn't find the matching clusterDeployment?

I'm not sure if map function should manipulate resources. but it does look like a nice trick. the only problem is handling internal errors in case of failure. 
@mhrivnak what do you think? is there another way to avoid creating a controller?

internal/controller/controllers/agentclusterinstall_controller.go
rollandf: I agree, watch on AgentClusterInstall for this is enough.

internal/controller/controllers/agentclusterinstall_controller.go
rollandf: So, only update the conditions if we get `IsNotFound`?
Makes sense to me.

danielerez: Sure, added an early return with requeue for other errors.

internal/controller/controllers/agentclusterinstall_controller.go
danielerez: Right, removed.

danielerez (commented):
> > Can't this logic be implemented in `mapClusterInstallToClusterDeployment` if it didn't find the matching clusterDeployment?
> 
> I'm not sure if map function should manipulate resources. but it does look like a nice trick. the only problem is handling internal errors in case of failure.
> @mhrivnak what do you think? is there another way to avoid creating a controller?

Yeah adding the new controller might be an overkill, but perhaps it would be useful for later on any way?
I.e. could we have other scenarios in which we want to watch the AgentClusterInstall resource specifically?
WDYT

nmagnezi:
> > > Can't this logic be implemented in `mapClusterInstallToClusterDeployment` if it didn't find the matching clusterDeployment?
> > 
> > 
> > I'm not sure if map function should manipulate resources. but it does look like a nice trick. the only problem is handling internal errors in case of failure.
> > @mhrivnak what do you think? is there another way to avoid creating a controller?
> 
> Yeah adding the new controller might be an overkill, but perhaps it would be useful for later on any way?
> I.e. could we have other scenarios in which we want to watch the AgentClusterInstall resource specifically?
> WDYT

We do a lot of things with AgentClusterInstall in clusterDeployments controller, notable examples would be the AgentClusterInstall finalizer, modifying AgentClusterInstall Conditions etc.

This is why I'm suggesting that this belongs to the clusterDeployments controller.

internal/controller/controllers/agentclusterinstall_controller.go
mhrivnak: Makes sense to me.

filanov:
> > > > Can't this logic be implemented in `mapClusterInstallToClusterDeployment` if it didn't find the matching clusterDeployment?
> > > 
> > > 
> > > I'm not sure if map function should manipulate resources. but it does look like a nice trick. the only problem is handling internal errors in case of failure.
> > > @mhrivnak what do you think? is there another way to avoid creating a controller?
> > 
> > 
> > Yeah adding the new controller might be an overkill, but perhaps it would be useful for later on any way?
> > I.e. could we have other scenarios in which we want to watch the AgentClusterInstall resource specifically?
> > WDYT
> 
> We do a lot of things with AgentClusterInstall in clusterDeployments controller, notable examples would be the AgentClusterInstall finalizer, modifying AgentClusterInstall Conditions etc.
> 
> This is why I'm suggesting that this belongs to the clusterDeployments controller.

Cluster deployment controller is watching for cluster deployment resource, the request that it gets search for a cluster deployment, the case that we are trying to resolve is when agent cluster install point to a none existing cluster deployment.
As far as i know reconcile function should get a request for a specific resource, you cannot identify resource type by name + namespace.

internal/controller/controllers/agentclusterinstall_controller.go
```diff
@@ -0,0 +1,105 @@
...
+			return ctrl.Result{Requeue: true}, nil
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *AgentClusterInstallReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&hiveext.AgentClusterInstall{}).
```
mhrivnak, line 103: Here I suggest adding option `builder.WithPredicates(predicate.GenerationChangedPredicate{})` so reconcile isn't called for every status update. We mainly just need this reconcile to run when the resource gets created or has a spec change.

subsystem/kubeapi_test.go
```diff
@@ -2274,6 +2274,23 @@ spec:
 			return err
 		}, "1m", "10s").Should(HaveOccurred())
 	})
+
+	It("deploy AgentClusterInstall with missing ClusterDeployment", func() {
```
filanov, line 2278: suggesting to add another step that will add cluster deployment to "fix" the issue and check that sync state is consistently true.

danielerez: done

internal/controller/controllers/agentclusterinstall_controller.go
danielerez: sure, done.

nmagnezi:
> > > > > Can't this logic be implemented in `mapClusterInstallToClusterDeployment` if it didn't find the matching clusterDeployment?
> > > > 
> > > > 
> > > > I'm not sure if map function should manipulate resources. but it does look like a nice trick. the only problem is handling internal errors in case of failure.
> > > > @mhrivnak what do you think? is there another way to avoid creating a controller?
> > > 
> > > 
> > > Yeah adding the new controller might be an overkill, but perhaps it would be useful for later on any way?
> > > I.e. could we have other scenarios in which we want to watch the AgentClusterInstall resource specifically?
> > > WDYT
> > 
> > 
> > We do a lot of things with AgentClusterInstall in clusterDeployments controller, notable examples would be the AgentClusterInstall finalizer, modifying AgentClusterInstall Conditions etc.
> > This is why I'm suggesting that this belongs to the clusterDeployments controller.
> 
> Cluster deployment controller is watching for cluster deployment resource, the request that it gets search for a cluster deployment, the case that we are trying to resolve is when agent cluster install point to a none existing cluster deployment.
> As far as i know reconcile function should get a request for a specific resource, you cannot identify resource type by name + namespace.

The idea I proposed was to handle this case in`mapClusterInstallToClusterDeployment`.
We can do that without creating a reconcile request for it.

mhrivnak:
> > Cluster deployment controller is watching for cluster deployment resource, the request that it gets search for a cluster deployment, the case that we are trying to resolve is when agent cluster install point to a none existing cluster deployment.
> > As far as i know reconcile function should get a request for a specific resource, you cannot identify resource type by name + namespace.
> 
> The idea I proposed was to handle this case in`mapClusterInstallToClusterDeployment`.
> We can do that without creating a reconcile request for it.

We should not change state from within a map handler function. There's no Context, no logger, and no way to return an error because the handler pattern is intended to only translate events into reconcile requests. Reconcilers are the intended place to make state changes, and I think it's fine to have this small additional Reconciler in order to facilitate specific status updates.

internal/controller/controllers/agentclusterinstall_controller.go
```diff
@@ -0,0 +1,129 @@
...
+						Name:      aci.Name,
+					},
+				},
+			}
+		},
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&hiveext.AgentClusterInstall{}).
+		Watches(&source.Kind{Type: &hiveext.AgentClusterInstall{}}, handler, predicates).
```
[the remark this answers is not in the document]
danielerez: Got it, done.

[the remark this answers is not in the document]
danielerez: Seems the 'For' call is still required. Getting "must provide an object for reconciliation" after dropping it. I guess it's verified here: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/builder/controller.go#L185

internal/controller/controllers/agentclusterinstall_controller.go
```diff
@@ -0,0 +1,129 @@
...
+// AgentClusterInstallReconciler reconciles a AgentClusterInstall object
+type AgentClusterInstallReconciler struct {
+	client.Client
+	Log              logrus.FieldLogger
+	CRDEventsHandler CRDEventsHandler
+}
+
+// +kubebuilder:rbac:groups=extensions.hive.openshift.io,resources=agentclusterinstalls,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=extensions.hive.openshift.io,resources=agentclusterinstalls/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=extensions.hive.openshift.io,resources=agentclusterinstalls/finalizers,verbs=update
```
nmagnezi, line 50: This controller doesn't do anything with finalizers currently.
As long as this is the case, I think this should be removed - to avoid conflicts with the CD controller handling of the ACI finalizer.

internal/controller/controllers/agentclusterinstall_controller.go
```diff
@@ -0,0 +1,129 @@
...
+		clusterSpecSynced(clusterInstall, err)
+		setClusterConditionsUnknown(clusterInstall)
+
+		if updateErr := r.Status().Update(ctx, clusterInstall); updateErr != nil {
+			log.WithError(updateErr).Error("failed to update AgentClusterInstall Status")
+			return ctrl.Result{Requeue: true}, nil
+		}
+	}
+
+	return ctrl.Result{}, nil
```
nmagnezi, line 103: If the controller was able to fetch a `clusterDeployment` for that `AgentClusterInstall`, it will print nothing and end the iteration.

It might make it unclear to the person who reads the log, what exactly happened here.
The user will see:
`AgentClusterInstall Reconcile started` 
immediately followed by
`AgentClusterInstall Reconcile ended`

I would suggest adding an info log saying that a clusterDeployment resource was detected for that `AgentClusterInstall` resource, so the rest of the handling will be done by the clusterDeployments controller.

subsystem/kubeapi_test.go
```diff
@@ -1954,7 +1964,7 @@ var _ = Describe("[kube-api]cluster installation", func() {
 				}
 			}
 			return true
-		}, "1m", "2s").Should(BeTrue())
+		}, "3m", "2s").Should(BeTrue())
```
nmagnezi, line 1967: why?

nmagnezi (changes requested):

internal/controller/controllers/agentclusterinstall_controller.go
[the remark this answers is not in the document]
filanov: Did you see it unit-test or subsystem test? at least for unit-tests 
`Watches(&source.Kind{Type: &hiveext.AgentClusterInstall{}}, &handler.EnqueueRequestForObject{}, predicates).` worked for me

filanov (approved):

PR #2112, opened 2021-06-28
Bug 1966586: 200 OK returned when setting invalid BaseDNSdomain via API

sagidayan:
# Description

Updated `ValidateDNSName`.                                                                 
The function was validating against `apps.<clusterName>.<baseDNSDomain>` .
This is why via the API `baseDNSDomain=example` is responding with 200. because it actually validates `apps.<cluster_name>.example`.

The update adds a validation on the baseDNS itself in addition to the validations we already have.

Updated subsytem tests to ensure validation

- [ ] Cloud
- [x] None

# How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

# Assignees

this PR directly to someone.

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/dns/dns.go
+++ b/internal/dns/dns.go
@@ -192,14 +192,14 @@ func (h *handler) GetDNSDomain(clusterName, baseDNSDomainName string) (*DNSDomai
 	}, nil
 }
 
-//ValidateDNSName checks if a combination of cluster name and base DNS domain
+// ValidateDNSName checks baseDNSDomainName and the combination of cluster name and base DNS domain
 // leaves enough room for automatically added domain names,
 // e.g. "alertmanager-main-openshift-monitoring.apps.test-infra-cluster-assisted-installer.example.com").
 // The max total length of a domain name is 255 bytes, including the dots. An individual label can be
 // up to 63 bytes. A single char may occupy more than one byte in Internationalized Domain Names (IDNs).
 func (h *handler) ValidateDNSName(clusterName, baseDNSDomainName string) error {
 	appsDomainNameSuffix := fmt.Sprintf(appsDomainNameFormat, clusterName, baseDNSDomainName)
-	if err := validations.ValidateDomainNameFormat(appsDomainNameSuffix); err != nil {
+	if err := validations.ValidateDomainNameFormat(baseDNSDomainName); err != nil {
 		return err
 	}
 	if len(appsDomainNameSuffix) > dnsDomainPrefixMaxLen {
--- a/subsystem/cluster_test.go
+++ b/subsystem/cluster_test.go
@@ -843,6 +843,79 @@ var _ = Describe("cluster install - DHCP", func() {
 	})
 })
 
+var _ = Describe("Validate BaseDNSDomain when creating a cluster", func() {
+	var (
+		ctx         = context.Background()
+		clusterCIDR = "10.128.0.0/14"
+		serviceCIDR = "172.30.0.0/16"
+	)
+	AfterEach(func() {
+		clearDB()
+	})
+	type DNSTest struct {
+		It            string
+		BaseDNSDomain string
+		ShouldThrow   bool
+	}
+	createClusterWithBaseDNS := func(baseDNS string) (*installer.RegisterClusterCreated, error) {
+		return userBMClient.Installer.RegisterCluster(ctx, &installer.RegisterClusterParams{
+			NewClusterParams: &models.ClusterCreateParams{
+				BaseDNSDomain:            baseDNS,
+				ClusterNetworkCidr:       &clusterCIDR,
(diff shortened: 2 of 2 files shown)


swagger.yaml
```diff
@@ -3811,6 +3811,7 @@ definitions:
       base_dns_domain:
         type: string
         description: Base domain of the cluster. All DNS records must be sub-domains of this base and include the cluster name.
+        pattern: '^([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,}$'
```
michaellevy101, line 3814: Where did you find this pattern?
is the error message indicative to the user?

michaellevy101 (commented):
What about kubeapi? (You can create a cluster from the kubeapi as well)
What about testing?

swagger.yaml
sagidayan: Pattern was found In the UI validation (https://github.com/openshift-assisted/assisted-ui-lib/blob/76d85794aa565e98fb3820df879baf17525fc9a2/src/components/ui/formik/validationSchemas.ts#L15). So it uses same validation.

Regarding the error message, It uses the swagger default message, same as any other patterned parameter

sagidayan:
@michaellevy101 
> What about kubeapi? (You can create a cluster from the kubeapi as well)

We have two epics that will tackle kubeapi validations.
 - MGMT-5247
 - MGMT-6677


> What about testing?
Added

subsystem/cluster_test.go
```diff
@@ -821,6 +821,55 @@ var _ = Describe("cluster install - DHCP", func() {
...
+				ClusterNetworkHostPrefix: 23,
+				Name:                     swag.String("test-cluster"),
+				OpenshiftVersion:         swag.String(openshiftVersion),
+				PullSecret:               swag.String(pullSecret),
+				ServiceNetworkCidr:       &serviceCIDR,
+				SSHPublicKey:             sshPublicKey,
+			},
+		})
+	}
+	It("Should thow en error with invalid BaseDNS 'example'", func() {
```
slaviered, line 847: 1. Spelling: "throw an error"
2. Use loop of tests instead a copy paste of the same test logic
3. Please use more descriptive test names. for example: "Illegal character in domain name", "too short domain name" etc.

subsystem/cluster_test.go
```diff
@@ -821,6 +821,55 @@ var _ = Describe("cluster install - DHCP", func() {
 	})
 })
 
+var _ = Describe("Create cluster with BaseDNS", func() {
```
slaviered, line 824: Use more descriptive name: such validation of baseDNS during create cluster

swagger.yaml
```diff
@@ -4101,6 +4103,7 @@ definitions:
...
```
slaviered, line 4106: Please consult Avishai if we do not need to extract base_dns_domain to the description part and use ref here

subsystem/cluster_test.go
sagidayan: 👍🏽 Thanks for the input!

subsystem/cluster_test.go
sagidayan: 👍🏽

swagger.yaml
sagidayan: We decided to remove the swagger changes altogether.
Please see description.

internal/dns/dns.go
```diff
@@ -192,13 +192,16 @@ func (h *handler) GetDNSDomain(clusterName, baseDNSDomainName string) (*DNSDomai
...
+// ValidateDNSName checks baseDNSDomainName and the combination of cluster name and base DNS domain
 // leaves enough room for automatically added domain names,
 // e.g. "alertmanager-main-openshift-monitoring.apps.test-infra-cluster-assisted-installer.example.com").
 // The max total length of a domain name is 255 bytes, including the dots. An individual label can be
 // up to 63 bytes. A single char may occupy more than one byte in Internationalized Domain Names (IDNs).
 func (h *handler) ValidateDNSName(clusterName, baseDNSDomainName string) error {
 	appsDomainNameSuffix := fmt.Sprintf(appsDomainNameFormat, clusterName, baseDNSDomainName)
+	if err := validations.ValidateDomainNameFormat(baseDNSDomainName); err != nil {
+		return err
+	}
```
ori-amizur, line 204: Is there a reason to validate **appsDomainNameFormat** now ?

sagidayan: You are right. The format of `appsDomainNameSuffix` should be valid if `clusterName` is valid and the `baseDNS` is valid.
The only thing to validate is `appsDomainNameSuffix`'s is length. 
Ill update.

PR #2113, opened 2021-06-28
MGMT-6768: Allow users to choose between SDN or OVN network types

oshercc:
# Description

give the user the option to choose a network type out of the options of OVS and SDN with the limitation on ipv6 has to use OVN.

- [x] Cloud
- [ ] None

# How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

# Assignees

this PR directly to someone.

/cc @ori-amizur 
/cc @avishayt 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -458,6 +458,7 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 			HTTPSProxy:               swag.StringValue(params.NewClusterParams.HTTPSProxy),
 			NoProxy:                  swag.StringValue(params.NewClusterParams.NoProxy),
 			VipDhcpAllocation:        params.NewClusterParams.VipDhcpAllocation,
+			NetworkType:              params.NewClusterParams.NetworkType,
 			UserManagedNetworking:    params.NewClusterParams.UserManagedNetworking,
 			AdditionalNtpSource:      swag.StringValue(params.NewClusterParams.AdditionalNtpSource),
 			MonitoredOperators:       monitoredOperators,
@@ -2035,6 +2036,10 @@ func (b *bareMetalInventory) updateNetworkParams(params installer.UpdateClusterP
 		machineCidr = ""
 		setMachineNetworkCIDRForUpdate(updates, machineCidr)
 	}
+	if params.ClusterUpdateParams.NetworkType != nil && params.ClusterUpdateParams.NetworkType != cluster.NetworkType {
+		updates["network_type"] = swag.StringValue(params.ClusterUpdateParams.NetworkType)
+	}
+
 	if !userManagedNetworking {
 		if vipDhcpAllocation {
 			err = b.updateDhcpNetworkParams(updates, params, log, &machineCidr)
--- a/internal/cluster/validation_id.go
+++ b/internal/cluster/validation_id.go
@@ -19,6 +19,7 @@ const (
 	IsMachineCidrEqualsToCalculatedCidr = ValidationID(models.ClusterValidationIDMachineCidrEqualsToCalculatedCidr)
 	IsApiVipDefined                     = ValidationID(models.ClusterValidationIDAPIVipDefined)
 	IsApiVipValid                       = ValidationID(models.ClusterValidationIDAPIVipValid)
+	isNetworkTypeValid                  = ValidationID(models.ClusterValidationIDNetworkTypeValid)
 	IsIngressVipDefined                 = ValidationID(models.ClusterValidationIDIngressVipDefined)
 	IsIngressVipValid                   = ValidationID(models.ClusterValidationIDIngressVipValid)
 	AllHostsAreReadyToInstall           = ValidationID(models.ClusterValidationIDAllHostsAreReadyToInstall)
@@ -33,8 +34,9 @@ const (
 
 func (v ValidationID) Category() (string, error) {
 	switch v {
-	case IsMachineCidrDefined, IsMachineCidrEqualsToCalculatedCidr, IsApiVipDefined, IsApiVipValid, IsIngressVipDefined, IsIngressVipValid,
-		isClusterCidrDefined, isServiceCidrDefined, noCidrOverlapping, networkPrefixValid, IsDNSDomainDefined, IsNtpServerConfigured:
+	case IsMachineCidrDefined, IsMachineCidrEqualsToCalculatedCidr, IsApiVipDefined, IsApiVipValid, IsIngressVipDefined,
+		IsIngressVipValid, isClusterCidrDefined, isServiceCidrDefined, noCidrOverlapping, networkPrefixValid,
+		IsDNSDomainDefined, IsNtpServerConfigured, isNetworkTypeValid:
 		return "network", nil
 	case AllHostsAreReadyToInstall, SufficientMastersCount:
 		return "hosts-data", nil
--- a/internal/cluster/validator.go
+++ b/internal/cluster/validator.go
@@ -113,7 +113,7 @@ func (v *clusterValidator) printIsMachineCidrDefined(context *clusterPreprocessC
 	case ValidationPending:
 		return "Hosts have not been discovered yet"
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -128,7 +128,7 @@ func (v *clusterValidator) printIsClusterCidrDefined(context *clusterPreprocessC
 	case ValidationSuccess:
 		return "The Cluster Network CIDR is defined."
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -143,7 +143,7 @@ func (v *clusterValidator) printIsServiceCidrDefined(context *clusterPreprocessC
 	case ValidationSuccess:
 		return "The Service Network CIDR is defined."
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -174,7 +174,7 @@ func (v *clusterValidator) printIsMachineCidrEqualsToCalculatedCidr(context *clu
 	case ValidationFailure:
 		return fmt.Sprintf("The Cluster Machine CIDR %s is different than the calculated CIDR %s.", context.cluster.MachineNetworkCidr, context.calculateCidr)
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -208,7 +208,7 @@ func (v *clusterValidator) printIsApiVipDefined(context *clusterPreprocessContex
 		}
 		return "The API virtual IP is defined."
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -222,6 +222,40 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
 	err := network.VerifyVip(c.cluster.Hosts, c.cluster.MachineNetworkCidr, c.cluster.APIVip, ApiVipName,
 		true, v.log)
 	return boolValue(err == nil)
+}
+
+func (v *clusterValidator) isNetworkTypeValid(c *clusterPreprocessContext) ValidationStatus {
+	validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN, models.ClusterCreateParamsNetworkTypeAutoAssign}
+	if !funk.ContainsString(validNetworkTypes, swag.StringValue(c.cluster.NetworkType)) && c.cluster.NetworkType != nil {
+		return ValidationFailure
+	}
+	if (network.IsIPv6CIDR(c.cluster.ClusterNetworkCidr) || network.IsIPv6CIDR(c.cluster.MachineNetworkCidr) || network.IsIPv6CIDR(c.cluster.ServiceNetworkCidr)) &&
+		swag.StringValue(c.cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes && c.cluster.NetworkType != nil {
+		return ValidationFailure
+	}
+	return ValidationSuccess
+}
+
+func (v *clusterValidator) printIsNetworkTypeValid(context *clusterPreprocessContext, status ValidationStatus) string {
+	switch status {
+	case ValidationSuccess:
+		if context.cluster.NetworkType == nil {
+			return "The cluster will use a default network type"
+		}
+		return "The cluster has a valid network type"
+	case ValidationFailure:
+		validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN, models.ClusterCreateParamsNetworkTypeAutoAssign}
+		if !funk.ContainsString(validNetworkTypes, swag.StringValue(context.cluster.NetworkType)) && context.cluster.NetworkType != nil {
+			return "The network type is not valid; the valid network types are OpenShiftSDN or OVNKubernetes"
+		}
+		if network.IsIPv6CIDR(context.cluster.ClusterNetworkCidr) && swag.StringValue(context.cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes {
+			return "The cluster is configured with IPv6 which is not supported by OpenShiftSDN; use OVNKubernetes instead"
+		} else {
+			return "Network type is invalid for an unknown reason"
+		}
+	default:
+		return fmt.Sprintf("Unexpected status %s", status)
+	}
 }
 
 func (v *clusterValidator) printIsApiVipValid(context *clusterPreprocessContext, status ValidationStatus) string {
@@ -239,7 +273,7 @@ func (v *clusterValidator) printIsApiVipValid(context *clusterPreprocessContext,
 	case ValidationFailure:
 		return fmt.Sprintf("%s %s does not belong to the Machine CIDR or is already in use.", ApiVipName, context.cluster.APIVip)
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -273,7 +307,7 @@ func (v *clusterValidator) printIsIngressVipDefined(context *clusterPreprocessCo
 		}
 		return "The Ingress virtual IP is defined."
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 func (v *clusterValidator) isIngressVipValid(c *clusterPreprocessContext) ValidationStatus {
@@ -386,7 +420,7 @@ func (v *clusterValidator) printSufficientMastersCount(context *clusterPreproces
 		return fmt.Sprintf("Clusters must have exactly %d dedicated masters. Please either add hosts, or disable the worker host",
 			common.MinMasterHostsNeededForInstallation)
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -415,7 +449,7 @@ func (v *clusterValidator) printAllHostsAreReadyToInstall(context *clusterPrepro
 	case ValidationFailure:
 		return "The cluster has hosts that are not ready to install."
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -430,7 +464,7 @@ func (v *clusterValidator) printIsDNSDomainDefined(context *clusterPreprocessCon
 	case ValidationSuccess:
 		return "The base domain is defined."
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -462,7 +496,7 @@ func (v *clusterValidator) printNoCidrsOverlapping(c *clusterPreprocessContext, 
 		}
 		return "At least one of the CIDRs (Machine Network, Cluster Network, Service Network) is undefined."
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
@@ -477,7 +511,7 @@ func (v *clusterValidator) printIsPullSecretSet(context *clusterPreprocessContex
 	case ValidationSuccess:
 		return "The pull secret is set."
 	default:
-		return fmt.Sprintf("Unexpected status %s.", status)
+		return fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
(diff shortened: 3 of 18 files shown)


filanov:
Is there a default network? what should be used for ipv4/ipv6?

oshercc:
> Is there a default network? what should be used for ipv4/ipv6?

* default type will be SDN (https://github.com/openshift/assisted-service/pull/2113/files#diff-f9d8c8224b172499b34cb44479e6e3606f19036989fa80449d085d1f3780e6eeR159)
* for ipv6 network type must be OVN, for ipv4 both can be used

internal/cluster/validator.go
```diff
@@ -224,6 +224,30 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+func (v *clusterValidator) isNetworkTypeValid(c *clusterPreprocessContext) ValidationStatus {
+	if network.IsIPv6CIDR(c.cluster.ClusterNetworkCidr) && swag.StringValue(c.cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes {
+		return ValidationFailure
+	}
+	return ValidationSuccess
+}
+
+func (v *clusterValidator) printIsNetworkTypeValid(context *clusterPreprocessContext, status ValidationStatus) string {
+	switch status {
+	case ValidationPending:
```
gamli75, line 236: we don't return ValidationPending in isNetworkTypeValid so we can remove this and the UI are not displaying pending option

swagger.yaml
```diff
@@ -4022,6 +4022,11 @@ definitions:
         description: Enable/disable hyperthreading on master nodes, worker nodes, or all nodes.
         enum: ['masters', 'workers', 'all', 'none']
         x-nullable: true
+      network_type:
```
gamli75, line 4025: we need to add this parameter also to the cluster model

swagger.yaml
```diff
@@ -5123,6 +5133,7 @@ definitions:
       - 'lso-requirements-satisfied'
       - 'ocs-requirements-satisfied'
       - 'cnv-requirements-satisfied'
+      - 'network-valid'
```
gamli75, line 5136: I think that we should call it "network-type-valid"

gamli75:
@avishayt please review the swagger changes 
I also think that we need to validate the network type in the updateCluster API - not sure we need it as another cluster validation - WDYT?

swagger.yaml
avishayt: Why have a validation and not just return 400 upon update if it's invalid?

avishayt:
> @avishayt please review the swagger changes
> I also think that we need to validate the network type in the updateCluster API - not sure we need it as another cluster validation - WDYT?

I didn't see this comment and commented the same thing :)
Yes I agree - should return 400 if updating to an invalid setting

swagger.yaml
oshercc: the networkType will depend on the network (ipv6 vs 4) so there is the option it will not be define when the user updates the cluster params, no ?

ori-amizur: In hive, when we define the network-type we don't know if it will be valid for the current family (ipv4/ipv6).  Therefore, we need validation for it.

internal/bminventory/inventory.go
```diff
@@ -458,6 +461,7 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 			HTTPSProxy:               swag.StringValue(params.NewClusterParams.HTTPSProxy),
 			NoProxy:                  swag.StringValue(params.NewClusterParams.NoProxy),
 			VipDhcpAllocation:        params.NewClusterParams.VipDhcpAllocation,
+			NetworkType:              swag.StringValue(params.NewClusterParams.NetworkType),
```
ori-amizur, line 464: We should validate that the provided network type is valid - either OpenshiftSDN , or OVNKubernetes

ori-amizur (commented):
What about UpdateCluster ? I think we should support it there too

internal/bminventory/inventory.go
```diff
@@ -2062,6 +2066,15 @@ func (b *bareMetalInventory) updateNetworkParams(params installer.UpdateClusterP
 		return common.NewApiError(http.StatusBadRequest, err)
 	}
 
+	if err = validations.ValidateNetworkTypeWithIPv6(params.ClusterUpdateParams.NetworkType, machineCidr); err != nil {
+		validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
+		if !funk.ContainsString(validNetworkTypes, swag.StringValue(params.ClusterUpdateParams.NetworkType)) {
```
avishayt, line 2071: The swagger defines an enum, doesn't the code generated from the swagger check this?

internal/bminventory/inventory.go
```diff
@@ -2062,6 +2066,15 @@ func (b *bareMetalInventory) updateNetworkParams(params installer.UpdateClusterP
...
+			err = errors.Errorf("NetworkType %s is not a type is not valid networkType", swag.StringValue(params.ClusterUpdateParams.NetworkType))
```
avishayt, line 2072:
```suggestion
			err = errors.Errorf("NetworkType %s is not a valid networkType", swag.StringValue(params.ClusterUpdateParams.NetworkType))
```

internal/cluster/validations/validations.go
```diff
@@ -378,6 +380,15 @@ func ValidateVipDHCPAllocationWithIPv6(vipDhcpAllocation bool, machineNetworkCID
 	return nil
 }
 
+//ValidateNetworkTypeWithIPv6 returns an error in case of networkType selected OVNKubernetes
+//being used with IPv6 machine network
+func ValidateNetworkTypeWithIPv6(networkType *string, machineNetworkCIDR string) error {
+	if network.IsIPv6CIDR(machineNetworkCIDR) && swag.StringValue(networkType) != models.ClusterNetworkTypeOVNKubernetes {
+		return errors.Errorf("ipv6 can only use OVNKubernetes network type, current networkType: %s", swag.StringValue(networkType))
```
avishayt, line 387:
```suggestion
		return errors.Errorf("IPv6 can only use OVNKubernetes network type, current networkType is %s", swag.StringValue(networkType))
```

internal/bminventory/inventory.go
avishayt, line 2071: I would expect to first check that the type itself is valid, then check if it's compatible with the machineCIDR

ori-amizur: Not just machineCIDR,but with serviceCIDR and clusterCIDR

ori-amizur: Is this also enforced in non-REST API?

oshercc: @ori-amizur can you please explain what you mean by non-REST API?

internal/bminventory/inventory.go
```diff
@@ -2062,6 +2066,15 @@ func (b *bareMetalInventory) updateNetworkParams(params installer.UpdateClusterP
 		return common.NewApiError(http.StatusBadRequest, err)
 	}
 
+	validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
```
gamli75, line 2069: Please move this validation to the network package - similar to network.VerifyClusterCIDRsNotOverlap and move the testing as well. 
@ori-amizur WDYT?

avishayt:
Looks OK to me.  @ori-amizur ?

oshercc:
@avishayt is that ok the default for ipv4 will be now OVN instead of SDN

swagger.yaml
avishayt: We know it's valid if it doesn't return 400 right?

avishayt:
> @avishayt is that ok the default for ipv4 will be now OVN instead of SDN

Why are we changing the default?

oshercc:
> > @avishayt is that ok the default for ipv4 will be now OVN instead of SDN
> 
> Why are we changing the default?

as I see it, the problem is we can't tell if the user changed the networkType 
now the networkType is generated while the install config is generated by the network configuration (IPv4/6) 
if we give the user the option to select the network type (even pre-knowing the network configuration) at what point can we know what networkType to select and if it was selected by the user or as default.

internal/bminventory/inventory.go
ori-amizur: Hive ?

internal/bminventory/inventory.go
ori-amizur: Seems OK

swagger.yaml
ori-amizur: Still, if it was defined during create cluster to be OpenshiftSDN, and later the discovered networks are IPv6, who will verify that we should not start installation this way ?

ori-amizur: So 400 will not be returned here

oshercc: so what about creating with a default (OVM) with no option for SDN?

internal/bminventory/inventory.go
oshercc: this ?
https://github.com/openshift/assisted-service/pull/2113/files#diff-72619caf255df33ccaa77f4a5ea4b9980c41cea8fbbbde96f8c62e9085282195R391

swagger.yaml
avishayt: @ori-amizur If the discovered networks are IPv6 then we'll automatically choose IPv6 CIDR?

ori-amizur: What other choice we have? There are no IPv4 networks.

internal/bminventory/inventory.go
ori-amizur: This is Register and the link above is for update

swagger.yaml
avishayt: I thought the user was choosing - I didn't realize the service was doing this automatically.
How does it work in [validations.ValidateVipDHCPAllocationWithIPv6](https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory.go#L2062)?

avishayt: Another point to keep in mind, Moran wants us as a default to use whatever OpenShift uses as a default in whatever version we're installing.  So we'll need to display to the user what type will be used.

oshercc: so @ori-amizur what do you think ? 
using a validation (and if yes can you maybe explain how does it work with  [validations.ValidateVipDHCPAllocationWithIPv6](https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory.go#L2062) or not giving an option to select a type when creating cluster?
or any alternative ?

ori-amizur: Maybe we can avoid the problem by checking the VIPs during cluster create.  If they are IPv6, then network-type must be OVN.
Same can be done during update cluster.

avishayt: Please rename to network-type-valid

internal/cluster/validator.go
```diff
@@ -224,6 +224,36 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+	if network.IsIPv6CIDR(c.cluster.ClusterNetworkCidr) && swag.StringValue(c.cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes {
+		return ValidationFailure
+	}
+	return ValidationSuccess
+}
+
+func (v *clusterValidator) printIsNetworkTypeValid(context *clusterPreprocessContext, status ValidationStatus) string {
+	switch status {
+	case ValidationSuccess:
+		return "The cluster has a sufficient network type."
```
avishayt, line 241:
```suggestion
		return "The cluster has a valid network type."
```

internal/cluster/validator.go
```diff
@@ -224,6 +224,36 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+	case ValidationFailure:
+		validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
+		if !funk.ContainsString(validNetworkTypes, swag.StringValue(context.cluster.NetworkType)) {
+			return "NetworkType is not valid, valid network types can be OpenShiftSDN or OVNKubernetes"
```
avishayt, line 245:
```suggestion
			return "The network type is not valid; the valid network types are OpenShiftSDN or OVNKubernetes"
```

Also consider using the validNetworkTypes variable in the string

internal/cluster/validator.go
```diff
@@ -224,6 +224,36 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+		}
+		if network.IsIPv6CIDR(context.cluster.ClusterNetworkCidr) && swag.StringValue(context.cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes {
+			return "ipv6 can only use OVNKubernetes network type."
```
avishayt, line 248:
```suggestion
			return "The cluster is configured with IPv6 which is not supported by OpenShiftSDN;  use OVNKubernetes instead."
```

internal/cluster/validator.go
```diff
@@ -224,6 +224,36 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+		} else {
+			return "Network type is undefined"
```
avishayt, line 250: What does that mean?

internal/cluster/validation_id.go
```diff
@@ -19,6 +19,7 @@ const (
 	IsMachineCidrEqualsToCalculatedCidr = ValidationID(models.ClusterValidationIDMachineCidrEqualsToCalculatedCidr)
 	IsApiVipDefined                     = ValidationID(models.ClusterValidationIDAPIVipDefined)
 	IsApiVipValid                       = ValidationID(models.ClusterValidationIDAPIVipValid)
+	isNetworkTypeValid                  = ValidationID(models.ClusterValidationIDNetworkValid)
```
avishayt, line 22:
```suggestion
	isNetworkTypeValid                  = ValidationID(models.ClusterValidationIDNetworkTypeValid)
```

internal/cluster/validator.go
oshercc: the validation failed for an unclear reason
changed "Network type is invalid", will that work?

avishayt:
Thanks @oshercc , looks good to me now.  Now we're changing behavior, so:
1. we should add this to the UI @jtomasek @mareklibra 
2. the controller needs to be adjusted @filanov

internal/bminventory/inventory.go
```diff
@@ -329,6 +329,9 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(_ context.Context,
 	if params.NewClusterParams.Hyperthreading == nil {
 		params.NewClusterParams.Hyperthreading = swag.String(models.ClusterHyperthreadingAll)
 	}
+	if params.NewClusterParams.NetworkType == nil {
+		params.NewClusterParams.NetworkType = swag.String(models.ClusterNetworkTypeOVNKubernetes)
+	}
```
avishayt, line 334: Can we keep the current default behavior?  Leave NetworkType as optional, and if it's not set when we generate the install config then we'll choose as before?

oshercc: I don't think so, it will be conflicted between the non-rest call if no type will be selected we can't validate it.

avishayt: What do you mean by non-rest call?  We can say that the validation succeeds automatically if no type is selected.

oshercc: NM, added a commit that will set default type in case a type is not selected

internal/cluster/validator.go
```diff
@@ -224,6 +224,39 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+		return ValidationFailure
+	}
+	return ValidationSuccess
+}
+
+func (v *clusterValidator) printIsNetworkTypeValid(context *clusterPreprocessContext, status ValidationStatus) string {
+	switch status {
+	case ValidationSuccess:
+		if context.cluster.NetworkType == nil {
+			return "The cluster will use a default validationType"
```
avishayt, line 242:
```suggestion
			return "The cluster will use a default network type"
```

internal/cluster/validator.go
```diff
@@ -224,6 +224,39 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+		}
+		return "The cluster has a valid network type."
+	case ValidationFailure:
+		validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
+		if !funk.ContainsString(validNetworkTypes, swag.StringValue(context.cluster.NetworkType)) && context.cluster.NetworkType != nil {
+			return "The network type is not valid; the valid network types are OpenShiftSDN or OVNKubernetes"
+		}
+		if network.IsIPv6CIDR(context.cluster.ClusterNetworkCidr) && swag.StringValue(context.cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes {
+			return "The cluster is configured with IPv6 which is not supported by OpenShiftSDN;  use OVNKubernetes instead."
```
avishayt, line 251:
```suggestion
			return "The cluster is configured with IPv6 which is not supported by OpenShiftSDN; use OVNKubernetes instead"
```

internal/cluster/validator.go
```diff
@@ -224,6 +224,39 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+		} else {
+			return "Network type is invalid"
+		}
+	default:
+		return fmt.Sprintf("Unexpected status %s.", status)
```
avishayt, line 256:
```suggestion
		return fmt.Sprintf("Unexpected status %s", status)
```

avishayt:
> Thanks @oshercc , looks good to me now. Now we're changing behavior, so:
> 
> 1. we should add this to the UI @jtomasek @mareklibra
> 2. the controller needs to be adjusted @filanov

Note that @oshercc changed the PR to keep the current behavior as the default.  So now we can choose to expose it in the UI and k8s API, but no changes are required at the moment.

internal/cluster/validator.go
```diff
@@ -224,6 +224,39 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+	return ValidationSuccess
+}
+
+func (v *clusterValidator) printIsNetworkTypeValid(context *clusterPreprocessContext, status ValidationStatus) string {
+	switch status {
+	case ValidationSuccess:
+		if context.cluster.NetworkType == nil {
+			return "The cluster will use a default network type"
+		}
+		return "The cluster has a valid network type."
```
avishayt, line 244:
```suggestion
		return "The cluster has a valid network type"
```

avishayt (approved):

internal/cluster/validator.go
```diff
@@ -224,6 +224,39 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
...
+		return "The cluster has a valid network type"
+	case ValidationFailure:
+		validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
+		if !funk.ContainsString(validNetworkTypes, swag.StringValue(context.cluster.NetworkType)) && context.cluster.NetworkType != nil {
+			return "The network type is not valid; the valid network types are OpenShiftSDN or OVNKubernetes"
+		}
+		if network.IsIPv6CIDR(context.cluster.ClusterNetworkCidr) && swag.StringValue(context.cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes {
+			return "The cluster is configured with IPv6 which is not supported by OpenShiftSDN; use OVNKubernetes instead"
+		} else {
+			return "Network type is invalid"
```
ori-amizur, line 253: How is this possible?

internal/cluster/validator.go
```diff
@@ -224,6 +224,39 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
 	return boolValue(err == nil)
 }
 
+func (v *clusterValidator) isNetworkTypeValid(c *clusterPreprocessContext) ValidationStatus {
+	validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
+	if !funk.ContainsString(validNetworkTypes, swag.StringValue(c.cluster.NetworkType)) && c.cluster.NetworkType != nil {
+		return ValidationFailure
+	}
+	if network.IsIPv6CIDR(c.cluster.ClusterNetworkCidr) && swag.StringValue(c.cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes {
```
ori-amizur, line 232: Other places we check also the service CIDR and cluster CIDR.  Also keep in mind none platform ->no machine CIDR

avishayt: @oshercc please fix

internal/cluster/validator.go
avishayt: It's shouldn't be possible, it's in case of an unknown reason for validation failure.  Maybe better to say "Network type is invalid for an unknown reason"?

oshercc: changed to  "Network type is invalid for an unknown reason"

internal/cluster/validator.go
```diff
@@ -224,6 +224,40 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
 	return boolValue(err == nil)
 }
 
+func (v *clusterValidator) isNetworkTypeValid(c *clusterPreprocessContext) ValidationStatus {
+	validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
+	if !funk.ContainsString(validNetworkTypes, swag.StringValue(c.cluster.NetworkType)) && c.cluster.NetworkType != nil {
```
ori-amizur, line 229: Isn't the situation that c.cluster.NetworkType == nil a valid value? If yes it has to be treated here.

ori-amizur (commented):
@avishayt what will be shown in the UI in case the network type is not set ?

avishayt:
> @avishayt what will be shown in the UI in case the network type is not set ?

I assumed empty.  Maybe we should we add a default value of auto-assign like in roles?

avishayt (approved):

internal/bminventory/inventory.go
```diff
@@ -2035,6 +2036,10 @@ func (b *bareMetalInventory) updateNetworkParams(params installer.UpdateClusterP
 		machineCidr = ""
 		setMachineNetworkCIDRForUpdate(updates, machineCidr)
 	}
+	if params.ClusterUpdateParams.NetworkType != nil && params.ClusterUpdateParams.NetworkType != cluster.NetworkType {
+		updates["network_type"] = swag.StringValue(params.ClusterUpdateParams.NetworkType)
+	}
+
```
slaviered, line 2042: Does this feature also need to be reported to feature_usage?

gamli75: good point - I think we should add it

oshercc: I see a separate ticket was added.

slaviered: better to add it here . it's a one liner and has to do with the feature itself

avishayt (approved):

internal/cluster/validator.go
```diff
@@ -224,6 +224,40 @@ func (v *clusterValidator) isApiVipValid(c *clusterPreprocessContext) Validation
 	return boolValue(err == nil)
 }
 
+func (v *clusterValidator) isNetworkTypeValid(c *clusterPreprocessContext) ValidationStatus {
+	validNetworkTypes := []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN, models.ClusterCreateParamsNetworkTypeAutoAssign}
+	if !funk.ContainsString(validNetworkTypes, swag.StringValue(c.cluster.NetworkType)) && c.cluster.NetworkType != nil {
+		return ValidationFailure
```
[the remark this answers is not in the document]
ori-amizur: Since we have a valid default pending is not suitable here.  Pending should be used when we are waiting for information such as inventory, or hosts were not discovered yet.

PR #2081, opened 2021-06-23
BUG 1970063: Don't force ca-bundle.crt when mirroring

djzager:
# Description

We shouldn't force users to provide a ca-bundle when mirroring content.

Potential Use Cases:
1. Mirror registry has an official CA signed certificate, so no need for additional certificate.
2. Mirrored content into a network-unrestricted environment. Need an image mapping via registries.conf but bundle is not required.

# What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

# How was this code tested?

Please, select one or more if needed:

- [ ] assisted-test-infra environment
- [x] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

I manually tested all of the relevant scenarios that I could think of with respect to the mirror-registry configuration:
1. [User specifies a mirror-registry configmap that doesn't exist](https://github.com/openshift/assisted-service/pull/2081#issuecomment-868933476) - working as expected
2. [User specifies a mirror-registry configmap that is missing the registries.conf](https://github.com/openshift/assisted-service/pull/2081#issuecomment-868933486) - working as expected
3. [User specifies a mirror-registry configmap that is missing the ca-bundle.crt key](https://github.com/openshift/assisted-service/pull/2081#issuecomment-868934713) - this is the new use-case we are trying to support...working as expected
4. [User specifies a mirror-registry configmap that contains both the registries.conf and ca-bundle.crt keys](https://github.com/openshift/assisted-service/pull/2081#issuecomment-868946740) - this is the expected use case...working as expected

# Assignees

Please, add one or two reviewers that could help review this PR. Use `/assign` if you want to assign
this PR directly to someone.

/cc @mkowalski 
/cc @yevgeny-shnaidman 

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] Reviewers have been listed
- [x] This change **DOES** require a documentation update (docstring, `docs`, README, etc) **AND** documentation is updated
- [x] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- [ ] Are the title and description (in both PR and commit) meaningful and clear?
- [ ] Is there a bug required (and linked) for this change?
- [ ] Should this PR be backported?

[...]

PR #2137, opened 2021-06-30
Bug 1965117: Keep assisted-service configmap mount updated

djzager:
# Assisted Pull Request

## Description

This PR hashes the generated, mirror registry, and unsupported configMap(s) and uses that information to restart the assisted-service deployment if necessary. The general rule is that, if it's a part of our supported configuration, then we will restart the assisted-service. This simply means that if there is a mismatch between the "unsupported" configmap and what is currently running with the assisted-service, we will log (and emit) a warning and move on...we will take no direct action.

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @flaper87
/cc @mkowalski 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -70,6 +70,9 @@ const (
 	defaultIngressCertCMNamespace string = "openshift-config-managed"
 
 	configmapAnnotation = "unsupported.agent-install.openshift.io/assisted-service-configmap"
+
+	assistedConfigHashAnnotation = "agent-install.openshift.io/config-hash"
+	mirrorConfigHashAnnotation   = "agent-install.openshift.io/mirror-hash"
 )
 
 // AgentServiceConfigReconciler reconciles a AgentServiceConfig object
@@ -672,6 +675,14 @@ func (r *AgentServiceConfigReconciler) newIngressCertCM(instance *aiv1beta1.Agen
 }
 
 func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (*appsv1.Deployment, controllerutil.MutateFn, error) {
+	var assistedConfigHash, mirrorConfigHash, userConfigHash string
+
+	// Get hash of generated assisted config
+	assistedConfigHash, err := r.getCMHash(ctx, types.NamespacedName{Name: serviceName, Namespace: r.Namespace})
+	if err != nil {
+		return nil, nil, err
+	}
+
 	envSecrets := []corev1.EnvVar{
 		// database
 		newSecretEnvVar("DB_HOST", "db.host", databaseName),
@@ -695,16 +706,17 @@ func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(ctx context.
 		},
 	}
 
-	// User is responsible for knowing to restart assisted-service
-	annotations := instance.ObjectMeta.GetAnnotations()
-	configmapName, ok := annotations[configmapAnnotation]
+	// User is responsible for:
+	// - knowing to restart assisted-service when specifying the configmap via annotation
+	// - removing the annotation when the configmap is deleted
+	userConfigName, ok := instance.ObjectMeta.GetAnnotations()[configmapAnnotation]
 	if ok {
-		log.Infof("ConfigMap %s from namespace %s being used to configure assisted-service deployment", configmapName, r.Namespace)
+		log.Infof("ConfigMap %s from namespace %s being used to configure assisted-service deployment", userConfigName, r.Namespace)
 		envFrom = append(envFrom, []corev1.EnvFromSource{
 			{
 				ConfigMapRef: &corev1.ConfigMapEnvSource{
 					LocalObjectReference: corev1.LocalObjectReference{
-						Name: configmapName,
+						Name: userConfigName,
 					},
 				},
 			},
@@ -829,6 +841,11 @@ func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(ctx context.
 			return nil, nil, err
 		}
 
+		mirrorConfigHash, err = checksumMap(cm.Data)
+		if err != nil {
+			return nil, nil, err
+		}
+
 		// require that the registries config key be specified before continuing
 		if _, ok := cm.Data[mirrorRegistryRefRegistryConfKey]; !ok {
 			err = fmt.Errorf("Mirror registry configmap %s missing key %s", instance.Spec.MirrorRegistryRef.Name, mirrorRegistryRefRegistryConfKey)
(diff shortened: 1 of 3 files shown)


djzager:
This PR relies on https://github.com/openshift/assisted-service/pull/2081

/hold

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -922,13 +950,34 @@ func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(log logrus.F
...
+		}
+		deployment.Spec.Template.Annotations[assistedConfigHashAnnotation] = assistedConfigHash
+		if mirrorConfigHash != "" {
+			deployment.Spec.Template.Annotations[mirrorConfigHashAnnotation] = mirrorConfigHash
+		}
+		if userConfigHash != "" {
+			runningUserConfigHash, ok := deployment.Spec.Template.Annotations[userConfigHashAnnotation]
+			if ok {
+				if runningUserConfigHash != userConfigHash {
+					msg := fmt.Sprintf("User provided and running configuration differ on assisted-service deployment. Remove the '%s' annotation from deployment to force restart", userConfigHashAnnotation)
```
mkowalski, line 966: The error message confuses me a bit... If user CM and running user CMs do not match shouldn't we propose user to just restart/redeploy so that his changes are applied ?

djzager: This feels like a downside to this approach. Because in order to detect a difference between the specified config and what the assisted-service is running, we need something like this hash. But now the only way to resolve it is to either:
1) Delete the assisted-service deployment, allowing the controller to recreate it with the correct annotations
2) Remove the specific hash annotation from the assisted-service deployment, the controller then can update the deployment

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -922,13 +950,34 @@ func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(log logrus.F
...
+					r.Recorder.Event(instance, "Warning", "UserConfigMismatch", msg)
+					log.Warning(msg)
+				}
+			}
+			deployment.Spec.Template.Annotations[userConfigHashAnnotation] = userConfigHash
```
djzager, line 971:
```suggestion
			if ok {
				if runningUserConfigHash != userConfigHash {
					msg := fmt.Sprintf("User provided and running configuration differ on assisted-service deployment. Remove the '%s' annotation from deployment to force restart", userConfigHashAnnotation)
					r.Recorder.Event(instance, "Warning", "UserConfigMismatch", msg)
					log.Warning(msg)
				}
			} else { // only add the hash if the annotation doesn't exist
				deployment.Spec.Template.Annotations[userConfigHashAnnotation] = userConfigHash
			}
```

internal/controller/controllers/agentserviceconfig_controller.go
djzager: I now see a [bug in how I wrote this section](https://github.com/openshift/assisted-service/pull/2137/files#r661585822) but the suggestion I linked is how I envision it working in my comment :arrow_up:

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -717,16 +707,31 @@ func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(log logrus.F
...
+		// when the applied config differs from the running config
+		userConfigHash, err = r.getCMHash(ctx, types.NamespacedName{Name: userConfigName, Namespace: r.Namespace})
+		if err != nil {
+			// We will take no explicit action when the user provided configMap
+			// is not found. But we will log a warning and emit a 'Warning' event
+			if !errors.IsNotFound(err) {
+				return nil, nil, err
+			}
+			msg := fmt.Sprintf("A ConfigMap was specified on the agent via annotation '%s', but that ConfigMap was not found", configmapAnnotation)
+			r.Recorder.Event(instance, "Warning", "UserConfigMapNotFound", msg)
```
mhrivnak, line 726: This seems reasonable. The next step is to figure out how to communicate the resolution. If the user fixes this by creating the configmap or fixing the reference, we want to give them some sort of confirmation that this warning no longer applies, even though it still may show up in the list of events.

Would it be possible to emit a "Normal" event when such a configmap is first identified?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -922,13 +950,34 @@ func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(log logrus.F
...
+		deployment.Spec.Template.Annotations[assistedConfigHashAnnotation] = assistedConfigHash
+		if mirrorConfigHash != "" {
+			deployment.Spec.Template.Annotations[mirrorConfigHashAnnotation] = mirrorConfigHash
+		}
+		if userConfigHash != "" {
+			runningUserConfigHash, ok := deployment.Spec.Template.Annotations[userConfigHashAnnotation]
+			if ok {
+				if runningUserConfigHash != userConfigHash {
+					msg := fmt.Sprintf("User provided and running configuration differ on assisted-service deployment. Remove the '%s' annotation from deployment to force restart", userConfigHashAnnotation)
+					r.Recorder.Event(instance, "Warning", "UserConfigMismatch", msg)
```
mhrivnak, line 967: If we're adding a hash of the configmap to the pod spec template, won't that automatically cause a new rollout? Or have we not fully implemented that?

djzager: Yeah. This will cause a rollout, however we already do that when we update the env because the user supplied the "unsupported" configmap (https://github.com/openshift/assisted-service/pull/2137/files#diff-c619bb36181b5e1dd76fb1fcc1f54068b16a8761f77106052fcbf5e2f5f9d356R730-R737) quoting so it doesn't get lost:

```
	userConfigName, ok := instance.ObjectMeta.GetAnnotations()[configmapAnnotation]
	if ok {
		log.Infof("ConfigMap %s from namespace %s being used to configure assisted-service deployment", userConfigName, r.Namespace)
		...code omitted for space...
		envFrom = append(envFrom, []corev1.EnvFromSource{
			{
				ConfigMapRef: &corev1.ConfigMapEnvSource{
					LocalObjectReference: corev1.LocalObjectReference{
						Name: userConfigName,
					},
				},
			},
		}...,
		)
	}
```

Let me know if you see a fault in my line of thinking here.

mhrivnak: If you modify a Deployment and change anything about the pod template, the deployment controller will automatically roll out new pods to replace the old pods. It doesn't matter what the change was. Putting a hash in an annotation on the pod is a technique for using that behavior to our advantage. All we have to do is keep the hash current, and the Deployment controller will do a roll-out any time it changes.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -967,6 +994,15 @@ func (r *AgentServiceConfigReconciler) getOpenshiftVersions(log logrus.FieldLogg
 	return string(encodedVersions)
 }
 
+func (r *AgentServiceConfigReconciler) getCMHash(ctx context.Context, namespacedName types.NamespacedName) (string, error) {
+	cm := &corev1.ConfigMap{}
+	if err := r.Get(ctx, namespacedName, cm); err != nil {
+		r.Log.Error(err, "Failed to get configmap", "NamespacedName", namespacedName)
+		return "", err
+	}
+	return checksumMap(cm.Data)
```
flaper87, line 1008: What do you think (for safety measures) about logging errors coming from `checksumMap` and ignore them?

Maybe I'm being too paranoid but the scenario I would like to prevent is that some unhandled encoding issue (I realize that everything is b64 in this function) or some other internal problem, could cause this function to fail and, therefore, all the steps after it.

flaper87 (commented):
looking good, hope to see #2081 go in so we can rebase this one and complete the review flow.

flaper87:
/hold cancel

Sending this through now since we gotta get this in and backported asap. @djzager is out today

PR #2144, opened 2021-07-01
NO-ISSUE: enhancement proposal to avoid agent reboots

mhrivnak:
When an InfraEnv's ISO gets regenerated, particularly when that is due
to a NMStateConfig being added or modified, only those agents affected
should be rebooted.

### DIFF
--- a/docs/enhancements/bmac-reboot-hosts-on-iso-change.md
+++ b/docs/enhancements/bmac-reboot-hosts-on-iso-change.md
@@ -1,0 +1,221 @@
+---
+title: bmac-reboot-hosts-on-iso-change
+authors:
+  - "@mhrivnak"
+  - "@avishayt"
+  - "@flaper87"
+creation-date: 2021-06-30
+last-updated: 2021-07-01
+---
+
+# BMAC reboots only those hosts impacted by an ISO change
+
+## Summary
+
+When an InfraEnv's ISO gets regenerated based on a change such as adding or
+changing an NMStateConfig resource, there may be only a subset of that
+InfraEnv's Agents that are affected by the change. Rather than reboot all of
+them with the newly-generated ISO, only those hosts that are affected by the
+new or modified NMStateConfig should be rebooted.
+
+It is assumed that in all cases, Agents that have started installing will not
+be restarted due to an ISO change.
+
+## Motivation
+
+If there are 500 NMStateConfigs and 500 Agents all related to the same
+InfraEnv, and then someone adds the 501st NMStateConfig, we do not want to
+reboot all 500 Agents. Rebooting is an expensive and time-consuming operation.
+We need a way to determine which Agents or hosts should be rebooted in response
+to an NMStateConfig.
+
+### Goals
+
+* Avoid rebooting un-provisioned BareMetalHosts when it is not needed.
+* Leave the door open for other platforms (include None/boot-it-yourself) to also avoid host reboots that are not needed.
+
+### Non-Goals
+
+* 
+
+## Proposal
+
+The assisted-service will track whether a particular Agent was booted using an
+old InfraEnv or NMStateConfig. It will ensure that any Agent that was booted
+using obsolete artifacts will have:
+* a label indicating that it was booted from obsolete artifacts
+* an array in its Status of references to the resources that are obsolete
+
+### assisted-service InfraEnv controller
+
+The InfraEnv ISO URL changes when the ISO contents change. This makes it
+obvious to an API user that they have booted a host from an obsolete ISO,
+because the URLs won't match.
+
+When assisted-service creates an ISO, it will capture and include in the ISO:
+* the name, namespace, UID, and generation of the InfraEnv
+* the name, namespace, UID, and generation of each NMStateConfig
+
+At runtime, those resource identifiers for the InfraEnv and the NMStateConfig
+it used will be passed by the agent to assisted-service and added to the
+Agent's Status.
+
+### assisted-service Agent controller
+
+When the Agent resource gets created, its status should include the
+above-described reference details for:
+* The InfraEnv
+* The NMStateConfig it utilized during its boot
+
+Upon comparing those references to the resources that currently exist, if any
+of the referenced resources have changed, then the controller will add a label
+to the Agent resource indicating such. It will also add a reference to each
+obsolete resource to an array in the Agent's Status.
+
+### Bare Metal Agent Controller
+
+The Bare Metal Agent Controller will watch for Agents to have that label.
+
+If the Agent's NMStateConfig is obsolete, then it will reboot the corresponding
+BareMetalHost with the latest ISO URL.
+
+Else if the Agent's InfraEnv is obsolete, and the Agent does not have an associated
+NMStateConfig, then the BMAC will reboot the host only if finds a new
+NMStateConfig that matches the Agent.
+
+The BMAC will also watch for the InfraEnv URL to change, and if there are any
+BareMetalHosts that were booted with an obsolete URL but do not yet have an
+Agent resource, those BareMetalHosts will be rebooted with the latest ISO URL.
+
+ +---------------------------+
+ |                           |
+ |   Agent has the label?    |
+ |                           |
+ +----+-----------+----------+
+      |           |
+      v           v
+ +----+--+    +---+---+      +-------------+
+ |       |    |       |      |             |
+ |  Yes  |    |  No   +----->+  No action  |
+ |       |    |       |      |             |
+ +---+---+    +-------+      +-------------+
+     |
+     v
+ +---+----------------+
+ |                    |
+ |  Agent Status      |
+ |  cites an obsolete |
+ |  NMStateConfig?    |
+ |                    |
+ +---+------------+---+
+     |            |
+     v            v
+ +---+---+    +---+---+
+ |       |    |       |
+ |  Yes  |    |  No   |
+ |       |    |       |
+ +---+---+    +---+---+
+     |            |
+     |            v
+     |         +--+-------------------------+
+     |         |                            |
+     |         |  Is there a NMStateConfig  |
+     |         |  that matches the Agent?   |
+     |         |                            |
+     |         +-----------+------------+---+
+     |                     |            |
+     |                     v            v
++----v-----+           +---+---+    +---+---+     +-------------+
+|          |           |       |    |       |     |             |
+|  Reboot  +<----------+  Yes  |    |  No   +---->+  No action  |
+|          |           |       |    |       |     |             |
++----------+           +-------+    +-------+     +-------------+
+
+
+### User Stories
+
+#### Story 1
+
+As a user of Zero Touch Provisioning, when I add and modify NMStateConfig
+resources, unaffected hosts will not reboot. Only those hosts directly impacted
+will be rebooted with a current ISO URL.
+
+#### Story 2
+
+As a boot-it-yourself user, I will manually or with automation watch for Agents
+to have the obsolete artifact label and restart them with a fresh ISO.
+
+### Implementation Details/Notes/Constraints [optional]
+
+* Any Agent that has already started provisioning will be exempt from rebooting.
+* Similar to current behavior, a new ISO URL will not be used until a quiet
+  period has elapsed (currently 1 minute). This ensures that if multiple
+  changes are happening in serial, for example many NMStateConfig resources are
+  getting created or modified, that there will not be continuous churn of
+  rebooting hosts with each change.
+
+#### Label Algorithm
+
+Given:
+* The Agent's Status includes a reference to the InfraEnv resource that was used to create its ISO.
+* The Agent's Status includes an optional reference to a NMStateConfig resource that was used to create the network config the agent used.
+
+IF:
+
+* The existing InfraEnv generation or UID don't match the corresponding values reported in the Agent's Status.
+
+OR
+
+* The existing NMStateConfig generation or UID don't match the corresponding values reported in the Agent's Status, or the NMStateConfig does not exist.
+
+THEN
+
+* Ensure that a label exists on the Agent signifying that it was booted using obsolete artifacts.
+* Ensure that the obsolete InfraEnv and/or NMStateConfig are listed in an array of obsolete resource references in the Agent status.
+
+ELSE
+
+* Ensure that a label **does not exist** on the Agent signifying that it was booted using obsolete artifacts.
+* Ensure that the Agent Status does not have an array of references to obsolete resources.
+
+### Risks and Mitigations
+
+
+## Design Details [optional]
+
+
+### Open Questions
+
+* What should the label's key and value be?
+* How exactly would the resource references get embedded into the ISO, and how
+  would the Agent communicate those back to assisted-service?
+
+### UI Impact
+
+
+### Test Plan
+
+Test Cases:
+* Create InfraEnv, add NMStateConfigs one by one.
+* Create InfraEnv, add invalid NMStateConfig, fix NMStateConfig
+* Update NMStateConfig for existing Agent
+* Have multiple NMStateConfigs for single Agent
+* Remove NMStateConfig for existing Agent
+* Add NMStateConfig after Agent is already up
+* Create InfraEnv and NMStateConfig, wait for the Agent to be up, then update proxy in InfraEnv
+
+## Drawbacks
+
+
+## Alternatives
+
+Each NMStateConfig resource could have a unique ISO URL. The InfraEnv would
+provide a base URL, and the NMStateConfig could be identified by query
+parameter or similar. This way, each host would get a unique URL and a unique
+ISO. When a particular NMStateConfig gets added or modified, it would
+implicitly only affect any BMH with that NMStateConfig's specific URL.
+
+Then BMAC would have to determine the correct NMStateConfig for a BareMetalHost
+in advance and assign the correct URL to the BMH.
+
+This option is more complicated, especially from an API perspective, than the proposal.


docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,131 @@
...
+The assisted-service will track whether a particular Agent was booted using an
+old InfraEnv or NMStateConfig. It will ensure that any Agent that was booted
+using obsolete artifacts will have:
+* a label indicating that it was booted from obsolete artifacts
+* an array in its Status of references to the resources that are obsolete
+
+### assisted-service InfraEnv controller
+
+The InfraEnv ISO URL will change when the ISO contents change. This will make
+it obvious to an API user that they have booted a host from an obsolete ISO,
```
ronniel1, line 48: On one hand, it will make it obvious if they have an ISO with old content, but in the case of a change that just added nmstate, all 500 users will think that they booted from an obsolete ISO, which is not the case

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,131 @@
...
+
+### assisted-service Agent controller
+
+When the Agent resource gets created, its status should include the
+above-described reference details for:
+* The InfraEnv
+* The NMStateConfig it utilized during its boot
+
+If those details indicate that any of the referenced resources have changed,
+then the controller will add a label to the Agent resource indicating such. It
```
ronniel1, line 69: Maybe the label should be a counter or timestamp. This way we will not have any races or issues between setting and clearing these labels

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Risks and Mitigations
+
+
+## Design Details [optional]
+
+
+### Open Questions
+
+* Should the BMAC delete an Agent after it reboots the corresponding host?
```
ronniel1, line 106: For late binding it must not delete it.
Besides, the agent is a resource of the Assisted-Service, not the BMAC, so I don't think it makes sense for BMAC to delete it

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Test Plan
+
+
+## Drawbacks
+
+
+## Alternatives
+
+Each NMStateConfig resource could have a unique ISO URL. The InfraEnv would
```
ronniel1, line 122: Actually, with streaming and Ironic boot, that might be a good solution.
The reason for having a single ISO is mainly for BIY scenarios, I think

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,131 @@
...
+provide a base URL, and the NMStateConfig could be identified by query
+parameter or similar. This way, each host would get a unique URL and a unique
+ISO. When a particular NMStateConfig gets added or modified, it would
+implicitly only affect any BMH with that NMStateConfig's specific URL.
+
+Then BMAC would have to determine the correct NMStateConfig for a BareMetalHost
```
ronniel1, line 128: BMAC should not care about the correct NMStateConfig. It should create a URL that embeds the ID of the host (BMC MAC?) .
The Assisted-Service, while streaming this URL should generate the appropriate ISO for that host ID

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,131 @@
...
+authors:
+  - "@mhrivnak"
+  - "@avishayt"
+creation-date: 2021-06-30
+last-updated: 2021-06-30
+---
+
+# BMAC reboots only those hosts impacted by an ISO change
+
+## Summary
```
avishayt, line 12: We should mention that this only relates to Agents that have not begun installation.  Even today any Agent that is installing/installed/error will not be rebooted.

omertuc:
Due to the ISO being served off of AI / Ironic cache dynamically, and due to the time consuming nature of booting a host (it might get delayed for whatever reasons like BIOS, networking, etc) - wouldn't it be racey to have hosts still in booting stage run off of older ISOs that are no longer served by AI/cache (because they have been replaced by newer revisions)? 

I'm not familiar with the exact timing / behavior of ironic cache or AI but I assume they don't retain ISO files forever. 

Those ISOs might not be there by the time the host needs them. Rebooting the host as we do now prevents / hides such problems. I wonder if getting rid of these reboots might uncover them. 

I might be wrong though, just something to think about

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,131 @@
...
+* a label indicating that it was booted from obsolete artifacts
+* an array in its Status of references to the resources that are obsolete
+
+### assisted-service InfraEnv controller
+
+The InfraEnv ISO URL will change when the ISO contents change. This will make
+it obvious to an API user that they have booted a host from an obsolete ISO,
+because the URLs won't match. A side benefit of ensuring that the ISO URL
+uniquely identifies the resource's contents, is that pull-through caches will
+have an easy time understanding when an file is no longer relevant.
```
avishayt, line 51: It should be clarified that this exists today

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
avishayt: I agree.  What's the motivation for having BMAC delete the Agent?

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: update to say: the url does change today when the ISO contents change.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
avishayt: We'll stay with the same Agent so no reason to delete it

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,132 @@
...
+
+When assisted-service creates an ISO, it will capture and include in the ISO:
+* the name, namespace, UID, and generation of the InfraEnv
+* the name, namespace, UID, and generation of each NMStateConfig
+
+At runtime, those resource identifiers for the InfraEnv and the NMStateConfig
+it used will be passed by the agent to assisted-service and compared against
```
filanov, line 59: Maybe need to specify that it will happen only on registration right? 
What will be compared? does the service need to search for a matching nmstate config?

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,132 @@
...
+the current resource generations to determine if they are obsolete.
+
+### assisted-service Agent controller
+
+When the Agent resource gets created, its status should include the
+above-described reference details for:
+* The InfraEnv
+* The NMStateConfig it utilized during its boot
```
filanov, line 67: Is that a new logic that need to be added to the agent? when it boots it will just send the inventory no?
How will it know what is the relevant nmstate cr?

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,132 @@
...
+If those details indicate that any of the referenced resources have changed,
+then the controller will add a label to the Agent resource indicating such. It
+will also add a reference to each resource to an array in the Agent's Status.
+
+### Bare Metal Agent Controller
+
+The Bare Metal Agent Controller will watch for Agents to have that label and
+reboot the corresponding BareMetalHost with the latest ISO URL.
+
+The BMAC will also watch for the InfraEnv URL to change, and if there are any
```
filanov, line 78: feels like it can delay boot time if the user adding nmstate configurations manually.
for example it will take 10 min to boot the machine then it may boot only after user finished to configure all his nmstate configurations.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,132 @@
...
+BareMetalHosts that were booted with an obsolete URL but do not yet have an
+Agent resource, those BareMetalHosts will be rebooted with the latest ISO URL.
+
+### User Stories
```
filanov, line 82: i think that we need to make sure that we cover all those cases:
* Create infra env, add nmstate config one by one.
* Create infra env, add invalid nmstate config, fix nmstate config
* Update nmstate config for existing agent
* Have multiple nmstate configs for single  agent
* Remove nmstate config for existing agent
* Add nmstate config after agent is already up
* Create infra env and nmstate, agent is up, update proxy in infra env

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: Apparently the URL already changes with every ISO change. Given that, this proposal is about helping users (human or machine) determine if that change impacts a particular agent.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: I'm adding a section below under implementation details that describes the algorithm for applying and removing the label. Perhaps you can point out where there is a potential race condition.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: I agree that's worth exploring. That should be a different proposal though.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,132 @@
...
+### assisted-service InfraEnv controller
+
+The InfraEnv ISO URL will change when the ISO contents change. This will make
+it obvious to an API user that they have booted a host from an obsolete ISO,
+because the URLs won't match. A side benefit of ensuring that the ISO URL
+uniquely identifies the resource's contents, is that pull-through caches will
+have an easy time understanding when an file is no longer relevant.
+
+When assisted-service creates an ISO, it will capture and include in the ISO:
+* the name, namespace, UID, and generation of the InfraEnv
```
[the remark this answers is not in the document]
mhrivnak: See the [`ObjectMeta` docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#objectmeta-v1-meta) and the `generation` field.

"A sequence number representing a specific generation of the desired state. Populated by the system. Read-only."

It's commonly used to determine if a resource's Spec has changed.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,132 @@
...
+* the name, namespace, UID, and generation of each NMStateConfig
+
+At runtime, those resource identifiers for the InfraEnv and the NMStateConfig
```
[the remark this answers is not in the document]
mhrivnak: It's just explaining how the references will be used, and thus why the controller needs to add them.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: Does it matter if it only happens on registration?

I'm adding a section below with the specific algorithm for adding and removing the label.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: Yes this requires new logic. Today the ISO includes data from many NMStateConfigs, and the Agent at runtime picks the one relevant to it (by matching a MAC address). We'll need to enhance that task so that the data also includes a reference to the k8s resource.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: Yes, that's an inefficiency that this proposal accepts.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: Thanks. I'm adding these below as test cases.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
filanov: adding this label will require going over all the nmstates and see if something changed not?
If user update nmstate and now it need to move to a different host or any other update
it may affect performance no?

mhrivnak: The agent will have in its status a reference to one specific NMStateConfig that it used. Then when reconciling the Agent, we only have to determine if that single resource changed or was removed.

Perhaps there's an edge case in what you're raising though. If a host is able to boot without a NMStateConfig and it creates an Agent resource, but then an NMStateConfig gets created for it, BMAC should recognize that and reboot the host. I'll add detail below to the BMAC behavior, and see what you think.

That should be a rare situation. Had the NMStateConfig resource been created at any time before the Agent resource appeared, then the host would have been restarted. 

Even if/when the controller does need to iterate all of the nmstateconfig resources, they're all in RAM anyway, so it shouldn't be a significant hit.

filanov: i'm not concerned about executing a list, i'm thinking about that for every nmstate we need to unmarshal it and understand it before we create a link between agnet and nmstate.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,221 @@
...
+ +---+---+    +---+---+
+ |       |    |       |
+ |  Yes  |    |  No   |
+ |       |    |       |
+ +---+---+    +---+---+
+     |            |
+     |            v
+     |         +--+-------------------------+
+     |         |                            |
+     |         |  Is there a NMStateConfig  |
```
filanov, line 122: why does nmstate need to search for match isn't it something that the agent will do?

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
```diff
@@ -0,0 +1,221 @@
...
+NMStateConfig, then the BMAC will reboot the host only if finds a new
+NMStateConfig that matches the Agent.
+
+The BMAC will also watch for the InfraEnv URL to change, and if there are any
+BareMetalHosts that were booted with an obsolete URL but do not yet have an
+Agent resource, those BareMetalHosts will be rebooted with the latest ISO URL.
+
+ +---------------------------+
+ |                           |
+ |   Agent has the label?    |
```
filanov, line 92: just to see the full picture consider adding `No agent resource`

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: Assuming we are matching by MAC address, there is no unmarshalling required for [that field](https://github.com/openshift/assisted-service/blob/913aba6/internal/controller/api/v1beta1/nmstate_config_types.go#L52).

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
mhrivnak: 1. A host boots, registers, and an Agent resource gets created.
2. Time passes.
3. Someone creates a new NMStateConfig resource matching that host.

In that scenario, the Agent is not going to look for new NMStateConfig in the API. The only way to apply the new NMStateConfig will be to reboot that host with a new ISO.

filanov: So it require host to match only creation, that mean that when someone delete nmstate then it will require reboot as well.
it's related to my comments from https://github.com/openshift/assisted-service/pull/2144#discussion_r662490993 
Maybe i misunderstand something

mhrivnak: Can you increase the verbosity of your concern? Perhaps describe a specific scenario step-by-step? I'm not sure what you're getting at.

filanov: sorry you can resolve this comment, i read it again and now it's clear.

docs/enhancements/bmac-reboot-hosts-on-iso-change.md
filanov: Is there a case to support only layer 3 configuration?

mhrivnak: What do you mean exactly? The basic use case is to identify a layer 2 device by MAC address and then apply layer 3 configuration to it.

There is some layer 2 configuration we can do, such as bonding, but that's optional.

filanov: In https://docs.google.com/document/d/1PEOfM3-f_JUtR0rjZ74W_j3_8HGbKDrgrU3yVUTdhZQ/edit#heading=h.9cdnoxyarh8q
 there is a section related to `Add L3 connectivity check` i wasn't sure if it mean that we don't use l2 there. maybe it's not relevant

mhrivnak: Checking L3 connectivity would just be a validation. I suspect that's based on the prior cases where we've seen a host configured without a default route.

flaper87:
@mhrivnak Are you planning to submit a new revision to this enhancement? Wondering what the next steps are

mhrivnak:
> @mhrivnak Are you planning to submit a new revision to this enhancement? Wondering what the next steps are

@flaper87  I think @filanov asked that we review the testing section and make sure the various cases are accounted for. Do you have a little time to do that? I can push a change if you want to send me a patch or just a new version of that section.

filanov:
> > @mhrivnak Are you planning to submit a new revision to this enhancement? Wondering what the next steps are
> 
> @flaper87 I think @filanov asked that we review the testing section and make sure the various cases are accounted for. Do you have a little time to do that? I can push a change if you want to send me a patch or just a new version of that section.

You added the test section that i suggested,  if no one have nothing to add to it then i think that we can approve this enhancement

flaper87:
> You added the test section that i suggested, if no one have nothing to add to it then i think that we can approve this enhancement

Thanks for confirming this...

@mhrivnak I think the only thing left to do would be to amend the commit messages to make CI happy.

mhrivnak:
> @mhrivnak I think the only thing left to do would be to amend the commit messages to make CI happy.

What's the actual problem? I see its complaint, but my commit message does start with `NO-ISSUE`. Do my two additional commits also have to use the same? Or is squashing a requirement?

filanov (approved):

PR #2159, opened 2021-07-04
NO-ISSUE: Enable Day 2 cluster in kube api

rollandf:
## Description
Enable Day 2 cluster in kube api

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @nmagnezi 
/cc @danielerez 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -381,7 +381,7 @@ _run_subsystem_test:
 	$(MAKE) _test TEST_SCENARIO=subsystem TIMEOUT=120m TEST="$(or $(TEST),./subsystem/...)"
 
 enable-kube-api-for-subsystem: $(BUILD_FOLDER)
-	$(MAKE) deploy-service-requirements AUTH_TYPE=local ENABLE_KUBE_API=true ENABLE_KUBE_API_DAY2=false
+	$(MAKE) deploy-service-requirements AUTH_TYPE=local ENABLE_KUBE_API=true ENABLE_KUBE_API_DAY2=true
 	$(call restart_service_pods)
 	$(MAKE) wait-for-service
 
--- a/docs/hive-integration/README.md
+++ b/docs/hive-integration/README.md
@@ -135,6 +135,14 @@ Installed	The installation is in progress: Waiting for control plane
 
 Once the cluster is installed, the ClusterDeployment is set to Installed and secrets for kubeconfig and credentials are created and referenced in the AgentClusterInstall.
 
+## Day 2 worker
+
+In case of none SNO deployment, after that the cluster is installed, the original cluster is deleted and a Day 2 cluster is created instead in the Assisted Service database.
+
+Additional nodes can be added by booting from the new generated ISO. Each additional host will start installation once the Agent is Approved and the Host is in known state.
+
+Note that the user needs to approved the additional nodes in the installed cluster.
+
 ## Bare Metal Operator Integration
 
 In case that the Bare Metal Operator is installed, the Baremetal Agent Controller will sync between the Agent CR and the matching BareMetalHost CR:
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -389,6 +389,7 @@ func (r *AgentServiceConfigReconciler) newAssistedCM(log logrus.FieldLogger, ins
 			"CHECK_CLUSTER_VERSION":       "True",
 			"CREATE_S3_BUCKET":            "False",
 			"ENABLE_KUBE_API":             "True",
+			"ENABLE_KUBE_API_DAY2":        "True",
 			"ENABLE_SINGLE_NODE_DNSMASQ":  "True",
 			"IPV6_SUPPORT":                "True",
 			"JWKS_URL":                    "https://api.openshift.com/.well-known/jwks.json",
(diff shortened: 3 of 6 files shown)


filanov:
@avishayt @ravidbro @yevgeny-shnaidman currently once cluster is deleted from the DB we delete the agents in this cluster. 
What will happen in late binding?

ravidbro:
I see two options:
1. return them to the pool.
2. block deletion of cluster until the user will bind them to the pool first

filanov:
> I see two options:
> 
> 1. return them to the pool.
> 2. block deletion of cluster until the user will bind them to the pool first

Then in this case we cannot delete the day1 cluster, maybe instead of creating a new cluster we should just change installed cluster into a day2 cluster ?
Otherwise we will need to hold both day1 and day2.

yevgeny-shnaidman:
> > I see two options:
> > 
> > 1. return them to the pool.
> > 2. block deletion of cluster until the user will bind them to the pool first
> 
> Then in this case we cannot delete the day1 cluster, maybe instead of creating a new cluster we should just change installed cluster into a day2 cluster ?
> Otherwise we will need to hold both day1 and day2.

Maybe we can delete day1 cluster without doing anything with the agents: we do have connections between InfraEnv and installed agents, so in case we want to return the installed agents to the pool, we will be able to find them and move them as needed. Maybe we will need to handle  a special case where the currently attached cluster does not exists, but this is just a minor issue

filanov:
> > > I see two options:
> > > 
> > > 1. return them to the pool.
> > > 2. block deletion of cluster until the user will bind them to the pool first
> > 
> > 
> > Then in this case we cannot delete the day1 cluster, maybe instead of creating a new cluster we should just change installed cluster into a day2 cluster ?
> > Otherwise we will need to hold both day1 and day2.
> 
> Maybe we can delete day1 cluster without doing anything with the agents: we do have connections between InfraEnv and installed agents, so in case we want to return the installed agents to the pool, we will be able to find them and move them as needed. Maybe we will need to handle a special case where the currently attached cluster does not exists, but this is just a minor issue

sounds a bit hacky to me. thinking about the cloud it will require the user to follow the same pattern.
User install a cluster and want to add hosts from the pool to it, he will need to have both clusters that point to same installed cluster. IMO keeping one entity is clearer

rollandf:
/hold
Waiting for decision regarding late binding behavior

filanov:
> /hold
> Waiting for decision regarding late binding behavior

@rollandf I don't think that you need to hold this PR, there is some time before late binding is implemented and we need to handle adding remote workers to SNO as well so it will be probably with a day2 cluster as well.
For now this change LGTM

filanov (approved):

yevgeny-shnaidman:
> > > > I see two options:
> > > > 
> > > > 1. return them to the pool.
> > > > 2. block deletion of cluster until the user will bind them to the pool first
> > > 
> > > 
> > > Then in this case we cannot delete the day1 cluster, maybe instead of creating a new cluster we should just change installed cluster into a day2 cluster ?
> > > Otherwise we will need to hold both day1 and day2.
> > 
> > 
> > Maybe we can delete day1 cluster without doing anything with the agents: we do have connections between InfraEnv and installed agents, so in case we want to return the installed agents to the pool, we will be able to find them and move them as needed. Maybe we will need to handle a special case where the currently attached cluster does not exists, but this is just a minor issue
> 
> sounds a bit hacky to me. thinking about the cloud it will require the user to follow the same pattern.
> User install a cluster and want to add hosts from the pool to it, he will need to have both clusters that point to same installed cluster. IMO keeping one entity is clearer

it means that you will have to keep day1 cluster indefinitely, even if the user does not want to add any hosts in day2.

filanov:
> > > > > I see two options:
> > > > > 
> > > > > 1. return them to the pool.
> > > > > 2. block deletion of cluster until the user will bind them to the pool first
> > > > 
> > > > 
> > > > Then in this case we cannot delete the day1 cluster, maybe instead of creating a new cluster we should just change installed cluster into a day2 cluster ?
> > > > Otherwise we will need to hold both day1 and day2.
> > > 
> > > 
> > > Maybe we can delete day1 cluster without doing anything with the agents: we do have connections between InfraEnv and installed agents, so in case we want to return the installed agents to the pool, we will be able to find them and move them as needed. Maybe we will need to handle a special case where the currently attached cluster does not exists, but this is just a minor issue
> > 
> > 
> > sounds a bit hacky to me. thinking about the cloud it will require the user to follow the same pattern.
> > User install a cluster and want to add hosts from the pool to it, he will need to have both clusters that point to same installed cluster. IMO keeping one entity is clearer
> 
> it means that you will have to keep day1 cluster indefinitely, even if the user does not want to add any hosts in day2.

If you are working with a pool then yes. otherwise you will not know when to return a host to cluster. 
In hive when you delete cluster deployment it is expected that the manged cluster get destroyed.

ravidbro:
maybe when you delete a cluster...
we will look per host to the infra-env it belongs to, if the infra_env has the optional cluster_id set to the one we deleted - we delete the agent.
if the cluster_id of the associated infra_env is null we the agent just reset its cluster_id and getting back to the pool.

PR #2168, opened 2021-07-06
MGMT-7098: Add installation_failed_at field for monitoring failed ins…

oshercc:
…tallations

# Assisted Pull Request

## Description

this will help in two eras:
1. Give the option to locate clusters that have failed in the past and simplify the process of finding clusters that need to be examined (triaged)
2. Help monitor general activity in the service and help find bugs and issuse. 

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @YuviGold 
/cc @avishayt 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/common.go
+++ b/internal/cluster/common.go
@@ -55,6 +55,9 @@ func updateClusterStatus(ctx context.Context, log logrus.FieldLogger, db *gorm.D
 		if funk.ContainsString(installationCompletedStatuses, swag.StringValue(&newStatus)) {
 			extra = append(extra, "install_completed_at", now)
 		}
+		if swag.StringValue(&newStatus) == models.ClusterStatusError {
+			extra = append(extra, "install_failed_at", now)
+		}
 	}
 
 	if cluster, err = UpdateCluster(log, db, clusterId, srcStatus, extra...); err != nil ||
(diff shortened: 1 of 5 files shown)


avishayt:
Why not use install_completed_at?

oshercc:
> Why not use install_completed_at?
install_completed_at is for all installations completed (installed, cancel, and error)
we need to signify all the failed installations so we can know what clusters had failed for inspecting it.
because the job that opens the ticket is run periodically there are cases the state already passed from error when it's sampled and we can't know if it's a new error if it even happen.

avishayt:
> > Why not use install_completed_at?
> install_completed_at is for all installations completed (installed, cancel, and error)

It should also be set for failed installations

> we need to signify all the failed installations so we can know what clusters had failed for inspecting it.
> because the job that opens the ticket is run periodically there are cases the state already passed from error when it's sampled and we can't know if it's a new error if it even happen.

So this field won't be reset?

This sounds like a hack to help with collecting data about clusters that got to 'error' and then reset, is that right?

oshercc:
> It should also be set for failed installations
yes
> So this field won't be reset?
no
> This sounds like a hack to help with collecting data about clusters that got to 'error' and then reset, is that right?
I  would not  describe it as a hack 
but yes

YuviGold:
Decided to move to a different approach.

PR #2170, opened 2021-07-06
OCPBUGSM-30766: Removing "Unauthorized access to cluster *" log.

ybettan:
# Assisted Pull Request

## Description

This log is printed in one of 2 cases:

    * A user tries to access a non-existing cluster
    * A user tries to access a cluster that he doesn't own

The UI is getting a list of subscriptions from AMS which contains all the
clusters in the user's organization, then for each cluster it queries AI
in order to get some additional information, therefore it may:

    * Try to access a GC cluster
    * Try to access a cluster that someone else in the organization owns

Until we have an organization model in AI aligned with OCM we should
remove this log, otherwise, we will be spammed.

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

On the new code, I didn't get this log.

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/auth/rhsso_authenticator.go
+++ b/pkg/auth/rhsso_authenticator.go
@@ -312,7 +312,6 @@ func (a *RHSSOAuthenticator) CreateAuthenticator() func(name, in string, authent
 				return true, nil, common.NewApiError(http.StatusInternalServerError, err)
 			}
 			if !ownedBy {
-				log.Errorf("Unauthorized access to cluster %s by a user other than the owner\n", clusterID)
 				return true, nil, common.NewApiError(http.StatusNotFound, errors.New("Cluster Not Found"))
 			}
 


ybettan:
/hold
Waiting for manual tests results.

ybettan:
/unhold
Updated the test outuput in the PR template.

ybettan:
/hold
Will add unit-tests

danielerez:
I think that this logic actually makes sense for a user, as we don't want to expose any internal data for an unauthorized user.
I.e. if we remove that validation, an unauthorized user could reveal the existence of any cluster, which could be considered as a security breach.
Maybe we could just add a warning log instead for this case, as an indication for the issue.

slaviered:
+1 to what @danielerez said. The practice is not to expose the cluster IDsby returning something else. This code was added at a request to not allow users to query other user's clusters and it is important. If there is a lot of such messages it indicates a deeper problem with the interaction between kube-api and rest api

ybettan:
@danielerez @slaviered  This PR doesn't change the behavior you described, the user still gets an error indicating that there is no such cluster, the only difference is that it returns `500` instead of `404` and I will fix it.

```
$ curl -s $(minikube service assisted-service -n assisted-installer --url)/api/assisted-install/v1/clusters/93a682c5-5c5e-43d2-9c55-3205af6a8681 -H "Authorization: bearer $(ocm token)"
{"code":"500","href":"","id":500,"kind":"Error","reason":"record not found"}
```

In addition to printing the wrong log in case an authorized user try to access a non-existing cluster, it doesn't print the log when an unauthorized user try to access a cluster that it doesn't own.

pkg/auth/rhsso_authenticator.go
```diff
@@ -312,7 +312,6 @@ func (a *RHSSOAuthenticator) CreateAuthenticator() func(name, in string, authent
 				return true, nil, common.NewApiError(http.StatusInternalServerError, err)
 			}
 			if !ownedBy {
-				log.Errorf("Unauthorized access to cluster %s by a user other than the owner\n", clusterID)
```
danielerez, line 315: Maybe worth just reducing it to a debug log, and add a comment that it might flood the logs otherwise.
Also, a warning log in line 274 might be helpful.

avishayt (approved):

pkg/auth/rhsso_authenticator.go
ybettan: > Also, a warning log in line 274 might be helpful

This will just move the problem to somewhere else since we currently not differentiating between access to non-existing clusters and unauthorized access.

> Maybe worth just reducing it to a debug log

But then we will just flood the debug logs, how does that solve the problem.

Please see @avishayt 's comment https://coreos.slack.com/archives/CUPJTHQ5P/p1625663070173400?thread_ts=1625642128.124500&cid=CUPJTHQ5P.

@danielerez

danielerez: ok got it, so we would just eliminate the log for now to mitigate the issue.

avishayt (approved):

PR #2175, opened 2021-07-07
MGMT-7104: Add new API for updating VMWare credentials in the assisted service

eliorerz:
Update API for supporting multiple installation platforms.
Add to cluster properties a platform structure that contains the available platforms (Vsphere, Baremetal).
Note that `user_managed_networking` effects differently on different Platforms:

- BM: None platfrom
- Vsphere: UPI or IPI

Note: On vsphere platform, If API user don't want to set his vCenter credentials he can set the platform type to `vsphere` and not set the `vsphere_platform` structure. At the current implementation user can't set vsphere credentials partially

Currently this PR is just for the API review and without any validations.

https://issues.redhat.com/browse/MGMT-7104

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees
/cc @filanov 
/cc @avishayt 
/cc @tsorya 
/cc @gamli75 

## Reviewers Checklist


swagger.yaml
```diff
@@ -3153,6 +3153,48 @@ paths:
...
+    properties:
+      provisioningNetwork:
+        type: string
+        enum: [Disabled, Enabled]
+
+  vsphere_platform:
+    type: object
+    description: Vsphere platform specific configuration upon which to perform the installation
+    required:
+      - username
```
tsorya, line 3168: why required? didn't we say it is not must?

swagger.yaml
```diff
@@ -3153,6 +3153,48 @@ paths:
...
+      - password
```
tsorya, line 3169: all those params are required to make it work? or they are required only n day2?

eliorerz: For day1 I don't think we need any  vsphere credentials

swagger.yaml
eliorerz: Required for day2 operations to work

eliorerz: I added the `type` under platform, if type is set and no details (`vsphere_platform`) is given it will act as vsphere platform without credentials. Those fields are mandatory when the `vsphere` platform is selected and the `vsphere_platform` is initialized.

swagger.yaml
```diff
@@ -3153,6 +3153,48 @@ paths:
...
 
 definitions:
+  baremetal_platform:
+    description: Baremetal platform specific configuration upon which to perform the installation
+    type: object
+    properties:
+      provisioningNetwork:
+        type: string
+        enum: [Disabled, Enabled]
+
```
avishayt, line 3163: Why is this needed?  We don't support setting this property.

swagger.yaml
```diff
@@ -3153,6 +3153,48 @@ paths:
...
+      - vCenter
+      - cluster
+    properties:
+      username:
+        type: string
+        description: The user name to use to connect to the vCenter instance with. This user must have at least the roles and privileges that are required for static or dynamic persistent volume provisioning in vSphere.
+      password_set:
+        type: boolean
+        description: True if the vCenter password has been set.
+        readOnly: true
```
avishayt, line 3180: We can remove this for now.  I'm not sure if anyone will complain.

swagger.yaml
eliorerz: @avishayt Do we not want the API to support setting all platform properties?
And if not, completely remove the `baremetal_platform` structure?

avishayt: We want to expose only the properties that we want to support.  Yes, remove the `baremetal_platform` structure.

swagger.yaml
avishayt: If it can be installed successfully without passing any parameters then this is the best way to go (either pass everything or nothing).

tsorya: @avishayt yep, but i think it will be very helpful to allow to provide all params except user/password
This is will allow to make day2 operation much easier

avishayt: @tsorya Why?  What's the difference?  It's possible to update the user/password but not the other information?

tsorya: User password is copied to secret and all other info should be updated in some other place
@eliorerz can you please check where we should update vspere details? 
My point it that if we can minimize user day2 actions by given him to set all params without credentials it will allow us to reduce his day2 operations and make process more friendly

tsorya: In any case we will need to push some junk data there as it is required by installer? @eliorerz ?

avishayt: Is the vmware team OK with this?  This seems kind of hacky.

avishayt: @eliorerz ?

eliorerz: We decided not to put placeholders at first and so the `install_config.yaml` platform (if properties are not set) will look something like this:
```
paltform:
  vsphere:
    apiVIP: 2.2.2.2
    ingressVIP: 1.1.1.1
```

avishayt:
API looks good

avishayt:
Don't forget regexs on each of the vsphere properties (can be a separate PR)

internal/bminventory/inventory.go
```diff
@@ -1925,6 +1930,28 @@ func (b *bareMetalInventory) updateDhcpNetworkParams(updates map[string]interfac
 	}
 	return nil
 }
+func (b *bareMetalInventory) updatePlatformParams(params installer.UpdateClusterParams, cluster *common.Cluster, updates map[string]interface{}) {
+	platform := params.ClusterUpdateParams.Platform
+	if platform == nil || platform.Type == "" {
```
filanov, line 1935: How can user unset a profile?

eliorerz: You will always have a platform.  Platform is nil when it is not part of the `UpdateClusterParams`.
For replacing the cluster just patch `{ "platform": { "type": "baremetal" } }`

filanov: please make sure that when you define vsphere then switch to baremetal that you don't get the old vscphere parameterse.

internal/bminventory/inventory.go
```diff
@@ -1925,6 +1930,28 @@ func (b *bareMetalInventory) updateDhcpNetworkParams(updates map[string]interfac
 	}
 	return nil
 }
+func (b *bareMetalInventory) updatePlatformParams(params installer.UpdateClusterParams, cluster *common.Cluster, updates map[string]interface{}) {
+	platform := params.ClusterUpdateParams.Platform
+	if platform.Type == "" {
```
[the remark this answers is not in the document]
eliorerz: This is the Platform we get from the user (and not as part of the cluster).

internal/bminventory/inventory.go
```diff
@@ -340,6 +340,10 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 
 	id := strfmt.UUID(uuid.New().String())
 	url := installer.GetClusterURL{ClusterID: id}
+	platform := models.Platform{
+		Type:    models.PlatformTypeBaremetal,
```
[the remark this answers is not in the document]
eliorerz: It's auto generated from swagger
`platform_type` is the enum and `baremetal` is the enum value

internal/installcfg/installcfg.go
```diff
@@ -271,6 +272,41 @@ func (i *installConfigBuilder) setImageContentSources(cfg *InstallerConfigBareme
...
+			Username:         *vspherePlatform.Username,
+			Password:         *vspherePlatform.Password,
+			Datacenter:       *vspherePlatform.Datacenter,
+			DefaultDatastore: *vspherePlatform.DefaultDatastore,
+			Network:          *vspherePlatform.Network,
+			Cluster:          *vspherePlatform.Cluster,
+			APIVIP:           cluster.APIVip,
+			IngressVIP:       cluster.IngressVip,
+		}
+		if len(vspherePlatform.Folder) > 0 {
```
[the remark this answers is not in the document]
eliorerz: it changed to 
``` 
if vspherePlatform.Folder != nil {
	p.Folder = *vspherePlatform.Folder
}
``` 
But we need it because Folder can be nil, and `p.Folder` is string, we have to verify `vspherePlatform.Folder` before assign it

internal/bminventory/inventory.go
```diff
@@ -1827,6 +1832,10 @@ func (b *bareMetalInventory) updateClusterInternal(ctx context.Context, params i
 		host.FreeAddresses = ""
 	}
 
+	//if cluster.Platform.Vsphere != nil && cluster.Platform.Vsphere.VCenter == nil{
```
filanov, line 1835: please remove it not used.

internal/bminventory/inventory.go
```diff
@@ -340,6 +340,10 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 
 	id := strfmt.UUID(uuid.New().String())
 	url := installer.GetClusterURL{ClusterID: id}
+	platform := models.Platform{
+		Type:    models.PlatformTypeBaremetal,
+		Vsphere: nil,
```
filanov, line 345: default is nil you don't have to define it specifically.

eliorerz:
@avishayt @tsorya @filanov 
I had to remove the `required` tag from the `vsphere_platform` properties due to assisted-service-client failure.
The `vsphere_platform` object is not mandatory so its properties were set to null without omitempty tag (due to the required tag) but the client is validating that all required properties are not null (so it's failing).

mhrivnak:
> Note that user_managed_networking effects differently on different Platforms:
> 
> BM: None platfrom
> Vsphere: UPI or IPI

Practically speaking, what do you mean here by UPI vs IPI?

gamli75 (approved):

PR #2180, opened 2021-07-07
NO-ISSUE: Remove -u flag from go get commands to use fixed versions

mkowalski:
This is a manual backport of https://github.com/openshift/assisted-service/pull/2174.

/cc @YuviGold


YuviGold:
Pushing with force as all jobs failing on this.

PR #2183, opened 2021-07-07
MGMT-7136: Add metric report to all installation complete scenarios

slaviered:
# Assisted Pull Request

## Description

Currently only complete after finalizing is measured by metrics.
This code add metric events to all branches of installation, including
ones that are prematurely ended with error

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @gamli75 
/cc @YuviGold 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -645,11 +645,19 @@ func (m *Manager) VerifyClusterUpdatability(c *common.Cluster) (err error) {
 func (m *Manager) CancelInstallation(ctx context.Context, c *common.Cluster, reason string, db *gorm.DB) *common.ApiErrorResponse {
 	eventSeverity := models.EventSeverityInfo
 	eventInfo := "Cancelled cluster installation"
+	lastState := newStateCluster(c)
+	installationStates := []string{
+		models.ClusterStatusPreparingForInstallation, models.ClusterStatusInstalling, models.ClusterStatusFinalizing}
 	defer func() {
 		m.eventsHandler.AddEvent(ctx, *c.ID, nil, eventSeverity, eventInfo, time.Now())
+		//metrics for cancel as final state are calculated only when the transition to cancel was made
+		//from one of the installing states
+		if funk.Contains(installationStates, lastState.srcState) {
+			m.metricAPI.ClusterInstallationFinished(ctx, models.ClusterStatusCancelled, lastState.srcState, c.OpenshiftVersion, *c.ID, c.EmailDomain, c.InstallStartedAt)
+		}
 	}()
 
-	err := m.sm.Run(TransitionTypeCancelInstallation, newStateCluster(c), &TransitionArgsCancelInstallation{
+	err := m.sm.Run(TransitionTypeCancelInstallation, lastState, &TransitionArgsCancelInstallation{
 		ctx:    ctx,
 		reason: reason,
 		db:     db,
@@ -659,7 +667,6 @@ func (m *Manager) CancelInstallation(ctx context.Context, c *common.Cluster, rea
 		eventInfo = fmt.Sprintf("Failed to cancel installation: %s", err.Error())
 		return common.NewApiError(http.StatusConflict, err)
 	}
-	m.metricAPI.ClusterInstallationFinished(ctx, "canceled", c.OpenshiftVersion, *c.ID, c.EmailDomain, c.InstallStartedAt)
 	return nil
 }
 
@@ -1064,6 +1071,11 @@ func (m *Manager) CompleteInstallation(ctx context.Context, db *gorm.DB,
 	severity := models.EventSeverityInfo
 	eventMsg := fmt.Sprintf("Successfully finished installing cluster %s", cluster.Name)
 
+	defer func() {
+		m.metricAPI.ClusterInstallationFinished(ctx, result, models.ClusterStatusFinalizing, cluster.OpenshiftVersion,
+			*cluster.ID, cluster.EmailDomain, cluster.InstallStartedAt)
+	}()
+
 	if successfullyFinished {
 		destStatus = models.ClusterStatusInstalled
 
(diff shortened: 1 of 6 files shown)


internal/cluster/cluster_test.go
```diff
@@ -243,7 +243,7 @@ var _ = Describe("TestClusterMonitoring", func() {
 				expectedState = models.ClusterStatusFinalizing
 			})
 			It("installing -> error", func() {
-				mockMetric.EXPECT().ClusterInstallationFinished(gomock.Any(), "error", gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+				mockMetric.EXPECT().ClusterInstallationFinished(gomock.Any(), "error", "installing", gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
```
YuviGold, line 246: Use constants

internal/cluster/cluster.go
```diff
@@ -645,11 +645,13 @@ func (m *Manager) VerifyClusterUpdatability(c *common.Cluster) (err error) {
 func (m *Manager) CancelInstallation(ctx context.Context, c *common.Cluster, reason string, db *gorm.DB) *common.ApiErrorResponse {
 	eventSeverity := models.EventSeverityInfo
 	eventInfo := "Cancelled cluster installation"
+	lastState := newStateCluster(c)
 	defer func() {
 		m.eventsHandler.AddEvent(ctx, *c.ID, nil, eventSeverity, eventInfo, time.Now())
+		m.metricAPI.ClusterInstallationFinished(ctx, "canceled", lastState.srcState, c.OpenshiftVersion, *c.ID, c.EmailDomain, c.InstallStartedAt)
```
YuviGold, line 651: Should be a function to be used on both places

internal/metrics/metricsManager.go
```diff
@@ -101,7 +101,7 @@ type API interface {
 	InstallationStarted(clusterVersion string, clusterID strfmt.UUID, emailDomain string, userManagedNetworking string)
 	ClusterHostInstallationCount(emailDomain string, hostCount int, clusterVersion string)
 	Duration(operation string, duration time.Duration)
-	ClusterInstallationFinished(ctx context.Context, result, clusterVersion string, clusterID strfmt.UUID, emailDomain string, installationStartedTime strfmt.DateTime)
+	ClusterInstallationFinished(ctx context.Context, result, prevState, clusterVersion string, clusterID strfmt.UUID, emailDomain string, installationStartedTime strfmt.DateTime)
```
gamli75, line 104: why do we need the prevState?

internal/cluster/cluster.go
slaviered: Those 2 cases (The user facing event and the metrics) are not really the same - they are bundled here by chance. Such function will have too long parameters and will not be re-used too much. For 2 line defer this style is clearer

internal/cluster/cluster_test.go
slaviered: Using constants for the states. Leaving the dashboard labels as-is so they won't be used for other uses

internal/metrics/metricsManager.go
slaviered: For the dashboards - it will show the calculation and statistics of the installation based on the stage from which we arrived.

oshercc, line 104: isn't prev State always be installing?

slaviered: No. With error and canceled states the prevState varies. But, it will be one of the installing states

internal/cluster/transition.go
```diff
@@ -482,6 +483,21 @@ func (th *transitionHandler) PostRefreshCluster(reason string) stateswitch.PostT
 				reason, params.eventHandler, extra...)
 		}
 
+		//report cluster install duration metrics in case of error. Cancel and Installed cases are
+		//treated separately in CancelInstallation and CompleteInstallation respectively
+		if sCluster.srcState != swag.StringValue(sCluster.cluster.Status) {
```
oshercc, line 488:
```suggestion
		if funk.ContainsString(["error","pending-user-action], sCluster.cluster.Status)
				params.metricApi.ClusterInstallationFinished(params.ctx, sCluster.cluster.Status, sCluster.srcState,
					sCluster.cluster.OpenshiftVersion, *sCluster.cluster.ID, sCluster.cluster.EmailDomain,
					sCluster.cluster.InstallStartedAt)
```

slaviered: Here's the reason is important and not the state. RefreshCluster is called with other reasons on the same statuses

internal/cluster/transition.go
```diff
@@ -482,6 +483,21 @@ func (th *transitionHandler) PostRefreshCluster(reason string) stateswitch.PostT
...
+			if reason == statusInfoError {
+				params.metricApi.ClusterInstallationFinished(params.ctx, "error", sCluster.srcState,
```
gamli75, line 490: please use models.ClusterStatusError instead of "error"

internal/cluster/transition.go
```diff
@@ -482,6 +483,21 @@ func (th *transitionHandler) PostRefreshCluster(reason string) stateswitch.PostT
...
+					sCluster.cluster.OpenshiftVersion, *sCluster.cluster.ID, sCluster.cluster.EmailDomain,
+					sCluster.cluster.InstallStartedAt)
+			}
+			if reason == statusInfoInstallingPendingUserAction {
+				params.metricApi.ClusterInstallationFinished(params.ctx, "pending-user-action", sCluster.srcState,
```
gamli75, line 495: Currently, we have only 3 installation results: done, error & canceled. After this code, we will see also pending-user-action result - Not sure it's correct.

slaviered: This is a stopping point for the installation. We really can not measure the installation time including the user's manual intervention

internal/cluster/transition.go
slaviered: It's looks the same but it's not. It's just the label of the dashboard. I did not want to put those label in constants so they will be used for other places by mistake. And I can not use the states either because it will create discrepancies with previous dashboards (for the other states other then error)

slaviered:
/retitle MGMT-7136: Add metric report to all installation complete scenarios

internal/cluster/transition.go
slaviered: After some thought I accept the proposal (state should be taken from the updated cluster though. See Fix

internal/cluster/transition.go
slaviered: FIXED

gamli75 (approved):

PR #2098, opened 2021-06-24
Bug 1974085: fixes network&latency host validations

jordigilh:
# Description

Resolves BZ-1974085. The issue was that the IPs stored in the inventory include the CIDR, and the original code expected to find just the IP without the CIDR. 

I've tested it manually in one of my environments with success: during failure, the validation message now contains the name of the remote host that fails the validation.

# What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [x] Cloud
- [x] Operator Managed Deployments
- [] None

# How was this code tested?

Please, select one or more if needed:

- [x] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

It's recommended to take a few extra minutes to provide more information about
how this code was tested. Here are some questions that may be worth answering:

- Should this PR be tested by the reviewer?
- Is this PR relying on CI for an e2e test run?
- Should this PR be tested in a specific environment?
- Any logs, screenshots, etc that can help with the review process?

# Assignees

Please, add one or two reviewers that could help review this PR. Use `/assign` if you want to assign
this PR directly to someone.

/cc @
/cc @

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- [ ] Are the title and description (in both PR and commit) meaningful and clear?
- [ ] Is there a bug required (and linked) for this change?
- [ ] Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #2187, opened 2021-07-08
MGMT-7085: Add metrics network latency packet loss

jordigilh:
# Assisted Pull Request

## Description

~~**This PR contains the commit from #2098**~~

Defines 2 new prometheus metrics of type histogram for capturing host network latency and packet loss between hosts:

Network latency:
```
metric basename: assisted_installer_host_network_latency_in_ms
labels: clusterVersion,sourceRole,targetRole
```
Packet loss:
```
metric basename: assisted_installer_packet_loss_percentage
labels: clusterVersion,sourceRole,targetRole
```

For more information about histograms, head to the [prometheus documentation](https://prometheus.io/docs/concepts/metric_types/#histogram) website.

- [X] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

- [X] Are the title and description (in both PR and commit) meaningful and clear?

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -22,6 +22,7 @@ import (
 	"github.com/openshift/assisted-service/internal/metrics"
 	"github.com/openshift/assisted-service/internal/operators"
 	"github.com/openshift/assisted-service/models"
+	"github.com/openshift/assisted-service/pkg/commonutils"
 	"github.com/openshift/assisted-service/pkg/leader"
 	logutil "github.com/openshift/assisted-service/pkg/log"
 	"github.com/pkg/errors"
@@ -535,10 +536,21 @@ func (m *Manager) SetUploadLogsAt(ctx context.Context, h *models.Host, db *gorm.
 
 func (m *Manager) UpdateConnectivityReport(ctx context.Context, h *models.Host, connectivityReport string) error {
 	if h.Connectivity != connectivityReport {
+		//retrieve cluster
+		var cluster common.Cluster
+		err := m.db.First(&cluster, "id = ?", h.ClusterID).Error
+		if err != nil {
+			return errors.Wrapf(err, "failed to find cluster %s", h.ClusterID)
+		}
+		if len(connectivityReport) > 0 {
+			m.captureConnectivityReportMetrics(ctx, cluster.OpenshiftVersion, h, connectivityReport, cluster.Hosts)
+		}
+
 		// Only if the connectivity between the hosts changed change the updated_at field
 		if err := m.db.Model(h).Update("connectivity", connectivityReport).Error; err != nil {
 			return errors.Wrapf(err, "failed to set connectivity to host %s", h.ID.String())
 		}
+
 	}
 	return nil
 }
@@ -1137,4 +1149,26 @@ func (m *Manager) GetHostByKubeKey(key types.NamespacedName) (*common.Host, erro
 
 func (m *Manager) UnRegisterHost(ctx context.Context, hostID, clusterID string) error {
 	return m.db.Where("id = ? and cluster_id = ?", hostID, clusterID).Delete(&common.Host{}).Error
+}
+
+func (m *Manager) captureConnectivityReportMetrics(ctx context.Context, openshiftVersion string, h *models.Host, report string, hosts []*models.Host) {
+	log := logutil.FromContext(ctx, logrus.New())
+	defer commonutils.MeasureOperation("CaptureConnectivityReportMetrics", log, m.metricApi)
+
+	connectivityReport, err := hostutil.UnmarshalConnectivityReport(h.Connectivity)
+	if err != nil {
+		log.Errorf("unable to unmarshal connectivity report for host ID %s:%v", h.ID, err)
+		return
+	}
+	for _, r := range connectivityReport.RemoteHosts {
+		for _, l3 := range r.L3Connectivity {
+			_, targetRole, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
+			if err != nil {
+				log.Warn(err)
+				continue
+			}
+			m.metricApi.NetworkLatencyBetweenHosts(openshiftVersion, h.Role, targetRole, l3.AverageRTTMs)
+			m.metricApi.PacketLossBetweenHosts(openshiftVersion, h.Role, targetRole, l3.PacketLossPercentage)
+		}
+	}
 }
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -709,10 +709,10 @@ func (v *validator) hasSufficientNetworkLatencyRequirementForRole(c *validationC
 }
 
 func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host) (ValidationStatus, []string, error) {
-	var connectivityReport *models.ConnectivityReport
 	connectivityReport, err := hostutil.UnmarshalConnectivityReport(host.Connectivity)
 	if err != nil {
-		return ValidationFailure, nil, fmt.Errorf("unable to unmarshall host connectivity for %s:%s", host.ID, err)
+		v.log.Errorf("Unable to unmarshall host connectivity for %s:%s", host.ID, err)
+		return ValidationError, nil, nil
 	}
 	failedHostIPs := map[string]struct{}{}
 	failedHostNames := []string{}
@@ -722,7 +722,8 @@ func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRole
 				if _, ok := failedHostIPs[l3.RemoteIPAddress]; !ok {
 					hostname, role, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
 					if err != nil {
-						return ValidationFailure, nil, fmt.Errorf("unable to determine role and hostname for IP %s: %s", l3.RemoteIPAddress, err)
+						v.log.Error(err)
+						return ValidationFailure, nil, err
 					}
 					if role == host.Role {
 						failedHostIPs[l3.RemoteIPAddress] = struct{}{}
@@ -751,6 +752,8 @@ func (v *validator) printSufficientNetworkLatencyRequirementForRole(c *validatio
 		return fmt.Sprintf("Network latency requirements of less or equals than %.3f ms not met for connectivity between %s and %s.", *c.clusterHostRequirements.Total.NetworkLatencyThresholdMs, c.host.ID, strings.Join(hostNames, ","))
 	case ValidationPending:
 		return "Missing network latency information."
+	case ValidationError:
+		return "Parse error while attempting to process the connectivity report"
 	default:
 		return fmt.Sprintf("Unexpected status %s", status)
 	}
@@ -774,8 +777,8 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
 func (v *validator) validatePacketLossForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host) (ValidationStatus, []string, error) {
 	connectivityReport, err := hostutil.UnmarshalConnectivityReport(host.Connectivity)
 	if err != nil {
-		v.log.Warnf("unable to unmarshall host connectivity for %s:%s", host.ID, err)
-		return ValidationFailure, nil, err
+		v.log.Errorf("Unable to unmarshall host connectivity for %s:%s", host.ID, err)
+		return ValidationError, nil, nil
 	}
 	failedHostIPs := map[string]struct{}{}
 	failedHostNames := []string{}
@@ -785,8 +788,8 @@ func (v *validator) validatePacketLossForRole(host *models.Host, clusterRoleReqs
 				if _, ok := failedHostIPs[l3.RemoteIPAddress]; !ok {
 					hostname, role, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
 					if err != nil {
-						v.log.Warnf("unable to determine role and hostname for IP %s: %v", l3.RemoteIPAddress, err)
-						return ValidationFailure, nil, fmt.Errorf("unable to determine role and hostname for IP %s: %s", l3.RemoteIPAddress, err)
+						v.log.Error(err)
+						return ValidationFailure, nil, err
 					}
 					if role == host.Role {
 						failedHostIPs[l3.RemoteIPAddress] = struct{}{}
@@ -815,6 +818,8 @@ func (v *validator) printSufficientPacketLossRequirementForRole(c *validationCon
 		return fmt.Sprintf("Packet loss percentage requirement of less or equals than %.2f%% not met for connectivity between %s and %s.", *c.clusterHostRequirements.Total.PacketLossPercentage, c.host.ID, strings.Join(hostNames, ","))
 	case ValidationPending:
 		return "Missing packet loss information."
+	case ValidationError:
+		return "Parse error while attempting to process the connectivity report"
 	default:
 		return fmt.Sprintf("Unexpected status %s", status)
 	}
--- a/internal/metrics/metricsManager.go
+++ b/internal/metrics/metricsManager.go
@@ -41,6 +41,8 @@ const (
 	counterFilesystemUsagePercentage              = "assisted_installer_filesystem_usage_percentage"
 	counterMonitoredHosts                         = "assisted_installer_monitored_hosts"
 	counterMonitoredClusters                      = "assisted_installer_monitored_clusters"
+	histogramNetworkLatencyMilliseconds           = "assisted_installer_host_network_latency_in_ms"
+	histogramPacketLossPercentage                 = "assisted_installer_packet_loss_percentage"
 )
 
 const (
@@ -65,6 +67,8 @@ const (
 	counterDescriptionFilesystemUsagePercentage              = "The percentage of the filesystem usage by the service"
 	counterDescriptionMonitoredHosts                         = "Number of hosts monitored by host monitor"
 	counterDescriptionMonitoredClusters                      = "Number of clusters monitored by cluster monitor"
+	histogramDescriptionNetworkLatencyMilliseconds           = "Histogram/sum/count of the L3 network latency in milliseconds between hosts"
+	histogramDescriptionPacketLossPercentage                 = "Histogram/sum/count of the L3 packet loss percentage between hosts"
 )
 
 const (
@@ -90,6 +94,8 @@ const (
 	imageLabel                 = "imageName"
 	hosts                      = "hosts"
 	clusters                   = "clusters"
+	sourceRoleLabel            = "sourceRole"
+	targetRoleLabel            = "targetRole"
 )
 
 type API interface {
@@ -108,6 +114,8 @@ type API interface {
 	FileSystemUsage(usageInPercentage float64)
 	MonitoredHostsCount(monitoredHosts int64)
 	MonitoredClusterCount(monitoredClusters int64)
+	NetworkLatencyBetweenHosts(clusterVersion string, sourceRole, targetRole models.HostRole, latency float64)
+	PacketLossBetweenHosts(clusterVersion string, sourceRole, targetRole models.HostRole, packetLoss float64)
 }
 
 type MetricsManager struct {
@@ -135,6 +143,8 @@ type MetricsManager struct {
 	serviceLogicFilesystemUsagePercentage              *prometheus.GaugeVec
 	serviceLogicMonitoredHosts                         *prometheus.GaugeVec
 	serviceLogicMonitoredClusters                      *prometheus.GaugeVec
+	serviceLogicNetworkLatencyMilliseconds             *prometheus.HistogramVec
+	serviceLogicPacketLossPercentage                   *prometheus.HistogramVec
 }
 
 var _ API = &MetricsManager{}
@@ -310,6 +320,21 @@ func NewMetricsManager(registry prometheus.Registerer, eventsHandler events.Hand
 			Name:      counterMonitoredClusters,
 			Help:      counterDescriptionMonitoredClusters,
 		}, []string{hosts}),
+
+		serviceLogicNetworkLatencyMilliseconds: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      histogramNetworkLatencyMilliseconds,
+			Help:      histogramDescriptionNetworkLatencyMilliseconds,
+			Buckets:   []float64{0, 25, 50, 75, 100, 150, 200, 300, 400, 500, 1000},
+		}, []string{openshiftVersionLabel, sourceRoleLabel, targetRoleLabel}),
+		serviceLogicPacketLossPercentage: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      histogramPacketLossPercentage,
+			Help:      histogramDescriptionPacketLossPercentage,
+			Buckets:   []float64{0, 10, 20, 30, 40, 50, 60, 70, 80, 90},
+		}, []string{openshiftVersionLabel, sourceRoleLabel, targetRoleLabel}),
 	}
 
 	registry.MustRegister(
@@ -332,6 +357,8 @@ func NewMetricsManager(registry prometheus.Registerer, eventsHandler events.Hand
 		m.serviceLogicFilesystemUsagePercentage,
 		m.serviceLogicMonitoredHosts,
 		m.serviceLogicMonitoredClusters,
+		m.serviceLogicNetworkLatencyMilliseconds,
+		m.serviceLogicPacketLossPercentage,
 	)
 	return m
 }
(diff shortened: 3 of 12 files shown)


pkliczewski:
@filanov Who can review this PR?

internal/host/validator.go
```diff
@@ -674,50 +676,50 @@ func (v *validator) hasSufficientNetworkLatencyRequirementForRole(c *validationC
 		return ValidationSuccess
 	}
 
+	v.log.Debugf("Maximum network latency allowed for host %s: %f", c.host.ID, *c.clusterHostRequirements.Total.NetworkLatencyThresholdMs)
```
gamli75, line 679: I guess that you added it for debug - please remove it. this method is called every 8 seconds.

internal/host/validator.go
```diff
@@ -674,50 +676,50 @@ func (v *validator) hasSufficientNetworkLatencyRequirementForRole(c *validationC
...
+		return ValidationFailure, nil, fmt.Errorf("unable to unmarshall host connectivity for %s:%s", host.ID, err)
 	}
 	failedHostIPs := map[string]struct{}{}
 	failedHostNames := []string{}
 	for _, r := range connectivityReport.RemoteHosts {
 		for _, l3 := range r.L3Connectivity {
+			hostname, role, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
+			if err != nil {
+				v.log.Warnf("unable to determine host's role and hostname for IP %s: %s", l3.RemoteIPAddress, err)
+				return ValidationPending, nil, fmt.Errorf("unable to determine role and hostname for IP %s: %s", l3.RemoteIPAddress, err)
```
gamli75, line 700: I think it should be ValidationFailure and not ValidationPending

internal/host/validator.go
```diff
@@ -674,50 +676,50 @@ func (v *validator) hasSufficientNetworkLatencyRequirementForRole(c *validationC
...
+			}
+			v.metricsAPI.NetworkLatencyBetweenHosts(openshiftVersion, emailDomain, host.Role, role, l3.AverageRTTMs)
```
gamli75, line 702: it means that we will send a metric every 8 seconds even if nothing was changed, right? 
I think we should send the metric in case of failure and once before installation started 
@avishayt WDYT?

internal/host/validator.go
```diff
@@ -737,51 +739,50 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
 		return ValidationSuccess
 	}
 
+	v.log.Debugf("Maximum packet loss percentage allowed for host %s: %f", c.host.ID, *c.clusterHostRequirements.Total.PacketLossPercentage)
```
gamli75, line 742: I guess that you added it for debug - please remove it. this method is called every 8 seconds

internal/host/validator.go
```diff
@@ -737,51 +739,50 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
...
+		return ValidationFailure, nil, err
 	}
 	failedHostIPs := map[string]struct{}{}
 	failedHostNames := []string{}
 	for _, r := range connectivityReport.RemoteHosts {
 		for _, l3 := range r.L3Connectivity {
+			hostname, role, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
+			if err != nil {
+				v.log.Warnf("unable to determine host's role and hostname for IP %s: %s", l3.RemoteIPAddress, err)
+				return ValidationPending, nil, fmt.Errorf("unable to determine role and hostname for IP %s: %s", l3.RemoteIPAddress, err)
```
gamli75, line 763: same here - ValidationFailure instead of ValidationPending

internal/host/validator.go
jordigilh: If the debug level is no enabled by default, this line will not have any impact. Do you still want me to remove it? It's good to have for debugging purposes.

internal/host/validator.go
jordigilh: Pending translates for "missing packet loss information", which I considered in the same scope as missing inventory data for other hosts.
When a host has not yet reported its inventory, does it show as pending of failure in the UI?

internal/host/validator.go
gamli75: we can check the thresholds in the saas.yaml, right? if the answer is yes so I think we should remove it

internal/host/validator.go
jordigilh: Same comment as before: is the default log level set to debug in production systems? If not, this line won't have any impact.

internal/host/validator.go
gamli75: Pending means that the user needs to do something and it's not the case here.
when host has not yet reported its inventory - we display it as failure, see: https://github.com/openshift/assisted-service/blob/master/internal/host/validator.go#L206

jordigilh: ok, I'll change it to failure then

internal/host/validator.go
jordigilh: Ok I'll remove it.

internal/host/validator.go
```diff
@@ -674,50 +676,50 @@ func (v *validator) hasSufficientNetworkLatencyRequirementForRole(c *validationC
 		return ValidationSuccess
 	}
 
+	v.log.Debugf("Maximum network latency allowed for host %s: %f", c.host.ID, *c.clusterHostRequirements.Total.NetworkLatencyThresholdMs)
 	if len(c.host.Connectivity) == 0 {
 		return ValidationPending
```
jordigilh, line 681: @gamli75 should this be Failure as well?

gamli75: yes

internal/host/validator.go
ronniel1: That is not a problem.
The metric is not really sent every time we call this api. It is just collected in the background

internal/host/common.go
```diff
@@ -102,20 +102,16 @@ func GetHostnameAndRoleByIP(ip string, hosts []*models.Host) (string, models.Hos
...
-						return getRealHostname(h, inv), h.Role, nil
-					}
-				}
-			}
+			return "", "", fmt.Errorf("unable to unmarshall cluster inventory for host %s: %s", h.RequestedHostname, err)
 		}
 		for _, i := range inv.Interfaces {
-			for _, ipv6 := range i.IPV6Addresses {
-				if ip == ipv6 {
+			ips := append(i.IPV4Addresses, i.IPV6Addresses...)
```
ronniel1, line 108: Please add a UT that shows that this logic is working correctly in all cases

internal/host/validator.go
```diff
@@ -740,49 +741,47 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
...
 	if err != nil {
-		return nil, fmt.Errorf("unable to unmarshall host connectivity for %s:%s", host.ID, err)
+		v.log.Warnf("unable to unmarshall host connectivity for %s:%s", host.ID, err)
+		return ValidationFailure, nil, err
 	}
 	failedHostIPs := map[string]struct{}{}
 	failedHostNames := []string{}
 	for _, r := range connectivityReport.RemoteHosts {
 		for _, l3 := range r.L3Connectivity {
+			hostname, role, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
```
ronniel1, line 758: Why did we remove the detailed logging in validateNetworkLatencyForRole() above, but added it here?

internal/host/validator.go
```diff
@@ -677,48 +679,47 @@ func (v *validator) hasSufficientNetworkLatencyRequirementForRole(c *validationC
...
-						v.log.Warnf("unable to determine host's role and hostname for IP: %s", err)
-						continue
-					}
-					if role == host.Role {
-						failedHostIPs[l3.RemoteIPAddress] = struct{}{}
-						failedHostNames = append(failedHostNames, hostname)
-					}
+				if _, ok := failedHostIPs[l3.RemoteIPAddress]; !ok && role == host.Role {
+					failedHostIPs[l3.RemoteIPAddress] = struct{}{}
+					failedHostNames = append(failedHostNames, hostname)
```
ronniel1, line 705: What about sending metrics?

internal/host/validator.go
```diff
@@ -740,49 +741,47 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
...
 	failedHostIPs := map[string]struct{}{}
 	failedHostNames := []string{}
 	for _, r := range connectivityReport.RemoteHosts {
 		for _, l3 := range r.L3Connectivity {
+			hostname, role, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
+			if err != nil {
+				v.log.Warnf("unable to determine host's role and hostname for IP %s: %s", l3.RemoteIPAddress, err)
+				return ValidationFailure, nil, fmt.Errorf("unable to determine role and hostname for IP %s: %s", l3.RemoteIPAddress, err)
+			}
+			v.metricsAPI.PacketLossBetweenHosts(openshiftVersion, emailDomain, host.Role, role, l3.PacketLossPercentage)
```
ronniel1, line 763: Metric should always be reported. Not just on high values

internal/metrics/metricsManager.go
```diff
@@ -108,6 +114,8 @@ type API interface {
 	FileSystemUsage(usageInPercentage float64)
 	MonitoredHostsCount(monitoredHosts int64)
 	MonitoredClusterCount(monitoredClusters int64)
+	NetworkLatencyBetweenHosts(clusterVersion string, emailDomain string, sourceRole, targetRole models.HostRole, latency float64)
+	PacketLossBetweenHosts(clusterVersion string, emailDomain string, sourceRole, targetRole models.HostRole, packetLoss float64)
```
ronniel1, line 118: No need for the emailDomain.
Is there a readon to add the source and target roles?

internal/host/common.go
jordigilh: Will this be sufficient?
https://play.golang.org/p/OYhsH6TkKR-

internal/host/validator.go
jordigilh: It was not considered when I discussed this with @gamli75 and @avishayt . What  we are aiming to do is capture the latency in a histogram. Knowing the latency limits between roles, we can easily extract the percentage of installations where the limit was crossed for a given set of roles.

internal/host/validator.go
jordigilh: Agreed, and that's the case here. The next line is the one that checks for the threshold. This line will report the packet loss value for each entry.

internal/metrics/metricsManager.go
jordigilh: The labels were discussed with @gamli75 and @avishayt to use these labels in the metric. The reason for adding the roles is to provide a better visibility between hosts with the same role and different role. Currently we are only validating against hosts with the same role, but knowing the value between master<->worker can also help us get a better picture of the average cluster latency/packet loss overall.

internal/host/validator.go
jordigilh: No reason. I'll remove the log entry and push the log message as part of the error in both cases.

internal/metrics/metricsManager.go
ronniel1: roles is OK, but email domain should definitely not be in the metric
@avishayt , @gamli75

ronniel1: About the roles, we need to measure the latency between all the hosts, regardless of the roles.
Validation is something else.
The metrics should be sent on the values we measured and not on the result of the validations

internal/host/common.go
ronniel1: This is not a unit test
will it run anc check the logic everytime someone makes a change?

ronniel1: This is not a unit test
will it run anc check the logic everytime someone makes a change?

jordigilh: Ok, I'll add unit test for this

internal/metrics/metricsManager.go
jordigilh: As discussed, I'll remove the `emailDomain` from both metrics.

filanov:
cc: @ori-amizur

internal/host/validator.go
jordigilh: :+1:

internal/host/validator.go
jordigilh: Removed.

internal/host/validator.go
jordigilh: OTOH, I prefer to leave the log message for now in case of error, just as it was initially. I'll make sure that both validations have a log entry.

jordigilh:
@ronniel1 @gamli75 @ori-amizur I have addressed all comments. If you have no further questions, please help approve this PR.

ronniel1 (commented):
/lgtm /approve

internal/host/refresh_status_preprocessor.go
```diff
@@ -37,12 +38,13 @@ type refreshPreprocessor struct {
 	disabledHostValidations DisabledHostValidations
 }
 
-func newRefreshPreprocessor(log logrus.FieldLogger, hwValidatorCfg *hardware.ValidatorCfg, hwValidator hardware.Validator, operatorsApi operators.API, disabledHostValidations DisabledHostValidations) *refreshPreprocessor {
+func newRefreshPreprocessor(log logrus.FieldLogger, hwValidatorCfg *hardware.ValidatorCfg, hwValidator hardware.Validator, operatorsApi operators.API, disabledHostValidations DisabledHostValidations, metricsAPI metrics.API) *refreshPreprocessor {
```
ori-amizur, line 41: I am not sure that refreshStatusPreprocessor is the correct location to measure latency metrics.  It can be called every 10s when there is no change occurred between subsequent calls, and it may be called every 5 minutes when many connectivity changes occurred in-between.  
Why not measure in the location that connectivity is received and processed here https://github.com/openshift/assisted-service/blob/8758e8197b6bdd44bfa2e652f344ed6c8e3485fe/internal/host/host.go#L536 ?

ori-amizur (commented):
I think the refreshStatusPreprocessor is not the correct location to analyze the metrics.  https://github.com/openshift/assisted-service/blob/8758e8197b6bdd44bfa2e652f344ed6c8e3485fe/internal/host/host.go#L536 seems more appropriate.

internal/host/refresh_status_preprocessor.go
ronniel1: Seems that the correct place to measure latency is when getting the results, no?

ori-amizur: Right.  That's the function above

jordigilh: Makes sense. thanks for the suggestion.

internal/host/validator.go
```diff
@@ -703,12 +706,12 @@ func (v *validator) hasSufficientNetworkLatencyRequirementForRole(c *validationC
...
 		return ValidationFailure
 	}
-	status, _, _ := v.validateNetworkLatencyForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts)
+	status, _, _ := v.validateNetworkLatencyForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts, c.cluster.OpenshiftVersion)
 	return status
 }
 
-func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host) (ValidationStatus, []string, error) {
-	var connectivityReport *models.ConnectivityReport
+func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host, openshiftVersion string) (ValidationStatus, []string, error) {
```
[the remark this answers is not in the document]
jordigilh: It was proposed with @gamli75 and @avishayt as part of the metrics labels. On a second thought, the latency and packet loss are intrinsic values of the hosts, not something that will change based on the openshift version. If so, I'll remove it and leave only the `sourceRole` and `targetRole`.

internal/host/validator.go
```diff
@@ -763,31 +766,30 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
 	if len(c.host.Connectivity) == 0 {
 		return ValidationFailure
 	}
-	status, _, _ := v.validatePacketLossForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts)
+	status, _, _ := v.validatePacketLossForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts, c.cluster.OpenshiftVersion)
 	return status
 }
 
-func (v *validator) validatePacketLossForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host) (ValidationStatus, []string, error) {
+func (v *validator) validatePacketLossForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host, openshiftVersion string) (ValidationStatus, []string, error) {
```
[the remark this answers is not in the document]
jordigilh: I'll remove it from the metric labels.

internal/host/validator.go
```diff
@@ -717,16 +720,16 @@ func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRole
 	failedHostNames := []string{}
 	for _, r := range connectivityReport.RemoteHosts {
 		for _, l3 := range r.L3Connectivity {
+			hostname, role, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
+			if err != nil {
+				v.log.Warnf("unable to unmarshall host connectivity for %s:%s", host.ID, err)
+				return ValidationFailure, nil, err
```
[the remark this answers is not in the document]
jordigilh: How do you suggest to handle this use case? @gamli75 confirmed that `Pending` state is only when the host needs user intervention. I don't think it makes sense to continue the execution in case of unmarshalling failure and end up returning success because the slice is empty.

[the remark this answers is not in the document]
jordigilh: Note that QE just opened a BZ on this: https://bugzilla.redhat.com/show_bug.cgi?id=1983412
Should we then ignore the error and continue execution or return as `Pending`? /cc @gamli75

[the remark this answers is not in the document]
jordigilh: Ok, `ValidationError` should do.

internal/host/validator.go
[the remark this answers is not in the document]
ori-amizur: Just wanted to know.  If required then leave it.

internal/host/validator.go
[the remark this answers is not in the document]
ori-amizur: My remark was for the error message. The connectivity unmarshall is on line 715.  Here it is a different error.

internal/host/validator.go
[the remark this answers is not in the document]
jordigilh: You have a point anyway. I actually don't see much value in adding the openshift version here since it won't change the outcome of the latency and packet loss. It might make more sense to include which virt provider (if any) is being used.

internal/host/validator.go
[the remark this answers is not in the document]
gamli75: @ori-amizur you can see here: https://github.com/openshift/assisted-service/blob/master/internal/host/validator.go#L464 
if inventory == nil - it's pending
if the unmarshal failed - it's failure

internal/host/validator.go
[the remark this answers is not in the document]
gamli75: @jordigilh don't remove it - let's leave it as is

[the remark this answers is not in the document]
jordigilh: Ok, I'll keep it then.

internal/host/validator.go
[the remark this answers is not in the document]
jordigilh: At the end I did not remove the label.

internal/host/refresh_status_preprocessor.go
jordigilh: At the end I added it here:
https://github.com/openshift/assisted-service/blob/98f8f74399ded5f9ad68c58ada0a3878fce8ab6d/internal/cluster/transition.go#L533-L534
Let me know what you think.

internal/host/validator.go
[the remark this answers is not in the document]
jordigilh: Changed to log the error message verbatim. Also I change to return a `ValidationError` if the unmarshal of the connectivity report fails. WDYT @gamli75 @ori-amizur ?

internal/host/validator.go
```diff
@@ -763,31 +766,30 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
...
-		return ValidationFailure, nil, err
+		return ValidationFailure, nil, fmt.Errorf("unable to unmarshall host connectivity for %s:%s", host.ID, err)
 	}
 	failedHostIPs := map[string]struct{}{}
 	failedHostNames := []string{}
 	for _, r := range connectivityReport.RemoteHosts {
 		for _, l3 := range r.L3Connectivity {
+			hostname, role, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
+			if err != nil {
+				v.log.Warnf("unable to unmarshall host connectivity for %s:%s", host.ID, err)
```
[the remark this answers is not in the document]
jordigilh: Agreed, changed to just log the error verbatim.

jordigilh:
@ori-amizur @ronniel1 now that I moved the logic to collect the metrics outside the validation loop, I think it is no longer necessary to modify the unit tests to make sure each host has its own unique IP. Do you want me to revert the changes to the unit tests?

jordigilh:
@ori-amizur @ronniel1 @gamli75 I've implemented the suggested changes. Please review and approve if you have no more comments.

internal/cluster/transition.go
```diff
@@ -532,6 +534,8 @@ func (th *transitionHandler) InstallCluster(sw stateswitch.StateSwitch, args sta
 	// send metric and event that installation process has been started
 	params.metricApi.InstallationStarted(cluster.OpenshiftVersion, *cluster.ID, cluster.EmailDomain, strconv.FormatBool(swag.BoolValue(cluster.UserManagedNetworking)))
 	params.metricApi.ClusterHostInstallationCount(cluster.EmailDomain, len(cluster.Hosts), cluster.OpenshiftVersion)
+	captureMetricsForNetworkLatency(ctx, params.metricApi, cluster.OpenshiftVersion, cluster.Hosts)
```
ori-amizur, line 537: I don't think that InstallCluster is the right location to collect latency metrics.  What about metrics before installation?

internal/host/refresh_status_preprocessor.go
ori-amizur: I think this https://github.com/openshift/assisted-service/blob/8758e8197b6bdd44bfa2e652f344ed6c8e3485fe/internal/host/host.go#L536 location is better because it is the place that new connectivity report is handled.
In addition, it should be handled at host level.

internal/cluster/transition.go
jordigilh: What do you mean before installation? 
Doing it once here reduces the amount of cycles required to process the connectivity report and also guarantees that the reporting information contains only the latest update. Correct me if I'm mistaken, but doing it at the place you suggested can lead to situations where the latency and packet loss information of a host can be reprocessed more than once if there are new hosts being added to the cluster, or even disabled/re-enabled, or new interfaces added, right?

internal/host/refresh_status_preprocessor.go
jordigilh: My concern is about potential re-processing of the connectivity report before the cluster is created due to changes in the cluster topology during pre-installation. Capturing the report only once right before installation gives a clearer picture of these values.

internal/cluster/transition.go
ori-amizur: I think users expect to see the metrics before starting installation.  In addition, should these metrics captured once only, or when connectivity results arrive ?

jordigilh: I'm not sure the metrics are targeted to be displayed to the users. If we want to display the current latency information, that could be shown in the UI, but at this point the only way to visualize the metrics will be in Grafana.
As for when the metrics should be captured, it's up to us to decide. I favor the option to capture them when the cluster is about to be installed so that we don't have polluted values, in which hosts that have high latency are not eventually added to the cluster for this or other reasons, or even glitches in the network that cause high latency during the discovery.  
If the general consensus is to capture any latency as it is reported to the service, then your suggestion is the right place to record the metrics. WDYT @avishayt @gamli75 ?

gamli75: The decision that was made by Ronnie & Avishay is to report the metric every time we receive the connectivity report and not only once before installation.

internal/host/host.go
```diff
@@ -1138,3 +1148,25 @@ func (m *Manager) GetHostByKubeKey(key types.NamespacedName) (*common.Host, erro
...
+		return
+	}
+	for _, r := range connectivityReport.RemoteHosts {
+		for _, l3 := range r.L3Connectivity {
+			_, targetRole, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
+			if err != nil {
+				log.Warn(err)
+				continue
+			}
+			m.metricApi.NetworkLatencyBetweenHosts(openshiftVersion, h.Role, targetRole, l3.AverageRTTMs)
```
ori-amizur, line 1168: Seems strange the host is not part of the metrics.  If there was a problem - don't we want to know which hosts were involved ?

internal/host/validator.go
```diff
@@ -722,7 +722,8 @@ func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRole
 				if _, ok := failedHostIPs[l3.RemoteIPAddress]; !ok {
 					hostname, role, err := GetHostnameAndRoleByIP(l3.RemoteIPAddress, hosts)
 					if err != nil {
-						return ValidationFailure, nil, fmt.Errorf("unable to determine role and hostname for IP %s: %s", l3.RemoteIPAddress, err)
+						v.log.Error(err)
```
ori-amizur, line 725: This log message will be meaningless without the context

internal/host/host.go
jordigilh: Adding the host as a label would literally explode the metrics' cardinality.

internal/host/validator.go
jordigilh: I figured that the incoming error message already contains sufficient information, wrapping it would not add any value:

https://github.com/openshift/assisted-service/blob/581f79c8b3b42dfe92467bfeaf644fde404fd2e4/internal/host/common.go#L103-L106

https://github.com/openshift/assisted-service/blob/581f79c8b3b42dfe92467bfeaf644fde404fd2e4/internal/host/common.go#L119-L120

gamli75 (approved):

PR #2207, opened 2021-07-13
MGMT-7214: Add InfraEnv model

ravidbro:
Add InfraEnv model to the swagger definitions and create a new table for the InfraEnv resource.
This PR doesn't introduce yet any usage of the new resource, it's a step for the future work of late-binding


internal/common/db.go
```diff
@@ -76,6 +76,13 @@ type Host struct {
 	TriggerMonitorTimestamp time.Time
 }
 
+type InfraEnv struct {
+	models.InfraEnv
+
+	// The pull secret that obtained from the Pull Secret page on the Red Hat OpenShift Cluster Manager site.
+	PullSecret string `json:"pull_secret" gorm:"type:TEXT"`
```
yevgeny-shnaidman, line 83: we need to add KubeKeyNamespace string `json:"kube_key_namespace"` as a preparation for creating InfraEnv via specific API

PR #2208, opened 2021-07-13
MGMT-7197: Implementing cluster's installation progress bar in BE.

ybettan:
# Assisted Pull Request

The progress bar is currently implemented in the UI and should move to
BE. It will also allow other APIs (like kube-api) to use that feature.

**(This part is not a part of the commit message)**
The implementation follows https://github.com/openshift/assisted-service/pull/2042#discussion_r659296212 with a small change - the finalizing progress will be handled by the operators' API instead of inside the monitoring process.

- [x] New Feature
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

I have run an infinite loop that checks the progress during the cluster installation (using test-infra) and those are the results:
```
{
  "preparing_for_installation_stage_percentage": 100,
  "total_percentage": 10
}
...
{
  "installing_stage_percentage": 12,
  "preparing_for_installation_stage_percentage": 100,
  "total_percentage": 18
}
...
{
  "installing_stage_percentage": 54,
  "preparing_for_installation_stage_percentage": 100,
  "total_percentage": 47
}
...
{
  "installing_stage_percentage": 79,
  "preparing_for_installation_stage_percentage": 100,
  "total_percentage": 65
}
...
{
  "installing_stage_percentage": 100,
  "preparing_for_installation_stage_percentage": 100,
  "total_percentage": 80
}
...
{
  "finalizing_stage_percentage": 100,
  "installing_stage_percentage": 100,
  "preparing_for_installation_stage_percentage": 100,
  "total_percentage": 100
}
```

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


internal/operators/handler/handler.go
```diff
@@ -22,15 +23,16 @@ import (
...
+	db                 *gorm.DB
+	log                logrus.FieldLogger
+	eventsHandler      events.Handler
+	clusterProgressAPI cluster.ProgressAPI
 }
 
 // NewHandler creates new handler
-func NewHandler(operatorsAPI operators.API, log logrus.FieldLogger, db *gorm.DB, eventsHandler events.Handler) *Handler {
-	return &Handler{operatorsAPI: operatorsAPI, log: log, db: db, eventsHandler: eventsHandler}
+func NewHandler(operatorsAPI operators.API, log logrus.FieldLogger, db *gorm.DB, eventsHandler events.Handler, clusterProgressAPI cluster.ProgressAPI) *Handler {
```
jakub-dzon, line 34: :+1:

internal/operators/handler/handler.go
```diff
@@ -67,6 +69,11 @@ func (h *Handler) ReportMonitoredOperatorStatus(ctx context.Context, params rest
 	if err != nil {
 		return common.GenerateErrorResponder(err)
 	}
+
+	//FIXME: should this 2 updates be in a transaction? - if yes extract the transaction of UpdateMonitoredOperatorStatus one layer above
```
jakub-dzon, line 73: IMO it should be one transaction. Especially one case can be problematic without it:
- Update with operator status = Available;
- Monitored Operator gets updated;
- Progress update fails;

As a result the installation progress visible to the user is stuck even if the operator has completed its deployment. It will be this way until the agent retries sending the status update. At best it will be intermittent inconsistency, at worst the inconsistency will never be fixed.

ybettan: That was my thought as well, I just saw other places with some non-transcationed API calls so I was wondering - I will get it inside a transaction then.

Thanks

internal/cluster/cluster.go
```diff
@@ -668,6 +674,70 @@ func (m *Manager) UpdateLogsProgress(ctx context.Context, c *common.Cluster, pro
...
+	var totalHostsDoneStages, totalHostsStages float64
+	for _, h := range cluster.Hosts {
+		stages := m.hostAPI.GetStagesByRole(h.Role, h.Bootstrap)
+		currentIndex := m.hostAPI.IndexOfStage(h.Progress.CurrentStage, stages)
+		totalHostsDoneStages += float64(currentIndex + 1)
+		totalHostsStages += float64(len(stages))
+	}
+
+	installingStagePercentage := int64((totalHostsDoneStages / totalHostsStages) * 100)
+	totalPercentage := int64(common.ProgressWeightPreparingForInstallationStage*float64(cluster.Progress.PreparingForInstallationStagePercentage) +
```
ori-amizur, line 695: If we are in installation-in-progress shouldn't the preparing-for-installation percentage currently be complete?

internal/cluster/cluster.go
```diff
@@ -668,6 +674,70 @@ func (m *Manager) UpdateLogsProgress(ctx context.Context, c *common.Cluster, pro
...
+		// failing OLM operators will lead to dgraded cluster but are still considered as a progress
+		if o.OperatorType == models.OperatorTypeOlm && o.Status != models.OperatorStatusProgressing {
+			doneOperatorsCount++
+		}
+	}
+
+	finalizingStagePercentage := int64((doneOperatorsCount / float64(len(cluster.MonitoredOperators))) * 100)
+	totalPercentage := int64(common.ProgressWeightPreparingForInstallationStage*float64(cluster.Progress.PreparingForInstallationStagePercentage) +
+		common.ProgressWeightInstallingStage*float64(cluster.Progress.InstallingStagePercentage) +
+		common.ProgressWeightFinalizingStage*float64(finalizingStagePercentage))
```
ori-amizur, line 731: Currently - only finalizingStagePercentage should be less than 100

internal/host/host.go
```diff
@@ -458,20 +468,29 @@ func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, pro
...
+			}
+			if currentIndex < m.IndexOfStage(previousProgress.CurrentStage, stages) &&
+				!m.allowStageOutOfOrder(h, progress.CurrentStage) {
+				return errors.Errorf("Can't assign lower stage \"%s\" after host has been in stage \"%s\"",
+					progress.CurrentStage, previousProgress.CurrentStage)
+			}
 		}
+
+		stages := m.GetStagesByRole(h.Role, h.Bootstrap)
+		currentIndex := m.IndexOfStage(progress.CurrentStage, stages)
```
ori-amizur, line 491: Is -1 a valid index stage because it can be returned from the function ?

internal/cluster/cluster.go
```diff
@@ -668,6 +674,70 @@ func (m *Manager) UpdateLogsProgress(ctx context.Context, c *common.Cluster, pro
...
+	cluster, err := common.GetClusterFromDB(m.db, clusterID, common.UseEagerLoading)
+	if err != nil {
+		log.WithError(err).Error("Failed to get cluster from DB")
+		return err
+	}
+
+	var totalHostsDoneStages, totalHostsStages float64
+	for _, h := range cluster.Hosts {
+		stages := m.hostAPI.GetStagesByRole(h.Role, h.Bootstrap)
+		currentIndex := m.hostAPI.IndexOfStage(h.Progress.CurrentStage, stages)
```
ori-amizur, line 689: Is -1 a valid index here?

ybettan: Yes, as you can see I add `+1`, that means that the first stage gets 1 instead of 0 and therefore we get a real positive percentage.

If the index doesn't exist, we will get `-1 + 1 = 0` therefore the progress will be `0`.

internal/cluster/cluster.go
ybettan: It is complete, but it was already updated at an earlier stage therefore we just need to update the new part (the other fields of the embedded struct in DB will remain unchanged.

ybettan: If you mean why I use the old data from DB instead of just writing `100` instead it is just because it felt better coding for me (if anyhow something didn't reach 100% we should reflect it in `totalPercentage` so we can fix it - otherwise we won't see the issue.

internal/cluster/cluster.go
ybettan: Same response as https://github.com/openshift/assisted-service/pull/2208/files/9ab80aad5efd3881d8ec06c362c071a221118f90#r673715328.

internal/host/host.go
ybettan: Same as https://github.com/openshift/assisted-service/pull/2208/files/9ab80aad5efd3881d8ec06c362c071a221118f90#r673712867

avishayt (approved):

avishayt (approved):

PR #2212, opened 2021-07-14
NO-ISSUE: Move ZTP flow to Ansible

djzager:
# Assisted Pull Request

## Description

This will allow developers in the future to better grok what is
happening in the ZTP flow by reading through an Ansible playbook that
mimicks what a user would do.

- [ ] New Feature
- [ ] Bug fix
- [x] Tests
- [x] Documentation
- [x] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @lranjbar 
/cc @YuviGold 

## Checklist

## Reviewers Checklist


djzager:
@lranjbar @YuviGold I sincerely doubt that this will work on my first rough attempt. Before I go through the effort of cleaning this up and expanding it to what remains of `deploy_spoke_cluster.sh`, I want to make sure others like this approach.

YuviGold:
In general, this looks good to me.

Some points:
* Could you elaborate on what's the benefit of moving to pure Ansible? Does it has any new advantages?
* I'd prefer maintaining the defaults on the shell script, as it is now. Otherwise, there would be two places with defaults (which we would like to avoid).
* How does the Ansible would apply the manifests? Does it ignore already created resources?
like I had to do for `oc get namespace "${SPOKE_NAMESPACE}" || oc create namespace "${SPOKE_NAMESPACE}"`
* Do the manifests being printed to stdout? 
being done with `    tee < "${manifest}" >(oc apply -f -)`
It is really helpful for debugging the workflow.

/cc @osherdp

djzager:
> * Could you elaborate on what's the benefit of moving to pure Ansible? Does it has any new advantages?
It's idempotent, easier to read, execution order is more clear. Modifying it as a developer to help me recreate scenarios is more straightforward.

> * I'd prefer maintaining the defaults on the shell script, as it is now. Otherwise, there would be two places with defaults (which we would like to avoid).
I wouldn't intend to have them defaulted in both. However, I think a lionshare of what exists currently in the shell script could simply be handled in the playbook.

> * How does the Ansible would apply the manifests? Does it ignore already created resources?
>   like I had to do for `oc get namespace "${SPOKE_NAMESPACE}" || oc create namespace "${SPOKE_NAMESPACE}"`
Ansible is idempotent. It would see that the namespace already exists and "skip" it.

> * Do the manifests being printed to stdout?
>   being done with `    tee < "${manifest}" >(oc apply -f -)`
>   It is really helpful for debugging the workflow.
I think it's right around `ansible-playbook -vvvv` that would show you the call to the k8s apiserver, but I don't remember clearly. I can tinker with this if it's important.

djzager:
This isn't really meant to be the start of an effort to replace all of our shell scripts with Ansible. However, scenarios like this one where what we use in test, could be useful as a developer, and as a sort of documentation on how to do the thing (here it's deploy a spoke cluster) I believe that writing Ansible is beneficial.

djzager:
Let me know what you think @YuviGold @osherdp . I don't want to clobber any plans that you have with this.

YuviGold:
> > * I'd prefer maintaining the defaults on the shell script, as it is now. Otherwise, there would be two places with defaults (which we would like to avoid).
> 
> I wouldn't intend to have them defaulted in both. However, I think a lionshare of what exists currently in the shell script could simply be handled in the playbook.
> 

IMO defaults should be at the higher place possible. In case the shell script is going to override the Ansible scripts, either way, then they should be defined in there.

> > * Do the manifests being printed to stdout?
> >   being done with `    tee < "${manifest}" >(oc apply -f -)`
> >   It is really helpful for debugging the workflow.
> 
> I think it's right around `ansible-playbook -vvvv` that would show you the call to the k8s apiserver, but I don't remember clearly. I can tinker with this if it's important.

Yes, please. That's really important when I'm looking at a backlog of a CI run to understand what was executed.

> Let me know what you think @YuviGold @osherdp . I don't want to clobber any plans that you have with this.

Just fix the points above, and that's good for me. :)

djzager:
this isn't on the top of my priority list but I should be able to rebase it today and address the review comments.

deploy/operator/ztp/assisted-installer-crds-playbook.yaml
```diff
@@ -1,55 +1,182 @@
...
-      src: "baremetalHost.j2"
-      dest: "generated/baremetalHost.yaml"
+  - name: Create spoke namespace
+    community.kubernetes.k8s:
+      state: present
+      definition:
+        apiVersion: v1
+        kind: Namespace
+        metadata:
+          name: "{{ spoke_namespace }}"
```
pawanpinjarkar, line 34: Please change it to 
```
- name: Create spoke namespace
  community.kubernetes.k8s:
    state: present
    api_version: v1
    kind: Namespace
    name: "{{ spoke_namespace }}"
```
Ref: https://docs.ansible.com/ansible/latest/collections/community/kubernetes/k8s_module.html

deploy/operator/ztp/assisted-installer-crds-playbook.yaml
```diff
@@ -1,55 +1,182 @@
...
+        spec:
+          clusterDeploymentRef:
+            name: "{{ cluster_deployment_name }}"
+          imageSetRef:
+            name: "{{ cluster_image_set_name }}"
+          apiVIP: ""
+          ingressVIP: ""
+          networking:
+            clusterNetwork:
+            - cidr: {{ cluster_subnet }}
```
pawanpinjarkar, line 138: `"`{{ cluster_subnet }}`"`

deploy/operator/ztp/assisted-installer-crds-playbook.yaml
```diff
@@ -1,55 +1,182 @@
...
+              hostPrefix: {{ cluster_host_prefix }}
```
pawanpinjarkar, line 139: `"`{{ cluster_host_prefix }}`"`

deploy/operator/ztp/assisted-installer-crds-playbook.yaml
```diff
@@ -1,55 +1,182 @@
...
+            machineNetwork:
+            - cidr: {{ external_subnet }}
```
pawanpinjarkar, line 141: `"`{{ external_subnet }}`"`

deploy/operator/ztp/assisted-installer-crds-playbook.yaml
```diff
@@ -1,55 +1,182 @@
...
+          labels:
+            infraenvs.agent-install.openshift.io: "{{ infraenv_name }}"
+          annotations:
+            bmac.agent-install.openshift.io/hostname: "{{ item.name }}"
+        spec:
+          online: true
+          bootMACAddress: "{{ item.ports[0].address }}"
+          bmc:
+            address: "{{ item.driver_info.address}}"
+            credentialsName: "{{ item.name }}-bmc-secret'
```
pawanpinjarkar, line 181: "{{ item.name }}-bmc-secret`"`

deploy/operator/ztp/assisted-installer-crds-playbook.yaml
```diff
@@ -1,55 +1,182 @@
...
-    template:
-      src: "agentClusterInstall.j2"
-      dest: "generated/agentClusterInstall.yaml"
-
-  - name: write the baremetalHost crd
-    template:
-      src: "baremetalHost.j2"
-      dest: "generated/baremetalHost.yaml"
+  - name: Create spoke namespace
+    community.kubernetes.k8s:
```
pawanpinjarkar, line 28: Got below error- 
`ERROR! couldn't resolve module/action 'community.kubernetes.k8s'. This often indicates a misspelling, missing collection, or incorrect module path.`

Ref: https://github.com/ansible-collections/kubernetes.core/issues/159

lranjbar:
>>> I'd prefer maintaining the defaults on the shell script, as it is now. Otherwise, there would be two places with defaults (which we would like to avoid).
>>
>> I wouldn't intend to have them defaulted in both. However, I think a lionshare of what exists currently in the shell script could simply be handled in the playbook.
>
> IMO defaults should be at the higher place possible. In case the shell script is going to override the Ansible scripts, either way, then they should be defined in there.

I would like to move the configuration of the variables into a file so that we can easily maintain different values for different release branches. @YuviGold @osherdp @djzager WDYT?

We can do it in the another PR I just think it might be a good place to put the variable defaults.

flaper87:
Closing this one as it's not a priority and @djzager has moved on to a different team

/close

PR #2224, opened 2021-07-14
WIP MGMT-7148: Add webhooks validation for ACI

rollandf:

# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


djzager:
Either the webhook(s):
* need to come out of the box when someone subscribes to the `infrastructure-operator`, this would require a move into `config/` and updates the clusterserviceversion. I know that there are a bunch of issues around OLM's support for webhooks so this may not be the best option
* be deployed by the agentserviceconfig_controller directly as part of the procedure for deploying the assisted-service. I think this is the correct path.

rollandf:
> Either the webhook(s):
> 
>     * need to come out of the box when someone subscribes to the `infrastructure-operator`, this would require a move into `config/` and updates the clusterserviceversion. I know that there are a bunch of issues around OLM's support for webhooks so this may not be the best option
> 
>     * be deployed by the agentserviceconfig_controller directly as part of the procedure for deploying the assisted-service. I think this is the correct path.

Thanks for the comment.
Yes, agree. Operator support will be in a different Jira: https://issues.redhat.com/browse/MGMT-7147

filanov (commented):
Do you think that we should add the deployment to `enable-kube-api-for-subsystem` ?

tools/deploy_webhooks.py
```diff
@@ -0,0 +1,138 @@
...
+import utils
+import yaml
+import deployment_options
+import textwrap
+
+log = utils.get_logger('deploy_webhooks')
+
+def main():
+    deploy_options = deployment_options.load_deployment_options()
+    if deploy_options.enable_kube_api:
```
nmagnezi, line 11: maybe
```python3
if not deploy_options.enable_kube_api:
    return
```
so you won't have to indent this entire logic

nmagnezi: also, note that this if statement will always return `True`, won't that create a problem?
Consider this example: https://gist.github.com/nmagnezi/8e307cbf0bb8c53f19000fb49a275fc4

Testing it:
```sh
$ python3 args.py --enable-kube-api xyz 
in if deploy_options.enable_kube_api statement
$ python3 args.py --enable-kube-api True
in if deploy_options.enable_kube_api statement
$ python3 args.py --enable-kube-api Flase
in if deploy_options.enable_kube_api statement
```

nmagnezi (changes requested):

tools/deploy_webhooks.py
rollandf: I see the same pattern in a lot of Python scripts under tools.
I am not sure that this is an issue.

djzager:
> Thanks for the comment.
> Yes, agree. Operator support will be in a different Jira: https://issues.redhat.com/browse/MGMT-7147

is the idea that the operator is going to consume the yamls being put in the `deploy/` directory? We haven't done that for any resource yet.

I really do feel like I'm missing something fundamental with the operator support being in a separate Jira. Webhooks are only valid for operator managed deployments (ie. not cloud), but with this PR, there are no webhooks installed when using the operator to install the assisted service. From my perspective, if I install assisted-service via the `infrastructure-operator` either OLM managed or via ACM and there are no validating webhooks; we haven't added validating webhooks. Is the idea that we install them manually after the fact?

rollandf:
> > Thanks for the comment.
> > Yes, agree. Operator support will be in a different Jira: https://issues.redhat.com/browse/MGMT-7147
> 
> is the idea that the operator is going to consume the yamls being put in the `deploy/` directory? We haven't done that for any resource yet.
> 
> I really do feel like I'm missing something fundamental with the operator support being in a separate Jira. Webhooks are only valid for operator managed deployments (ie. not cloud), but with this PR, there are no webhooks installed when using the operator to install the assisted service. From my perspective, if I install assisted-service via the `infrastructure-operator` either OLM managed or via ACM and there are no validating webhooks; we haven't added validating webhooks. Is the idea that we install them manually after the fact?

This is needed for subsystem tests, test-infra, or local development.
I did not want to make this PR bigger, and will implement the operator later.
I will move the yamls to config in the next PR if this the right place for the operator to consume from.

rollandf:
Output while editing manually ACI, after install started:

```
# kubectl edit agentclusterinstalls.extensions.hive.openshift.io -n assisted-installer test-infra-cluster-assisted-installer-agent-cluster-install 
error: agentclusterinstalls.extensions.hive.openshift.io "test-infra-cluster-assisted-installer-agent-cluster-install" could not be patched: admission webhook "agentclusterinstallvalidators.admission.agentinstall.openshift.io" denied the request: Attempted to change AgentClusterInstall.Spec which is immutable after install started, except for ClusterMetadata fields. Unsupported change: 
	APIVIP: (192.168.126.100 => 192.168.126.10)
```

docs/hive-integration/kube-api-webhook.md
```diff
@@ -0,0 +1,49 @@
...
+Admission webhooks are HTTP callbacks that receive admission requests and do something with them. There are two types of admission webhooks: validating admission webhook and mutating admission webhook.
+
+- Mutating Admission Webhook: These can mutate the object while it’s being created or updated, before it gets stored. It can be used to default fields in a resource requests, e.g. fields in Deployment that are not specified by the user.
+
+- Validating Admission Webhook: These can validate the object while it’s being created or updated, before it gets stored. It allows more complex validation than pure schema-based validation. e.g. cross-field validation.
+
+Read more about admission webhooks [here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/).
+
+Webhook validation enable adding validations based on the state of the CRDs.
+For example, changing the Spec of AgentClusterInstall or Agent after the installation started is not permitted.
```
lranjbar, line 12: I think we can add some info about why I would run this webhook. From conversations I gathered that we are using this validating admission web hook to validate the CRDs before starting the installation. This isn't really clear here. Also I would remove the mutating webhook unless there are plans to add one (to fix configuration, maybe)? It could be confusing if our webhook isn't a mutating one and we mention that.

deploy/webhooks/agentclusterinstall-webhook.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: agentclusterinstallvalidators.admission.agentinstall.openshift.io
+webhooks:
+- name: agentclusterinstallvalidators.admission.agentinstall.openshift.io
+  clientConfig:
+    service:
+      # reach the webhook via the registered aggregated API
+      namespace: default
```
mhrivnak, line 10: These seem unlikely to be what we want. Should they be changed?

deploy/webhooks/agentinstalladmission_rbac_role.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  annotations:
+  name: system:openshift:assisted-installer:agentinstalladmission
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
```
mhrivnak, line 10: What does it do with configmaps?

docs/hive-integration/kube-api-webhook.md
```diff
@@ -0,0 +1,49 @@
+# Webhook validation
```
mhrivnak, line 1: Some of this text appears to be lifted from here: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/admission-webhook.md

We cannot copy text without attribution. How about just linking to k8s docs that explain these concepts?

docs/hive-integration/kube-api-webhook.md
```diff
@@ -0,0 +1,49 @@
...
+- Mutating Admission Webhook: These can mutate the object while it’s being created or updated, before it gets stored. It can be used to default fields in a resource requests, e.g. fields in Deployment that are not specified by the user.
+
+- Validating Admission Webhook: These can validate the object while it’s being created or updated, before it gets stored. It allows more complex validation than pure schema-based validation. e.g. cross-field validation.
+
+Read more about admission webhooks [here](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/).
+
+Webhook validation enable adding validations based on the state of the CRDs.
+For example, changing the Spec of AgentClusterInstall or Agent after the installation started is not permitted.
+
+## Implementation
```
mhrivnak, line 14: The rest of this doc looks relevant only to developer environments. We'll never expect an end-user to deploy a webhook like this, or even have much awareness that webhooks exist. How about moving this doc into https://github.com/openshift/assisted-service/tree/master/docs/dev ?

pkg/validating-webhooks/hiveextension/v1beta1/agentclusterinstall_admission_hook.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	// We're only validating creates and updates at this time, so all other operations are explicitly allowed.
+	contextLogger.Info("Successful validation")
+	return &admissionv1.AdmissionResponse{
+		Allowed: true,
+	}
+}
+
+// shouldValidate explicitly checks if the request should validated. For example, this webhook may have accidentally been registered to check
```
mhrivnak, line 109: s/should/should be/

pkg/validating-webhooks/hiveextension/v1beta1/agentclusterinstall_admission_hook.go
```diff
@@ -0,0 +1,292 @@
...
+	contextLogger := log.WithFields(log.Fields{
+		"operation": admissionSpec.Operation,
+		"group":     admissionSpec.Resource.Group,
+		"version":   admissionSpec.Resource.Version,
+		"resource":  admissionSpec.Resource.Resource,
+		"method":    "validateCreate",
+	})
+
+	aci := &hiveext.AgentClusterInstall{}
+	if err := a.decoder.DecodeRaw(admissionSpec.Object, aci); err != nil {
```
mhrivnak, line 151: Are we just testing that we can unmarshal the object? Do we need to validate creation at all then?

pkg/validating-webhooks/hiveextension/v1beta1/agentclusterinstall_admission_hook.go
```diff
@@ -0,0 +1,292 @@
...
+	agentClusterInstallGroup    = "extensions.hive.openshift.io"
+	agentClusterInstallVersion  = "v1beta1"
+	agentClusterInstallResource = "agentclusterinstalls"
+
+	agentClusterInstallAdmissionGroup   = "admission.agentinstall.openshift.io"
+	agentClusterInstallAdmissionVersion = "v1"
+)
+
+var (
+	mutableFields = []string{"ClusterMetadata"}
```
mhrivnak, line 30: Eventually this should include things in `Compute`, for example so workers can be created with different settings for hyperthreading, or any other settings we put in there (like the custom ignition endpoint we discussed earlier today). I'm just stating this to confirm the design will accommodate that.

mhrivnak (changes requested):

deploy/webhooks/agentinstalladmission_rbac_role.yaml
rollandf: This is needed to access the `extension-apiserver-authentication` configmap

pkg/validating-webhooks/hiveextension/v1beta1/agentclusterinstall_admission_hook.go
rollandf: What fields do you think can be changed after the installation started?
If you refer for day2 operations, then the `installAlreadyStarted` function will return false on Day2 clusters.

pkg/validating-webhooks/hiveextension/v1beta1/agentclusterinstall_admission_hook.go
rollandf: Done

pkg/validating-webhooks/hiveextension/v1beta1/agentclusterinstall_admission_hook.go
rollandf: Agree, no validation on Create for now, so I will remove this.

pkg/validating-webhooks/hiveextension/v1beta1/agentclusterinstall_admission_hook.go
mhrivnak: Ah, so this is only preventing change during installation, but allows changes again after installation finishes? That'll work. It would help to make that distinction clear in the docs, such as `docs/hive-integration/README.md`.

deploy/webhooks/agentclusterinstall-webhook.yaml
djzager: This is only for the subsystem tests. Deploying this via the operator is covered in https://issues.redhat.com/browse/MGMT-7147

docs/hive-integration/kube-api-webhook.md
rollandf: Done

docs/hive-integration/kube-api-webhook.md
rollandf: Done

docs/hive-integration/kube-api-webhook.md
rollandf: Done

pkg/validating-webhooks/hiveextension/v1beta1/agentclusterinstall_admission_hook.go
rollandf: Done

pkg/validating-webhooks/hiveextension/v1beta1/agentclusterinstall_admission_hook.go
rollandf: Done

djzager (commented):
This is taking shape. My only real concern at this stage is the amount of permission our operator requires, it may be worth considering (if only briefly) whether we can/want to let OLM deploy this stuff for us. It may not be worth the added complexity but worth talking about.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -100,6 +103,11 @@ type AgentServiceConfigReconciler struct {
 // +kubebuilder:rbac:groups=route.openshift.io,resources=routes,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=ClusterRole,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=ClusterRoleBinding,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
mhrivnak: Good catch. It doesn't look like the webhook even has an API client. Do we really need to create a cluster role for it? Ideally it takes all of its input in the body of the http request.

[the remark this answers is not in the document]
rollandf: @mhrivnak what do you mean by "It doesn't look like the webhook even has an API client" ?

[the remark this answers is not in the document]
rollandf: Reference of configuration is from hive: https://github.com/openshift/hive/tree/master/config/hiveadmission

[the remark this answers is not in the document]
mhrivnak: When I looked at the code for the admission webhook, I didn't see anywhere that it establishes a connection to the k8s API service.

deploy/operator/ztp/deploy_spoke_cluster.sh
```diff
@@ -89,5 +89,8 @@ echo "Cluster installation has been stopped (either for good or bad reasons)"
 wait_for_condition "agentclusterinstall/${ASSISTED_AGENT_CLUSTER_INSTALL_NAME}" "Completed" "1m" "${SPOKE_NAMESPACE}"
 echo "Cluster has been installed successfully!"
 
+echo "Test webhook - try to update ACI after install"
+oc -n ${SPOKE_NAMESPACE} patch agentclusterinstall.extensions.hive.openshift.io ${ASSISTED_AGENT_CLUSTER_INSTALL_NAME} -p '{"spec":{"provisionRequirements":{"controlPlaneAgents":"7"}}}' --type merge
+
```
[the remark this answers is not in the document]
lranjbar: It isn't really clear why this is being changed as well. Please add to the description the reason the base CI script is being changed here. This will effect all CI jobs so we need to understand that is actually desirable.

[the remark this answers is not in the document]
rollandf: This is still WIP.
As we discussed some time ago, this is the only place where I can test the webhook.
The idea is to attempt to change the ACI after it is installed and verify that it will be revoked.

I need to check the return value and validate that it fails.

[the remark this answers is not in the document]
lranjbar: You will need to make a new CI job for this test case. I suggest removing this until you understand how to add this check without breaking all the existing jobs with this change.

docs/hive-integration/README.md
```diff
@@ -25,6 +25,8 @@ In the AgentClusterInstall, the user can specify requirements like Networking, n
 
 The installation will start automatically if the required number of hosts is available, the hosts are ready to be installed and the Agents are approved.
 
+Once the installation started, changes to the AgentClusterInstall Spec will be revoked.
```
djzager, line 28:
```suggestion
Once the installation started, changes to the AgentClusterInstall Spec will be rejected.
```

djzager (commented):
One nitpick. Otherwise this looks good to me.

PR #2232, opened 2021-07-16
MGMT-7235 - change AI implementation to use InfraEnv instead of Image…

yevgeny-shnaidman:
…Info

Moving the AI code to use InfraEnv instead of Cluster.ImageInfo both for
storing ISO creation parameters and for Disocvery ISO ignition generation
At this stage InfraEnv ID is the same as the Cluster ID it creates the ISO
for

# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -245,12 +245,12 @@ func (b *bareMetalInventory) updatePullSecret(pullSecret string, log logrus.Fiel
 func (b *bareMetalInventory) GetDiscoveryIgnition(ctx context.Context, params installer.GetDiscoveryIgnitionParams) middleware.Responder {
 	log := logutil.FromContext(ctx, b.log)
 
-	c, err := b.getCluster(ctx, params.ClusterID.String())
+	infraEnv, err := common.GetInfraEnvFromDB(b.db, params.ClusterID)
 	if err != nil {
 		return common.GenerateErrorResponder(err)
 	}
 
-	cfg, err := b.IgnitionBuilder.FormatDiscoveryIgnitionFile(c, b.IgnitionConfig, false, b.authHandler.AuthType())
+	cfg, err := b.IgnitionBuilder.FormatDiscoveryIgnitionFile(infraEnv, b.IgnitionConfig, false, b.authHandler.AuthType())
 	if err != nil {
 		log.WithError(err).Error("Failed to format ignition config")
 		return common.GenerateErrorResponder(err)
@@ -271,8 +271,10 @@ func (b *bareMetalInventory) UpdateDiscoveryIgnition(ctx context.Context, params
 func (b *bareMetalInventory) UpdateDiscoveryIgnitionInternal(ctx context.Context, params installer.UpdateDiscoveryIgnitionParams) error {
 	log := logutil.FromContext(ctx, b.log)
 
+	//[TODO] - change the code to use InfraEnv once we move the code and test to use InfraEnv CRUD
 	c, err := b.getCluster(ctx, params.ClusterID.String())
 	if err != nil {
+		log.WithError(err).Errorf("Failed to get cluster %s", params.ClusterID)
 		return err
 	}
 
@@ -475,16 +477,6 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 		KubeKeyName:             kubeKey.Name,
 		KubeKeyNamespace:        kubeKey.Namespace,
 		TriggerMonitorTimestamp: time.Now(),
-	}
-
-	proxyHash, err := computeClusterProxyHash(params.NewClusterParams.HTTPProxy,
-		params.NewClusterParams.HTTPSProxy,
-		params.NewClusterParams.NoProxy)
-	if err != nil {
-		err = errors.Wrapf(err, "failed to compute cluster proxy hash")
-		return nil, common.NewApiError(http.StatusInternalServerError, errors.Errorf("Failed to compute cluster proxy hash"))
-	} else {
-		cluster.ProxyHash = proxyHash
 	}
 
 	pullSecret := swag.StringValue(params.NewClusterParams.PullSecret)
@@ -740,17 +732,17 @@ func (b *bareMetalInventory) DeregisterClusterInternal(ctx context.Context, para
 
 func (b *bareMetalInventory) DownloadClusterISO(ctx context.Context, params installer.DownloadClusterISOParams) middleware.Responder {
 	log := logutil.FromContext(ctx, b.log)
-	var cluster common.Cluster
 
-	if err := b.db.First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
-		log.WithError(err).Errorf("failed to get cluster %s", params.ClusterID)
-		return common.NewApiError(http.StatusNotFound, err)
+	infraEnv, err := common.GetInfraEnvFromDB(b.db, params.ClusterID)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get infra env %s", params.ClusterID)
+		return common.GenerateErrorResponder(err)
 	}
 
-	imgName := getImageName(*cluster.ID)
+	imgName := getImageName(infraEnv.ID)
 	exists, err := b.objectHandler.DoesObjectExist(ctx, imgName)
 	if err != nil {
-		log.WithError(err).Errorf("Failed to get ISO for cluster %s", cluster.ID.String())
+		log.WithError(err).Errorf("Failed to get ISO for cluster %s", infraEnv.ID.String())
 		b.eventsHandler.AddEvent(ctx, params.ClusterID, nil, models.EventSeverityError,
 			"Failed to download image: error fetching from storage backend", time.Now())
 		return installer.NewDownloadClusterISOInternalServerError().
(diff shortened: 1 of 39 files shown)


internal/common/db.go
```diff
@@ -199,6 +201,16 @@ func GetHostFromDBWhere(db *gorm.DB, where ...interface{}) (*Host, error) {
 	return &host, nil
 }
 
+func GetInfraEnvFromDB(db *gorm.DB, infraEnvID strfmt.UUID) (*InfraEnv, error) {
+	var infraEnv InfraEnv
+
+	err := db.First(&infraEnv, "id = ?", infraEnvID.String()).Error
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get infra env %s", infraEnvID)
```
filanov, line 209: Wrapping the error will hide the original error, i suggest to make check if the error is not found and return `ApiError` or the original error.

yevgeny-shnaidman: this is an infrastructure function. I think it is more appropriate to make the decision (ApiError| original error) at the upper level. this is the way GetClusterFromDB is also implemented, and the examination can be done by using errors.Is

filanov: Does error is work on a wrapped error?

yevgeny-shnaidman: yes

yevgeny-shnaidman: // Is reports whether any error in err's chain matches target.
//
// The chain consists of err itself followed by the sequence of errors obtained by
// repeatedly calling Unwrap.

internal/bminventory/inventory.go
```diff
@@ -732,17 +724,17 @@ func (b *bareMetalInventory) DeregisterClusterInternal(ctx context.Context, para
 
 func (b *bareMetalInventory) DownloadClusterISO(ctx context.Context, params installer.DownloadClusterISOParams) middleware.Responder {
 	log := logutil.FromContext(ctx, b.log)
-	var cluster common.Cluster
+	var infraEnv common.InfraEnv
 
-	if err := b.db.First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
-		log.WithError(err).Errorf("failed to get cluster %s", params.ClusterID)
+	if err := b.db.First(&infraEnv, "id = ?", params.ClusterID).Error; err != nil {
```
filanov, line 729: maybe it should be a helper function? or maybe use `GetInfraEnvFromDB`?

internal/bminventory/inventory.go
```diff
@@ -732,17 +724,17 @@ func (b *bareMetalInventory) DeregisterClusterInternal(ctx context.Context, para
...
+		log.WithError(err).Errorf("failed to get infra env %s", params.ClusterID)
 		return common.NewApiError(http.StatusNotFound, err)
```
filanov, line 731: It can be internal error as well. 
```suggestion
		return common.GenerateErrorResponder(err)
```

internal/bminventory/inventory.go
```diff
@@ -773,17 +765,17 @@ func (b *bareMetalInventory) DownloadClusterISO(ctx context.Context, params inst
...
 func (b *bareMetalInventory) DownloadClusterISOHeaders(ctx context.Context, params installer.DownloadClusterISOHeadersParams) middleware.Responder {
 	log := logutil.FromContext(ctx, b.log)
-	var cluster common.Cluster
+	var infraEnv common.InfraEnv
 
-	if err := b.db.First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
-		log.WithError(err).Errorf("failed to get cluster %s", params.ClusterID)
+	if err := b.db.First(&infraEnv, "id = ?", params.ClusterID).Error; err != nil {
+		log.WithError(err).Errorf("failed to get infra env %s", params.ClusterID)
 		return common.NewApiError(http.StatusNotFound, err)
```
filanov, line 772: It can be internal error as well. 
```suggestion
		return common.GenerateErrorResponder(err)
```

internal/bminventory/inventory.go
```diff
@@ -773,17 +765,17 @@ func (b *bareMetalInventory) DownloadClusterISO(ctx context.Context, params inst
 
 func (b *bareMetalInventory) DownloadClusterISOHeaders(ctx context.Context, params installer.DownloadClusterISOHeadersParams) middleware.Responder {
 	log := logutil.FromContext(ctx, b.log)
-	var cluster common.Cluster
+	var infraEnv common.InfraEnv
 
-	if err := b.db.First(&cluster, "id = ?", params.ClusterID).Error; err != nil {
-		log.WithError(err).Errorf("failed to get cluster %s", params.ClusterID)
+	if err := b.db.First(&infraEnv, "id = ?", params.ClusterID).Error; err != nil {
```
filanov, line 770: maybe it should be a helper function? or maybe use `GetInfraEnvFromDB`?

internal/bminventory/inventory.go
```diff
@@ -2505,12 +2535,28 @@ func (b *bareMetalInventory) GetClusterInternal(ctx context.Context, params inst
...
 	for _, host := range cluster.Hosts {
-		if err := b.customizeHost(host); err != nil {
+		if err = b.customizeHost(host); err != nil {
 			return nil, err
 		}
 		// Clear this field as it is not needed to be sent via API
 		host.FreeAddresses = ""
 	}
+
+	infraEnv, err := common.GetInfraEnvFromDB(b.db, *cluster.ID)
```
filanov, line 2545: Probably need to handle the error?

internal/controller/controllers/agent_controller.go
```diff
@@ -562,6 +562,7 @@ func requirementsMet(agent *aiv1beta1.Agent, status string) {
 func (r *AgentReconciler) updateNtpSources(log logrus.FieldLogger, host *models.Host, agent *aiv1beta1.Agent) error {
 	if host.NtpSources == "" {
 		log.Debugf("Skip update NTP Sources: Host %s NTP sources not set", agent.Name)
+		log.Infof("Skip update NTP Sources: Host %s NTP sources not set", agent.Name)
```
ravidbro, line 565: Do we need both lines?

internal/bminventory/inventory.go
```diff
@@ -912,43 +904,74 @@ func (b *bareMetalInventory) GenerateClusterISOInternal(ctx context.Context, par
...
 
+	if !cluster.PullSecretSet {
+		errMsg := "Can't generate cluster ISO without pull secret"
+		log.Error(errMsg)
+		return nil, common.NewApiError(http.StatusBadRequest, errors.New(errMsg))
+	}
+
+	infraEnv, err := common.GetInfraEnvFromDB(tx, params.ClusterID)
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: createInfraEnv will call this API, as does controller. Unitl controller is moved to support infra env generation, this code will stay in this function, but once controller is updated, we can remove it. Another option is to break up this function in 2 layers,the upper one will create infra env. controller will call the upper layer, while createInfraEnv will call the lower layer, and once controller is updated we can remove the upper layer. i don't see much difference between those 2 options

internal/controller/controllers/agent_controller.go
yevgeny-shnaidman: actually i added the second line to debug the subsystem failure, i will remove it

internal/migrations/20210713123129_populate_infra_env_test.go
```diff
@@ -47,8 +47,9 @@ var _ = Describe("populateInfraEnv", func() {
 		hostID = strfmt.UUID(uuid.New().String())
 		clusterID = strfmt.UUID(uuid.New().String())
 		host := models.Host{
-			ID:        &hostID,
-			ClusterID: clusterID,
+			ID:         &hostID,
+			InfraEnvID: clusterID,
+			ClusterID:  clusterID,
```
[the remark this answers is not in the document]
yevgeny-shnaidman: in my swagger infra env is already defined as primary key, so it host cannot be created with empty InfraEnvID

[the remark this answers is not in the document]
ravidbro: but since this is running before AutoMigrate then the primary key shouldn't be defined yet,
do you get an error if you remove that line?

internal/bminventory/inventory.go
yevgeny-shnaidman: fixed

internal/bminventory/inventory.go
yevgeny-shnaidman: fixed

internal/bminventory/inventory.go
yevgeny-shnaidman: fixed

internal/bminventory/inventory.go
yevgeny-shnaidman: fixed

internal/bminventory/inventory.go
yevgeny-shnaidman: fixed

internal/migrations/20210713123129_populate_infra_env_test.go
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed with dummy value

internal/common/db.go
dylwong: The InfraEnv AI service resource to not be created until the discovery ISO is first downloaded for a given cluster. As a result, the ignition download API endpoints are not available following a single call to the cluster creation API. This is problem for my Ansible create-cluster tasks, I am downloading the ignition before the discovery iso. What steps do I need to change in my tasks with this PR?

filanov: Thanks for the comment, @yevgeny-shnaidman will handle this issue in a different PR.

dylwong: Thank you, I want to make clear also that I need to patch the discovery iso ignition before downloading it, so downloading it twice woudln't make much sense

PR #2248, opened 2021-07-19
Bug 1974077: Missing source hostname in validation error

jordigilh:
# Assisted Pull Request

## Description

When network latency and packet loss validations fail, the user gets an error message that contains the source hostname and the target hostnames against which the validations failed. 

```
"Network latency requirements of less or equals than 100.000 ms not met for connectivity between and test-infra-cluster-gfdhfg-master-0"
```
It should say something like this:
```
"Network latency requirements of less or equals than 100.000 ms not met for connectivity between test-infra-cluster-gfdhfg-master-1 and test-infra-cluster-gfdhfg-master-0"
```

The source hostname is retrieved from the `host.RequestedHostname` field, which it seems it's not always populated at validation time. This change retrieves the source host name from `host.RequstedHostname` and, if empty, from `host.Inventory.Hostname`.

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/assign @ori-amizur 
/cc @lalon4 

## Checklist

## Reviewers Checklist


internal/host/validator.go
```diff
@@ -700,6 +700,10 @@ func (v *validator) hasSufficientNetworkLatencyRequirementForRole(c *validationC
 		return ValidationSuccess
 	}
 
+	if len(c.host.Inventory) == 0 {
+		return ValidationPending
+	}
+
 	if len(c.host.Connectivity) == 0 {
 		return ValidationFailure
```
jordigilh, line 708: I'm changing this to be `ValidationPending` in #2251

internal/host/validator.go
```diff
@@ -747,7 +751,11 @@ func (v *validator) printSufficientNetworkLatencyRequirementForRole(c *validatio
 			return fmt.Sprintf("Error while attempting to validate network latency: %s", err)
 		}
 		sort.Strings(hostNames)
-		return fmt.Sprintf("Network latency requirements of less or equals than %.3f ms not met for connectivity between %s and %s.", *c.clusterHostRequirements.Total.NetworkLatencyThresholdMs, c.host.RequestedHostname, strings.Join(hostNames, ","))
+		inv, err := hostutil.UnmarshalInventory(c.host.Inventory)
```
ori-amizur, line 754: The inventory is already part of the validationContext

internal/host/validator.go
```diff
@@ -760,6 +768,10 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
 		return ValidationSuccess
 	}
 
+	if len(c.host.Inventory) == 0 {
```
ori-amizur, line 771: I am not sure we should say the validation is pending just because we cannot print the hostname in the validation message.  You can use the id instead.

internal/host/validator.go
jordigilh: If it's part of the context, do you still think it's better to use hostID instead of the hostname in the error message? If I don't need to unmarshal the inventory, it is not going to be  expensive to retrieve the hostname.

internal/host/validator.go
jordigilh: Ok, I'll use the ID then. Ignore my previous comment then :smile:

internal/host/validator.go
jordigilh: see my comment below, I will use the hostID instead.

jordigilh: Done.

internal/host/validator.go
jordigilh: Removed this conditional. Using the host.ID removes the need for the inventory availability.

jordigilh:
@ori-amizur I've implemented the suggested changes. Please review and approve if you have no more comments.

internal/host/validator.go
```diff
@@ -747,7 +747,9 @@ func (v *validator) printSufficientNetworkLatencyRequirementForRole(c *validatio
 			return fmt.Sprintf("Error while attempting to validate network latency: %s", err)
 		}
 		sort.Strings(hostNames)
-		return fmt.Sprintf("Network latency requirements of less or equals than %.3f ms not met for connectivity between %s and %s.", *c.clusterHostRequirements.Total.NetworkLatencyThresholdMs, c.host.RequestedHostname, strings.Join(hostNames, ","))
+		return fmt.Sprintf("Network latency requirements of less or equals than %.3f ms not met for connectivity between %s and %s.", *c.clusterHostRequirements.Total.NetworkLatencyThresholdMs, c.host.ID, strings.Join(hostNames, ","))
+	case ValidationPending:
```
ori-amizur, line 751: The validation function does not return pending for connectivity

internal/host/validator.go
```diff
@@ -808,7 +810,9 @@ func (v *validator) printSufficientPacketLossRequirementForRole(c *validationCon
 			return fmt.Sprintf("Error while attempting to validate packet loss validation: %s", err)
 		}
 		sort.Strings(hostNames)
-		return fmt.Sprintf("Packet loss percentage requirement of less or equals than %.2f%% not met for connectivity between %s and %s.", *c.clusterHostRequirements.Total.PacketLossPercentage, c.host.RequestedHostname, strings.Join(hostNames, ","))
+		return fmt.Sprintf("Packet loss percentage requirement of less or equals than %.2f%% not met for connectivity between %s and %s.", *c.clusterHostRequirements.Total.PacketLossPercentage, c.host.ID, strings.Join(hostNames, ","))
+	case ValidationPending:
```
ori-amizur, line 814: The validation function does not return pending for connectivity

jordigilh: you're right. I mixed it up with #2251

PR #2257, opened 2021-07-21
MGMT-7141: Add VMWare to feature_usage

eliorerz:
Feature usage is set to show us if different platform (rather then baremetal) is set and if credentials are set.
https://issues.redhat.com/browse/MGMT-7141

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

/cc @slaviered
/cc @gamli75 
/cc @michaellevy101

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -1933,7 +1933,7 @@ func (b *bareMetalInventory) updateDhcpNetworkParams(updates map[string]interfac
 	}
 	return nil
 }
-func (b *bareMetalInventory) updatePlatformParams(params installer.UpdateClusterParams, cluster *common.Cluster, updates map[string]interface{}) {
+func (b *bareMetalInventory) updatePlatformParams(params installer.UpdateClusterParams, updates map[string]interface{}) {
 	platform := params.ClusterUpdateParams.Platform
 	if platform == nil || platform.Type == "" {
 		return
@@ -1979,7 +1979,7 @@ func (b *bareMetalInventory) updateClusterData(_ context.Context, cluster *commo
 
 	b.setProxyUsage(params.ClusterUpdateParams.HTTPProxy, params.ClusterUpdateParams.HTTPSProxy, params.ClusterUpdateParams.NoProxy, usages)
 
-	b.updatePlatformParams(params, cluster, updates)
+	b.updatePlatformSources(params, updates, usages)
 
 	if err = b.updateNetworkParams(params, cluster, updates, usages, log, interactivity); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


internal/bminventory/inventory.go
```diff
@@ -2135,6 +2135,23 @@ func setCommonUserNetworkManagedParams(params *models.ClusterUpdateParams, singl
 	return nil, false
 }
 
+func (b *bareMetalInventory) updatePlatformSources(params installer.UpdateClusterParams, cluster *common.Cluster, updates map[string]interface{}, usages map[string]models.Usage) {
+	b.updatePlatformParams(params, cluster, updates)
+
+	if params.ClusterUpdateParams.Platform.Type != models.PlatformTypeBaremetal {
```
gamli75, line 2141: we should update the platform type not only for vsphere so this condition is not necessary.
we should always call b.setUsage(true, usage.PlatformSelectionUsage, &map[string]interface{}{
			"platform_type": params.ClusterUpdateParams.Platform.Type, "with_credentials": withCredentials}, usages)

internal/usage/consts.go
```diff
@@ -17,4 +17,6 @@ const (
 	DiskSelectionUsage string = "Disk Selection"
 	//user networkType selection
 	NetworkTypeSelectionUsage string = "NetworkType"
+	//usage of platform provider other than baremetal
+	PlatformSelectionUsage string = "Platform selection"
```
gamli75, line 21: Please change it to platform type
```suggestion
	PlatformTypeUsage string = "Platform type"
```

gamli75:
@eliorerz please fix the PR title - it should be MGMT-7141 and not MGMT-7280

internal/usage/consts.go
eliorerz: @gamli75  This is not just platform type. It for changing the platform (platform type different than baremetal + has credentials). I thought on "platform type" but it's not that accurate. WDYT?

internal/bminventory/inventory.go
eliorerz: If nothing is changed why should we call it? We want to monitor the usage of this feature. If platform type is Baremetal you define it as feature usage?

slaviered: I agree with @eliorerz the usage is when we define a non bare metal platform. In the additional widget we will also break it up to the platform used and the relative percentage of the different platforms.

gamli75: we use the feature usage in our data dashboard and we would like to see in the future - how many clusters were with baremetal, vsphere, libvirt ... it will make our data analysis much easier

internal/usage/consts.go
slaviered: +1 for Platform selection

gamli75: agree

internal/bminventory/inventory.go
```diff
@@ -2135,6 +2135,23 @@ func setCommonUserNetworkManagedParams(params *models.ClusterUpdateParams, singl
...
+		withCredentials := false
+
+		if params.ClusterUpdateParams.Platform.Type == models.PlatformTypeVsphere {
+			if params.ClusterUpdateParams.Platform.Vsphere != nil && params.ClusterUpdateParams.Platform.Vsphere.VCenter != nil {
+				withCredentials = true
+			}
+		}
```
slaviered, line 2148: you can set it as a compound logical expression: withCredentials := params.ClusterUpdateParams.Platform.Type == models.PlatformTypeVsphere && params.ClusterUpdateParams.Platform.Vsphere != nil && params.ClusterUpdateParams.Platform.Vsphere.VCenter != nil

slaviered: It will be cleared that way

internal/bminventory/inventory.go
eliorerz: We can determine how clusters are with baremetal using the total clusters about minus clusters with platform set

gamli75: the decision was to report only on changes and not the default

PR #2261, opened 2021-07-21
OCPBUGSM-31996: Preparing for Installation failed with message: Failed to prepare the installation due to an unexpected error. Please retry later

ori-amizur:
# Assisted Pull Request

## Description

In order to identify BMH file as master, the name was expected to contain "-master-" .  This changed
Now the name contains the hostname.  In order to identify master correctly, a match is done by
getting hostnames for masters and workers and matching them to the name in the BMH file.
If no match is found, then the previous logic (search for "-master-") is applied.

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x]  UT
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @flaper87 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/ignition/ignition.go
+++ b/internal/ignition/ignition.go
@@ -42,6 +42,7 @@ import (
 	"github.com/openshift/assisted-service/pkg/staticnetworkconfig"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/thoas/go-funk"
 	"github.com/vincent-petithory/dataurl"
 	"golang.org/x/sync/errgroup"
 	"gopkg.in/yaml.v2"
@@ -528,7 +529,24 @@ func (g *installerGenerator) bootstrapInPlaceIgnitionsCreate(installerPath strin
 	return nil
 }
 
-func bmhIsMaster(bmh *bmh_v1alpha1.BareMetalHost) bool {
+func getHostnames(hosts []*models.Host) []string {
+	ret := make([]string, 0)
+	for _, h := range hosts {
+		ret = append(ret, hostutil.GetHostnameForMsg(h))
+	}
+	return ret
+
+}
+
+func bmhIsMaster(bmh *bmh_v1alpha1.BareMetalHost, masterHostnames, workerHostnames []string) bool {
+	if funk.ContainsString(masterHostnames, bmh.Name) {
+		return true
+	}
+	if funk.ContainsString(workerHostnames, bmh.Name) {
+		return false
+	}
+
+	// For backward compatibility in case the name is not in the (masterHostnames, workerHostnames)
 	return strings.Contains(bmh.Name, "-master-")
 }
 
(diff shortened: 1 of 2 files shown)


internal/ignition/ignition.go
```diff
@@ -615,8 +637,19 @@ func (g *installerGenerator) updateBootstrap(bootstrapPath string) error {
...
+			)
+			masterHostnames, err = getHostnames(masters)
+			if err == nil {
+				workerHostnames, err = getHostnames(workers)
+			}
+
+			if err != nil {
+				return err
+			}
+			if bmhIsMaster(bmh, masterHostnames, workerHostnames) {
```
gamli75, line 652: it is not clear why we need both masterHostnames & workerHostnames in order to check if bmhIsMaster and not checking it only in masterHostnames - please add a comment

ori-amizur: Since there is backward compatibility logic afterwards, we need to verify that the host is not a worker before falling back to the backward compatibility logic.

ori-amizur: There is a comment before the backward compatibility logic that explains that

filanov (commented):
Something confusing me, BMH refer to a specific host by hostname but then the list of workers/masters may contain a different order so `modifyBMHFile` may be called with a host that doesn't match to it? 
Then i looks like the bmh have masterX with `RequestedHostname` of hostY

ori-amizur:
> modifyBMHFile

@filanov seems you are right.  But I think the problem existed before.  Will try to fix.

flaper87:
@ori-amizur this also affects operator too. Could you flag `Operator Managed Deployments` in the description?

ori-amizur:
> Something confusing me, BMH refer to a specific host by hostname but then the list of workers/masters may contain a different order so `modifyBMHFile` may be called with a host that doesn't match to it?
> Then i looks like the bmh have masterX with `RequestedHostname` of hostY

After looking at it, the way it was designed before is that the files are sorted here https://github.com/openshift/assisted-service/blob/92bfccfd261d2edcc8ad3bf4209594d8f21cd109/internal/installcfg/installcfg.go#L324 and the code that is changed now relies on the hosts being sorted.  Out implementation also sorts the hosts so their order should lead us to do the right correlation.
If we keep the current implementation then it should be OK since it follows the same guidelines as before.  I agree that it is not the best design since it relies on side effects (sorting) from different packages.
Another option is to do the correlation in our code.

filanov:
Can you please add a comment in the code explaining why the sorting is done? so next time we will know the reason, the sorting of the files is done in a different place so it's confusing to see that we are sorting the hosts here.

ori-amizur:
> @ori-amizur this also affects operator too. Could you flag `Operator Managed Deployments` in the description?

@flaper87 I am not sure I understand

ori-amizur:
> Can you please add a comment in the code explaining why the sorting is done? so next time we will know the reason, the sorting of the files is done in a different place so it's confusing to see that we are sorting the hosts here and depend on the order.

OK

filanov (approved):

PR #2275, opened 2021-07-25
MGMT-7358: change host validation to support infra env

yevgeny-shnaidman:
currently host validations also require existence of the cluster for the host.
in late binding cluster does not necessary exists for the host.
we split the validtion in 2, based on clusterID. the validations for infra-env host
ignore the operators validations, and the cluster conditions

# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/hardware/validator.go
+++ b/internal/hardware/validator.go
@@ -30,12 +30,14 @@ type Validator interface {
 	GetHostValidDisks(host *models.Host) ([]*models.Disk, error)
 	GetHostInstallationPath(host *models.Host) string
 	GetClusterHostRequirements(ctx context.Context, cluster *common.Cluster, host *models.Host) (*models.ClusterHostRequirements, error)
+	GetInfraEnvHostRequirements(ctx context.Context, infraEnv *common.InfraEnv, host *models.Host) (*models.ClusterHostRequirements, error)
 	DiskIsEligible(ctx context.Context, disk *models.Disk, cluster *common.Cluster, host *models.Host) ([]string, error)
 	ListEligibleDisks(inventory *models.Inventory) []*models.Disk
 	GetInstallationDiskSpeedThresholdMs(ctx context.Context, cluster *common.Cluster, host *models.Host) (int64, error)
 	// GetPreflightHardwareRequirements provides hardware (host) requirements that can be calculated only using cluster information.
 	// Returned information describe requirements coming from OCP and OLM operators.
 	GetPreflightHardwareRequirements(ctx context.Context, cluster *common.Cluster) (*models.PreflightHardwareRequirements, error)
+	GetPreflightInfraEnvHardwareRequirements(ctx context.Context, infraEnv *common.InfraEnv) (*models.PreflightHardwareRequirements, error)
 }
 
 func NewValidator(log logrus.FieldLogger, cfg ValidatorCfg, operatorsAPI operators.API) Validator {
@@ -160,7 +162,7 @@ func (v *validator) GetClusterHostRequirements(ctx context.Context, cluster *com
 		return nil, err
 	}
 
-	ocpRequirements, err := v.getOCPHostRoleRequirementsForVersion(cluster, host.Role)
+	ocpRequirements, err := v.getOCPClusterHostRoleRequirementsForVersion(cluster, host.Role)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


internal/hardware/validator.go
```diff
@@ -173,6 +175,27 @@ func (v *validator) GetClusterHostRequirements(ctx context.Context, cluster *com
 	}, nil
 }
 
+func (v *validator) GetInfraEnvHostRequirements(ctx context.Context, infraEnv *common.InfraEnv, host *models.Host) (*models.ClusterHostRequirements, error) {
+	/*
```
ravidbro, line 179: I don't think that a new code block should have code in comments

internal/hardware/validator.go
```diff
@@ -173,6 +175,27 @@ func (v *validator) GetClusterHostRequirements(ctx context.Context, cluster *com
 	}, nil
 }
 
+func (v *validator) GetInfraEnvHostRequirements(ctx context.Context, infraEnv *common.InfraEnv, host *models.Host) (*models.ClusterHostRequirements, error) {
```
ravidbro, line 178: Is it intentional that 'GetInfraEnvHostRequirements' returns 'ClusterHostRequirements'?
Shouldn't we have a dedicated structure for 'InfraEnvHostRequirements'?

internal/hardware/validator.go
```diff
@@ -244,8 +285,20 @@ func (v *validator) getOCPHostRoleRequirementsForVersion(cluster *common.Cluster
...
+func (v *validator) getOCPInfraEnvHostRoleRequirementsForVersion(infraEnv *common.InfraEnv, role models.HostRole) (models.ClusterHostRequirementsDetails, error) {
+	requirements, err := v.getOCPRequirementsForVersion(infraEnv.OpenshiftVersion)
+	if err != nil {
+		return models.ClusterHostRequirementsDetails{}, err
+	}
+
+	if role == models.HostRoleMaster {
+		return *requirements.MasterRequirements, nil
+	}
+	return *requirements.WorkerRequirements, nil
```
ravidbro, line 297: I know that today if it's not a master it is a worker but I think it will be nicer to have 'else worker' as more structured and maybe in the future it won't be the only two options

internal/hardware/validator.go
```diff
@@ -244,8 +285,20 @@ func (v *validator) getOCPHostRoleRequirementsForVersion(cluster *common.Cluster
...
+}
+
 func (v *validator) getPreflightOCPRequirements(cluster *common.Cluster) (*models.HostTypeHardwareRequirementsWrapper, error) {
```
ravidbro, line 300: Should we add 'Cluster' to the name of the method to be aligned with the other methods?

internal/hardware/validator.go
yevgeny-shnaidman: fixed

internal/hardware/validator.go
yevgeny-shnaidman: Currently it is intentional. This commit is not the final design. It should support current day1 flow and allow us to push changes in state machine and instruction manager. The final design will be done in MGMT-7366 by Ori

internal/hardware/validator.go
yevgeny-shnaidman: actually in such case golang does not like if /else

internal/hardware/validator.go
yevgeny-shnaidman: fixed

internal/hardware/validator.go
ravidbro: it will if after the if/else you will return an error which is what is expected if you get something that is not master/worker

internal/hardware/validator.go
ori-amizur: Why do we need requirements for infra-env ? is it mandatory for late binding ? Do infra-env and cluster both have requirements ?

internal/host/refresh_status_preprocessor.go
```diff
@@ -96,28 +96,30 @@ func (r *refreshPreprocessor) preprocess(c *validationContext) (map[string]bool,
...
-	}
-	for _, result := range results {
-		id := validationID(result.ValidationId)
-		conditions[id.String()] = result.Status == api.Success
-		category, err := id.category()
+	if c.infraEnv == nil {
+		// Validate operators
+		results, err := r.operatorsApi.ValidateHost(context.TODO(), c.cluster, c.host)
 		if err != nil {
-			logrus.WithError(err).Warn("id.category()")
```
ori-amizur, line 109: Any reason to remove the error ?

internal/host/validator.go
```diff
@@ -142,30 +155,72 @@ func (c *validationContext) loadGeneralMinRequirements(hwValidator hardware.Vali
...
-		db:      db,
-		cluster: c,
+		host:     host,
+		db:       db,
+		cluster:  c,
+		infraEnv: nil,
 	}
-	err := ret.loadCluster()
-	if err == nil {
+	if host.ClusterID != "" {
```
ori-amizur, line 177: newValidationContext is called with cluster parameter.  Can the ClusterID be empty in this case?

internal/hardware/validator.go
yevgeny-shnaidman: Infra env also has requirement for hardware, but without operators requirements etc'

internal/host/refresh_status_preprocessor.go
yevgeny-shnaidman: The log is not removed, it is just presented so by the diff. You can see on line 110

yevgeny-shnaidman: the error is not removed, it is just how diff presents it, you can see it on line 110

internal/host/validator.go
yevgeny-shnaidman: it is called with host and cluster. cluster might be nil, even if it exists. in case cluster is not passed, it will be loaded by the loadCluster function. host.ClusterIDF condition differentiates between day1 cluster hosts and infra-env (pool) hosts

internal/host/conditions.go
```diff
@@ -29,36 +29,60 @@ func (c conditionId) String() string {
 }
 
 func (v *validator) isInstallationDiskSpeedCheckSuccessful(c *validationContext) bool {
+	if c.infraEnv != nil {
```
ori-amizur, line 32: Not sure all these changes are needed because they are checked in certain states.  For example here https://github.com/openshift/assisted-service/blob/fa49588aa60a275a2e9d6bb4f11ab9320e99f40c/internal/host/statemachine.go#L314

internal/hardware/validator.go
yevgeny-shnaidman: fixed

internal/host/conditions.go
yevgeny-shnaidman: since it is preprocessed, i want to make sure that cluster is not access in case of infra-env host. In addition, you will probably change this code when dealing with MGMT-7366 :)

internal/host/validator.go
ori-amizur: But where is it called from (with InfraEnv only) ?

internal/host/refresh_status_preprocessor.go
ori-amizur: OK

internal/host/validator.go
yevgeny-shnaidman: it will be called during validation for infra-env host statemachine. this code is preparation for MGMT-7093

ravidbro:
I'm OK with the fixes for the changes I asked but I'll let @ori-amizur lgtm when he is satisfied

PR #2289, opened 2021-07-27
NO-ISSUE: Move 4.8 support_level from beta to production

oshercc:
@romfreiman @gamli75


gamli75:
@oshercc please move the default:true from 4.7 to 4.8 - see https://github.com/openshift/assisted-service/blob/master/data/default_ocp_versions.json#L19

gamli75:
/hold

hold until we will change the default flag as well

YuviGold (approved):

PR #2293, opened 2021-07-28
MGMT-7333 Add a HoldInstallation flag to AgentClusterInstall

filanov:
# Assisted Pull Request

## Description

Hold day1 instalaltion
This feature is relevant mostly for UI, to give users an option to
review the deployment before we trigger the intallation
day2 flow will start installation automatically once hosts are added to
the cluster when they are in a ready state.

No effect on ZTP

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @rollandf 
/cc @avishayt 

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/hiveextension/v1beta1/agentclusterinstall_types.go
+++ b/api/hiveextension/v1beta1/agentclusterinstall_types.go
@@ -55,6 +55,8 @@ const (
 	ClusterInstallationFailedMsg        string = "The installation has failed:"
 	ClusterInstallationNotStartedReason string = "InstallationNotStarted"
 	ClusterInstallationNotStartedMsg    string = "The installation has not yet started"
+	ClusterInstallationOnHoldReason     string = "InstallationOnHold"
+	ClusterInstallationOnHoldMsg        string = "The installation is on hold. To unhold set holdInstallation to false"
 	ClusterInstallationInProgressReason string = "InstallationInProgress"
 	ClusterInstallationInProgressMsg    string = "The installation is in progress:"
 	ClusterUnknownStatusReason          string = "UnknownStatus"
@@ -138,6 +140,12 @@ type AgentClusterInstallSpec struct {
 	// IngressVIP is the virtual IP used for cluster ingress traffic.
 	// +optional
 	IngressVIP string `json:"ingressVIP,omitempty"`
+
+	// HoldInstallation will prevent installation from happening when true.
+	// Inspection and validation will proceed as usual, but once the RequirementsMet condition is true,
+	// installation will not begin until this field is set to false.
+	// +optional
+	HoldInstallation bool `json:"holdInstallation,omitempty"`
 }
 
 // AgentClusterInstallStatus defines the observed state of the AgentClusterInstall.
--- a/internal/controller/controllers/clusterdeployments_controller.go
+++ b/internal/controller/controllers/clusterdeployments_controller.go
@@ -209,24 +209,7 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 
 	// In case the Cluster is a Day 1 cluster and is installed, update the Metadata and create secrets for credentials
 	if *cluster.Status == models.ClusterStatusInstalled && swag.StringValue(cluster.Kind) == models.ClusterKindCluster {
-		if !isInstalled(clusterDeployment, clusterInstall) {
-			// create secrets and update status
-			err = r.updateClusterMetadata(ctx, log, clusterDeployment, cluster, clusterInstall)
-			if err != nil {
-				log.WithError(err).Error("failed to update cluster metadata")
-			}
-			return r.updateStatus(ctx, log, clusterInstall, cluster, err)
-		} else if r.EnableDay2Cluster && !r.isSNO(clusterInstall) {
-			// Delete Day1 Cluster
-			_, err = r.deregisterClusterIfNeeded(ctx, log, req.NamespacedName)
-			if err != nil {
-				log.WithError(err).Error("failed to deregister cluster")
-				return r.updateStatus(ctx, log, clusterInstall, cluster, err)
-			}
-			//Create Day2 cluster
-			return r.createNewDay2Cluster(ctx, log, req.NamespacedName, clusterDeployment, clusterInstall)
-		}
-		return r.updateStatus(ctx, log, clusterInstall, cluster, err)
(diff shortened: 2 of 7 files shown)


filanov:
/test subsystem-kubeapi-aws
infra error.

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -237,7 +220,7 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 		}
 	}
 
-	if swag.StringValue(cluster.Kind) == models.ClusterKindCluster {
+	if swag.StringValue(cluster.Kind) == models.ClusterKindCluster && !clusterInstall.Spec.HoldInstallation {
```
nmagnezi, line 223: perhaps we should log this case, or show this in conditions to the user?
people might miss the reason as for why the installation never started

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -138,6 +138,11 @@ type AgentClusterInstallSpec struct {
 	// IngressVIP is the virtual IP used for cluster ingress traffic.
 	// +optional
 	IngressVIP string `json:"ingressVIP,omitempty"`
+
+	// HoldInstallation if true it will prevent installation to start automatically.
```
nmagnezi, line 142: when an agent gets approved we have:
`Approved = True`

with `HoldInstallation` it kinda the opposite, where `True` means 'not approved'
for consistency, maybe we should have `ApproveInstallation` instead?

nmagnezi (changes requested):

internal/controller/controllers/clusterdeployments_controller.go
filanov: user is the one that set the spec so i'm not sure what we should add?

api/hiveextension/v1beta1/agentclusterinstall_types.go
filanov: We had this discussion with UX and architects, we prefer the default to be automatic. 
We can add defaults with webhooks but it's breaking the API for existing clusters/ZTP flows.

internal/controller/controllers/clusterdeployments_controller.go
filanov: buy maybe i'm wrong, @avishayt @mhrivnak what do you think?

nmagnezi: True. but is there a way to show that the installation did not start because HoldInstallation = True?

api/hiveextension/v1beta1/agentclusterinstall_types.go
nmagnezi: ok

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -138,6 +138,11 @@ type AgentClusterInstallSpec struct {
...
+	// relevant only for day1 installation
```
avishayt, line 143:
```suggestion
	// Cluster installation will not start as long as HoldInstallation is set to true.
```

internal/controller/controllers/clusterdeployments_controller.go
avishayt: Not sure there is what to show.  It's in the Spec and like @filanov said, the user set it.

nmagnezi: ok

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -138,6 +138,10 @@ type AgentClusterInstallSpec struct {
 	// IngressVIP is the virtual IP used for cluster ingress traffic.
 	// +optional
 	IngressVIP string `json:"ingressVIP,omitempty"`
+
+	// Cluster installation will not start as long as HoldInstallation is set to true.
```
mhrivnak, line 142: I think it's worth adding some detail that validation and such will still happen. We don't want this to be confused with a generic "Pause" button that will stop progress toward installation; it just adds a gate at a specific part of the process, and we should describe exactly where that is.

mhrivnak (changes requested):

api/hiveextension/v1beta1/agentclusterinstall_types.go
mhrivnak: That's an improvement. I suggest being explicit about when the behavior does have an effect. Also I noticed that this needs to be re-phrased to start with the name of the field, which is [a golang doc convention that supports good doc generation](https://blog.golang.org/godoc). Maybe something like this:

`HoldInstallation will prevent installation from happening when true. Inspection and validation will proceed as usual, but once the RequirementsMet condition is true, installation will not begin until this field is set to false.`

internal/controller/controllers/clusterdeployments_controller.go
filanov: Thanks for the feedback, i added the condition and changed the message as well. 
Let me know if you think it should be different.

filanov:
test /continuous-integration/jenkins/pr-merge

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -55,6 +55,8 @@ const (
 	ClusterInstallationFailedMsg        string = "The installation has failed:"
 	ClusterInstallationNotStartedReason string = "InstallationNotStarted"
 	ClusterInstallationNotStartedMsg    string = "The installation has not yet started"
+	ClusterInstallationOnHoldReason     string = "InstallationOnHold"
+	ClusterInstallationOnHoldMsg        string = "The installation is on hold, to unhold set holdInstallation to false"
```
mhrivnak, line 59: For better grammar, I would make this two sentences. "The installation is on hold. To unhold set holdInstallation to false"
Write

mhrivnak (changes requested):
Looks great. I just suggested one change to the human-readable message.

api/hiveextension/v1beta1/agentclusterinstall_types.go
filanov: Sure, added documentation for the condition as well.

mhrivnak (approved):

PR #2296, opened 2021-07-28
MGMT-7093: implement pool image state machine and instruction manager

yevgeny-shnaidman:
implement pool image host statmachine. it is implemented in a different file
and added to the statemachine that implements day1 and day2 states.In addition
an instruction manager is expended to support the states of the pool hosts

# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/db.go
+++ b/internal/common/db.go
@@ -188,12 +188,12 @@ func GetClustersFromDBWhere(db *gorm.DB, eagerLoading EagerLoadingState, include
 	return clusters, nil
 }
 
-func GetHostFromDB(db *gorm.DB, clusterId, hostId string) (*Host, error) {
+func GetHostFromDB(db *gorm.DB, infraEnvId, hostId string) (*Host, error) {
 	var host Host
 
-	err := db.First(&host, "id = ? and cluster_id = ?", hostId, clusterId).Error
+	err := db.First(&host, "id = ? and infra_env_id = ?", hostId, infraEnvId).Error
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get host %s in cluster %s", hostId, clusterId)
+		return nil, errors.Wrapf(err, "failed to get host %s in infra_env %s", hostId, infraEnvId)
 	}
 	return &host, nil
 }
--- a/internal/host/hostcommands/instruction_manager.go
+++ b/internal/host/hostcommands/instruction_manager.go
@@ -31,8 +31,9 @@ const (
 )
 
 type StepsStruct struct {
-	Commands      []CommandGetter
-	NextStepInSec int64
+	Commands       []CommandGetter
+	NextStepInSec  int64
+	PostStepAction string
 }
 
 type stateToStepsMap map[string]StepsStruct
@@ -42,6 +43,7 @@ type InstructionManager struct {
 	db                            *gorm.DB
 	installingClusterStateToSteps stateToStepsMap
 	addHostsClusterToSteps        stateToStepsMap
+	poolHostToSteps               stateToStepsMap
 	disabledStepsMap              map[models.StepType]bool
 }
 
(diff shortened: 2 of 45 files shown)


swagger.yaml
```diff
@@ -3848,6 +3849,12 @@ definitions:
           - resetting
           - added-to-existing-cluster
           - cancelled
+          - waiting-to-be-registered
+          - known-in-pool
+          - disconnected-in-pool
+          - insufficient-in-pool
+          - disabled-in-pool
+          - discovering-in-pool
```
avishayt, line 3857:
```suggestion
          - binding
          - known-unbound
          - disconnected-unbound
          - insufficient-unbound
          - disabled-unbound
          - discovering-unbound
```

internal/host/transition.go
```diff
@@ -529,6 +552,14 @@ func (th *transitionHandler) IsDay2Host(sw stateswitch.StateSwitch, args statesw
 	return hostutil.IsDay2Host(sHost.host), nil
 }
 
+func (th *transitionHandler) IsPoolHost(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) (bool, error) {
```
avishayt, line 555:
```suggestion
func (th *transitionHandler) IsUnboundHost(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) (bool, error) {
```

internal/host/hostcommands/empty_cmd.go
```diff
@@ -0,0 +1,24 @@
+package hostcommands
+
+import (
```
ori-amizur, line 3: Why not use this one https://github.com/openshift/assisted-service/blob/master/internal/host/hostcommands/noop_cmd.go ?

swagger.yaml
```diff
@@ -4011,6 +4018,7 @@ definitions:
       - installation-disk-speed-check
       - container-image-availability
       - domain-resolution
+      - empty
```
ori-amizur, line 4021: Why is it needed ?

internal/host/hostcommands/instruction_manager.go
```diff
@@ -31,8 +31,9 @@ const (
 )
 
 type StepsStruct struct {
-	Commands      []CommandGetter
-	NextStepInSec int64
+	Commands       []CommandGetter
+	NextStepInSec  int64
+	PostStepAction string
```
ori-amizur, line 36: What is the usage of this field ?

internal/common/db.go
```diff
@@ -188,12 +188,12 @@ func GetClustersFromDBWhere(db *gorm.DB, eagerLoading EagerLoadingState, include
 	return clusters, nil
 }
 
-func GetHostFromDB(db *gorm.DB, clusterId, hostId string) (*Host, error) {
+func GetHostFromDB(db *gorm.DB, infraEnvId, hostId string) (*Host, error) {
 	var host Host
 
-	err := db.First(&host, "id = ? and cluster_id = ?", hostId, clusterId).Error
+	err := db.First(&host, "id = ? and infra_env_id = ?", hostId, infraEnvId).Error
```
ori-amizur, line 194: All the locations that use this function were not changed - since they are still passing cluster-id.  Why change it ?

swagger.yaml
yevgeny-shnaidman: fixed

internal/host/transition.go
yevgeny-shnaidman: fixed

internal/host/hostcommands/empty_cmd.go
yevgeny-shnaidman: noop_cmd does not suit the purpose, it does not return any steps to be sent to the agent. W need a step sent to agent, but that its command is empty

swagger.yaml
yevgeny-shnaidman: since we define an empty command, and it is sent to the agent, it is helpful that there will be a type assigned to it. For example it helps during verifications in unit test

internal/host/hostcommands/instruction_manager.go
yevgeny-shnaidman: This field is used to tell the agent what to do after it finished executing the current bunch of the command (steps) sent to it. By default , we send Continue action, which tells agents to continue processing (sleep and then send get next step command. In case of empty command we send the agent action of exit, which causes it to exit the current loop and to initiate registration to the AI again

internal/common/db.go
yevgeny-shnaidman: Starting from this commit we are may be using unbound hosts, without cluster id, so the function must be updated. The whole move from cluster-id to infra-env will be done in MGMT-7321

internal/host/hostcommands/empty_cmd.go
ori-amizur: This cause an error since we are trying to run empty command.  Why is there a need to send a step to the agent that does nothing ?

yevgeny-shnaidman: we want agent to re-register. we also don't want to change agent's code. So we send it an empty command ( since we don't want it to do anything), and tell it to exit when it is done. also the response in the AI does not affect host state

ori-amizur: So use the noop command above.  It will do the same, but without the error

yevgeny-shnaidman: fixed

PR #2297, opened 2021-07-28
MGMT-7389: Create cluster with a platform

michaellevy101:
# Assisted Pull Request

## Description
Currently for creating a vsphere cluster we have to call 2 APIs:
registerCluster and then updateCluster to set the platform
Now we can create a vsphere cluster with a single API call
Relevant for the test-infra and where the user comes from the UI vSphere section rather than Bare Metal.

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -4191,6 +4191,10 @@ definitions:
       network_configuration:
         $ref: '#/definitions/network_configuration'
         x-nullable: true
+      platform:
+        type: object
+        $ref: '#/definitions/platform'
+        x-nullable: true
 
   cluster-update-params:
     type: object
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -334,6 +334,11 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(_ context.Context, 
 	if params.NewClusterParams.SchedulableMasters == nil {
 		params.NewClusterParams.SchedulableMasters = swag.Bool(false)
 	}
+	if params.NewClusterParams.Platform == nil {
+		params.NewClusterParams.Platform = &models.Platform{
+			Type: models.PlatformTypeBaremetal,
+		}
+	}
 
 	return params
 }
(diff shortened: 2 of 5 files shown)


eliorerz:
Please add description and unit-tests

gamli75:
@avishayt why do we need it in the create cluster?

avishayt:
If I know I want vSphere, why do I need 2 API calls?  For example, k8s API, or in the cloud UI if the user comes from vSphere section rather than Bare Metal.

gamli75:
@michaellevy101 please retitle your PR to: MGMT-7389**:** Create cluster with a platform

michaellevy101:
@avishayt you need to create the cluster with the vsphere platform...

swagger.yaml
```diff
@@ -4184,6 +4184,10 @@ definitions:
       network_configuration:
         $ref: '#/definitions/network_configuration'
         x-nullable: true
+      platform:
+        type: object
+        $ref: '#/definitions/platform'
+        x-nullable: true
```
avishayt, line 4190: If you do this, can you remove the default logic in the code?

```suggestion
        default: 'baremetal'
```

michaellevy101: What about kube-api which calls internally?

avishayt: Good point.  @filanov what's our convention?

filanov: As far as i remember default doesn't really work. fist need to check that.
In kube-api we use internal functions so the default (if it works) is done by the infra wrappers that we don't call. 
So if we use this feature we will have to handle it in kube-api. 
There are two options There:
* If it's part of the API (AgentClusterInstall) then we will can add default with webhooks (wip).
* If it's not part of the API we will need to handle it in the code of the controller when calling create.

michaellevy101: I think that instead of dealing with it twice we can set it on the registration operation (All the method has default values)

avishayt: I agree, the way it's implemented here looks best.

filanov: +1

avishayt (approved):

avishayt (approved):

filanov:
@michaellevy101 just making sure but are existing clusters covered in this RP?
What will happen to clusters that were created before?

PR #2298, opened 2021-07-28
MGMT-6911: Validate that DNS wildcard entry is not configured.

bkhizgiy:
Add new validation that checks if the host has DNS wildcard entry for
*.cluster_name.base_dns_domain_name

If a wildcard was configured in the cluster network
it will cause insight pods to fail report to cloud.redhat.com
And cluster version will complain about the insight operator's degraded status.

By testing domain name resolution for a non-existing domain name under 
cluster_name.base_dns_domain_name we can determine if wildcard was
configured and alert the user accordingly.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees
/cc @tsorya 
/cc @omertuc 
/cc @eranco74 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/test_configuration.go
+++ b/internal/common/test_configuration.go
@@ -86,7 +86,7 @@ var TestDefaultConfig = &TestConfiguration{
 
 	ImageName: "image",
 
-	ClusterName: "test",
+	ClusterName: "test-cluster",
 
 	BaseDNSDomain: "example.com",
 
@@ -110,9 +110,10 @@ var TestImageStatusesFailure = &models.ContainerImageAvailability{
 	Result: models.ContainerImageAvailabilityResultFailure,
 }
 
-var DomainAPI = "api.test.example.com"
-var DomainAPIInternal = "api-int.test.example.com"
-var DomainApps = fmt.Sprintf("%s.apps.test.example.com", constants.AppsSubDomainNameHostDNSValidation)
+var DomainAPI = "api.test-cluster.example.com"
+var DomainAPIInternal = "api-int.test-cluster.example.com"
+var DomainApps = fmt.Sprintf("%s.apps.test-cluster.example.com", constants.AppsSubDomainNameHostDNSValidation)
+var WildcardDomain = fmt.Sprintf("%s.test-cluster.example.com", constants.DNSWildcardFalseDomainName)
 
 var DomainResolution = []*models.DomainResolutionResponseDomain{
 	{
@@ -129,6 +130,11 @@ var DomainResolution = []*models.DomainResolutionResponseDomain{
 		DomainName:    &DomainApps,
 		IPV4Addresses: []strfmt.IPv4{"7.8.9.10/24"},
 		IPV6Addresses: []strfmt.IPv6{"1003:db8::10/120"},
+	},
+	{
+		DomainName:    &WildcardDomain,
+		IPV4Addresses: []strfmt.IPv4{},
+		IPV6Addresses: []strfmt.IPv6{},
 	}}
 
 var TestDomainNameResolutionSuccess = &models.DomainResolutionResponse{
@@ -231,6 +237,23 @@ func GenerateTestDefaultVmwareInventory() string {
 	b, err := json.Marshal(inventory)
 	Expect(err).To(Not(HaveOccurred()))
 	return string(b)
+}
+
+func CreateWildcardDomainNameResolutionReply(name string, baseDomain string) *models.DomainResolutionResponse {
+
+	domain := fmt.Sprintf("%s.%s.%s", constants.DNSWildcardFalseDomainName, name, baseDomain)
+
+	var domainNameWildcardConfig = []*models.DomainResolutionResponseDomain{
+		{
+			DomainName:    &domain,
+			IPV4Addresses: []strfmt.IPv4{},
+			IPV6Addresses: []strfmt.IPv6{},
+		}}
+
+	var testDomainNameResolutionWildcard = &models.DomainResolutionResponse{
+		Resolutions: domainNameWildcardConfig}
+
+	return testDomainNameResolutionWildcard
 }
 
 type NetAddress struct {
--- a/internal/constants/files.go
+++ b/internal/constants/files.go
@@ -7,3 +7,6 @@ const KubeconfigNoIngress = "kubeconfig-noingress"
 const AppsSubDomainNameHostDNSValidation = "console-openshift-console"
 const APIName = "api"
 const APIInternalName = "api-int"
+
+//Non existing domain name under clusterName.baseDomain for wildcard configuration check
+const DNSWildcardFalseDomainName = "validateNoWildcardDNS"
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -1113,3 +1113,34 @@ func printIsDomainNameResolvedCorrectly(c *validationContext, status ValidationS
 		return "Unexpected status"
 	}
 }
+
+func (v *validator) isDNSWildcardNotConfigured(c *validationContext) ValidationStatus {
+	var response *models.DomainResolutionResponse
+	if err := json.Unmarshal([]byte(c.host.DomainNameResolutions), &response); err != nil {
+		return ValidationError
+	}
+	dnsWildcardName := domainNameToResolve(c, constants.DNSWildcardFalseDomainName)
+
+	// Note that we're validating that the wildcard DNS *.<cluster_name>.<base_domain> is NOT configured, since this causes known problems for OpenShift
+	for _, domain := range response.Resolutions {
+		if domain.DomainName != nil && *domain.DomainName == dnsWildcardName {
+			if len(domain.IPV4Addresses) == 0 && len(domain.IPV6Addresses) == 0 {
+				return ValidationSuccess
+			}
+		}
+	}
+	return ValidationFailure
+}
+
+func (v *validator) printIsDNSWildcardNotConfigured(c *validationContext, status ValidationStatus) string {
+	switch status {
+	case ValidationSuccess:
+		return "DNS wildcard check was succesful"
+	case ValidationFailure:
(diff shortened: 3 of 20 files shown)


internal/common/test_configuration.go
```diff
@@ -119,6 +120,9 @@ var DomainResolution = []*models.DomainResolutionResponseDomain{
 		DomainName:    &DomainApps,
 		IPV4Addresses: []strfmt.IPv4{"7.8.9.10/24"},
 		IPV6Addresses: []strfmt.IPv6{"1003:db8::10/120"},
+	},
+	{
+		DomainName: &WildcardDomain,
```
omertuc, line 125: Please be more explicit about the lack of response:
```go
	{
		DomainName:    &WildcardDomain,
		IPV4Addresses: []strfmt.IPv4{},
		IPV6Addresses: []strfmt.IPv6{},
	},
```

internal/constants/files.go
```diff
@@ -7,3 +7,6 @@ const KubeconfigNoIngress = "kubeconfig-noingress"
 const AppsSubDomainNameHostDNSValidation = "console-openshift-console"
 const APIName = "api"
 const APIInternalName = "api-int"
+
+//Non existing domain name under clusterName.baseDomain for waildcard configuratin check
```
omertuc, line 11: `wildcard configuration`

internal/host/validator.go
```diff
@@ -1032,3 +1032,40 @@ func printIsDomainNameResolvedCorrectly(c *validationContext, status ValidationS
 		return "Unexpected status"
 	}
 }
+
+func (v *validator) isDNSWildcardNotConfigured(c *validationContext) ValidationStatus {
+	if !isNonePlatformMultiNodeInstallation(c) {
```
omertuc, line 1037: @eranco74 do we want [this](https://issues.redhat.com/browse/MGMT-6911) only on none-platform, or all the time?

internal/host/validator.go
```diff
@@ -1032,3 +1032,40 @@ func printIsDomainNameResolvedCorrectly(c *validationContext, status ValidationS
...
+
+	var response *models.DomainResolutionResponse
+	if err := json.Unmarshal([]byte(c.host.DomainNameResolutions), &response); err != nil {
+		return ValidationError
+	}
+	dnsWildcardName := domainNameToResolve(c, constants.DNSWildcardFalseDomainName)
+
+	for _, domain := range response.Resolutions {
+		if domain.DomainName != nil && *domain.DomainName == dnsWildcardName {
+			if len(domain.IPV4Addresses) == 0 && len(domain.IPV6Addresses) == 0 {
```
omertuc, line 1049: Please add comment above this line: 
```go
// Note that we're validating that the wildcard DNS *.<cluster_name>.<base_domain> is NOT configured, because this causes known problems for OpenShift 
```

internal/host/validator.go
```diff
@@ -1032,3 +1032,40 @@ func printIsDomainNameResolvedCorrectly(c *validationContext, status ValidationS
...
+
+func (v *validator) printIsDNSWildcardNotConfigured(c *validationContext, status ValidationStatus) string {
+	switch status {
+	case ValidationSuccess:
+		if !isNonePlatformMultiNodeInstallation(c) {
+			return "DNS wildcard configuration check is not required (managed networking)"
+		}
+		return "DNS wildcard check was succesful"
+	case ValidationFailure:
+		return fmt.Sprintf("DNS wildcard configuration was detected for domain *.%s.%s The installation will not be able to complete until the entry exists. Please remove it to proceed.", c.cluster.Name, c.cluster.BaseDNSDomain)
```
omertuc, line 1065: `s/until/while/`

internal/host/validator.go
omertuc: @eranco74 also please take a look at this PR in general, it's your ticket so you know best

tsorya: all the time, it effects all the platforms

tsorya:
@bkhizgiy thanks for pushing it, it will save tons of debug time

internal/host/validator.go
omertuc: Great. So @bkhizgiy please remove this `if` statement and also don't forget to correct the commit message

internal/common/test_configuration.go
bkhizgiy: done

internal/host/validator.go
bkhizgiy: Ack. removed the check from the validation.

internal/host/validator.go
bkhizgiy: done

internal/host/validator.go
bkhizgiy: done

bkhizgiy: done

internal/constants/files.go
```diff
@@ -7,3 +7,6 @@ const KubeconfigNoIngress = "kubeconfig-noingress"
...
+const DNSWildcardFalseDomainName = "testWildCardConfig"
```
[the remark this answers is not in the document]
bkhizgiy: done

internal/host/validator.go
omertuc: Commit message + PR description still mention "user managed networking"

bkhizgiy:
/unhold
The wildcard configuration validation was manually checked, everything is working and the behavior was as expected.

eranco74:
Please no more retests without looking at the failure.
The tests fail due to a host in insufficient status, probably caused by the new validation added in this PR.

internal/common/test_configuration.go
```diff
@@ -233,6 +239,23 @@ func GenerateTestDefaultVmwareInventory() string {
 	return string(b)
 }
 
+func CreatWildcardDomainNameResolutionReply(name string, baseDomain string) *models.DomainResolutionResponse {
```
tsorya, line 242: typo: Create

subsystem/day2_cluster_test.go
```diff
@@ -141,6 +141,7 @@ var _ = Describe("Day2 cluster tests", func() {
 		By("checking insufficient state state - one host, no connectivity check")
 		ips := hostutil.GenerateIPv4Addresses(2, defaultCIDRv4)
 		generateEssentialHostSteps(ctx, h, "h1host", ips[0])
+		generateDomain(ctx, h, "test-cluster", "")
```
tsorya, line 144: "test-cluster" consts?

subsystem/kubeapi_test.go
```diff
@@ -691,6 +691,9 @@ var _ = Describe("[kube-api]cluster installation", func() {
 			hosts = append(hosts, host)
 		}
 		generateFullMeshConnectivity(ctx, ips[0], hosts...)
+		for _, h := range hosts {
+			generateDomain(ctx, h, clusterDeploymentSpec.ClusterName, "hive.example.com")
```
tsorya, line 695: "hive.example.com" -> consts or test var?

subsystem/day2_cluster_test.go
bkhizgiy: It's a var, depending on the test, the cluster name+Base domain defers between different tests.

tsorya: i mean that adding it as string in all the cases looks less pretty than creating 
var cluterName or using some const

internal/common/test_configuration.go
```diff
@@ -110,9 +110,10 @@ var TestImageStatusesFailure = &models.ContainerImageAvailability{
 	Result: models.ContainerImageAvailabilityResultFailure,
 }
 
-var DomainAPI = "api.test.example.com"
-var DomainAPIInternal = "api-int.test.example.com"
-var DomainApps = fmt.Sprintf("%s.apps.test.example.com", constants.AppsSubDomainNameHostDNSValidation)
+var DomainAPI = "api.test-cluster.example.com"
```
ybettan, line 113: I think it will be more clear if we use `example.invalid` as baseDnsDomain (in the test).

See https://github.com/openshift/assisted-service/blob/dd818a5650a66ff6693f632ddb83d48d29692755/internal/network/manifests_generator.go#L343

subsystem/ams_subscriptions_test.go
```diff
@@ -343,12 +343,15 @@ var _ = Describe("test AMS subscriptions", func() {
...
 				generateClusterISO(clusterID, models.ImageTypeMinimalIso)
-				registerHostsAndSetRoles(clusterID, minHosts)
+				registerHostsAndSetRoles(clusterID, minHosts, "test-ams-subscriptions-cluster", "example.com")
 				reply, err = userBMClient.Installer.InstallCluster(context.Background(), &installer.InstallClusterParams{ClusterID: clusterID})
 				Expect(err).NotTo(HaveOccurred())
 				c := reply.GetPayload()
 				Expect(*c.Status).Should(Equal(models.ClusterStatusPreparingForInstallation))
 				generateEssentialPrepareForInstallationSteps(ctx, c.Hosts...)
+				for _, host := range c.Hosts {
+					updateProgress(*host.ID, clusterID, models.HostStageDone)
```
ybettan, line 353: This cannot work, the test title is `UpdateSubscription failed`.

You are trying to simulate the host's installation on a cluster that is failing to install,  you can look on how it is done in `happy flow` test - https://github.com/openshift/assisted-service/blob/dd818a5650a66ff6693f632ddb83d48d29692755/subsystem/ams_subscriptions_test.go#L310

bkhizgiy: you right, thanks! I have been trying to understand why it keeps failing, removing these lines.

internal/common/test_configuration.go
bkhizgiy: All the DNS validation runs with the same resolution result so the cluster name+domain should be the same for all 4 resolved entries or else the unit tests will fail, and I think this name IS less fitting to the normal DNS validation so I will leave it with "example.com".
Also for the subsystem test, 99% defined with "example.com" in the test files already so if I change the name they will fail as well.

PR #2305, opened 2021-07-29
MGMT-7309: enhancement proposal for ARM64 CPU architecture support

danielerez:
An enhancement proposal for 'ARM64 CPU architecture support':
In order to support additional platforms, the assisted-service
should accept and handle selection of other platforms.
The focus of this enhancement is specifically arm64, but the introduced
infrastructure should facilitate future support of more platforms (e.g. PPC).

### DIFF
--- a/docs/enhancements/arm64-cpu-architecture.md
+++ b/docs/enhancements/arm64-cpu-architecture.md
@@ -1,0 +1,192 @@
+---
+title: arm64-cpu-architecture
+authors:
+  - "@danielerez"
+creation-date: 2021-07-29
+last-updated: 2021-07-29
+---
+
+# ARM64 CPU architecture support
+
+## Summary
+
+The default host CPU architecture supported for installation is x86_64 in Assisted Installer. In order to support additional CPU architectures, the assisted-service should accept and handle selection of arm64 CPU architecture as well.
+
+## Motivation
+
+Although x86_64 is popular in many use cases, other CPU architectures are common in other segments.  For example, ARM is popular in edge use cases.  OpenShift has recently added support for arm64 and it is important to add this support to Assisted Installer.
+
+### Goals
+
+* Allow installation of clusters that are running on arm64 hosts.
+* The introduced infrastructure should facilitate future support of more architectures (e.g. PPC).
+
+### Non-Goals
+
+* The assisted-service and assisted-image-service will not be built for arm64.
+* No support for clusters with hosts of multiple architectures (e.g. x86_64 + arm64).
+* Implementation for kube-api as ClusterImageSet and AgentServiceConfig CRDs. Though the feature design should consider implications.
+* Day2 and Late binding won't be supported in first phase.
+* On-Prem flows, for now the focus is for SaaS users.
+* No baremetal platform support for arm64 in 4.9. I.e. only SNO for now.
+
+## Proposal
+
+### Host images
+
+The following images should also be built for arm64:
+  * assisted-installer
+  * assisted-installer-controller
+  * assisted-installer-agent
+
+This should be handled in similar to the current flows by modifying relevant Dockerfiles to support arm64 build. The images should be published to the same registries and follow same naming conventions. I.e. The best approach would be using the same image URI for any architecture. Which means, minimal changes in the assisted-service, as the architecture specific image would be fetched according to the machine pulling the image.
+
+### OCP versions
+
+[Versions](https://github.com/openshift/assisted-service/blob/master/internal/versions/versions.go) should
+have CPU architecture granularity for each API. So that OCP release images and RHCOS images of multiple architectures could be supported.
+
+The assisted-service should support the arm64 variant of RHCOS images and OCP release image (for OpenShift >= 4.9).
+
+As a preparation for supporting multiple CPU architectures in Version, the RHCOS images information from OPENSHIFT_VERSIONS environment variable should be extracted into a new variable (e.g. RHCOS_VERSIONS).
+We should probably use a simple array structure instead of mapping, to align with the k8s API convention in [AgentServiceConfig](https://github.com/openshift/assisted-service/blob/77cb9d3348dafd2abd61fd7c8066e7e7a7805d95/config/samples/agent-install.openshift.io_v1beta1_agentserviceconfig.yaml#L18-L22) structure.
+
+To simplify backwards compatibility, we can still support the old format of OPENSHIFT_VERSIONS by keeping the current properties for old versions (<4.9).
+
+E.g. 
+
+#### OPENSHIFT_VERSIONS
+```json
+"4.8": {
+      "display_name": "4.8.2",
+      "release_version": "4.8.2",
+      "release_image": "quay.io/openshift-release-dev/ocp-release:4.8.2-x86_64",
+      "rhcos_image": "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-4.8.2-x86_64-live.x86_64.iso",
+      "rhcos_rootfs": "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-live-rootfs.x86_64.img",
+      "rhcos_version": "48.84.202107202156-0",
+      "support_level": "production",
+      "default": true
+},
+"4.9": {
+   "display_name": "4.9.0-0.nightly",
+   "support_level": "beta",
+   "release": {
+      "x86_64": {
+            "cpu_architecture": "x86_64",
+            "release_version": "4.9.0-0.nightly-2021-07-05-092650",
+            "release_image": "quay.io/openshift-release-dev/ocp-release-nightly:4.9.0-0.nightly-2021-07-05-092650"
+      },
+      "arm64": {
+            "cpu_architecture": "arm64",
+            "release_version": "4.9.0-0.nightly-arm64-2021-07-05-092650",
+            "release_image": "quay.io/openshift-release-dev/ocp-release-nightly@sha256:74493f2ad08a86f3f001ab398735965d0253755dbe4186ee7aa9e6d6a809ae3b"
+      }
+   }
+}
+```
+
+#### RHCOS_VERSIONS
+```json
+"os_images": [
+    {
+      "openshift_version": "4.9",
+      "cpu_architecture": "x86_64",
+      "url": "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/latest-4.9/rhcos-live.x86_64.iso",
+      "rootfs_url": "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/latest-4.9/rhcos-live-rootfs.x86_64.img",
+      "version": "49.84.202107032011-0"
+    },
+    {
+      "openshift_version": "4.9",
+      "cpu_architecture": "arm64",
+      "url": "https://mirror.openshift.com/pub/openshift-v4/aarch64/dependencies/rhcos/pre-release/latest-4.9/rhcos-live.aarch64.iso",
+      "rootfs_url": "https://mirror.openshift.com/pub/openshift-v4/aarch64/dependencies/rhcos/pre-release/latest-4.9/rhcos-live-rootfs.aarch64.img",
+      "version": "49.84.202106272247-0"
+    }
+  ]
+```
+
+### Cluster creation and ISO generation
+
+When creating a new cluster, `cpu_architecture` property should be supported to define which hosts can be registered to it. This property would be added to cluster-create-params and stored in cluster upon creation. If not specified, defaulted to x86_64 as a fallback.
+
+An example for the swagger schema:
+
+```
+cpu_architecture:
+   type: string
+   default: 'x86_64'
+   description: The CPU architecture of the image (x86_64/arm64/etc)
+```
+
+When generating a discovery ISO, the proper RHCOS/Release images should be used according to CPU architecture specified in the cluster. The `cpu_architecture` property should be stored in `infraenv` after the ISO is generated.
+
+### User Stories
+
+#### Story 1
+
+As an Assisted-Installer user, I need to install a cluster on non x86_64 hosts.
+
+#### Story 2
+
+As an Assisted-Installer user, I need to install a Single-Node OpenShift cluster on an arm64 host.
+
+### Implementation Details/Notes/Constraints [optional]
+
+* The current focus of the enhancement is for SaaS users.
+
+* Only SNO is supported as there's no Baremetal support for arm64 in 4.9 (arm packages for Ironic, etc).
+
+* Late binding is not covered by this enhancement, but decoupling the discovery ISO generation from the cluster should be simple enough. I.e. adding `cpu_architecture` property to `infraenv_create_params` and override the architecture defined in the cluster. Also, architecture validation should be probably added to host discovery conditions (i.e. to ensure that the host CPU architecture is similar to the cluster's architecture).
+
+* When supporting Day2, it should be handled by adding `cpu_architecture` property in `AddHostsClusterCreateParams`, which is needed for generating the proper discovery ISO. We could probably rely on the current validations of host discovery, i.e. the user should ensure architecture when generating the ISO.
+
+* Need to verify IPv6 flows: [OVN-Kubernetes](https://github.com/ovn-org/ovn-kubernetes)
+
+* LSO/OCS/CNV operators: add a validation for x86_64 CPU architecture. See:
+  * 'lso-requirements-satisfied'
+  * 'ocs-requirements-satisfied'
+  * 'cnv-requirements-satisfied'
+
+### Open Questions
+
+#### kube-api considerations
+
+##### ClusterImageSet
+Since we extract the openshift-baremetal-install binary in the assisted-service (to create the manifests of the cluster), the x86_64 OCP release image has to be used. Which means that both arm64 and x86_64 release images must be provided. The OCP release image probably won't support the manifest list feature, so we couldn't rely on it either.
+
+Alternatives:
+  * We could add another URL for the x86_64 release image in ClusterImageSet.
+  * As a workaround, we could use something like [qemu-user-static](https://github.com/multiarch/qemu-user-static), which enables invocation of multi-architecture containers using QEMU. I.e it might be sufficient just for running the arm64 openshift-baremetal-install binary, though not sure we want to rely on such solution.
+
+For now, we'll try to start with supporting None platform only. So we could use the x86_64 openshift-install binary embedded in the arm release. See: [extract Linux binaries for multiple architectures](https://github.com/openshift/oc/pull/816)
+
+##### AgentServiceConfig
+`osImages` property currently contains a single set of images for each OCP version.
+E.g.
+```
+osImages:
+- openshiftVersion: '4.8'
+  rootFSUrl: https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-live-rootfs.x86_64.img
+  url: https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-4.8.2-x86_64-live.x86_64.iso
+  version: 48.84.202107202156-0
+```
+
+We should consider API changes to support multiple architectures for each version. Perhaps just add `cpuArchitecture` property for each image, i.e. multiple images for each version. Would need to modify parsing handling obviously.
+
+### UI Impact
+
+* New cluster dialog:
+  * Add 'CPU Architecture' select box.
+  * Should be displayed only for versions >= 4.9.
+  * Should be disabled on update cluster dialog.
+* Generate Discovery ISO:
+  * Add an indication for the selected CPU architecture.
+
+### Test Plan
+
+- Need dedicated arm64 machines for running tests in the CI.
+- Need to consider which tests to run for arm64 clusters in the CI (using assisted-test-infra). We could probably start with merely the positive flow of full SNO installation, and add tests afterwards if required.
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


docs/enhancements/multiple-cpu-platforms.md
```diff
@@ -0,0 +1,144 @@
...
+* Cluster installation flow should be identical from user perspective regardless of the platform.
+
+### Non-Goals
+
+* No need to build assisted-service to arm64. I.e. The assisted-service would remain for x86_64 only.
+* Hybrid clusters - no need to support clusters with hosts of multiple platforms (e.g. x86_64 + arm64).
+* Implementation for kube-api (though design should take into account).
+* Only SNO will be supported in first phase (i.e. no day2/late binding/etc).
+
+## Proposal
```
avishayt, line 35: Need to add a host validation that checks that its CPU architecture matches the architecture defined in the cluster property.

docs/enhancements/multiple-cpu-platforms.md
```diff
@@ -0,0 +1,144 @@
+---
+title: multiple-cpu-platforms
+authors:
+  - "@danielerez"
+creation-date: 2021-07-29
+last-updated: 2021-07-29
+---
+
+# Multiple CPU platforms support
```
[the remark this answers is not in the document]
ybettan: +1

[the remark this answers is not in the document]
danielerez: Yeah, just wanted to be consistent with the naming in docs and code:) Can change if looks better...

docs/enhancements/multiple-cpu-platforms.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Summary
+
+The default host CPU platform supported for installation is x86_64 in Assisted Installer. In order to support additional platforms, the assisted-service should accept and handle selection of other platforms. The focus of this enhancement is specifically arm64, but the introduced infrastructure should facilitate future support of more platforms (e.g. PPC).
+
+## Motivation
+
+Users own machines of numerous variants, hence, support for platforms other than x86_64
+could be highly significant. E.g. creating clusters with arm64 hosts could be crucial
+for some workloads.
```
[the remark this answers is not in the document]
danielerez: thanks, sounds good.

docs/enhancements/multiple-cpu-platforms.md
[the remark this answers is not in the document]
avishayt: OpenShift docs and code?  If so we should be consistent.  I just hope it won't cause confusion with the other platform (baremetal/vsphere/etc).

docs/enhancements/multiple-cpu-platforms.md
```diff
@@ -0,0 +1,144 @@
...
+  - "@danielerez"
+creation-date: 2021-07-29
+last-updated: 2021-07-29
+---
+
+# Multiple CPU platforms support
+
+## Summary
+
+The default host CPU platform supported for installation is x86_64 in Assisted Installer. In order to support additional platforms, the assisted-service should accept and handle selection of other platforms. The focus of this enhancement is specifically arm64, but the introduced infrastructure should facilitate future support of more platforms (e.g. PPC).
```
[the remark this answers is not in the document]
danielerez: sure makes sense, will rephrase.

docs/enhancements/multiple-cpu-platforms.md
```diff
@@ -0,0 +1,144 @@
...
+
+Users own machines of numerous variants, hence, support for platforms other than x86_64
+could be highly significant. E.g. creating clusters with arm64 hosts could be crucial
+for some workloads.
+
+### Goals
+
+* The host side images should be built also to arm64.
+* The assisted-service should support the arm64 variant of RHCOS images and OCP release image (for OpenShift >= 4.9).
+* Cluster creation should support specification of CPU platform.
```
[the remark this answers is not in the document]
danielerez: ok I'll move it to the proposal part then.

docs/enhancements/multiple-cpu-platforms.md
```diff
@@ -0,0 +1,144 @@
...
+* Cluster installation flow should be identical from user perspective regardless of the platform.
+
+### Non-Goals
+
+* No need to build assisted-service to arm64. I.e. The assisted-service would remain for x86_64 only.
+* Hybrid clusters - no need to support clusters with hosts of multiple platforms (e.g. x86_64 + arm64).
+* Implementation for kube-api (though design should take into account).
```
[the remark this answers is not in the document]
danielerez: Just the kube-api support and impl is not part of the enhancement, but we still need to consider it in the design.

docs/enhancements/multiple-cpu-platforms.md
```diff
@@ -0,0 +1,144 @@
...
+* Only SNO will be supported in first phase (i.e. no day2/late binding/etc).
```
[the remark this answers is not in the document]
danielerez: The phrasing was probably too harsh, meant that the arm support is currently implemented mainly with SNO in mind. So for day2, there's an issue with getting the right cpu arch from the cluster. For late binding, there's some missing impl. So just adding more masters/workers shouldn't be an issue indeed.

docs/enhancements/multiple-cpu-platforms.md
```diff
@@ -0,0 +1,144 @@
...
+
+### OCP versions
+
+[Versions](https://github.com/openshift/assisted-service/blob/master/internal/versions/versions.go) should
+have CPU platform granularity for each API. So that OCP release images and RHCOS images of multiple platforms could be supported.
+
+The [default_ocp_versions](https://github.com/openshift/assisted-service/blob/master/data/default_ocp_versions.json) json file should include Release/RHCOS images of multiple platforms for versions >= 4.9.
+E.g.
+```
+"4.9": {
```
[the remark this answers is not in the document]
danielerez: Not sure we want to force it actually:/ But can use some env var indeed for supported archs.

docs/enhancements/multiple-cpu-platforms.md
```diff
@@ -0,0 +1,144 @@
...
+### Cluster creation and ISO generation
+
+When creating a new cluster, `cpu_platform` property should be supported to define which hosts can be registered to it. This property would be added to cluster-create-params and stored in cluster upon creation. If not specified, defaulted to x86_64 as a fallback.
+
+An example for the swagger schema:
+
+```
+cpu_platform:
+   type: string
+   enum: ['x86_64', 'arm64']
```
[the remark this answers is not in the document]
danielerez: good idea, so we could list only the supported ones in the UI, and filter according to it, etc.

docs/enhancements/multiple-cpu-platforms.md
danielerez: Isn't it relevant only for late-binding scenarios? I mean, when generating a discovery ISO, we would use only the platform defined in the cluster any way. So no need for additional validation.

avishayt: Yes, my brain always works in late binding mode :)

docs/enhancements/multiple-cpu-platforms.md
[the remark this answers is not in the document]
avishayt: I mentioned the env variable below.  What I meant here is given the list of supported architectures, we want to force all of them for all versions.

docs/enhancements/multiple-cpu-platforms.md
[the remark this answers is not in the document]
danielerez: It's being used interchangeably in ocp docs and code. So I'll just change to architecture to be consistent.

docs/enhancements/multiple-cpu-platforms.md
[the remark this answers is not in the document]
danielerez: This is problematic for old versions of ocp, as only >=4.9 would support arm64. So we could enforce it only for newer versions. Do you think it make sense, or better just keep it simple for now and not validate it? :/

docs/enhancements/multiple-cpu-platforms.md
danielerez: assumed that :) ok, I'll add a comment about it then.

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,149 @@
...
+Although x86_64 is popular in many use cases, other CPU architectures are common in other segments.  For example, ARM is popular in edge use cases.  OpenShift has recently added support for arm64 and it is important to add this support to Assisted Installer.
+
+### Goals
+
+* Allow installation of clusters that are running on arm64 hosts.
+* The introduced infrastructure should facilitate future support of more architectures (e.g. PPC).
+
+### Non-Goals
+
+* The assisted-service will remain for x86_64 only.
```
flaper87, line 26: I was a bit confused by the phrasing here. What do you say if we say something like: 

```suggestion
* The assisted-service and assisted-image-service will not be built for arm64.
```

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,149 @@
...
+* No support for clusters with hosts of multiple architectures (e.g. x86_64 + arm64).
+* Implementation for kube-api (though design should take into account).
```
flaper87, line 28: Unsure of what you mean by this?

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,149 @@
...
+* Day2 and late-binding won't be supporting in first phase.
+
+## Proposal
+
+### Host images
+
+The following images should be built also to arm64:
```
flaper87, line 35:
```suggestion
The following images should also be built for arm64:
```

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,149 @@
...
+This should be handled in similar to the current flows by modifying relevant Dockerfiles to support arm64 build. The images should be published to the same registries and follow same naming conventions. I.e. The best approach would be using the same image URI for any architecture. Which means, minimal changes in the assisted-service, as the architecture specific image would be fetched according to the machine pulling the image.
+
+### OCP versions
+
+[Versions](https://github.com/openshift/assisted-service/blob/master/internal/versions/versions.go) should
+have CPU architecture granularity for each API. So that OCP release images and RHCOS images of multiple architectures could be supported.
+
+The assisted-service should support the arm64 variant of RHCOS images and OCP release image (for OpenShift >= 4.9).
+
+The OPENSHIFT_VERSIONS environment variable should include Release/RHCOS images of multiple architectures for versions >= 4.9.
```
flaper87, line 49: Is backwards compatibility going to be guaranteed? 

Is it expected that users specifying a `4.9` image will have to use only the new format? Anyway to automatically convert from the old version of this variable to the new one?

Just mentioning things that may be worth noting in this enhancement.

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,149 @@
...
+E.g.
+```
+osImages:
+- openshiftVersion: '4.8'
+  rootFSUrl: https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-live-rootfs.x86_64.img
+  url: https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-4.8.2-x86_64-live.x86_64.iso
+  version: 48.84.202107202156-0
+```
+
+We should consider API changes to support multiple architectures for each version. Perhaps just add `cpuArchitecture` property for each image, i.e. multiple images for each version. Would need to modify parsing handling obviously.
```
flaper87, line 131: I like the idea of just adding a `cpuArchitecture` field but I also think it would be better for us to be consistent with the structure used elsewhere.

That said, the current structure is already inconsistent with the `OPENSHIFT_VERSIONS` one so I guess adding the field should do it

docs/enhancements/arm64-cpu-architecture.md
danielerez: indeed, will add.

docs/enhancements/arm64-cpu-architecture.md
danielerez: Just that changes in ClusterImageSet CRD and the operator are out of the scope in this phase.

docs/enhancements/arm64-cpu-architecture.md
danielerez: sure, will fix:)

docs/enhancements/arm64-cpu-architecture.md
danielerez: So the old format will still be supported (for backwards compatibility indeed). But, in order to use multiple architectures the proper images would have to be specified using the new format. So it is expected to use the new format if additional architectures are needed.

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,152 @@
...
+* Late binding is not covered by this enhancement, but decoupling the discovery ISO generation from the cluster should be simple enough. I.e. adding `cpu_architecture` property to `infraenv_create_params` and override the architecture defined in the cluster. Also, architecture validation should be probably added to host discovery conditions (i.e. to ensure that the host CPU architecture is similar to the cluster's architecture).
+
+### Open Questions
+
+#### kube-api considerations
+
+##### ClusterImageSet
+Since we extract the openshift-baremetal-install binary in the assisted-service (to create the manifests of the cluster), the x86_64 OCP release image has to be used. Which means that both arm64 and x86_64 release images must be provided.
+  * Should we perhaps add another URL for the x86_64 release image in ClusterImageSet?
+  * Maybe we could assume that a x86_64 release image is available by default for each 'y-stream' version?
```
mhrivnak, line 120: What do you mean by "for each `y-stream` version"? I'm not sure how that relates to a ClusterImageSet, which references a specific openshift release.

docs/enhancements/arm64-cpu-architecture.md
mhrivnak: If we're breaking compatibility, this may be the right opportunity to separate the definition of openshift release images from RHCOS images, and specify each separately as the kube API does. Perhaps `OPENSHIFT_VERSIONS` and `RHCOS_VERSIONS`. That would eliminate a lot of confusion and duplication. @avishayt WDYT?

avishayt: @mhrivnak Can you provide an example of `OPENSHIFT_VERSIONS` and `RHCOS_VERSIONS`?  What would this break for existing customers?

mhrivnak: Something like what's below. The enhancement proposes maintaining compatibility with the old OPENSHIFT_VERSIONS format in parallel with a new format. I'm just adjusting that new format here to put RHCOS image data elsewhere. The new format is backward-incompatible, but the key is to support both formats concurrently.

OS_IMAGES
```
"4.9": {
      "x86_64": {
            "url": "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/latest-4.9/rhcos-live.x86_64.iso",
            "rootfsurl": "https://mirror.openshift.com/pub/openshift-v4/x86_64/dependencies/rhcos/pre-release/latest-4.9/rhcos-live-rootfs.x86_64.img",
            "version": "49.84.202107032011-0"
      },
}
```

OPENSHIFT_VERSIONS
```
"4.9": {
   "display_name": "4.9.0-0.nightly",
   "support_level": "beta",
   "version": "4.9.0-0.nightly-2021-07-05-092650",
   "image": "quay.io/openshift-release-dev/ocp-release-nightly:4.9.0-0.nightly-2021-07-05-092650",
}
```

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,152 @@
...
+As an Assisted-Installer user, I need to install a Single-Node OpenShift cluster on an arm64 host.
+
+### Implementation Details/Notes/Constraints [optional]
+
+* The current focus is for SNO flows, but other scenarios could be supported.
+
+* Late binding is not covered by this enhancement, but decoupling the discovery ISO generation from the cluster should be simple enough. I.e. adding `cpu_architecture` property to `infraenv_create_params` and override the architecture defined in the cluster. Also, architecture validation should be probably added to host discovery conditions (i.e. to ensure that the host CPU architecture is similar to the cluster's architecture).
+
+### Open Questions
+
```
eranco74, line 114: Do we need to do anything in order to accommodate CNV and LSO operators?
Users can enable them from AI UI, unsure what will be the result on ARM.

docs/enhancements/multiple-cpu-platforms.md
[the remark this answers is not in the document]
danielerez: @avishayt do you think we can support SNO and multiple hosts, but without day2 for now? I mean, as supporting day2 is an additional effort, could we keep it disabled for arm64 for now? or should it be in the scope of this enhancement?

docs/enhancements/arm64-cpu-architecture.md
danielerez: @gamli75 do you know if these operators support arm? maybe we could just disable the installation on arm for now?

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,152 @@
...
+#### kube-api considerations
+
+##### ClusterImageSet
+Since we extract the openshift-baremetal-install binary in the assisted-service (to create the manifests of the cluster), the x86_64 OCP release image has to be used. Which means that both arm64 and x86_64 release images must be provided.
```
mhrivnak, line 118: It looks like we will not need an entire x86_64 payload just to get the binary: https://github.com/openshift/oc/pull/816

docs/enhancements/arm64-cpu-architecture.md
mhrivnak: Both are out of scope for the first OCP deliverable of arm support.

avishayt: Good call @eranco74 .  We have the following validations:
      - 'lso-requirements-satisfied'
      - 'ocs-requirements-satisfied'
      - 'cnv-requirements-satisfied'
We already check various cluster properties when deciding if those validations pass or fail.  We should add a check to ensure that the CPU architecture equals x86_64.

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,192 @@
...
+
+### OCP versions
+
+[Versions](https://github.com/openshift/assisted-service/blob/master/internal/versions/versions.go) should
+have CPU architecture granularity for each API. So that OCP release images and RHCOS images of multiple architectures could be supported.
+
+The assisted-service should support the arm64 variant of RHCOS images and OCP release image (for OpenShift >= 4.9).
+
+As a preparation for supporting multiple CPU architectures in Version, the RHCOS images information from OPENSHIFT_VERSIONS environment variable should be extracted into a new variable (e.g. RHCOS_VERSIONS).
+We should probably use a simple array structure instead of mapping, to align with the [AgentServiceConfig](https://github.com/openshift/assisted-service/blob/77cb9d3348dafd2abd61fd7c8066e7e7a7805d95/config/samples/agent-install.openshift.io_v1beta1_agentserviceconfig.yaml#L18-L22) structure.
```
carbonin, line 52: I'm not sure this is an important consideration. I'd rather the structure @mhrivnak suggested in https://github.com/openshift/assisted-service/pull/2305#discussion_r681169284.

danielerez: Good point, perhaps we should just keep using the current structure, for example as already used in https://github.com/openshift/assisted-image-service#configuration (RHCOS_VERSIONS).
Since the operator have to parse the json object any way (and only once), maybe it's better to keep the mapping?
@flaper87 @mhrivnak wdyt?

docs/enhancements/multiple-cpu-platforms.md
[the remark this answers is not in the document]
danielerez: Will support only none platform for now, as baremetal not available yet for arm64 in 4.9.

docs/enhancements/arm64-cpu-architecture.md
```diff
@@ -0,0 +1,152 @@
...
+
+* Late binding is not covered by this enhancement, but decoupling the discovery ISO generation from the cluster should be simple enough. I.e. adding `cpu_architecture` property to `infraenv_create_params` and override the architecture defined in the cluster. Also, architecture validation should be probably added to host discovery conditions (i.e. to ensure that the host CPU architecture is similar to the cluster's architecture).
+
+### Open Questions
+
+#### kube-api considerations
+
+##### ClusterImageSet
+Since we extract the openshift-baremetal-install binary in the assisted-service (to create the manifests of the cluster), the x86_64 OCP release image has to be used. Which means that both arm64 and x86_64 release images must be provided.
+  * Should we perhaps add another URL for the x86_64 release image in ClusterImageSet?
```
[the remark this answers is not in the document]
danielerez: Seems that manifest list won't be supported for payload images in 4.9.

docs/enhancements/arm64-cpu-architecture.md
danielerez: So we'll try to align with the k8s API convention if not too cumbersome: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

docs/enhancements/arm64-cpu-architecture.md
danielerez: Turns out it's only supported for none platform, so we'll try to use openshift-install

flaper87 (commented):
/approve  

I would rather have 2 approvals for this one since there were comments from multiple folks. This has my lgtm.

danielerez:
> /approve
> 
> I would rather have 2 approvals for this one since there were comments from multiple folks. This has my lgtm.

Thanks:)
@avishayt @mhrivnak Can you please take another look?

flaper87 (approved):
Alright, this has been pending long enough. I'll send it through now and we can edit/fix things in follow-up PRs if needed.

/cc @avishayt @mhrivnak

PR #2313, opened 2021-07-30
OCPBUGSM-32618: Create worker's spoke machines in the "openshift-machine-api" namespace

pawanpinjarkar:
# Assisted Pull Request

## Description
The previous code was trying to create spoke resources such as BMH, machine, secret, in the same namespace on the spoke cluster as in the hub cluster i.e. assisted-spoke-cluster namespace. This namespace is not present on the spoke cluster and results in a failure while creating a spoke BMH (surely we can create the namespace but we shouldn't).

This change fixes this bug and correctly creates the resources such as spoke BMH, and spoke machine on the spoke cluster when adding workers on day2. Use "openshift-machine-api" namespace just like the other master agent/BMH/machines on the spoke cluster. 

- [ ] New Feature
- [x] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
    Create OCP cluster using dev-scripts. Deploy assisted service operator and deploy spoke cluster. Create a BMH on the hub cluster with worker role on a day2 cluster.  Spoke BMH should be created in openshift-machine-api namespace on the spoke cluster.
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -73,6 +73,7 @@ const (
 	MACHINE_ROLE                        = "machine.openshift.io/cluster-api-machine-role"
 	MACHINE_TYPE                        = "machine.openshift.io/cluster-api-machine-type"
 	MCS_CERT_NAME                       = "ca.crt"
+	OPENSHIFT_MACHINE_API_NAMESPACE     = "openshift-machine-api"
 )
 
 var (
@@ -700,6 +701,7 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 	}
 
 	_, err = r.ensureSpokeBMH(ctx, log, spokeClient, bmh, machine, agent)
+
 	if err != nil {
 		log.WithError(err).Errorf("failed to create or update spoke BareMetalHost")
 		return reconcileError{err}
@@ -1054,7 +1056,7 @@ func (r *BMACReconciler) ensureSpokeBMHSecret(ctx context.Context, log logrus.Fi
 	if err != nil {
 		return secret, err
 	}
-	secretSpoke, mutateFn := r.newSpokeBMHSecret(secret, bmh)
+	secretSpoke, mutateFn := r.newSpokeBMHSecret(secret)
 	if result, err := controllerutil.CreateOrUpdate(ctx, spokeClient, secretSpoke, mutateFn); err != nil {
 		return nil, err
 	} else if result != controllerutil.OperationResultNone {
@@ -1067,7 +1069,7 @@ func (r *BMACReconciler) newSpokeBMH(bmh *bmh_v1alpha1.BareMetalHost, machine *m
 	bmhSpoke := &bmh_v1alpha1.BareMetalHost{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      bmh.Name,
-			Namespace: bmh.Namespace,
+			Namespace: OPENSHIFT_MACHINE_API_NAMESPACE,
 		},
 	}
 	mutateFn := func() error {
@@ -1099,11 +1101,11 @@ func (r *BMACReconciler) newSpokeBMH(bmh *bmh_v1alpha1.BareMetalHost, machine *m
 	return bmhSpoke, mutateFn
 }
 
-func (r *BMACReconciler) newSpokeBMHSecret(secret *corev1.Secret, bmh *bmh_v1alpha1.BareMetalHost) (*corev1.Secret, controllerutil.MutateFn) {
+func (r *BMACReconciler) newSpokeBMHSecret(secret *corev1.Secret) (*corev1.Secret, controllerutil.MutateFn) {
 	secretSpoke := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      secret.Name,
-			Namespace: bmh.Namespace,
+			Namespace: OPENSHIFT_MACHINE_API_NAMESPACE,
 		},
 	}
 	mutateFn := func() error {
@@ -1119,7 +1121,7 @@ func (r *BMACReconciler) newSpokeMachine(bmh *bmh_v1alpha1.BareMetalHost, cluste
 	machine := &machinev1beta1.Machine{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      machineName,
-			Namespace: bmh.Namespace,
+			Namespace: OPENSHIFT_MACHINE_API_NAMESPACE,
 		},
 	}
 	mutateFn := func() error {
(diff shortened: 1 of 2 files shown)


flaper87:
> This change fixes a bug to correctly create the resources such as spoke BMH, and spoke machine on the spoke cluster when adding workers on day2. Use "openshift-machine-api" namespace just like the other master agent/BMH/machines on the spoke cluster.

Could you elaborate on why this is needed? The above explains what the fix is but it's unclear what the problem is. Please, elaborate on what the actual problem is and why this changes fixes it :)

flaper87:
/retitle OCPBUGSM-32618: Create worker's spoke machines in the "openshift-machine-api" namespace

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -687,7 +688,7 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		return reconcileError{err}
 	}
 
-	machine, err := r.ensureSpokeMachine(ctx, log, spokeClient, bmh, cd)
+	spokeMachine, err := r.ensureSpokeMachine(ctx, log, spokeClient, bmh, cd)
```
flaper87, line 691: As much as I like the explicitness this change brings, I gotta say that it's generally better to avoid "unnecessary" changes as they have the potential to create conflicts in the future. Taking this PR as an example, if you revert the variables renames, the change itself should be quite small (~4 LOC in this file)

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -73,6 +73,7 @@ const (
 	MACHINE_ROLE                        = "machine.openshift.io/cluster-api-machine-role"
 	MACHINE_TYPE                        = "machine.openshift.io/cluster-api-machine-type"
 	MCS_CERT_NAME                       = "ca.crt"
+	OPENSHIFT_MACHINE_API               = "openshift-machine-api"
```
flaper87, line 76:
```suggestion
	OPENSHIFT_MACHINE_API_NAMESPACE               = "openshift-machine-api"
```

flaper87 (changes requested):

internal/controller/controllers/bmh_agent_controller.go
pawanpinjarkar: Done

internal/controller/controllers/bmh_agent_controller.go
pawanpinjarkar: Done.

flaper87:
> (surely we can create the namespace but we shouldn't).

Could you elaborate on why we shouldn't create the namespace? I don't see anything wrong with creating the namespace, but maybe I am missing something.

To me, this change makes sense for the following reason: These resources belong to the cluster they are being created in and `openshift-machine-api` is where normally these Machine/BMH resources live.

flaper87:
> To me, this change makes sense for the following reason: These resources belong to the cluster they are being created in and `openshift-machine-api` is where normally these Machine/BMH resources live.

A counterpoint to what I just said would be that, from a user perspective, it would make more sense to find the resources in the same namespace as in the hub cluster. We will have to document this properly, regardless.

pawanpinjarkar:
> > (surely we can create the namespace but we shouldn't).
> 
> Could you elaborate on why we shouldn't create the namespace? I don't see anything wrong with creating the namespace, but maybe I am missing something.
> 
> To me, this change makes sense for the following reason: These resources belong to the cluster they are being created in and `openshift-machine-api` is where normally these Machine/BMH resources live.

That's correct.

pawanpinjarkar:
> > To me, this change makes sense for the following reason: These resources belong to the cluster they are being created in and `openshift-machine-api` is where normally these Machine/BMH resources live.
> 
> A counterpoint to what I just said would be that, from a user perspective, it would make more sense to find the resources in the same namespace as in the hub cluster. We will have to document this properly, regardless.

If that's the case, then I think even the master nodes should also be created same as in the hub cluster. In other words, everything should only be created in assisted-spoke-cluster namespace on hub and spoke cluster.

pawanpinjarkar:
> > > To me, this change makes sense for the following reason: These resources belong to the cluster they are being created in and `openshift-machine-api` is where normally these Machine/BMH resources live.
> > 
> > 
> > A counterpoint to what I just said would be that, from a user perspective, it would make more sense to find the resources in the same namespace as in the hub cluster. We will have to document this properly, regardless.
> 
> If that's the case, then I think even the master resources on the spoke should also be created same as in the hub cluster. In other words, everything should only be created in assisted-spoke-cluster namespace on hub and spoke cluster.

@flaper87 , I think this will help. The machine controller only checks openshift-machine-api namespace during reconciliaiton so we have to always use that namespace.


```
[root@hardprov-fx2-21 ~]# oc logs machine-api-controllers-7bd66b7c5-rwff4  -c machine-controller -n openshift-machine-api --kubeconfig=clusterconfig
I0802 03:04:32.991345       1 main.go:191] baremetal-controller-manager "msg"="Found API group metal3.io/v1alpha1"  
I0802 03:04:32.991524       1 main.go:121] Watching machine-api objects only in namespace "openshift-machine-api" for reconciliation.
```

flaper87:
> > > > To me, this change makes sense for the following reason: These resources belong to the cluster they are being created in and `openshift-machine-api` is where normally these Machine/BMH resources live.
> > > 
> > > 
> > > A counterpoint to what I just said would be that, from a user perspective, it would make more sense to find the resources in the same namespace as in the hub cluster. We will have to document this properly, regardless.
> > 
> > 
> > If that's the case, then I think even the master resources on the spoke should also be created same as in the hub cluster. In other words, everything should only be created in assisted-spoke-cluster namespace on hub and spoke cluster.
> 
> @flaper87 , I think this will help. The machine controller only checks openshift-machine-api namespace during reconciliaiton so we have to always use that namespace.
> 
> ```
> [root@hardprov-fx2-21 ~]# oc logs machine-api-controllers-7bd66b7c5-rwff4  -c machine-controller -n openshift-machine-api --kubeconfig=clusterconfig
> I0802 03:04:32.991345       1 main.go:191] baremetal-controller-manager "msg"="Found API group metal3.io/v1alpha1"  
> I0802 03:04:32.991524       1 main.go:121] Watching machine-api objects only in namespace "openshift-machine-api" for reconciliation.
> ```

This is because, by default, BMO only watches the `openshift-machine-api` namespace :)

flaper87 (approved):

mkowalski:
>This is because, by default, BMO only watches the openshift-machine-api namespace :)

And for performance reasons we should try to keep it like this for as long as possible. I can't find the reference now, but when discussing it with someone from the metal team we have discovered that enabling BMO watching all the namespaces can increase the resource consumption of the whole cluster

PR #2323, opened 2021-08-02
MGMT-7426: Allow installing on OpenStack in case of userManagedNetwor…

eranco74:
…king

# Assisted Pull Request

## Description
Currently Assisted installer isn't allowing installations on OpenStack.
It seems that the issues we had with OpenStack are related to the baremetal platform and not to the assisted installer itself.
hence this PR enable installing on OpenStack VMs in case or userManagedNetworking (None platform).
 
- Is this PR relying on CI for an e2e test run? no 
- Should this PR be tested in a specific environment? yes it should be tested with OpenStack platform
- Any logs, screenshots, etc that can help with the review process? no

-->

- [x] New Feature
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?
CI for regression.
Pending OpenStack environment for more tests

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

No need for review yet
/hold

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/transition_test.go
+++ b/internal/host/transition_test.go
@@ -1993,7 +1993,7 @@ var _ = Describe("Refresh Host", func() {
 			{
 				name:          "insufficient virtualization cpu",
 				hostID:        strfmt.UUID("054e0100-f50e-4be7-874d-73861179e40d"),
-				inventory:     hostutil.GenerateMasterInventoryWithHostnameAndCpuFlags("master", []string{"fpu", "vme", "de", "pse", "tsc", "msr"}),
+				inventory:     hostutil.GenerateMasterInventoryWithHostnameAndCpuFlags("master", []string{"fpu", "vme", "de", "pse", "tsc", "msr"}, "RHEL"),
 				role:          models.HostRoleMaster,
 				srcState:      models.HostStatusDiscovering,
 				dstState:      models.HostStatusInsufficient,
@@ -2838,7 +2838,7 @@ var _ = Describe("Refresh Host", func() {
 				imageStatuses:      map[string]*models.ContainerImageAvailability{common.TestDefaultConfig.ImageName: common.TestImageStatusesSuccess},
 				role:               models.HostRoleMaster,
 				statusInfoChecker: makeValueChecker(formatStatusInfoFailedValidation(statusInfoInsufficientHardware,
-					"Platform OpenStack Compute is forbidden")),
+					fmt.Sprintf("Platform %s is allowed only for Single Node OpenShift or user-managed networking", OpenStackPlatform))),
 				validationsChecker: makeJsonChecker(map[validationID]validationCheckResult{
 					IsConnected:                   {status: ValidationSuccess, messagePattern: "Host is connected"},
 					HasInventory:                  {status: ValidationSuccess, messagePattern: "Valid inventory exists for the host"},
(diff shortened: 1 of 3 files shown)


internal/host/validator.go
```diff
@@ -395,15 +395,24 @@ func (v *validator) isValidPlatform(c *validationContext) ValidationStatus {
 	if c.inventory.SystemVendor == nil {
 		return ValidationError
 	}
-	return boolValue(!funk.ContainsString(invalidPlatforms, c.inventory.SystemVendor.ProductName))
+	if funk.ContainsString(invalidPlatforms, c.inventory.SystemVendor.ProductName) {
+		// In case there is no cluster validation is pending
```
filanov, line 399: Or late binding, @yevgeny-shnaidman @ravidbro how does refresh status will filter out validations for the pool?

eranco74: no cluster == late binding no?

yevgeny-shnaidman: currently validations are either containing cluster (day1, day2) or infra-env (pool). To filter out validation for pool, simply check if infraenv in validation context s not null. see example:
https://github.com/openshift/assisted-service/blob/8f89ccc17c91559b00d9b5e49d4dacc1d63ad526/internal/host/validator.go#L338

filanov: `no cluster` sounds clearer then `infraenv == nil`, i would like to make sure that we don't have two different ways in the code.

yevgeny-shnaidman: currently all validations are using infraenv == nil. You can change it as you like , as long as it is the same for all validaitons

internal/host/validator.go
```diff
@@ -395,15 +395,24 @@ func (v *validator) isValidPlatform(c *validationContext) ValidationStatus {
...
+	return boolValue(true)
 }
 
 func (v *validator) printValidPlatform(c *validationContext, status ValidationStatus) string {
 	switch status {
 	case ValidationSuccess:
 		return fmt.Sprintf("Platform %s is allowed", c.inventory.SystemVendor.ProductName)
 	case ValidationFailure:
-		return fmt.Sprintf("Platform %s is forbidden", c.inventory.SystemVendor.ProductName)
+		return fmt.Sprintf("Platform %s is forbidden, the platform is supported only for user managed networking", c.inventory.SystemVendor.ProductName)
```
avishayt, line 415: Simplify sentence, and also don't assume that users know the relationship between SNO and user-managed networking.

```suggestion
		return fmt.Sprintf("Platform %s is only supported for Single Node OpenShift or user-managed networking", c.inventory.SystemVendor.ProductName)
```

internal/host/validator.go
eranco74: I'm not sure what c.infraEnv != nil means (that's what we check in isMachineCidrDefined !)
```
func (v *validator) isMachineCidrDefined(c *validationContext) ValidationStatus {
	if c.infraEnv != nil {
		return ValidationSuccessSuppressOutput
	}
```
If c.infraEnv == nil host beolngs to a cluster?

In order to check `c.cluster.UserManagedNetworking` I need to make sure `c.cluster != inl` anyway.
So I prefer to keep the code as is (assuming logic is correct)

avishayt: What does this mean?  How can a host not have an InfraEnv?  @ravidbro any idea?

filanov: yes, because of a logic in https://github.com/openshift/assisted-service/blob/master/internal/host/validator.go#L171-L227

internal/host/transition_test.go
```diff
@@ -4266,6 +4266,62 @@ var _ = Describe("Refresh Host", func() {
...
+
+		tests := []struct {
+			name                  string
+			hostPlatform          string
+			userManagedNetworking bool
+			dstState              string
+		}{
+			{
+				name:                  "validate platform OpenStack and userManagedNetwork false",
+				hostPlatform:          "OpenStack Compute",
```
filanov, line 4286: maybe it should be a const that will be used in `invalidPlatforms` as well? it will be easier to find references this way.

avishayt: Good call.  Can follow the vsphere code in internal/common/platform.go

internal/host/validator.go
filanov: Suggesting to use `if c.infraEnv != nil` here and then fix everything with https://issues.redhat.com/browse/MGMT-7459

avishayt (approved):

filanov (approved):

PR #2324, opened 2021-08-02
MGMT-7114: support CPU architectures in Versions

danielerez:
# Assisted Pull Request

## Description

As a preparation to support arm64 CPU architecture, introduced the following changes:
- Added openshift-images object to swagger.
   - Added openshift-images object that represents rhcos/release images and versions for a specific cpu architecture.
   - Added images property to openshift-version: contains a map of OCP version to openshift-images object.
-  Added 'cpuArchitecture' to Versions API signatures.
   - Each API retrieves the requested value from openshiftVersions[key].Images according to the specified CPU architecture.
   - For older OCP versions (<4.9), fallback to x86_64 as default if 'Images' map doesn't exist.
   - Until arm64 architecture is added, every invocation to the API is used with the default architecture (x86_64).
   - Updated uploadISOs func to support multiple CPU architectures. The only effect for now is the addition of architecture to the rhcos object templates.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

As this PR is adding an infrastructure for later arm64 support, no substantial impact in behaviour should occur. I.e. added tests should cover the changes.

## Assignees

/cc @YuviGold 
/cc @tsorya 
/cc @filanov 
/cc @carbonin 
/cc @osherdp 
/cc @avishayt 

## Checklist

## Reviewers Checklist


swagger.yaml
```diff
@@ -5605,6 +5605,11 @@ definitions:
       rhcos_version:
         type: string
         description: Build ID of the RHCOS image.
+      images:
+        type: object
+        additionalProperties:
+          $ref: '#/definitions/openshift-images'
+        description: RHCOS/Release images and versions for each CPU architectures.
```
osherdp, line 5612: Am I wrong or it duplicates the above fields?
In case we do need to maintain backwards compatibility, maybe change the name of the field from ``images`` to ``extra_versions`` or maybe ``per_arch_versions``

swagger.yaml
```diff
@@ -5613,6 +5618,31 @@ definitions:
         type: boolean
         description: Indication that the version is the recommended one.
 
+  openshift-images:
```
osherdp, line 5621: openshift-images -> openshift-version-info

swagger.yaml
osherdp, line 5612: *for each CPU architecture.

osherdp (changes requested):

internal/bminventory/inventory.go
```diff
@@ -1621,11 +1622,15 @@ func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, c
...
 	}
 
-	releaseImage, err := b.versionsHandler.GetReleaseImage(cluster.OpenshiftVersion)
+	releaseImage, err := b.versionsHandler.GetReleaseImage(cluster.OpenshiftVersion, versions.DefaultCPUArchitecture)
+	if err != nil {
+		msg := fmt.Sprintf("failed to get release image for cluster %s with openshift version %s", cluster.ID, cluster.OpenshiftVersion)
+		return errors.Wrapf(err, msg)
+	}
 
 	if err != nil {
```
filanov, line 1631: Isn't it a dead code? you added `if err != nil {` with return few lines above

danielerez: right, some leftovers in rebase, need to remove lines 1631-1634.

internal/assistedserviceiso/assistedserviceiso_api.go
```diff
@@ -90,7 +90,7 @@ func (a *assistedServiceISOApi) CreateISOAndUploadToS3(ctx context.Context, para
 	ignitionConfig := reIgnition.Replace(ignitionConfigSource)
 
 	username := ocm.UserNameFromContext(ctx)
-	srcISOName, err := a.objectHandler.GetBaseIsoObject(params.AssistedServiceIsoCreateParams.OpenshiftVersion)
+	srcISOName, err := a.objectHandler.GetBaseIsoObject(params.AssistedServiceIsoCreateParams.OpenshiftVersion, "x86_64")
```
YuviGold, line 93:
```suggestion
	srcISOName, err := a.objectHandler.GetBaseIsoObject(params.AssistedServiceIsoCreateParams.OpenshiftVersion, common.TestDefaultConfig.CPUArchitecture
```

internal/bminventory/inventory.go
```diff
@@ -421,10 +421,11 @@ func (b *bareMetalInventory) RegisterClusterInternal(
...
 	}
 
-	openshiftVersion, err := b.versionsHandler.GetVersion(swag.StringValue(params.NewClusterParams.OpenshiftVersion))
+	releaseImage, releaseVersion, err := b.versionsHandler.GetRelease(
+		swag.StringValue(params.NewClusterParams.OpenshiftVersion), versions.DefaultCPUArchitecture)
 	if err != nil {
-		err = errors.Errorf("Openshift version %s is not supported",
-			swag.StringValue(params.NewClusterParams.OpenshiftVersion))
+		err = errors.Errorf("Openshift version %s is not supported: %v",
+			swag.StringValue(params.NewClusterParams.OpenshiftVersion), err)
```
YuviGold, line 428:
```suggestion
		err = errors.Wrapf(err, "Openshift version %s is not supported",
			swag.StringValue(params.NewClusterParams.OpenshiftVersion))
```

internal/bminventory/inventory.go
```diff
@@ -1084,7 +1085,7 @@ func (b *bareMetalInventory) createAndUploadNewImage(ctx context.Context, log lo
 			return common.NewApiError(http.StatusInternalServerError, err)
 		}
 	} else {
-		baseISOName, err := b.objectHandler.GetBaseIsoObject(infraEnv.OpenshiftVersion)
+		baseISOName, err := b.objectHandler.GetBaseIsoObject(infraEnv.OpenshiftVersion, versions.DefaultCPUArchitecture)
```
YuviGold, line 1088: Don't we save the arch as part of the cluster?
If it is supposed to come on a following PR - then add a TODO with the ticket key every place needs to be changed.

swagger.yaml
danielerez: Right, that's should be a map of CPU arch to images, so it includes the same properties for each arch. The current properties are kept for backwards compatibility indeed of old versions. So not sure about the naming, as we'll probably just remove these properties when deleting the old versions. I mean, 'images' should be available only for new versions (> 4.9) instead of the current properties. Maybe 'images_per_arch'? Or just add more explanation in the description it make it clearer.

internal/versions/versions.go
```diff
@@ -26,17 +26,22 @@ type Versions struct {
...
-	GetVersion(openshiftVersion string) (*models.OpenshiftVersion, error)
-	IsOpenshiftVersionSupported(versionKey string) bool
+	GetRelease(openshiftVersion, cpuArchitecture string) (string, string, error)
+	IsOpenshiftVersionSupported(versionKey, cpuArchitecture string) bool
 	AddOpenshiftVersion(ocpReleaseImage, pullSecret string) (*models.OpenshiftVersion, error)
 }
 
+const (
+	DefaultCPUArchitecture = "x86_64"
+)
```
YuviGold, line 43: I think this property should be part of the cluster and `versions` should work only with what it gets from the calling function, it shouldn't be familiar with a default configuration.

swagger.yaml
danielerez: hmm, do you think it reflects the object better? as it includes just the info about the images. Maybe openshift-images-info?:/

internal/versions/versions.go
```diff
@@ -76,112 +81,192 @@ func (h *handler) ListSupportedOpenshiftVersions(ctx context.Context, params ope
...
+		if images, ok := h.openshiftVersions[versionKey].Images[cpuArchitecture]; ok {
+			return *images.ReleaseImage, nil
+		}
+		return "", errors.Errorf("Release image was missing for openshift version %s, CPU architecture %s", versionKey, cpuArchitecture)
 	}
 
 	if h.openshiftVersions[versionKey].ReleaseImage == nil {
-		return "", errors.Errorf("Release image was missing for openshift version %s", versionKey)
+		return "", errors.Errorf("Release image was missing for openshift version %s, CPU architecture %s", versionKey, cpuArchitecture)
 	}
```
YuviGold, line 102: These two logs look the same. They should be different so it would be easier to understand what's the failure.

YuviGold, line 102: I see a lot of code duplication between the functions. Please take a look into the [reflect](https://pkg.go.dev/reflect) module to see if we can use it to get properties by name.

YuviGold (commented):
Please notice to update the bump OCP job script, and the handle_versions verification script.

internal/assistedserviceiso/assistedserviceiso_api.go
filanov: probably not the one from a `TestDefault` maybe the one from versions.

internal/versions/versions.go
danielerez: It's still needed for functions as GetCPUArchitectures (for returning the default arch for older versions), and for IsOpenshiftVersionSupported (for having a default to fallback). We can move the constant though if it makes more sense, maybe in inventory?

internal/bminventory/inventory.go
danielerez: Right, it'll be part of an upcoming PR. will add a TODO.

swagger.yaml
osherdp: I think image means two different things (docker image and ISO image) and version means something more unified in this case
anyway, just a nit

swagger.yaml
osherdp: add move explanation to the description and also change the name to the one you suggested
sounds good

internal/assistedserviceiso/assistedserviceiso_api.go
YuviGold: right :)

danielerez:
Closing. Covered by https://github.com/openshift/assisted-service/pull/2443

PR #2143, opened 2021-06-30
[WIP] Bug 1976991: Add Spoke Machine ProviderSpec and BareMetalHost annotation

rwsu:
## Description

The ProviderSpec and BareMetalHost annotation are required for
the machine-controller to reconcile the Machine. If they are
not provided, the Machine does not reconcile, and the Machine
status is unavailable.

The Machine status is required for Discovery CSR auto approval.
The status contains the machine address which is used to match
with the Node name during CSR approval.

This patch partially fixes https://bugzilla.redhat.com/show_bug.cgi?id=1976991.

## List all the issues related to this PR

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [x] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- [ ] Are the title and description (in both PR and commit) meaningful and clear?
- [ ] Is there a bug required (and linked) for this change?
- [ ] Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #2325, opened 2021-08-02
OCPBUGSM-31544: Spoke worker BMH does not join cluster because CSR was not auto approved

pawanpinjarkar:
# Assisted Pull Request

## Description
The ProviderSpec and BareMetalHost annotation are required for the machine-controller to reconcile the Machine. If they are not provided, the Machine does not reconcile, and the Machine status is unavailable.

The Machine status is required for Discovery CSR auto-approval. The status contains the machine address which is used to match with the Node name during CSR approval.

This patch is built on top of the original patch suggested by Richard here https://github.com/openshift/assisted-service/pull/2143

This patch retrieves checksum and URL values from the ProviderSpec object of a master machine from the day 2 spoke cluster and sets the values when creating a spoke machine for a worker on day2 spoke cluster. 
All the master machines use the same value so it is safe to set the same value in worker machines as well instead of the dummy values originally used as  a temporary fix here https://github.com/openshift/assisted-service/pull/2143

co-author: Richard Su rwsu@redhat.com

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
  Add a worker node on the day 2 spoke cluster. After adding the ProviderSpec and the metal3.io/BareMetalHost annotation for the Machine, the first Discovery CSR is auto-approved. You can verify if the master and worker machines have the same checksum and image URL in the ProviderSpec object.
- [] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -67,6 +67,8 @@ const (
 	BMH_AGENT_MACHINE_CONFIG_POOL       = "bmac.agent-install.openshift.io/machine-config-pool"
 	BMH_INFRA_ENV_LABEL                 = "infraenvs.agent-install.openshift.io"
 	BMH_AGENT_INSTALLER_ARGS            = "bmac.agent-install.openshift.io/installer-args"
+	BMH_ANNOTATION                      = "metal3.io/BareMetalHost"
+	BMH_API_VERSION                     = "baremetal.cluster.k8s.io/v1alpha1"
 	BMH_DETACHED_ANNOTATION             = "baremetalhost.metal3.io/detached"
 	BMH_INSPECT_ANNOTATION              = "inspect.metal3.io"
 	BMH_HARDWARE_DETAILS_ANNOTATION     = "inspect.metal3.io/hardwaredetails"
@@ -691,7 +693,13 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		return reconcileError{err}
 	}
 
-	machine, err := r.ensureSpokeMachine(ctx, log, spokeClient, bmh, cd)
+	checksum, url, err := r.getChecksumAndURL(ctx, log, spokeClient)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get checksum and url value from master spoke machine")
+		return reconcileError{err}
+	}
+
+	machine, err := r.ensureSpokeMachine(ctx, log, spokeClient, bmh, cd, checksum, url)
 	if err != nil {
 		log.WithError(err).Errorf("failed to create or update spoke Machine")
 		return reconcileError{err}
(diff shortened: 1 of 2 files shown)


internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -944,8 +953,29 @@ func (r *BMACReconciler) ensureSpokeBMH(ctx context.Context, log logrus.FieldLog
...
+	}
+	providerSpecValue := string(machineList.Items[0].Spec.ProviderSpec.Value.Raw)
+
+	var providerSpecValueObj map[string]interface{}
+	err = json.Unmarshal([]byte(providerSpecValue), &providerSpecValueObj)
+	if err != nil {
+		return checksum, url, err
+	}
+	image := providerSpecValueObj["image"].(map[string]interface{})
+	checksum = fmt.Sprint(image["checksum"])
```
tsorya, line 972: image["checksum"] is not string?

flaper87 (commented):
I guess it's fine, for now, to re-use the same values as the `Master` machine. However, these values are not correct since the image linked there is not used to deploy the machine.

Another option would be to use the line to the `rhcos` image we actually used, since we know this info.

As for the hostname issue, I created #2382 which should make sure both, the `BMH` and the `Machine` will get the right hostname.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -944,8 +953,29 @@ func (r *BMACReconciler) ensureSpokeBMH(ctx context.Context, log logrus.FieldLog
...
-	machineSpoke, mutateFn := r.newSpokeMachine(bmh, clusterDeployment)
+// get spokeMachineMaster and retrieve checksum , url to set into spokeMachineWorker
+func (r *BMACReconciler) getCheckSumURL(ctx context.Context, log logrus.FieldLogger, spokeClient client.Client) (string, string, error) {
+	var checksum, url string
+	machineList := &machinev1beta1.MachineList{}
+	err := spokeClient.List(ctx, machineList, client.MatchingLabels{MACHINE_TYPE: string(models.HostRoleMaster)})
+	if err != nil {
+		return checksum, url, err
+	}
+	providerSpecValue := string(machineList.Items[0].Spec.ProviderSpec.Value.Raw)
```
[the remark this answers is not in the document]
pawanpinjarkar: @flaper87 TBH, I didn't think why not to use Raw value. I ported the changes from @rwsu original PR here https://github.com/openshift/assisted-service/pull/2143/files#diff-86415614c6ed40c56c5ca2b5f751a65746ef5c83b61e6f021a4f5d817e4425baR1022

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -1133,6 +1162,40 @@ func (r *BMACReconciler) newSpokeMachine(bmh *bmh_v1alpha1.BareMetalHost, cluste
 		},
 	}
 	mutateFn := func() error {
+		if machine.ObjectMeta.Annotations == nil {
+			machine.ObjectMeta.Annotations = make(map[string]string)
+		}
+		machine.ObjectMeta.Annotations[BMH_ANNOTATION] = fmt.Sprintf("%s/%s", OPENSHIFT_MACHINE_API_NAMESPACE, bmh.Name)
+		providerSpecValueFormat := `{
```
flaper87, line 1169: I'm not sure using the values from the master machine is what we want here. I understand these are not being "consumed" but I think the image used for worker nodes is different from the master's and I would rather have this value represent the real state of affairs. Considering that BMAC is the one creating the resource, I think we don't really have an excuse to add the right value here.

flaper87 (approved):
Alright, this may take a bit longer than desired so let's get this PR in and work on adjusting the values for the `providerSpec` in a follow-up PR

PR #2333, opened 2021-08-04
MGMT-7420 - Introducing disk encryption to cluster properties.

ybettan:
# Assisted Pull Request

## Description

OCP allows encrypting the installation disks, therefore,
assisted-service should allow it too.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


ybettan:
Still missing:
* subsystem-tests
* manual system-tests

/assign @avishayt

internal/network/manifests_generator.go
```diff
@@ -238,6 +239,126 @@ func (m *ManifestsGenerator) AddSchedulableMastersManifest(ctx context.Context,
...
+
+	case models.DiskEncryptionEnableOnAll:
+
+		manifestParams["ROLE"] = "master"
+		if err := createDiskEncryptionManifest(manifestParams); err != nil {
+			return err
+		}
+
+		manifestParams["ROLE"] = "worker"
+		if err := createDiskEncryptionManifest(manifestParams); err != nil {
```
eranco74, line 332: Sp all manifest params are the same, just different ROLE.
Seems like this will override the manifest created in line 327.

internal/network/manifests_generator.go
```diff
@@ -238,6 +239,126 @@ func (m *ManifestsGenerator) AddSchedulableMastersManifest(ctx context.Context,
 	return nil
 }
 
+const (
+	diskEncryptionManifest = `apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  name: {{.NAME}}
```
eranco74, line 246: You don't define this key in manifestParams

internal/network/manifests_generator.go
```diff
@@ -238,6 +239,126 @@ func (m *ManifestsGenerator) AddSchedulableMastersManifest(ctx context.Context,
...
+      version: 3.2.0
+    storage:
+      luks:
+        - name: root
+          device: /dev/disk/by-partlabel/root
+          clevis:
+		  {{ if .ENCRYPTION_MODE_TPMV2 }}
+            tpm2: true
+		  {{ end }}
+		  {{ if .ENCRYPTION_MODE_TANG }}
```
eranco74, line 261: Sure this is the right indentation?

internal/network/manifests_generator.go
ybettan: Missed that - good catch.

internal/network/manifests_generator.go
ybettan: I think that github show you a wrong indention - this is not how it is in the code

internal/network/manifests_generator.go
ybettan: `Sp` ?

I will fix it, thanks.

ybettan: Fixed.

internal/network/manifests_generator.go
ybettan: Fixed.

swagger.yaml
```diff
@@ -5193,6 +5203,19 @@ definitions:
       finalizing_stage_percentage:
         type: integer
 
+  #FIXME: add to get/update/create cluster params as well
```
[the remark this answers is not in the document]
ybettan: I don't see any `cluster-update-params`, as long as `disk-encryption` is part of the cluster object it should be fetched be `get` no?!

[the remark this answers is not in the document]
avishayt: Correct - `get` and `list` return `cluster` - you just need to add it there

swagger.yaml
```diff
@@ -5193,6 +5203,19 @@ definitions:
...
+  disk-encryption:
+    type: object
+    properties:
+      enable_on:
+        type: string
+        description: Enable/disable disk encryption on master nodes, worker nodes, or all nodes.
+        enum: ['non', 'all', 'masters', 'workers']
```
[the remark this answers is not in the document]
ybettan: Fixed.

swagger.yaml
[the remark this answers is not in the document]
ybettan: It is already in the PR, closing.

swagger.yaml
```diff
@@ -5193,6 +5203,18 @@ definitions:
...
+    type: object
+    properties:
+      enable_on:
+        type: string
+        description: Enable/disable disk encryption on master nodes, worker nodes, or all nodes.
+        enum: ['none', 'all', 'masters', 'workers']
+      mode:
+        type: string
+        description: The disk encryption mode to use.
+        enum: ['tpmv2', 'tang']
```
lack, line 5216: For tang encryption, a user would need to specify a list of Tang URLs and Thumbprints.

swagger.yaml
```diff
@@ -5193,6 +5203,18 @@ definitions:
       finalizing_stage_percentage:
         type: integer
 
+  disk-encryption:
+    type: object
+    properties:
+      enable_on:
+        type: string
+        description: Enable/disable disk encryption on master nodes, worker nodes, or all nodes.
+        enum: ['none', 'all', 'masters', 'workers']
```
lack, line 5212: This configuration per node pool is nice.

How should this work if there are more disk partitions than just a singular root partition; does this need to have an additional opt-in/opt-out per disk partition?  Or even a unique disk encryption mode per disk partition?

swagger.yaml
ybettan: Seems like we will need additional info in the API as you first suggested after all @avishayt.

I didn't notice that the hard-coded URL is an example one.

swagger.yaml
ybettan: From (openshift docs)[https://docs.openshift.com/container-platform/4.7/installing/install_config/installing-customizing.html#installation-special-config-encrypt-disk-tpm2_installing-customizing]:
```
If you configure additional data partitions, they will not be encrypted unless encryption is explicitly requested
```

I guess that we will need additional objects in `luks` list in that case, we can have a different task for it.
WDYT @avishayt ?

swagger.yaml
avishayt: Good call.  For tang, we should add two parameters: URL and thumbprint.  See the ignition spec [here](https://coreos.github.io/ignition/configuration-v3_2/).  Note that this is unrelated to the "options" that you can also see in the spec, and which we removed for now:

> options (list of strings): any additional options to be passed to the cryptsetup utility.

swagger.yaml
avishayt: I assumed that the configuration in the docs was the suggested one for OpenShift and would encrypt the relevant partitions.  In what case would additional partitions be created?

swagger.yaml
lack: A list of {url, thumbprint} please :)

avishayt: Correct

swagger.yaml
ybettan: @lack? Can you answer @avishayt 's question?

swagger.yaml
ybettan: Fixed.

internal/network/manifests_generator.go
```diff
@@ -238,6 +241,138 @@ func (m *ManifestsGenerator) AddSchedulableMastersManifest(ctx context.Context,
...
+			URL        string
+			Thumbprint string
+		}
+		if err := json.Unmarshal([]byte(c.DiskEncryption.TangServers), &tangServers); err != nil {
+			log.WithError(err).Error("faild to unmarshal tang_server from cluster object")
+			return err
+		}
+		var tangInfoStr string
+		for _, ts := range tangServers {
+			tangInfoStr += fmt.Sprintf("\n%s- url: %s\n%sthumbprint: %s", strings.Repeat(" ", 14), ts.URL, strings.Repeat(" ", 16), ts.Thumbprint)
```
lack, line 321: It's probably cleaner to use a range function in the template?

swagger.yaml
```diff
@@ -5190,6 +5203,22 @@ definitions:
...
+      mode:
+        type: string
+        description: The disk encryption mode to use.
+        enum: ['tpmv2', 'tang']
+      tang_servers:
+        type: string
+        description: JSON-formatted string containing additional information regarding tang's configuration
```
lack, line 5219: Does this follow the ignition {url, thumbprint} convention, or the clevis {url, thp} convention?

Would it be better to just have the spec ask for an array of {url, thumbprint} pairs to make it unambiguous and less prone to typos or json-in-a-string-in-yaml escaping errors?

swagger.yaml
lack: I'm trying to figure out the specific needs of Telco, where there may be a need for a separate partition set up for the exclusive use of the LocalStorageOperator.  I haven't yet gotten a straight answer about whether that's needed or not, though.

But a read of the luks section of the ignition spec makes it clear that if there are any additional partitions besides the root (a configuration that is supported by OpenShift I think), they each need their own luks section too if encryption is desired.

If these multiple partitions are something AssistedService allows users to set up, the encryption should be as flexible as the partition definition.

swagger.yaml
ybettan: > Does this follow the ignition {url, thumbprint} convention, or the clevis {url, thp} convention

It follows the ignition, you can see it here https://coreos.github.io/ignition/configuration-v3_2/

> Would it be better to just have the spec ask for an array of {url, thumbprint} pairs

This will require maintaining 2 different objects in swagger - 1 for the cluster object and one for create/update cluster params.

The reason for that is that we don't know the length of this list, therefore, the ways to store it in DB are by Marshaling it to a json-string or to have a separate table in DB for them.

The second option seems a bit of overkill to me, so I decided to go with the first option as we do for `validations_info`.

If you have a better option I am open for suggestions.
@lack

swagger.yaml
ybettan: > If these multiple partitions are something AssistedService allows users to set up

Do we @avishayt?

Anyway, I think this PR is big enough, we have a different task for disk partitioning https://issues.redhat.com/browse/MGMT-7461.

internal/network/manifests_generator.go
ybettan: I wasn't aware of that option :)
Fixing it now.

swagger.yaml
lack: It definitely meets the requirements for what NBDE needs, and given the marshalling requirement makes sense?  Unless something can let you de-couple the data format here from that implementation detail?  I don't know enough about your architecture to say!  I wasn't sure if the consumers of your API are administrators creating CRs or other non-human actors, so I worry about the potential errors of having to pass structured data via an unstructured string.

Maybe we should just make sure the description is clear about what the format is, since there's are a couple different possibilities of how this data may legitimately be structured?

Clevis:
```
[{"url": "http://host.example.com:7500”, "thp": "xyz"},...]
```

Ignition:
```
[{"url": "http://host.example.com:7500”, "thunbprint": "xyz"},...]
```

Your example of capitalized ignition:
```
[{"URL": "http://host.example.com:7500”, "Thumbprint": "xyz"},...]
```

ybettan:
/hold
Waiting for test-infra results using `tpmv2` and `tang` modes.

ybettan:
PR is currently blocked on https://bugzilla.redhat.com/show_bug.cgi?id=1994051.

swagger.yaml
ybettan: This was just an example a got when playing with the template, in the code they are called `url` and `thumbprint`

ybettan: Fixed.

internal/network/manifests_generator.go
```diff
@@ -238,6 +240,138 @@ func (m *ManifestsGenerator) AddSchedulableMastersManifest(ctx context.Context,
...
+			log.WithError(err).Error("faild to unmarshal tang_server from cluster object")
+			return err
+		}
+
+		manifestParams["MODE"] = "tang"
+		manifestParams["TANG_SERVERS"] = tangServers
+
+	default:
+
+		return errors.New("Non of the encryption modes were specified")
```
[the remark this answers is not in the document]
ybettan: Thanks, fixed.

ybettan:
Actually, https://github.com/openshift/assisted-service/pull/2333#issuecomment-899733747 is not the blocker, by default our VMs don't enable `tpm` in BIOS, working on enabling it in order to retest.

ybettan:
/unhold
System tests have been too time-consuming, delegating this responsibility to QE.
@gamli75 @lalon4 @iovadia

internal/network/manifests_generator.go
```diff
@@ -238,6 +240,135 @@ func (m *ManifestsGenerator) AddSchedulableMastersManifest(ctx context.Context,
...
+
+		manifestParams["MODE"] = "tang"
+		manifestParams["TANG_SERVERS"] = tangServers
+
+	default:
+
+		return errors.New("None of the encryption modes were specified")
+	}
+
+	switch c.DiskEncryption.EnableOn {
```
eranco74, line 328: This should be an enum.

internal/network/manifests_generator.go
```diff
@@ -238,6 +240,135 @@ func (m *ManifestsGenerator) AddSchedulableMastersManifest(ctx context.Context,
...
+
+	case models.DiskEncryptionEnableOnAll:
+
+		manifestParams["ROLE"] = "master"
+		filename := fmt.Sprintf("99-openshift-%s-%s-encryption.yaml", manifestParams["ROLE"], manifestParams["MODE"])
+		if err := m.createDiskEncryptionManifest(ctx, log, c, manifestParams, filename); err != nil {
```
eranco74, line 334: Consider passing the role instead of the filename to `m.createDiskEncryptionManifest` and move these 2 lines to that function:
manifestParams["ROLE"] =
filename :=

e.g.
if err := m.createDiskEncryptionManifest(ctx, log, c, manifestParams, "master"); err != nil {

internal/network/manifests_generator.go
ybettan: Already enum.

internal/network/manifests_generator.go
ybettan: Fixed.

PR #2286, opened 2021-07-27
OCPBUGSM-32293: Cache only required secrets

rollandf:
# Assisted Pull Request

## Description

The internal cache now filters out secrets without the label "agentclusterinstalls.extensions.hive.openshift.io":"true".
The ClusterDeployment controller takes care of applying such label to the CD Pull secrets only.
Signed-off-by: Fred Rolland <frolland@redhat.com>

## List all the issues related to this PR

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

Tested with an env with 20k Secrets. ( [creation script](https://gist.github.com/rollandf/3670353cf1b8e379f3cd4bcc2f3dc48a) )

Without this change:
On start of Service:
```
# kubectl top pods -n assisted-installer 
NAME                                CPU(cores)   MEMORY(bytes)   
assisted-service-64b956dfcf-d547r   3m           1148Mi          
object-store-558498d8-fvbkp         1m           71Mi            
postgres-664877c76f-hf2c6           2m           37Mi            
wiremock-7c674958d7-wdcc8           1m           67Mi 
```  
After some time:
```
# kubectl top pods -n assisted-installer
NAME                                CPU(cores)   MEMORY(bytes)   
assisted-service-64b956dfcf-d547r   4m           688Mi           
object-store-558498d8-fvbkp         1m           71Mi            
postgres-664877c76f-hf2c6           3m           38Mi            
wiremock-7c674958d7-wdcc8           1m           67Mi 
```

With this change:
On start of Service:
```
# kubectl top pods -n assisted-installer
NAME                                CPU(cores)   MEMORY(bytes)
assisted-service-59c55b6b95-hfmbm   4m           62Mi
object-store-558498d8-fvbkp         1m           67Mi
postgres-664877c76f-hf2c6           2m           38Mi
wiremock-7c674958d7-wdcc8           1m           67Mi
```

After some time:
```
# kubectl top pods -n assisted-installer
NAME                                CPU(cores)   MEMORY(bytes)   
[...]

PR #2335, opened 2021-08-04
OCPBUGSM-32941: Cache secrets for spoke cluster

pawanpinjarkar:
# Assisted Pull Request

## Description
The internal cache now filters out secrets without the AI custom label "agent-install.openshift.io/watch": "true". Secrets are retrieved using the non-caching client if it fails to find the Secret in the current cache.

This is a combined fix for [OCPBUGSM-32941](https://issues.redhat.com/browse/OCPBUGSM-32941) and [OCPBUGSM-32966](https://issues.redhat.com/browse/OCPBUGSM-32966). 

Without this fix, the BMAC and ClusterDeployment controllers can't find the required secrets even if they are actually present.

Ref: 
https://issues.redhat.com/browse/OCPBUGSM-32941
https://issues.redhat.com/browse/OCPBUGSM-32966

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
- [] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -505,9 +505,10 @@ func main() {
 			}).SetupWithManager(ctrlMgr), "unable to create controller Agent")
 
 			failOnError((&controllers.BMACReconciler{
-				Client: ctrlMgr.GetClient(),
-				Log:    log,
-				Scheme: ctrlMgr.GetScheme(),
+				Client:    ctrlMgr.GetClient(),
+				APIReader: ctrlMgr.GetAPIReader(),
+				Log:       log,
+				Scheme:    ctrlMgr.GetScheme(),
 			}).SetupWithManager(ctrlMgr), "unable to create controller BMH")
 
 			failOnError((&controllers.AgentClusterInstallReconciler{
@@ -753,7 +754,11 @@ func createControllerManager() (manager.Manager, error) {
 			NewCache: cache.BuilderWithOptions(cache.Options{
 				SelectorsByObject: cache.SelectorsByObject{
 					&corev1.Secret{}: {
-						Label: labels.SelectorFromSet(labels.Set{controllers.SecretLabelName: controllers.SecretLabelValue}),
+						Label: labels.SelectorFromSet(
+							labels.Set{
+								controllers.WatchResourceLabel: controllers.WatchResourceValue,
+							},
+						),
 					},
 				},
 			}),
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -54,6 +54,7 @@ import (
 // BMACReconciler reconciles a Agent object
 type BMACReconciler struct {
 	client.Client
+	APIReader   client.Reader
 	Log         logrus.FieldLogger
 	Scheme      *runtime.Scheme
 	spokeClient client.Client
@@ -674,11 +675,13 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 	if !installed {
 		return reconcileComplete{}
 	}
-	secret, ok, err := r.getSecret(ctx, log, agent, cd)
+
+	key := types.NamespacedName{
+		Namespace: agent.Spec.ClusterDeploymentName.Namespace,
+		Name:      fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name),
+	}
+	secret, err := getSecret(ctx, r.Client, r.APIReader, key)
 	if err != nil {
-		if !ok {
-			return reconcileComplete{}
-		}
 		return reconcileError{err}
 	}
 
@@ -694,7 +697,11 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		return reconcileError{err}
 	}
 
-	_, err = r.ensureSpokeBMHSecret(ctx, log, spokeClient, bmh)
+	key = types.NamespacedName{
+		Namespace: bmh.Namespace,
+		Name:      bmh.Spec.BMC.CredentialsName,
+	}
+	_, err = r.ensureSpokeBMHSecret(ctx, log, spokeClient, key)
 	if err != nil {
 		log.WithError(err).Errorf("failed to create or update spoke BareMetalHost Secret")
 		return reconcileError{err}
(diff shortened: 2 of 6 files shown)


flaper87:
Another option to consider, which came from a chat convo with @rollandf , is to use a separate kubeclient instance for these specific calls. This would prevent us from adding secrets to the cache (which we don't really need to watch), and it would also improve the situation as we wouldn't have to keep a list of labels.

cmd/main.go
```diff
@@ -753,7 +753,18 @@ func createControllerManager() (manager.Manager, error) {
...
+						Label: labels.SelectorFromSet(
+							labels.Set{
+								controllers.SecretLabelName: controllers.SecretLabelValue,
+							},
+						),
+					},
+					&corev1.Secret{}: {
+						Label: labels.SelectorFromSet(
+							labels.Set{
+								"hive.openshift.io/secret-type": "kubeconfig",
```
flaper87, line 765: What do you think we construct the label/value using [Hive's constants](https://github.com/openshift/hive/blob/5990617a5f027cc834f1b14d6a0ece0c7dc35696/pkg/installmanager/installmanager.go#L1146)? 

This should help us stay closer to what they are doing if any of these values are changed in the future.

pawanpinjarkar: Sure, will update.

flaper87 (changes requested):
Summarizing a longer conversation that happened offline:

* Use the r.Get calls as they are now.
  * If a secret is not found, use the non-caching client to fetch the secret, add the label, and save it.

All the primitives to do this should be there.

In addition, this sounds like a good case to move from using Hive's label/value to something custom for bmac/AI (edited)

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -74,6 +75,8 @@ const (
 	MACHINE_TYPE                        = "machine.openshift.io/cluster-api-machine-type"
 	MCS_CERT_NAME                       = "ca.crt"
 	OPENSHIFT_MACHINE_API_NAMESPACE     = "openshift-machine-api"
+	CustomSecretLabelName               = "some-custom-label-for-assisted-service-secrets"
+	CustomSecretLabelValue              = "some-value-for-custom-label-of-assisted-service-secrets"
```
pawanpinjarkar, line 79: Please suggest better names here.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -1049,10 +1068,9 @@ func (r *BMACReconciler) createIgnitionWithMCSCert(ctx context.Context, log logr
 
 }
 
-func (r *BMACReconciler) ensureSpokeBMHSecret(ctx context.Context, log logrus.FieldLogger, spokeClient client.Client, bmh *bmh_v1alpha1.BareMetalHost) (*corev1.Secret, error) {
-	secretName := bmh.Spec.BMC.CredentialsName
+func (r *BMACReconciler) ensureSpokeBMHSecret(ctx context.Context, log logrus.FieldLogger, spokeClient client.Client, key types.NamespacedName) (*corev1.Secret, error) {
 	secret := &corev1.Secret{}
-	err := r.Get(ctx, types.NamespacedName{Namespace: bmh.Namespace, Name: secretName}, secret)
+	err := r.Get(ctx, types.NamespacedName{Namespace: key.Namespace, Name: key.Name}, secret)
```
flaper87, line 1073:
```suggestion
	err := r.Get(ctx, key, secret)
```

pawanpinjarkar:
I also updated/rearranged some code originally added by @frolland for his PR https://github.com/openshift/assisted-service/pull/2286 so that it can be reused and we can minimize the code duplication.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -694,7 +694,18 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
...
+	key := types.NamespacedName{
+		Namespace: bmh.Namespace,
+		Name:      bmh.Spec.BMC.CredentialsName,
+	}
+	// Make sure that the BMC Credentials Secret has the required label
+	err = ensureSecretLabel(ctx, key.Name, key.Namespace, CustomSecretLabelName, CustomSecretLabelValue, r.Client, r.APIReader)
+	if err != nil {
+		log.WithError(err).Error("error setting label on BMC Credentials Secret")
+		return reconcileError{err}
+	}
```
flaper87, line 706: This will query the secret (using the non-cached client) on every reconcile. An alternative implementation would be to modify the `ensureSpokeBMHSecret` function by catching the `NotFound` error and calling `ensureSecretLabel` *only* in that case.

The `ensureSecretLabel` function could return the secret so that we don't have to fetch it twice.

Let me know if this makes sense.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -74,6 +75,8 @@ const (
 	MACHINE_TYPE                        = "machine.openshift.io/cluster-api-machine-type"
 	MCS_CERT_NAME                       = "ca.crt"
 	OPENSHIFT_MACHINE_API_NAMESPACE     = "openshift-machine-api"
+	CustomSecretLabelName               = "some-custom-label-for-assisted-service-secrets"
```
flaper87, line 78:
```suggestion
	WatchResourceLabel               = "agent-install.openshift.io/watch"
```

We could move this to `common.go` and use it for `ClusterDeploymentController`

internal/controller/controllers/bmh_agent_controller.go
flaper87:
```suggestion
	WatchResourceValue              = "true"
```

flaper87 (changes requested):
Some initial comments, doing another pass now.

internal/controller/controllers/bmh_agent_controller.go
mhrivnak: Yes, let's please do this. We should only use the non-caching client if we fail to find the Secret in the current cache.

pawanpinjarkar: Sure

internal/controller/controllers/common.go
```diff
@@ -34,22 +34,33 @@ const (
...
+	WatchResourceLabel               = "agent-install.openshift.io/watch"
+	WatchResourceValue               = "true"
 )
 
-func getSecret(ctx context.Context, c client.Client, r client.Reader, namespace, name string) (*corev1.Secret, error) {
-	secret := &corev1.Secret{}
-	key := types.NamespacedName{
-		Namespace: namespace,
-		Name:      name,
+func ensureSecretLabel(ctx context.Context, c client.Client, r client.Reader, key types.NamespacedName) error {
```
mhrivnak, line 41: Since labeling is just an implementation detail of how we are getting the secret (with a selective cache), WDYT about moving this entirely within the below `getSecret` function? That way any code in this package can just call `getSecret` and not worry about the cache or label stuff. Within `getSecret`:

* try to get from cache
* if not found, get from API and add the label

pawanpinjarkar: yes, sounds better. Fixed.

flaper87:
LGTM!

/cc @rollandf for a sanity check

rollandf:
Looks fine, thanks! LGTM

flaper87 (approved):

PR #2344, opened 2021-08-05
Adapt service to use only V2 APIs for agent Register, GetNextSteps, PostStepReply

ori-amizur:
# Assisted Pull Request

## Description
[MGMT-7413](https://issues.redhat.com/browse/MGMT-7413): Change post step reply tools to use V2 structure
- Make post step reply to use only V2 structures
[MGMT-7462](https://issues.redhat.com/browse/MGMT-7462): Remove V1 agent APIs from service
- RegisterHost, GetNextSteps, PostStepReply

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @ravidbro 
/cc @yevgeny-shnaidman 

## Checklist

## Reviewers Checklist


ravidbro:
General question: don't we need to support a case where the service is upgraded but the agents are still old and calling v1 APIs?

flaper87:
> General question: don't we need to support a case where the service is upgraded but the agents are still old and calling v1 APIs?

For the operator-based workflow, I think the requirement is for there not to be any ongoing installation during an upgrade. For the cloud workflow, I don't think this is true. It is possible to communicate a maintenance window.

That said, I think I would rather keep backward compatibility and make sure existing agents won't break.

There are other, a bit more complex, solutions that would provide a seamless experience. It may not be worth digging too much into those as keeping backward compatibility or (explicitly communicating the incompatibility to users) should be enough.

yevgeny-shnaidman:
> General question: don't we need to support a case where the service is upgraded but the agents are still old and calling v1 APIs?

It was agreed that in case there are still active agents while the assisted service is upgraded to the new version, we/user will need to reboot those agent.

PR #2346, opened 2021-08-05
MGMT-7089: Add binding host API

ravidbro:
# Assisted Pull Request

## Description

Add API for Binding host to a cluster.
At this stage, the host was already created with InfraEnv and isn't attached yet to any cluster.
The API will use the state machine to change the state and update cluster_id so the agent will register itself to the cluster

- [X] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] Unit Tests
- [ ] No tests needed

## Assignees

/assign @yevgeny-shnaidman 

/cc @ori-amizur 
/cc @machacekondra 
/cc @filanov 

## Checklist

## Reviewers Checklist


swagger.yaml
```diff
@@ -3754,6 +3754,70 @@ paths:
...
+      tags:
+        - installer
+      security:
+        - agentAuth: []
+      description: Bind host to a cluster
+      operationId: BindHost
+      parameters:
+        - in: path
+          name: infra_env_id
+          description: The infra env of the host that is posting results.
```
avishayt, line 3768:
```suggestion
          description: The infra env of the host that is being bound.
```

swagger.yaml
```diff
@@ -3754,6 +3754,70 @@ paths:
...
+          type: string
+          format: uuid
+          required: true
+        - in: path
+          name: host_id
+          description: The host that is posting results.
```
avishayt, line 3774:
```suggestion
          description: The host that is being bound.
```

swagger.yaml
```diff
@@ -3754,6 +3754,70 @@ paths:
...
+          type: string
+          format: uuid
+          required: true
+        - name: bind-host-params
+          description: The params for binding host (e.g. cluster).
```
avishayt, line 3779:
```suggestion
          description: The parameters for the host binding.
```

internal/bminventory/inventory.go
```diff
@@ -5551,3 +5553,39 @@ func (b *bareMetalInventory) V2PostStepReply(ctx context.Context, v2Params insta
 func (b *bareMetalInventory) V2GetHost(ctx context.Context, params installer.V2GetHostParams) middleware.Responder {
 	return installer.NewV2GetHostNotImplemented()
 }
+
+func (b *bareMetalInventory) BindHost(ctx context.Context, params installer.BindHostParams) middleware.Responder {
+	log := logutil.FromContext(ctx, b.log)
+	log.Infof("Bind host %s to cluster %s", params.HostID, params.BindHostParams.ClusterID)
```
avishayt, line 5559:
```suggestion
	log.Infof("Binding host %s to cluster %s", params.HostID, params.BindHostParams.ClusterID)
```

internal/bminventory/inventory.go
```diff
@@ -5551,3 +5553,39 @@ func (b *bareMetalInventory) V2PostStepReply(ctx context.Context, v2Params insta
...
+	host, err := common.GetHostFromDB(b.db, params.InfraEnvID.String(), params.HostID.String())
+	if err != nil {
+		log.WithError(err).Errorf("failed to find host <%s> in infraEnv <%s>",
+			params.HostID, params.InfraEnvID)
+		return common.NewApiError(http.StatusNotFound, err)
+	}
+	if host.ClusterID != nil {
```
eranco74, line 5566: Is this even possible? a host with InfraEnvId and ClusterId?

internal/bminventory/inventory.go
```diff
@@ -5551,3 +5553,39 @@ func (b *bareMetalInventory) V2PostStepReply(ctx context.Context, v2Params insta
...
+	if err != nil {
+		return common.NewApiError(http.StatusBadRequest, errors.Errorf("Failed to find cluster %s", params.BindHostParams.ClusterID))
+	}
+	if err = b.clusterApi.AcceptRegistration(cluster); err != nil {
+		log.WithError(err).Errorf("failed to bind host <%s> to cluster %s due to: %s",
+			params.HostID.String(), *params.BindHostParams.ClusterID, err.Error())
+		return common.NewApiError(http.StatusConflict, err)
+	}
+
+	if err = b.hostApi.BindHost(ctx, &host.Host, *params.BindHostParams.ClusterID, b.db); err != nil {
```
eranco74, line 5579: Where are we removing the InfraEnvId?
Seems that `PostBindHost` only adds the ClusterId
 
I'm asking because TTBOMK `if c.infraEnv != nil ` is used for skipping validation while the host isn't bound to a cluster.

internal/bminventory/inventory.go
```diff
@@ -5551,3 +5553,39 @@ func (b *bareMetalInventory) V2PostStepReply(ctx context.Context, v2Params insta
...
+		log.WithError(err).Errorf("Failed to bind host <%s> to cluster <%s>",
+			params.HostID, *params.BindHostParams.ClusterID)
+		return common.NewApiError(http.StatusInternalServerError, err)
+	}
+
+	host, err = common.GetHostFromDB(b.db, params.InfraEnvID.String(), params.HostID.String())
```
eranco74, line 5585: Why is this required?
You got the host from DB, the `b.hostApi.BindHost` updated it, I think you can just return it.

internal/bminventory/inventory.go
eranco74: If it's possible, consider adding: 
```suggestion
	if host.ClusterID != nil {
	    if host.ClusterID == *params.BindHostParams.ClusterID {
	        return 204
	    }
```

ravidbro: It is possible to have both, InfraEnvID always exists since it's part of the identifier of the host (PK), if the host is currently part of a cluster then ClusterID will also have values.
Since this is an imperative action, I thought that it will be right to return an error if it's already bound and the user asks to bind even if the request is to bind to the cluster it's currently bounded to.
Anyhow, we defined that this API return Host object then I don't see a reason in some cases to return 204

internal/bminventory/inventory.go
ravidbro: There will be a dedicated API for Unbind action that will clear that field

ravidbro: don't confuse between the fields on the validation context (where looking fo c.infraEnv != nil) to the fields on the host object that is used here

internal/bminventory/inventory.go
ravidbro: I'm not sure how the state machine works, are you sure that it updated the object I sent to it?
@filanov can you confirm that?

internal/bminventory/inventory.go
avishayt: 1. @eranco74 , I think you may have gotten confused by your PR, where the unit tests checked if infraenv was nil.  That's just confusing test code and there is a task to fix it.  A host always belongs to the InfraEnv whose ISO it booted from.  It can also optionally be bound to a cluster.
2. I think the original code was good as is.

internal/bminventory/inventory.go
yevgeny-shnaidman: state machine updates the host in DB, and then updates the pointer to the host in stateHost, and not the host pointed by stateHost, so we need to read from DB if we want to get the right cluster id in the returned host

internal/bminventory/inventory.go
```diff
@@ -5413,11 +5413,13 @@ func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installe
 	}
 
 	if err = b.customizeHost(host); err != nil {
+		// TODO Need event for infra-env instead of cluster
```
filanov, line 5416: please add ticket number to those comments `MGMT-7322` will be easier to find them later.

internal/bminventory/inventory.go
yevgeny-shnaidman: /lgtm

ori-amizur (commented):
Should there be timeout to the period that host is in binding status?

ravidbro:
> Should there be timeout to the period that host is in binding status?

I don't think so

ori-amizur:
> > Should there be timeout to the period that host is in binding status?
> 
> I don't think so

So host can be in binding status indefinitely? Seems we expect host to register here.

ravidbro:
> > > Should there be timeout to the period that host is in binding status?
> > 
> > 
> > I don't think so
> 
> So host can be in binding status indefinitely? Seems we expect host to register here.

We are expecting, but if the agent won't register it will stay 'binding'

ori-amizur:
> > > > Should there be timeout to the period that host is in binding status?
> > > 
> > > 
> > > I don't think so
> > 
> > 
> > So host can be in binding status indefinitely? Seems we expect host to register here.
> 
> We are expecting, but if the agent won't register it will stay 'binding'

Which shouldn't happen.  If agent didn't register then something went wrong

ravidbro:
had to rebase, please LGTM

PR #2367, opened 2021-08-10
MGMT-7485: move infra-env generation for V1 to cluster generation

yevgeny-shnaidman:
Currently we create infra-env for cluster during GenerateClusterISO API.
This allows to check the behaviour of the system when infra-env is created asynchronously
from cluster creation. Since now we have 2 more additional APIs that rely on infra-env
existence for V1 (discovery-ignition and ramdisk query), we move it back to be created/deleted
whenever its cluster is created/deleted

# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


internal/cluster/registrar.go
```diff
@@ -69,27 +70,23 @@ func (r *registrar) registerCluster(ctx context.Context, cluster *common.Cluster
...
+	if err != nil {
+		r.log.WithError(err).Errorf("Failed to create Infra Env along the cluster %s", cluster.ID)
 		return err
 	}
 
-	success = true
-	return nil
-}
-
-func (r *registrar) RegisterAddHostsOCPCluster(c *common.Cluster, db *gorm.DB) error {
```
carbonin, line 85: Still not clear why this was removed.

internal/cluster/registrar.go
```diff
@@ -113,11 +110,21 @@ func (r *registrar) DeregisterCluster(ctx context.Context, cluster *common.Clust
 		return errors.Errorf("failed to deregister host while unregistering cluster %s", cluster.ID)
 	}
 
+	fmt.Printf("YEV - deleting cluster %s", cluster.ID)
```
carbonin, line 113: Please remove debug prints.

internal/common/db.go
```diff
@@ -288,3 +290,39 @@ func ToSqlList(strs []string) string {
 	res = fmt.Sprintf("('%s')", res)
 	return res
 }
+
+func CreateTransactionWithDefer(db *gorm.DB, txSuccess *bool) (func(), *gorm.DB, error) {
```
carbonin, line 294: Why is this created here? I don't see it being used in this PR.

internal/common/db.go
```diff
@@ -288,3 +290,39 @@ func ToSqlList(strs []string) string {
...
+			tx.Rollback()
+		}
+		if r := recover(); r != nil {
+			tx.Rollback()
+		}
+	}
+	return deferFunc, tx, nil
+}
+
+func CreateInfraEnvForCluster(db *gorm.DB, cluster *Cluster) error {
```
carbonin, line 311: What about things like static network config, ignition overrides, and ntp sources?

internal/cluster/registrar.go
```diff
@@ -69,27 +70,23 @@ func (r *registrar) registerCluster(ctx context.Context, cluster *common.Cluster
...
 	}
 
+	fmt.Printf("YEV - creating cluster %s\n", cluster.ID)
 	if err := tx.Create(cluster).Error; err != nil {
 		r.log.Errorf("Error registering cluster %s", cluster.Name)
 		return err
 	}
 
-	if err := tx.Commit().Error; err != nil {
+	err := common.CreateInfraEnvForCluster(tx, cluster)
```
filanov, line 79: Why not create it in the inventory.go level? We will probably use the same code for v2 no? so why not create it with internal function in inventory.go and with v1 you will create it and with v2 you will not.

internal/bminventory/inventory.go
```diff
@@ -5326,7 +5310,6 @@ func (b *bareMetalInventory) V2UpdateHostInstallProgress(ctx context.Context, pa
 		return common.NewApiError(http.StatusInternalServerError, err)
 	}
 
-	//FIXME: do we need a transaction here? I am guessing that if the call failed the host will retry anyway no?!
```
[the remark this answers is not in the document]
yevgeny-shnaidman: yes, by @ybettan , he asked me to removed this commeny

internal/cluster/registrar.go
```diff
@@ -69,27 +70,23 @@ func (r *registrar) registerCluster(ctx context.Context, cluster *common.Cluster
 		tx = common.LoadTableFromDB(tx, tableName)
 	}
 
+	fmt.Printf("YEV - creating cluster %s\n", cluster.ID)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/cluster/registrar.go
yevgeny-shnaidman: see my comment above

internal/cluster/registrar.go
yevgeny-shnaidman: fixed

internal/common/db.go
yevgeny-shnaidman: we have a lot of repeating code for transaction initialization. This function unifies this code. I push it here, so that we can replace all those occurrences in a later PR

internal/common/db.go
yevgeny-shnaidman: static network is passed as a parameter only during GenerateISO request, and we are setting it there into the InfraEnv. 
ntp is not used as part of Discovery generation at all.
ignition overrides in V1 is part of cluster update. we are updating Infraenv with relevant value before calling FormatDiscoveryIgnitionFile. In V2, it will be updating on the InfraEnv directly in a dedicated API

internal/cluster/registrar.go
yevgeny-shnaidman: if we create it in inventory.go, then we must define a transaction at that level and pass it down via cluster API. I prefer to create it directly during registrar, without changing inventory code. In V2 InfraEnv for cluster will be created by a InfraEnv API, and not as part of Cluster creation.

yevgeny-shnaidman: in addition , using registrar also covers the cases of Day2 cluster (at least until @nmagnezi  delivers) and covers the unregister of the infra-env

filanov: What is the problem creating the transaction in inventory.go and how will you handle v2 cluster create?

yevgeny-shnaidman: as agreed, leaving the creation/destruction of the InfraEnv in the registar, passing a flag that defines it is a v1/v2 implementation (create or not infra env)
This way we have only one place where we create/delete infra-env for V1, and it will take care of day1 cluster, day2 cluster and cluster garbage collector

internal/bminventory/inventory.go
```diff
@@ -250,11 +250,27 @@ func (b *bareMetalInventory) updatePullSecret(pullSecret string, log logrus.Fiel
...
+	}
+
 	infraEnv, err := common.GetInfraEnvFromDB(b.db, params.ClusterID)
 	if err != nil {
+		log.WithError(err).Errorf("failed to get infra-env: %s", params.ClusterID)
 		return common.GenerateErrorResponder(err)
 	}
 
+	// update pull secret and proxy, since they could have been set during cluster update
+	proxy := models.Proxy{
```
ori-amizur, line 266: This means that we ignore these parameters in the infraEnv and take it from the cluster even for v2.  Is this the intention ?

rollandf:
looks good

internal/cluster/cluster.go
```diff
@@ -200,10 +198,6 @@ func (m *Manager) RegisterAddHostsCluster(ctx context.Context, c *common.Cluster
 	return err
 }
 
-func (m *Manager) RegisterAddHostsOCPCluster(c *common.Cluster, db *gorm.DB) error {
```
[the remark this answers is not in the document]
carbonin: How does that have anything to do with changing where we create the infraenv?

internal/cluster/registrar.go
carbonin: Same here, we should limit changes in a PR to a single topic.

carbonin:
Also, please don't ignore the PR template.

internal/cluster/cluster.go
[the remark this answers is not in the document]
yevgeny-shnaidman: just "going along for the ride", since if we make it a task, i have no idea when it will be pushed

internal/bminventory/inventory.go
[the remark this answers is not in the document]
yevgeny-shnaidman: When adding V2UpdateHostInstallProgress, we copied part of the code from the UpdateHostInstallProgress, and so we copied this comment also. It was left there by mistake by @ybettan (as a private comment to himself, and he forgot to remove before push). He removed it in 3889757c6131e30096f5f178238253dce5cf3f44

internal/cluster/cluster.go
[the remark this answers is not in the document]
carbonin: That's not a good reason to lump it in with unrelated changes.
If it's tech debt you can make a `NO-ISSUE` PR and describe why it's no longer needed in the commit message.

internal/common/db.go
yevgeny-shnaidman: removed

internal/common/db.go
yevgeny-shnaidman: not sure i understand your comment:  where am i changing them in this PR?

carbonin: I'm saying that we'll need a way to set static networking that isn't generating the image when generating the image outside of assisted service.

But that doesn't need to be handled in this PR. I'll resolve this comment.

internal/cluster/registrar.go
yevgeny-shnaidman: fixed

internal/cluster/cluster.go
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/bminventory/inventory.go
```diff
@@ -250,11 +250,27 @@ func (b *bareMetalInventory) updatePullSecret(pullSecret string, log logrus.Fiel
...
+		return common.GenerateErrorResponder(err)
+	}
+
 	infraEnv, err := common.GetInfraEnvFromDB(b.db, params.ClusterID)
 	if err != nil {
+		log.WithError(err).Errorf("failed to get infra-env: %s", params.ClusterID)
 		return common.GenerateErrorResponder(err)
 	}
 
+	// update pull secret and proxy, since they could have been set during cluster update
```
carbonin, line 265: Would it not make more sense to set these in the infra env when they are set in the cluster?
Then we wouldn't have to worry about doing this all over again in the next endpoint that needs these values.

internal/bminventory/inventory.go
```diff
@@ -1020,6 +1021,7 @@ func (b *bareMetalInventory) GenerateClusterISOInternal(ctx context.Context, par
 	}
 
 	updates := map[string]interface{}{}
+	updates["pull_secret"] = cluster.PullSecret
```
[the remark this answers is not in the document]
carbonin: Similar comment as I had previously. If we're creating the infraenv when we create the cluster shouldn't we keep this up to date as it changes in the cluster?

Then we wouldn't have to duplicate this in https://github.com/openshift/assisted-service/pull/2367/files#diff-747a429a0e12307ef65d754c83b6898ce3858b621005e1e7af6e420da1f1657cR271

internal/bminventory/inventory.go
yevgeny-shnaidman: We are in the middle of moving to V2. any new endpoints will use V2 infra env, which will be updated using dedicated API, without going through cluster. We can start adding code to ClusterUpdate API, but that function is confusing, and we are going to dissemble it anyway shortly

internal/bminventory/inventory.go
```diff
@@ -250,11 +250,27 @@ func (b *bareMetalInventory) updatePullSecret(pullSecret string, log logrus.Fiel
...
 	}
 
+	// update pull secret and proxy, since they could have been set during cluster update
+	proxy := models.Proxy{
+		HTTPProxy:  swag.String(cluster.HTTPProxy),
+		HTTPSProxy: swag.String(cluster.HTTPSProxy),
+		NoProxy:    swag.String(cluster.NoProxy),
+	}
+	infraEnv.PullSecret = cluster.PullSecret
+	infraEnv.Proxy = &proxy
```
carbonin, line 272: Do we also need to handle ignition config override and static networking here?
Presumably these could also have been updated since the cluster was created.

carbonin: I guess static networking is less of a concern because it's only set in the generate endpoint and it looks like we're updating the infra env there, but I think this still applies for the override because that's a separate endpoint.

yevgeny-shnaidman: Good catch. Fixed the ignition config overide. Static Networking is only passed during Generate Cluster ISO call, and infra env is updated with it

carbonin (approved):
Thanks for making all those changes @yevgeny-shnaidman 

If we're going to need to maintain this longer term then I would really like to keep the infraenv as up to date as possible by reflecting changes from the cluster in the actual cluster update endpoints, but if this works for now and can get us fully to v2 faster then that works too.

PR #2369, opened 2021-08-10
MGMT-7444: Support OS images listing in OS_IMAGES variable

danielerez:
# Assisted Pull Request

## Description

As a preparation for multiple CPU architecture support, extracting the RHCOS images information from OPENSHIFT_VERSIONS to simplify version object, see [arm64 support enhancement](https://github.com/openshift/assisted-service/blob/d5e953b1e8d68fa3470c433fc8141d882bc3e01e/docs/enhancements/arm64-cpu-architecture.md#ocp-versions).

This PR is introducing support to Versions API for fetching OS images info from OS_IMAGES instead of OPENSHIFT_VERSIONS. The implementation ensures backwards compatibility by fallback to the current fields in OpenshiftVersion if RhcosVersions is missing. 
* Swagger:
  * Added an os-image object to swagger represent information on an OS image.
  * The rhcos-versions is an array for aligning with [k8s API convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps) and easier handling in AgentServiceConfig.
* Versions:
  * Added support for fetching OS info from a specified OsImages array.
  * Added GetOsImage func for retrieving the OsImage struct.
  * Field validations is done once when invoking NewHandler, instead of in each Getter.
* Added default_os_images.json file
  * Contains rhcos images info for the current versions.
  * default_ocp_versions.json remains unchanged for now until integrating with relevant projects. 
* Added OS_IMAGES to Makefile and generate.
  * For adding the (currently) optional env var to the configuration files.
* TODO: Handle test-infra/operator/assisted-image-service/etc
  * See issues in epic.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @YuviGold 
/cc @osherdp 
/cc @filanov 
/cc @flaper87 
/cc @avishayt 
/cc @mhrivnak 

## Checklist

## Reviewers Checklist


internal/versions/versions.go
```diff
@@ -284,3 +264,44 @@ func (h *handler) AddOpenshiftVersion(ocpReleaseImage, pullSecret string) (*mode
 
 	return openshiftVersion, nil
 }
+
+// Returns a string value by a specified field name
+func (h *handler) getRhcosField(openshiftVersion, fieldName string) (string, error) {
```
mhrivnak, line 269: Is there another way we could parse the data without subverting go's strong typing? And especially without using `unsafe`? What's motivating this approach vs. something like unmarshalling json into a struct?

models/rhcos_version.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/go-openapi/errors"
+	"github.com/go-openapi/strfmt"
+	"github.com/go-openapi/swag"
+	"github.com/go-openapi/validate"
+)
+
+// RhcosVersion rhcos version
+//
+// swagger:model rhcos-version
+type RhcosVersion struct {
```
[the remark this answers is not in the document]
danielerez: Yeah, makes sense, will be more consistent.

internal/versions/versions.go
danielerez: That means that any validation currently done in these getters will be done by the accessing code. So it won't be any cleaner than the current situation.
I've pushed a new version without using 'unsafe'. So maybe we could keep the reflection:/

mhrivnak: What validation are you referring to? The data doesn't change at runtime, so presumably it can be validated once at startup.

danielerez: There are runtime validations, like ensuring that the specified version exists when creating a cluster.

mhrivnak: Ok, `GetVersion` can either return a models.OpenshiftVersion or an error if one doesn't exist. We're just checking if something exists in a map, right? Seems like standard stuff. Any other validations?

data/default_rhcos_versions.json
```diff
@@ -0,0 +1,23 @@
...
+        "cpu_architecture": "x86_64",
+        "rhcos_image": "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.6/4.6.8/rhcos-4.6.8-x86_64-live.x86_64.iso",
+        "rhcos_rootfs": "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.6/4.6.8/rhcos-live-rootfs.x86_64.img",
+        "rhcos_version": "46.82.202012051820-0"
+    },
+    {
+        "openshift_version": "4.7",
+        "cpu_architecture": "x86_64",
+        "rhcos_image": "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.7/4.7.13/rhcos-4.7.13-x86_64-live.x86_64.iso",
+        "rhcos_rootfs": "https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.8/4.8.2/rhcos-live-rootfs.x86_64.img",
```
YuviGold, line 13: This should be 4.7.13

danielerez: good catch:) fixed

internal/versions/versions.go
danielerez: Ok, so added a GetOsImage funk for fetching the entire struct, and the rest of the validation is done when creating a new Versions handler. We could do the same for other funcs in the API later, to reduce duplication.

internal/versions/versions_test.go
```diff
@@ -40,6 +40,24 @@ var defaultOpenShiftVersions = models.OpenshiftVersions{
 	},
 }
 
+var newOpenShiftVersions = models.OpenshiftVersions{
+	"4.9": models.OpenshiftVersion{
```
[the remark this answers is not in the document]
danielerez: good enough for tests:) fixed

cmd/main.go
```diff
@@ -204,7 +205,17 @@ func main() {
 	failOnError(json.Unmarshal([]byte(Options.OpenshiftVersions), &openshiftVersionsMap),
 		"Failed to parse supported openshift versions JSON %s", Options.OpenshiftVersions)
 
-	log.Println(fmt.Sprintf("Started service with OCP versions %v", Options.OpenshiftVersions))
+	var rhcosVersionsArray models.RhcosVersions
+	if Options.RhcosVersions == "" {
+		log.Warn("RhcosVersions list is empty - using only OpenshiftVersions")
```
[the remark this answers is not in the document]
danielerez: Right, until supporting RhcosVersions in all relevant projects, we should keep it as optional. Added a TODO.

data/default_rhcos_versions.json
```diff
@@ -0,0 +1,23 @@
+[
+    {
```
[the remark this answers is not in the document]
danielerez: We've decided to go with an array format for aligning with [k8s API convention](https://github.com/openshift/assisted-service/pull/2369#:~:text=for%20aligning%20with-,k8s%20API%20convention,-and%20easier%20handling) and easier handling in AgentServiceConfig: https://github.com/openshift/assisted-service/blob/d5e953b1e8d68fa3470c433fc8141d882bc3e01e/docs/enhancements/arm64-cpu-architecture.md#ocp-versions

YuviGold (approved):
LGTM

models/os_image.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/go-openapi/errors"
+	"github.com/go-openapi/strfmt"
+	"github.com/go-openapi/swag"
+	"github.com/go-openapi/validate"
+)
+
+// OsImage os image
+//
+// swagger:model os-image
+type OsImage struct {
```
mhrivnak, line 18: I think this would still be better if it matched the [field names used in the k8s API](https://github.com/openshift/assisted-service/blob/72eb9bb/api/v1beta1/agentserviceconfig_types.go#L27-L37). Any reason not to do that?

FWIW part of the reasoning for the naming there avoiding "RHCOS" is that this could all theoretically be used with OKD and fedora coreos.

models/rhcos_versions.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/go-openapi/errors"
+	"github.com/go-openapi/strfmt"
+	"github.com/go-openapi/swag"
+)
+
+// RhcosVersions rhcos versions
+//
+// swagger:model rhcos-versions
+type RhcosVersions []*OsImage
```
mhrivnak, line 19: remame to `OsImages` ?

mhrivnak (changes requested):

models/os_image.go
mhrivnak: To be clear, I'm not talking about case. I'm talking about the names of all the fields below that are part of this struct. They are very different from the names of the same fields in the k8s API struct.

models/rhcos_versions.go
mhrivnak: Yes I think it makes sense to name everything consistently and go with "OS Image" as the base name across the board.

models/os_image.go
danielerez: oh ok, kept the old keys just for an easier transition to the new format. Changed to correlate with k8s API.

models/rhcos_versions.go
danielerez: sure, changed to OS_IMAGES then.

YuviGold (approved):
LGTM

PR #2378, opened 2021-08-12
MGMT-7443: skip removable disks format

slaviered:
Avoid formatting of USB and SD cards that suppose to belong
to the users and should be controlled by them

Avoid formatting mmblk devices (NVIDIA servers case where such
a device was part of the server infrastructure

# Assisted Pull Request

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees
/assign @avishayt 
/assign @omertuc 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/hostcommands/install_cmd.go
+++ b/internal/host/hostcommands/install_cmd.go
@@ -252,7 +252,8 @@ func (i *installCmd) getDiskUnbootableCmd(ctx context.Context, host models.Host)
 	formatCmds := ""
 	for _, disk := range inventory.Disks {
 		isFcIscsi := strings.Contains(disk.ByPath, "-fc-") || strings.Contains(disk.ByPath, "-iscsi-")
-		if disk.Bootable && !isFcIscsi && !disk.IsInstallationMedia {
+		isMmcblk := strings.Contains(disk.ByPath, "mmcblk") //mmc devices should be treated as removable
+		if disk.Bootable && !disk.Removable && !isMmcblk && !isFcIscsi && !disk.IsInstallationMedia {
 			formatCmds += fmt.Sprintf("dd if=/dev/zero of=%s bs=512 count=1 ; ", hostutil.GetDeviceIdentifier(disk))
 			i.eventsHandler.AddEvent(
 				ctx,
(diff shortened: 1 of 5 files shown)


omertuc:
Are you going to add a corresponding PR to https://github.com/openshift/assisted-installer-agent that will populate the Removable field?

omertuc:
@tsorya we saw that the NVIDIA `mmblk` is not detected as bootable after it has been re-formatted, so why are we excluding it specifically?

mmblk devices can potentially interfere with the boot process just as regular disks might, so if they have an MBR, why are they getting special treatment that other disks don't? Especially that we so that after it has been re-formatted it didn't even have an MBR, so what are we solving here exactly?

avishayt:
> @tsorya we saw that the NVIDIA `mmblk` is not detected as bootable after it has been re-formatted, so why are we excluding it specifically?
> 
> mmblk devices can potentially interfere with the boot process just as regular disks might, so if they have an MBR, why are they getting special treatment that other disks don't? Especially that we saw that after it has been re-formatted it didn't even have an MBR, so what are we solving here exactly?

I would say that we never want to format an SD card in a server.  It will either hold something special like in NVIDIA's case, or not belong to the server.

internal/host/hostcommands/install_cmd.go
```diff
@@ -252,7 +252,8 @@ func (i *installCmd) getDiskUnbootableCmd(ctx context.Context, host models.Host)
 	formatCmds := ""
 	for _, disk := range inventory.Disks {
 		isFcIscsi := strings.Contains(disk.ByPath, "-fc-") || strings.Contains(disk.ByPath, "-iscsi-")
-		if disk.Bootable && !isFcIscsi && !disk.IsInstallationMedia {
+		isMmcblk := strings.Contains(disk.ByPath, "mmcblk") //mmc devices should be treated as removable
```
avishayt, line 255: Consider moving this logic to the agent, which would set removable to True if:
1. The disk is actually removable, or
2. The device starts with mmcblk

slaviered: Similar logic are placed here. Moving it to the agent will cause maintenance issues

avishayt: OK

avishayt (approved):

PR #2393, opened 2021-08-16
MGMT-7330: Create a job in prow CI to run assisted-service unit tests.

pawanpinjarkar:
# Assisted Pull Request

## Description
This adds a new target "ci-unit-test" which will run go tests in CI. A new dockerfile "Dockerfile.unit-test" builds
a new image to be used by CI. The docker file installs postgres database on centos8 and keeps it running in the
background. When CI executes "make ci-unit-tests", the container gets all the required dependencies like go and
postgres db.

- [x] New Feature
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
Building the dockerfile "Dockerfile.unit-test" locally and running docker run. Once inside the container, invoke "make ci-unit-test"
- [] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


Dockerfile.unit-test
```diff
@@ -0,0 +1,22 @@
...
+USER postgres
+RUN mkdir -p /tmp/postgres/data
+
+# start postgres server
+RUN /usr/pgsql-13/bin/initdb -D /tmp/postgres/data/
+
+# start postgres db
+RUN /usr/pgsql-13/bin/pg_ctl -D /tmp/postgres/data/ -l /tmp/postgres/logfile start 
+
+FROM registry.ci.openshift.org/openshift/release:golang-1.16
```
pawanpinjarkar, line 21: This overrides Postgres layer.

pawanpinjarkar:
Some tests are failing which doesn't fail locally with "skipper make unit-test". A bug is created here https://bugzilla.redhat.com/show_bug.cgi?id=1996694

pawanpinjarkar:
The test failures are fixed with the additional fix added here https://github.com/openshift/assisted-service/pull/2393/files#diff-565b2e4180459ba8936345e8130a83201308bbad0596421b519db9281e93b12bR5-R6

Here is slack discussion https://coreos.slack.com/archives/CUPJTHQ5P/p1629727308274500

lranjbar:
You will need to update the commit message to include the JIRA issue number

YuviGold (commented):
In general, It's a bad practice to have two different courses of action.
This approach multiplies the makefile targets, Dockerfile images and eventually runs differently from what we do locally.
If that's possible, I would strongly prefer having the same course of action for both the local running and the CI job. Otherwise, the inconsistency would break one another from time to time.

Other possible solutions to replace the Postgres container would be:
* Deploy Postgres as a pod.
Instead of running an executable in the background, run the DB as a k8s pod. This is how it is going to run in prod which would benefit from similar behavior. OpenShift CI does support deploying pods. That's what we do for the subsystem tests.
* Use `sqlite` for UT.
Most of the tests we have don't require an actual DB. SQLite should be a proper replacement for most of the tests. In case we have any special logic for the DB integration - it should be as part of the component subsystem tests and not a unit test.

cc @filanov @eranco74

osherdp:
@YuviGold I'm voting for trying out sqlite and see if it's working
I understand work has been already done here, but adding more makefile targets and dockerfiles will add work over time for everyone triaging, fixing and augmenting CI

carbonin:
>Deploy Postgres as a pod.
>Instead of running an executable in the background, run the DB as a k8s pod. This is how it is going to run in prod which would benefit from similar behavior. OpenShift CI does support deploying pods. That's what we do for the subsystem tests.

If this is true, why wasn't this investigated first? I actually thought the reason we couldn't run our unit tests in prow was exactly this restriction. @lranjbar ?

>Use sqlite for UT.
>Most of the tests we have don't require an actual DB. SQLite should be a proper replacement for most of the tests. In case we have any special logic for the DB integration - it should be as part of the component subsystem tests and not a unit test.

I would very much prefer not to do this. We do have some logic specific to postgres and I don't think this is a compelling enough reason to use a separate DBMS in tests than we do in prod.

I'd much rather have to figure out an issue due to the different ways we deploy PG and have to figure out why something works in sqlite and not in PG.

carbonin:
>This approach multiplies the makefile targets, Dockerfile images and eventually runs differently from what we do locally.

I'm actually thinking this approach might be a better option then how we run our unit tests currently. Why not standardize on the method that starts only one container rather than managing two running locally?

pawanpinjarkar:
I also think that we should not have a different DBMS just for the sake of running unit tests.

osherdp:
> > This approach multiplies the makefile targets, Dockerfile images and eventually runs differently from what we do locally.
> 
> I'm actually thinking this approach might be a better option then how we run our unit tests currently. Why not standardize on the method that starts only one container rather than managing two running locally?

@carbonin this PR doesn't do this standardization, it leaves the two implementations
Also, I think it's the less standard thing. You'll see much more advocates for one process per container

Either way, let's close a loop on this with a meeting. @pawanpinjarkar you wanna schedule?

pawanpinjarkar:
> > > This approach multiplies the makefile targets, Dockerfile images and eventually runs differently from what we do locally.
> > 
> > 
> > I'm actually thinking this approach might be a better option then how we run our unit tests currently. Why not standardize on the method that starts only one container rather than managing two running locally?
> 
> @carbonin this PR doesn't do this standardization, it leaves the two implementations
> Also, I think it's the less standard thing. You'll see much more advocates for one process per container
> 
> Either way, let's close a loop on this with a meeting. @pawanpinjarkar you wanna schedule?

Sent an invite

lranjbar:
> > Deploy Postgres as a pod.
> > Instead of running an executable in the background, run the DB as a k8s pod. This is how it is going to run in prod which would benefit from similar behavior. OpenShift CI does support deploying pods. That's what we do for the subsystem tests.
> 
> If this is true, why wasn't this investigated first? I actually thought the reason we couldn't run our unit tests in prow was exactly this restriction. @lranjbar ?
> 

Subsystem tests are using a cluster claim approach to set up thier pods. Which means it is claiming an already made OCP cluster outside of the running CI OCP cluster. Claiming an OCP cluster for the unit tests its too much resources for a unit test. Not to mention adding in a cluster claim adds 15mins to an hour to the test to claim a cluster. This approach runs in one container on the CI OCP cluster, like all the other unit tests. The set up time of this approach would be to build the additional image. Our image build today around 5mins so this would probably add a minute or two tops.

To be perfectly honest for unit tests we should not need to spin up a database, OCP cluster, etc. It should just need the base libraries of the language we are using, and the libraries needed by our product. If we want to go down the route of re-implementing our unit tests the route we should take is to simply mock out the database. Move database tests to integration test suites and leave the unit tests in the unit testing suite. Then we wouldn't even need this container.

This approach is the least invasive on our current testing suite and was not seeking to bring up the question of re-implementation. If others want to drive that conversation, that is fine. We were seeking the least changes needed while using the least amount of time and resources for the tests. I think people will raise eyebrows if unit tests take an hour to complete in the CI. Which is possible with the cluster claim approach, if there isn't an OCP cluster in hibernation it has to build one.

> This approach multiplies the makefile targets, Dockerfile images and eventually runs differently from what we do locally.
> 
> > I'm actually thinking this approach might be a better option then how we run our unit tests currently. Why not standardize on the method that starts only one container rather than managing two running locally?
> 

It is not hard to run the tests locally using this from the testing we've done. We could replace existing Makefile targets with this approach. We simply were trying to be the least invasive with the change.

pawanpinjarkar:
Thank you @lranjbar for adding more details. 

@YuviGold @osherdp we can remove the current `unit-test` target from the Makefile or update the same target to use underlying functionality added with the new `ci-unit-test` target. If we do the later part, we will be able to test our unit tests almost in the same manner, be it in CI or in local machines. When we meet tomorrow, I will demo how easy it is to use the new target even locally. 

cc @celebdor

YuviGold:
@lranjbar 

> > > Deploy Postgres as a pod.
> > > Instead of running an executable in the background, run the DB as a k8s pod. This is how it is going to run in prod which would benefit from similar behavior. OpenShift CI does support deploying pods. That's what we do for the subsystem tests.
> > 
> > 
> > If this is true, why wasn't this investigated first? I actually thought the reason we couldn't run our unit tests in prow was exactly this restriction. @lranjbar ?
> 
> Subsystem tests are using a cluster claim approach to set up thier pods. Which means it is claiming an already made OCP cluster outside of the running CI OCP cluster. Claiming an OCP cluster for the unit tests its too much resources for a unit test. Not to mention adding in a cluster claim adds 15mins to an hour to the test to claim a cluster. This approach runs in one container on the CI OCP cluster, like all the other unit tests. The set up time of this approach would be to build the additional image. Our image build today around 5mins so this would probably add a minute or two tops.

We would have to check the timing for that. But we anyhow run all the tests together concurrently. So I don't care about unit tests running faster than the subsystem. Honestly, I think it's fine in case we can claim an OCP cluster for that. If we didn't have this option and we were required to provision a bare-metal machine I'd agree it's too much just for unit tests. But using a hibernated cluster for our CI is fine in my opinion.

> To be perfectly honest for unit tests we should not need to spin up a database, OCP cluster, etc. It should just need the base libraries of the language we are using, and the libraries needed by our product. If we want to go down the route of re-implementing our unit tests the route we should take is to simply mock out the database. Move database tests to integration test suites and leave the unit tests in the unit testing suite. Then we wouldn't even need this container.
> 
> This approach is the least invasive on our current testing suite and was not seeking to bring up the question of re-implementation. If others want to drive that conversation, that is fine. We were seeking the least changes needed while using the least amount of time and resources for the tests. I think people will raise eyebrows if unit tests take an hour to complete in the CI. Which is possible with the cluster claim approach, if there isn't an OCP cluster in hibernation it has to build one.
> 
> > This approach multiplies the makefile targets, Dockerfile images and eventually runs differently from what we do locally.
> > > I'm actually thinking this approach might be a better option then how we run our unit tests currently. Why not standardize on the method that starts only one container rather than managing two running locally?
> 
> It is not hard to run the tests locally using this from the testing we've done. We could replace existing Makefile targets with this approach. We simply were trying to be the least invasive with the change.

We can continue using Jenkins as our temporary solution for UT.
But since we already started to put some thinking and effort into this transition, I suggest that we shouldn't hurry just to make a non-invasive solution. Otherwise, we will have future discussions about UT process again.



> Thank you @lranjbar for adding more details.
> 
> @YuviGold @osherdp we can remove the current `unit-test` target from the Makefile or update the same target to use underlying functionality added with the new `ci-unit-test` target. If we do the later part, we will be able to test our unit tests almost in the same manner, be it in CI or in local machines. When we meet tomorrow, I will demo how easy it is to use the new target even locally.
> 
> cc @celebdor

@pawanpinjarkar  It would be better since that's what I am most concerned about, having two courses of action for local running and CI running.

eranco74:
> > Deploy Postgres as a pod.
> > Instead of running an executable in the background, run the DB as a k8s pod. This is how it is going to run in prod which would benefit from similar behavior. OpenShift CI does support deploying pods. That's what we do for the subsystem tests.
> 
> If this is true, why wasn't this investigated first? I actually thought the reason we couldn't run our unit tests in prow was exactly this restriction. @lranjbar ?
> 
> > Use sqlite for UT.
> > Most of the tests we have don't require an actual DB. SQLite should be a proper replacement for most of the tests. In case we have any special logic for the DB integration - it should be as part of the component subsystem tests and not a unit test.
> 
> I would very much prefer not to do this. We do have some logic specific to postgres and I don't think this is a compelling enough reason to use a separate DBMS in tests than we do in prod.
> 
> I'd much rather have to figure out an issue due to the different ways we deploy PG than have to figure out why something works in sqlite and not in PG.

We can try this, seems to be a good mock for PG
https://betterprogramming.pub/how-to-unit-test-a-gorm-application-with-sqlmock-97ee73e36526

carbonin:
### We all met up, here's a summary:

- We want there to be one single way of running the unit tests that is used locally and in CI
- We want CI unit tests to run as a single pod and don't want to require a cluster of some sort to run unit tests locally
- Ideally we wouldn't add yet another Dockerfile
- We should maintain current workflows when building a new, unified way to run unit tests
    - The current implementation doesn't allow for passing env vars to the test context in `run-ci-unit-test`

### Proposed solution

Replace current `unit-test` target implementation with what is currently in `ci-unit-test`.
Install PG in `Dockerfile.assisted-service-build` and use that image to run unit tests in prow using the new `unit-test` target.

Locally, users should continue to run `skipper make unit-test` (or whatever the new name is) which will now only start a single container locally.

PR #2420, opened 2021-08-18
MGMT-7206: add RELEASE_IMAGES env var for listing release images

danielerez:
# Assisted Pull Request

## Description

In order to support multiple CPU architectures in Versions API, added support for RELEASE_IMAGES env var for listing release images:
- Added release-image object and release-images array to swagger:
   - Each release-image represents the url and version for a specific cpu architecture and ocp version.
- Added the env var to main:
  - Parse and pass to VersionHandler (validates its fields if not empty).
  - The list is optional for now to support additional CPU architectures, i.e. x86_64 will be used as a fallback.
- The default images are listed in default_release_images.json.
  - Contains a list of release images of a specific CPU architecture and OCP version.
- Updated configuration files (configmap/templates).
- To do In the following PRs:
  - ReleaseImages will be used in GetOpenshiftVersion func.
  - Update handle_ocp_versions.py.
  - Support cpu_arechitecture in a new cluster.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

As this PR is adding an infrastructure for later arm64 support, no substantial impact in behaviour should occur. I.e. added tests should cover the changes.

## Assignees

/cc @YuviGold
/cc @filanov
/cc @osherdp
/cc @avishayt

## Checklist

## Reviewers Checklist


data/default_release_images.json
```diff
@@ -0,0 +1,20 @@
+[
+    {
+        "openshift_version": "4.6",
+        "cpu_architecture": "x86_64",
```
flaper87, line 4: should we add some `arm` example? Or would it fail if we add it now?

internal/versions/versions.go
```diff
@@ -266,6 +269,16 @@ func (h *handler) validateVersions() error {
 		if osImage.Version == nil {
 			return errors.Errorf(fmt.Sprintf(missingValueTemplate, "Version"))
 		}
+
+		missingValueTemplate = "Missing value in ReleaseImage for '%s' field"
```
flaper87, line 273: The check below implies that we will have a default architecture. IIRC, the manifests list API doesn't allow for setting a default when using multi-arch repositories, which means we would have to pick the default.

Can we document that `x86_64` will be the default architecture? 

TBH, I would be inclined to make the `cpu_architecture` field required. It is better to be explicit and it will simplify the implementation, debugging, and maintenance.

data/default_release_images.json
danielerez: Yeah, multiple architectures not yet supported by this PR. So I'll add an arm image in an upcoming PR when adding the 4.9 arm64 release image.

internal/versions/versions.go
danielerez: Sure, makes sense. I'll add a validation for the architecture as well, can be mandatory when using this list.

internal/versions/versions.go
```diff
@@ -266,6 +269,19 @@ func (h *handler) validateVersions() error {
...
+		for _, release := range h.releaseImages {
+			if release.URL == nil {
+				return errors.Errorf(fmt.Sprintf(missingValueTemplate, "URL"))
```
flaper87, line 276: Sorry for not noticing this earlier, but do you think we should use the `json` field name instead of Go's? 

Considering this is a user-facing message, we may want to tell them exactly what's missing in the json config.

data/default_release_images.json
```diff
@@ -0,0 +1,20 @@
+[
+    {
+        "openshift_version": "4.6",
```
YuviGold, line 3: I believe this new `releases` json is to support different cluster versions than OCP like OKD.
So maybe this field should be named `version` and have an additional `type` field for `openshift` value?

internal/versions/versions.go
```diff
@@ -266,6 +269,19 @@ func (h *handler) validateVersions() error {
 		if osImage.Version == nil {
 			return errors.Errorf(fmt.Sprintf(missingValueTemplate, "Version"))
 		}
+
+		missingValueTemplate = "Missing value in ReleaseImage for '%s' field"
+		for _, release := range h.releaseImages {
```
YuviGold, line 274: I think this for loop can be outside of the current for loop scope

data/default_release_images.json
danielerez: Actually the main incentive is for an easier multiple architecture support, i.e. having a simple list as os images json instead of a complex structure in openshift_versions. So since we already have a 'version' property for the release and to keep the object simple, we could just add okd_version property later on if required.

internal/versions/versions.go
danielerez: right, done.

YuviGold (approved):

PR #2424, opened 2021-08-19
MGMT-6670: Create sample ACI manifests for dual-stack OCP

mkowalski:
# Assisted Pull Request

## Description

This PR creates a sample AgentClusterInstall manifests for OCP
installation using dual-stack. ACI for SNO cluster provides
required clusterNetwork, machineNetwork and serviceNetwork whereas the
ACI for the multi-node cluster provides clusterNetwork and
serviceNetwork, as machineNetwork values are calculated automatically by
the assisted-service.

Contributes-to: [MGMT-6670](https://issues.redhat.com/browse/MGMT-6670)

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/hive-integration/crds/agentClusterInstall-dualstack-SNO.yaml
+++ b/docs/hive-integration/crds/agentClusterInstall-dualstack-SNO.yaml
@@ -1,0 +1,25 @@
+apiVersion: extensions.hive.openshift.io/v1beta1
+kind: AgentClusterInstall
+metadata:
+  name: test-agent-cluster-install
+  namespace: spoke-cluster
+spec:
+  clusterDeploymentRef:
+    name: single-node
+  imageSetRef:
+    name: openshift-v4.8.0
+  networking:
+    clusterNetwork:
+    - cidr: 10.128.0.0/14
+      hostPrefix: 23
+    - cidr: fd01::/48
+      hostPrefix: 64
+    machineNetwork:
+    - cidr: 192.168.126.0/24
+    - cidr: 1001:db8::/120
+    serviceNetwork:
+    - 172.30.0.0/16
+    - fd02::/112
+  provisionRequirements:
+    controlPlaneAgents: 1
+ #sshPublicKey: ssh-rsa your-public-key-here (optional)
--- a/docs/hive-integration/crds/agentClusterInstall-dualstack.yaml
+++ b/docs/hive-integration/crds/agentClusterInstall-dualstack.yaml
@@ -1,0 +1,34 @@
+apiVersion: extensions.hive.openshift.io/v1beta1
+kind: AgentClusterInstall
+metadata:
+  name: test-agent-cluster-install
+  namespace: spoke-cluster
+spec:
+  apiVIP: 1.2.3.8
+  clusterDeploymentRef:
+    name: test-cluster
+  imageSetRef:
+    name: openshift-v4.8.0
+  ingressVIP: 1.2.3.9
+  networking:
+    clusterNetwork:
+    - cidr: 10.128.0.0/14
+      hostPrefix: 23
+    - cidr: fd01::/48
+      hostPrefix: 64
+    serviceNetwork:
+    - 172.30.0.0/16
+    - fd02::/112
+  provisionRequirements:
+    controlPlaneAgents: 3
+ #sshPublicKey: ssh-rsa your-public-key-here (optional)
+  # By default, SMT (or hyperthreading) is enabled to increase the performance of your machines' cores.
+  # Therefore, you can omit this section unless you wish to disable hyperthreading.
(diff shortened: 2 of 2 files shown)


docs/hive-integration/crds/agentClusterInstall-dualstack-SNO.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: extensions.hive.openshift.io/v1beta1
+kind: AgentClusterInstall
+metadata:
+  name: test-agent-cluster-install
+  namespace: spoke-cluster
+spec:
+  clusterDeploymentRef:
+    name: single-node-dual-stack
```
nmagnezi, line 8: nit: maybe just use 'single-node' here , since this is what we have in CRD examples? 
https://github.com/openshift/assisted-service/blob/577382f491b06a18c61c04076c1eba6c2e71cb4d/docs/hive-integration/crds/clusterDeployment.yaml#L4

mkowalski: Yes, makes very much sense. Fixed

djzager:
My only question for this is how do we bring attention to it. Three weeks from now when I imagine myself wanting info about dual-stack, what breadcrumbs are there that would lead me to these files and the fields/values I should be paying attention to?

docs/hive-integration/crds/agentClusterInstall-dualstack-SNO.yaml
```diff
@@ -0,0 +1,25 @@
...
+  clusterDeploymentRef:
+    name: single-node
+  imageSetRef:
+    name: openshift-v4.8.0
+  networking:
+    clusterNetwork:
+    - cidr: 10.128.0.0/14
+      hostPrefix: 23
+    - cidr: fd01::/48
+      hostPrefix: 64
```
lranjbar, line 16: So for dual stack clearly we will need to define two cidr blocks... but it isn't clear which cidr block is used for what. Another way to think about that question is what happens if I put two IP4 or two IP6 cidr blocks here? How does this differ from that?

docs/hive-integration/crds/agentClusterInstall-dualstack.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: extensions.hive.openshift.io/v1beta1
+kind: AgentClusterInstall
+metadata:
+  name: test-agent-cluster-install
+  namespace: spoke-cluster
+spec:
+  apiVIP: 1.2.3.8
```
lranjbar, line 7: I'm assuming this has to be IP4 due to other discussions I've heard but would be good to mention that somewhere.

lranjbar (commented):
Probably out of the scope of this PR but it would be helpful to have a doc that explains the fields in our CRDs and when you will configure them differently. We have some explanation of what the CRDs are used for in a different document... I'm seeing us enumerate on these CRDs and but there is little explanation of the samples and why there are so many.

docs/hive-integration/crds/agentClusterInstall-dualstack-SNO.yaml
mkowalski: Right, so I think it still revolves around the fact that we need to have a place where we put what's supported and so on... To answer specifically your questions
1) Your pods will be getting IP address from this block(s). Depending on your manifest, your pods will be IPv4 or dual-stack, that's why you have 2 of them, but they are both citizens with the same rights
2) If you put two IPv4 and/or 2 IPv6 blocks, we will fail to deploy your cluster (via the validator) because this is not a legal combination from the perspective of the OCP (or maybe even Kubernetes). 

I think in principle, given that we are starting to do more and more advanced networking, it would be good if we had a Networking Intro document, but then again, question is "where?"

docs/hive-integration/crds/agentClusterInstall-dualstack.yaml
mkowalski: Correct. For dual-stack API and Ingress IPs have to be IPv4 (until Metal supports an array of addresses here; luckily there is already a Jira for this). As above, I'm happy to craft the doc explaining all this as soon as we have _some_ idea where it should live

flaper87: +1 for the doc!

Let's write this doc, and then add a comment with a link to the networking docs somewhere in this yaml.

flaper87 (approved):

PR #2427, opened 2021-08-19
MGMT-7603: Update static network config and image type in infra-env PATCH

carbonin:
## Description

Previously, you could send these parameters, but they would not get used.
This commit also passes the image type from the update parameters through the generate calls as it could be different from what is currently set in the database and we don't want to update the value until after the image has been created successfully.

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Deployed an image built using my branch to minikube and tested that an infraenv was updated properly.

## Assignees

/cc @yevgeny-shnaidman 
/cc @avishayt 

## Checklist

## Reviewers Checklist


carbonin:
cc @djzager

flaper87 (commented):
lgtm, will let someone with more knowledge in the API take a look

internal/bminventory/inventory.go
```diff
@@ -1101,7 +1101,7 @@ func (b *bareMetalInventory) GenerateClusterISOInternal(ctx context.Context, par
 	return b.GetClusterInternal(ctx, installer.GetClusterParams{ClusterID: params.ClusterID})
 }
 
-func (b *bareMetalInventory) GenerateInfraEnvISOInternal(ctx context.Context, infraEnv *common.InfraEnv) error {
```
yevgeny-shnaidman, line 1104: Lets change it to generateInfraEnvISOInternal, it is used only internally

internal/bminventory/inventory.go
```diff
@@ -1160,15 +1160,15 @@ func (b *bareMetalInventory) GenerateInfraEnvISOInternal(ctx context.Context, in
 	}
 
 	if imageExists {
-		if err = b.updateImageInfoPostUpload(ctx, infraEnv, infraEnv.ProxyHash, infraEnv.Type, false); err != nil {
```
yevgeny-shnaidman, line 1163: I think that we can set the image type into the InfraEnv at the same time we set all the rest of the parameters (before calling GenerateClusterISOInternal). That way it will be more consistent with the update once we move to image service. And we have Generated field that speicifies if image was created or not

carbonin: I was trying to be consistent with the v1 generate here where we only update the image type after it's created successfully.

If we fail to create, then we'll end up with the wrong image type, right?

carbonin: Plus we have to pass something here.

yevgeny-shnaidman: before creating image we set the "generated field to false, and updated it to true only after generation has succeeded. So even if we set the image type before creation, and creation fails, then next time GenerateclusterISO is called we check the "generated" and since it false we try to create the image again. So eventually the mage is created with the right type ( i think  :)).

PR #2432, opened 2021-08-20
MGMT-7464: Pass multiple machine networks to install-config.yaml

mkowalski:
# Assisted Pull Request

## Description

This PR changes the logic of install-config.yaml generation regarding
the machine networks. Till now we have been always passing only single
IPv4 subnet (coming either from the cluster parameters or discovered
from the bootstrap host).

With this change the logic is as follows

* if cluster params contains multiple machine networks, we pass them all
  to the installcfg structure
* if cluster params do not have any machine networks, we generate them
  from the cluster's bootstrap node's inventory; we pick at most one
  IPv4 network and at most one IPv6 network

Contributes-to: [MGMT-7464](https://issues.redhat.com/browse/MGMT-7464)

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @YuviGold 
/cc @flaper87 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/installcfg/installcfg.go
+++ b/internal/installcfg/installcfg.go
@@ -391,11 +391,14 @@ func (i *installConfigBuilder) getInstallConfig(cluster *common.Cluster, addRhCa
 
 		bootstrapCidr := network.GetPrimaryMachineCidrForUserManagedNetwork(cluster, i.log)
 		if bootstrapCidr != "" {
-			i.log.Infof("None-Platform: Selected bootstrap machine network CIDR %s for cluster %s", bootstrapCidr, cluster.ID.String())
-			cfg.Networking.MachineNetwork = []MachineNetwork{{Cidr: bootstrapCidr}}
-			cluster.MachineNetworks = network.CreateMachineNetworksArray(bootstrapCidr)
+			i.log.Infof("None-Platform or SNO: Selected bootstrap machine network CIDR %s for cluster %s", bootstrapCidr, cluster.ID.String())
+			machineNetwork := []MachineNetwork{}
+			cluster.MachineNetworks = network.GetMachineNetworksFromBoostrapHost(cluster, i.log)
+			for _, net := range cluster.MachineNetworks {
+				machineNetwork = append(machineNetwork, MachineNetwork{Cidr: string(net.Cidr)})
+			}
+			cfg.Networking.MachineNetwork = machineNetwork
 			cfg.Networking.NetworkType = swag.StringValue(cluster.NetworkType)
-
 		} else {
 			cfg.Networking.MachineNetwork = nil
 		}
--- a/internal/installcfg/installcfg_test.go
+++ b/internal/installcfg/installcfg_test.go
@@ -48,12 +48,13 @@ var _ = Describe("installcfg", func() {
 			NetworkType:            swag.String("OpenShiftSDN"),
 		}}
 		id := strfmt.UUID(uuid.New().String())
+		// By default all the hosts in the cluster are IPv4-only
 		host1 = models.Host{
 			ID:        &id,
 			ClusterID: &clusterId,
 			Status:    swag.String(models.HostStatusKnown),
 			Role:      "master",
-			Inventory: getInventoryStr("hostname0", "bootMode", false),
+			Inventory: getInventoryStr("hostname0", "bootMode", true, false),
 		}
 		id = strfmt.UUID(uuid.New().String())
 		host2 = models.Host{
@@ -61,7 +62,7 @@ var _ = Describe("installcfg", func() {
 			ClusterID: &clusterId,
 			Status:    swag.String(models.HostStatusKnown),
 			Role:      "worker",
-			Inventory: getInventoryStr("hostname1", "bootMode", false),
+			Inventory: getInventoryStr("hostname1", "bootMode", true, false),
 		}
 
 		host3 = models.Host{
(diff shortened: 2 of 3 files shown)


internal/installcfg/installcfg_test.go
```diff
@@ -528,20 +572,23 @@ var _ = Describe("ValidateInstallConfigPatch", func() {
...
+				IPV4Addresses: []string{},
+				IPV6Addresses: []string{},
 				MacAddress:    "some MAC address",
 			},
 		},
 	}
-	if ipv6only {
-		inventory.Interfaces[0].IPV4Addresses = nil
+	if ipv4 {
+		inventory.Interfaces[0].IPV4Addresses = append(make([]string, 0), "10.35.20.10/24")
```
djzager, line 588: Is there a reason this isn't just:
```suggestion
		inventory.Interfaces[0].IPV4Addresses = []string{"10.35.20.10/24"}
```

djzager: Is one preferred over the other? I suspect there are likely compiler things to make them equally efficient...but I feel like I'm missing something.

internal/installcfg/installcfg.go
```diff
@@ -391,11 +391,14 @@ func (i *installConfigBuilder) getInstallConfig(cluster *common.Cluster, addRhCa
...
-			i.log.Infof("None-Platform: Selected bootstrap machine network CIDR %s for cluster %s", bootstrapCidr, cluster.ID.String())
-			cfg.Networking.MachineNetwork = []MachineNetwork{{Cidr: bootstrapCidr}}
-			cluster.MachineNetworks = network.CreateMachineNetworksArray(bootstrapCidr)
+			i.log.Infof("None-Platform or SNO: Selected bootstrap machine network CIDR %s for cluster %s", bootstrapCidr, cluster.ID.String())
+			machineNetwork := []MachineNetwork{}
+			for _, net := range network.GetMachineNetworksFromBoostrapHost(cluster, i.log) {
+				machineNetwork = append(machineNetwork, MachineNetwork{Cidr: string(net.Cidr)})
+			}
+			cfg.Networking.MachineNetwork = machineNetwork
+			cluster.MachineNetworks = network.GetMachineNetworksFromBoostrapHost(cluster, i.log)
```
djzager, line 400: I'm struggling to understand all that is happening in this block of code. I don't think all of it has to do with your specific changes.

* Why are we calling the same function twice?
* Why do we call it a MachineNetwork in the installcfg package when we have a MachineNetwork from models? I guess one is the API facing representation versus the internal one? either way it's weird.

Here is a suggestion that makes sense to me, tell me if/why it's crazy and wouldn't work (or why you don't like it).

```suggestion
			i.log.Infof("None-Platform or SNO: Selected bootstrap machine network CIDR %s for cluster %s", bootstrapCidr, cluster.ID.String())
			machineNetwork := []MachineNetwork{}
			cluster.MachineNetworks = network.GetMachineNetworksFromBoostrapHost(cluster, i.log)
			for _, net := range cluster.MachineNetworks {
				machineNetwork = append(machineNetwork, MachineNetwork{Cidr: string(net.Cidr)})
			}
			cfg.Networking.MachineNetwork = machineNetwork
```

One more question. Why do we need, here, a `[]MachineNetwork{}` and a `[]*models.MachineNetwork{}`?

djzager (commented):
A few hopefully reasonable comments. Not sure how to exercise this PR to make sure it's functioning as expected. Thoughts? Screenshots? or anything we can derive from prow results?

internal/installcfg/installcfg_test.go
mkowalski: They are indeed equivalent, it's just my syntax is much more convoluted and yours is much more clear :)
Fixed.

internal/installcfg/installcfg.go
mkowalski: >Why are we calling the same function twice?

That was stupid and is fixed now

>Why do we call it a MachineNetwork in the installcfg package when we have a MachineNetwork from models? I guess one is the API facing representation versus the internal one? either way it's weird.

Your guess is correct. If you look at `models.MachineNetwork` you can see it stores CIDR as well as clusterID because this is how we store it in the DB. On the other hand the `MachineNetwork` from the `installcfg` package is a local struct used as a wrapper around a single property (string, CIDR) which is used in the generation of install-config.yaml.

`models.MachineNetwork` has been created very very recently whereas `MachineNetwork` from the `installcfg` exists since forever. I guess it could be somehow possible to incorporate the former into the latter, but my quick attempt in doing so resulted in a lot of compiler complains, so I think it would be a much bigger refactoring than we want to do right now.

>Why do we need, here, a []MachineNetwork{} and a []*models.MachineNetwork{}?

We use `[]MachineNetwork{}` in order to locally generate a part of install-config.yaml because this is a structure that stores `networking.MachineNetwork` part from this yaml. On the other hand `[]*models.MachineNetwork{}` is how we store it globally, as a `cluster` parameter. In general the piece of code we are touching here is a moment when `cluster.MachineNetwork` gets converted to the `networking.MachineNetwork` in install-config.yaml; as those two are represented by slightly different structures (because of the mentioned above `clusterID`), we have a mix of structs.

Just to be clear, I would happily collapse both of those into a single `something.MachineNetwork` type, but I don't see how to easily do it. Maybe @YuviGold has some ideas, though

YuviGold: I looked as well if it is possible to merge these structs but I couldn't find how.
First, there's the `clusterID` field and possible other metadata fields we store in our DB.
Second, there's the `yaml` tags that aren't generated by swagger. (swagger automatically generates `json` tags).

djzager (approved):
/lgtm

PR #2412, opened 2021-08-18
MGMT-7588: Fix event wrapper to notify kube-api

arielireni:
Bug fix in `controller_event_wrapper.go` file, where the functions `SendClusterEvent` and `SendHostEvent` should notify kube-api according to the event type.

This PR continues #1854

Signed-off-by: arielireni <aireni@redhat.com>

# Assisted Pull Request

## List all the issues related to this PR

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [x] No tests needed

## Assignees

/cc @masayag 
/cc @filanov 

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- [ ] Are the title and description (in both PR and commit) meaningful and clear?
- [ ] Is there a bug required (and linked) for this change?
- [ ] Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #1854, opened 2021-05-30
MGMT-4872: Initial events generation

masayag:
Events generation is designed to expose a uniformed method of
initializing the assisted-service events, based on each event definition.
The event definition will require a set of parameters for each event type.
It will be used for generating a function for initializing the event with
the provided parameters, hence enforcing to provide the parameters at
initialization and not to miss any needed information.
The events will be defined in `./data/events.yaml`.

In order to introduce a new event, the developer will have to follow the
next steps:
1. Add event definition to `data/events.yaml` (event definition will be
   discussed ahead).
2. Generate the code for creating the event by: `skipper make generate
   generate-events`
3. Use the generated function for the event from
   `./internal/common/events.event.go` for creating the event to be
   emitted by the event handler.

Event definition should specify the following attributes:

-  name: The name of the event which needs to remain unique and constant
   as it may be referred by the service's clients (e.g. by the UI).
- message: A template of the message that will be rendered if it
  contains references to the properties. E.g. the message "Install
  cluster {cluster_id}" expects the existence of a property named
  cluster_id
- event_type: Can be either "cluster" or "host". "cluster" type requires
  the existence of "cluster_id" in properties and "host" type requires
  the existence of "host_id".
- severity: Any of info, warning, error or critical
- properties: A list of properties to be rendered into the message (if
  referred by) or added as identifier of the message (e.g. cluster_id, host_id)
  with the type of the property.

Event's severity is defined as a field of the event definition in ./data/events.yaml.
In special events (e.g. generic events) where the severity value of the event is set
by the caller, the severity value can be provided as a property.
In this case a priority will be given to the severity value provided as a property.

PR #2444, opened 2021-08-22
MGMT-4872: Generate cluster registration\validation events

arielireni:
# Assisted Pull Request

## Description
Generate cluster events related to cluster registration\validation from internal\cluster\cluster.go file by the new events API.

This PR continues #2412

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

Signed-off-by: arielireni <aireni@redhat.com>


internal/cluster/cluster_test.go
```diff
@@ -2138,7 +2140,7 @@ var _ = Describe("insufficient_state", func() {
 			BaseDNSDomain:   "test.com",
 			PullSecretSet:   true,
 		}}
-		mockEvents.EXPECT().AddEvent(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
+		mockEvents.EXPECT().SendClusterEvent(gomock.Any(), gomock.Any()).AnyTimes()
```
masayag, line 2143: Can this expectation be reduced to use eventMatcher to verify an expected event was emitted in this context ?

internal/events/eventstest/events_test_utils.go
```diff
@@ -0,0 +1,86 @@
+package eventstest
```
masayag, line 1: @carbonin based on your comment on https://github.com/openshift/assisted-service/pull/1854#discussion_r687079898, WDYT ?

internal/events/eventstest/events_test_utils.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		return false
+	}
+}
+
+type EventMatcher struct {
+	matchers []eventPartMatcher
+}
+
+func (e EventMatcher) Matches(input interface{}) bool {
```
masayag, line 71: pls add comment `// implements gomock.Matcher`

internal/events/eventstest/events_test_utils.go
```diff
@@ -0,0 +1,86 @@
...
+	for _, matcher := range e.matchers {
+		if !matcher(event) {
+			return false
+		}
+	}
+	return true
+}
+
+func (e EventMatcher) String() string {
+	return "input contains Ids"
```
masayag, line 85: pls change message to "verifies event fields"

internal/cluster/cluster_test.go
arielireni: done

internal/events/eventstest/events_test_utils.go
arielireni: done

internal/events/eventstest/events_test_utils.go
arielireni: done

arielireni:
@filanov @YuviGold Please review

internal/events/eventstest/events_test_utils.go
filanov: Sorry for the late reply, we don't have a convention for this. 
IMO if we have something that is used by multiple tests we can have a separate package like here. 
we have `test_utils.go` in multiple packages and i don't like it. 
So having a separate model for this is actually a good idea.

internal/common/events/events.go
```diff
@@ -375,3 +375,633 @@ func (e *HostRegistrationFailedSettingPropertiesEvent) FormatMessage() string {
     return e.format(&s)
 }
 
+//
```
YuviGold, line 378: I believe this file is generated. GitHub doesn't recognize it that way. Please take a look how it could be marked as a generated file so GitHub would collapse it accordingly. 
This is not a blocker.

data/events.yaml
```diff
@@ -39,3 +39,75 @@ x-events:
...
+    cluster_id: UUID
+    validation_id: string
+
+- name: cluster_validation_fixed
+  message: "Cluster validation '{validation_id}' is now fixed"
+  event_type: cluster
+  severity: "info"
+  properties:
+    cluster_id: UUID
+    validation_id: string
```
YuviGold, line 106: I see some inconsistency with how fields are being handled. Either presented as part of the message or as a property.
IMO a property only is preferred than the string interpolation inside the message could be avoided.
Either way, this scheme should be aligned (since I see some places the property is being used and some not)

data/events.yaml
```diff
@@ -39,3 +39,75 @@ x-events:
...
+  message: "Failed to register add-hosts cluster with name '{cluster_name}'. Error: {error}"
+  event_type: cluster
+  severity: "error"
+  properties:
+    cluster_id: UUID
+    cluster_name: string
+    error: string
+
+- name: registered_add_hosts_cluster
+  message: "Registered add-hosts cluster {cluster_name}"
```
YuviGold, line 70: Merge this `registered_add_hosts_cluster` with `registered_cluster` event and add the ClusterKind property

data/events.yaml
YuviGold, line 106: Add the validation message as a property. This could be really useful for debugging and not required to go into logs.

filanov: things like cluster id is probably something that we don't need in the message but i think that host name or validation id can help because ui will probably not show all the properties.

data/events.yaml
YuviGold: IMO the user doesn't have any benefit from knowing that the cluster is "add-hosts" cluster (let's say he actually understand that, although I believe he won't).
So I think the event could be simpler with a single "Register cluster" as the primer event. The property of kind could be added for our usage.

PR #2512, opened 2021-08-30
MGMT-6670: Fix handling of empty network in various network params

mkowalski:
# Assisted Pull Request

## Description

This PR fixes handling of Machine, Cluster and Service Networks when
the value passed is a non-empty array with empty objects.

The current assumption that the MachineNetworks (or other) variable is
either empty or contains a valid set of CIDRs is incorrect, as it has
been discovered that the following requests arrive to the service

```
--data '{"machine_networks": [{}]}'
```

Those should be rather arriving as

```
--data '{"machine_networks": []}'
```

but in order to handle all the possibilities, we add a scenario for the
former in this change.

Contributes-to: [MGMT-6670](https://issues.redhat.com/browse/MGMT-6670)

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @YuviGold 
/cc @flaper87 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -536,14 +536,14 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 	}
 
 	// TODO MGMT-7365: Deprecate single network
-	if len(params.NewClusterParams.ClusterNetworks) > 0 {
+	if common.IsSliceNonEmpty(params.NewClusterParams.ClusterNetworks) {
 		cluster.ClusterNetworkCidr = string(params.NewClusterParams.ClusterNetworks[0].Cidr)
 		cluster.ClusterNetworkHostPrefix = params.NewClusterParams.ClusterNetworks[0].HostPrefix
 	}
-	if len(params.NewClusterParams.ServiceNetworks) > 0 {
+	if common.IsSliceNonEmpty(params.NewClusterParams.ServiceNetworks) {
 		cluster.ServiceNetworkCidr = string(params.NewClusterParams.ServiceNetworks[0].Cidr)
 	}
-	if len(params.NewClusterParams.MachineNetworks) > 0 {
+	if common.IsSliceNonEmpty(params.NewClusterParams.MachineNetworks) {
 		cluster.MachineNetworkCidr = string(params.NewClusterParams.MachineNetworks[0].Cidr)
 	}
 
@@ -2200,7 +2200,7 @@ func (b *bareMetalInventory) updateNonDhcpNetworkParams(updates map[string]inter
 		ingressVip = *params.ClusterUpdateParams.IngressVip
 	}
 	if params.ClusterUpdateParams.MachineNetworks != nil &&
-		len(params.ClusterUpdateParams.MachineNetworks) > 0 {
+		common.IsSliceNonEmpty(params.ClusterUpdateParams.MachineNetworks) {
 		err := errors.New("Setting Machine network CIDR is forbidden when cluster is not in vip-dhcp-allocation mode")
 		log.WithError(err).Warnf("Set Machine Network CIDR")
 		return common.NewApiError(http.StatusBadRequest, err)
@@ -2389,7 +2389,7 @@ func (b *bareMetalInventory) updateNetworks(db *gorm.DB, params installer.Update
 		cluster.ClusterNetworks = params.ClusterUpdateParams.ClusterNetworks
 
 		// TODO MGMT-7365: Deprecate single network
-		if len(params.ClusterUpdateParams.ClusterNetworks) > 0 {
+		if common.IsSliceNonEmpty(params.ClusterUpdateParams.ClusterNetworks) {
 			updates["cluster_network_cidr"] = string(params.ClusterUpdateParams.ClusterNetworks[0].Cidr)
 			updates["cluster_network_host_prefix"] = params.ClusterUpdateParams.ClusterNetworks[0].HostPrefix
 		} else {
@@ -2407,7 +2407,7 @@ func (b *bareMetalInventory) updateNetworks(db *gorm.DB, params installer.Update
 		cluster.ServiceNetworks = params.ClusterUpdateParams.ServiceNetworks
 
 		// TODO MGMT-7365: Deprecate single network
-		if len(params.ClusterUpdateParams.ServiceNetworks) > 0 {
+		if common.IsSliceNonEmpty(params.ClusterUpdateParams.ServiceNetworks) {
 			updates["service_network_cidr"] = string(params.ClusterUpdateParams.ServiceNetworks[0].Cidr)
 		} else {
 			updates["service_network_cidr"] = ""
@@ -2415,25 +2415,22 @@ func (b *bareMetalInventory) updateNetworks(db *gorm.DB, params installer.Update
 	}
 
 	if params.ClusterUpdateParams.MachineNetworks != nil {
+		for _, machineNetwork := range params.ClusterUpdateParams.MachineNetworks {
+			if err = network.VerifyMachineCIDR(string(machineNetwork.Cidr)); err != nil {
+				return common.NewApiError(http.StatusBadRequest, errors.Wrapf(err, "Machine network CIDR %s", string(machineNetwork.Cidr)))
+			}
+		}
 		cluster.MachineNetworks = params.ClusterUpdateParams.MachineNetworks
 
 		// TODO MGMT-7365: Deprecate single network
-		if len(params.ClusterUpdateParams.MachineNetworks) > 0 {
+		if common.IsSliceNonEmpty(params.ClusterUpdateParams.MachineNetworks) {
 			updates["machine_network_cidr"] = string(params.ClusterUpdateParams.MachineNetworks[0].Cidr)
 		} else {
 			updates["machine_network_cidr"] = ""
 		}
 	}
 
-	for index := range cluster.MachineNetworks {
-		if string(cluster.MachineNetworks[index].Cidr) != "" {
-			if err = network.VerifyMachineCIDR(string(cluster.MachineNetworks[index].Cidr)); err != nil {
-				return common.NewApiError(http.StatusBadRequest, errors.Wrapf(err, "Machine network CIDR %s", string(cluster.MachineNetworks[index].Cidr)))
-			}
-		}
-	}
-
-	if len(params.ClusterUpdateParams.MachineNetworks) > 0 {
+	if common.IsSliceNonEmpty(params.ClusterUpdateParams.MachineNetworks) {
 		if err = validations.ValidateVipDHCPAllocationWithIPv6(vipDhcpAllocation, string(cluster.MachineNetworks[0].Cidr)); err != nil {
 			return common.NewApiError(http.StatusBadRequest, err)
 		}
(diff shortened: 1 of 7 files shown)


YuviGold (commented):
There are other places with this simple condition
Find and replace all the 
```
Networks) > 0
```

I would suggest to make a generic function `IsArrayNonEmpty(val)` if possible

mkowalski:
I'm trying to craft a test for this scenario, but typing is blocking me from doing it in an easy way. Any advice appreciated

The attempt of writing
```
						ClusterUpdateParams: &models.ClusterUpdateParams{
							ClusterNetworks: []*models.ClusterNetwork{},
							ServiceNetworks: []*models.ServiceNetwork{},
							MachineNetworks: []*models.MachineNetwork{{}},
						},
```
is not valid because `{}` is `nil` and not `models.somethingsomething` interface

YuviGold (commented):
Add UT

internal/common/common.go
```diff
@@ -150,8 +151,19 @@ func GetNetworkCidrAttr(obj interface{}, fieldName string) []*string {
...
+		address := funk.Get(elem, "Cidr")
+		if address != nil {
+			addresses = append(addresses, swag.String(string(address.(models.Subnet))))
+		}
 	})
 
 	return addresses
 }
+
+func IsArrayNonEmpty(arg interface{}) bool {
```
lranjbar, line 163: This is a nit, but I think this should be called `IsSliceNonEmpty()` because it isn't actually an array. That's the reason reflect is required here after playing around with the reflect package a bit today.

internal/bminventory/inventory.go
```diff
@@ -2406,33 +2406,30 @@ func (b *bareMetalInventory) updateNetworks(db *gorm.DB, params installer.Update
...
-		if len(params.ClusterUpdateParams.ServiceNetworks) > 0 {
+		if common.IsNetworksSliceNonEmpty(params.ClusterUpdateParams.ServiceNetworks) {
 			updates["service_network_cidr"] = string(params.ClusterUpdateParams.ServiceNetworks[0].Cidr)
 		} else {
 			updates["service_network_cidr"] = ""
 		}
 	}
 
 	if params.ClusterUpdateParams.MachineNetworks != nil {
+		for _, machineNetwork := range params.ClusterUpdateParams.MachineNetworks {
```
mkowalski, line 2417: @YuviGold: I'm moving this validator a bit earlier, i.e. before setting the values for the first time. Let me know if you think this is an issue and if we should really first set and then (re)validate

internal/common/common.go
```diff
@@ -150,8 +151,25 @@ func GetNetworkCidrAttr(obj interface{}, fieldName string) []*string {
 	}
 
 	funk.ForEach(field, func(elem interface{}) {
-		addresses = append(addresses, swag.String(string(funk.Get(elem, "Cidr").(models.Subnet))))
+		address := funk.Get(elem, "Cidr")
+		if address != nil {
```
mkowalski, line 155: This is a bugfix I'm piggybacking here. Without this check it could happen that we try to get `Cidr` from an object without `Cidr` and then cast it to `models.Subnet`. Runtime complains then that we try to convert interfaces in an illegal way.

internal/common/common.go
mkowalski: Totally yes. And given it's a helper specifically for network subnets, I renamed it to be even more specific

mkowalski:
/approve cancel

I don't feel anymore confident enough to self-approve. Plz someone else do approval and someone else lgtm :)

internal/bminventory/inventory.go
```diff
@@ -2627,7 +2624,7 @@ func validateUserManagedNetworkConflicts(params *models.ClusterUpdateParams, sin
 		log.WithError(err)
 		return common.NewApiError(http.StatusBadRequest, err)
 	}
-	if params.MachineNetworks != nil && len(params.MachineNetworks) > 0 && !singleNodeCluster {
+	if common.IsNetworksSliceNonEmpty(params.MachineNetworks) && !singleNodeCluster {
```
mkowalski, line 2627: This `if` got simplified because `IsNetworksSliceNonEmpty` handles the case of getting `nil` as an argument and simply returns `False`

internal/common/common.go
```diff
@@ -150,8 +151,25 @@ func GetNetworkCidrAttr(obj interface{}, fieldName string) []*string {
...
+		}
 	})
 
 	return addresses
 }
+
+// IsNetworksSliceNonEmpty checks whether the provided slice of networks, i.e. objects containing
+// CIDR is non-empty. The slice is assumed to be non-empty if at least one of its elements
+// contains a non-empty CIDR.
+func IsNetworksSliceNonEmpty(arg interface{}) bool {
```
flaper87, line 166: To some extent, it seems like this is something that we should validate at the source or have specific types/functions to validate `MachineNetworks` and `ClusterNetworks`.

One option could be to have an interface with a `HasCIDR` / `IsEmpty` function, which both `MachineNetworks` and `ClusterDeployment` implement. This should provide the same results in a hopefully more maintainable/easier-to-follow way. 

Not opposed to the solution in this PR, just throwing another idea out there :)

internal/common/common.go
```diff
@@ -150,8 +151,25 @@ func GetNetworkCidrAttr(obj interface{}, fieldName string) []*string {
...
+			addresses = append(addresses, swag.String(string(address.(models.Subnet))))
+		}
 	})
 
 	return addresses
 }
+
+// IsNetworksSliceNonEmpty checks whether the provided slice of networks, i.e. objects containing
+// CIDR is non-empty. The slice is assumed to be non-empty if at least one of its elements
+// contains a non-empty CIDR.
```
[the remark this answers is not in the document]
mkowalski: Done, added bunch of examples with explanations

internal/bminventory/inventory.go
YuviGold: I think that the initial purpose was to verify that the calculated machine cidr is valid. Although in case it wasn't it means we have a bug in our code... So I don't know if that's a requirement. We actually can also validate the machine CIDR during the calculation and assume it is valid (as we don't validate any other value that passes between our own components)

WDYT? @ori-amizur @tsorya

mkowalski: I'm not sure I'm not missing something -- machine CIDR is not always calculated by the code; there are cases when user provides it explicitly and then we need to validate it _somewhere_

YuviGold: Right.
So either:
* machine CIDR can be calculated and assigned to the cluster; just like when it is provided - it is assigned, and eventually the cluster value is verified. (previous solution)
* It can be calculated and assigned, without the validation (assuming it is valid); only when the value is provided by the user - it is validated before the assignment (your solution)

mkowalski: >machine CIDR can be calculated and assigned to the cluster; just like when it is provided - it is assigned, and eventually the cluster value is verified. (previous solution)

I like the fact that we validate values that were calculated, not only those that were coming from the user. But what I don't really like is the order of those actions, i.e.

```
	if params.ClusterUpdateParams.MachineNetworks != nil {
		cluster.MachineNetworks = params.ClusterUpdateParams.MachineNetworks
[...]
	}
 	for index := range cluster.MachineNetworks { 
[...] validate
	}
```

This flow means that we blindy take what came from the user, assign it to the cluster and if it fails later we throw an error, but we don't really rollback the change. I'm not sure however how to make this flow so that all the scenarios are completely covered...

YuviGold: There's a rollback. In case of an error - the transaction would be canceled and rolleback.
https://github.com/openshift/assisted-service/blob/51ee4ce418dd208fcc902a4e54064bc74dbbd57b/internal/bminventory/inventory.go#L2067-L2078

But I think we can validate the machine network as well in the calculation itself.
https://github.com/openshift/assisted-service/blob/51ee4ce418dd208fcc902a4e54064bc74dbbd57b/internal/bminventory/inventory.go#L2219-L2239

I'm fine with either way.

internal/bminventory/inventory_test.go
```diff
@@ -5038,6 +5038,90 @@ var _ = Describe("cluster", func() {
...
+						ClusterUpdateParams: &models.ClusterUpdateParams{
+							ClusterNetworks: []*models.ClusterNetwork{{Cidr: ""}},
+							ServiceNetworks: []*models.ServiceNetwork{},
+							MachineNetworks: []*models.MachineNetwork{},
+						},
+					})
+					verifyApiError(reply, http.StatusBadRequest)
+				})
+
+				It("Empty networks 3 - invalid HostPrefix, ClusterNetwork", func() {
```
YuviGold, line 5065:
```suggestion
				It("Empty networks - invalid HostPrefix, ClusterNetwork", func() {
```

internal/bminventory/inventory_test.go
```diff
@@ -5038,6 +5038,90 @@ var _ = Describe("cluster", func() {
...
+						ClusterUpdateParams: &models.ClusterUpdateParams{
+							ClusterNetworks: []*models.ClusterNetwork{},
+							ServiceNetworks: []*models.ServiceNetwork{},
+							MachineNetworks: []*models.MachineNetwork{{}},
+						},
+					})
+					verifyApiError(reply, http.StatusBadRequest)
+				})
+
+				It("Empty networks 5 - invalid CIDR, MachineNetwork", func() {
```
YuviGold, line 5113:
```suggestion
				It("Empty networks - invalid CIDR, MachineNetwork", func() {
```

internal/bminventory/inventory_test.go
```diff
@@ -5038,6 +5038,90 @@ var _ = Describe("cluster", func() {
...
+				It("Empty networks - invalid empty ClusterNetwork", func() {
+					reply := bm.UpdateCluster(ctx, installer.UpdateClusterParams{
+						ClusterID: clusterID,
+						ClusterUpdateParams: &models.ClusterUpdateParams{
+							ClusterNetworks: []*models.ClusterNetwork{{}},
+							ServiceNetworks: []*models.ServiceNetwork{},
+							MachineNetworks: []*models.MachineNetwork{},
+						},
+					})
+					verifyApiError(reply, http.StatusBadRequest)
```
YuviGold, line 5050: Check specific errors with `verifyApiErrorString`

internal/common/common.go
```diff
@@ -150,8 +151,37 @@ func GetNetworkCidrAttr(obj interface{}, fieldName string) []*string {
...
+//   that may have been currently configured
+// - `[]*models.MachineNetwork{{}}` - invalid, as it means that we are trying to configure
+//    a single machine network that is empty; a valid network contains at least a CIDR which is
+//    missing in this case
+// - `[]*models.MachineNetwork{{Cidr: ""}}` - invalid, as it means we are trying to configure
+//   a single machine network that has empty CIDR; a valid network should contain a non-empty
+//   CIDR
+// - `[]*models.ClusterNetwork{{HostPrefix: 0}}` - invalid, as it means we are trying to configure
+//   a single cluster network with host prefix with a value 0; this is not a valid subnet lenght
+func IsNetworksSliceNonEmpty(arg interface{}) bool {
```
YuviGold, line 178: I think this function could be simplified to be `IsSliceNonEmpty`
With going over all the fields of the elements and checking in case they are a pointer, validate that they are not nil
pseudo:
```golang
v := reflect.ValueOf(arg)
for i := 0; i < v.NumField(); i++ {
    res = res || v.Kind() == reflect.Ptr && reflectv.Field(i).IsNil()
}

internal/bminventory/inventory_test.go
mkowalski: Done

internal/bminventory/inventory_test.go
mkowalski: Done

internal/bminventory/inventory_test.go
mkowalski: Done

internal/common/common.go
mkowalski: Done, it covers a generic case now

YuviGold (approved):
LGTM

PR #2536, opened 2021-09-01
MGMT-7736: Do not blindly load the whole database in InfraEnv migration

mkowalski:
# Assisted Pull Request

## Description

This PR changes the behaviour of the 20210713123129_populate_infra_env
database migration. What happens currently is that the migration is
using the application code (namely `common.GetClustersFromDBWhere`) and
preloads the following tables without checking whether they are actually
needed

```
var ClusterSubTables = [...]string
{
  HostsTable,
  MonitoredOperatorsTable,
  ClusterNetworksTable,
  ServiceNetworksTable,
  MachineNetworksTable
}
```

This is not a correct behaviour for those reasons

* the application code should not be used inside the migration itself,
  as the migrations should be agnostic to the version that executes it
* only the data actually needed should be loaded, as the initial state
  of the DB at the moment of running the migration may not be the same
  as the initial state of the DB at the moment of writing the code of
  the migration

Please note this PR is a bugfix for enabling the upgrade from ocm-2.3
version to the current master. The correct fix should include replacing
the `common.GetClustersFromDBWhere` call with `tx.Model().Select()`, but
this is outside of the scope of this commit.

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

### Manual test

The following image has been created with the code from this PR

```
quay.io/mkowalski/assisted-service:fixing-some-migration-ordering-blablabla-4
```

AI was deployed from the ocm-2.3 build and the DB schema looked as follows

```
installer=# \d
                List of relations
 Schema |        Name         |   Type   | Owner
--------+---------------------+----------+-------
 public | clusters            | table    | admin
 public | events              | table    | admin
 public | events_id_seq       | sequence | admin
 public | hosts               | table    | admin
 public | migrations          | table    | admin
 public | monitored_operators | table    | admin
(6 rows)
```

AI was upgraded to the current master and after the service started, DB schema looked as follows

```
installer=# \d                                  
                List of relations               
 Schema |        Name         |   Type   | Owner
--------+---------------------+----------+-------
 public | cluster_networks    | table    | admin
 public | clusters            | table    | admin
 public | events              | table    | admin
 public | events_id_seq       | sequence | admin
 public | hosts               | table    | admin
 public | infra_envs          | table    | admin
 public | machine_networks    | table    | admin 
 public | migrations          | table    | admin
 public | monitored_operators | table    | admin
 public | service_networks    | table    | admin
(10 rows)  
```

Test performed together with @pawanpinjarkar and @carbonin on the operator-deployed AI.

## Assignees

/cc @pawanpinjarkar 
/cc @carbonin 
/cc @YuviGold 
/cc @flaper87 

## Checklist

## Reviewers Checklist


mkowalski:
The issue has not been spotted in THE CLOUD because current PROD version deployed in THE CLOUD has the migration `20210713123129_populate_infra_env` already applied.

As the version ocm-2.3 is older, the only migrations applied there are the following ones
```
20201019194303_change_overrides_to_text                                  
20201202140700_change_image_ssh_key_text                                                  
20210218160100_validations_info_to_text                                              
20210223090000_host_validations_info_to_text
```

mkowalski:
/retest

CI seems to be unhappy
```
 INFO[2021-09-01T18:08:51Z] Loading configuration from https://config.ci.openshift.org for openshift/assisted-service@master
ERRO[2021-09-01T18:09:21Z] Failed to load arguments. error=failed to load configuration: got unexpected http 504 status code from configresolver: 
```

docs/dev/migrations.md
```diff
@@ -19,4 +19,12 @@ Migrations are run at startup right after the automigration is run.
...
     - This will give the migration files a proper timestamp as well as (empty) tests
 
-To be run, every migration function should be added to the list in `migrations.all()`
+To be run, every migration function should be added to the list in `postMigrations`
+
+### Pre-migrations
+
+We currently have a single migration in `preMigrations`, i.e. running before `AutoMigrate` step that is responsible for synchronizing the database schema.
+
+It is a bad decision, but given that it has already been rolled out, it would be extremely difficult to move it again to `postMigrations` with the guarantee that no deployment is going to break or end up in a state that is difficult to recover.
```
carbonin, line 28: I think rewording this would be better.

IIUC running this before the automigrate was necessary, so I don't think "bad decision" is fair.
Maybe here is a good place to document the reasoning behind why we had to run this migration before auto migrate even if it isn't exactly the best situation.

From a slack conversation at the time I have this info:
> because the automigrate will fail in this case unless a column will be created and populated first since we added to the model a column that will be part of the PK and that cannot be NULL

Since Ravid left and didn't put any info in the PR or commit can @avishayt or @yevgeny-shnaidman maybe clear up the reasoning for this?

internal/migrations/20210713123129_populate_infra_env.go
```diff
@@ -31,7 +31,7 @@ func populateInfraEnv() *gormigrate.Migration {
 				return err
 			}
 
-			dbClusters, err := common.GetClustersFromDBWhere(tx, common.UseEagerLoading, common.IncludeDeletedRecords)
+			dbClusters, err := common.GetClustersFromDBWhere(tx, common.SkipEagerLoading, common.IncludeDeletedRecords)
```
carbonin, line 34: How much work/time would it take to move this away from using `GetClustersFromDBWhere` entirely?

It's tough to see someone getting back to fixing this "properly" without running into another issue unless you want to do it in a follow up now?

mkowalski: I can't answer that. I have no connection to the origin of this migration and apparently I was not able to find out anyone who has more context than us here after reading through the code.

docs/dev/migrations.md
mkowalski: I completely agree that it was probably the best solution back then (if not the only possible). But as you said, today we don’t seem to have the knowledge on why it’s been done like this and we want to discourage people from doing the same in the future. So yeah, I do hope someone in the team has the knowledge we could write here in the doc

pawanpinjarkar:
Adding my testing results. I tested the operator upgrade manually on top of IPV4 OCP cluster. The AI operator was successfully upgraded and after using the custom service image from this PR, I see no errors while starting the assisted-service pod after the upgrade. I was able to successfully upgrade and install the SNO cluster.

pawanpinjarkar:
This looks good to me but not sure if we want to ahead with this fix. @carbonin

docs/dev/migrations.md
flaper87:
```suggestion
The use of `preMigrations` is (currently) discouraged, as it may cause unexpected failures due to the existing usage of application code in some of the already-released migrations.
```

internal/migrations/20210713123129_populate_infra_env.go
flaper87: I did a quick grep and it looks like we are not using it in many places:

```
internal/bminventory/inventory.go
2999:    dbClusters, err := common.GetClustersFromDBWhere(db, common.UseEagerLoading,

internal/cluster/common_test.go
236:     clusters, err := common.GetClustersFromDBWhere(db, common.UseEagerLoading, true, "")
254:     clusters, err := common.GetClustersFromDBWhere(db, common.UseEagerLoading, true, "")
272:     clusters, err := common.GetClustersFromDBWhere(db, common.UseEagerLoading, true, "")
290:     clusters, err := common.GetClustersFromDBWhere(db, common.SkipEagerLoading, true, "")

internal/common/db.go
203:func GetClustersFromDBWhere(db *gorm.DB, eagerLoading EagerLoadingState, includeDeleted DeleteRecordsState, where ...interface{}) ([]*Cluster, error) {

internal/migrations/20210713123129_populate_infra_env.go
34:          dbClusters, err := common.GetClustersFromDBWhere(tx, common.UseEagerLoading, common.IncludeDeletedRecords)

internal/migrations/20210822134659_multiple_networks.go
12:      dbClusters, err := common.GetClustersFromDBWhere(tx, common.UseEagerLoading, common.IncludeDeletedRecords)
```

We should be able to write an equivalent of this function to be used during migrations. It's important to avoid using any of the internal structures since those may change in the future.

Let's do this in a follow-up PR, tho.

flaper87 (changes requested):
Let's change the wording of the docs, and get this one in.

flaper87 (approved):
Sending through based on @pawanpinjarkar feedback. Given the urgency of this fix, I will submit a fix for the docs in a follow-up PR.

PR #2546, opened 2021-09-02
Add Providers Support - enhancement implementation

Gal-Zaidman:
# Assisted Pull Request

## Description

This PR implements the "Add Providers Support" [enhancement](https://github.com/openshift/assisted-service/pull/2272).
In this PR we:
- Move all related provider logic was added on [MGMT-7067](https://issues.redhat.com/browse/MGMT-7067) to the provider interface.
- Add the provider registry to control the currently used provider and add it to the different related structs.
- Separates the installcfg builder to its own package due to dependency cycle
- Fix existing unit tests.

Questions:

>It would be great

>Yes

>Vsphere and Baremetal

- [X] New Feature
- [X] Bug fix
- [X] Tests
- [ ] Documentation

- [ ] Cloud
- [x] None

## How was this code tested?
- [ ] dev-scripts environment
- [X] Reviewer's test appreciated
- [] No tests needed

## Checklist

## Reviewers Checklist


YuviGold:
Let's get more eyes on this one
/cc @eliorerz 
/cc @michaellevy101 
/cc @filanov

internal/provider/registry.go
```diff
@@ -0,0 +1,91 @@
...
+	Register(provider Provider)
+	// Get returns a provider registered to a name.
+	// if provider is not registered returns an ErrNoSuchProvider
+	Get(name string) (Provider, error)
+	// GetCurrentProvider returns the current provider.
+	// if provider has not been set returns ErrProviderUnSet.
+	GetCurrentProvider() (Provider, error)
+	// SetCurrentProvider sets the current provider to the specific provider.
+	// if provider is not registered returns an ErrNoSuchProvider
+	SetCurrentProvider(name string) error
```
YuviGold: Assisted-service running in parallel with multiple clusters. Each cluster could have a different provider. 
I suggest to remove the `CurrentProvider` functionality although it makes it more simple, we can't have this caching.

internal/provider/base.go
```diff
@@ -30,4 +41,19 @@ func (p *DefaultProvider) Name() models.PlatformType {
 // AddPlatformToInstallConfig adds the current platform to the installconfig.
 func (p *DefaultProvider) AddPlatformToInstallConfig(_ *installcfg.InstallerConfigBaremetal, _ *common.Cluster) error {
 	return nil
-}
\ No newline at end of file
+}
```
YuviGold, line 47: Let's remove the `DefaultProvider` to avoid this error-prone behavior that would occur only in runtime.
I'd prefer to have explicitly every provider implementation (even if it's empty) and to fail on compilation time when something is missing.

internal/provider/base.go
```diff
@@ -25,6 +25,8 @@ type Provider interface {
 	CleanPlatformValuesFromDBUpdates(updates map[string]interface{}) error
 	// SetPlatformUsages uses the usageApi to update platform specific usages
 	SetPlatformUsages(platformParams *models.Platform, usages map[string]models.Usage, usageApi usage.API) error
+	// AreHostsSupported checks if the provider supports the hosts from values discovered in the inventory data
+	AreHostsSupported(hostInventories []*models.Inventory) (bool, error)
```
YuviGold, line 30: 1. Let's have two functions. One for single host validation and one for cluster hosts validations.
I think a validation for a single host could be easier to manipulate and run in a loop if necessary instead of just supporting arrays.
2. I think the wrapping models.Host for the Inventory is better since we can add more business logic metadata to the host without it being detected by the agent and as such sent as part of the Inventory model.

YuviGold: Should be merged within the validator https://github.com/openshift/assisted-service/blob/9d64b25d5fcd2c90c360cab28b15b32288124c7e/internal/host/validator.go#L437-L454

internal/host/validator.go
```diff
@@ -318,9 +320,12 @@ func (v *validator) compatibleWithClusterPlatform(c *validationContext) Validati
 	if c.inventory == nil || c.cluster.Platform.Type == "" {
 		return ValidationPending
 	}
-	hostAvailablePlatforms := common.GetSupportedPlatformsFromInventory(*c.inventory)
-
-	for _, platform := range *hostAvailablePlatforms {
+	// TODO: check if we can return an error with info here
+	hostAvailablePlatforms, err := v.providerRegistry.GetSupportedProvidersByHosts([]*models.Host{c.host})
```
YuviGold, line 324: `IsHostSupportedForProvider`

internal/bminventory/inventory_test.go
```diff
@@ -10414,17 +10413,18 @@ func createInventory(db *gorm.DB, cfg Config) *bareMetalInventory {
 	mockCRDUtils = NewMockCRDUtils(ctrl)
 	mockOperatorManager = operators.NewMockAPI(ctrl)
 	mockIgnitionBuilder = ignition.NewMockIgnitionBuilder(ctrl)
-	mockProviderRegistry := registry.NewMockProviderRegistry(ctrl)
 	mockInstallConfigBuilder = installcfg.NewMockInstallConfigBuilder(ctrl)
 	mockHwValidator = hardware.NewMockValidator(ctrl)
 	mockStaticNetworkConfig = staticnetworkconfig.NewMockStaticNetworkConfig(ctrl)
 	dnsApi := dns.NewDNSHandler(cfg.BaseDNSDomains, common.GetTestLog())
 	gcConfig := garbagecollector.Config{DeregisterInactiveAfter: 20 * 24 * time.Hour}
+	pr := registry.InitProviderRegistry(common.GetTestLog())
```
YuviGold, line 10333: The way we test `bminventory` is using mocks for UT. When we want to test the integration between the components we do it in the subsystem tests.

YuviGold (commented):
Please add subsystem tests for the providers.
Please squash the commits and organize them for <5 commits. Something like (1) Basic infrastructure (2) Implementation (3) host/cluster validations (4) extra tests (5) generations

internal/provider/base.go
Gal-Zaidman: Done

internal/provider/registry.go
Gal-Zaidman: done

internal/provider/base.go
Gal-Zaidman: Done,

Regarding the isValidPlatform function I prefer to do it on a separate PR, so I opened `OCPRHV-622` to track it

internal/host/validator.go
Gal-Zaidman: done

internal/provider/vsphere/consts.go
```diff
@@ -0,0 +1,23 @@
...
+const (
+	PhCluster          = "clusterplaceholder"
+	PhVcenter          = "vcenterplaceholder"
+	PhNetwork          = "networkplaceholder"
+	PhDefaultDatastore = "defaultdatastoreplaceholder"
+	PhUsername         = "usernameplaceholder"
+	PhPassword         = "passwordplaceholder"
+	PhDatacenter       = "datacenterplaceholder"
+
+	DbFieldVsphere          = "platform_vsphere_username"
```
mmartinv, line 12: This field is wrong and probably not needed

internal/provider/vsphere/inventory.go
```diff
@@ -0,0 +1,49 @@
+package vsphere
+
+import (
+	"github.com/openshift/assisted-service/internal/usage"
+	"github.com/openshift/assisted-service/models"
+)
+
+func (p *vsphereProvider) CleanPlatformValuesFromDBUpdates(updates map[string]interface{}) error {
+	updates[DbFieldVsphere] = nil
```
mmartinv, line 9: As mentioned before, "DbFieldVsphere" is wrong and probably not needed. However, If we reset the vsphere database fields we might need to update "DbFieldPlatformType" to the default value, which should be "models.PlatformTypeBaremetal"

internal/provider/vsphere/consts.go
Gal-Zaidman: This is part of the initial vsphere implementation, I didn't want to touch any of the original logic (at least to try and change as little as I can).
If you feel this should be removed then please create a followup pr that removes this field and it can be discussed there

internal/provider/vsphere/inventory.go
Gal-Zaidman: This is why its best to save it for a follow up PR as commented, all that discussion should be made on a dedicated PR for your proposed change

YuviGold (approved):
LGTM

PR #2569, opened 2021-09-05
MGMT-7726: Handle non existing cluster in Agent controller

rollandf:
# Assisted Pull Request

## Description
Before this patch, if the referenced ClusterDeployment in Agent was not found in DB or kubernetes, the Agent CR was deleted.

With this patch:
- In case that the referenced cluster is not found in DB, update SpecSync condition with Backend Error, retry
- In case that the referenced cluster is not found in kubernetes, update SpecSync condition with Input Error, no retry
 
- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @nmagnezi 
/cc @yevgeny-shnaidman 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -154,13 +154,6 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 
 		// Retrieve clusterDeployment
 		if err = r.Get(ctx, kubeKey, clusterDeployment); err != nil {
-			if k8serrors.IsNotFound(err) {
-				// Delete the agent, using a finalizer with pre-delete to deregister the host.
-				log.Infof("Cluster Deployment name: %s namespace: %s not found, deleting Agent",
-					agent.Spec.ClusterDeploymentName.Name, agent.Spec.ClusterDeploymentName.Namespace)
-				return r.deleteAgent(ctx, log, req.NamespacedName)
-			}
-
 			errMsg := fmt.Sprintf("failed to get clusterDeployment with name %s in namespace %s",
 				agent.Spec.ClusterDeploymentName.Name, agent.Spec.ClusterDeploymentName.Namespace)
 			log.WithError(err).Error(errMsg)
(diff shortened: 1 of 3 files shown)


internal/controller/controllers/agent_controller.go
```diff
@@ -154,13 +154,6 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 
 		// Retrieve clusterDeployment
 		if err = r.Get(ctx, kubeKey, clusterDeployment); err != nil {
-			if k8serrors.IsNotFound(err) {
-				// Delete the agent, using a finalizer with pre-delete to deregister the host.
-				log.Infof("Cluster Deployment name: %s namespace: %s not found, deleting Agent",
-					agent.Spec.ClusterDeploymentName.Name, agent.Spec.ClusterDeploymentName.Namespace)
-				return r.deleteAgent(ctx, log, req.NamespacedName)
```
flaper87, line 161: Are agents for non-existing clusters not going to be deleted?

internal/controller/controllers/agent_controller.go
```diff
@@ -174,14 +167,10 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 			// Retrieve cluster by ClusterDeploymentName from the database
 			cluster, err2 := r.Installer.GetClusterByKubeKey(kubeKey)
 			if err2 != nil {
-				if errors.Is(err2, gorm.ErrRecordNotFound) {
-					// Delete the agent, using a finalizer with pre-delete to deregister the host.
-					log.Infof("Cluster name: %s namespace: %s not found in backend, deleting Agent",
-						agent.Spec.ClusterDeploymentName.Name, agent.Spec.ClusterDeploymentName.Namespace)
-					return r.deleteAgent(ctx, log, req.NamespacedName)
-				}
+				log.WithError(err2).Errorf("Fail to get cluster name: %s namespace: %s in backend",
```
yevgeny-shnaidman, line 170: just wondering: backend is doing all the same validations before binding: checks that cluster exists, checks if it is in appropriate state, etc. What if we just call BindHost in case h.ClusterID == nil , and handle the return error as we handle it right now?

internal/controller/controllers/agent_controller.go
filanov: Maybe we need to change the condition to delete them if infra env is deleted.

rollandf: There is another Jira ticket for this flow : https://issues.redhat.com/browse/MGMT-7758

internal/controller/controllers/agent_controller_test.go
```diff
@@ -445,6 +449,43 @@ var _ = Describe("agent reconcile", func() {
 		Expect(conditionsv1.FindStatusCondition(agent.Status.Conditions, v1beta1.SpecSyncedCondition).Status).To(Equal(corev1.ConditionTrue))
 	})
 
+	It("Agent bind, cluster not found in DB", func() {
```
nmagnezi, line 452: I would also add a test that after a retry cluster is found in DB, so we can be sure it recovers as we expected it to and update conditions accordingly.

nmagnezi (changes requested):
Looks good overall.
Added a comment related to tests.

internal/controller/controllers/agent_controller_test.go
rollandf: Done

internal/controller/controllers/agent_controller.go
rollandf: I guess, it could work. You want me to change it?

yevgeny-shnaidman: yes, let's do it.  no need to do the same validation twice

rollandf: Second thoughts on it. I don't have the Cluster ID so I need to retrieve the cluster by kube key first.

yevgeny-shnaidman: you are right, scratch that

nmagnezi (approved):
/lgtm

internal/controller/controllers/agent_controller_test.go
nmagnezi: 👍

PR #2577, opened 2021-09-08
MGMT-7282: Operator upgrade CI job

pawanpinjarkar:
# Assisted Pull Request

## Description

For the operator upgrade CI job, we need to set some env vars during the initial part of the upgrade and later change the values after the upgrade. Setting the values in the scripts with this PR seems the easiest and most straightforward option for CI use. Also, it can be used for local testing.
 
This PR adds new variables to be used by the operator upgrade CI job.

- `ASSISTED_UPGRADE_OPERATOR` -  This is useful to determine which `ASSISTED_OPENSHIFT_VERSION` to be used. Before upgrading the operator, the value of this variable is `true`, which signals that the current operator needs to be updated. The value is derived from the default_ocp_versions.json. When set to true, `ASSISTED_OPENSHIFT_VERSION` gets the value `4.8`. Otherwise, `4.9`. After upgrading the operator, the value of this variable is `false`.

- `ASSISTED_STOP_AFTER_AGENT_DISCOVERY` - This is useful when waiting for a full cluster deployment is not needed. For example, during an upgrade CI job, we may not need to wait for the full deployment in some of the phases of the job. Before upgrading the operator, the value of this variable is `false`, which preserves the existing behavior.
After upgrading the operator, the value of this variable is `true`.

- `ASSISTED_CLUSTER_NAME`, `ASSISTED_CLUSTER_DEPLOYMENT_NAME`, `ASSISTED_AGENT_CLUSTER_INSTALL_NAME`, `ASSISTED_INFRAENV_NAME` - These are used for the upgraded cluster.

Consumed by: https://github.com/openshift/release/pull/21539

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [X] CI/CD 

- [ ] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
- [] No tests needed

## Assignees

/cc @lranjbar 
/cc @flaper87 

## Checklist

## Reviewers Checklist


deploy/operator/after_upgrade.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+export UPGRADE_OPERATOR="false"
```
lranjbar, line 2: I'm pretty bad at naming, but I'm wondering if this is too generic. Maybe `IS_OPERATOR_UPGRADE_FLOW` or `UPGRADE_ASSISTED_OPERATOR`? Since this is used to denote select which OCP version to start with.

Just trying to think about reading this. 🤔

deploy/operator/before_upgrade.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+export SERVICE_IMAGE=""
+export INDEX_IMAGE="quay.io/ocpmetal/assisted-service-index:ocm-2.3"
```
lranjbar, line 3: Where do we set the `INDEX_IMAGE` to be the newly built operator? I didn't see another value in `after_upgrade.sh`

lranjbar (commented):
This looks mostly right to me, just a couple of questions.

deploy/operator/before_upgrade.sh
pawanpinjarkar: There are 3 phases in the upgrading job, before the upgrade, upgrade and after the upgrade. So, when `baremetalds-assisted-operator-upgrade-ref.yaml` calls  `baremetalds-assisted-operator-upgrade-commands.sh`, I am setting the latest index image and patching the catalogsource with that image.
https://github.com/openshift/release/pull/21539/files#diff-27e52a9f8e18f6f4a309185c25377a2839112270b55e25429bca81156182bd51R35

deploy/operator/after_upgrade.sh
pawanpinjarkar: `UPGRADE_ASSISTED_OPERATOR` sounds good.

deploy/operator/common.sh
```diff
@@ -25,14 +25,21 @@ export HIVE_NAMESPACE="${HIVE_NAMESPACE:-hive}"
 # Versions #
 ############
 
+export UPGRADE_ASSISTED_OPERATOR="${UPGRADE_ASSISTED_OPERATOR:-false}"
 export OPENSHIFT_VERSIONS="${OPENSHIFT_VERSIONS:-$(cat ${__root}/data/default_ocp_versions.json)}"
 
 if [ -z "${ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE:-}" ]; then
-    export ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE=$(echo ${OPENSHIFT_VERSIONS} | jq -rc '[.[].release_image]|max')
+    if [ "${UPGRADE_ASSISTED_OPERATOR}" = "true" ]; then
+        export ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE=$(echo ${OPENSHIFT_VERSIONS} | jq -rc '[.[].release_image][-2]')
```
YuviGold, line 33: why `-2`? I don't understand these code section. In any case, we must not depend on indexing that might get change from time to time. (e.g. 4.6 and 4.7 will soon be removed)

deploy/operator/after_upgrade.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+export UPGRADE_ASSISTED_OPERATOR="false"
```
YuviGold, line 2: Rename to be aligned with the same naming convention
```suggestion
export ASSISTED_UPGRADE_OPERATOR="false"
```

deploy/operator/after_upgrade.sh
```diff
@@ -0,0 +1,7 @@
...
+export ASSISTED_STOP_AFTER_AGENT_DISCOVERY="true"
```
YuviGold, line 3: I think that it would be more intuitive if we will inverse it. Similar to what we have in assisted-test-infra. There's `deploy_nodes` and `deploy_nodes_with_install`
```suggestion
export ASSISTED_WAIT_FOR_INSTALLATION="false"
```
WDYT? @flaper87

deploy/operator/before_upgrade.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
YuviGold, line 1: The logic of the upgrade process shouldn't be split between repositories. I can see `before_upgrade.sh` and `after_upgrade.sh` here but not their usage.

Please move https://github.com/openshift/release/pull/21539/files#diff-27e52a9f8e18f6f4a309185c25377a2839112270b55e25429bca81156182bd51R20-R36 logic to be a `./deploy/operator/upgrade.sh` script maintained under the assisted-service repository. This way it would be accessible locally as well.

deploy/operator/before_upgrade.sh
lranjbar: Ah right it is more clear seeing it in code than talking through it. Okay so we will need to ability to override these exports therefore we will need to pass in the variables like we do in other places like:

`export INDEX_IMAGE=$INDEX_IMAGE-:quay.io/ocpmetal/assisted-service-index:ocm-2.3`

But for all of the exports to be safe

deploy/operator/common.sh
lranjbar: Before the upgrade we are looking to install the version prior to the most current one of OCP. The most current version of OCP we are installing is 4.9, and the version previous to that is 4.8.

TLDR of the current flow is:
- Install hub cluster
- Install operator
- Install OCP 4.8 SNO spoke cluster
- Update operator
- Install OCP 4.9 SNO spoke cluster

pawanpinjarkar: Thank you, Lisa. I can add comments explaining the logic but I don't see any other alternative atm.

deploy/operator/after_upgrade.sh
pawanpinjarkar: Done

deploy/operator/common.sh
pawanpinjarkar: Added comments in the script explaining the need and logic.

deploy/operator/before_upgrade.sh
pawanpinjarkar: Done. Moved to `./deploy/operator/upgrade.sh`. `before_upgrade.sh` and `after_upgrade.sh` are to be used by CI job. My priority is to get the job ready before feature freeze and then focus on running the upgrade job locally.

deploy/operator/common.sh
YuviGold: Thank you for the clarification

deploy/operator/before_upgrade.sh
YuviGold: Great.
Now looking at the CI job, I don't understand why do we need 3 different steps. I'd appreciate if also `after_upgrade` step could be as part of the `upgrade` script.
https://github.com/openshift/release/pull/21539/files#diff-10db92a2ad655eae770c3c0383bf65fa41baf1fe82a8a388e00ea02634e02199R53-R57

The same goes for ZTP. We verify our installation stages as part of the ZTP script.
Then, all the variables will be assigned inside the script and the `after_upgrade.sh` and `before_upgrade.sh` won't be necessairy. In case someone would want to override them - then it should be possible.

pawanpinjarkar: I suggest making these code style changes later and get the CI job-ready first. I will create a follow-up PR later.

lranjbar:
There has been a decent amount of churn on this PR and it is in a working state so I'm ready to merge it to make it easier to test in `openshift/release`

The following need to be addressed in a follow up PRs:
- How are we piping in the openshift releases into `ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE`? I think this stanza needs to be tested: https://github.com/openshift/release/pull/21539/files#diff-b4ec9980e779ed206c8f8c154f750816537b6dde361289606477872c7fd319aaR27-R28 I'm not entirely convinced we wont need `ASSISTED_OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` but also I believe it will be easier to test this after this PR merges.
- Reusing the same worker node vs having two worker nodes this can be moved to another JIRA issue as an enhancement to the upgrade job.

PR #2587, opened 2021-09-11
MGMT-4475: validations to consider effective role

slaviered:
# Assisted Pull Request

## Description

  - role based validations now takes into account the
     suggested role in case of auto-assign. This results
     in timely failures in case the role to be assigned
     is incompatible with the validation's logic

  - role evaluation is now being done in its own function
    within the monitoring instead of in RefreshStatus

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -100,6 +100,7 @@ type LogTimeoutConfig struct {
 type Config struct {
 	LogTimeoutConfig
 	EnableAutoReset         bool                    `envconfig:"ENABLE_AUTO_RESET" default:"false"`
+	EnableAutoAssign        bool                    `envconfig:"ENABLE_AUTO_ASSIGN" default:"true"`
 	ResetTimeout            time.Duration           `envconfig:"RESET_CLUSTER_TIMEOUT" default:"3m"`
 	MonitorBatchSize        int                     `envconfig:"HOST_MONITOR_BATCH_SIZE" default:"100"`
 	DisabledHostvalidations DisabledHostValidations `envconfig:"DISABLED_HOST_VALIDATIONS" default:""` // Which host validations to disable (should not run in preprocess)
@@ -134,6 +135,7 @@ type API interface {
 	IsInstallable(h *models.Host) bool
 	// auto assign host role
 	AutoAssignRole(ctx context.Context, h *models.Host, db *gorm.DB) (bool, error)
+	RefreshRole(ctx context.Context, h *models.Host, db *gorm.DB) error
 	IsValidMasterCandidate(h *models.Host, c *common.Cluster, db *gorm.DB, log logrus.FieldLogger) (bool, error)
 	SetUploadLogsAt(ctx context.Context, h *models.Host, db *gorm.DB) error
 	UpdateLogsProgress(ctx context.Context, h *models.Host, progress string) error
@@ -375,6 +377,28 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster, 
 	}).Error
 }
 
+func (m *Manager) refreshRoleInternal(ctx context.Context, h *models.Host, db *gorm.DB, forceRefresh bool) error {
+	//update suggested role, if not yet set
+	var suggestedRole models.HostRole
+	var err error
+	if m.Config.EnableAutoAssign || forceRefresh {
+		//because of possible hw changes, suggested role should be calculated
+		//periodically even if the suggested role is already set
+		if h.Role == models.HostRoleAutoAssign &&
+			funk.ContainsString(hostStatusesBeforeInstallation[:], *h.Status) {
+			if suggestedRole, err = m.autoRoleSelection(ctx, h, db); err == nil {
+				if h.SuggestedRole != suggestedRole {
+					if err = updateRole(m.log, h, h.Role, suggestedRole, db, string(h.Role)); err == nil {
+						h.SuggestedRole = suggestedRole
+						m.log.Infof("suggested role for host %s is %s", *h.ID, suggestedRole)
+					}
+				}
+			}
+		}
+	}
+	return err
+}
+
 func (m *Manager) refreshStatusInternal(ctx context.Context, h *models.Host, c *common.Cluster, i *common.InfraEnv, db *gorm.DB) error {
 	if db == nil {
 		db = m.db
@@ -409,15 +433,6 @@ func (m *Manager) refreshStatusInternal(ctx context.Context, h *models.Host, c *
 		}
 	}
 
-	//update suggested role, if not yet set
-	var suggestedRole models.HostRole
-	if h.Role == models.HostRoleAutoAssign && h.SuggestedRole == models.HostRoleAutoAssign &&
-		funk.ContainsString(hostStatusesBeforeInstallation[:], *h.Status) {
-		if suggestedRole, err = m.autoRoleSelection(ctx, h, db); err == nil {
-			_ = updateRole(m.log, h, h.Role, suggestedRole, db, string(h.Role))
-		}
-	}
-
 	err = m.sm.Run(TransitionTypeRefresh, newStateHost(h), &TransitionArgsRefreshHost{
 		ctx:               ctx,
 		db:                db,
@@ -429,6 +444,13 @@ func (m *Manager) refreshStatusInternal(ctx context.Context, h *models.Host, c *
 		return common.NewApiError(http.StatusConflict, err)
 	}
 	return nil
+}
+
+func (m *Manager) RefreshRole(ctx context.Context, h *models.Host, db *gorm.DB) error {
+	if db == nil {
+		db = m.db
+	}
+	return m.refreshRoleInternal(ctx, h, db, true)
 }
 
 func (m *Manager) RefreshStatus(ctx context.Context, h *models.Host, db *gorm.DB) error {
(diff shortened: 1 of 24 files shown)


internal/host/validator.go
```diff
@@ -343,38 +343,38 @@ func (v *validator) printCompatibleWithClusterPlatform(c *validationContext, sta
...
 
-func isDiskEncryptionRelevantForHost(c *validationContext) bool {
-	switch *c.cluster.DiskEncryption.EnableOn {
+func isDiskEncryptionEnabledForRole(encryption models.DiskEncryption, role models.HostRole) bool {
+	switch swag.StringValue(encryption.EnableOn) {
 	case models.DiskEncryptionEnableOnAll:
 		return true
 	case models.DiskEncryptionEnableOnMasters:
-		return c.host.Role == models.HostRoleMaster || c.host.Role == models.HostRoleBootstrap || c.host.Role == models.HostRoleAutoAssign
+		return role == models.HostRoleMaster || role == models.HostRoleBootstrap
```
ybettan, line 351: Are we sure we will never reach that code with `h.Role == autoAssigned`?

internal/host/validator.go
```diff
@@ -343,38 +343,38 @@ func (v *validator) printCompatibleWithClusterPlatform(c *validationContext, sta
...
 
-func isAutoAassignAllowed(c *validationContext) bool {
-	return *c.cluster.DiskEncryption.EnableOn == models.DiskEncryptionEnableOnAll ||
-		*c.cluster.DiskEncryption.EnableOn == models.DiskEncryptionEnableOnNone
-}
-
 func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) ValidationStatus {
 
-	if c.infraEnv != nil || reflect.DeepEqual(c.cluster.DiskEncryption, &models.DiskEncryption{}) || !isDiskEncryptionRelevantForHost(c) {
+	if c.infraEnv != nil || reflect.DeepEqual(c.cluster.DiskEncryption, &models.DiskEncryption{}) {
```
ybettan, line 361: Same question

internal/host/validator.go
```diff
@@ -343,38 +343,38 @@ func (v *validator) printCompatibleWithClusterPlatform(c *validationContext, sta
...
 	}
 
-	if c.host.Role == models.HostRoleAutoAssign && !isAutoAassignAllowed(c) {
-		return ValidationFailure
+	if c.inventory == nil {
+		return ValidationPending
 	}
 
-	if c.inventory == nil {
+	role := common.GetEffectiveRole(c.host)
```
ybettan, line 369: Why don't you check directly `c.host.Role` ?

internal/host/validator.go
```diff
@@ -343,38 +343,38 @@ func (v *validator) printCompatibleWithClusterPlatform(c *validationContext, sta
...
+	if role == models.HostRoleAutoAssign {
```
ybettan, line 370: Please merge this condition with the previous one, if both cases the result is the same.

internal/host/validator.go
```diff
@@ -387,17 +387,14 @@ func (v *validator) printDiskEncryptionRequirementsSatisfied(c *validationContex
...
-		}
 		if c.inventory.TpmVersion == models.InventoryTpmVersionNone {
 			return "TPM version could not be found, make sure TPM is enalbed in host's BIOS"
 		} else {
 			return fmt.Sprintf("The host's TPM version is not supported, expected-version: %s, actual-version: %s",
 				models.InventoryTpmVersionNr20, c.inventory.TpmVersion)
 		}
 	case ValidationPending:
-		return "Missing host inventory"
+		return "Missing host inventory or role assignment"
```
ybettan, line 397: Please split the return value to be specific for each case.

internal/host/validator.go
slaviered: Because it has to take into account suggested role as well

internal/host/validator.go
slaviered: I wished to stress the condition of auto-assign

internal/host/validator.go
slaviered: Done

internal/operators/cnv/cnv_operator.go
```diff
@@ -91,10 +91,12 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
 		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{"CPU does not have virtualization support "}}, nil
 	}
 
+	role := common.GetEffectiveRole(host)
 	// If the Role is set to Auto-assign for a host, it is not possible to determine whether the node will end up as a master or worker node.
-	if host.Role == models.HostRoleAutoAssign {
-		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{"All host roles must be assigned to enable CNV"}}, nil
+	if role == models.HostRoleAutoAssign {
```
pkliczewski, line 96: Please add a test for changes in this file.

pkliczewski:
@rewantsoni please review

internal/operators/cnv/cnv_operator.go
slaviered: There was already test in place for this line and I changed the expected value to Pending. Semantically, failure and pending are the same BTW so overall the behavior did not changed

internal/operators/ocs/ocs_operator.go
```diff
@@ -93,23 +93,26 @@ func (o *operator) ValidateHost(_ context.Context, cluster *common.Cluster, host
...
 
+	role := common.GetEffectiveRole(host)
+
+	// auto-assign effective role (meaning both role and suggested-role) should be a transient
+	// state. return pending status in this case
+	if role == models.HostRoleAutoAssign {
+		status := "Host role must be assigned or evaluated to master or worker."
+		return api.ValidationResult{Status: api.Pending, ValidationId: o.GetHostValidationID(), Reasons: []string{status}}, nil
+	}
+
```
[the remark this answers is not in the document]
slaviered: So basically leave I should leave the logic as it was before with the distinction between compact mode and standard mode right?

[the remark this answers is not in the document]
rewantsoni: Yes, either that or if you are changing it here to show validation pending in case of auto-assign then even in ValidateCluster in case of auto-assign role(both in mode), we should show validation pending

internal/host/host.go
```diff
@@ -375,6 +377,28 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
 	}).Error
 }
 
+func (m *Manager) refreshRoleInternal(ctx context.Context, h *models.Host, db *gorm.DB) error {
+	//update suggested role, if not yet set
+	var suggestedRole models.HostRole
+	var err error
+	if m.Config.EnableAutoAssign {
```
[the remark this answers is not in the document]
slaviered: If the suggested role is not calculated, GetEffectiveRole returns h.Role so no need to reference it there

internal/host/host.go
```diff
@@ -1009,7 +1031,7 @@ func (m *Manager) AutoAssignRole(ctx context.Context, h *models.Host, db *gorm.D
 	if h.Role == models.HostRoleAutoAssign {
 		log := logutil.FromContext(ctx, m.log)
 		// If role is auto-assigned calculate the suggested roles
-		// This logic will moved to the monitor soon
+		// to make sure the suggestion is fresh
```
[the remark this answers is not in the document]
slaviered: In this case we calculate it one more time before installation, if the calculation was missed in the monitor. Here we do not need the AUTO_ASSIGN flag because it is the legacy place and code for such calculation. When the flag is set to false, there will be no calculation in the monitoring and only here we will decide on the actual role

internal/host/validator.go
```diff
@@ -121,11 +121,11 @@ func (v *validator) getBootDeviceInfo(host *models.Host) (*models.DiskInfo, erro
 }
 
 func (c *validationContext) validateRole() error {
-	switch c.host.Role {
+	switch common.GetEffectiveRole(c.host) {
```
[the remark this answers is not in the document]
slaviered: There are edge cases where suggested role is not calculated, or if validateRole is called before the second cycle of monitoring (very unlikely but still possible). So auto-assign is still 'valid' role. The purpose of this function really is to check that there are any role at all. If we have a problem with the inventory the role is empty

internal/host/validator.go
```diff
@@ -389,17 +389,17 @@ func (v *validator) printDiskEncryptionRequirementsSatisfied(c *validationContex
...
 		} else {
 			return fmt.Sprintf("The host's TPM version is not supported, expected-version: %s, actual-version: %s",
 				models.InventoryTpmVersionNr20, c.inventory.TpmVersion)
 		}
 	case ValidationPending:
-		return "Missing host inventory"
+		if c.inventory == nil {
+			return "Missing host inventory"
+		}
+		return "Missing role assignment"
```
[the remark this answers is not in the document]
slaviered: Correct. I will fix it

internal/host/monitor.go
```diff
@@ -80,6 +80,13 @@ func (m *Manager) clusterHostMonitoring() int64 {
 					if err != nil {
 						log.WithError(err).Errorf("failed to refresh host %s state", *host.ID)
 					}
+					//the refreshed role will be taken into account
+					//on the next monitor cycle. The force flag is a workaround
+					//until the feature flag is removed
+					err = m.refreshRoleInternal(ctx, host, m.db, false)
```
ori-amizur, line 86: I think the refreshRole has to be before the refresh status.  Otherwise the status will be changed as a result only in the next cycle

ori-amizur: OK

PR #2619, opened 2021-09-14
MGMT-7804: Create healthchecks for monitoring routines

eranco74:
# Assisted Pull Request

## Description
The assisted-installer cluster-monitor seems to hang for hours from time to time, whenever it hangs clusters dont' get updated by the monitor.
This cause few issues, that lead to installation failures (mainly we've seen clusters stuck in timeout while preparing for installaiton until they timeout).

Adding last run timestamp to each thread object that will run monitor.
Updated liveness probe to watch the last run timestamp with 5 minutes timeout
Adding logs around cluster monitor in order to see what takes so long.

- Is this PR relying on CI for an e2e test run? For regression

This PR was tested using this [branch](https://github.com/eranco74/assisted-service/tree/malformed_monitor) in subsystem with a malformed monitor (that sleeps) 
The liveness prob failed and the pod got restarted.
-->

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @filanov 
/cc @tsorya 

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/app/middleware.go
+++ b/pkg/app/middleware.go
@@ -4,8 +4,10 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/openshift/assisted-service/pkg/thread"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rs/cors"
+	"github.com/sirupsen/logrus"
 )
 
 // WithMetricsResponderMiddleware Returns middleware which responds to /metrics endpoint with the prometheus metrics
@@ -21,10 +23,19 @@ func WithMetricsResponderMiddleware(next http.Handler) http.Handler {
 }
 
 // WithHealthMiddleware returns middleware which responds to the /health endpoint
-func WithHealthMiddleware(next http.Handler) http.Handler {
+func WithHealthMiddleware(next http.Handler, threads []*thread.Thread, logger logrus.FieldLogger, timeout time.Duration) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.Method == http.MethodGet && r.URL.Path == "/health" {
-			w.WriteHeader(http.StatusOK)
+			status := http.StatusOK
+			for _, th := range threads {
+				if time.Since(th.LastRunStartedAt()) > timeout {
+					logger.Errorf("thread %s live probe validation failed, last run timestamp "+
+						"is %v, current time %s", th.Name(), th.LastRunStartedAt(), time.Now())
+					status = http.StatusInternalServerError
+					break
+				}
+			}
+			w.WriteHeader(status)
 			return
 		}
 		next.ServeHTTP(w, r)
--- a/pkg/thread/thread.go
+++ b/pkg/thread/thread.go
@@ -18,11 +18,12 @@ import (
 //    ....
 //
 type Thread struct {
-	log      logrus.FieldLogger
-	exec     func()
-	done     chan struct{}
-	name     string
-	interval time.Duration
+	log              logrus.FieldLogger
+	exec             func()
+	done             chan struct{}
+	name             string
+	interval         time.Duration
+	lastRunStartedAt time.Time
 }
 
 func New(log logrus.FieldLogger, name string, interval time.Duration, exec func()) *Thread {
@@ -38,6 +39,7 @@ func New(log logrus.FieldLogger, name string, interval time.Duration, exec func(
 // Start thread
 func (t *Thread) Start() {
 	t.log.Infof("Started %s", t.name)
+	t.lastRunStartedAt = time.Now()
 	go t.loop()
 }
 
@@ -47,6 +49,14 @@ func (t *Thread) Stop() {
 	t.done <- struct{}{}
 	<-t.done
 	t.log.Infof("Stopped %s", t.name)
+}
+
+func (t *Thread) LastRunStartedAt() time.Time {
+	return t.lastRunStartedAt
+}
+
+func (t *Thread) Name() string {
+	return t.name
 }
 
 func (t *Thread) loop() {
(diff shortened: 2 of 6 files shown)


pkg/app/middleware.go
```diff
@@ -21,10 +23,19 @@ func WithMetricsResponderMiddleware(next http.Handler) http.Handler {
...
 
 // WithHealthMiddleware returns middleware which responds to the /health endpoint
-func WithHealthMiddleware(next http.Handler) http.Handler {
+func WithHealthMiddleware(next http.Handler, threads []*thread.Thread, logger logrus.FieldLogger) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.Method == http.MethodGet && r.URL.Path == "/health" {
-			w.WriteHeader(http.StatusOK)
+			status := http.StatusOK
+			for _, th := range threads {
+				if th.LastRunTimestamp().IsZero() && time.Since(th.LastRunTimestamp()).Minutes() > 5 {
```
filanov, line 31: lets have the timeout configurable to avoid pushing another hotfix in case we will need to adjust it.

filanov: Please don't forget to add the env var into https://github.com/openshift/assisted-service/blob/master/openshift/template.yaml

pkg/app/middleware.go
```diff
@@ -21,10 +23,19 @@ func WithMetricsResponderMiddleware(next http.Handler) http.Handler {
...
 
 // WithHealthMiddleware returns middleware which responds to the /health endpoint
-func WithHealthMiddleware(next http.Handler) http.Handler {
+func WithHealthMiddleware(next http.Handler, threads []*thread.Thread, logger logrus.FieldLogger, timeout time.Duration) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.Method == http.MethodGet && r.URL.Path == "/health" {
-			w.WriteHeader(http.StatusOK)
+			status := http.StatusOK
+			for _, th := range threads {
+				if th.LastRunTimestamp().IsZero() && time.Since(th.LastRunTimestamp()) > timeout {
```
filanov, line 31: Should it be or?

ybettan:
Can we test using subsystem?

pkg/app/middleware.go
filanov: Actually it doesn't matter. if it's not zero then we have something and the second part will do the work

filanov: OK no it should be OR

filanov: https://play.golang.org/p/U37w8T5p2lb

eranco74: Should be 
`if !th.LastRunTimestamp().IsZero() && time.Since(th.LastRunTimestamp()) > timeout`
Fixed

pkg/thread/thread.go
```diff
@@ -18,11 +18,12 @@ import (
...
-	exec     func()
-	done     chan struct{}
-	name     string
-	interval time.Duration
+	log              logrus.FieldLogger
+	exec             func()
+	done             chan struct{}
+	name             string
+	interval         time.Duration
+	lastRunStartedAt time.Time
```
filanov, line 26: lets init the time in the New Call, we don't know how long the first call will run so we can get into a state that we started the thread the first exec takes 1 minute but the health check will already run and see that the time is zero and reboot the service.

flaper87: We are now setting the `lastRunStartedAt` right before the `exec` so, I think it's ok to skip with the `IsZero` check. Both work, as in we could init it and remove the check but as it is it should have the same effect. Thoughts?

filanov: Actually it should be set in the Start function.

eranco74: I changed the var to lastRunStartedAt due to that we will set it before the operation starts.

filanov: It can create a different bug that if we never start the monitor health will report that everything is ok. 
it's happening right now as well but at least if we already adding something that validate it then we should cover this case as well. 
So even if you set the time at New we will know that something is wrong.

filanov: I think that we should set it in the start just before the `go` call

pkg/app/middleware.go
filanov: A comment related to https://github.com/openshift/assisted-service/pull/2619#discussion_r708166865 
I know that right now it will work but if we already there so maybe we should avoid `IsZero`?

pkg/thread/thread.go
eranco74: Agree, fixed

carbonin (approved):

filanov (approved):

eranco74:
ERROR: Packet setup failed
/test e2e-metal-assisted

eranco74:
> Can we test using subsystem?

Tested it locally with a malformed monitor (that sleeps)
The liveness prob failed and the pod got restarted.

```
time="2021-09-14T12:41:38Z" level=error msg="thread Cluster State Monitor live probe validation failed, last run timestamp is 2021-09-14 12:36:34.453717637 +0000 UTC m=+938.815037264, current time 2021-09-14 12:41:38.484370021 +0000 UTC m=+1242.845689757" func=github.com/openshift/assisted-service/pkg/app.WithHealthMiddleware.func1 file="/go/src/github.com/openshift/origin/pkg/app/middleware.go:32" pkg=healthcheck

Warning  Unhealthy  45s (x3 over 65s)   kubelet            Liveness probe failed: HTTP probe failed with statuscode: 500

assisted-installer   assisted-service-7c69c7f746-59bcd   1/1     Running   1          21m

```

PR #2620, opened 2021-09-14
MGMT-7806: Change host events API in order to solve segmentation fault

yevgeny-shnaidman:
Currently host/cluster events store ClusterID as strfmt.UUID. As a result, APIs
that create events also expect clusterId parameter as strfmt.UUID. In case of
unbound host this may cause issues, since ClusterID will be nil.
We change the host event to store ClusterID as pointer, and propagate it as is
to SendHostEvent* APIs, which already handle the case where it is nil

# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -582,7 +582,7 @@ func (m *Manager) SetBootstrap(ctx context.Context, h *models.Host, isbootstrap 
 		if err != nil {
 			return errors.Wrapf(err, "failed to set bootstrap to host %s", h.ID.String())
 		}
-		eventgen.SendHostSetBootstrapEvent(ctx, m.eventsHandler, *h.ClusterID, *h.ID, h.InfraEnvID, hostutil.GetHostnameForMsg(h))
+		eventgen.SendHostSetBootstrapEvent(ctx, m.eventsHandler, h.ClusterID, *h.ID, h.InfraEnvID, hostutil.GetHostnameForMsg(h))
 	}
 	return nil
 }
@@ -710,7 +710,7 @@ func (m *Manager) UpdateImageStatus(ctx context.Context, h *models.Host, newImag
 				newImageStatus.Time, newImageStatus.SizeBytes, newImageStatus.DownloadRate)
 		}
 
-		eventgen.SendUpdateImageStatusEvent(ctx, m.eventsHandler, *h.ClusterID, *h.ID, h.InfraEnvID,
+		eventgen.SendUpdateImageStatusEvent(ctx, m.eventsHandler, h.ClusterID, *h.ID, h.InfraEnvID,
 			hostutil.GetHostnameForMsg(h), newImageStatus.Name, string(newImageStatus.Result), eventInfo)
 	}
 	marshalledStatuses, err := common.MarshalImageStatuses(hostImageStatuses)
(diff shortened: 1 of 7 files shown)


tools/generate_events.py
```diff
@@ -96,7 +96,7 @@
 }
 {%- else %}
 func (e *{{eventName}}) GetClusterId() *strfmt.UUID {
-    {% if event.properties.cluster_id -%}     return &e.ClusterId
+    {% if event.properties.cluster_id -%}     return e.ClusterId
```
masayag, line 99: Could you describe the issues that occurred in the unbound host ? 
If it reduced only to invocation of this method, perhaps it can be changed to
```
    {% if event.properties.cluster_id -%}     if len(e.ClusterId) > 0 {
        return &e.ClusterId
    } else {
      return nil
    }
```
to be nil-safe.

If the problem with with the call to the eventshandler.SendHostEvent(...), then there is a need to change their signatures to support nil values as cluster-id.

masayag:
pls fix typo (Chahe) in PR message and in the commit message as well.

internal/host/host.go
```diff
@@ -830,10 +830,10 @@ func (m *Manager) ResetHost(ctx context.Context, h *models.Host, reason string,
 	defer func() {
 		if shouldAddEvent {
 			if isFailed {
-				eventgen.SendHostResetInstallationFailedEvent(ctx, m.eventsHandler, h.InfraEnvID, *h.ID,
+				eventgen.SendHostResetInstallationFailedEvent(ctx, m.eventsHandler, h.ClusterID, *h.ID,
```
nmagnezi, line 833: good catch

tools/generate_events.py
yevgeny-shnaidman: The specific issue was with the call to the SendHostValidationFallingEvent function (but it may happen to other calls for host events). in previous implementation SendHostValidationFallingEvent received clusterId as strfmt.UUID. in case of the unbound host, clusterId is nil, so calling SendHostValidationFallingEvent (*host.ClusterID) caused segmentation fault. I think it is better simply to store ClusterID as pointer in the host events, since we already have a code that checks if it is nil and calls AddEvent function with appropriate params

masayag: Okay, in that case @nmagnezi we'll have to use the practice for the infra-env event type which may have the cluster-id optionals in that context

nmagnezi (changes requested):
looks good overall, but this fixes an issue that we didn't have tests for.
please add unit-test(s) to trigger those events with nil cluster ID.

yevgeny-shnaidman:
> looks good overall, but this fixes an issue that we didn't have tests for.
> please add unit-test(s) to trigger those events with nil cluster ID.

added

masayag:
LGTM, @nmagnezi pls take another look.

nmagnezi (approved):
/lgtm
test looks good. thanks!

PR #2625, opened 2021-09-14
Adding monitoring metric with object id and threshold

tsorya:
NO-ISSUE: Adding monitoring metric with object id and threshold
NO-ISSUE: Adding cluster monitoring metrics per function

# Assisted Pull Request

## Description
Adding operation metrics with object id
Use those new metrics in cluster monitoring

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


carbonin:
@tsorya @ori-amizur 
How does this PR relate to https://github.com/openshift/assisted-service/pull/2624?

If they're mutually exclusive I prefer the approach here as it provides metrics rather than logs.

tsorya:
Yes they are mutually exclusive, we just didn't know which way is the best and decided to open both

ori-amizur:
> @tsorya @ori-amizur
> How does this PR relate to #2624?
> 
> If they're mutually exclusive I prefer the approach here as it provides metrics rather than logs.

In general they are similar except the metrics, and the other PR implements a trap, meaning that no data will be written if a threshold is not crossed.

internal/cluster/cluster.go
```diff
@@ -337,6 +340,8 @@ func (m *Manager) RefreshStatus(ctx context.Context, c *common.Cluster, db *gorm
 }
 
 func (m *Manager) refreshStatusInternal(ctx context.Context, c *common.Cluster, db *gorm.DB) (*common.Cluster, error) {
+
+	defer commonutils.MeasureOperationWithThresholdAndId("ClusterMonitoring-refreshStatusInternal", m.log, m.metricAPI, OperationThresholdInSec, c.ID.String())()
```
carbonin, line 344: I thought it wasn't a good practice to put a specific id into a metric.
Unless we're interested in filtering metrics for a specific cluster I feel like this will create a whole lot of data.

Didn't we have an issue with putting cluster ids into metrics before or am I misremembering @slaviered ?

carbonin:
I think I'd still prefer metrics over logs for this.

Also, it looks like this implementation only creates a metric if we're over a threshold. I'm not even sure we want that. Why not always record the metric and do the threshold definition in an alert if we need to?

Or we could combine the approaches and log for a specific cluster if we're over the threshold, but always record the metric without a cluster tag.

flaper87:
/close

another PR based on this one went in already

flaper87 (commented):
/reopen

whooops, confused the PRs

tsorya:
> I think I'd still prefer metrics over logs for this.
> 
> Also, it looks like this implementation only creates a metric if we're over a threshold. I'm not even sure we want that. Why not always record the metric and do the threshold definition in an alert if we need to?
> 
> Or we could combine the approaches and log for a specific cluster if we're over the threshold, but always record the metric without a cluster tag.

the idea was to put metric with ID to be able to catch clusters that takes long time to monitor.
We shouldn't have many metrics like this as there is a threshold and we should not reach it

slaviered:
Also please note that we try to reduce the metrics that are based on object IDs because it inflates the database records of Prometheus

tsorya:
@slaviered totally agree but this metric shouldn't be pushed in regular case. If we have it, it is sort of alert

internal/cluster/cluster.go
tsorya: The idea is to find cluster/host that creates the issue. In regular case we shouldn't create the metric as threshold will not be reached.

pkg/commonutils/common_utils.go
```diff
@@ -18,3 +18,16 @@ func MeasureOperation(operation string, log logrus.FieldLogger, metricsApi metri
 		}
 	}
 }
+
+func MeasureOperationWithThresholdAndId(operation string, log logrus.FieldLogger, metricsApi metrics.API, thresholdInSec float64, objectId string) func() {
+	start := time.Now()
+	return func() {
+		duration := time.Since(start)
+		if duration.Seconds() >= thresholdInSec {
+			log.Debugf("%s for %s took : %v", operation, objectId, duration)
```
filanov, line 27: suggesting to make it a warning

filanov:
@tsorya is that still relevant? can we close this PR?

PR #2640, opened 2021-09-18
MGMT-7687: Propagate Infrastructure Operator node selection and tolerations to its workloads

celebdor:
# Assisted Pull Request

## Description

There is a common pattern in OpenShift to create [infrastructure machinesets](https://docs.openshift.com/container-platform/4.8/machine_management/creating-infrastructure-machinesets.html). These allow the cluster admins to have machines specialized for running the things like Hive, ACM, assisted, etc.

The upcoming Open Cluster Management 2.4 release adds support for running the operators in the infrastructure nodes. In order to satisfy the expectation that the OCM components run in the infrastructure nodes, we need to make sure that the workloads that we spawn remain in the same node selection as the Infrastructure operator does.

The way this commit seeks to achieve that, is by retrieving the node selection and the toleration that OCM set for the  infrastructure operator and set it for the image service and assisted service deployments as well.

### How to test

This PR should be tested by:
- Labeling some worker node in a specific way
- Set nodeSelector and/or tolerations in the Subscription object
- See that both the infrastructure, assisted-service and image-service all run in the selected node and that their deployments all have both the nodeSelector and the Tolerations set to what was put in Subscriptions

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @djzager 
/cc @flaper87 

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -43,7 +43,6 @@ import (
 const (
 	NamespaceEnvVar string = "NAMESPACE"
 	PodNameEnvVar   string = "POD_NAME"
-	K8sSvcHost      string = "KUBERNETES_SERVICE_HOST"
 )
 
 var (
@@ -94,11 +93,6 @@ func main() {
 		setupLog.Error(fmt.Errorf("%s environment variable must be set (commonly set automatically in every Pod)", NamespaceEnvVar), "unable to get namespace")
 		os.Exit(1)
 	}
-	podName, found := os.LookupEnv(PodNameEnvVar)
-	if !found {
-		setupLog.Error(fmt.Errorf("%s environment variable must be set (commonly set automatically in every Pod)", PodNameEnvVar), "unable to get pod name")
-		os.Exit(1)
-	}
 
 	// must have openshift versions specified on the operator
 	// this prevents us from having to include the full json in source
@@ -121,8 +115,9 @@ func main() {
 	client := mgr.GetClient()
 	var nodeSelector map[string]string
 	var tolerations []corev1.Toleration
-	_, found = os.LookupEnv(K8sSvcHost)
-	if found { // Running inside a Kubelet pod
+
+	podName, found := os.LookupEnv(PodNameEnvVar)
+	if found {
 		operatorPod := &corev1.Pod{}
 		if err = client.Get(context.TODO(), types.NamespacedName{Name: podName, Namespace: ns}, operatorPod); err != nil {
 			setupLog.Error(err, "Unable to get Infrastructure Operator Pod")
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -43,6 +43,7 @@ import (
 const (
 	NamespaceEnvVar string = "NAMESPACE"
 	PodNameEnvVar   string = "POD_NAME"
+	K8sSvcHost      string = "KUBERNETES_SERVICE_HOST"
 )
 
 var (
@@ -118,10 +119,17 @@ func main() {
 	}
 
 	client := mgr.GetClient()
-	operatorPod := &corev1.Pod{}
-	if err = client.Get(context.TODO(), types.NamespacedName{Name: podName, Namespace: ns}, operatorPod); err != nil {
-		setupLog.Error(err, "Unable to get Infrastructure Operator Pod")
-		os.Exit(1)
+	var nodeSelector map[string]string
+	var tolerations []corev1.Toleration
+	_, found = os.LookupEnv(K8sSvcHost)
+	if found { // Running inside a Kubelet pod
+		operatorPod := &corev1.Pod{}
+		if err = client.Get(context.TODO(), types.NamespacedName{Name: podName, Namespace: ns}, operatorPod); err != nil {
+			setupLog.Error(err, "Unable to get Infrastructure Operator Pod")
+			os.Exit(1)
+		}
+		nodeSelector = operatorPod.Spec.NodeSelector
+		tolerations = operatorPod.Spec.Tolerations
 	}
 
 	if err = (&controllers.AgentServiceConfigReconciler{
@@ -130,8 +138,8 @@ func main() {
 		Scheme:       mgr.GetScheme(),
 		Recorder:     mgr.GetEventRecorderFor("agentserviceconfig-controller"),
 		Namespace:    ns,
-		NodeSelector: operatorPod.Spec.NodeSelector,
-		Tolerations:  operatorPod.Spec.Tolerations,
+		NodeSelector: nodeSelector,
+		Tolerations:  tolerations,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "AgentServiceConfig")
 		os.Exit(1)
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -36,6 +36,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
@@ -112,12 +113,17 @@ func main() {
 		os.Exit(1)
 	}
 
-	client := mgr.GetClient()
 	var nodeSelector map[string]string
 	var tolerations []corev1.Toleration
 
 	podName, found := os.LookupEnv(PodNameEnvVar)
 	if found {
+		client, clientErr := client.New(mgr.GetConfig(), client.Options{Scheme: mgr.GetScheme()})
+		if err != nil {
+			setupLog.Error(clientErr, "Unable to create new client")
+			os.Exit(1)
+		}
+
 		operatorPod := &corev1.Pod{}
 		if err = client.Get(context.TODO(), types.NamespacedName{Name: podName, Namespace: ns}, operatorPod); err != nil {
 			setupLog.Error(err, "Unable to get Infrastructure Operator Pod")
@@ -128,7 +134,7 @@ func main() {
 	}
 
 	if err = (&controllers.AgentServiceConfigReconciler{
-		Client:       client,
+		Client:       mgr.GetClient(),
 		Log:          logrus.New(),
 		Scheme:       mgr.GetScheme(),
 		Recorder:     mgr.GetEventRecorderFor("agentserviceconfig-controller"),
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -28,7 +29,9 @@ import (
 	"github.com/openshift/assisted-service/models"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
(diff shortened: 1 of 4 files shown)


cmd/operator/main.go
```diff
@@ -108,12 +112,21 @@ func main() {
 		os.Exit(1)
 	}
 
+	client := mgr.GetClient()
+	operatorDeployment := &appsv1.Deployment{}
+	if err = client.Get(context.TODO(), types.NamespacedName{Name: operatorDeploymentName, Namespace: ns}, operatorDeployment); err != nil {
+		setupLog.Error(err, "Unable to get Infrastructure Operator Deployment")
+		os.Exit(1)
+	}
```
flaper87, line 120: I was considering whether this belongs to the `Reconcile` function or not. I think it's fine here since there's no way to change these values without restarting the `infrastructure-operator` pod, which will restart the reconcile and, therefore, the assisted-service deployment.

flaper87 (changes requested):
This is exactly how I was planning to implemented when we talked about this last week. :+1: 

It'd be great to add some unit-tests for it (just for sanity). Meanwhile, I can help testing this live.

cmd/operator/main.go
celebdor: I was thinking about it for quite some time too :-)

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -798,6 +802,8 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
 		deployment.Spec.Template.Spec.Containers = []corev1.Container{container}
 		deployment.Spec.Template.Spec.Volumes = volumes
 		deployment.Spec.Template.Spec.ServiceAccountName = imageServiceName
+		deployment.Spec.Template.Spec.NodeSelector = r.NodeSelector
```
mhrivnak, line 805: I suggest making copies. The resource will end up in the client cache, and who knows what may happen to it from there. We don't want some future changes to leak into the map and slice that are stored on the Reconciler.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1082,6 +1088,9 @@ func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(ctx context.
 		deployment.Spec.Template.Spec.Volumes = volumes
 		deployment.Spec.Template.Spec.ServiceAccountName = serviceAccountName
 
+		deployment.Spec.Template.Spec.NodeSelector = r.NodeSelector
```
mhrivnak, line 1091: I suggest making copies. The resource will end up in the client cache, and who knows what may happen to it from there. We don't want some future changes to leak into the map and slice that are stored on the Reconciler.

mhrivnak:
I remembered one other use case that this breaks. I think we should preserve the ability to run the operator process outside a k8s cluster, for example for a dev running `go run cmd/main.go` or `operator-sdk run` to get quick feedback in their dev workflow. It's a common operator developer workflow, and it requires some way to provide this value other than reading it off the operator's own Pod. Maybe a command-line arg to skip loading the Pod definition?

celebdor:
> I remembered one other use case that this breaks. I think we should preserve the ability to run the operator process outside a k8s cluster, for example for a dev running `go run cmd/main.go` or `operator-sdk run` to get quick feedback in their dev workflow. It's a common operator developer workflow, and it requires some way to provide this value other than reading it off the operator's own Pod. Maybe a command-line arg to skip loading the Pod definition?

config/manager/manager.yaml
```diff
@@ -54,6 +54,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: POD_NAME
```
mhrivnak, line 57: @djzager what needs to be done to get this into the bundle? We'll also need to make sure it gets into ACM's chart, which might just require re-generating it from our updated bundle. But we should give them a heads-up about the change.

djzager: Need to run `make generate-manifests` (that's why pr-merge is failing) to get it into the clusterserviceversion. Next time we push to community-operators-prod, ACM will be able to pick it up.

djzager:
> I remembered one other use case that this breaks. I think we should preserve the ability to run the operator process outside a k8s cluster, for example for a dev running `go run cmd/main.go` or `operator-sdk run` to get quick feedback in their dev workflow. It's a common operator developer workflow, and it requires some way to provide this value other than reading it off the operator's own Pod. Maybe a command-line arg to skip loading the Pod definition?

No one is really trying to do that right now. I'm comfortable with creating a ticket to support `go run cmd/operator/main.go`, that we can prioritize later.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -799,6 +803,13 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
 		deployment.Spec.Template.Spec.Containers = []corev1.Container{container}
 		deployment.Spec.Template.Spec.Volumes = volumes
 		deployment.Spec.Template.Spec.ServiceAccountName = imageServiceName
+
+		nodeSelector := make(map[string]string)
```
djzager, line 807: Is it possible for node selectors or tolerations to be manually modified by a user? 

said another way, do we really want to clobber the nodeSelector + tolerations on every reconcile?

mhrivnak:
> > I remembered one other use case that this breaks. I think we should preserve the ability to run the operator process outside a k8s cluster, for example for a dev running `go run cmd/main.go` or `operator-sdk run` to get quick feedback in their dev workflow. It's a common operator developer workflow, and it requires some way to provide this value other than reading it off the operator's own Pod. Maybe a command-line arg to skip loading the Pod definition?
> 
> No one is really trying to do that right now. I'm comfortable with creating a ticket to support `go run cmd/operator/main.go`, that we can prioritize later.

Works for me as long as it doesn't get forgotten. If someone needs to work on this from the ACM release branch, for example to fix a bug after feature freeze or release, it'll slow them down if the process can't run locally. If we go into feature freeze that way, it's a compromise that needs to be accepted.

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: I don't think there's any expectation that manual changes can be applied to this Deployment.

cmd/operator/main.go
```diff
@@ -108,12 +118,28 @@ func main() {
...
+	if found { // Running inside a Kubelet pod
+		operatorPod := &corev1.Pod{}
+		if err = client.Get(context.TODO(), types.NamespacedName{Name: podName, Namespace: ns}, operatorPod); err != nil {
+			setupLog.Error(err, "Unable to get Infrastructure Operator Pod")
+			os.Exit(1)
+		}
+		nodeSelector = operatorPod.Spec.NodeSelector
+		tolerations = operatorPod.Spec.Tolerations
+	}
+
```
djzager, line 134: My idea.

```suggestion
	var nodeSelector map[string]string
	var tolerations []corev1.Toleration

	podName, found := os.LookupEnv(PodNameEnvVar)
	if found {
		client := mgr.GetClient()
		operatorPod := &corev1.Pod{}
		if err = client.Get(context.TODO(), types.NamespacedName{Name: podName, Namespace: ns}, operatorPod); err != nil {
			setupLog.Error(err, "Unable to get Infrastructure Operator Pod")
			os.Exit(1)
		}
		nodeSelector = operatorPod.Spec.NodeSelector
		tolerations = operatorPod.Spec.Tolerations
	}
```

djzager:
> Works for me as long as it doesn't get forgotten. If someone needs to work on this from the ACM release branch, for example to fix a bug after feature freeze or release, it'll slow them down if the process can't run locally. If we go into feature freeze that way, it's a compromise that needs to be accepted.

ended up fixing it anyway :sunglasses:

djzager:
Tested manually using dev-scripts cluster by adding:

```
nodeSelector:
    kubernetes.io/arch: amd64
```

to the `spec.install.spec.deployments[0].spec.template.spec` in the assisted-service-operator's clusterservice version. 

```
$ kubectl get pod/infrastructure-operator-7585847f5c-2b5dd -o yaml | yq eval '.spec.nodeSelector' -
kubernetes.io/arch: amd64

$ kubectl get pod/assisted-image-service-76b7f6bc88-fsnrz -o yaml | yq eval '.spec.nodeSelector' -
kubernetes.io/arch: amd64

$ kubectl get pod/assisted-service-7fc4c4c669-5r7l9 -o yaml | yq eval '.spec.nodeSelector' -
kubernetes.io/arch: amd64
```

Then adding a toleration, first adding a taint to a node with `kubectl taint nodes master-2 key1=value1:NoSchedule`:

```
tolerations:
  - key: "key1"
    operator: "Equal"
    value: "value1"
    effect: "NoSchedule"
```

to the `spec.install.spec.deployments[0].spec.template.spec` in the assisted-service-operator's clusterservice version.

```
$ kubectl get pod/infrastructure-operator-5894ccf84b-5wth7 -o yaml | yq eval '.spec.tolerations' -
- effect: NoSchedule
  key: key1
  operator: Equal
  value: value1
- effect: NoExecute
  key: node.kubernetes.io/not-ready
  operator: Exists
  tolerationSeconds: 300
- effect: NoExecute
  key: node.kubernetes.io/unreachable
  operator: Exists
  tolerationSeconds: 300
- effect: NoSchedule
  key: node.kubernetes.io/memory-pressure
  operator: Exists

$ kubectl get pod/assisted-service-784dfbd65c-7ppx4 -o yaml | yq eval '.spec.tolerations' -
- effect: NoSchedule
  key: key1
  operator: Equal
  value: value1
- effect: NoExecute
  key: node.kubernetes.io/not-ready
  operator: Exists
  tolerationSeconds: 300
- effect: NoExecute
  key: node.kubernetes.io/unreachable
  operator: Exists
  tolerationSeconds: 300
- effect: NoSchedule
  key: node.kubernetes.io/memory-pressure
  operator: Exists

$ kubectl get pod/assisted-image-service-5b499654b4-vn6q9 -o yaml | yq eval '.spec.tolerations' -
- effect: NoSchedule
  key: key1
  operator: Equal
  value: value1
- effect: NoExecute
  key: node.kubernetes.io/not-ready
  operator: Exists
  tolerationSeconds: 300
- effect: NoExecute
  key: node.kubernetes.io/unreachable
  operator: Exists
  tolerationSeconds: 300
- effect: NoSchedule
  key: node.kubernetes.io/memory-pressure
  operator: Exists
```

I noticed if I remove the nodeSelector/tolerations that we treat it as a no-op. What this means is that if there are existing nodeSelector||tolerations on the assisted(-image)-service pods then we won't touch them if ours are unset.

djzager:
I noticed @mhrivnak that you said in the Jira:

> The AgentClusterInstall resource should add new fields for the operand with the same names: NodeSelector and Tolerations, and then it should add those values (if specified) to any Pods it creates.

Did you mean the operands for the AgentServiceConfig resource? I suspect you did...but want to clarify before we merge it :sunglasses:

djzager:
Assuming we are happy with the behavior described in https://github.com/openshift/assisted-service/pull/2640#issuecomment-924514605 I believe this is ready to go.

mhrivnak:
> I noticed @mhrivnak that you said in the Jira:
> 
> > The AgentClusterInstall resource should add new fields for the operand with the same names: NodeSelector and Tolerations, and then it should add those values (if specified) to any Pods it creates.
> 
> Did you mean the operands for the AgentServiceConfig resource? I suspect you did...but want to clarify before we merge it

Ah yes, indeed I meant AgentServiceConfig. That is the approach that the ACM team preferred. I also marginally preferred it since it would avoid the need to read our own pod, make assumptions about propagating values from operator to operand, and would avoid the need for the command-line argument to skip all that. But it's notable that the hive team decided to do the same thing this PR does. I'm happy either way, and given the deadline, we should probably stick with this.

mhrivnak:
> I noticed if I remove the nodeSelector/tolerations that we treat it as a no-op. What this means is that if there are existing nodeSelector||tolerations on the assisted(-image)-service pods then we won't touch them if ours are unset.

That sounds like a bug. Might be unusual, but if someone wants to move this off of their infrastructure nodes, or even get rid of the infrastructure nodes entirely, seems like that should work.

flaper87:
>  What this means is that if there are existing nodeSelector||tolerations on the assisted(-image)-service pods then we won't touch them if ours are unset.

When you say *ours*, do you mean the operator NodeSelector/Tolerations? 

This looks good enough to move along, let's file a bug for the above if needed

/override ci/prow/e2e-metal-assisted-networking 
/lgtm

:pizza:

PR #2649, opened 2021-09-19
MGMT-7223: Changes for adding remote worker node jobs

lranjbar:
# Assisted Pull Request

## Description

This playbook is meant to be run after `remote_nodes.sh` that lives in dev-scripts. After running `remote_nodes.sh` there is a file called `remote_baremetalhosts.json` that is similar to `extra_baremetalhosts.json`

While dev-scripts generates it's own manifests we still need to apply the correct `infraenv` to the host. This reuses the existing code to do this.

I also renamed `extra_baremetalhosts` to `baremetalhosts` to make it more generic.

TODO: Find validation for the remote hosts once they are applied.

- [X] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [X] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [X] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @flaper87
/cc @mkowalski 

## Checklist

## Reviewers Checklist

### DIFF
--- a/deploy/operator/ztp/assisted-installer-crds-playbook.yaml
+++ b/deploy/operator/ztp/assisted-installer-crds-playbook.yaml
@@ -25,13 +25,14 @@
     - cluster_host_prefix_additional: "{{ lookup('env', 'CLUSTER_HOST_PREFIX_ADDITIONAL') }}"
     - external_subnet_additional: "{{ lookup('env', 'EXTERNAL_SUBNET_ADDITIONAL') }}"
     - service_subnet_additional: "{{ lookup('env', 'SERVICE_SUBNET_ADDITIONAL') }}"
-    - extra_baremetalhosts: "{{ lookup('file', lookup('env', 'EXTRA_BAREMETALHOSTS_FILE')) | from_json }}"
+    - baremetalhosts: "{{ lookup('file', lookup('env', 'EXTRA_BAREMETALHOSTS_FILE')) | from_json }}"
 
   tasks:
   - name: create directory for generated CRDs
     file:
       name: generated
       state: directory
+
   - name: write the cluster image set crd
     template:
       src: "clusterImageSet.j2"
--- a/deploy/operator/ztp/add-remote-nodes-playbook.yaml
+++ b/deploy/operator/ztp/add-remote-nodes-playbook.yaml
@@ -1,0 +1,23 @@
+- name: Add remote nodes for Assisted Installer
+  hosts: localhost
+  collections:
+   - community.general
+  gather_facts: no
+  vars:
+    - baremetalhosts: "{{ lookup('file', lookup('env', 'REMOTE_BAREMETALHOSTS_FILE')) | from_json }}"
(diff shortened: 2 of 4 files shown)


deploy/operator/ztp/assisted-installer-crds-playbook.yaml
```diff
@@ -25,13 +25,14 @@
     - cluster_host_prefix_additional: "{{ lookup('env', 'CLUSTER_HOST_PREFIX_ADDITIONAL') }}"
     - external_subnet_additional: "{{ lookup('env', 'EXTERNAL_SUBNET_ADDITIONAL') }}"
     - service_subnet_additional: "{{ lookup('env', 'SERVICE_SUBNET_ADDITIONAL_ADDITIONAL') }}"
-    - extra_baremetalhosts: "{{ lookup('file', lookup('env', 'EXTRA_BAREMETALHOSTS_FILE')) | from_json }}"
+    - baremetalhosts: "{{ lookup('file', lookup('env', 'EXTRA_BAREMETALHOSTS_FILE')) | from_json }}"
```
flaper87, line 28: `EXTRA_BAREMETALHOSTS_FILE` contains the definition of all the extra nodes created by dev-script, which are currently used to deploy define the `master` nodes. 

I'm saying this because I think we should find a way to differentiate between `masters` and `workers` here. Here's an idea:

Have 2 variables `SPOKE_NUM_MASTERS` and `SPOKE_NUM_WORKERS`. Master nodes will be the first `$NUM_MASTERS` hosts defined in the json and workers will use the rest of the hosts. thoughts?

lranjbar: I like this idea but unsure if we actually need it for RWN. Though we should look into defining the `SPOKE_NUM_MASTERS` and `SPOKE_NUM_WORKERS` to clarify what is going on here. 🤔 

I think also with Pawan's flow he initally wanted to use different VMs for the different SNO clusters. He wanted to put them in separate manifest files as well. With that said perhaps I should think about how we can interact with this a bit more hm.

The flow I'm thinking is:
1) Setup hub cluster w/ dev-scripts, define `EXTRA_BAREMETAL_HOSTS` and use this for the spoke masters.
2) Install assisted operator and set up spoke cluster
3) Run `remote_nodes.sh` in dev-scripts, this will define `REMOTE_BAREMETAL_HOSTS` and we can use this for the spoke workers.
4) Finally we should be able to apply the remote nodes to the spoke namespace and assisted should it pick it up from there? I think.. 😅

flaper87: Sounds reasonable to me! 

While reviewing this PR yesterday, I forgot that `dev-scripts` creates a separate hosts file when calling `remote_nodes.sh`. Now that I remembered this, the changes in this PR make a lot more sense :) 

Let me know if I can help with tests or something :)

flaper87 (approved):
/override ci/prow/e2e-metal-assisted-networking
:pizza:

PR #2653, opened 2021-09-21
MGMT-7852 - Implement V2 InfraEnv Host download Ignition API

yevgeny-shnaidman:
# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


omertuc:
- Why is the path different from https://github.com/openshift/assisted-service/blob/5b1886c1f78a08b79d770755e63baa6699b4b170/swagger.yaml#L2355 ? (no downloads)
- Is the host ignition available even before the host is bound to a cluster?
	- What happens if you try to download the ignition before the host is bound to a cluster?
- Can you please add a commit message / description of what you're trying to do in general? Is it exactly the same as the old API? How it behaves when the host is unbound? etc.

yevgeny-shnaidman:
> * Why is the path different from https://github.com/openshift/assisted-service/blob/5b1886c1f78a08b79d770755e63baa6699b4b170/swagger.yaml#L2355
>    ?
>   (no `/downloads/` in the old path)
> * Is the host ignition available even before the host is bound to a cluster?
>   
>   * If not, what is expected to happen if you try to download the ignition before the host is bound to a cluster?
>   * Does the code handle the above case correctly?
> * Can you please add a commit message of what you're trying to do in general? Is it exactly the same as the old v1 API? How it behaves when the host is unbound? etc.
> 
> EDIT: I see that the implementation (and endpoint without `/downloads/` for v2 host ignition download has already been added in #2381 , so is this PR just creating _another v2_ new endpoint for that function, just with the `/downloads/` added? Why do we need 2 v2 endpoints?
> 
> Can you please edit the PR title / commit message to clarify that this is creating another V2 endpoint (on top of the existing v2 endpoint), and not just creating a v2 endpoint for an existing v1 endpoint?

You can see that in v1 there is additional endpoint with "download" https://github.com/openshift/assisted-service/blob/51c724621cc833b1dd9c5987c86cd2d3f08d679b/swagger.yaml#L2455. This PR adds an endpoint for download for v2.
This difference betwee 2 endpoint is that the "download" one returns a stream. 
In case of the unbound host, this endpoint returns an error.

yevgeny-shnaidman:
> So we have 4 `/ignition` endpoints, 2 in v1, and 2 in v2?
> 
> Their swagger descriptions don't mention anything about there being a difference between `/download/` and non-`/download/` (it's hard to notice that one has a `produces: ... octet` entry and the other doesn't), can you please modify the `description:` for all of them to emphasize this difference?
> 
> Also the description for the infra-env ones should mention that this endpoint only works for bound hosts.
> 
> Also, can you please modify the commit message to explain that
> 
> ```
> There was already:
> 
> `/v1/clusters/.../ignition`
> `/v1/clusters/.../downloads/ignition`
> `/v2/infra-envs/.../ignition`
> 
> And this commit only adds 
> `/v2/infra-envs/.../downloads/ignition`
> 
> Which will allow streaming host ignition for an infra-env host, just like 
> `/v1/clusters/.../download/ignition` allowed streaming of the ignition file for cluster hosts
> ```
> 
> This is very not obvious, it assumes a lot of very specific knowledge about the various ignition endpoints, and it would've been very hard and time-consuming for me to figure out those differences on my own, please take time to give some background in your PRs and commit messages / JIRA tickets. A simple `Add a streaming ignition endpoint for infra envs on top of the non-streaming infra-env ignition endpoint already added by #2381` would have saved a lot of confusion and time I put in trying to understand this PR and writing my previous message.

Fixed, please check

omertuc:
Thank you
/lgtm
/unhold

PR #2662, opened 2021-09-24
Bug 2007515: Pass copy-network when infraEnv.StaticNetworkConfig is set

flaper87:
# Assisted Pull Request

## Description

Originally, assisted-service would add the `--copy-network` argument to
the coreos-install command when the cluster.StaticNetworkConfigured flag
was set. This flag, however, is only been set to true for cloud-managed
deployments.

The cluster.StaticNetworkConfigured flag became a cached value of data
that should be considered dynamic, which broke the abstraction and the
separation of concerns of the various objects managed internally.
Specifically, it put in the cluster data that should only belong to the
infraenv.

This commit removes the use of `cluster.StaticNetworkConfigured` when
deciding to pass the `--copy-network` flag in favor of just checking
that the `infraenv.StaticNetworkConfig` is not empty.

The field in the model is not being removed in this commit as this will
be backported and it'd be preferable to not do database migrations in
backports. A follow-up commit will take care of removing the DB column.

Signed-off-by: Flavio Percoco <flavio@redhat.com>

## List all the issues related to this PR

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [x] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [x] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

Did a deployment with static IP addresses, here's the result:

```
Sep 24 05:50:43 localhost installer[5731]: time="2021-09-24T05:50:43Z" level=info msg="Assisted installer started. Configuration is:\n struct Config {\n\tRole: \"master\",\n\tClusterID: \"064958fa-e558-4775-95c6-6f92ae0e088c\",\n\tInfraEnvID: \"6c7ea4bb-6709-4234-8004-8e5772b7c497\",\n\tHostID: \"4d462bd1-91a2-de92-a9
[...]

PR #2667, opened 2021-09-27
Bug 2007515: Fix regression introduced by copy-network bugfix

omertuc:
# Assisted Pull Request

## Description
Commit 8cf51023fecb0ef0d213b8d0d216176f903f0190 (#2662) introduced a bug
which broke late binding.

The `GetInfraEnvByClusterFromDB` function introduced by that commit
wrongly assumed that every infra env must have a corresponding cluster
ID. This is not the case for late-binding infra-envs where infra envs
can have no cluster ID.

This commit gets rid of the `GetInfraEnvByClusterFromDB` function and instead fetches
the infra-env database object during install command creation by taking the infraenv
ID from the host and using the existing `GetInfraEnvFromDB` function.

All "modern" hosts are guaranteed to have an infra env ID.

Older hosts in the database might not have an infra env ID, but we
do not expect any user to actually try to go through with "installation"
of these old hosts. In any case, the `constructHostInstallerArgs` has
been modified to handle a `nil` `*InfraEnv`.

Unit tests have been adjusted to use the correct infra env ID which is
actually set in the host, rather then re-generating a new infra env ID
when creating test-specific infra env objects.

- [ ] New Feature
- [x] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?
- [ ] dev-scripts environment
- [ ] No tests needed

(Currently there are no E2E tests for late binding, which is the reason this
 regression was missed in the first place)

This PR is being tested manually by attempting to install a late-binding cluster locally.
Proper late-binding tests are in the works and should be merged in the upcoming days.

## Assignees

/cc @flaper87
/cc @carbonin

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/hostcommands/install_cmd.go
+++ b/internal/host/hostcommands/install_cmd.go
@@ -73,9 +73,12 @@ func (i *installCmd) GetSteps(ctx context.Context, host *models.Host) ([]*models
 		return nil, err
 	}
 
-	infraEnv, err := common.GetInfraEnvByClusterFromDB(i.db, *cluster.ID)
-	if err != nil {
-		return nil, err
+	var infraEnv *common.InfraEnv = nil
+	if host.InfraEnvID != "" {
+		infraEnv, err = common.GetInfraEnvFromDB(i.db, host.InfraEnvID)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	fullCmd, err := i.getFullInstallerCommand(cluster, host, infraEnv, bootdevice)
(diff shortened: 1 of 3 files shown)


internal/host/hostcommands/install_cmd.go
```diff
@@ -73,7 +73,11 @@ func (i *installCmd) GetSteps(ctx context.Context, host *models.Host) ([]*models
 		return nil, err
 	}
 
-	infraEnv, err := common.GetInfraEnvByClusterFromDB(i.db, *cluster.ID)
+	if host.InfraEnvID == "" {
+		return nil, errors.New("Hosts without infra envs can no longer be installed")
+	}
```
flaper87, line 78: What do you think about loading the infraEnv only when `host.InfraEnvID` is set. This would help us preserve backwards compatibility, at least for the 2.4 time frame.

This would require checking if `infraEnv` is `nil` here: https://github.com/openshift/assisted-service/blob/4cb3d5365d90873eea60d5f3094167ab4914e132/internal/host/hostcommands/install_cmd.go#L291

flaper87 (changes requested):
Small optimization to guarantee some extra compatibility

flaper87 (approved):
:pizza: 

Thanks for this PR! TIL that hosts now have an `InfraEnvID` :pray:

omertuc:
Force pushed to update out-of-date commit message

flaper87 (approved):

PR #2669, opened 2021-09-28
NO-ISSUE - Increase operator deployment wait-for-service-pod-ready timeout

omertuc:
# Assisted Pull Request

## Description
When deploying the operator, we wait 22m for the service pods to become ready,
and if they don't become ready, we give up and exit with an error.

I've had a run fail due to this timeout being too short, where the pod did eventually
became ready - if the timeout were to be higher, it would've probably worked just fine.

- [ ] New Feature
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @yuvigold
/cc @carbonin

## Checklist

## Reviewers Checklist

### DIFF
--- a/deploy/operator/utils.sh
+++ b/deploy/operator/utils.sh
@@ -35,7 +35,7 @@ function wait_for_pod() {
     namespace="${2:-}"
     selector="${3:-}"
 
-    wait_for_condition "pod" "Ready" "22m" "${namespace}" "${selector}"
+    wait_for_condition "pod" "Ready" "30m" "${namespace}" "${selector}"
 }
 
 function hash() {


carbonin:
I'm fine with this, but with the switch to the image service I wonder if we should be waiting for that pod also?
I'd actually expect that to be the one that takes a while now since it is the one downloading the rhcos images.

omertuc:
> I'm fine with this, but with the switch to the image service I wonder if we should be waiting for that pod also? I'd actually expect that to be the one that takes a while now since it is the one downloading the rhcos images.

Does the assisted service become ready before the image service does?

omertuc:
> I'm fine with this, but with the switch to the image service I wonder if we should be waiting for that pod also? I'd actually expect that to be the one that takes a while now since it is the one downloading the rhcos images.

Anyway it sounds like a good idea but I'd like to get this merged first, because it's messing up my runs

carbonin:
> Does the assisted service become ready before the image service does? (As in, I kinda would expect the former would wait for the latter to become ready)

There's no explicit dependency between them like this.
As of now, yeah I would expect assisted-service to come up first.

carbonin (approved):

PR #2677, opened 2021-09-30
Bug 2007781: Multiple concurrent request for initrd download crashes …

yevgeny-shnaidman:
…the service

In case InfraEnv contains static network configuration, each request for initrd spawns
a process in the AI pod that executes nmstatectl command. Each such command may take ~500ms.
unrestricted concurrent request eventually lead to OOM killer being activated, which may kill the pod
itself. This PR add a API to executer interface that allows executing only a predefined number of concurrent processes.
Static Network generator is move to use this API, wiht default value of 30 concurrent execution

# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/staticnetworkconfig/generator.go
+++ b/pkg/staticnetworkconfig/generator.go
@@ -2,6 +2,7 @@ package staticnetworkconfig
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -13,12 +14,17 @@ import (
 	"github.com/openshift/assisted-service/pkg/executer"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/semaphore"
 	"gopkg.in/ini.v1"
 	"gopkg.in/yaml.v2"
 )
 
 const staticNetworkConfigHostsDelimeter = "ZZZZZ"
 const hostStaticNetworkDelimeter = "HHHHH"
+
+type Config struct {
+	MaxConcurrentGenerations int64 `envconfig:"MAX_CONCURRENT_NMSTATECTL_GENERATIONS" default:"30"`
+}
 
 type StaticNetworkConfigData struct {
 	FilePath     string
@@ -27,25 +33,30 @@ type StaticNetworkConfigData struct {
 
 //go:generate mockgen -source=generator.go -package=staticnetworkconfig -destination=mock_generator.go
 type StaticNetworkConfig interface {
-	GenerateStaticNetworkConfigData(hostsYAMLS string) ([]StaticNetworkConfigData, error)
+	GenerateStaticNetworkConfigData(ctx context.Context, hostsYAMLS string) ([]StaticNetworkConfigData, error)
 	FormatStaticNetworkConfigForDB(staticNetworkConfig []*models.HostStaticNetworkConfig) string
-	ValidateStaticConfigParams(staticNetworkConfig []*models.HostStaticNetworkConfig) error
+	ValidateStaticConfigParams(ctx context.Context, staticNetworkConfig []*models.HostStaticNetworkConfig) error
 }
 
 type StaticNetworkConfigGenerator struct {
+	Config
 	log logrus.FieldLogger
+	sem *semaphore.Weighted
 }
 
-func New(log logrus.FieldLogger) StaticNetworkConfig {
-	return &StaticNetworkConfigGenerator{log: log}
+func New(log logrus.FieldLogger, cfg Config) StaticNetworkConfig {
+	return &StaticNetworkConfigGenerator{
+		Config: cfg,
+		log:    log,
+		sem:    semaphore.NewWeighted(cfg.MaxConcurrentGenerations)}
 }
 
-func (s *StaticNetworkConfigGenerator) GenerateStaticNetworkConfigData(hostsYAMLS string) ([]StaticNetworkConfigData, error) {
+func (s *StaticNetworkConfigGenerator) GenerateStaticNetworkConfigData(ctx context.Context, hostsYAMLS string) ([]StaticNetworkConfigData, error) {
 	hostsConfig := strings.Split(hostsYAMLS, staticNetworkConfigHostsDelimeter)
 	s.log.Infof("Start configuring static network for %d hosts", len(hostsConfig))
 	filesList := []StaticNetworkConfigData{}
 	for i, hostConfig := range hostsConfig {
-		hostFileList, err := s.generateHostStaticNetworkConfigData(hostConfig, fmt.Sprintf("host%d", i))
+		hostFileList, err := s.generateHostStaticNetworkConfigData(ctx, hostConfig, fmt.Sprintf("host%d", i))
 		if err != nil {
 			s.log.WithError(err).Errorf("Failed to create static config for host")
 			return nil, err
@@ -55,7 +66,7 @@ func (s *StaticNetworkConfigGenerator) GenerateStaticNetworkConfigData(hostsYAML
 	return filesList, nil
 }
 
-func (s *StaticNetworkConfigGenerator) generateHostStaticNetworkConfigData(hostConfigString, hostDir string) ([]StaticNetworkConfigData, error) {
+func (s *StaticNetworkConfigGenerator) generateHostStaticNetworkConfigData(ctx context.Context, hostConfigString, hostDir string) ([]StaticNetworkConfigData, error) {
 	hostConfig := strings.Split(hostConfigString, hostStaticNetworkDelimeter)
 	if len(hostConfig) != 2 {
 		msg := fmt.Sprintf("Invalid format of the host config string %s", hostConfig)
@@ -64,7 +75,7 @@ func (s *StaticNetworkConfigGenerator) generateHostStaticNetworkConfigData(hostC
 	}
 	hostYAML := hostConfig[0]
 	macInterfaceMapping := hostConfig[1]
-	result, err := s.executeNMStatectl(hostYAML)
+	result, err := s.executeNMStatectl(ctx, hostYAML)
 	if err != nil {
 		return nil, err
 	}
@@ -81,7 +92,14 @@ func (s *StaticNetworkConfigGenerator) generateHostStaticNetworkConfigData(hostC
 	return filesList, nil
 }
 
-func (s *StaticNetworkConfigGenerator) executeNMStatectl(hostYAML string) (string, error) {
+func (s *StaticNetworkConfigGenerator) executeNMStatectl(ctx context.Context, hostYAML string) (string, error) {
+	err := s.sem.Acquire(ctx, 1)
+	if err != nil {
+		s.log.WithError(err).Errorf("Failed to lock semaphore for nmstatectl execution")
+		return "", err
+	}
+	defer s.sem.Release(1)
+
 	executer := &executer.CommonExecuter{}
 	f, err := executer.TempFile("", "host-config")
 	if err != nil {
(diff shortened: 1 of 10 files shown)


carbonin (approved):

pkg/staticnetworkconfig/generator.go
```diff
@@ -20,6 +20,11 @@ import (
 const staticNetworkConfigHostsDelimeter = "ZZZZZ"
 const hostStaticNetworkDelimeter = "HHHHH"
 
+type Config struct {
+	MaxConcurrentGenerations int `envconfig:"MAX_CONCURRENT_NMSTATECTL_GENERATIONS" default:"30"`
```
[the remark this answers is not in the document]
carbonin: I'm still investigating how this should all work together. These are all pretty extreme upper limits and they should all be configurable after deploy.

That said, this is a much more expensive operation than anything we're doing directly in the image service because this spawns a new process every request. If there is no static network config or if we're using the full iso this won't get called and the image service would be able to handle more.

mhrivnak (changes requested):
This is the right direction and solves the problem, but I'd like to propose three changes that I think will benefit us long-term.

## Context
When we call [`GenerateStaticNetworkConfigData`](https://github.com/openshift/assisted-service/blob/0be2147c0e5128c664d0223469816047c2c94399/internal/bminventory/inventory.go#L1458), we have a context that can be passed in. It already has a timeout set by the web handler layer. Let's not have a new timeout setting and create a new context that's just going to be shadowed by the web handler's timeout and context. Not to mention those two contexts might end up with different timeout values, which would lead to weird behavior. Using one Context for the lifetime of a request matches the pattern/best practice expected by the language.

Passing one Context down the stack enables us to call [CommandContext](https://pkg.go.dev/os/exec#CommandContext) as well, which will kill the process if it's hung.

I see that `GenerateStaticNetworkConfigData` also gets called [from the validation code path](https://github.com/openshift/assisted-service/blob/f9b2f3d3246f91f7e80eab0b85eb9c34742eaa46/internal/ignition/ignition.go#L1406). We can create a context there, either without a timeout or add a timeout if needed.

## Semaphore acquire/release
The semaphore is created and owned by the `StaticNetworkConfigGenerator`. I think it should do the acquiring and releasing in the same package, specifically in [`executeNMStatectl`](https://github.com/openshift/assisted-service/blob/a714f40bf720c30d8e491bdf54e847fb468c0654/pkg/staticnetworkconfig/generator.go#L84). Assuming we are passing one context down the whole stack and utilizing its timeout, then we don't need any semaphore logic spilling into the executor.

## Semaphore implementation
There's a nice weighted semaphore resource [included in the Go project](https://pkg.go.dev/golang.org/x/sync/semaphore) outside the core language. It does exactly what we need, and it eliminates our select block and channel interaction. It's also more readable. I think we should use it.

## Summary
With that, we can:
* Pass one Context all the way down the stack.
* Put semaphore logic in `executeNMStatectl`, using the go project's weighted semaphore.
* replace `ExecuteWithSemaphore` with `ExecuteContext`; it'll be just like `Execute` but take a context as the first arg. This will mirror the relationship between the golang exec package's `Command` and `CommandContext`.
* No longer need a setting for the nmstate queue timeout.

WDYT?

yevgeny-shnaidman:
> This is the right direction and solves the problem, but I'd like to propose three changes that I think will benefit us long-term.
> 
> ## Context
> When we call [`GenerateStaticNetworkConfigData`](https://github.com/openshift/assisted-service/blob/0be2147c0e5128c664d0223469816047c2c94399/internal/bminventory/inventory.go#L1458), we have a context that can be passed in. It already has a timeout set by the web handler layer. Let's not have a new timeout setting and create a new context that's just going to be shadowed by the web handler's timeout and context. Not to mention those two contexts might end up with different timeout values, which would lead to weird behavior. Using one Context for the lifetime of a request matches the pattern/best practice expected by the language.
> 
> Passing one Context down the stack enables us to call [CommandContext](https://pkg.go.dev/os/exec#CommandContext) as well, which will kill the process if it's hung.
> 
> I see that `GenerateStaticNetworkConfigData` also gets called [from the validation code path](https://github.com/openshift/assisted-service/blob/f9b2f3d3246f91f7e80eab0b85eb9c34742eaa46/internal/ignition/ignition.go#L1406). We can create a context there, either without a timeout or add a timeout if needed.
> 
> ## Semaphore acquire/release
> The semaphore is created and owned by the `StaticNetworkConfigGenerator`. I think it should do the acquiring and releasing in the same package, specifically in [`executeNMStatectl`](https://github.com/openshift/assisted-service/blob/a714f40bf720c30d8e491bdf54e847fb468c0654/pkg/staticnetworkconfig/generator.go#L84). Assuming we are passing one context down the whole stack and utilizing its timeout, then we don't need any semaphore logic spilling into the executor.
> 
> ## Semaphore implementation
> There's a nice weighted semaphore resource [included in the Go project](https://pkg.go.dev/golang.org/x/sync/semaphore) outside the core language. It does exactly what we need, and it eliminates our select block and channel interaction. It's also more readable. I think we should use it.
> 
> ## Summary
> With that, we can:
> 
> * Pass one Context all the way down the stack.
> * Put semaphore logic in `executeNMStatectl`, using the go project's weighted semaphore.
> * replace `ExecuteWithSemaphore` with `ExecuteContext`; it'll be just like `Execute` but take a context as the first arg. This will mirror the relationship between the golang exec package's `Command` and `CommandContext`.
> * No longer need a setting for the nmstate queue timeout.
> 
> WDYT?

1) i agree regarding the contetxts. Validation code path should also not be an issue since it still in the context of the web handler
2) I think we should keep ExecuteWithSemaphore: it provides a nice interface that can be used in other places in the code where we will need some kind of semaphore behavior on the executor.

carbonin:
> I think we should keep ExecuteWithSemaphore: it provides a nice interface that can be used in other places in the code where we will need some kind of semaphore behavior on the executor.

I think this isn't (and shouldn't be) particularly common. Ideally we should be striving for not exec-ing additional processes whenever possible.
Also, from a code-design perspective I'd prefer the semephore logic to live with the struct that owns the semephore. We already have a function that encapsulates running nmstatectl in the static network package so it made more sense to me for the semephore-related blocking logic to live there.

carbonin:
Also using the package @mhrivnak suggested makes the logic simple enough to the point that it isn't really worth trying to deduplicate it.

pkg/staticnetworkconfig/generator.go
```diff
@@ -13,39 +14,49 @@ import (
...
+	"golang.org/x/sync/semaphore"
 	"gopkg.in/ini.v1"
 	"gopkg.in/yaml.v2"
 )
 
 const staticNetworkConfigHostsDelimeter = "ZZZZZ"
 const hostStaticNetworkDelimeter = "HHHHH"
 
+type Config struct {
+	MaxConcurrentGenerations int `envconfig:"MAX_CONCURRENT_NMSTATECTL_GENERATIONS" default:"30"`
```
carbonin, line 26: This can just be an int64 to avoid the type conversion later.

pkg/staticnetworkconfig/generator.go
```diff
@@ -81,7 +92,18 @@ func (s *StaticNetworkConfigGenerator) generateHostStaticNetworkConfigData(hostC
...
 
-func (s *StaticNetworkConfigGenerator) executeNMStatectl(hostYAML string) (string, error) {
+func (s *StaticNetworkConfigGenerator) executeNMStatectl(ctx context.Context, hostYAML string) (string, error) {
+	err := s.sem.Acquire(ctx, 1)
+	if err != nil {
+		s.log.WithError(err).Errorf("Failed to lock semaphore for nmstatectl execution")
+		return "", err
+	}
+	defer s.sem.Release(1)
+	if ctx.Err() == context.Canceled {
```
carbonin, line 102: Will `Acquire` return a nil if the context was canceled? And if it would, we should probably check for anything in `ctx.Err()` here, not just `context.Canceled`, right?

yevgeny-shnaidman: in case context was canceled the Acquire will succeed , so i am checking the status of the context with Err. It returns only cCanceled in case of context cancelation

yevgeny-shnaidman: on second thought... i will also need to check the DeadlineExceeds to be sure

pkg/staticnetworkconfig/generator.go
yevgeny-shnaidman: fixed

pkg/staticnetworkconfig/generator.go
yevgeny-shnaidman: fixed

pkg/staticnetworkconfig/generator.go
```diff
@@ -81,7 +92,19 @@ func (s *StaticNetworkConfigGenerator) generateHostStaticNetworkConfigData(hostC
...
+func (s *StaticNetworkConfigGenerator) executeNMStatectl(ctx context.Context, hostYAML string) (string, error) {
+	err := s.sem.Acquire(ctx, 1)
+	if err != nil {
+		s.log.WithError(err).Errorf("Failed to lock semaphore for nmstatectl execution")
+		return "", err
+	}
+	defer s.sem.Release(1)
+	contextError := ctx.Err()
+	if contextError != nil {
+		s.log.WithError(err).Errorf("Failed to lock semaphore for nmstatectl execution before context cancellation/timeout")
```
carbonin, line 104: This should be `WithError(contextError)`

carbonin (approved):

carbonin:
Will wait on @mhrivnak's review to un-hold

pkg/staticnetworkconfig/generator.go
```diff
@@ -81,7 +92,19 @@ func (s *StaticNetworkConfigGenerator) generateHostStaticNetworkConfigData(hostC
...
 
-func (s *StaticNetworkConfigGenerator) executeNMStatectl(hostYAML string) (string, error) {
+func (s *StaticNetworkConfigGenerator) executeNMStatectl(ctx context.Context, hostYAML string) (string, error) {
+	err := s.sem.Acquire(ctx, 1)
+	if err != nil {
+		s.log.WithError(err).Errorf("Failed to lock semaphore for nmstatectl execution")
+		return "", err
+	}
+	defer s.sem.Release(1)
+	contextError := ctx.Err()
```
mhrivnak, line 102: As I understand the behavior of `Acquire`, it will normally [return the context's error](https://cs.opensource.google/go/x/sync/+/036812b2:semaphore/semaphore.go;l=78;drc=09787c993a3ab68e3d1f5c9b2394ab9433f391be) if it has to wait for capacity (there are already `MaxConcurrentGenerations` acquired by other goroutines) AND the context completes while waiting.

There is an edge case where the context becomes "Done" at the same time as capacity for the acquire becomes available. If those happen at the same time, `Acquire` just succeeds. This scenario is not significantly different than if the context expired 1ns after Acquire succeeds. Thus I don't think we need to check for a context error here. We can let such a condition be caught by `ExecuteWithContext` just as it would if the context became done at any point while executing the lines below.

mhrivnak (changes requested):

pkg/staticnetworkconfig/generator.go
yevgeny-shnaidman: fixed

mhrivnak (approved):
Looks great. Thanks!

PR #2678, opened 2021-09-30
Bug 2001973: Fix handling nil value when setting disk encryption

mkowalski:
# Assisted Pull Request

## Description

This commit fixes handling of nil values when setting a feature usage
for disk encryption.

Currently it happens that when sending an update payload with
`disk_encryption: {}` we were trying to access fields inside the
payload, what was causing null pointer exceptions.

This commit also changes the way of determining the status of the
feature from looking at the cluster object into looking into the cluster
update object. Previously the check for nil values was taking into
account the current state of the cluster and not the desired state. With
this change, we are always comparing against the requested state from
the update payload.

Closes-bug: #2001973

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @ybettan 
/cc @flaper87 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -2602,8 +2602,7 @@ func (b *bareMetalInventory) updatePlatformParams(params installer.V2UpdateClust
 }
 
 func (b *bareMetalInventory) setDiskEncryptionUsage(c *models.Cluster, diskEncryption *models.DiskEncryption, usages map[string]models.Usage) {
-
-	if c.DiskEncryption == nil {
+	if diskEncryption == nil || diskEncryption.EnableOn == nil {
 		return
 	}
 
@@ -2612,7 +2611,7 @@ func (b *bareMetalInventory) setDiskEncryptionUsage(c *models.Cluster, diskEncry
 		"mode":         *diskEncryption.Mode,
 		"tang_servers": diskEncryption.TangServers,
 	}
-	b.setUsage(*c.DiskEncryption.EnableOn != models.DiskEncryptionEnableOnNone, usage.DiskEncryption, &props, usages)
+	b.setUsage(*diskEncryption.EnableOn != models.DiskEncryptionEnableOnNone, usage.DiskEncryption, &props, usages)
 }
 
 func (b *bareMetalInventory) updateClusterData(_ context.Context, cluster *common.Cluster, params installer.V2UpdateClusterParams, usages map[string]models.Usage, db *gorm.DB, log logrus.FieldLogger, interactivity Interactivity) error {


mkowalski:
/retest

```
STEP 18: FROM quay.io/ocpmetal/oc-image:bug-1823143-multi-arch AS oc-image
STEP 19: FROM image-registry.openshift-image-registry.svc:...18ab8443bc20d97263136870d34bc14f4a8f95fa089ad636101f7ae3358
error: build error: error creating build container: Error ...image:bug-1823143-multi-arch: too many requests to registry
```

internal/bminventory/inventory.go
```diff
@@ -2602,8 +2602,7 @@ func (b *bareMetalInventory) updatePlatformParams(params installer.V2UpdateClust
 }
 
 func (b *bareMetalInventory) setDiskEncryptionUsage(c *models.Cluster, diskEncryption *models.DiskEncryption, usages map[string]models.Usage) {
-
-	if c.DiskEncryption == nil {
+	if diskEncryption == nil || diskEncryption.EnableOn == nil {
```
ybettan, line 2605: Maybe we can extend `setDiskEncryptionWithDefaultValues` to fit to `updateCluster` as well as `registerCluster` instead of rechecking inner fields of `DiskEncryption` ?

Seems like I forgot to set the default in `updateCluster` as well.

mkowalski: As we already discussed, this seems to be a bit bigger issue than it initially seemed. However, given that fixing it properly would be a bigger change (and we need to have a backported fix to ocm-2.4), I'm not sure whether it wouldn't be better to have 2 fixes

1) as simple as possible fix, just in order to fix the error raised in the BZ
2) proper fix for handling default values in all the scenarios

All the points you raised would be perfect for (2), but then again, it would be a bigger change and backporting it creates a potential for missing something very close to the code freeze

cc @flaper87 @celebdor, as I'm still not sure what is the level of support of this feature in 2.4

flaper87: I think I would rather do this in 2 steps. Let's fix the bug first, then do the bigger change to improve the implementation.

When backporting (or in general, really :stuck_out_tongue_closed_eyes: ), the smaller the PR, the better :)

flaper87 (changes requested):
Should there be a test for this?

flaper87 (commented):
Looks like this work is not really planned for 2.4 so the need to backport is questionable. 

The change is quite small so I'd be ok with backporting it, regardless. Let's maybe not spend too much time on this right now, tho.

internal/bminventory/inventory.go
ybettan: After searching for this is issue I discovered that there is a bug in swagger that doesn't seem to be fixed and there is a workaround for it but it will complicate things in another way - once defaults work correctly, we are unable to update the cluster with DiskEncryption without overriding all the non-mentioned field in the object to the default values unless we duplicate the object in swagger.

I rather just fix the merge in that case.

PR #2731, opened 2021-10-08
Bug 2005440: Use current and new cluster data for dual-stack check

mkowalski:
# Assisted Pull Request

## Description

This commit extends the dual-stack validator to use both existing
cluster data and new cluster params when checking whether a
dual-stackness is requested.

The required behaviour comes from the fact that when using Kube API and
creating a new cluster, the initial payload contains empty Machine
Networks and only afterwards the reconcile loop is creating a cluster
update request with desired Machine Networks. This means that during
creation of the cluster we never have Cluster/Service Networks and
Machine Networks available in the same payload, so that a single
validation is not enough.

In order to overcome this issue, dual-stackness validation takes into
account the current payload as well as the previous state of the
cluster. This allows us to handle a scenario where Cluster/Service
Networks are updated separately from Machine Networks, even if all of
them were passed in the same CR.

Closes-bug: #2005440

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @flaper87 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/network/dual_stack_validations.go
+++ b/internal/network/dual_stack_validations.go
@@ -1,6 +1,7 @@
 package network
 
 import (
+	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/models"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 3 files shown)


flaper87 (commented):
> The required behaviour comes from the fact that when using Kube API and
creating a new cluster, the initial payload contains empty Machine
Networks and only afterwards the reconcile loop is creating a cluster
update request with desired Machine Networks. 

It almost seems like this may be the actual bug.

I'm not opposed to this PR, but I wonder if we can do something to change this behavior. If we already have the payload, why don't we send it? Is the whole service built on the assumption that we may not have the full payload when resources are created?

internal/network/dual_stack_validations.go
```diff
@@ -59,3 +60,35 @@ func VerifyClusterNetworksDualStack(networks []*models.ClusterNetwork, isDualSta
 
 	return nil
 }
+
+func CheckIfDualStack(c *common.Cluster) bool {
```
flaper87, line 64: Could you document this function?

The function is called `CheckIfDualStack` but the variable holding the boolean is called `reqDualStack`. This makes it unclear on what the function should do. To this, we can add that the variable holding the returned value from this function is called `alreadyDualStack`. 

Are we checking if it is dual-stack? Or whether it should be dual stack? Or only if it could be dual stack?

flaper87 (changes requested):
I see the PR is still marked as WIP, still dropping some comments. (ignoring the choco-logs)

mkowalski:
> > The required behaviour comes from the fact that when using Kube API and
> > creating a new cluster, the initial payload contains empty Machine
> > Networks and only afterwards the reconcile loop is creating a cluster
> > update request with desired Machine Networks.
> 
> It almost seems like this may be the actual bug.
> 
> I'm not opposed to this PR, but I wonder if we can do something to change this behavior. If we already have the payload, why don't we send it? Is the whole service built on the assumption that we may not have the full payload when resources are created?

Given that REST API does not have this behaviour, I'm also eager to say that this may be a bug (if it was a conscious decision, then I don't see any comment on why it would behave like this).

At the same time, I'm not sure that we'd really want to change this behaviour for 2.4; it would be for sure a bigger change than this one here and given that I don't see a reason for the current behaviour, also much less understood. I'd prefer to just tweak the validators so that we cover the required usecases in 2.4 without changing the logic.

For master, yes, we should revise the behaviour by trying to understand
1) why Machine Networks are empty in the initial cluster creation payload even if provided in the very first CR
2) why behaviour between Kube and REST APIs is different

PR #2733, opened 2021-10-08
MGMT-7923: Image service cloud authentication enhancement

carbonin:
## Description

Authentication between the image service and assisted service when deployed
in the cloud requires a new strategy because the existing tokens used in the cloud
are too large and too short lived to be included in a pre-signed URL. Instead
of the existing authentication used in the cloud, assisted service will issue JWTs
using a symmetric key specific to each infraEnv. This token will be used
as authentication for requests to assisted service, specifically for the endpoints
required by the image service.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @mhrivnak 
/cc @avishayt 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/image-service-cloud-authentication.md
+++ b/docs/enhancements/image-service-cloud-authentication.md
@@ -1,0 +1,153 @@
+---
+title: image-service-cloud-authentication
+authors:
+  - "@carbonin"
+creation-date: 2021-10-07
+last-updated: 2021-10-18
+---
+
+# Image Service Cloud Authentication
+
+## Summary
+
+Authentication between the image service and assisted service when deployed
+in the cloud requires a new strategy because the existing tokens used in the cloud
+are too large and too short lived to be included in a pre-signed URL. Instead
+of the existing authentication used in the cloud, assisted service will issue JWTs
+using a symmetric key specific to each infraEnv. This token will be used
+as authentication for requests to assisted service, specifically for the endpoints
+required by the image service.
+
+## Motivation
+
+When the assisted service is deployed in the cloud, it relies on Red Hat
+SSO for user authentication and authentication tokens from Red Hat SSO are not
+usable with pre-signed URLs for two main reasons:
+
+* Size - these tokens are typically ~2000 characters long
+* Expiration - these tokens have a maximum lifespan of 15 minutes
+
+### Goals
+
+* Define an authentication strategy suitable for pre-signed URLs in the cloud. This includes a
+token with a longer expiration (several hours) and a reasonable size (10s to 100s of characters)
+which will be included directly in a download URL
+
+### Non-Goals
+
+* Tie an image download URL to a Red Hat account or organization
+* Maintain public/private key pairs as service configuration to generate and issue tokens
+* Add additional external identity management components
+
+## Proposal
+
+The new authentication mechanism will be a JWT signed by the assisted service
+using a randomly generated key stored with the infraEnv. The JWT will include
+an `exp` claim as defined in RFC7519. A token will be included as a parameter
+in the image download URL and assisted service authentication will pass if the
+token in the URL is not expired and validates using the key in the infraEnv record.
+
+Managing the signed URL and key will require a few API changes. The REST API will no
+longer return a `download_url` as a part of an infraEnv as the token in the URL could
+be expired at the time of the call. A new API will be added to fetch pre-signed URL
+which will include a new token. Another new API will be added to force a new key to be
+generated. This will allow a user to invalidate all existing non-expired tokens in case
+a download URL is leaked.
+
+### User Stories
+
+#### Story 1
+
+As a user of assisted service via console.redhat.com, I want a simple download
+URL to access the discovery ISO. This URL should be usable with standard download
+utilities (i.e. `curl`, `wget`) without the need for request headers or additional tools
+like the `ocm` CLI.
+
+#### Story 2
+
+As a user of assisted service via console.redhat.com, I want my sensitive credentials
+and infrastructure information contained within the discovery ISO to be kept secure.
+
+#### Story 3
+
+As a user of assisted service via console.redhat.com, I want the discovery ISO download
+URL to be usable by anyone I choose to give it to without disclosing my personal Red Hat
+account credentials or API tokens.
+
+### Implementation Details/Notes/Constraints [optional]
+
+Assisted service will implement a new security definition to serve this purpose.
+The existing definitions already have well-defined use cases and scopes that do not
+fully match the requirements of this proposal. Specifically the existing urlAuth security
+definition applies to more endpoints than we would want to expose for this enhancement.
+
+This new security definition will apply only to the following endpoints:
+
+* `/v2/infra-envs/{infra_env_id}/downloads/files`
+* `/v2/infra-envs/{infra_env_id}/downloads/minimal-initrd`
+
+These are the endpoints that the image service uses to fetch image customization information.
+
+The symmetric key will be stored in a new column in the `infra_envs` database table and will not
+be accessible through the infraEnv API. The expiration time for tokens will be 4 hours by default
+(the same as the current image expiration time), but will also continue to be configurable.
+
+New API endpoints:
+
+* `GET /v2/infra-envs/{infra_env_id}/downloads/image-url`
+  - Get a new pre-signed download URL for an infraEnv's discovery image
+  - Example response:
+    - `{"url": "https://image-service.example.com/images/{infra_env_id}?image_token=<some-jwt>&version=4.9&arch=x86_64&type=full-iso", "exp": "1634571105"}`
+* `POST /v2/infra-envs/{infra_env_id}/regenerate-signing-key`
+  - Create a new symmetric key for an infraEnv (invalidates all existing JWTs)
+
+These new endpoints and will be protected by SSO user credential authentication and authorization.
+
+The image service will accept a new URL parameter, `image_token`, which will
+then be forwarded to assisted service in the `Image-Token` header key.
+
+### Risks and Mitigations
+
+Accessing a database during authentication has a performance impact, but we should be able
+to cache keys by infraEnv id which would mitigate most of this. Additionally we already
+check the database for authorization and the added time to fetch the key should be small
+compared to the total download time so it likely won't be something users notice.
+
+Introducing a new authentication system always involves some risk that it will be implemented
+incorrectly, but this addition should be easy enough to understand and address any issues that
+could come up.
+
+## Design Details [optional]
+
+### Open Questions
+
+* Should we also support RHSSO user tokens in a header for these endpoints?
+  - The image service could also pass a header through the assisted service
+
+### UI Impact
+
+The UI is currently reading the image download URL from the infraEnv. This will need to change
+to request the URL separately after the infraEnv is created.
+
+### Test Plan
+
+- Image downloads from the cloud with the image service and the new authentication work correctly.
+- An image can not be downloaded without a token or with an invalid token.
+- The `image-url` endpoint returns a URL with a new token with a new expiration claim.
+- The regenerate signing key endpoint invalidates all existing tokens (requests no longer succeed with any previously generated URL).
+- URLs with an existing token expire after 4 hours
+
+## Drawbacks
+
+Implementing our own authentication system is generally risky. We already have a system
+(RHSSO) that works in the cloud, gives us authorization as well, and we don't maintain.
+The only reason we can't use this system for this use case is that we want to make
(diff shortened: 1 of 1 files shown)


docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,164 @@
...
+like the `ocm` CLI.
+
+#### Story 2
+
+As a user of assisted service via console.redhat.com, I want my sensitive credentials
+and infrastructure information contained within the discovery ISO to be kept secure.
+
+#### Story 3
+
+As a user of assisted service via console.redhat.com, I want the discover ISO download
```
omertuc, line 78: nit: discovery

docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,164 @@
...
+
+The new authentication mechanism will be a simple secret and expiration date
+both of which are tied to an infraEnv. This secret will be included as a
+parameter in the image download URL and assisted service authentication will
+pass if the secret in the URL matches the one currently in the infraEnv record
+and the request is processed before the expiration time. A new secret will be
+generated any time a user updates the infraEnv in a way that would change the
+image content. Additionally a new endpoint will be required to force a new
+secret to be created in the event that the current one has expired and no new
+changes to the infraEnv are necessary.
```
avishayt, line 52: Why not have the service create a new one?

docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,164 @@
...
+downloaded by anyone with access to the URL itself.
+
+### User Stories
+
+#### Story 1
+
+As a user of assisted service via console.redhat.com, I want a simple download
+URL to access the discovery ISO. This URL should be usable with standard download
+utilities (i.e. `curl`, `wget`) without the need for request headers or additional tools
+like the `ocm` CLI.
```
avishayt, line 69: And also avoid special characters, unlike S3 URLs (MGMT-1916)

docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,164 @@
...
+
+Assisted service will implement a new security definition to serve this purpose.
+The existing definitions already have well-defined use-cases and scopes that do not
+fully match the requirements of this proposal. Specifically the existing urlAuth security
+definition applies to more endpoints than we would want to expose using just a secret.
+
+This security definition will apply only to the following endpoints:
+
+* `/v2/infra-envs/{infra_env_id}/downloads/files`
+* `/v2/infra-envs/{infra_env_id}/downloads/minimal-initrd`
```
avishayt, line 93: Other than the discovery ISO, we have 2 other APIs that require presigned URLS:
GET /v2/clusters/{cluster_id}/downloads/files-presigned
GET /v2/clusters/{cluster_id}/downloads/credentials-presigned (new in v2, similar to files but we moved the kubeconfig and password here because they require stricter permissions)

How would those work?

carbonin: The goal here is not to fully remove all s3 presigned urls. So I don't think those need to change.

We only need to handle authentication for the image download because when we start using the image service it won't be stored in s3 anymore.

docs/enhancements/image-service-cloud-authentication.md
avishayt: I think a background job is more user-friendly, no?

docs/enhancements/image-service-cloud-authentication.md
avishayt: True

docs/enhancements/image-service-cloud-authentication.md
carbonin: I suppose we already have one for expiring images. :man_shrugging: 

How would we decide when _not_ to renew the secret on a particular infra-env though?
In the cloud a user probably doesn't care about the infra env once they've installed the cluster, but the modeling doesn't make it super easy to determine when that's done. Or do we just keep regenerating the secret until the infra env gets removed?

docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,164 @@
...
+
+These are the endpoints that the image service uses to fetch image customization information
+and are the only ones that are needed for this process.
+
+The secret and expiration date will be stored as columns in the infraEnv database table
+and will not be accessible through the infraEnv API. The expiration will remain the same (4 hours)
+
+A new API endpoint will be added:
+
+* `/v2/infra-envs/{infra_env_id}/downloads/refresh-image-secret`
```
filanov, line 103: maybe it should be called image-secret if we have a CRUD for it then refresh doesn't sounds right, @avishayt what do you think?

docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,164 @@
...
+## Alternatives
+
+* Mint JWTs in our service which we then verify in the request.
+  - This provides us a place to put context for the token (username, expiration, etc),
+    but will require us to maintain our own private and public keys and build out
+    the ability to rotate those keys periodically without breaking existing downlaod URLS.
+  - This is effectively what we do in the operator case, but with the added work of having
+    to cleanly rotate the keys.
+
+* Use RHSSO and a token in the header as we do for all other API requests
```
filanov, line 161: personally it sounds like the right thing to do, is UI our limitations for this one?

docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,164 @@
...
+* Maintain public/private key pairs to generate and issue tokens
+* Add additional external identity management components
+
+## Proposal
+
+The new authentication mechanism will be a simple secret and expiration date
+both of which are tied to an infraEnv. This secret will be included as a
+parameter in the image download URL and assisted service authentication will
+pass if the secret in the URL matches the one currently in the infraEnv record
+and the request is processed before the expiration time. A new secret will be
```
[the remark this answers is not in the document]
avishayt: Instead of 1 hour or 4 hours, why not 24 hours, or a week?

docs/enhancements/image-service-cloud-authentication.md
avishayt: The UI and BMCs don't know how to pass auth headers

docs/enhancements/image-service-cloud-authentication.md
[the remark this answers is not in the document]
mhrivnak: So even today if I get the URL and perhaps even start the download, that will fail if it's not done by the 4h mark? There's no graceful rotation?

docs/enhancements/image-service-cloud-authentication.md
carbonin: I'm not sure we'll have a CRUD. Maybe the endpoint should update the secret and return the URL to the image?

So something like `GET /v2/infra-envs/{infra_env_id}/downloads/image-url` would work I suppose.
But that doesn't make a ton of sense given that the image url is already in the infra env as a property.

docs/enhancements/image-service-cloud-authentication.md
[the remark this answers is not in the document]
carbonin: > What would the value be for including it unsigned? Would we expect someone to use it with SSO auth?

No, the idea was just to avoid duplicating information while also separating concerns. If we want a separate endpoint that is specifically meant to fetch the signed url and possibly (see comment below) refresh the token I wouldn't also want a signed url in the infraenv resource where we'd have to decide whether to re-implement the token refresh logic or knowingly give the user an expired or soon-expiring token.

Specifically I don't think it's a good idea to trigger a token refresh when we `GET` an infraenv because that could leave a user with an invalid URL that they reasonably thought would be valid for some time just because someone else queried the infraenv.

So I guess the options are give the user a URL that might no longer be valid when they query an infraenv (which is probably what we're doing today), or leave the URL without auth.

> It seems like a reasonable goal to not invalidate a URL that was generated 30s ago just because a new URL was retrieved.

So it sounds like we want something like an initial expiration and some lower time where we'd force a refresh if someone requests the url even if the token isn't actually expired yet. Say, if there's < 30 minutes left on a 4 hour token. Maybe even a parameter to force a refresh at any time.

docs/enhancements/image-service-cloud-authentication.md
carbonin: Sorry I just saw that you also responded to this @avishayt :sweat_smile: 

I'm not sure BMC is relevant for this as it would only be used in the cloud and we won't support serving directly to a virtual media interface in any case.

docs/enhancements/image-service-cloud-authentication.md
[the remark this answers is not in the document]
mhrivnak: > So I guess the options are give the user a URL that might no longer be valid when they query an infraenv (which is probably what we're doing today), or leave the URL without auth.

The third option is to not put any download URL in the infraenv resource representation, and only make such URLs available from a dedicated endpoint. Then when someone uses that endpoint, they will get a URL that works for some documented period of time, and it won't affect previously-retrieved download URLs.

docs/enhancements/image-service-cloud-authentication.md
flaper87: > The UI and BMCs don't know how to pass auth headers

Are we expecting users to paste the image download URL inside the BMC directly? Right now, users are likely not doing this since the S3 URLs are massive and the BMC can't handle them. Am I missing something?

docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,164 @@
...
+- The refresh image secret endpoint invalidates the old token (requests no longer succeed with the old URL) and replaces the URL with a valid one
+- URLs with an existing token expire after 4 hours
+
+## Drawbacks
+
+Implementing our own authentication system is generally risky. We already have a system
+(RHSSO) that works in the cloud, gives us authorization as well, and we don't maintain.
+The only reason we can't use this system for this use case is that we want to make
+the download easier for users. The most secure, and most flexible option would be
+to require an OCM JWT token for this download, just as we do for every other API call.
```
flaper87, line 150: This paragraph summarizes my main concern. I would really prefer it if we didn't create another authentication system. 

Have we asked the RHSSO folks what they think? Maybe they know of a way for us to accomplish what we need. It be nice to have their feedback as part of this proposal.

docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,151 @@
...
+
+* Tie an image download URL to a Red Hat account or organization
+* Maintain public/private key pairs as service configuration to generate and issue tokens
+* Add additional external identity management components
+
+## Proposal
+
+The new authentication mechanism will be a JWT signed by the assisted service
+using a randomly generated key stored with the infraEnv. The JWT will include
+an `exp` claim as defined in RFC7519. A token will be included as a parameter
```
mhrivnak, line 46: Will we also include a `sub` claim identifying the infraenv? I don't think it would strictly be required, since the signature won't be valid unless using the token is used to access the matching infraenv, but it seems reasonable for observability to include an identifier in the JWT.

carbonin: Not adding it saves about 60 characters in the token. I guess it depends on how much we care about that length.

docs/enhancements/image-service-cloud-authentication.md
```diff
@@ -0,0 +1,153 @@
...
+* `GET /v2/infra-envs/{infra_env_id}/downloads/image-url`
+  - Get a new pre-signed download URL for an infraEnv's discovery image
+  - Example response:
+    - `{"url": "https://image-service.example.com/images/{infra_env_id}?image_token=<some-jwt>&version=4.9&arch=x86_64&type=full-iso", "exp": "1634571105"}`
+* `POST /v2/infra-envs/{infra_env_id}/regenerate-signing-key`
+  - Create a new symmetric key for an infraEnv (invalidates all existing JWTs)
+
+These new endpoints and will be protected by SSO user credential authentication and authorization.
+
+Both the assisted service and image service will accept a new URL parameter `image_token`
```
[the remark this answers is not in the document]
mhrivnak: Agreed. A header is preferable for many reasons. I think it's reasonable to start there, and we could add query-param support if a use case appears.

PR #2754, opened 2021-10-12
MGMT-7932: added PodDisruptionBudget to template.yaml

isaacdorfman:
Added PodDisruptionBudget on the assisted-service to help prevent unavailability.

### DIFF
--- a/openshift/template.yaml
+++ b/openshift/template.yaml
@@ -318,3 +318,14 @@ objects:
         targetPort: 8090
     selector:
       app: assisted-service
+- apiVersion: policy/v1
+  kind: PodDisruptionBudget
+  metadata:
+    labels:
+      app: assisted-service
+    name: assisted-service
+  spec:
+    minAvailable: 2
+    selector:
+      matchLabels:
+        app: assisted-service


filanov:
cc: @osherdp

openshift/template.yaml
```diff
@@ -318,3 +318,14 @@ objects:
         targetPort: 8090
     selector:
       app: assisted-service
+- apiVersion: policy/v1
```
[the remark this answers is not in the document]
filanov: we don't use this template for the operator, in the operator we have only one replica.

osherdp, line 321: I wonder if that's the right apiVersion for the cluster version that we use, because in this example they used betav1:
https://github.com/app-sre/github-mirror/blob/e03ea308941d1dbc14f9530a4c7b751b3e45936b/openshift/github-mirror.yaml#L7

openshift/template.yaml
```diff
@@ -318,3 +318,14 @@ objects:
...
+  kind: PodDisruptionBudget
+  metadata:
+    labels:
+      app: assisted-service
```
osherdp, line 325: seems like the deployment is not labeled with this label, no?

PR #2688, opened 2021-10-02
NO-ISSUE: Use 4.8 redhat-operators catalog on OCP 4.9

omertuc:
# Assisted Pull Request

## Description
See the # comment in the diff for the rationale

This should solve the widespread LSO CI errors we've been seeing recently

## List all the issues related to this PR

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [x] CI/CD 

## What environments does this code impact?

- [x] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Assignees

/cc @YuviGold
/cc @osherdp

## Checklist

- [x] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] Reviewers have been listed
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- [ ] Are the title and description (in both PR and commit) meaningful and clear?
- [ ] Is there a bug required (and linked) for this change?
- [ ] Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #2802, opened 2021-10-20
Revert "Use 4.8 redhat-operators catalog on OCP 4.9 (#2688)"

mkowalski:
# Assisted Pull Request

## Description

With OCP 4.9 released and all the required operators availalble in the
4.9 catalog we do not need to override manually anymore.

Reverts: #2688

## Assignees

/cc @omertuc 
/cc @osherdp


mkowalski:
FYI: It has never been backported to the release-ocm-2.4, so no need to revert in any other branch

PR #2824, opened 2021-10-26
MGMT-7897: Add Image-Token authentication

carbonin:
## Description

This PR is the implementation of the [image service cloud authentication enhancement](https://github.com/openshift/assisted-service/blob/f5b56b1923b534c16a13c17ef29bde902466e4cc/docs/enhancements/image-service-cloud-authentication.md).

It adds a new column to `infra_envs` to hold an encryption key which is set whenever a new infraEnv is created. This key is used sign tokens which are added to the infraEnv image download url when the image service is deployed in the cloud.

Request flow using this new authentication scheme:
1. Request arrives to one of the infraEnv download endpoints with an `Image-Token` header
2. Assisted service middleware stores infraEnv ID from the request path in the request context
3. Authentication validates the token using the key associated with the infraEnv for the id in the token `sub` claim
4. If a token is valid, authentication stores the `sub` claim in the request context
5. Authorization ensures the infraEnv ID from the `sub` claim in the request context matches the infraEnv ID in the request path (also from the context)
6. Request is processed

This auth scheme is used only on the ignition and minimal ISO initrd download endpoints and is intended to be used when the image service downloads those artifacts when building an image stream.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

I deployed assisted service using the `deploy-test` make target and configured wiremock as we do in subsystem test setup.
I was then able to create a cluster and validated that I could make successful requests using tokens signed with the key in the automatically created infraEnv.

## Assignees

/cc @rollandf 
/cc @filanov 
/cc @mhrivnak 

## Checklist

## Reviewers Checklist


carbonin:
I don't think we can create a subsystem test for this currently.

We would need a way for the service to generate a token for us to use, but we only do that when we're using the image service (which we are not doing currently).

When the image service is added to subsystem I'll add tests for this.

carbonin:
I did forget to add the other new endpoint `POST /v2/infra-envs/{infra_env_id}/regenerate-signing-key`

/hold

carbonin:
@filanov @rollandf anything else you think is needed here (other than passing tests)?

filanov (approved):

carbonin:
@filanov can you re-lgtm? Had to rebase

pkg/auth/rhsso_authenticator.go
```diff
@@ -258,6 +259,34 @@ func (a *RHSSOAuthenticator) AuthURLAuth(_ string) (interface{}, error) {
...
+		return nil, err
+	}
+
+	return []byte(infraEnv.ImageTokenKey), nil
+}
+
+func (a *RHSSOAuthenticator) AuthImageAuth(token string) (interface{}, error) {
+	parsedToken, err := jwt.Parse(token, a.getInfraEnvKey)
+	if err != nil {
+		return nil, err
```
carbonin, line 284: I think I should return some kind of API error here.
Right now if I give an invalid token I'm getting a 500 back. Presumably we want that to be a 401, right?

carbonin: Found this while testing my image service changes.

mhrivnak: Right. 500 is an absolute last resort, indicative of a failure on the service side, not the user side.

swagger.yaml
```diff
@@ -4476,6 +4480,7 @@ paths:
         - userAuth: [admin, read-only-admin, user]
         - agentAuth: []
         - urlAuth: []
+        - imageAuth: []
```
filanov, line 4483: what will happen before we have the image service
What will happen with existing infra-envs?

carbonin:
/hold

Need to create a migration to generate a token for all existing infraEnvs.

swagger.yaml
carbonin: > what will happen before we have the image service

This API will still accept imageAuth tokens if they're present, but we won't issue any before the image service is deployed.

> What will happen with existing infra-envs?

Need to create a migration to create keys for them. Put a hold on the PR until I get to it.

carbonin:
/unhold

Added the migration

filanov:
want to make sure that the following flows are covered.
1. What will happen with existing infra envs - migration handle the creation of the token, is there any change for the URL that we store? 
2. New infra env with image generation (without image service)
3. New infra env with image service
4. What will happen to the URL/images once we move to work with the image service?

carbonin:
@filanov and I met to discuss his questions:

>1. What will happen with existing infra envs - migration handle the creation of the token, is there any change for the URL that we store?

The existing URL in the infra env (s3 presigned, presumably) will continue to be valid until it expires. At this point a new URL will have to be generated (more on this after) which will point to the image service.

>2. New infra env with image generation (without image service)

After this PR is merged new infraEnvs will get a key, but we will continue to generate and download images as we currently do until we deploy the image service.

>3. New infra env with image service

New infraEnvs will get a key and will also get a URL pointing to the image service on create.

>4. What will happen to the URL/images once we move to work with the image service?

Same as question 1. The images and existing presigned URLs will expire and be removed/become invalid respectively.

--------------------------------------------------

We did realize that we have a gap here related to issuing new signed urls. With the v1 API, we have an endpoint to regenerate the ISO and issue a new url, but we don't have something like that for v2 as it's all done during infraEnv update.

If a user wants (or needs) a new download url, but doesn't want to change anything related to the infraEnv, it's a bit strange to require them to call infraEnv update.

Because of this we need another endpoint to re-issue the download URL for an infraEnv. This should be callable externally and should also be used when we create a new key.

Because we're not actually issuing any of these urls currently I think it's safe to leave this new endpoint for a follow-up PR.

filanov (approved):

filanov (approved):

carbonin:
Looks like a flaky test: https://search.ci.openshift.org/?search=%2Fgo%2Fsrc%2Fgithub.com%2Fopenshift%2Fassisted-service%2Fsubsystem%2Fhost_test.go%3A237&maxAge=48h&context=1&type=bug%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

PR #2830, opened 2021-10-28
MGMT-7330: Create a job in prow CI to run assisted-service unit tests

pawanpinjarkar:
Install postgresql in Dockerfile.assisted-service-build and use that image to run unit tests in prow using the updated unit-test target.
Locally, users should continue to run 'skipper make unit-test' which will now only start a single container locally.

Signed-off-by: Pawan Pinjarkar

# Assisted Pull Request

- [x] New Feature
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
Run `skipper make unit-test` to start the postgres db in a container created using skipper and runs the existing unit tests
- [x] No tests needed

## Assignees

/cc @carbonin 
/cc @lranjbar 

## Checklist

## Reviewers Checklist


omertuc:
1. I'm not sure it's okay our unit-tests install postgres on dev machines, it was really nice when it was in a container, less destructive
2. The script assumes the developer has `yum` as a package manager. We probably assume this in other places, but it's best to keep these assumptions at a minimum
3. When running in a container, we placed the postgres database data inside a tmpfs filesystem, which significantly improved the performance of the unit tests, by an order of magnitude, this really important benefit has been silently eliminated in this PR

carbonin:
I think the intention here was to run `skipper make unit-test` so that nothing was installed on the local machine.

You're right in that we don't want to assume anything about or change the user's machine.

omertuc:
> I think the intention here was to run `skipper make unit-test` so that nothing was installed on the local machine.
> 
> You're right in that we don't want to assume anything about or change the user's machine.

If that's the intention, then as @osherdp specified on Slack, this installation should be part of the Skipper's image build Dockerfile rather than part of the unit-test Makefile recipe

hack/setup_db.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -euo pipefail
+set -o errexit
+
+su - postgres -c "mkdir -p /tmp/postgres/data"
```
carbonin, line 7: When I was testing this with you I thought we were trying to avoid this approach because it required root or a password prompt.

What changed?

pawanpinjarkar: @carbonin the `continuous-integration/jenkins/pr-merge` was failing  https://github.com/openshift/assisted-service/pull/2830#issuecomment-954083198 as Jenkins runs as a root user.

pawanpinjarkar:
> 1. I'm not sure it's okay our unit-tests install postgres on dev machines, it was really nice when it was in a container, less disruptive

@omertuc We want to be able to run the unit tests in the CI. With the original make unit-test target, we get the docker-in-docker issue in CI. To circumvent that, an initial PR was added here https://github.com/openshift/assisted-service/pull/2393 but based on the discussions at that point we decided to use the assisted-service-build dockerfile and be able to run unit tests using skipper. We definitely don't want to run docker in docker so not sure what are my other options than the current approach added in this PR.

> 2. The script assumes the developer has yum as a package manager. We probably assume this in other places, but it's best to keep these assumptions at a minimum

So on my local machine, I don't have yum installed but the Dockerfile.assisted-service-build gets the `yum` from the underlying base image `registry.ci.openshift.org/openshift/release:golang-1.16`

```
pawanpinjarkar@unknown94e6f7912530 ~/go/src/github.com/openshift/assisted-service (unit-tests) $ skipper make unit-test
[skipper] No build container tag was provided
[skipper] Building image using docker file: Dockerfile.assisted-service-build
Sending build context to Docker daemon    189MB
Step 1/9 : FROM registry.ci.openshift.org/openshift/release:golang-1.16
 ---> 784484e27a9c
Step 2/9 : ENV GO111MODULE=on
 ---> Using cache
 ---> fd80b6d79321
Step 3/9 : ENV GOFLAGS=""
 ---> Using cache
 ---> 65380a944400
Step 4/9 : RUN yum install -y     make     jq     git     https://download.postgresql.org/pub/repos/yum/reporpms/EL-8-x86_64/pgdg-redhat-repo-latest.noarch.rpm
 ---> Using cache
 ---> d2fc1d8b8673
```

omertuc:
> We definitely don't want to run docker in docker so not sure what are my other options than the current approach added in this PR.

Here is an alternative option that doesn't disrupt the existing flow for developers:

You could break things apart, rather than have the unit-test target both create a docker db and run unit tests, have it call 2 separate targets (e.g. docker-db and just-unit-test), one that creates a container database and one that runs the unit tests against that database.

Then also create a third, new target (e.g. install-db), that sets up a yum postgres and starts it like you did in this PR.

Then have the prow CI use install-db and just-unit-test
And keep the regular target, for local development, as-is (unit-test calls docker-db and just-unit-test).

omertuc:
> So on my local machine, I don't have yum installed but the Dockerfile.assisted-service-build gets the yum from the underlying base image registry.ci.openshift.org/openshift/release:golang-1.16

This is assuming you use skipper, in that case - just install the postgres database inside the `Dockerfile.assisted-service-build` Dockerfile - the makefile has no business installing packages

pawanpinjarkar:
> > We definitely don't want to run docker in docker so not sure what are my other options than the current approach added in this PR.
> 
> Here is an alternative option that doesn't disrupt the existing flow for developers:
> 
> You could break things apart, rather than have the unit-test target both create a docker db and run unit tests, have it call 2 separate targets (e.g. docker-db and just-unit-test). docker-db just creates a container database and just-unit-test runs the unit tests against whatever database is provided to it (specified with env vars, for example)
> 
> Then also create a third, new target (e.g. install-db), that sets up a yum postgres and starts it like you did in this PR.
> 
> Then have the prow CI use install-db and just-unit-test - no container-in-container needed
> 
> And keep the regular unit-test target as-is for local development - unit-test calls docker-db and just-unit-test

So, I previously had a similar approach to this i.e. let CI use a new target and do whatever it needs but keep running the unit tests locally unchanged. PR here https://github.com/openshift/assisted-service/pull/2393/files but the discussion on that PR led to this PR :man_facepalming: 

The only reason I have modified the existing unit-test target is that we want to keep only one way of running unit tests locally as well as in CI but looking at the current discussions, I think it's better to go back to the changes I had proposed in the earlier PR https://github.com/openshift/assisted-service/pull/2393/files

@carbonin  @lranjbar @celebdor

carbonin:
> I think it's better to go back to the changes I had proposed in the earlier PR

I disagree. Composing the pieces of this differently is still much better than maintaining two completely different flows for running tests.

PR #2844, opened 2021-11-01
MGMT-8090: Add state unbinding-requires-user-action

rollandf:
# Assisted Pull Request

## Description

Add host state unbinding-requires-user-action.

The host will move to this state if the user unbind the host from one of
the following state:
Cancelled, Installed, Error, AddedToExistingCluster.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @yevgeny-shnaidman 
/cc @filanov 
/cc @avishayt 

## Checklist

## Reviewers Checklist


internal/host/statemachine.go
```diff
@@ -312,6 +312,24 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 		PostTransition:   th.PostUnbindHost,
 	})
 
+	sm.AddTransition(stateswitch.TransitionRule{
```
avishayt, line 315: What about the transition from unbinding-pending-user-action to discovering-unbound?

rollandf: My thought is that the only transition out of this state is to re-register the host.
Then the old host instance will be deleted from the DB and the state machine will start from scratch.

WDYT?

yevgeny-shnaidman: The only time when the host is deleted from DB is when DeregisterHost is called. Upon any other re-registration , the host will exist in DB and we must make sure that there is a valid transition from its current state to a new one (discovering-unbound)

internal/host/statemachine.go
```diff
@@ -312,6 +312,24 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
...
+			stateswitch.State(models.HostStatusInstallingInProgress),
+			stateswitch.State(models.HostStatusInstallingPendingUserAction),
+			stateswitch.State(models.HostStatusResettingPendingUserAction),
+			stateswitch.State(models.HostStatusInstalled),
+			stateswitch.State(models.HostStatusResetting),
+			stateswitch.State(models.HostStatusAddedToExistingCluster),
+		},
+		DestinationState: stateswitch.State(models.HostStatusUnbindingPendingUserAction),
+		PostTransition:   th.PostUnbindHost,
+	})
```
yevgeny-shnaidman, line 331: Since the host will continue sending state updates, you must make sure that it always stays in UnbindPedningUserAction, unless it has reregistered.  You can use Noop transactions (see in the statemachine.go file). 
Also, you must add transition from UnbindPedningUserAction to Discovering-Unbound in case node is rebooted with ISO

internal/host/statemachine.go
rollandf: So, the host is not deleted, need to handle the transition

internal/host/statemachine.go
rollandf: Done

internal/host/transition_test.go
```diff
@@ -1463,62 +1463,72 @@ var _ = Describe("Unbind", func() {
...
 			name:      "disconnected to unbinding",
 			srcState:  models.HostStatusDisconnected,
+			dstState:  models.HostStatusUnbinding,
 			success:   true,
 			sendEvent: true,
 		},
 		{
 			name:      "discovering to unbinding",
 			srcState:  models.HostStatusDiscovering,
+			dstState:  models.HostStatusUnbinding,
```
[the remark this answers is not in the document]
rollandf: @yevgeny-shnaidman WDYT?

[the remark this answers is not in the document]
yevgeny-shnaidman: currently host always becomes unbinding, and when the nodes will reconnect it will move to discovering-unbound. Maybe we need to fix it, but it needs to be done in a different PR, since we need to move it to a different transaction

PR #2861, opened 2021-11-02
MGMT-9259: Enhance the schedulable masters feature

mmartinv:
## Description

The current implementation of the schedulable masters feature allows to enable the scheduling of workloads in the control plane. However, it's not possible to disable the scheduling of workloads explicitly. Additionally, the current API always returns  the value of the 'schedulable_masters' property saved in the database at cluster creation time (defaults to "false" if omitted). However the UI needs the API to return the default correct value depending on the number of hosts in the cluster.

The proposed changes allow to effectively disable the scheduling of workloads on control plane and returns the expected default values for each platform in case the 'schedulable_masters' property was not actively set at cluster creation time. 

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @avishayt 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -293,9 +293,6 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(_ context.Context, 
 	if params.NewClusterParams.Hyperthreading == nil {
 		params.NewClusterParams.Hyperthreading = swag.String(models.ClusterHyperthreadingAll)
 	}
-	if params.NewClusterParams.SchedulableMasters == nil {
-		params.NewClusterParams.SchedulableMasters = swag.Bool(false)
-	}
 	if params.NewClusterParams.Platform == nil {
 		params.NewClusterParams.Platform = &models.Platform{
 			Type: common.PlatformTypePtr(models.PlatformTypeBaremetal),
@@ -309,6 +306,9 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(_ context.Context, 
 			EnableOn: swag.String(models.DiskEncryptionEnableOnNone),
 			Mode:     swag.String(models.DiskEncryptionModeTpmv2),
 		}
+	}
+	if params.NewClusterParams.UseSchedulingDefaults == nil {
+		params.NewClusterParams.UseSchedulingDefaults = swag.Bool(true)
 	}
 
 	return params
@@ -463,6 +463,7 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 			HighAvailabilityMode:  params.NewClusterParams.HighAvailabilityMode,
 			Hyperthreading:        swag.StringValue(params.NewClusterParams.Hyperthreading),
 			SchedulableMasters:    params.NewClusterParams.SchedulableMasters,
+			UseSchedulingDefaults: params.NewClusterParams.UseSchedulingDefaults,
 			Platform:              params.NewClusterParams.Platform,
 			ClusterNetworks:       params.NewClusterParams.ClusterNetworks,
 			ServiceNetworks:       params.NewClusterParams.ServiceNetworks,
@@ -473,6 +474,18 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 		KubeKeyName:             kubeKey.Name,
 		KubeKeyNamespace:        kubeKey.Namespace,
 		TriggerMonitorTimestamp: time.Now(),
+	}
+
+	if cluster.SchedulableMasters != nil {
+		cluster.UseSchedulingDefaults = swag.Bool(false)
+	}
+	if swag.BoolValue(cluster.UseSchedulingDefaults) {
+		var schedulableMasters bool
+		schedulableMasters, err = b.providerRegistry.GetActualSchedulableMasters(&cluster)
+		if err != nil {
+			return nil, common.NewApiError(http.StatusInternalServerError, err)
+		}
+		cluster.SchedulableMasters = swag.Bool(schedulableMasters)
 	}
 
 	pullSecret := swag.StringValue(params.NewClusterParams.PullSecret)
@@ -1934,10 +1947,28 @@ func (b *bareMetalInventory) updateClusterData(_ context.Context, cluster *commo
 			return common.NewApiError(http.StatusBadRequest, errors.Errorf(msg))
 		}
 	}
+
+	if params.ClusterUpdateParams.UseSchedulingDefaults != nil {
+		useSchedulingDefaults := swag.BoolValue(params.ClusterUpdateParams.UseSchedulingDefaults)
+		updates["use_scheduling_defaults"] = useSchedulingDefaults
+		if useSchedulingDefaults {
+			var schedulableMasters bool
+			schedulableMasters, err = b.providerRegistry.GetActualSchedulableMasters(cluster)
+			if err != nil {
+				msg := fmt.Sprintf("Can't update 'use_scheduling_defaults' to '%t' for cluster %s", useSchedulingDefaults, cluster.ID)
+				log.Error(msg)
+				return common.NewApiError(http.StatusInternalServerError, errors.Errorf(msg))
+			}
+			updates["schedulable_masters"] = schedulableMasters
+			b.setUsage(false, usage.SchedulableMasters, nil, usages)
+		}
+	}
+
 	if params.ClusterUpdateParams.SchedulableMasters != nil {
-		value := swag.BoolValue(params.ClusterUpdateParams.SchedulableMasters)
-		updates["schedulable_masters"] = value
-		b.setUsage(value, usage.SchedulableMasters, nil, usages)
+		schedulableMasters := swag.BoolValue(params.ClusterUpdateParams.SchedulableMasters)
+		updates["schedulable_masters"] = schedulableMasters
+		updates["use_scheduling_defaults"] = false
+		b.setUsage(true, usage.SchedulableMasters, nil, usages)
 	}
 
 	if params.ClusterUpdateParams.DiskEncryption != nil {
@@ -4627,12 +4658,38 @@ func (b *bareMetalInventory) BindHostInternal(ctx context.Context, params instal
 		return nil, common.NewApiError(http.StatusInternalServerError, err)
 	}
 
+	if err = b.refreshSchedulableMasters(ctx, cluster); err != nil {
+		log.WithError(err).Errorf("Failed to refresh schedulable_masters while binding host <%s> to cluster <%s>",
+			params.HostID, *params.BindHostParams.ClusterID)
+		return nil, common.NewApiError(http.StatusInternalServerError, err)
+	}
+
 	host, err = common.GetHostFromDB(b.db, params.InfraEnvID.String(), params.HostID.String())
 	if err != nil {
 		return nil, common.NewApiError(http.StatusInternalServerError, err)
 	}
 
 	return host, nil
+}
+
+func (b *bareMetalInventory) refreshSchedulableMasters(ctx context.Context, cluster *common.Cluster) error {
+	if swag.BoolValue(cluster.UseSchedulingDefaults) {
+		schedulableMasters, err := b.providerRegistry.GetActualSchedulableMasters(cluster)
+		if err != nil {
+			return errors.Errorf("Failed get actual schedulable masters for cluster %s", cluster.ID)
+		}
+		clusterUpdateParams := installer.V2UpdateClusterParams{
+			ClusterID: *cluster.ID,
+			ClusterUpdateParams: &models.V2ClusterUpdateParams{
+				SchedulableMasters: swag.Bool(schedulableMasters),
+			},
+		}
+		_, err = b.UpdateClusterNonInteractive(ctx, clusterUpdateParams)
+		if err != nil {
+			return errors.Errorf("Failed to refresh schedulable_masters on cluster %s", cluster.ID)
+		}
+	}
+	return nil
 }
 
 func (b *bareMetalInventory) UnbindHostInternal(ctx context.Context, params installer.UnbindHostParams) (*common.Host, error) {
@@ -4646,6 +4703,11 @@ func (b *bareMetalInventory) UnbindHostInternal(ctx context.Context, params inst
 	}
 	if host.ClusterID == nil {
 		return nil, common.NewApiError(http.StatusConflict, errors.Errorf("Host %s is already unbound", params.HostID))
+	}
+
+	cluster, err := common.GetClusterFromDB(b.db, *host.ClusterID, common.SkipEagerLoading)
+	if err != nil {
+		return nil, common.NewApiError(http.StatusInternalServerError, errors.Errorf("Failed to find cluster %s", host.ClusterID))
 	}
 
 	infraEnv, err := common.GetInfraEnvFromDB(b.db, params.InfraEnvID)
@@ -4664,6 +4726,12 @@ func (b *bareMetalInventory) UnbindHostInternal(ctx context.Context, params inst
 
 	if _, err = b.refreshClusterStatus(ctx, host.ClusterID, b.db); err != nil {
 		log.WithError(err).Warnf("Failed to refresh cluster after unbind of host <%s>", params.HostID)
+	}
+
+	if err = b.refreshSchedulableMasters(ctx, cluster); err != nil {
+		log.WithError(err).Errorf("Failed to refresh schedulable_masters while unbinding host <%s> from cluster <%s>",
+			params.HostID, *host.ClusterID)
+		return nil, common.NewApiError(http.StatusInternalServerError, err)
 	}
 
 	host, err = common.GetHostFromDB(b.db, params.InfraEnvID.String(), params.HostID.String())
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -125,11 +125,13 @@ func mockClusterRegisterSteps() {
 	mockVersions.EXPECT().GetReleaseImage(gomock.Any(), gomock.Any()).Return(common.TestDefaultConfig.ReleaseImage, nil).Times(1)
 	mockOperatorManager.EXPECT().GetSupportedOperatorsByType(models.OperatorTypeBuiltin).Return([]*models.MonitoredOperator{&common.TestDefaultConfig.MonitoredOperator}).Times(1)
 	mockProviderRegistry.EXPECT().SetPlatformUsages(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
+	mockProviderRegistry.EXPECT().GetActualSchedulableMasters(gomock.Any()).AnyTimes()
 }
 
 func mockClusterRegisterSuccess(withEvents bool) {
 	mockClusterRegisterSteps()
 	mockMetric.EXPECT().ClusterRegistered(common.TestDefaultConfig.ReleaseVersion, gomock.Any(), gomock.Any()).Times(1)
+	mockProviderRegistry.EXPECT().GetActualSchedulableMasters(gomock.Any()).AnyTimes()
 
 	if withEvents {
 		mockEvents.EXPECT().SendClusterEvent(gomock.Any(), eventstest.NewEventMatcher(
@@ -158,6 +160,13 @@ func mockInfraEnvUpdateSuccess() {
 	mockIgnitionBuilder.EXPECT().FormatDiscoveryIgnitionFile(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(discovery_ignition_3_1, nil).Times(1)
 	mockEvents.EXPECT().SendInfraEnvEvent(gomock.Any(), eventstest.NewEventMatcher(
 		eventstest.WithNameMatcher(eventgen.ImageInfoUpdatedEventName))).AnyTimes()
+}
+
+func mockInfraEnvUpdateSuccessNoImageGeneration() {
+	mockS3Client.EXPECT().UpdateObjectTimestamp(gomock.Any(), gomock.Any()).Return(true, nil).Times(1)
+	mockS3Client.EXPECT().GetObjectSizeBytes(gomock.Any(), gomock.Any()).Return(int64(100), nil).Times(1)
+	mockProviderRegistry.EXPECT().GetActualSchedulableMasters(gomock.Any()).AnyTimes()
+>>>>>>> d2145df4 (MGMT-8089: Reimplement the schedulable masters feature)
 }
 
 func mockInfraEnvDeRegisterSuccess() {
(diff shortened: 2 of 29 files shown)


filanov:
@avishayt @rollandf @danielerez please help with the review

internal/provider/baremetal/inventory.go
```diff
@@ -13,10 +17,54 @@ func (p *baremetalProvider) CleanPlatformValuesFromDBUpdates(_ map[string]interf
...
+// property in case it was actively set through the API. Otherwise it returns
+// the expected default value depending on the number of hosts in the cluster.
+func (p *baremetalProvider) GetActualSchedulableMasters(cluster *common.Cluster) (bool, error) {
+	if cluster == nil {
+		return false, errors.New("unexpected 'nil' cluster")
+	}
+	if cluster.SchedulableMasters != nil {
+		return *cluster.SchedulableMasters, nil
+	}
+	if common.CountActiveHosts(cluster) < 5 {
```
danielerez, line 30: * try using cluster.ReadyHostCount.
* probably worth extracting '5' to a const and explain that restriction in scheduling.

internal/common/common.go
```diff
@@ -320,3 +316,16 @@ func VerifyCaBundle(pemCerts []byte) error {
 	}
 	return nil
 }
+
+// CountActiveHosts returns the number of non-disabled hosts for a given cluster
+func CountActiveHosts(cluster *Cluster) int {
```
danielerez, line 321: try using cluster.ReadyHostCount instead. it should already contain this value.

internal/provider/baremetal/inventory.go
```diff
@@ -13,10 +17,54 @@ func (p *baremetalProvider) CleanPlatformValuesFromDBUpdates(_ map[string]interf
...
+
 func (p *baremetalProvider) SetPlatformUsages(
 	_ *models.Platform,
 	usages map[string]models.Usage,
 	usageApi usage.API) error {
 	usageApi.Remove(usages, usage.PlatformSelectionUsage)
 	return nil
 }
+
+func (p baremetalProvider) GenerateManifests(cluster *common.Cluster) (map[string][]byte, error) {
```
danielerez, line 44: this func seems similar in vsphere in ovirt providers. can we extract it to a common location.

internal/provider/baremetal/inventory.go
```diff
@@ -13,10 +17,54 @@ func (p *baremetalProvider) CleanPlatformValuesFromDBUpdates(_ map[string]interf
...
+		return nil, err
+	}
+	if filename == "" {
+		return nil, errors.New("template filename cannot be empty")
+	}
+	manifests[filename] = content
+	return manifests, nil
+}
+
+func (p baremetalProvider) generateSchedulableMastersManifest(cluster *common.Cluster) (string, []byte, error) {
```
danielerez, line 57: same

internal/bminventory/inventory.go
```diff
@@ -1719,6 +1716,11 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params
 		return nil, common.NewApiError(http.StatusInternalServerError, errors.New("Failed to generated additional cluster manifest"))
 	}
 
+	if err = b.providerRegistry.GenerateProviderManifests(ctx, cluster); err != nil {
```
danielerez, line 1719: shouldn't we keep this call in GenerateAdditionalManifests?

internal/bminventory/inventory_test.go
```diff
@@ -10979,31 +10996,6 @@ var _ = Describe("TestRegisterCluster", func() {
...
-
-		reply := bm.RegisterCluster(ctx, installer.RegisterClusterParams{
-			NewClusterParams: getDefaultClusterCreateParams(),
-		})
-		Expect(reply).Should(BeAssignableToTypeOf(installer.NewRegisterClusterCreated()))
-		actual := reply.(*installer.RegisterClusterCreated)
-		Expect(actual.Payload.SchedulableMasters).To(Equal(swag.Bool(false)))
-	})
-
-	It("SchedulableMasters non default value", func() {
```
rollandf, line 10994: This should still work, no?

internal/provider/vsphere/inventory.go
```diff
@@ -47,3 +63,31 @@ func (p *vsphereProvider) SetPlatformUsages(
...
+		return nil, err
+	}
+	if filename == "" {
+		return nil, errors.New("template filename cannot be empty")
+	}
+	manifests[filename] = content
+	return manifests, nil
+}
+
+func (p *vsphereProvider) generateSchedulableMastersManifest(cluster *common.Cluster) (string, []byte, error) {
```
rollandf, line 80: This methods seem duplicated 3 times.
Can we have a way to have it in only one place?

rollandf (changes requested):

internal/provider/baremetal/inventory.go
filanov: It's used in get cluster as well, so if we start the installation you probably want to get the right reply there as well. 
`ReadyHostCount` count hosts only in known state.

internal/bminventory/inventory_test.go
```diff
@@ -123,11 +123,13 @@ func mockClusterRegisterSteps() {
 	mockVersions.EXPECT().GetReleaseImage(gomock.Any(), gomock.Any()).Return(common.TestDefaultConfig.ReleaseImage, nil).Times(1)
 	mockOperatorManager.EXPECT().GetSupportedOperatorsByType(models.OperatorTypeBuiltin).Return([]*models.MonitoredOperator{&common.TestDefaultConfig.MonitoredOperator}).Times(1)
 	mockProviderRegistry.EXPECT().SetPlatformUsages(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
+	mockProviderRegistry.EXPECT().GetActualSchedulableMasters(gomock.Any()).AnyTimes()
```
filanov, line 126: What about testing bad flows?

internal/provider/registry/registry.go
```diff
@@ -174,10 +194,59 @@ func (r *registry) PostCreateManifestsHook(cluster *common.Cluster, envVars *[]s
...
+		ClusterID: *cluster.ID,
+		CreateManifestParams: &models.CreateManifestParams{
+			Content:  swag.String(base64.StdEncoding.EncodeToString(content)),
+			FileName: &filename,
+			Folder:   swag.String(models.ManifestFolderOpenshift),
+		},
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to create manifest %s", filename)
```
filanov, line 248: suggesting to wrap the error

internal/provider/baremetal/inventory.go
danielerez: so EnabledHostCount then? https://github.com/openshift/assisted-service/blob/b9ab5d3c77580ab93c29bb43cb78cc6b02d54e88/internal/common/db.go#L361

filanov: yes that will be better

mmartinv: done

internal/common/common.go
mmartinv: done

internal/provider/baremetal/inventory.go
mmartinv: done

internal/provider/baremetal/inventory.go
mmartinv: done

internal/bminventory/inventory.go
mmartinv: I moved this to the cluster.GenerateAdditionalManifests method

internal/provider/vsphere/inventory.go
mmartinv: done

internal/provider/registry/registry.go
mmartinv: done

internal/bminventory/inventory_test.go
mmartinv: Unit tests are not implemented yet, I will do it in my next push, hopefully

filanov: Still relevant no?

internal/cluster/cluster.go
```diff
@@ -1160,16 +1163,17 @@ func (m *Manager) GenerateAdditionalManifests(ctx context.Context, cluster *comm
...
 		return errors.Wrap(err, "failed to add operator manifests")
 	}
-	if err := m.manifestsGeneratorAPI.AddTelemeterManifest(ctx, log, cluster); err != nil {
-		return errors.Wrap(err, "failed to add telemeter manifest")
-	}
 
-	if common.AreMastersSchedulable(cluster) {
-		if err := m.manifestsGeneratorAPI.AddSchedulableMastersManifest(ctx, log, cluster); err != nil {
-			return errors.Wrap(err, "failed to add schedulable masters manifest")
+	if m.providerRegistry != nil {
```
filanov, line 1167: in which case should it be nil?

internal/cluster/cluster_test.go
```diff
@@ -2372,6 +2375,7 @@ var _ = Describe("GenerateAdditionalManifests", func() {
 		manifestsGenerator.EXPECT().AddTelemeterManifest(ctx, gomock.Any(), &c).Return(nil)
 		manifestsGenerator.EXPECT().AddDiskEncryptionManifest(ctx, gomock.Any(), &c).Return(nil)
 		mockOperatorMgr.EXPECT().GenerateManifests(gomock.Any(), gomock.Any()).Return(nil).Times(1)
+		mockProviderRegistry.EXPECT().GenerateProviderManifests(gomock.Any(), gomock.Any()).Return(nil).Times(1)
```
filanov, line 2378: please add bad flows as well

internal/provider/ovirt/inventory.go
```diff
@@ -45,6 +45,18 @@ func (p *ovirtProvider) SetPlatformValuesInDBUpdates(
 	return nil
 }
 
+// GetActualSchedulableMasters returns the value of the SchedulableMasters cluster
+// property in case it was actively set through the API or 'false' otherwise.
+func (p *ovirtProvider) GetActualSchedulableMasters(cluster *common.Cluster) (bool, error) {
```
filanov, line 50: why is it different from `func (p *baremetalProvider) GetActualSchedulableMasters(cluster *common.Cluster) (bool, error) {` ?

internal/cluster/cluster.go
mmartinv: Ouch, you are right, I will fix it

internal/bminventory/inventory_test.go
mmartinv: Actually, the "SchedulableMasters non default value" test doesn't work because the GetActualSchedulableMasters is now mocked and it returns always 'false'

internal/provider/ovirt/inventory.go
mmartinv: In oVirt platform (and Vsphere most probably) we need to disable the scheduling of workloads on masters by default. 

This is because the oVirt provider always sets the compute replicas to 0 in the 'install-config.yaml' file to prevent the machine operator to provision additional compute machines at installation time (the compute nodes installed by the Assisted Installer will be `Nodes` only and not part of the `MachineSet`).

The side effect of doing this is that the generated manifests will enable the scheduling of workloads in the control plane (no compute nodes in the 'install-config.yaml'). Therefore, It does not matter how many nodes belong to the cluster, the scheduling of workloads on the control plane would be always enabled.

internal/bminventory/inventory_test.go
mmartinv: done

internal/provider/ovirt/inventory.go
```diff
@@ -45,6 +45,18 @@ func (p *ovirtProvider) SetPlatformValuesInDBUpdates(
...
+	if cluster == nil {
+		return false, errors.New("unexpected 'nil' cluster")
+	}
+	if cluster.SchedulableMasters != nil {
+		return *cluster.SchedulableMasters, nil
+	}
+	return false, nil
```
filanov, line 57:
```suggestion
	return swag.BoolValue(cluster.SchedulableMasters), nil
```

internal/provider/registry/registry.go
```diff
@@ -81,6 +110,12 @@ func (r *registry) Name() models.PlatformType {
 	return ""
 }
 
+func (r *registry) InitProviders(log logrus.FieldLogger) {
```
filanov, line 113: Suggesting to switch to a more dependency injection approach

internal/provider/registry/registry.go
```diff
@@ -174,10 +209,87 @@ func (r *registry) PostCreateManifestsHook(cluster *common.Cluster, envVars *[]s
...
-	providerRegistry.Register(baremetal.NewBaremetalProvider(log))
-	return providerRegistry
+func (r *registry) GetActualSchedulableMasters(cluster *common.Cluster) (bool, error) {
+	if cluster == nil || cluster.Platform == nil {
+		return false, errors.New("unable to get the platform type")
+	}
+	currentProviderStr := string(common.PlatformTypeValue(cluster.Platform.Type))
+	currentProvider, err := r.Get(currentProviderStr)
+	if err != nil {
+		return false, fmt.Errorf("error while getting the actual schedulable masters value on platform %s, error %w",
```
filanov, line 219:
```suggestion
		return false, errors.Wrapf(err, "error while getting the actual schedulable masters value on platform %s", currentProviderStr)
```

internal/provider/vsphere/inventory.go
```diff
@@ -13,6 +16,18 @@ func (p *vsphereProvider) SetPlatformValuesInDBUpdates(_ *models.Platform, _ map
...
+// GetActualSchedulableMasters returns the value of the SchedulableMasters cluster
+// property in case it was actively set through the API or 'false' otherwise.
+func (p *vsphereProvider) GetActualSchedulableMasters(cluster *common.Cluster) (bool, error) {
+	if cluster == nil {
+		return false, errors.New("unexpected 'nil' cluster")
+	}
+	if cluster.SchedulableMasters != nil {
+		return *cluster.SchedulableMasters, nil
+	}
+	return false, nil
```
filanov, line 28:
```suggestion
	return swag.BoolValue(cluster.SchedulableMasters), nil
```

flaper87 (commented):
thanks for working on this, @mmartinv 

I am going to hold this PR for now as I would like to have a chance to review it. There's an ongoing investigation (no PR/actions yet) on how to allow for other topologies like `3+1` where there are 3 schedulable masters and 1 worker.

Just want to make sure that we won't break anything and that we get the time to discuss this PR properly. It may take a few more days since this is the OCP shift week and we are all a bit distracted :) 

/hold

omertuc:
Note there are other PRs touching code in that area - 
https://github.com/openshift/assisted-service/pull/3255
https://github.com/openshift/assisted-service/pull/3247

internal/bminventory/inventory.go
```diff
@@ -3619,6 +3616,12 @@ func (b *bareMetalInventory) listClustersInternal(ctx context.Context, params in
 			// Clear this field as it is not needed to be sent via API
 			h.FreeAddresses = ""
 		}
+		schedulableMasters, err := b.providerRegistry.GetActualSchedulableMasters(c)
+		if err != nil {
+			log.WithError(err).Error("failed to retrieve the actual schedulable masters value")
+			return nil, common.NewApiError(http.StatusInternalServerError, err)
+		}
+		c.SchedulableMasters = &schedulableMasters
```
omertuc, line 3624: We don't typically make such on-request adjustments to data returned from the database, could this property be stored in the database somehow rather than computed on each request?

mmartinv: By default the "schedulable_masters" flag is not set and we need to respond with the expected value depending on the provider/number of active hosts, etc. Having the actual value saved in the database would need the database to be updated with any state change (each time the number of hosts or the provider changed...). I don't think doing the computation and saving it on every state change is better than just compute the actual value on the fly when needed.

internal/cluster/cluster_test.go
mmartinv: this is already implemented in latest push

internal/provider/ovirt/inventory.go
mmartinv: done

internal/provider/registry/registry.go
mmartinv: done

internal/provider/ovirt/inventory.go
```diff
@@ -45,6 +45,18 @@ func (p *ovirtProvider) SetPlatformValuesInDBUpdates(
 	return nil
 }
 
+// GetActualSchedulableMasters returns the value of the SchedulableMasters cluster
+// property in case it was actively set through the API or 'false' otherwise.
```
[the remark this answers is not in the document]
mmartinv: done

internal/bminventory/inventory.go
omertuc: I remember @avishayt bringing up this concern about a year ago on one of my PRs, I'm not entirely sure what's the motivation behind this constraint, maybe he can weigh in

flaper87: I think I agree with @mkowalski  and @omertuc  here. I would rather have the data properly reflected in the DB rather than dynamically computed before handing it off to `openshift-insdtall` The motivation is that we are not the ones doing the deployment itself and there are some decisions being made dynamically on the `openshift-installer` side already.

internal/provider/baremetal/inventory.go
```diff
@@ -13,6 +16,19 @@ func (p *baremetalProvider) CleanPlatformValuesFromDBUpdates(_ map[string]interf
 	return nil
 }
 
+func (p *baremetalProvider) GetActualSchedulableMasters(cluster *common.Cluster) (bool, error) {
```
flaper87, line 19: Don't mean to nitpick on naming but, why do you think we need `Actual` in the name of this function? Is there a way to get a list of schedulable masters that are not really schedulable?

flaper87 (changes requested):
Thanks for the PR and the patience. I'm sorry it took this long to get back to it. 

Could you please share an example of what the final `install-config` looks like when masters are schedulable?

internal/bminventory/inventory.go
mmartinv: > Thanks for the PR and the patience. I'm sorry it took this long to get back to it.
> 
> Could you please share an example of what the final `install-config` looks like when masters are schedulable?

This implementation does not touch/alter the `install-config.yaml`, it just adds an additional manifest (`50-schedulable_masters.yaml`) to the installation to enable/disable the scheduling.
Something like:
```
apiVersion: config.openshift.io/v1
kind: Scheduler
metadata:
  name: cluster
spec:
  mastersSchedulable: true
  policy:
    name: ""
status: {}
```

This is because the current implementation does not allow to effectively disable the scheduling if needed. It just adds and additional manifest to enable the scheduling if `schedulable_masters=true`, otherwise it does nothing and the default value generated by the 'openshift-install' command is used (which is different depending on the number of hosts in the cluster). 

The idea is to use the expected defaults for each platform if `schedulable_masters=nil` (the default) or the value saved in the database otherwise (the user explicitly wants to enable/disable the scheduling).

On the other hand, we need to disable the scheduling in **oVirt** and **vSphere** providers by default, no matter the number of hosts in the cluster. This is because we set the number of compute replicas to 0 in the install config to prevent the MCO trying to provision new worker nodes and has the side effect of enabling the scheduling on masters.

internal/provider/baremetal/inventory.go
mmartinv: "Actual" in this function name means "Get the actual value of the scheduling property" because it's not always the value contained in the database.

internal/bminventory/inventory.go
mmartinv: > I think I agree with @mkowalski and @omertuc here. I would rather have the data properly reflected in the DB rather than dynamically computed before handing it off to `openshift-insdtall` The motivation is that we are not the ones doing the deployment itself and there are some decisions being made dynamically on the `openshift-installer` side already.

One of the motivation of this change is that we need to return the real value of the `schedulable_masters` property on each call (for example the UI needs to know it to show the correct information, now it's using a workaround to fix this). That real value depends on the number of hosts and it varies over time, so lets say `schedulable_masters=nil` (the user didn't actively set the property at cluster creation time) and we have three masters. In this case the expected value is `true`. Then the user adds one more host so now the expected value is `false`. If the user configured the property, then the saved value in the database must be returned no matter how many hosts are in the cluster. 

I guess this could be done by computing the expected default value when a new host is added/removed from the cluster and updating the database accordingly or do nothing if the user explicitly set the property. Would you prefer this approach? I am not really sure about the implications of this and I would probably need some help to get this implemented in this way.

mkowalski: >so lets say schedulable_masters=nil (the user didn't actively set the property at cluster creation time) and we have three masters. In this case the expected value is true. Then the user adds one more host so now the expected value is false. If the user configured the property, then the saved value in the database must be returned no matter how many hosts are in the cluster.

We already do similar updates in the network options. This is a very natural behaviour - as the state of the cluster changes (by e.g. adding or removing hosts), some of the properties need to be recalculated. That's perfectly fine and it's not a bad pattern at all.

One more important thing here. We should have proper and reasonable defaults. If "user didn't actively set the property at cluster creation time" we should not just put `nil` to the DB, but we should agree on a default value and roll like this. You can use `disk_encryption` property as an example - we have introduced defaults there and have explicit code that handles the scenario when user did not provide value at all. But in principle, if we are introducing on/off field, we should
1) create a migration that will backfill the DB for already existing clusters
2) only allow true/false values, because `nil` has no semantics for a feature that can be either enabled or disabled

romfreiman:
/hold
@jelkosz @nirfarkas lets discuss the necessity of this?

jelkosz:
@jkilzi can you please weigh in?

jkilzi:
> /hold @jelkosz @nirfarkas let's discuss the necessity of this?

Honestly, I wasn't aware of this enhancement. But seems to be necessary.
There was a problem with the current implementation ([bug](https://issues.redhat.com/browse/MGMT-8089) open by @batzionb) and that forces the UI to hack the cluster.schedulable_master value according to the number of discovered hosts. If this PR corrects that behavior it will allow the UI to remove that bug-prone hack that it currently has.

flaper87:
@mmartinv said that this was discussed with @batzionb

batzionb:
The workaround in the UI has been working for a while, it's good enough. so no need for this anymore

flaper87:
> The workaround in the UI has been working for a while, it's good enough. so no need for this anymore

Is a workaround enough? Should this be implemented "properly"? 

I understand it's been too long since this was asked for. 

@mmartinv, if the API is not needed anymore, we may still want to move on with the changes you did on how the manifests are handled. I need to review the latest version of this PR, tho.

batzionb:
I think by now the workaround is enough

gamli75 (approved):

osherdp:
@mmartinv can you unhold? or is there something else that's needed to be changed

mmartinv:
> @mmartinv can you unhold? or is there something else that's needed to be changed

IIRC @flaper87 held it because he wanted to take a deeper look.

PR #2868, opened 2021-11-03
Bug 2019415: Cluster installation stages missing from cluster event logs

arielireni:
# Assisted Pull Request

## Description

Cluster event logs are missing in cluster installation events. I.E. stage Starting installation.

Added `cluster_id` as a property to any event that is related to the cluster in order to solve this.

Signed-off-by: arielireni <aireni@redhat.com>

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/events.yaml
+++ b/docs/events.yaml
@@ -660,6 +660,7 @@ x-events:
   properties:
     host_id: UUID
     infra_env_id: UUID
+    cluster_id: UUID_PTR
     message: string
 
 - name: host_install_progress_updated
@@ -669,6 +670,7 @@ x-events:
   properties:
     host_id: UUID
     infra_env_id: UUID
+    cluster_id: UUID_PTR
     host_name: string
     event: string
 
@@ -679,6 +681,7 @@ x-events:
   properties:
     host_id: UUID
     infra_env_id: UUID
+    cluster_id: UUID_PTR
     message: string
 
 - name: host_registration_failed
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4170,7 +4170,7 @@ func (b *bareMetalInventory) DisableHost(ctx context.Context, params installer.D
 		return common.GenerateErrorResponderWithDefault(err, http.StatusConflict)
 	}
 
-	c, err := b.refreshHostAndClusterStatuses(ctx, "disable host", &params.HostID, &params.ClusterID, tx)
+	c, err := b.refreshHostAndClusterStatuses(ctx, "disable host", host, &params.ClusterID, tx)
 	if err != nil {
 		return common.GenerateErrorResponder(err)
 	}
@@ -4231,7 +4231,7 @@ func (b *bareMetalInventory) EnableHost(ctx context.Context, params installer.En
 		return common.GenerateErrorResponderWithDefault(err, http.StatusConflict)
 	}
 
-	c, err := b.refreshHostAndClusterStatuses(ctx, "enable host", &params.HostID, &params.ClusterID, tx)
+	c, err := b.refreshHostAndClusterStatuses(ctx, "enable host", host, &params.ClusterID, tx)
 	if err != nil {
 		return common.GenerateErrorResponder(err)
 	}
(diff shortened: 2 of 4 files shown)


docs/events.yaml
```diff
@@ -658,6 +658,7 @@ x-events:
   event_type: host
   severity: "error"
   properties:
+    cluster_id: UUID_PTR
```
masayag, line 661: why UUID_PTR and not just UUID ? 
The result will be the same, but according to the calling methods, we pass one cluster-id as a reference (clusterID), and one as its value (*clusterID):

```
eventgen.SendRefreshHostOrClusterStatusesFailedEvent(ctx, b.eventsHandler, clusterID, *hostID, *clusterID, err.Error())
```
unless there is a future plan to replace the first clusterID param with infraEnvID

internal/bminventory/inventory.go
```diff
@@ -4263,13 +4263,21 @@ func (b *bareMetalInventory) refreshHostAndClusterStatuses(
 
 	var err error
 
+	infraEnv, err := common.GetInfraEnvFromDB(b.db, *clusterID)
```
arielireni: done

docs/events.yaml
arielireni: replaced clusterID with infraEnvID

internal/bminventory/inventory.go
masayag, line 4266: I don't think we should re-fetch it from the DB.
Pls have a look at the caller of `refreshHostAndClusterStatuses` and see that in both cases the host was already retrieved from the DB, therefore I'd suggest to change the signature of this function to accept the entire host struct and use the value of the infraenv-id from it.

internal/bminventory/inventory.go
```diff
@@ -4789,7 +4792,7 @@ func (b *bareMetalInventory) UpdateHostInstallProgress(ctx context.Context, para
 		}
 
 		log.Info(fmt.Sprintf("Host %s in cluster %s: %s", host.ID, host.ClusterID, event))
-		eventgen.SendHostInstallProgressUpdatedEvent(ctx, b.eventsHandler, *host.ID, *host.ClusterID, hostutil.GetHostnameForMsg(&host.Host), event)
+		eventgen.SendHostInstallProgressUpdatedEvent(ctx, b.eventsHandler, host.ClusterID, *host.ID, *host.ClusterID, hostutil.GetHostnameForMsg(&host.Host), event)
```
masayag, line 4795: looks we can replace the clusterID with infraenvid here as well

docs/events.yaml
masayag: General comment about the uniformity of the events, for host-event it will be more acceptable to have the mandatory parameters first.
meaning, the first two params should be host-id and infraenv-id, later-on the cluster-id (when needed) and then the rest of the params. WDYT ?

internal/bminventory/inventory.go
```diff
@@ -4255,12 +4255,15 @@ func (b *bareMetalInventory) EnableHost(ctx context.Context, params installer.En
...
 	eventName string,
-	hostID *strfmt.UUID,
+	host *common.Host,
 	clusterID *strfmt.UUID,
 	db *gorm.DB) (*common.Cluster, error) {
 
 	logger := logutil.FromContext(ctx, b.log)
 
+	hostID := host.ID
+	infraEnvID := host.InfraEnvID
```
masayag, line 4265: don't have that strong opinion about it, but instead of defining of a variable only for the sake of the defer, this can be set as a parameter to the defer function, then it will be relevant in the context in which it is relevant.

internal/bminventory/inventory.go
arielireni: done

internal/bminventory/inventory.go
arielireni: done

docs/events.yaml
arielireni: changed for the events related to this PR, I will create a separate PR for changing the other host-type events

nmagnezi: Agreed. Please file a Jira ticket for this. Otherwise, PR looks good.

arielireni: https://issues.redhat.com/browse/MGMT-8188

PR #2872, opened 2021-11-03
MGMT-8129: Upgrade go-swagger to v0.28.0

ori-amizur:
# Assisted Pull Request

## Description

- Port to latest go-swagger
- This change causes all refs in swagger to become pointers
- The x-go-custom-tag should now be part of the referenced entity in
  case of $ref is used
- Added mechanism to remove 'Dash' and 'Dot' in generated symbols to be
  backward compatible and clearer symbols for golang

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


filanov:
/hold 
@ori-amizur wouldn't https://github.com/go-swagger/go-swagger/issues/2515 break the clients?

filanov:
OK i saw that you resolved it by https://github.com/openshift/assisted-service/pull/2872/files#diff-cd3c90d4f4561232dd163f1783d1fe993d24389537eda6e615219a98ca05716cR57 
Not sure how i feel about it

internal/bminventory/inventory.go
```diff
@@ -545,7 +545,6 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 			OpenshiftVersion:      *releaseImage.Version,
 			OcpReleaseImage:       *releaseImage.URL,
 			SSHPublicKey:          params.NewClusterParams.SSHPublicKey,
-			UpdatedAt:             strfmt.DateTime{},
```
filanov, line 548: why is that removed?

internal/cluster/cluster_test.go
```diff
@@ -2491,6 +2491,8 @@ var _ = Describe("Deregister inactive clusters", func() {
 		inactiveCluster2 := registerCluster()
 		inactiveCluster3 := registerCluster()
 
+		// To verify that lastActive is greater than the updatedAt field of inactiveCluster3
+		time.Sleep(time.Millisecond)
```
filanov, line 2495: why not use eventually? sleep can lead to flaky tests

internal/bminventory/inventory.go
ori-amizur: It is not needed since we set here default value.

internal/common/common.go
```diff
@@ -255,3 +255,36 @@ func GetHostNTPSources(db *gorm.DB, host *models.Host) (string, error) {
...
+func LogStatePtr(l models.LogsState) *models.LogsState {
+	return &l
+}
+
+func LogStateValue(l *models.LogsState) models.LogsState {
+	if l == nil {
+		return ""
+	}
+	return *l
+}
```
carbonin, line 290: I'd rather this not exist, but if it has to exist, can it go into its own file or package for type manipulations?

internal/migrations/20210713123129_populate_infra_env.go
```diff
@@ -16,8 +17,9 @@ func populateInfraEnv() *gormigrate.Migration {
 			if err := tx.Exec("ALTER TABLE hosts ADD COLUMN IF NOT EXISTS infra_env_id text NULL;").Error; err != nil {
 				return err
 			}
+
 			// Populate the infra_env_id column
-			if err := tx.Exec("UPDATE hosts SET infra_env_id=cluster_id;").Error; err != nil {
+			if err := tx.Exec("UPDATE hosts SET infra_env_id=cluster_id where infra_env_id is null or infra_env_id = '';").Error; err != nil {
```
carbonin, line 22: Why does this need to change?

models/cluster.go
```diff
@@ -151,7 +155,7 @@ type Cluster struct {
 	Kind *string `json:"kind"`
 
 	// The progress of log collection or empty if logs are not applicable
-	LogsInfo LogsState `json:"logs_info,omitempty" gorm:"type:varchar(2048)"`
+	LogsInfo LogsState `json:"logs_info,omitempty"`
```
carbonin, line 158: This will need a migration, right?

I think this is the same issue we ran into with ssh key.

I'm not reviewing all of the generated files, but we'll want to be sure we don't have any other changes like this (I just happened to open this one)

cc @filanov

internal/migrations/20210713123129_populate_infra_env.go
ori-amizur: It is incorrect to set the infra_env_id during migration if it already has a value

internal/migrations/20210713123129_populate_infra_env_test.go
```diff
@@ -55,6 +55,7 @@ var _ = Describe("populateInfraEnv", func() {
 		}
 		err := db.Create(&host).Error
 		Expect(err).NotTo(HaveOccurred())
+		Expect(db.Model(&models.Host{}).Where("id = ? and infra_env_id = ?", hostID.String(), host.InfraEnvID.String()).UpdateColumn("infra_env_id", "").Error).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
ori-amizur: The infra-env-id is created is not null column, but when setting it as empty string ("") it attempts to set null and we receive db error.

swagger.yaml
```diff
@@ -8992,6 +9044,7 @@ definitions:
 
   subnet:
     type: string
+    x-go-custom-tag: gorm:"primary_key"
```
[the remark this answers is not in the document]
ori-amizur: The subnet is part of other tables. Now the x-custom-tag fields have to be on the referenced object.. Otherwise go-swagger ignores them.

internal/common/common.go
ori-amizur: OK

internal/bminventory/inventory.go
```diff
@@ -2910,7 +2908,7 @@ func (b *bareMetalInventory) updateNetworkTables(db *gorm.DB, cluster *common.Cl
 
 	// TODO: Update machine CIDR only if necessary
 	// The machine cidr can be resetted, calculated and provided by the user
-	if err = db.Where("cluster_id = ?", *cluster.ID).Delete(&models.MachineNetwork{}).Error; err != nil {
+	if err = db.Unscoped().Where("cluster_id = ?", *cluster.ID).Delete(&models.MachineNetwork{}).Error; err != nil {
```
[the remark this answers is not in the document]
ori-amizur: Will remove

models/cluster.go
ori-amizur: This is a swagger change that I missed

internal/migrations/20210713123129_populate_infra_env.go
carbonin: How is this related to the go-swagger update though?

Generally I don't want to edit migrations unless we have a very compelling reason because they've already run in some environments and will not run again.

For example, I'd really expect to see a separate bug to make a change like this rather than slipping it into a PR about updating go-swagger.

internal/migrations/20210713123129_populate_infra_env_test.go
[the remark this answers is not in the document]
carbonin: > but when setting it as empty string ("") it attempts to set null and we receive db error.

This is very strange to me. Empty string and NULL are very different values. Is this something gorm is doing?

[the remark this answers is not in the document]
ori-amizur: gorm set default values to null values.  In this case the default value is empty

internal/migrations/20210713123129_populate_infra_env.go
ori-amizur: Actually I tackled into the problem while working on this.  You are right it is not a go-swagger issue.  Will remove

hack/generate.sh
```diff
@@ -56,6 +54,9 @@ function generate_from_swagger() {
     lint_swagger
     generate_go_client
     generate_go_server
+    for f in models/* ; do
+	    sed -i 's/Dash//g;s/Dot//g' $f
```
omertuc, line 58: Nit: This line has mixed indentation

ori-amizur: Changed

hack/generate.sh
```diff
@@ -56,6 +54,9 @@ function generate_from_swagger() {
     lint_swagger
     generate_go_client
     generate_go_server
+    for f in models/* ; do
+        sed -i 's/Dash//g;s/Dot//g' $f
```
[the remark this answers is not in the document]
omertuc: Also we may want to verify that our swagger doesn't contain the words "dash" and "dot" in enums literally, because that would be seriously confusing when they suddenly disappear

internal/migrations/20210713123129_populate_infra_env_test.go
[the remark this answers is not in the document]
carbonin: Even then I still don't understand why this is being added in a patch to upgrade go-swagger.

swagger.yaml
```diff
@@ -7647,7 +7676,6 @@ definitions:
       logs_info:
         $ref: '#/definitions/logs_state'
         description: The progress of log collection or empty if logs are not applicable
-        x-go-custom-tag: gorm:"type:varchar(2048)"
```
carbonin, line 7650: This should be removed for host as well, right?

swagger.yaml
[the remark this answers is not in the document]
carbonin: Hopefully we never have to use a subnet in something where it's not the primary key I guess :man_shrugging: 

Something to look out for in the future.

swagger.yaml
ori-amizur: Seems so. Yes

internal/migrations/20210713123129_populate_infra_env_test.go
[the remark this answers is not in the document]
ori-amizur: Removed

carbonin:
/lgtm

@filanov @ori-amizur is the reason for the hold resolved?

ori-amizur:
> /lgtm
> 
> @filanov @ori-amizur is the reason for the hold resolved?

https://github.com/openshift/assisted-service/pull/2872#issuecomment-962568902

carbonin:
I saw that comment, but the last word on it seems to have been "Not sure how i feel about it"
So I wanted to get @filanov 's approval as well.

ori-amizur:
> I saw that comment, but the last word on it seems to have been "Not sure how i feel about it" So I wanted to get @filanov 's approval as well.

@filanov said it is OK to approve
/unhold

PR #2884, opened 2021-11-05
MGMT-7254: Deploy and use image service for subsystem

carbonin:
## Description

This PR adds the image service deployment spec, script, and make
target. It also configures assisted service to direct image downloads to
the image service rather than generating the images and uploading them
to s3.

It also fixes issues in code and subsystem tests to handle the new image creation
strategy. See individual commits for details.

- [x] New Feature
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

1. Deployed using `skipper make deploy-test` to minikube
2. set up rhsso user in wiremock
3. Created cluster
4. Called generate iso endpoint
5. Successfully downloaded iso from image service using url in cluster image info

## Assignees

/cc @osherdp 
/cc @filanov 

## Checklist

## Reviewers Checklist


carbonin:
Also I still need to review the docs to see if anything needs updating.
Ideally with this and deploying the image service to the cloud we won't be using the generation anywhere and can start removing that. This step will clearly need a bunch of documentation updates, but maybe we want to do some if it here.

carbonin:
> Opening it a bit early to see what all blows up

Everything ... the answer is all the test infra jobs that use these scripts :man_facepalming: :sweat_smile:

internal/bminventory/inventory.go
```diff
@@ -1464,6 +1464,9 @@ func (b *bareMetalInventory) createAndUploadNewImage(ctx context.Context, log lo
 			return err
 		}
 
+		details := b.getIgnitionConfigForLogging(ctx, infraEnv, log, imageType)
+		eventgen.SendIgnitionConfigImageGeneratedEvent(ctx, b.eventsHandler, *infraEnv.ID, details)
```
carbonin, line 1468: I'm not sure if we really want this. But I had to either change the code or tests to get it working.

@filanov @avishayt what do you think? Should we still have an event here if we're not actually "generating" anything?

carbonin:
It looks like the subsystem test is failing because we do more validation on the nmstate configs when we actually generate the image vs when we're just saving those configs to the database.

I figure I can expand the validations we do pretty easily, then fix the test.

I looked into the e2e test for a bit, but I'm still not sure why we are showing `<pending>` for the image service host in just that case.

carbonin:
Additionally, after talking with QE about how they're deploying assisted to ocp I'll also need to update this PR to ensure that we're creating an ingress for the image service like we're doing for assisted [here](https://github.com/openshift/assisted-service/blob/140fe0cc2726d73077e6737a8a65f178d2ea6875/tools/deploy_inventory_service.py#L41-L79)

internal/bminventory/inventory.go
filanov: Maybe changing the event, something like new image is ready or something like that.

subsystem/image_test.go
```diff
@@ -39,48 +39,21 @@ var _ = Describe("system-test image tests", func() {
 		clearDB()
 	})
 
-	It("create_and_download_live_iso", func() {
```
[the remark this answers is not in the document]
filanov: If there are other tests that check image download i think that this test can be removed. 
The download api should return an error.

filanov:
> It looks like the subsystem test is failing because we do more validation on the nmstate configs when we actually generate the image vs when we're just saving those configs to the database.
> 
> I figure I can expand the validations we do pretty easily, then fix the test.
> 
> I looked into the e2e test for a bit, but I'm still not sure why we are showing `<pending>` for the image service host in just that case.

Pending on what?

internal/bminventory/inventory.go
carbonin: So create a new event? Right now the message is in the generated events code, right?

subsystem/image_test.go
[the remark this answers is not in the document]
carbonin: There are no other tests that check for downloading this specific image, but all the rest of the tests in this file are about downloading the discover iso.

I'll make the change to make this endpoint return an error and test for it here.

carbonin:
> Pending on what?

`<pending>` is literally what we get in the hostname of the image service. It think it's a problem when we can't detect the ip of the minikube instance or something.

This is the specific error line:
`requests.exceptions.ConnectionError: HTTPConnectionPool(host='%3cpending%3e', port=8080): Max retries exceeded with url: /images/7fbae65a-b9b8-4213-89a5-45ddd48967ae?api_key=eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJpbmZyYV9lbnZfaWQiOiI3ZmJhZTY1YS1iOWI4LTQyMTMtODlhNS00NWRkZDQ4OTY3YWUifQ.UUkl6uoNE5Iyj9gLA_a6fOpdXx6vZBpctqpMY0xk6QGpUmi6C-tqi7gmJohwvsJnk7sGlx3UUWMHSWOBKor0JA&arch=x86_64&type=minimal-iso&version=4.8 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f564a1a0e80>: Failed to establish a new connection: [Errno -2] Name or service not known',))
`

internal/bminventory/inventory.go
filanov: yes

filanov:
> > Pending on what?
> 
> `<pending>` is literally what we get in the hostname of the image service. It think it's a problem when we can't detect the ip of the minikube instance or something.
> 
> This is the specific error line: `requests.exceptions.ConnectionError: HTTPConnectionPool(host='%3cpending%3e', port=8080): Max retries exceeded with url: /images/7fbae65a-b9b8-4213-89a5-45ddd48967ae?api_key=eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJpbmZyYV9lbnZfaWQiOiI3ZmJhZTY1YS1iOWI4LTQyMTMtODlhNS00NWRkZDQ4OTY3YWUifQ.UUkl6uoNE5Iyj9gLA_a6fOpdXx6vZBpctqpMY0xk6QGpUmi6C-tqi7gmJohwvsJnk7sGlx3UUWMHSWOBKor0JA&arch=x86_64&type=minimal-iso&version=4.8 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f564a1a0e80>: Failed to establish a new connection: [Errno -2] Name or service not known',)) `

Maybe it's a race in the deployment? 
You apply the resource but k8s may take some time until the resource is actually ready

carbonin:
TODO on this PR still:

- Create a new event so that it makes sense when were using the image service
- Handle the other deployment targets
  - Currently the main one would be `oc-ingress` which is what QE uses when deploying to their environments.

That said, all the tests on this PR _should_ be passing now - or at least I don't know of any reason they wouldn't :sweat_smile:

carbonin:
Looks like lint failed building the image:
```
github.com/openshift/assisted-service/client imports
	github.com/go-openapi/strfmt imports
	go.mongodb.org/mongo-driver/bson: go.mongodb.org/mongo-driver@v1.5.1: stream error: stream ID 131; INTERNAL_ERROR
github.com/openshift/assisted-service/client imports
	github.com/go-openapi/strfmt imports
	go.mongodb.org/mongo-driver/bson/bsontype: go.mongodb.org/mongo-driver@v1.5.1: stream error: stream ID 131; INTERNAL_ERROR
github.com/openshift/assisted-service/client imports
	github.com/go-openapi/strfmt imports
	go.mongodb.org/mongo-driver/bson/primitive: go.mongodb.org/mongo-driver@v1.5.1: stream error: stream ID 131; INTERNAL_ERROR
```

/test lint

carbonin:
It looks like infra-env events are [not coming back as a part of the list events API](https://github.com/openshift/assisted-service/blob/03371a059d91942126f8e56c3bb20e2612341596/internal/events/event.go#L178).

Was that intentional? Should we open an issue to include them or make this event a cluster-focused one even though it really isn't?

This is why the subsystem tests were failing so for now I just removed the check for the image event.

cmd/main.go
```diff
@@ -482,6 +482,8 @@ func main() {
 			} else {
 				failOnError(uploadISOs(objectHandler, versionHandler, log), "Failed to upload boot files")
 			}
+		} else {
+			log.Infof("Skipping ISO upload, image service running at %s", bm.ImageServiceBaseURL)
```
eranco74, line 486: Does this mean the assisted-service will enable the API and report ready regardless of the image-service status and the availability of the images?

eranco74: OK, I see you adressed it in `wait_for_assisted_service.py`
Yet we might want to reflect the service isn't really functional as long the image-service images aren't avilable

eranco74 (commented):
/lgtm

cmd/main.go
carbonin: I mean it is to a degree. If you wanted to, you could create an infra-env then the request to download the image would just fail.

The assisted service doesn't directly depend on the image service (i.e. nothing will fail in the assisted service because the image service isn't up).

pkg/staticnetworkconfig/generator.go
```diff
@@ -204,7 +204,12 @@ func (s *StaticNetworkConfigGenerator) validateMacInterfaceName(hostIdx int, mac
 }
 
 func (s *StaticNetworkConfigGenerator) validateNMStateYaml(ctx context.Context, networkYaml string) error {
-	_, err := s.executeNMStatectl(ctx, networkYaml)
+	result, err := s.executeNMStatectl(ctx, networkYaml)
+	if err != nil {
+		return err
+	}
+
+	_, err = s.createNMConnectionFiles(result, "temphostdir")
```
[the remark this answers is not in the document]
carbonin: This isn't actually a directory we create locally, it's just used to render the path in the returned files which we don't care about in this case.

tools/wait_for_assisted_service.py
```diff
@@ -60,6 +59,18 @@ def main():
         expected_exceptions=(requests.exceptions.ConnectionError, requests.exceptions.ReadTimeout),
         sleep_seconds=SLEEP, waiting_for="assisted-service to be healthy")
 
+    waiting.wait(
+        lambda: is_service_ready(
+            service="assisted-image-service",
+            path="/health",
```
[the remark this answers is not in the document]
carbonin: This is the readiness endpoint.
https://github.com/openshift/assisted-image-service/blob/c53b4ee0adb81a6c581af30071bda2093c31f794/main.go#L71

pkg/staticnetworkconfig/generator.go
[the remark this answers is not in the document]
filanov: Then just please add a comment that will explain that files are not actually created on the filesystem

tools/wait_for_assisted_service.py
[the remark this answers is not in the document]
filanov: Very confusing, i would expect that health to report to livenessProbe and ready to reply to readinessProbe 
But it's out of context in this PR. so it's fine

osherdp (approved):

filanov:
@osherdp @carbonin if there is a commit that add the events maybe we can take only that?

PR #2887, opened 2021-11-07
Bug 2020928: Changing the condition that check if disk-encryption is enabled.

ybettan:
# Assisted Pull Request

## Description

Until now we were just checking that the disk-encryption object is
empty, but we recently added default values to disk-encryption feature
that always fill the object with data in such way that the 'enable_on'
field is set to 'none' by default, therefore, the condition for disabled
disk-encryption has changed.

Signed-off-by: Yoni Bettan <ybettan@redhat.com>

## List all the issues related to this PR

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [x] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- [ ] Are the title and description (in both PR and commit) meaningful and clear?
- [ ] Is there a bug required (and linked) for this change?
- [ ] Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #2889, opened 2021-11-07
NO-ISSUE: changing the condition that check if disk-encryption is enabled

osherdp:
Backported from #2887 

Until now we were just checking that the disk-encryption object is
empty, but we recently added default values to disk-encryption feature
that always fill the object with data in such a way that the 'enable_on'
field is set to 'none' by default, therefore, the condition for disabled
disk-encryption has changed.

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -15,7 +15,6 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	reflect "reflect"
 	"sort"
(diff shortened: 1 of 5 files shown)


osherdp:
/override ci/prow/e2e-metal-assisted
/override ci/prow/e2e-metal-assisted-olm
/override ci/prow/e2e-metal-assisted-single-node
(all are currently broken jobs in this branch)

PR #2892, opened 2021-11-07
MGMT-8167: Move `assisted_service_api.py` from test_infra to be included in the service

eliorerz:
The `assisted_service_api.py` is basically part of `assisted_service_client` python library.
Now it is also needed to be as part as `assisted-events-scrap` project.
After this PR is merged, https://github.com/openshift/assisted-test-infra/blob/master/discovery-infra/test_infra/assisted_service_api.py and https://github.com/openshift-assisted/assisted-events-scrap/blob/master/ai_client/assisted_service_api.py will be removed and deprecated

- [ ] Cloud
- [ ] None

## Assignees

/cc @osherdp 
/cc @michaellevy101

### DIFF
--- a/hack/generate.sh
+++ b/hack/generate.sh
@@ -33,8 +33,9 @@ function generate_python_client() {
         -v "${__root}"/tools/generate_python_client.sh:/script.sh:ro,Z \
         -e SWAGGER_FILE=/swagger.yaml -e OUTPUT=/local/assisted-service-client/ \
         quay.io/ocpmetal/swagger-codegen-cli:2.4.15 /script.sh
-     cd "${dest}"/assisted-service-client/ && python3 setup.py sdist --dist-dir "${dest}"
-     cd "${dest}"/assisted-service-client/ && python3 "${__root}"/tools/client_package_initializer.py "${dest}"/assisted-service-client/  https://github.com/openshift/assisted-service --build
+     cd "${dest}"/assisted-service-client/ && python3 "${__root}"/tools/client_package_initializer.py "${dest}"/assisted-service-client/  https://github.com/openshift/assisted-service "${__root}"/tools/python_api_client_extra_files/ --build
+     rm "${dest}"/assisted-service-client-1.0.0.tar.gz
+     cp "${dest}"/assisted-service-client/dist/assisted-service-client-*.tar.gz "${dest}"
 }
 
 function generate_mocks() {
--- a/tools/client_package_initializer.py
+++ b/tools/client_package_initializer.py
@@ -1,7 +1,9 @@
 import argparse
 import os
+import shutil
 import subprocess
-from typing import Dict
+from pathlib import Path
+from typing import Dict, List
 
 
 class Setup:
@@ -59,8 +61,9 @@ class SetupInitializer:
     DEFAULT_VERSION_PATH = "version.txt"
     DEFAULT_README_PATH = "README.md"
 
-    def __init__(self, project_path: str, url: str) -> None:
+    def __init__(self, project_path: str, url: str, client_added_files: str) -> None:
         self._project_path = project_path
+        self._client_added_files_dir = client_added_files
         self._setup_path = os.path.join(project_path, self.DEFAULT_SETUP_PATH)
         self._url = url
         self._version = None
@@ -73,6 +76,10 @@ class SetupInitializer:
 
         with open(os.path.join(self._project_path, "MANIFEST.in"), "w") as f:
             f.write(f"include {self.DEFAULT_README_PATH}")
+
+    def _copy_client_extra_modules(self) -> None:
+        for src_file in Path(self._client_added_files_dir).glob("*.py"):
+            shutil.copy(src_file, os.path.join(self._project_path, "assisted_service_client"))
 
     def _load_readme(self) -> None:
         with open(os.path.join(self._project_path, "README.md")) as f:
@@ -97,6 +104,7 @@ class SetupInitializer:
                 data += line
 
         setup = self._execute(data)["SETUP"]
+        self._append_missing_requirements(setup)
         for key in self._setup_data.keys():
             try:
                 setup.pop(key)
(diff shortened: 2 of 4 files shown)


tools/client_package_initializer.py
```diff
@@ -105,8 +113,23 @@ def _load_setup(self) -> None:
 
         self._setup_data.update(setup)
 
+    @classmethod
+    def _append_missing_requirements(cls, setup: Dict[str, List]):
+        # for assisted_service_api.py
+        if "kubernetes" not in "\n".join(setup[Setup.INSTALL_REQUIRES_KEY]):
+            setup[Setup.INSTALL_REQUIRES_KEY].append("kubernetes~=18.20.0")
```
osherdp, line 120: please change to a broader range when listing dependencies in ``install_requires``

tools/client_package_initializer.py
```diff
@@ -105,8 +113,23 @@ def _load_setup(self) -> None:
 
         self._setup_data.update(setup)
 
+    @classmethod
+    def _append_missing_requirements(cls, setup: Dict[str, List]):
+        # for assisted_service_api.py
```
osherdp, line 118: replace with docstring (and some more explanation)

tools/client_package_initializer.py
```diff
@@ -74,6 +77,10 @@ def __init__(self, project_path: str, url: str) -> None:
         with open(os.path.join(self._project_path, "MANIFEST.in"), "w") as f:
             f.write(f"include {self.DEFAULT_README_PATH}")
 
+    def _load_client_api(self) -> None:
```
osherdp, line 80: the function name seems not indicative to what it does

tools/client_package_initializer.py
```diff
@@ -148,11 +171,12 @@ def build(self) -> None:
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description="Swagger package builder")
+    parser.add_argument("client_extra_files_dir", help="External modules to add to the python api client", type=str)
```
osherdp, line 174: any ``*extra*`` argument should come after the main one :) 
also, I recommend changing to named argument as its role is not clear when looking usage in ``hack/generate.sh``

tools/client_package_initializer.py
```diff
@@ -105,8 +113,23 @@ def _load_setup(self) -> None:
...
+        if "kubernetes" not in "\n".join(setup[Setup.INSTALL_REQUIRES_KEY]):
+            setup[Setup.INSTALL_REQUIRES_KEY].append("kubernetes~=18.20.0")
+        if "waiting" not in "\n".join(setup[Setup.INSTALL_REQUIRES_KEY]):
+            setup[Setup.INSTALL_REQUIRES_KEY].append("waiting~=1.4.1")
+        if "requests" not in "\n".join(setup[Setup.INSTALL_REQUIRES_KEY]):
+            setup[Setup.INSTALL_REQUIRES_KEY].append("requests~=2.26.0")
+        if "urllib3" not in "\n".join(setup[Setup.INSTALL_REQUIRES_KEY]):
+            setup[Setup.INSTALL_REQUIRES_KEY].append("urllib3~=1.26.7")
+        if "retry" not in "\n".join(setup[Setup.INSTALL_REQUIRES_KEY]):
+            setup[Setup.INSTALL_REQUIRES_KEY].append("retry~=0.9.2")
```
osherdp, line 128: there must be a shorter way to write this code

hack/generate.sh
```diff
@@ -36,7 +36,9 @@ function generate_python_client() {
         -e SWAGGER_FILE=/swagger.yaml -e OUTPUT=/local/assisted-service-client/ \
         quay.io/ocpmetal/swagger-codegen-cli:2.4.15 /script.sh
      cd "${dest}"/assisted-service-client/ && python3 setup.py sdist --dist-dir "${dest}"
-     cd "${dest}"/assisted-service-client/ && python3 "${__root}"/tools/client_package_initializer.py "${dest}"/assisted-service-client/  https://github.com/openshift/assisted-service --build
+     cd "${dest}"/assisted-service-client/ && python3 "${__root}"/tools/client_package_initializer.py "${__root}"/tools/python_api_client_extra_files/ "${dest}"/assisted-service-client/  https://github.com/openshift/assisted-service --build
+     rm "${dest}"/assisted-service-client-1.0.0.tar.gz
+     cp "${dest}"/assisted-service-client/dist/assisted-service-client-*.tar.gz "${dest}"/
```
osherdp, line 41:
```suggestion
     cp "${dest}"/assisted-service-client/dist/assisted-service-client-*.tar.gz "${dest}"
```

osherdp, line 41: can you explain what happens here?

osherdp (changes requested):

tools/client_package_initializer.py
eliorerz: Create another requirements file and update the current one

hack/generate.sh
eliorerz: 1. Updating the client and generating wheel and tar.gz artifacts
2. Remove the old artifact that was created so far
3. Copy the new source artifact to where the previous one was

tools/client_package_initializer.py
eliorerz: It's not extra argument. I can change the order but it's mandatory argument.

osherdp: yeah, reorder
base always comes before extra :)

nshidlin:
trying to understand what the followup PR in test-infra would look. where would the client be imported from?

eliorerz:
> trying to understand what the followup PR in test-infra would look. where would the client be imported from?

```python
from assisted_service_client import assisted_service_api
```

osherdp:
I still don't like all this ``package>=version`` in ``setup.py``, but I'll roll with that
/lgtm

osherdp:
@eliorerz maybe you'll need to rebase

osherdp (approved):

osherdp:
/hold
(just holding a bit while we still evaluating it)

tools/python_api_client_extra_files/assisted_service_api.py
```diff
@@ -0,0 +1,533 @@
+# -*- coding: utf-8 -*-
+import base64
```
filanov, line 2: i don't think that this file belongs to assisted-service repo

osherdp:
I agree, looking at scoping right now it does seem weird to have it here
sorry :) 
/lgtm cancel

osherdp (changes requested):
maybe we can still use some stuff from this PR, but I agree the extra module should still exist in test-infra

eliorerz:
After discussing it, we decided to leave the wrapper in assisted-test-infra repository but as separate package.
Opened new ticket - https://issues.redhat.com/browse/MGMT-8367

PR #2912, opened 2021-11-11
MGMT-8079: reboot BMH on unbind-pending-user-action

slaviered:
# Assisted Pull Request

## Description
- Ask BMH to reboot the agent from ISO on late unbind scenario when the agent mark that it is ready by setting UnbindCondition to UnbindingPendingUserActionReason

- This is done by re-attaching the BMH and removing the image URL so the provisioning cycle will restart

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -256,6 +256,19 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
 		return result.Result()
 	}
 
+	result = r.reconcileUnboundAgent(log, bmh, agent)
+	if result.Dirty() {
+		err := r.Client.Update(ctx, bmh)
+		if err != nil {
+			log.WithError(err).Errorf("Error adding reset annotation on BMH for unbound agent")
+			return reconcileError{err}.Result()
+		}
+	}
+
+	if result.Stop(ctx) {
+		return result.Result()
+	}
+
 	result = r.ensureMCSCert(ctx, log, bmh, agent)
 	if result.Dirty() {
 		err := r.Client.Update(ctx, bmh)
@@ -502,6 +515,33 @@ func (r *BMACReconciler) reconcileAgentInventory(log logrus.FieldLogger, bmh *bm
 	log.Debugf("Agent Inventory reconciled to BMH \n %v \n %v", agent, bmh)
 	return reconcileComplete{dirty: true}
 
+}
+
+// Ask BMH to reboot the host if the agent is unbound after installation
+//
+// By re-attaching the BMH and clearing the Image field on it, BMAC will clear
+// the Image data to force the boot from ISO
+//
+func (r *BMACReconciler) reconcileUnboundAgent(log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent) reconcileResult {
+	log.Debugf("Started Unbound Agent reconcile for agent %s/%s and bmh %s/%s", agent.Namespace, agent.Name, bmh.Namespace, bmh.Name)
+
+	// proceed with the reconcile only when the agent ask for user action following unbinding
+	boundCondition := conditionsv1.FindStatusCondition(agent.Status.Conditions, aiv1beta1.BoundCondition)
+	if boundCondition == nil || boundCondition.Reason != aiv1beta1.UnbindingPendingUserActionReason {
+		log.Debugf("Skipping Unbound Agent reconcile \n %v", agent)
+		return reconcileComplete{}
+	}
+
+	// re-attach the bmh and clear the ISO url to force ironic image refresh
+	// and re-conciliation of BMH and agent. Also, clear the hw details just
+	// in case. They will be regenerated in the reconciles following the agent's
+	// reboot
+	delete(bmh.ObjectMeta.Annotations, BMH_DETACHED_ANNOTATION)
+	delete(bmh.ObjectMeta.Annotations, BMH_HARDWARE_DETAILS_ANNOTATION)
+	bmh.Spec.Image = nil
+
+	log.Infof("Unbound Agent reconciled to BMH \n %v \n %v", agent, bmh)
+	return reconcileComplete{dirty: true, stop: true}
 }
 
 // Utility to verify whether a BMH should be reconciled based on the InfraEnv
--- a/internal/controller/controllers/bmh_agent_controller_test.go
+++ b/internal/controller/controllers/bmh_agent_controller_test.go
@@ -1051,6 +1051,33 @@ var _ = Describe("bmac reconcile", func() {
 				Expect(updatedHost.Spec.Image.URL).ToNot(Equal(infraEnv.Status.ISODownloadURL))
 			})
 		})
+
+		Context("when Agent is unbound-pending-on-user-action", func() {
+			It("should remove the detached annotation and clear the ISO", func() {
+				agent.Status.Conditions = []conditionsv1.Condition{
+					{
+						Type:   v1beta1.BoundCondition,
+						Status: corev1.ConditionTrue,
+						Reason: v1beta1.UnbindingPendingUserActionReason,
+					},
+				}
+				Expect(c.Update(ctx, agent)).To(BeNil())
+				hostBeforeReconcile := &bmh_v1alpha1.BareMetalHost{}
+				_ = c.Get(ctx, types.NamespacedName{Name: host.Name, Namespace: testNamespace}, hostBeforeReconcile)
+				Expect(hostBeforeReconcile.Spec.Image).NotTo(BeNil())
+
+				result, err := bmhr.Reconcile(ctx, newBMHRequest(host))
+				Expect(err).To(BeNil())
+				Expect(result).To(Equal(ctrl.Result{}))
+
+				updatedHost := &bmh_v1alpha1.BareMetalHost{}
+				err = c.Get(ctx, types.NamespacedName{Name: host.Name, Namespace: testNamespace}, updatedHost)
+				Expect(err).To(BeNil())
+				Expect(updatedHost.ObjectMeta.Annotations).NotTo(HaveKey(BMH_DETACHED_ANNOTATION))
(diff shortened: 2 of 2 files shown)


internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -496,6 +510,42 @@ func (r *BMACReconciler) reconcileAgentInventory(log logrus.FieldLogger, bmh *bm
...
+
+	// bmhAnnotations := bmh.ObjectMeta.GetAnnotations()
+	// if _, ok := bmhAnnotations[BMH_REBOOT_ANNOTATION]; ok {
+	// 	return reconcileComplete{}
+	// }
+	// //use the default reset mode (soft reset)
+	// bmh.ObjectMeta.Annotations[BMH_REBOOT_ANNOTATION] = ""
+
+	log.Debugf("Unbound Agent reconciled to BMH \n %v \n %v", agent, bmh)
+	return reconcileComplete{dirty: true}
```
flaper87, line 546:
```suggestion
	return reconcileComplete{dirty: true, stop: true}
```

It'd be better to stop the reconcile loop as we know the following steps won't be useful for this specific BMH. We need the loop to start again so that the ISO can be copied into the BMH.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -496,6 +510,42 @@ func (r *BMACReconciler) reconcileAgentInventory(log logrus.FieldLogger, bmh *bm
...
+	}
+
+	if bmh.ObjectMeta.Annotations == nil {
+		bmh.ObjectMeta.Annotations = make(map[string]string)
+	}
+
+	// re-attach the bmh and clear the ISO url to force ironic image refresh
+	// and re-conciliation of BMH and agent
+	delete(bmh.ObjectMeta.Annotations, BMH_DETACHED_ANNOTATION)
+	bmh.Spec.Image = nil
```
flaper87, line 536: Not strictly required but it may be good to also clean the `HardwareDetails` from the BMH status so that it will be reconciled again once the new agent boots.

flaper87 (changes requested):
BMAC will get the (new) agent based on the following criteria: 

https://github.com/openshift/assisted-service/blob/6a0dcf6dae2e2ffc0a936fd6b8becd674a0fac71/internal/controller/controllers/bmh_agent_controller.go#L880-L895

It should all work as expected.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -496,6 +510,42 @@ func (r *BMACReconciler) reconcileAgentInventory(log logrus.FieldLogger, bmh *bm
...
+		return reconcileComplete{}
+	}
+
+	if bmh.ObjectMeta.Annotations == nil {
+		bmh.ObjectMeta.Annotations = make(map[string]string)
+	}
+
+	// re-attach the bmh and clear the ISO url to force ironic image refresh
+	// and re-conciliation of BMH and agent
+	delete(bmh.ObjectMeta.Annotations, BMH_DETACHED_ANNOTATION)
```
rollandf, line 535: Maybe only delete if annotations is not nil?
if bmh.ObjectMeta.Annotations != nil {
		delete(bmh.ObjectMeta.Annotations, BMH_DETACHED_ANNOTATION)
}

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -71,6 +71,7 @@ const (
 	BMH_API_VERSION                     = "baremetal.cluster.k8s.io/v1alpha1"
 	BMH_DETACHED_ANNOTATION             = "baremetalhost.metal3.io/detached"
 	BMH_INSPECT_ANNOTATION              = "inspect.metal3.io"
+	BMH_REBOOT_ANNOTATION               = "reboot.metal3.io"
```
rollandf, line 74: This is not used eventually. Also remove the code comments referring to it.

internal/controller/controllers/bmh_agent_controller.go
slaviered: why? delete will not fail if it is not there and the action is idempotent

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -496,6 +510,42 @@ func (r *BMACReconciler) reconcileAgentInventory(log logrus.FieldLogger, bmh *bm
...
+//
+func (r *BMACReconciler) reconcileUnboundAgent(log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent) reconcileResult {
+	log.Debugf("Started Unbound Agent reconcile for agent %s/%s and bmh %s/%s", agent.Namespace, agent.Name, bmh.Namespace, bmh.Name)
+
+	// proceed with the reconcile only when the agent ask for user action following unbinding
+	boundCondition := conditionsv1.FindStatusCondition(agent.Status.Conditions, aiv1beta1.BoundCondition)
+	if boundCondition == nil || boundCondition.Reason != aiv1beta1.UnbindingPendingUserActionReason {
+		return reconcileComplete{}
+	}
+
```
rollandf, line 528: Maybe add an info log that we are in this flow?

internal/controller/controllers/bmh_agent_controller.go
slaviered: After inspecting the code more closely: we do not need the if at all. delete will work both on nil and existing maps. Removing the redundant allocation

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -496,6 +509,32 @@ func (r *BMACReconciler) reconcileAgentInventory(log logrus.FieldLogger, bmh *bm
...
+	// re-attach the bmh and clear the ISO url to force ironic image refresh
+	// and re-conciliation of BMH and agent. Also, clear the hw details just
+	// in case. They will be regenerated in the reconciles following the agent's
+	// reboot
+	delete(bmh.ObjectMeta.Annotations, BMH_DETACHED_ANNOTATION)
+	delete(bmh.ObjectMeta.Annotations, BMH_HARDWARE_DETAILS_ANNOTATION)
+	bmh.Spec.Image = nil
+
+	log.Debugf("Unbound Agent reconciled to BMH \n %v \n %v", agent, bmh)
```
rollandf, line 534: I think that we can have this in Info. It is important enough and not going to happen a lot of times.

internal/controller/controllers/bmh_agent_controller_test.go
```diff
@@ -1051,6 +1051,29 @@ var _ = Describe("bmac reconcile", func() {
...
+				}
+				Expect(c.Update(ctx, agent)).To(BeNil())
+				result, err := bmhr.Reconcile(ctx, newBMHRequest(host))
+				Expect(err).To(BeNil())
+				Expect(result).To(Equal(ctrl.Result{}))
+
+				updatedHost := &bmh_v1alpha1.BareMetalHost{}
+				err = c.Get(ctx, types.NamespacedName{Name: host.Name, Namespace: testNamespace}, updatedHost)
+				Expect(err).To(BeNil())
+				Expect(updatedHost.ObjectMeta.Annotations).NotTo(HaveKey(BMH_DETACHED_ANNOTATION))
```
rollandf, line 1072: Maybe make sure we have the annotations and spec.Image in the BMH at the beginning of the test?

rollandf:
Looks good.
Let's have @flaper87 a chance to have another look before approving

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -496,6 +509,32 @@ func (r *BMACReconciler) reconcileAgentInventory(log logrus.FieldLogger, bmh *bm
...
+// By re-attaching the BMH and clearing the Image field on it, BMAC will clear
+// the Image data to force the boot from ISO
+//
+func (r *BMACReconciler) reconcileUnboundAgent(log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent) reconcileResult {
+	log.Debugf("Started Unbound Agent reconcile for agent %s/%s and bmh %s/%s", agent.Namespace, agent.Name, bmh.Namespace, bmh.Name)
+
+	// proceed with the reconcile only when the agent ask for user action following unbinding
+	boundCondition := conditionsv1.FindStatusCondition(agent.Status.Conditions, aiv1beta1.BoundCondition)
+	if boundCondition == nil || boundCondition.Reason != aiv1beta1.UnbindingPendingUserActionReason {
+		return reconcileComplete{}
```
flaper87, line 523: It'd be useful to have a debug line here mentioning that we are skipping the reconcile for unbound agents

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -1147,6 +1186,11 @@ func (r *BMACReconciler) newSpokeBMH(log logrus.FieldLogger, bmh *bmh_v1alpha1.B
 		if _, err := r.reconcileAgentInventory(log, bmhSpoke, agent).Result(); err != nil {
 			return err
 		}
+
+		// check if we need to reset the agent in case of unbinding after installation
+		if _, err := r.reconcileUnboundAgent(log, bmhSpoke, agent).Result(); err != nil {
+			return err
+		}
```
flaper87, line 1193: Could you elaborate on why this is needed here? Asking because this function is called *only* for day2 hosts and it modifies the BMH CR in the *spoke* cluster. This is done by copying the hub's BMH data and creating a new CR so we can expect the spoke BMH to be the same as the hub's.

internal/controller/controllers/bmh_agent_controller.go
slaviered: It was not called by design because in most case the call will be skipped. In Debug mode the log will be littered with these non useful logs. The case where we do find unbound case is clearly documented

flaper87 (approved):
:pizza:

PR #2916, opened 2021-11-11
MGMT-7330: Create a job in prow CI to run assisted-service unit tests

pawanpinjarkar:
# Assisted Pull Request

## Description
Instead of the current `unit-test` target which does multiple things in one target -  create a docker DB, run unit tests and clean/kill docker db container, create separate targets for each task. 
- `run-db-container` only creates a container database
- `run-unit-test` runs the unit tests
- `kill-db-container` kills the DB container

The current `unit-test` target functionality remains as-is for local development - `unit-test` calls `run-db-container`, ` run-unit-test` and `kill-db-container`.

For running unit tests in CI, the assisted-service-build dockerfile is updated to use `quay.io/edge-infrastructure/postgresql-12-centos7:latest` as the base image and installs required dependencies. When running unit tests in CI job, invoke `make ci-unit-test` which will first start the database, and then invoke `run-unit-test` which will run the unit tests.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
The current `unit-test` target functionality remains as-is for local development - unit-test calls run-db-container, run-unit-test, and kill-db-container. When running unit tests in CI job, invoke make ci-unit-test which will first start the database, and then invoke just-unit-test which will run the unit tests. Run `skipper make ci-unit-test` to test running the unit tests in CI.
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/Dockerfile.assisted-service-build
+++ b/Dockerfile.assisted-service-build
@@ -1,10 +1,21 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.16
+FROM quay.io/edge-infrastructure/postgresql-12-centos7:latest
 
-ENV GO111MODULE=on
-ENV GOFLAGS=""
+ENV GOPATH=/go
+ENV GOROOT=/usr/local/go
+ENV PATH=$PATH:$GOROOT/bin:$GOPATH/bin
+
+USER 0
+
+RUN yum install -y epel-release && yum install -y gcc genisoimage jq
+
+COPY --from=registry.ci.openshift.org/openshift/release:golang-1.16 /usr/bin/git /usr/bin/gotestsum /usr/bin/make /usr/bin/
+COPY --from=registry.ci.openshift.org/openshift/release:golang-1.16 /usr/local/go /usr/local/go
+
 
 COPY ./hack/setup_env.sh ./dev-requirements.txt ./
 RUN ./setup_env.sh assisted_service
+
+RUN chmod g+xw -R $GOPATH && chmod g+xw -R $(go env GOROOT)
 
 COPY --from=quay.io/openshift/origin-cli:latest /usr/bin/oc /usr/bin
 COPY --from=quay.io/operator-framework/upstream-opm-builder:v1.16.1 /bin/opm /bin
--- a/Makefile
+++ b/Makefile
@@ -446,13 +446,23 @@ ifeq ($(CI), true)
 	gocov convert $(REPORTS)/$(TEST_SCENARIO)_coverage.out | gocov-xml > $(REPORTS)/$(TEST_SCENARIO)_coverage.xml
 endif
 
-unit-test:
+run-db-container:
 	docker ps -q --filter "name=postgres" | xargs -r docker kill && sleep 3
 	docker run -d  --rm --tmpfs /var/lib/pgsql/data --name postgres -e POSTGRESQL_ADMIN_PASSWORD=admin -e POSTGRESQL_MAX_CONNECTIONS=10000 -p 127.0.0.1:5432:5432 \
 	quay.io/edge-infrastructure/postgresql-12-centos7:latest
 	timeout 5m ./hack/wait_for_postgres.sh
-	SKIP_UT_DB=1 $(MAKE) _test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./... | grep -v subsystem))" || (docker kill postgres && /bin/false)
+
+run-unit-test:
+	SKIP_UT_DB=1 $(MAKE) _test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./... | grep -v subsystem))"
+
+ci-unit-test:
+	./hack/start_db.sh
+	$(MAKE) run-unit-test
+
+kill-db-container:
 	docker kill postgres
+ 
+unit-test: run-db-container run-unit-test kill-db-container
 
 $(REPORTS):
 	-mkdir -p $(REPORTS)
--- a/hack/setup_env.sh
+++ b/hack/setup_env.sh
@@ -44,7 +44,7 @@ function assisted_service() {
   curl --retry 5 -L "https://dl.k8s.io/release/${latest_kubectl_version}/bin/linux/amd64/kubectl" -o /tmp/kubectl && \
     install -o root -g root -m 0755 /tmp/kubectl /usr/local/bin/kubectl && \
     rm -f /tmp/kubectl
-  yum install -y docker podman libvirt-clients awscli python3-pip postgresql genisoimage skopeo p7zip && \
+  yum install -y docker podman libvirt-clients awscli python3-pip genisoimage skopeo p7zip && \
     yum clean all
 
   kustomize
--- a/hack/start_db.sh
+++ b/hack/start_db.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -euo pipefail
+set -o errexit
+
+mkdir -p /tmp/postgres/data
+mkdir -p /tmp/postgres/sockets
+
+/usr/bin/initdb -D /tmp/postgres/data -U postgres
+/opt/rh/rh-postgresql12/root/usr/bin/pg_ctl -D /tmp/postgres/data -l /tmp/postgres/logfile -o'-k /tmp/postgres/sockets' start
--- a/internal/common/common_unitest_db.go
+++ b/internal/common/common_unitest_db.go
@@ -47,7 +47,7 @@ func InitializeDBTest() {
 		oldResource.Close()
 	}
 	resource, err := pool.RunWithOptions(&dockertest.RunOptions{
-		Repository: "postgresql-12-centos7",
+		Repository: "quay.io/edge-infrastructure/postgresql-12-centos7",
 		Tag:        "latest",
 		Env:        []string{"POSTGRESQL_ADMIN_PASSWORD=admin"},
 		Name:       dbDockerName,


Dockerfile.assisted-service-build
```diff
@@ -1,7 +1,12 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.16
+FROM quay.io/edge-infrastructure/postgresql-12-centos7:latest
 
-ENV GO111MODULE=on
-ENV GOFLAGS=""
```
carbonin, line 4: Do we still need to set these?

Makefile
```diff
@@ -445,13 +445,23 @@ ifeq ($(CI), true)
...
+	SKIP_UT_DB=1 $(MAKE) _test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./... | grep -v subsystem))"
+
+ci-unit-test:
+	./hack/setup_db.sh
+	$(MAKE) just-unit-test
+
+kill-db-container:
 	docker kill postgres
+ 
+unit-test: docker-db just-unit-test kill-db-container
```
carbonin, line 464: Have you locally tested both `make unit-test` and `skipper make unit-test`?
I think people probably do both.

Makefile
```diff
@@ -445,13 +445,23 @@ ifeq ($(CI), true)
...
 	docker ps -q --filter "name=postgres" | xargs -r docker kill && sleep 3
 	docker run -d  --rm --tmpfs /var/lib/postgresql/data --name postgres -e POSTGRESQL_ADMIN_PASSWORD=admin -e POSTGRESQL_MAX_CONNECTIONS=10000 -p 127.0.0.1:5432:5432 \
 	quay.io/edge-infrastructure/postgresql-12-centos7:latest
 	timeout 5m ./hack/wait_for_postgres.sh
-	SKIP_UT_DB=1 $(MAKE) _test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./... | grep -v subsystem))" || (docker kill postgres && /bin/false)
+
+just-unit-test:
+	SKIP_UT_DB=1 $(MAKE) _test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./... | grep -v subsystem))"
+
+ci-unit-test:
```
carbonin, line 457: What happens if you run `skipper make ci-unit-test` locally?
Does that work?

hack/setup_db.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -o nounset
+set -euo pipefail
+set -o errexit
+
+mkdir -p /tmp/postgres/data
+mkdir -p /tmp/postgres/sockets
+
+/usr/bin/initdb -D /tmp/postgres/data -U postgres
```
carbonin, line 10: Wont this fail when we're in an environment where the container is running as root?

carbonin:
Also, please don't ignore the PR template.

Dockerfile.assisted-service-build
pawanpinjarkar: no, those are leftovers. I will remove it.

Makefile
pawanpinjarkar: yes, I have tested running it locally and it works.

Makefile
pawanpinjarkar: I have always run existing unit tests with skipper. I tested running `make unit-test` locally from the master branch and it fails with a timeout. So, I think running locally needs `skipper` for exisiting `unit-test` target. 

```
~/go/src/github.com/openshift/assisted-service (master) $ make unit-test
docker ps -q --filter "name=postgres" | xargs -r docker kill && sleep 3
docker run -d  --rm --tmpfs /var/lib/postgresql/data --name postgres -e POSTGRESQL_ADMIN_PASSWORD=admin -e POSTGRESQL_MAX_CONNECTIONS=10000 -p 127.0.0.1:5432:5432 \
quay.io/edge-infrastructure/postgresql-12-centos7:latest
58819e63b8465af7ceb6ae4c4274c52f2bc2ef9a5f142c59f369409bb180f602
timeout 5m ./hack/wait_for_postgres.sh
make: *** [Makefile:451: unit-test] Error 124
```

Dockerfile.assisted-service-build
pawanpinjarkar: What is the other scenario that might need it? Maybe we can add later if it breaks something but as of now I don't know any other scenario than unit tests and it doesn't need it.

carbonin: :man_shrugging: It looks like there was an existing `GOFLAG` value in the previous base image that we were probably trying to remove.

As for `GO111MODULE` it looks like it's on by default as of go 1.16 ref: https://go.dev/blog/go116-module-changes

Makefile
carbonin: @omertuc wasn't this what we were trying to handle in the last PR?

You were running unit tests without skipper right?

Makefile
carbonin: This is eventually the direction I wanted to go for both ci and local, but the concern was that people could previously run unit tests without skipper and that lead to things being changed on their local system.

Relates to https://github.com/openshift/assisted-service/pull/2916#discussion_r750471511

@omertuc does this look like what you were thinking in the last PR?

hack/setup_db.sh
pawanpinjarkar: Yes, but what will be that use case? In CI, it will be run as a nonroot user, no?

Makefile
omertuc: I don't personally run without skipper, but I didn't want to lock our unit-tests flow to only RPM-based distros (our skipper build container technically being an "RPM-based" distro). 

It was nice when the unit tests flow launched postgres in a container, i.e. assuming the user has podman/docker is better  than assuming the user has a yum package manager

I wasn't aware `make unit-test` already doesn't work. Looking into it it simply has to do with the fact that `hack/wait_for_postgres.sh` silently fails when the host doesn't have a `psql` binary. It should not assume such thing, it should instead run the `psql` binary from inside the container to do this check. It works in skipper because we already [install](https://github.com/openshift/assisted-service/blob/56e04f4305f0514451a4b3ca9b47e70636344dbe/hack/setup_env.sh#L47) an entire `postgresql` package presumably just for this `psql` binary used in wait_for_postgres

Makefile
omertuc: Yeah this PR pretty much looks like what I had in mind, apart from the change of the `-build` Dockerfile base image - I expected the base image to stay the same with the `-build` Dockerfile installing postgres

hack/setup_db.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
omertuc, line 1: `setup` sounds a bit like installation, rename file to `start_db.sh`

hack/setup_db.sh
omertuc: > Wont this fail when we're in an environment where the container is running as root?

What container? Why would this fail when it's running as root?

pawanpinjarkar: this file will be used only from ci-unit-test. In earlier PR, the similar file was getting used from `unit-test` target and then the jenkins PR check was failing but now that we haven't changed any internal functionality for the `unit-test` target, it should not be an issue.

hack/setup_db.sh
pawanpinjarkar: sure, will do.

hack/setup_db.sh
carbonin: > Why would this fail when it's running as root?

IIRC postgres itself has Opinions about what user owns a data directory.

Makefile
carbonin: Yeah, turned out installing go in a PG image was easier than installing PG in a go image :man_shrugging:

Makefile
```diff
@@ -446,13 +446,23 @@ ifeq ($(CI), true)
 	gocov convert $(REPORTS)/$(TEST_SCENARIO)_coverage.out | gocov-xml > $(REPORTS)/$(TEST_SCENARIO)_coverage.xml
 endif
 
-unit-test:
+docker-db:
```
flaper87, line 449: Can we call this `run-db-container`?

This will match the `kill-db-container` name.

Makefile
```diff
@@ -446,13 +446,23 @@ ifeq ($(CI), true)
...
 	docker ps -q --filter "name=postgres" | xargs -r docker kill && sleep 3
 	docker run -d  --rm --tmpfs /var/lib/postgresql/data --name postgres -e POSTGRESQL_ADMIN_PASSWORD=admin -e POSTGRESQL_MAX_CONNECTIONS=10000 -p 127.0.0.1:5432:5432 \
 	quay.io/edge-infrastructure/postgresql-12-centos7:latest
 	timeout 5m ./hack/wait_for_postgres.sh
-	SKIP_UT_DB=1 $(MAKE) _test TEST_SCENARIO=unit TIMEOUT=30m TEST="$(or $(TEST),$(shell go list ./... | grep -v subsystem))" || (docker kill postgres && /bin/false)
+
+just-unit-test:
```
flaper87, line 455:
```suggestion
run-unit-test:
```

flaper87 (changes requested):
overall looks good. Just a few name suggestions and then I think we can move this forward

flaper87 (approved):
:pizza:

lranjbar (approved):
LGTM. Seems to run locally using skipper

PR #2970, opened 2021-11-22
MGMT-8279: kube-api - ensure release images exists upon install

danielerez:
# Assisted Pull Request

## Description

To avoid a missing release image on host installation, ClusterDeploymentsReconciler now ensures that the relevant image exists in version cache before invoking installation.

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @rollandf 
/cc @filanov 
/cc @eranco74 
/cc @nmagnezi 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/clusterdeployments_controller.go
+++ b/internal/controller/controllers/clusterdeployments_controller.go
@@ -229,19 +229,24 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 
 	// Create Kubeconfig no-ingress if needed
 	if *cluster.Status == models.ClusterStatusInstalling || *cluster.Status == models.ClusterStatusFinalizing {
-		if err := r.createNoIngressKubeConfig(ctx, log, clusterDeployment, cluster, clusterInstall); err != nil {
-			log.WithError(err).Error("failed to create kubeconfig no-ingress secret")
-			return r.updateStatus(ctx, log, clusterInstall, cluster, err)
+		if err1 := r.createNoIngressKubeConfig(ctx, log, clusterDeployment, cluster, clusterInstall); err1 != nil {
+			log.WithError(err1).Error("failed to create kubeconfig no-ingress secret")
+			return r.updateStatus(ctx, log, clusterInstall, cluster, err1)
 		}
+	}
+
+	pullSecret, err := getPullSecretData(ctx, r.Client, r.APIReader, clusterDeployment.Spec.PullSecretRef, req.Namespace)
+	if err != nil {
+		log.WithError(err).Error("failed to get pull secret")
+		return r.updateStatus(ctx, log, clusterInstall, nil, err)
 	}
 
 	if swag.StringValue(cluster.Kind) == models.ClusterKindCluster && !clusterInstall.Spec.HoldInstallation {
 		// Day 1
-		return r.installDay1(ctx, log, clusterDeployment, clusterInstall, cluster)
-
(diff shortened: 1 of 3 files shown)


internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -365,6 +370,13 @@ func (r *ClusterDeploymentsReconciler) installDay1(ctx context.Context, log logr
 			return ctrl.Result{Requeue: true, RequeueAfter: longerRequeueAfterOnError}, nil
 		}
 
+		// Ensure release image exists in versions cache
+		_, err = r.addReleaseImage(ctx, clusterInstall.Spec, pullSecret)
+		if err != nil {
+			log.WithError(err).Error("failed to add OCP version")
```
nmagnezi, line 376: this line is repeated both here and in `installDay2Hosts()`, maybe it should be included in `addReleaseImage()`?
Additionally, this is not informative enough to the reader. please also print for which image name this is failing.

nmagnezi (changes requested):
Please also add unit-test(s) that reconciles with some invalid image name and expect an error. We want to assert for the error and to make sure it does not Requeue for that reason.

internal/controller/controllers/clusterdeployments_controller.go
danielerez: sure, added.

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -229,19 +229,24 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 
 	// Create Kubeconfig no-ingress if needed
 	if *cluster.Status == models.ClusterStatusInstalling || *cluster.Status == models.ClusterStatusFinalizing {
-		if err := r.createNoIngressKubeConfig(ctx, log, clusterDeployment, cluster, clusterInstall); err != nil {
+		if err1 := r.createNoIngressKubeConfig(ctx, log, clusterDeployment, cluster, clusterInstall); err != nil {
```
filanov, line 232: why do you need to create err1? 
regardless there is a bug here 
```suggestion
		if err1 := r.createNoIngressKubeConfig(ctx, log, clusterDeployment, cluster, clusterInstall); err1 != nil {
```

filanov: it means that your unit test don't cover bad flows

filanov:
Why call it only on installation? 
What is the downside to call it all the time?

danielerez:
> Please also add unit-test(s) that reconciles with some invalid image name and expect an error. We want to assert for the error and to make sure it does not Requeue for that reason.

Added.

danielerez:
> Why call it only on installation? What is the downside to call it all the time?

So currently we actually don’t have caching based on release image url: https://github.com/openshift/assisted-service/blob/1b0a65263d27d00903f71ec8de65448380f739cc/internal/versions/versions.go

To avoid invoking ‘oc adm release info’ redundantly, added it only once before installation. We can add caching for that, but not sure we want to rely on the specified url (i.e. is it safe to assume users won't use 'latest' tag or such?)

filanov (approved):

carbonin:
Missing this in 2.4 is causing https://issues.redhat.com/browse/MGMT-11956

PR #2990, opened 2021-11-26
MGMT-8147: Expose OlmOperators through KubeApi

flaper87:
# Assisted Pull Request

## Description

Assisted service already supports enabling some olm operators that would be deployed as part of the day1 install
process. This is currently only exposed through the HTTP API. We need to also expose them through the KubeAPI for
consistency and because we want to be able to install operators like ODF and OCS as part of the day1 step on specific
scenarios (ZTP factory pipeline).

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] unit-tests
- [ ] No tests needed

## Assignees

/cc @avishayt 
/cc @rollandf 
/cc @jelkosz 

## Checklist

## Reviewers Checklist


flaper87:
@avishayt @rollandf @jelkosz I created this as a way to move the conversation forward. I don't think this is a blocker, but it would be useful, especially considering that we already support these operators from the HTTP API. The evaluation of what operators assisted will support can be done on a case-by-case basis.

In this implementation, I focused on keeping the API as generic and as similar to the HTTP API as possible. I believe there are no changes in how the user would interact with it.

Let me know what you think :)

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -331,6 +335,16 @@ type DiskEncryption struct {
...
 }
 
+type OlmOperator struct {
+	// Enable/disable one or more of the assisted-installer supported operators.
+
+	// The name of the operator
+	Name string `json:"name,omitempty"`
+
+	// Blob of operator-dependent parameters that are required for installation.
+	Properties string `json:"properties,omitempty" gorm:"type:text"`
```
avishayt, line 345: I hate that we do this in the REST API, and it's mostly because of the tooling we use that ties the API to the DB schema.  In this case, we didn't want to make new tables and schemas for these properties, but I'd like to have it well-defined in the k8s API at least.

flaper87: > I hate that we do this in the REST API, and it's mostly because of the tooling we use that ties the API to the DB schema. In this case, we didn't want to make new tables and schemas for these properties, but I'd like to have it well-defined in the k8s API at least.

you find me in full agreement on both fronts, the hate and the wish to have it well-defined.

PR #3056, opened 2021-12-15
MGMT-8564: Rename the UI deployment template

eliorerz:
# Assisted Pull Request

## Description
Fix failing CI due to a template rename on  https://github.com/openshift-assisted/assisted-ui/commit/1f499bfc5d94286039d99da3e7030fe6cb847884

## List all the issues related to this PR

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [x] No tests needed

## Assignees

/cc @osherdp

PR #3066, opened 2021-12-16
MGMT-8564: Rename the UI deployment template

osherdp:
Backport of #3056 
This should fix all jobs that try to run UI image (as location and name of the template have changed).
/cc @eliorerz


eliorerz (approved):
LGTM

PR #3067, opened 2021-12-16
MGMT-8571: Deploy the image service as a stateful set

carbonin:
## Description

Previously, when deploying the image service as a deployment, we were
writing the template images to the pod filesystem. This isn't good for
performance or the health of the node we're running on.

Because of this the image service now will run as a stateful set and
request a PV for the data directory.

https://issues.redhat.com/browse/MGMT-8571

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed\

Deployed the operator locally to a crc cluster using `operator-sdk run bundle`.
Saw the image service created as a stateful set with the PV attached properly.
Created an infra-env and downloaded the image.

## Assignees

/cc @avishayt
/cc @pawanpinjarkar 

## Checklist

## Reviewers Checklist


carbonin:
/hold

committed some stuff I shouldn't have

carbonin:
Rebased out what I shouldn't have committed, but leaving the hold as the upgrade doesn't work.
Two things still need to be sorted out:
1. Adding a new mandatory field to our agent-service-config CRD is causing the CSV to never complete the upgrade
  - If we want to keep the field (imageStorage) mandatory then I think we'll need a new major version.
  - Even with a new major version I'm not sure how the upgrade will look for customers
  - If we make the field optional (and maybe use an EmptyDir) volume if it's not specified then maybe we can get away with a minor version bump
  - Even then I'll need to test the case where someone goes from emptyDir to a PVC to make sure that works well.
2. When the upgrade does work, the operator as it exists now will leave the old image service deployment around; we need to remove this.
  - There isn't really anything in the operator-framework or OLM that will help us with this so assuming we want to keep our upgrade graph unconstrained then we'll need to write out some code to remove the deployment if we see it as a part of the reconcile loop.

carbonin:
Unsure if we want to wait to merge this before sorting out https://github.com/openshift/assisted-service/pull/3067#issuecomment-996973399, so I'll leave the hold for now.

carbonin:
> - If we make the field optional (and maybe use an EmptyDir volume if it's not specified) then maybe we can get away with a minor version bump
> - Even then I'll need to test the case where someone goes from emptyDir to a PVC to make sure that works well.

Tested this case out and it doesn't look like it's going to work unless we want to delete and re-create the entire statefulset when the storage information is updated. Specifically I got this error when I changed the CR to include storage information after an upgrade:

> "StatefulSet.apps \"assisted-image-service\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden"

@mhrivnak any advice here? Is there some PVC spec that would be safe for us to default to? I could base the size on the number of `osImage` entries, but I'm sure there are environments that will require more than just the size requirement.

mhrivnak:
> > "StatefulSet.apps "assisted-image-service" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden"
> 
> @mhrivnak any advice here? Is there some PVC spec that would be safe for us to default to? I could base the size on the number of `osImage` entries, but I'm sure there are environments that will require more than just the size requirement.

The statefulset has to be deleted and recreated to change the storage details. That's something the operator can automate. Then we can just expose settings like size on our CR. Maybe we can extrapolate from our previous recommendations what might be a sensible default or recommended size?

Is it important to migrate data in this case? Or can the new volume start empty?

carbonin:
> The statefulset has to be deleted and recreated to change the storage details. That's something the operator can automate.

Okay, so it makes sense to recreate the stateful set :+1: 

> Then we can just expose settings like size on our CR. Maybe we can extrapolate from our previous recommendations what might be a sensible default or recommended size?

If we're going to rebuild the whole stateful set then I don't think we need to change how we ask users to specify the storage details. We could validate that the size is over some reasonably minimum, but I think that's probably a different conversation.

> Is it important to migrate data in this case? Or can the new volume start empty?

No, the data itself will be recreated.

flaper87 (commented):
@carbonin could you elaborate (maybe in the PR description) on the choice of using `StatefulSet` vs `Deployment`?

carbonin:
Updated with a bit more detail @flaper87. Let me know if something still isn't clear.

carbonin:
Marking this as WIP.
Added a few comments in the code so I can keep track of what I still need to test and handle.

With this I was able to deploy the statefulset and update the agentserviceconfig CR to add storage and have the statefulset recreate successfully.

This should mean that we can keep the storage value in the CR optional which will unstick upgrade. Now I need to test a full upgrade from an older (master) version of the operator to the one in this branch which will move us from a deployment to a statefulset.

carbonin:
This can now handle upgrading from deployment and can move between pvc and emptydir by editing the agentserviceconfig without looping forever.

Still TODO:
- Handle monitoring (started in the last commit here)
- Recreate (or just delete) existing PVCs when size request changes

carbonin:
This is ready for review now.
Sorry for the size, each commit should be reviewable independently if that's helpful.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -903,31 +969,177 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
...
+		if err = mutateFn(); err != nil {
+			return err
+		}
+		if err = r.Client.Create(ctx, statefulSet); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	existing := statefulSet.DeepCopyObject().(*appsv1.StatefulSet)
```
mhrivnak, line 1091: Before doing this, I think we want to check if the `DeletionTimestamp` is set, and if so, return. I assume StatefulSet has a finalizer on it that stays until pods get deleted. So deletion will take some time, and the resource is basically immutable during that time. So there's no point in even attempting an update.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -903,31 +969,177 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
...
+	}
+
+	// need to delete and re-create statefulset because the volumes have changed
+	if err = r.Client.Delete(ctx, existing); err != nil {
+		return err
+	}
+
+	for _, pvc := range pvcs {
+		if err = r.Client.Delete(ctx, pvc); err != nil {
+			return err
```
mhrivnak, line 1125: If there is an error here, but the StatefulSet deletion was successful, will the code ever return here and re-try deleting these PVCs?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -903,31 +969,177 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
...
+	// reset object metadata fields for create
+	statefulSet.ObjectMeta = metav1.ObjectMeta{
+		Name:      imageServiceName,
+		Namespace: r.Namespace,
+	}
+	if err = controllerutil.SetControllerReference(instance, statefulSet, r.Scheme); err != nil {
+		return err
+	}
+
+	if err = r.Client.Create(ctx, statefulSet); err != nil {
```
mhrivnak, line 1138: This is likely to collide with the still-existing StatefulSet that hasn't yet completed deletion.

It's better anyway to just have one code path for creation. I suggest just returning after the deletes. When deletion completes, an event will fire, reconciliation will happen, it will notice the statefulset is missing, and it'll get created.

mhrivnak (changes requested):

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: No, I don't think it will.
I suppose we'd have to reconcile the resulting pvcs against the image storage spec directly.
But even then we'd still have to delete the stateful set to get them properly recreated.

Maybe the condition for recreate vs update should be "the statefulset claim templates or any of the pvcs are different from the image storage spec"? Then we would recreate the statefulset and pvcs if either are out of sync.

Do you think there's something more targeted we could do?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -903,31 +969,177 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
...
+			for i, claim := range statefulSet.Spec.VolumeClaimTemplates {
+				if claim.ObjectMeta.Name == "image-service-data" {
+					found = true
+					statefulSet.Spec.VolumeClaimTemplates[i].Spec.Resources.Requests = getStorageRequests(instance.Spec.ImageStorage)
+				}
+			}
+			if !found {
+				statefulSet.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{
+					{
+						ObjectMeta: metav1.ObjectMeta{
```
mhrivnak, line 1011: Could we add a label here to make it easier to identify these PVCs as belonging to this particular statefulset?

carbonin: It actually looks like the pvcs are inheriting the labels from the statefulset so we can probably just use `app: assisted-image-service`

Probably better than what I have here https://github.com/openshift/assisted-service/pull/3067/files/b206f2e1b8b8e4d1a5ab1c3605e324651a665193#diff-c619bb36181b5e1dd76fb1fcc1f54068b16a8761f77106052fcbf5e2f5f9d356R860 :sweat_smile:

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: We could:
* put a finalizer on the StatefulSet
* at the top of reconciling the StatefulSet, check the deletiontimestamp (as mentioned above), and when it's set, go into cleanup mode.

Cleanup mode:
* check if PVCs still exist
* if so, delete them and return
* if not, remove the finalizer

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: Yeah, and will enable cleanup if/when the pods are gone.

carbonin:
/hold

Still have a bit to do for cleanup here.
At least when the agentserviceconfig is removed we need to also remove the finalizer on the statefulset.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -153,6 +157,9 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
 		if errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+
+			// TODO: cleanup finalizer on statefulset here
```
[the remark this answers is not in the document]
mhrivnak: We should set an owner reference on the StatefulSet. Then if the ASC gets deleted, the statefulset will get deleted first. That should avoid the need for a finalizer on the ASC.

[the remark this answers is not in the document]
carbonin: We're already doing that, but the finalizer we added to the statefulset to clean up the PVCs is stopping it from getting cleaned up so we need a way to remove that when the ASC is removed.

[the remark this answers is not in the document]
mhrivnak: The ASC should still exist in that case, until its owned resources all get deleted. So you can deal with the SS finalizer just where you're fetching and reconciling the SS.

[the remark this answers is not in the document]
carbonin: :+1: thanks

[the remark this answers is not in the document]
carbonin: Tested this out and it doesn't really seem to be the case unfortunately.

```
$ oc delete agentserviceconfigs.agent-install.openshift.io agent 
agentserviceconfig.agent-install.openshift.io "agent" deleted
$ oc get agentserviceconfigs.agent-install.openshift.io 
No resources found
$ oc get statefulsets.apps 
NAME                     READY   AGE
assisted-image-service   1/1     7m25s
```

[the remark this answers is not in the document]
mhrivnak: Ah you're right. I got my wires crossed on the default deletion policy. So yes, we'll need to add our own finalizer to ASC to guarantee its dependent resources get deleted before it does.

carbonin:
The CAPI job has been broken on all PRs for over a week so I think that failure is safe to ignore.

carbonin:
@mhrivnak anything else you'd want to see here?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -170,6 +176,40 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
 		return reconcile.Result{}, nil
 	}
 
+	if instance.DeletionTimestamp.IsZero() {
+		if !funk.ContainsString(instance.GetFinalizers(), agentServiceConfigFinalizerName) {
```
mhrivnak, line 180: Consider also [using controllerutil](https://github.com/kubernetes-sigs/controller-runtime/blob/b8db76e/pkg/controller/controllerutil/controllerutil.go#L375) for this.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -910,31 +987,185 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
...
 		}
 		return nil
 	}
 
-	return deployment, mutateFn, nil
+	return statefulSet, mutateFn, nil
+}
+
+func (r *AgentServiceConfigReconciler) cleanupImageServiceFinalizer(ctx context.Context, statefulSet *appsv1.StatefulSet) error {
+	if !funk.ContainsString(statefulSet.GetFinalizers(), imageServiceStatefulSetFinalizerName) {
```
mhrivnak, line 1076: Also here consider controllerutil.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -170,6 +176,40 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
...
+	} else {
+		// do cleanup and remove finalizer
+		statefulSet := &appsv1.StatefulSet{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      imageServiceName,
+				Namespace: r.Namespace,
+			},
+		}
+		if err := r.Get(ctx, client.ObjectKeyFromObject(statefulSet), statefulSet); err != nil && !errors.IsNotFound(err) {
+			log.WithError(err).Error("failed to get image service stateful set for cleanup")
```
mhrivnak, line 196: If it no longer exists for whatever reason, what's the expectation?

I think once the StatefulSet is gone, we can't reliably determine if PVs were cleaned up or not, and should just remove the finalizer from ACI.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -170,6 +176,40 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
...
+			return ctrl.Result{Requeue: true}, err
+		}
+		if err := r.cleanupImageServiceFinalizer(ctx, statefulSet); err != nil {
+			log.WithError(err).Error("failed to cleanup image service stateful set")
+			return ctrl.Result{Requeue: true}, err
+		}
+
+		controllerutil.RemoveFinalizer(instance, agentServiceConfigFinalizerName)
```
mhrivnak, line 204: It may be wise to wait until the StatefulSet is actually gone. Consider someone who wants to delete and then re-create an ACI, perhaps even with a script. As soon as the ACI is deleted, they make a new one, and some error occurs because the reconciler can't create a new stateful set since the old one is still being deleted. Given the parent/child tree relationship, the safe thing to do seems to be to not delete parents until children are completely cleaned up.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -910,31 +987,185 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
...
+			return &ctrl.Result{Requeue: true}, err
+		}
+		return nil, nil
+	}
+
+	if !statefulSet.DeletionTimestamp.IsZero() {
+		if err = r.cleanupImageServiceFinalizer(ctx, statefulSet); err != nil {
+			return &ctrl.Result{Requeue: true}, err
+		}
+		return &ctrl.Result{Requeue: true}, nil
```
mhrivnak, line 1139: Why force a requeue here?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -910,31 +987,185 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
...
+	}
+
+	log.Info("Deleting image service stateful set on volume claim template update")
+
+	// need to delete and re-create statefulset because the volumes have changed
+	if err = r.Client.Delete(ctx, existing); err != nil {
+		return &ctrl.Result{Requeue: true}, err
+	}
+
+	return &ctrl.Result{Requeue: true}, nil
```
mhrivnak, line 1168: Why force a requeue here?

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: Also I see below that if the statefulset is being deleted, that code path also calls `cleanupImageServiceFinalizer`. It's probably better to call that from just one place, unless there's a reason it needs to be in both?

mhrivnak (changes requested):

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: We could be in the case where someone changed the volume size.
That flow actually requires three loops through reconcile.
1. ASC is edited to change volume size
  - We reconcile by requesting the statefulset be deleted then requeue (your other comment)
2. Remove the finalizer
  - We reconcile here and requeue
3. Create the statefulset

It's very possible that we could get triggered by the side-effects of these actions, but I figured it was better to force a requeue rather than relying on something like the statefulset or PVs being removed to trigger a reconcile.

Does that make sense or do you think we should remove these?

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: > If it no longer exists for whatever reason, what's the expectation?

I suppose this could be handled better. I do have a condition there to ensure that we don't return if the error returned by the `Get` call is `NotFound`, but it's probably best to not assume that `statefulSet` is in any way valid after that call.
I could also wrap the `cleanupImageServiceFinalizer` call in the same condition.

> Also I see below that if the statefulset is being deleted, that code path also calls cleanupImageServiceFinalizer. It's probably better to call that from just one place, unless there's a reason it needs to be in both?

I think it's better to have it in both places. Even if we force a requeue here after removing the finalizer on ASC then we could end up in a situation where we try to run the reconcile loop without an ASC instance and I think we fail on that rather early in this function. I don't think I'd want to let that go through just to not have to call this function again.

mhrivnak: Ah I missed the `!errors.IsNotFound(err)` part. Ok. But if it doesn't exist, does it make sense to still call `cleanupImageServiceFinalizer`?

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: I think the forced requeue just invites the possibility that there will be an extra reconcile before deletion is complete and the cache has been updated as such. The delete event on the StatefulSet will cause a reconcile of the ACI to happen regardless, so I would just rely on that as part of the normal pattern:

1. Something about a watched resource changes
2. Reconcile runs and makes sure real-world state matches desired state.

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: Won't that eventually resolve though?
Alternatively if we've already removed the finalizer on the statefulset couldn't you make the same argument for any of the deployments?

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: > But if it doesn't exist, does it make sense to still call cleanupImageServiceFinalizer?

This is what I was trying to address in the first part of my last comment.
If we assume that a failed `Get` call doesn't change `statefulSet` then it's safe to call that function because the finalizer won't be present on the little stub object and we'll just return. But I can make that explicit here by re-checking `errors.IsNotFound(err)` before calling the cleanup.

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: It'll resolve, but would need to be handled gracefully by the controller.

I'm not following your point about deployments?

carbonin: >would need to be handled gracefully by the controller.

You mean the statefulset controller here, right?

> I'm not following your point about deployments?

I mean, I don't think there's anything special about a statefulset in the situation you described.
There's a parent/child relationship between ASC and every object we create so if someone deletes it and recreates a new one we'll have issues with every object we create until the old one is removed. This applies to deployments, secrets, services, etc. I don't see why we would treat statefulsets specially here outside of ensuring the finalizer is removed.

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: :+1: I'll test it out to be sure but if you think that both calling `Delete` on an object with a finalizer and also removing the finalizer will both definitely cause another reconcile then I can roll back this change.

The other nice thing this gave us was in the tests I could call reconcile until we didn't request a requeue, but I can just reconcile some set number of times to test these changes.

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: Fair point. Since we're certain by this point that the finalizer on the StatefulSet has been removed, I'm fine leaving it as-is.

carbonin:
/override ci/prow/e2e-metal-assisted-operator-ztp

This is broken everywhere

carbonin:
/retest

Errors don't look related to this PR.

PR #3072, opened 2021-12-19
UNTESTED cnv: complete storage story

dankenigsberg:
# Assisted Pull Request

## Description

Currently, when OpenShift Virtualization (CNV) is installed on a Single-node cluster,
no storage solution is provided, despite most VMs requiring persistent storage.

Once this PR is tested and merged, if CNV is installed on a single-node cluster
we also configure local storage by the HostpathProvisioner (HPP) which is
shipped with CNV.

This code should eventually be replaced by applying manifests for
TopoLVM which would be shipped with ODF.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

not tested yet

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @awels

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/operators/cnv/manifest.go
+++ b/internal/operators/cnv/manifest.go
@@ -63,6 +63,18 @@ func Manifests(config Config) (map[string][]byte, []byte, error) {
 	return openshiftManifests, cnvHco, nil
 }
 
+// SingleNodeManifests returns manifests that are needed by single-node deployments
+func SingleNodeManifests(config Config) (map[string][]byte, []byte, error) {
+	openshiftManifests, cnvHco, err := Manifests(config)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	openshiftManifests["99_openshift-cnv_hpp.yaml"] = hpp
+	openshiftManifests["99_openshift-cnv_hpp_sc.yaml"] = hppStorageClass
+	return openshiftManifests, cnvHco, nil
+}
+
 func subscription(config manifestConfig) ([]byte, error) {
 	data := map[string]string{
 		"OPERATOR_NAMESPACE":         config.Namespace,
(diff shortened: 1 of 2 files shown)


filanov (changes requested):
Please add tests

internal/operators/cnv/manifest.go
```diff
@@ -140,3 +152,34 @@ metadata:
...
+      storagePools:
+        - name: "sno"
+    pvcTemplate:
+      storageClassName: local-block
+      volumeMode: Block
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: 50Gi
```
filanov, line 171: there is a basic requirement for SNO, should it be added to the validations in case CNV is enabled?

awels: So the local-block storage class would be an LSO storage class, and the PVs created by LSO will be the size of the disk. Kubernetes will match PVs with the PVC created from this template if the PV capacity >= request size. We don't want the request size to be larger than the capacity of the LSO PV.

filanov: Is that a dedicated disk?

romfreiman: @awels in assisted, we dont want users to fail and we base everything on the discovery data. If the disk size will be 40GB, your yaml will fail, right? And the user will not be able to identify it

rollandf: Does it means we need to have also LSO ?

awels: @romfreiman with the yaml as it is yes, we picked 50Gi as it seemed a reasonable minimum, but we could go with 1Mi if we fear that we might have smaller disks.
@rollandf The example assumes we have LSO installed (from what I was told, assisted installer, automatically installs LSO) since we are using LSOs discovery feature to discover disks, and then using the PVs associated with those disks to mount in the host, and create directories on. But there is no reason we cannot use non LSO PVs.

filanov: If we know that we require a minimal then we should validate it, 1M will not be really usable for the user, lets not make shortcuts here.
We have a lot of validations that we are running before the installation and we should make sure that we validate this as well

akalenyu:
Can be closed in favor of #3131 which will require more work testing wise
thanks for the direction

PR #3073, opened 2021-12-19
MGMT-7428: added ignition to apivip check response in swagger

danielerez:
# Assisted Pull Request

## Description

Added ignition property to api_vip_connectivity_response for populating when invoking apivip_check cmd in agent.
The ignition is required to fetch disk LUKS (disk encryption) information for workers in the cluster.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @ybettan 
/cc @nmagnezi 
/cc @filanov 
/cc @rollandf 
/cc @avishayt 

## Checklist

## Reviewers Checklist

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -8374,6 +8374,9 @@ definitions:
       is_success:
         type: boolean
         description: API VIP connectivity check result.
+      ignition:
+        type: string
+        description: Ignition fetched from the specified API VIP
 
   disk_speed_check_request:
     type: object
--- a/models/api_vip_connectivity_response.go
+++ b/models/api_vip_connectivity_response.go
@@ -17,6 +17,9 @@ import (
 // swagger:model api_vip_connectivity_response
 type APIVipConnectivityResponse struct {
 
+	// Ignition fetched from the specified API VIP
+	Ignition string `json:"ignition,omitempty"`
+
 	// API VIP connectivity check result.
 	IsSuccess bool `json:"is_success,omitempty"`
 }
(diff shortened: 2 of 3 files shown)


swagger.yaml
```diff
@@ -8374,6 +8374,9 @@ definitions:
       is_success:
         type: boolean
         description: API VIP connectivity check result.
+      ignition:
```
ybettan, line 8377: I don't see any change in the code, who set this value before the user get it from the API?

danielerez: It would be used by the agent: https://github.com/openshift/assisted-installer-agent/pull/263
Will push service change in an upcoming PR.

avishayt:
The idea here is to piggyback on the ignition check to have the agent send the actual ignition, and then what?  Store it in the DB and it will cause the disk encryption validations to run?  If so, that seems OK, but why not include the full logic in the PR?

danielerez:
> The idea here is to piggyback on the ignition check to have the agent send the actual ignition, and then what? Store it in the DB and it will cause the disk encryption validations to run? If so, that seems OK, but why not include the full logic in the PR?

Right, the response is stored in host->APIVipConnectivity, so the ignition would be parsed as part of the validation. 
The full logic is in: https://github.com/openshift/assisted-service/pull/3074
Pushed a separate PR just to update the agent first with the model: https://github.com/openshift/assisted-installer-agent/pull/263

swagger.yaml
ybettan: You mean "agent change" right?!

danielerez: Meant the full validation logic:) https://github.com/openshift/assisted-service/pull/3074

PR #3074, opened 2021-12-19
MGMT-7428: disk encryption validation for day2

danielerez:
# Assisted Pull Request

## Description

Added disk encryption validation for day2 in host validator.
Introduced the following changed to host/validator:
* Added getDiskEncryptionForDay2 func for fetching LUKS (disk encryption) from the ignition, which should be available in APIVipConnectivityResponse.
* In diskEncryptionRequirementsSatisfied:
  * Used luks info from the ignition for identifying disk encryption support for workers in the cluster.
  * If TPM2 is enabled in the cluster for workers - ensure TPM2 is available in the host. Otherwise, the disk could be encrypted only using tang.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @ybettan
/cc @nmagnezi
/cc @filanov
/cc @rollandf

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/validations_test.go
+++ b/internal/host/validations_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
 	"github.com/golang/mock/gomock"
@@ -87,7 +88,6 @@ var _ = Describe("Validations test", func() {
 	}
 
 	Context("Disk encryption validation", func() {
-
 		getDiskEncryptionValidationResult := func(validationsInfo string) (ValidationStatus, string, bool) {
 
 			var validationsRes ValidationsStatus
@@ -102,6 +102,32 @@ var _ = Describe("Validations test", func() {
 				}
 			}
 			return ValidationStatus(""), "", false
+		}
+
+		getIgnitionConfig := func(tpm2Enabled bool) types.Config {
+			return types.Config{
+				Ignition: types.Ignition{Version: "3.2.0"},
+				Storage: types.Storage{
+					Luks: []types.Luks{
+						{
+							Clevis: &types.Clevis{Tpm2: &tpm2Enabled},
+							Device: swag.String("/dev/disk"),
+						},
+					},
+				},
+			}
+		}
+
+		getDay2Host := func() models.Host {
+			h := hostutil.GenerateTestHostByKind(hostID, infraEnvID, &clusterID, models.HostStatusDiscovering, models.HostKindHost, models.HostRoleWorker)
+			*h.Kind = models.HostKindAddToExistingClusterHost
+			return h
+		}
+
+		createDay2Cluster := func() {
+			c := hostutil.GenerateTestCluster(clusterID, common.TestIPv4Networking.MachineNetworks)
+			c.DiskEncryption = &models.DiskEncryption{}
+			Expect(db.Create(&c).Error).ToNot(HaveOccurred())
 		}
 
 		It("disk-encryption not set", func() {
@@ -242,7 +268,7 @@ var _ = Describe("Validations test", func() {
 			validationStatus, validationMessage, found := getDiskEncryptionValidationResult(h.ValidationsInfo)
 			Expect(found).To(BeTrue())
 			Expect(validationStatus).To(Equal(ValidationFailure))
-			Expect(validationMessage).To(Equal("TPM version could not be found, make sure TPM is enalbed in host's BIOS"))
+			Expect(validationMessage).To(Equal("TPM version could not be found, make sure TPM is enabled in host's BIOS"))
 		})
 
 		It("host's TPM version is unsupported", func() {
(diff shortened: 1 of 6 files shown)


avishayt:
This PR looks a bit weird because we're using the response from one validation as the input to another.

internal/host/validator.go
```diff
@@ -372,6 +398,27 @@ func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) Va
...
+		}
+
+		c.cluster.DiskEncryption = &models.DiskEncryption{}
+		if swag.BoolValue(luks.Clevis.Tpm2) {
+			// If Tpm2 is enabled for workers, check whether supported by the host.
+			c.cluster.DiskEncryption.Mode = swag.String(models.DiskEncryptionModeTpmv2)
+			return boolValue(c.inventory.TpmVersion == models.InventoryTpmVersionNr20)
+		}
+
+		c.cluster.DiskEncryption.Mode = swag.String(models.DiskEncryptionModeTang)
```
ybettan, line 418: This looks a bit wired, consider having a switch-case for `tpm` and `tang` and return a failing validation if non of the above is the encryption mode.

danielerez: Actually the Clevis object in Luks contains only Tpm2 and Tang: https://github.com/coreos/ignition/blob/0cbe33fee45d012515479a88f0fe94ef58d5102b/config/v3_2/types/schema.go#L5
I'll make the check more explicit then, in case Tang is empty.

avishayt:
> This PR looks a bit weird because we're using the response from one validation as the input to another.

Can't think of a better way to do it though...

internal/host/validator.go
```diff
@@ -357,9 +359,33 @@ func isDiskEncryptionEnabledForRole(encryption models.DiskEncryption, role model
 	}
 }
 
+func (v *validator) getDiskEncryptionForDay2(host *models.Host) (*types.Luks, error) {
+	var response models.APIVipConnectivityResponse
+	if err := json.Unmarshal([]byte(host.APIVipConnectivity), &response); err != nil {
+		// APIVipConnectivityResponse is not available yet - retrying.
```
ybettan, line 365: Why does an error here means that `APIVipConnectivityResponse` is unavailable? can't it mean that we just failed to unmarshal?

internal/host/validator.go
```diff
@@ -372,6 +398,30 @@ func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) Va
...
+			return ValidationSuccessSuppressOutput
+		}
+
+		c.cluster.DiskEncryption = &models.DiskEncryption{}
+		if swag.BoolValue(luks.Clevis.Tpm2) {
+			// If Tpm2 is enabled for workers, check whether supported by the host.
+			c.cluster.DiskEncryption.Mode = swag.String(models.DiskEncryptionModeTpmv2)
+			return boolValue(c.inventory.TpmVersion == models.InventoryTpmVersionNr20)
+		} else if len(luks.Clevis.Tang) != 0 {
+			c.cluster.DiskEncryption.Mode = swag.String(models.DiskEncryptionModeTang)
```
ybettan, line 417: Same questions.

internal/host/validator.go
```diff
@@ -372,6 +398,30 @@ func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) Va
...
+			return ValidationSuccess
+		} else {
+			// Only Tpm2 and Tang are available for disk encryption
+			return ValidationSuccessSuppressOutput
```
ybettan, line 421: That will mean that we have some malformed ignition, don't we want to fail the validation in this case?

internal/host/validator.go
danielerez: So as we're using the same object in the service and agent it shouldn't fail on syntax and such. But it might takes a while till the response is available (as it invoked by apivip_check validation).

internal/host/validator.go
```diff
@@ -372,6 +398,30 @@ func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) Va
...
+		}
+		if luks == nil {
+			// Disk encryption is disabled for workers on day1 cluster
+			return ValidationSuccessSuppressOutput
+		}
+
+		c.cluster.DiskEncryption = &models.DiskEncryption{}
+		if swag.BoolValue(luks.Clevis.Tpm2) {
+			// If Tpm2 is enabled for workers, check whether supported by the host.
+			c.cluster.DiskEncryption.Mode = swag.String(models.DiskEncryptionModeTpmv2)
```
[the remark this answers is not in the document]
danielerez: We still use the same logic as v1: https://github.com/openshift/assisted-service/blob/8996b3e2f89ab1434a0a968cec326d65da0be18d/internal/bminventory/inventory.go#L782

The 'transform' flow is currently available only in kube-api: https://github.com/openshift/assisted-service/blob/11486d01ab53dac3abbd1c7f24cf4954d2722c01/internal/controller/controllers/clusterdeployments_controller.go#L1042

internal/host/validator.go
danielerez: See above.

internal/host/validator.go
danielerez: hmm, not sure it's necessarily means malformed. I mean, in case a new encryption type would be added, it might be a valid option later. Maybe I'll add a warning in this case then, to make it clearer.

danielerez:
> > This PR looks a bit weird because we're using the response from one validation as the input to another.
> 
> Can't think of a better way to do it though...

Yeah, it might look a bit convoluted:) But I think it's better than adding another command in agent just for that, seems like an overkill.

internal/host/validations_test.go
```diff
@@ -289,5 +304,176 @@ var _ = Describe("Validations test", func() {
 			Expect(validationStatus).To(Equal(ValidationSuccess))
 			Expect(validationMessage).To(Equal(fmt.Sprintf("Installation disk can be encrypted using %s", models.DiskEncryptionModeTpmv2)))
 		})
+
+		It("day2 host - TPM2 for worker enabled on cluster, TPM2 available on host", func() {
+			c := hostutil.GenerateTestCluster(clusterID, common.TestIPv4Networking.MachineNetworks)
```
filanov, line 309: you can put some of the tests under a single context and have the common code written once in before each

internal/host/validator.go
```diff
@@ -372,6 +398,30 @@ func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) Va
...
+			return boolValue(c.inventory.TpmVersion == models.InventoryTpmVersionNr20)
+		} else if len(luks.Clevis.Tang) != 0 {
+			c.cluster.DiskEncryption.Mode = swag.String(models.DiskEncryptionModeTang)
+			return ValidationSuccess
```
ybettan, line 418: I think we can return `ValidationSuccessSuppressOutput` here as we didn't do any actual validation.

internal/host/validator.go
ybettan: I don't think there will be one soon, and even if it will, ATM this can be caused only as an error in the ignition.
IMO, we better have a validation error here - in addition to what I said before, it will also ease the task of finding bugs if we reach this part of the code.

danielerez: ok will return ValidationFailure then. we could just remove it later on if a new mode is added.

internal/host/validations_test.go
danielerez: yeah, didn't want to make too much refactoring as part of this change. I'll try to extract some common code to funcs.

PR #3082, opened 2021-12-20
MGMT-8477: invoke cluster validations when updating host

slaviered:
# Assisted Pull Request

## Description
When updating a host the cluster level validation may be affected. For example, when
changing a role of a host from master to worker the sufficient count validation may fail.
If the installation is confirmed quickly using the UI, it may fail with out a warning because
the monitor has no time to run again and execute the validations

This code make sure that the cluster validation runs when the host parameters are updated
without having to wait for the monitor

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -6790,12 +6790,8 @@ func (b *bareMetalInventory) V2UpdateHostLogsProgress(ctx context.Context, param
 
 func (b *bareMetalInventory) V2UpdateHostInternal(ctx context.Context, params installer.V2UpdateHostParams) (*common.Host, error) {
 	log := logutil.FromContext(ctx, b.log)
-
-	host, err := common.GetHostFromDB(b.db, params.InfraEnvID.String(), params.HostID.String())
-	if err != nil {
-		log.WithError(err).Errorf("failed to find host <%s>, infra env <%s>", params.HostID, params.InfraEnvID)
-		return nil, common.NewApiError(http.StatusNotFound, err)
-	}
+	var c *models.Cluster
+	var cluster *common.Cluster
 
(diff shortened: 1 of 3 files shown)


internal/bminventory/inventory.go
```diff
@@ -6977,7 +6977,17 @@ func (b *bareMetalInventory) refreshAfterUpdate(ctx context.Context, host *commo
 	err := b.hostApi.RefreshStatus(ctx, &host.Host, db)
 	if err != nil {
 		log.WithError(err).Errorf("Failed to refresh host %s, infra env %s during update", host.ID, host.InfraEnvID)
+		return err
+	}
+
+	if cluster != nil {
```
filanov, line 6983: i'm thinking about how to test it, does cluster refresh include the changes in the host? 
Maybe it need to run with `for update` option? 
A scenario here is a ready cluster with 3 masters, user change one of the hosts role into worker, cluster status will remain ready until the next iteration of the monitor. i think that this TX need a `for update` to include the changes in the host. 
Because in v2 host is separated from the cluster UI will need to get the cluster again after they apply a configuration on the host @jkilzi

jkilzi: ack

ori-amizur: We need for update option on all entities that might be updated.  But in this case, the query must be inside the transaction

filanov (approved):

PR #3131, opened 2021-12-28
MGMT-8197: Deploy HPP when CNV is requested on SNO (so persistent VMs are possible)

akalenyu:
This PR replaces https://github.com/openshift/assisted-service/pull/3072
(Thanks @dankenigsberg for the directions).

# Assisted Pull Request

## Description
Currently, when OpenShift Virtualization (CNV) is installed on a Single-node cluster,
no storage solution is provided, despite most VMs requiring persistent storage.

Once this PR is tested and merged, if CNV is installed on a single-node cluster
we also configure local storage by the HostpathProvisioner (HPP) which is
shipped with CNV.

This code should eventually be replaced by applying manifests for
TopoLVM which would be shipped with ODF.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

figuring out testing
- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @awels

## Checklist

## Reviewers Checklist


internal/operators/cnv/manifest.go
```diff
@@ -140,3 +144,34 @@ metadata:
...
+const cnvHPPManifest = `apiVersion: hostpathprovisioner.kubevirt.io/v1beta1
+kind: HostPathProvisioner
+metadata:
+  name: hostpath-provisioner
+    spec:
+      imagePullPolicy: IfNotPresent
+      storagePools:
+        - name: sno
+    pvcTemplate:
+      storageClassName: localblock-sc
```
akalenyu, line 157: I have seen this storage class name referenced throughout the repo but couldn't find whats in charge of setting it
Example occurrence:
https://github.com/openshift/assisted-service/blob/master/internal/operators/lso/manifest.go#L71

rollandf: This will be available only if LSO is installed.
Is this a requirement for CNV+SNO?

rollandf:
No tests?

internal/operators/cnv/manifest.go
```diff
@@ -140,3 +144,34 @@ metadata:
...
+      volumeMode: Block
+      accessModes:
+      - ReadWriteOnce
+      resources:
+        requests:
+          storage: 50Gi
```
rollandf, line 163: What this number means?
What will happen if there is a disk with less space?

internal/operators/cnv/manifest.go
akalenyu: Yes, we want this to be a requirement for CNV+SNO
In what flows is LSO disabled?

We could also use other storage options but I don't think we have a better option than LSO
which can already be installed here

internal/operators/cnv/manifest.go
akalenyu: This is the capacity of the pool that the HPP can use to dynamically provision storage on the host's path.
We could work with less, but that would mean less persistent VMs running alongside each other, 50Gi seems fair.

Are you suggesting we derive this value from discovery? any tips on approaching that?

akalenyu:
> No tests?

Trying to figure out my way around tests, have a todo list up in PR's description. Thanks for ok-to-test btw.

internal/operators/cnv/manifest.go
filanov: The same comments from the original pr. if we have minimal requirement it should be validated.

filanov: https://github.com/openshift/assisted-service/pull/3072#discussion_r771971942

internal/operators/cnv/manifest.go
rollandf: Sorry, LSO is a requirement for CNV:
https://github.com/openshift/assisted-service/blob/master/internal/operators/cnv/cnv_operator.go#L59

internal/operators/cnv/manifest.go
```diff
@@ -14,6 +15,8 @@ const (
 
 	upstreamSourceName   string = "community-kubevirt-hyperconverged"
 	downstreamSourceName string = "kubevirt-hyperconverged"
+
+	discoverableDiskSizeThresholdGi int64 = 50
```
filanov, line 19: suggesting to make it configurable (env var) it will be easier to change, instead of releasing a new version.

dankenigsberg: Are env variables exposed to end users who install a cluster?

rollandf: No, it is a configuration that can be changed from the deployment yaml.
This is a validation, meaning that if the user have an additional smaller disk, he will not be able to continue with the installation.

internal/operators/cnv/cnv_operator.go
```diff
@@ -274,3 +281,16 @@ func sanitizeID(id string) string {
...
+// with certain size threshold for the hpp storage pool
+func validDiscoverableSNODisk(disks []*models.Disk, installationDiskID string) error {
+	for _, disk := range disks {
+		if (disk.DriveType == ocs.SsdDrive || disk.DriveType == ocs.HddDrive) && installationDiskID != disk.ID && disk.SizeBytes != 0 {
+			if disk.SizeBytes > conversions.GibToBytes(discoverableDiskSizeThresholdGi) {
+				return nil
+			}
+		}
+	}
+	return fmt.Errorf("CNV with SNO requires a discoverable disk with %d Gi", discoverableDiskSizeThresholdGi)
```
rollandf, line 295: Not sure about the message here. What discoverable means? Is CNV something that the user knows about?

rollandf: Though I see `CNV` used in other messages ....

internal/operators/cnv/manifest.go
dankenigsberg: So frankly, I do not see a point in making the size configurable. We may realize in the future that there are users who do not need storage for SNO+CNV; but then we should not lower discoverableDiskSizeThresholdGi, we'd need to drop the entire validation.

internal/operators/cnv/cnv_operator.go
dankenigsberg: But you are right: the user-facing term should be `OpenShift Virtualization`, and should be taken from a variable, to allow "whitelabeling" of assisted-installer for upstream OKD.

internal/operators/cnv/manifest.go
rollandf: The price in the code is not that high, comparing to have to release a new image.
Also it allows QE to tweak values if needed.

Need something like that:
`OCSMinDiskSizeGB                int64             envconfig:"OCS_MIN_DISK_SIZE_GB" default:"25"`

internal/operators/cnv/cnv_operator.go
akalenyu: Tweaked msg according to suggestion and changed the validation msgs from `CNV->OpenShift Virtualization`
Regarding whitelabeling assisted-service, would that be OK to not tackle in this PR?

internal/operators/cnv/manifest.go
akalenyu: Made is a configurable env var under the CNV Config

internal/operators/cnv/cnv_operator.go
dankenigsberg: Sure, particularly if defining a `VirtProduct` constant would cause you to touch more modules. PRs should be concise.

internal/operators/cnv/manifest.go
dankenigsberg: IMHO the most (only?) interesting config is to skip HPP validation and configuration if the value is 0.

akalenyu: Added opt-out-of-installing HPP env var

internal/operators/cnv/manifest.go
```diff
@@ -57,6 +58,14 @@ func Manifests(config Config) (map[string][]byte, []byte, error) {
 
 	openshiftManifests := make(map[string][]byte)
 
+	if isSingleNodeCluster && config.SNOInstallHPP {
+		cnvHpp, err := hpp(config.SNOPoolSizeRequestHPPGib)
+		if err != nil {
+			return nil, nil, err
+		}
+		openshiftManifests["99_openshift-cnv_hpp.yaml"] = cnvHpp
```
filanov, line 66: Suggesting to avoid using 99 because it will prevent form the user the option to override this config.

internal/operators/cnv/manifest.go
```diff
@@ -94,6 +103,13 @@ func hco(config manifestConfig) ([]byte, error) {
 	return executeTemplate(data, "cnvHCO", cnvHCOManifestTemplate)
 }
 
+func hpp(diskThresholdGi int64) ([]byte, error) {
```
pkliczewski, line 106: Please use more meaningful name

internal/operators/cnv/manifest.go
pkliczewski: We use this prefix for all the operator manifests. Do you want us to change it everywhere (work for other PR)?

akalenyu: We use this throughout quite often in other operators (OCS/LSO) as well:
https://github.com/openshift/assisted-service/blob/master/internal/operators/ocs/manifest.go#L53

internal/operators/cnv/config.go
```diff
@@ -13,6 +13,10 @@ type Config struct {
 	SupportedSRIOVNetworkIC DeviceIDDecoder `envconfig:"CNV_SUPPORTED_SRIOV_NICS" default:"8086:158b,15b3:1015,15b3:1017,15b3:1013,15b3:101b"`
 	// CNV operator mode. It defines whether to use upstream `false` or downstream `true`
 	Mode bool `envconfig:"CNV_MODE" default:"true"`
+	// HPP install flag. It defines whether we should install HPP when CNV is on SNO
+	SNOInstallHPP bool `envconfig:"CNV_SNO_INSTALL_HPP" default:"true"`
```
pkliczewski, line 17: Please update documentation so people know how to use it and add support in https://github.com/openshift/assisted-test-infra.

pkliczewski:
@machacekondra please take a look

internal/operators/cnv/config.go
akalenyu: Is there a doc of these I am missing?

internal/operators/cnv/manifest.go
akalenyu: This is the shorthand for hostpath provisioner, similarly to how we shorthand cnv/hco:
https://github.com/openshift/assisted-service/pull/3131/files#diff-448ecdc1e6c9432844dd5dd2c0fe9ed770432db8e41ca79ef62806c6fe677ad0R99

internal/operators/cnv/manifest.go
filanov: It can be done in a different PR, but it's something that needs to be handled, 99 will mean that any configuration that user will try to override will be ignored.

internal/operators/cnv/config.go
pkliczewski: Please add a section how to enable hpp [here](https://github.com/openshift/assisted-service/blob/140fe0cc2726d73077e6737a8a65f178d2ea6875/docs/dev/olm-operator-plugins.md)

machacekondra (approved):

internal/operators/cnv/config.go
filanov: and add the env var to https://github.com/openshift/assisted-service/blob/master/openshift/template.yaml

filanov:
/lgtm cancel 
still need to do some fixes.

internal/operators/cnv/config.go
akalenyu: > and add the env var to https://github.com/openshift/assisted-service/blob/master/openshift/template.yaml

Added, defaults to `false` until 4.10 testing begins, does that make sense? I have tested this against a hacked assisted-test-infra environment with 4.10 OC & CNV, but I understand we are still not testing those (waiting for FC?)

> Please add a section how to enable hpp [here](https://github.com/openshift/assisted-service/blob/140fe0cc2726d73077e6737a8a65f178d2ea6875/docs/dev/olm-operator-plugins.md)

No problem, I just want to make sure this is the correct place for it - it seems like an OLM operator addition development doc

filanov: Setting as default is fine, but if we want to enable it in the cloud i prefer not get another release, so adding it to the template will enable us just to change app-sre yaml and don't touch the code. 
As far as i know we didn't start to test 4.10, @osherdp correct me if i'm wrong

osherdp: no, we haven't integrated 4.10 yet
still waiting or a fc payload

filanov:
@pkliczewski @akalenyu as far as i understand this feature is relevant only for 4.10, is the code handles installations of other versions as well? what will happen if someone try to install 4.9?

pkliczewski:
> @pkliczewski @akalenyu as far as i understand this feature is relevant only for 4.10, is the code handles installations of other versions as well? what will happen if someone try to install 4.9?

We always use stable channel for particular ocp version so if 4.9 installation is supported it will attempt to use it as well. @akalenyu how this code would work for this version?

akalenyu:
> @pkliczewski @akalenyu as far as i understand this feature is relevant only for 4.10, is the code handles installations of other versions as well? what will happen if someone try to install 4.9?

Handled in last commit. No 4.10 - no party



> > @pkliczewski @akalenyu as far as i understand this feature is relevant only for 4.10, is the code handles installations of other versions as well? what will happen if someone try to install 4.9?
> 
> We always use stable channel for particular ocp version so if 4.9 installation is supported it will attempt to use it as well. @akalenyu how this code would work for this version?

Thanks for the explanation on the subscriptions, the code would now not install if we're not at 4.10

pkliczewski (commented):
/lgtm

pkliczewski:
@filanov @rollandf what do you think about this change?

internal/operators/cnv/cnv_operator.go
```diff
@@ -115,14 +122,14 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
 
 	cpu := requirements.CPUCores
 	if inventory.CPU.Count < cpu {
-		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{fmt.Sprintf("Insufficient CPU to deploy CNV. Required CPU count is %d but found %d ", cpu, inventory.CPU.Count)}}, nil
+		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{fmt.Sprintf("Insufficient CPU to deploy OpenShift Virtualization. Required CPU count is %d but found %d ", cpu, inventory.CPU.Count)}}, nil
```
filanov, line 125: please break long lines

internal/operators/cnv/manifest.go
```diff
@@ -57,6 +63,16 @@ func Manifests(config Config) (map[string][]byte, []byte, error) {
...
 
+	if shouldInstallHPP(config, cluster) {
+		cnvHpp, err := hpp(config.SNOPoolSizeRequestHPPGib)
+		if err != nil {
+			return nil, nil, err
+		}
+		// Add HostPathProvisioner to CNV manifest
+		cnvHco = append(cnvHco, []byte("\n---\n")...)
+		cnvHco = append(cnvHco, cnvHpp...)
+		openshiftManifests["99_openshift-cnv_hpp_sc.yaml"] = []byte(cnvHPPStorageClass)
```
filanov, line 74: Why 99?
@romfreiman  if i remember correctly we wanted to avoid using `99_`  prefix right?

internal/operators/cnv/manifest.go
```diff
@@ -57,6 +63,16 @@ func Manifests(config Config) (map[string][]byte, []byte, error) {
 
 	openshiftManifests := make(map[string][]byte)
 
+	if shouldInstallHPP(config, cluster) {
+		cnvHpp, err := hpp(config.SNOPoolSizeRequestHPPGib)
```
filanov, line 67: can/should users override that? is that what is going to be used or is it going to take full disk capacity?

internal/operators/cnv/manifest.go
akalenyu: I think we decided this should be handled outside of this PR?
https://github.com/openshift/assisted-service/pull/3131#discussion_r777884854
https://github.com/openshift/assisted-service/pull/3131#discussion_r777424333

internal/operators/cnv/manifest.go
akalenyu: Users can override this:
https://github.com/openshift/assisted-service/blob/68d297e4e44585c71d71a28990fab3a0a7d96aae/internal/operators/cnv/config.go#L16-L17
But they should not. If they want persistent VMs (wide use case) on SNO, they likely want a solution like HPP to allow them.
We are requesting `SNOPoolSizeRequestHPPGib` but get attached to the full LSO disk PV which can be bigger.

filanov: If thats only the minimum then you  probably should add a comment explaining that

internal/operators/cnv/manifest.go
filanov: OK do you have a ticket for this?

internal/operators/cnv/manifest.go
akalenyu: This is not specific to this situation; it has to do with how k8s will match a PVC with a PV
`(Kubernetes will match PVs with the PVC created from this template if the PV capacity >= request size)`

internal/operators/cnv/manifest.go
filanov: https://issues.redhat.com/browse/MGMT-8851

akalenyu: Thanks!

filanov (approved):

PR #3134, opened 2021-12-29
MGMT-8624: Enable downloading kubeconfig when he is available

michaellevy101:
Allow downloading after the installation regardless on the status (even for failed or pending-user-action statuses)

# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


internal/bminventory/inventory.go
```diff
@@ -4648,8 +4648,6 @@ func (b *bareMetalInventory) checkFileForDownload(ctx context.Context, clusterID
 	}
 
 	switch fileName {
-	case constants.Kubeconfig:
```
romfreiman, line 4651: can u verify that there are no issues in the UI?

filanov (commented):
Why there is no test coverage?

michaellevy101:
> Why there is no test coverage?

That why i set it on hold, need to test it manually first

romfreiman:
@michaellevy101 u're changing the behavior here. Are we certain that we're good with the new behavior?
@tsorya @eranco74 @avishayt

internal/bminventory/inventory.go
```diff
@@ -6741,11 +6744,19 @@ func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params
 }
 
 func (b *bareMetalInventory) V2DownloadClusterCredentials(ctx context.Context, params installer.V2DownloadClusterCredentialsParams) middleware.Responder {
+	fileName := params.FileName
 	respBody, contentLength, err := b.V2DownloadClusterCredentialsInternal(ctx, params)
+
+	if err != nil && params.FileName == constants.Kubeconfig {
```
eranco74, line 6750: Add a comment on why this is necessary.

eranco74: Consider moving this logic into `V2DownloadClusterCredentialsInternal`

eranco74:
> @michaellevy101 u're changing the behavior here. Are we certain that we're good with the new behavior? @tsorya @eranco74 @avishayt

Assuming no one care about the ingress cert.
I think the new logic is better.
I think the UI would like it as well (no need to calculate which kubeconfig to get)

internal/bminventory/inventory.go
tsorya: why not to verify if object exists?

avishayt:
1. If nobody cares about the ingress cert, then let's remove kubeconfig-noingress from the API, and give the user a single option for downloading kubeconfig (the logic here).
2. If some users do care, then this logic makes it awkward to get the kubeconfig with ingress.  I suggest either (1) a way to list the available files via API and let the user choose, (2) keeping this logic and adding a kubeconfig-ingress option that has the current logic.

internal/bminventory/inventory.go
michaellevy101: @tsorya @eranco74 @romfreiman The kubeconfig (with ca) isn't available before the finalize stage so we should downgrade to the no-ingress kubeconfig. on the other hand it changes the API.
It's just an option, still consulting with @avishayt

michaellevy101: We had a dedicated API for getting the kubeconfig file that did a similar logic but we want to keep it simple and have only single generic API for downloading files so we deprecated it.
The Kubeconfig ingress file is required only for oauth logic so for the most users the no-igress kubeconfig is enough.
The idea here is to provide the best kubeconfig we have, so at the beginning we provide the no-ingress kubeconfig and after the finalize stage we download the kubeconfig with the ingress ca.

avishayt:
Please address comments by @eranco74 and @tsorya

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -433,7 +433,15 @@ func (b *bareMetalInventory) RegenerateInfraEnvSigningKey(ctx context.Context, p
 func (b *bareMetalInventory) V2GetPresignedForClusterCredentials(ctx context.Context, params installer.V2GetPresignedForClusterCredentialsParams) middleware.Responder {
 	log := logutil.FromContext(ctx, b.log)
 
-	if err := b.checkFileDownloadAccess(ctx, params.FileName); err != nil {
+	fileName := params.FileName
```
filanov, line 436: please add tests to this function as well

internal/bminventory/inventory.go
michaellevy101: @eranco74 I prefer keep the internal method simple if possible and do this fallback on the external API layer. 
In addition please note the downloaded file name is set by the external method.
`	return filemiddleware.NewResponder(installer.NewV2DownloadClusterCredentialsOK().WithPayload(respBody), fileName, contentLength)`


(anyway, and external method sets the downloaded file name)

michaellevy101: @tsorya right, The V2DownloadClusterCredentialsInternal method do both: checks the file existence and downloads it.

internal/bminventory/inventory_v2_handlers.go
filanov: Right, i didn't see that it is in the same package.

internal/bminventory/inventory.go
avishayt: So please check if the error is not-found, because now you will fall back for every kind of error

michaellevy101: @avishayt now we have two potential errors StatusBadRequest or StatusConflict one for file not found and the second for wrong status and in the future we might have more errors to deal with. prefer keep the code simple and clean with no contract between the two methods. in the worse case we will check something twice.

avishayt (approved):

avishayt (approved):

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -443,13 +443,29 @@ func (b *bareMetalInventory) V2GetPresignedForClusterCredentials(ctx context.Con
...
+	fullFileName := fmt.Sprintf("%s/%s", params.ClusterID.String(), fileName)
 	duration, _ := time.ParseDuration("10m")
-	url, err := b.objectHandler.GeneratePresignedDownloadURL(ctx, fullFileName, params.FileName, duration)
+
+	// Kubeconfig-noingress has been created during the installation, but it does not have the ingress CA.
+	// At the finalizing phase, we create the kubeconfig file and add the ingress CA.
+	// An ingress CA isn't required for normal login but for oauth login which isn't a common use case.
+	// Here we fallback to the kubeconfig-noingress for the kubeconfig filename.
+	if fileName == constants.Kubeconfig {
+		exists, _ := b.objectHandler.DoesObjectExist(ctx, fullFileName)
```
eranco74, line 455: You are ignoring a potential error, 
I think you can keep trying the rest of the function but should log the error if it's not nil

PR #3154, opened 2022-01-04
MGMT-8047: Approve the CSRs after the installation

ori-amizur:
# Assisted Pull Request

## Description

Day1 CSRs are approved by assisted controller.  Since the controller is
not aware of day2 workers, until now these CSRs needed to be approved
manually.
This modification, adds capability to assisted service, to approve these
CSRs.  In order for this to work, cluster kubeconfig has to be available in S3
(which is not always true for day2 hosts), and the cluster needs to be
available when using this kubeconfig.
There are two parts to this solution.
In host monitoring, each day2 host that is not in done stage, is checked against its corresponding node
(using kubeapi).  If the node is ready then the host is moved to done.
In case there is a problem acccessing the day1 cluster and the day2 host
is rebooting, it is moved to done (as it is today).  In this case it is
the user responsibility to approve the CSRs.
In cluster monitoring, if there is a day2 cluster with hosts that are
not in done stage yet, it tries to approve all the CSRs using kubeapi.
For the usage of these monitoring loops, a package with the needed
kubeapi operations was created.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @tsorya 
/cc @filanov 

## Checklist

## Reviewers Checklist


filanov:
cc: @flaper87

filanov:
cc: @romfreiman

filanov:
cc: @avishayt

filanov:
I'm not sure that our monitor should access the installed cluster with kubeconfig, what will happen if the cluster is using proxy?

ori-amizur:
> 

How can we approve without it ?

eranco74:
is checked against its corresponding node
(using kubeapi).??
From where? the assisted-service?
The assisted-service don't have access to the user cluster.

ori-amizur:
> to the user cluster.

When using kubeapi, it accesses the user cluster.  Any difference ?

filanov:
The difference is that it's part of a different product that assumes that it have connection to the spoke cluster, here we cannot know that.

ori-amizur:
The question is, how can it be done otherwise ?

eranco74:
This feature seems like something that should just work for ZTP (create a machine with the matching InterlnaIP and hostname).
For the cloud use case, it requires some component that runs on the installed cluster.

filanov:
I agree the plan was for ZTP.
Here is how BMH controller take the client https://github.com/openshift/assisted-service/blob/ca9a803e06173cbfc6ec543ce31fa17a96ed35c0/internal/controller/controllers/bmh_agent_controller.go#L763

ori-amizur:
> I agree the plan was for ZTP. Here is how BMH controller take the client
> 
> https://github.com/openshift/assisted-service/blob/ca9a803e06173cbfc6ec543ce31fa17a96ed35c0/internal/controller/controllers/bmh_agent_controller.go#L763

If the assisted succeeds to connect the spoke from this location, it will succeed in the other place.

filanov:
But it will work only in disconnected environment, in the cloud we don't have access to the clusters.

tsorya:
We can't push such code to SAAS solution, we shouldn't even try to connect to users cluster.

tsorya:
please move it to controller and use it only in ztp flow

PR #3166, opened 2022-01-06
NO-ISSUE: Add a middleware to optionally disable v1 API

carbonin:
## Description

This will allow us to easily ensure that clients are no longer using v1
without removing the implementation entirely

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @osherdp 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -143,6 +143,7 @@ var Options struct {
 	EnableElasticAPM               bool          `envconfig:"ENABLE_ELASTIC_APM" default:"false"`
 	WorkDir                        string        `envconfig:"WORK_DIR" default:"/data/"`
 	LivenessValidationTimeout      time.Duration `envconfig:"LIVENESS_VALIDATION_TIMEOUT" default:"5m"`
+	V1APIEnabled                   bool          `envconfig:"V1_API_ENABLED" default:"true"`
 }
 
 func InitLogs() *logrus.Entry {
@@ -466,6 +467,10 @@ func main() {
 		OperatorsAPI:          operatorsHandler,
 	})
 	failOnError(err, "Failed to init rest handler")
+
+	if !Options.V1APIEnabled {
+		h = app.DisableV1Middleware(h)
+	}
 
 	if Options.Auth.AllowedDomains != "" {
 		allowedDomains := strings.Split(strings.ReplaceAll(Options.Auth.AllowedDomains, " ", ""), ",")
--- a/pkg/app/middleware.go
+++ b/pkg/app/middleware.go
@@ -2,6 +2,7 @@ package app
 
 import (
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/openshift/assisted-service/pkg/thread"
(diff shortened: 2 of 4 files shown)


carbonin:
Left this as WIP until I can deploy it locally with the new env var.

pkg/app/middleware.go
```diff
@@ -61,3 +62,14 @@ func SetupCORSMiddleware(handler http.Handler, domains []string) http.Handler {
 	})
 	return corsHandler.Handler(handler)
 }
+
+func DisableV1Middleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if strings.Contains(r.URL.Path, "/assisted-install/v1/") {
+			w.WriteHeader(http.StatusNotFound)
+			w.Write([]byte("v1 API is currently disabled"))
```
carbonin, line 70: Lint is failing here but I wouldn't do anything differently if this write fails.
Is it worth passing in a logger just so I can log that we failed to write this message?

filanov: I think that something like this will fix the lint error.
Besides log there is nothing we really can do. 
```suggestion
			_, _ = w.Write([]byte("v1 API is currently disabled"))
```

pkg/app/middleware.go
```diff
@@ -61,3 +62,14 @@ func SetupCORSMiddleware(handler http.Handler, domains []string) http.Handler {
 	})
 	return corsHandler.Handler(handler)
 }
+
+func DisableV1Middleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if strings.Contains(r.URL.Path, "/assisted-install/v1/") {
+			w.WriteHeader(http.StatusNotFound)
```
filanov, line 69: @avishayt what do you think is the right error code to return here?

avishayt: Agree with this - it's a path that doesn't exist

cmd/main.go
```diff
@@ -143,6 +143,7 @@ var Options struct {
 	EnableElasticAPM               bool          `envconfig:"ENABLE_ELASTIC_APM" default:"false"`
 	WorkDir                        string        `envconfig:"WORK_DIR" default:"/data/"`
 	LivenessValidationTimeout      time.Duration `envconfig:"LIVENESS_VALIDATION_TIMEOUT" default:"5m"`
+	V1APIEnabled                   bool          `envconfig:"V1_API_ENABLED" default:"true"`
```
filanov, line 146: Need to add it to https://github.com/openshift/assisted-service/blob/master/openshift/template.yaml 
it will enable use to use it on staging and prod

PR #3168, opened 2022-01-07
MGMT-8510: Duplicate hosts status messages on events

paul-maidment:
# Assisted Pull Request

## Description

Event messages on the assisted installer UI have redundant text in them, this text should be removed.

Steps to reproduce:
	Install cluster 
	Get events
	
Examples:
"Host master-0-0: updated status from preparing-successful to installing Installation is in progress(Installation is in progress)"
"Host master-0-0: updated status from installing to installing-in-progress Starting installation(Starting installation)"

The fix removes the duplication while leaving brackets around the status, the messages above will look like
"Host master-0-0: updated status from preparing-successful to installing (Installation is in progress)"
"Host master-0-0: updated status from installing to installing-in-progress (Starting installation)"

In addition to this update, changes have been made to tests to allow us to assert that this has been correctly applied. This also involved some updates to events so that these could be supported and checked in tests.

- [ ] New Feature
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

This has been tested this by running this branch of the assisted-service on assisted-test-infra and then visting the UI to perform the steps to reproduce.
On the current master branch the issue persists and on this branch it has been resolved. So a manual test has been performed.

This PR relies on only a change to unit tests and all unit tests are passing.

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @nmagnezi 
/cc @masayag 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/events/events.go
+++ b/internal/common/events/events.go
@@ -62,6 +62,8 @@ func (e *CancelInstallStartFailedEvent) GetClusterId() strfmt.UUID {
     return e.ClusterId
 }
 
+
+
 func (e *CancelInstallStartFailedEvent) format(message *string) string {
     r := strings.NewReplacer(
         "{cluster_id}", fmt.Sprint(e.ClusterId),
@@ -124,6 +126,8 @@ func (e *CancelInstallCommitFailedEvent) GetSeverity() string {
 func (e *CancelInstallCommitFailedEvent) GetClusterId() strfmt.UUID {
     return e.ClusterId
 }
+
+
 
 func (e *CancelInstallCommitFailedEvent) format(message *string) string {
     r := strings.NewReplacer(
@@ -208,6 +212,8 @@ func (e *HostRegistrationSettingPropertiesFailedEvent) GetInfraEnvId() strfmt.UU
     return e.InfraEnvId
 }
 
+
+
 func (e *HostRegistrationSettingPropertiesFailedEvent) format(message *string) string {
     r := strings.NewReplacer(
         "{host_id}", fmt.Sprint(e.HostId),
@@ -287,6 +293,8 @@ func (e *ClusterRegistrationFailedEvent) GetClusterId() strfmt.UUID {
     return e.ClusterId
 }
 
+
+
 func (e *ClusterRegistrationFailedEvent) format(message *string) string {
     r := strings.NewReplacer(
         "{cluster_id}", fmt.Sprint(e.ClusterId),
@@ -359,6 +367,8 @@ func (e *ClusterRegistrationSucceededEvent) GetClusterId() strfmt.UUID {
     return e.ClusterId
 }
 
+
+
 func (e *ClusterRegistrationSucceededEvent) format(message *string) string {
     r := strings.NewReplacer(
         "{cluster_id}", fmt.Sprint(e.ClusterId),
@@ -430,6 +440,8 @@ func (e *ClusterDeregisterFailedEvent) GetClusterId() strfmt.UUID {
     return e.ClusterId
 }
 
+
+
 func (e *ClusterDeregisterFailedEvent) format(message *string) string {
     r := strings.NewReplacer(
         "{cluster_id}", fmt.Sprint(e.ClusterId),
@@ -493,6 +505,8 @@ func (e *ClusterDeregisteredEvent) GetSeverity() string {
 func (e *ClusterDeregisteredEvent) GetClusterId() strfmt.UUID {
     return e.ClusterId
 }
+
+
 
 func (e *ClusterDeregisteredEvent) format(message *string) string {
     r := strings.NewReplacer(
@@ -570,6 +584,8 @@ func (e *ClusterValidationFailedEvent) GetSeverity() string {
 func (e *ClusterValidationFailedEvent) GetClusterId() strfmt.UUID {
     return e.ClusterId
 }
+
+
 
 func (e *ClusterValidationFailedEvent) format(message *string) string {
     r := strings.NewReplacer(
@@ -650,6 +666,8 @@ func (e *ClusterValidationFixedEvent) GetClusterId() strfmt.UUID {
     return e.ClusterId
 }
 
+
+
 func (e *ClusterValidationFixedEvent) format(message *string) string {
     r := strings.NewReplacer(
         "{cluster_id}", fmt.Sprint(e.ClusterId),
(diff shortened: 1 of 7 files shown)


internal/common/events/events.go
```diff
@@ -16,7 +16,7 @@ import (
 //
 type CancelInstallStartFailedEvent struct {
     eventName string
-    ClusterId strfmt.UUID
+        ClusterId strfmt.UUID
```
carbonin, line 19: All this extra indentation shouldn't be changing.

tools/generate_events.py
```diff
@@ -34,8 +34,17 @@
 type {{eventName}} struct {
     eventName string
 {%- for p, t in event.properties.items() %}
-    {{event.pascal_case(p)}} {{event.go_type(t)}}
+    {%- if event.type != "host" %}
+        {{event.pascal_case(p)}} {{event.go_type(t)}}
```
carbonin, line 38: Even though the indentation here makes this more readable, I think this is what's causing the indentation issues in the generated file. Probably just remove the indentation from the part that's going to render the go source.

carbonin (commented):
Looks good, just need to fix the indentation in the generated events code.

tools/generate_events.py
```diff
@@ -34,8 +34,17 @@
...
+    {%- if event.type != "host" %}
     {{event.pascal_case(p)}} {{event.go_type(t)}}
+    {%- else -%}
+    {%- if event.pascal_case(p) != "Info" %}
+    {{event.pascal_case(p)}} {{event.go_type(t)}}
+    {%- endif %}
+    {%- endif %}
 {%- endfor %}
+    {%- if event.type == "host" %}
+    Info string
```
carbonin, line 46: I think the right place to add this would be in the event definition file [here](https://github.com/openshift/assisted-service/blob/e781cea7e66fb4a75537961d5569a917491be8fd/docs/events.yaml).

That said, it makes me wonder if it's worth adding a currently unused field to all host event types seemingly just to be able to test this patch.

Will have to think about this a bit more and may want some input from @nmagnezi

nmagnezi (changes requested):
Hi Paul, welcome aboard!

Can you please explain the value in asserting SrcStatus and NewStatus for events?
We do not store those as fields for events in D.B. Those fields are merely meant to format the event msg correctly.

It looks like the problem we should address is how we format the message for HostStatusUpdatedEvent.

https://github.com/openshift/assisted-service/blob/5a661f1fedd3e3e36aaee7d988d86dcbd73a5ece/internal/common/events/events.go#L2245

What do you think?

P.S.
Some more helpful information about events: https://github.com/openshift/assisted-service/blob/master/docs/dev/events.md

eranco74:
/test e2e-assisted-capi-provider-agent-hypershift
The test was broken, verifying the fix using this PR...

paul-maidment:
> 

SourceStatus and NewStatus assertions were just for consistency with the other fields in the event, they do not add value to the PR and I will remove those.

Regarding the point on formatting, the issue looks a little more nuanced.

"Host {host_name}: updated status from {src_status} to {new_status} {info}" 

Looking at the reported message
"Host master-0-0: updated status from preparing-successful to installing Installation is in progress(Installation is in progress)"

```
{host_name} = master-0-0
{src_status} = preparing-successful
{new_status} = installing
{info} = Installation is in progress(Installation is in progress)
```

My interpretation of the ticket was that the repetition in info was to be addressed, making the desired output something like.
```
{host_name} = master-0-0
{src_status} = preparing-successful
{new_status} = installing
{info} = (Installation is in progress)
```

I therefore assumed the bug to be more around the content of the {info} field which was originally set using the following code

`statusInfo += fmt.Sprintf("(%s)", statusInfo)`

I simply removed the concatenation 

`statusInfo = fmt.Sprintf("(%s)", statusInfo)`

Assuming that we do not wish to remove the {info} completely from the event message, I believe that the format string should remain the same. Happy to correct this if I am mistaken though.

tools/generate_events.py
```diff
@@ -112,6 +112,12 @@
 }
 {%- endif %}
 
+{%- if event.name == "host_status_updated" %}
```
carbonin, line 115: I wonder if we can define this based on `event.properties.info` like is being done for `cluster_id` above?

internal/events/api/event.go
```diff
@@ -65,3 +65,8 @@ type InfraEnvEvent interface {
 	GetInfraEnvId() strfmt.UUID
 	GetClusterId() *strfmt.UUID
 }
+
+type HostUpdateStatusEvent interface {
+	HostEvent
```
carbonin, line 70: If we make sure that every event with an `Info` field implements this interface, then we can make this something like `InfoEvent` and only include `BaseEvent`'s interface here.

internal/events/eventstest/events_test_utils.go
```diff
@@ -114,6 +114,19 @@ func WithMessageContainsMatcher(expected string) eventPartMatcher {
 	}
 }
 
+func WithInfoMatcher(expected string) eventPartMatcher {
+	return func(event interface{}) bool {
+		e, ok := event.(eventsapi.HostUpdateStatusEvent)
```
carbonin, line 119: If we change the interface to only ensure we have a `GetInfo` method then this becomes much less specific to the event type.

carbonin (changes requested):
I like the direction you went here, but I think we can take it a bit further and it will end up more clean.

We're trying to be able to validate the result of the `GetInfo` method on any host with this new matcher, right? So we really want an interface that tells us that we have a `GetInfo`.

In the generator we can implement the method if the parameter is present and define an interface that is just about whether the event has additional info.

Using that new interface in the matcher makes it much more flexible. Now it will work for any type of event as soon as the `Info` field is added.

tools/generate_events.py
filanov: +1

internal/events/eventstest/events_test_utils.go
paul-maidment: @nmagnezi Nick has been helping me with my PR and we'd like to get your input on these thoughts if OK?

nmagnezi: I see those specific tests, as opposed to other tests, don't really use the db. so yes - `GetInfo` makes sense and it should be enough to test against it.

carbonin (approved):

PR #3169, opened 2022-01-07
MGMT-8047: Approve the CSRs after the installation

ori-amizur:
# Assisted Pull Request
Day1 CSRs are approved by assisted controller.  Since the controller is
not aware of day2 workers, until now these CSRs needed to be approved
manually.
This change applies on to ZTP flow, since in cloud assisted-service
cannot connect a cluster to attempt CSR approval.
The change is done in the agent controller that starts attempting to
approve CSRs when a day2 host moves to done.
At this point, the corresponding agent changes the stage to rebooting
which starts the CSR approval cycle.  It monitores the nodes in the
spoke cluster to find if a corresponding ready node is in the spoke
cluster.  If not, it keeps attempting to approve these CSRs.
Once a corresponding ready node is discovered, the stage in the agent is
moved to done and the cycle of CSR approval is stopped.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @avishayt 

## Checklist

## Reviewers Checklist


filanov:
cc: @flaper87

rollandf:
Do we need to have a timeout for trying to approve the CSR?
Assisted Service will not always have access to the spoke cluster. It really depends on the DNS settings.

filanov:
> Do we need to have a timeout for trying to approve the CSR? Assisted Service will not always have access to the spoke cluster. It really depends on the DNS settings.

Good point, because it's something that can probably be fixed by the user maybe we should try forever. increasing the timeout can help in this case. for example try again after 10m or more. if we can identify the error we can increase the timeout even more. 
@mhrivnak what do you think?

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +207,113 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+func isCsrApproved(csr *certificatesv1.CertificateSigningRequest) bool {
+	for _, c := range csr.Status.Conditions {
+		if c.Type == certificatesv1.CertificateApproved {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *AgentReconciler) ApproveAllCsrs(clients ClusterKubeClient) error {
```
filanov, line 271: how does it apply to a specific node? it looks like you are approving all the hosts in the cluster no?

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +207,113 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+	if err != nil {
+		return false, false, err
+	}
+	shouldApprove, err := r.shouldApproveCSRs(agent, clients)
+	if err != nil {
+		return false, true, err
+	}
+
+	// Even if node is already ready, we try approving last time
+	err = r.ApproveAllCsrs(clients)
```
filanov, line 310: Don't you need to approve only csrs related to a specific node?

internal/controller/controllers/agent_controller_test.go
```diff
@@ -1188,6 +1189,250 @@ var _ = Describe("agent reconcile", func() {
...
+			},
+			Data: map[string][]byte{
+				"kubeconfig": []byte(BASIC_KUBECONFIG),
+			},
+		}
+		Expect(c.Create(ctx, adminKubeconfigSecret)).ShouldNot(HaveOccurred())
+	})
+
+	It("No matching node - No csrs", func() {
+		host := newAgent(hostId.String(), testNamespace, v1beta1.AgentSpec{ClusterDeploymentName: &v1beta1.ClusterReference{Name: "clusterDeployment", Namespace: testNamespace}})
```
filanov, line 1260: please put all the things that repeat under context and add it to before each

internal/controller/controllers/agent_controller.go
```diff
@@ -54,19 +55,21 @@ import (
 )
 
 const (
-	AgentFinalizerName = "agent." + aiv1beta1.Group + "/ai-deprovision"
+	AgentFinalizerName         = "agent." + aiv1beta1.Group + "/ai-deprovision"
+	approveCsrsRequeueDuration = time.Minute
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/controller/controllers/agent_controller.go
ori-amizur: Yes.  I did something similar to this https://github.com/openshift/assisted-installer/blob/e3e0586daac9b07b93f5ca2d220ac0c458f441b3/src/assisted_installer_controller/assisted_installer_controller.go#L368-L384
I will have to find out if the CSR can be associated with the specific agent.

internal/controller/controllers/agent_controller.go
ori-amizur: I will try to see if the CSRs can be associated with the specific agent.

internal/controller/controllers/agent_controller_test.go
ori-amizur: OK

cmd/main.go
```diff
@@ -537,14 +537,15 @@ func main() {
...
-				AuthType:         Options.Auth.AuthType,
+				Client:                   ctrlMgr.GetClient(),
+				APIReader:                ctrlMgr.GetAPIReader(),
+				Log:                      log,
+				Scheme:                   ctrlMgr.GetScheme(),
+				Installer:                bm,
+				CRDEventsHandler:         crdEventsHandler,
+				ServiceBaseURL:           Options.BMConfig.ServiceBaseURL,
+				AuthType:                 Options.Auth.AuthType,
+				ClusterKubeClientFactory: controllers.NewClusterKubeClientFactory(log),
```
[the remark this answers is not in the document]
ori-amizur: This part repeats in the referenced link
https://github.com/openshift/assisted-service/blob/5a661f1fedd3e3e36aaee7d988d86dcbd73a5ece/internal/controller/controllers/spoke_cluster.go#L11-L25

internal/controller/controllers/agent_controller.go
ori-amizur: Seems to me that the association cannot be recognized

filanov: @flaper87 @tsorya @eranco74 maybe one of you know how to identify the correct agent on the spoke cluster?

ori-amizur: It can be done if the CSR itself is opened and the node itself is part of the CN field

ori-amizur: Do we want to go for it?

flaper87:
> Day1 CSRs are approved by assisted controller. Since the controller is not aware of day2 workers, until now these CSRs needed to be approved manually.

Can you elaborate under which scenarios the CSRs need to be manually approved? So far, using BMAC and day2, the CSRs have been automatically approved when the BMH, Machine, and node information are properly linked. MAPI has a `machine-approver`, `nodelink` and other pieces that take care of doing this automatically.

* Did you have a chance to look into why the CSR was not being automatically approved?
* Is this because of the `None` platform? 
* Is there a way for us to integrate with the OCP workflows rather than approving them manually?

Holding this PR for now until we clear the above points
/hold

internal/controller/controllers/agent_controller.go
flaper87: I have commented in the mail thread but, to provide a more specific comment here, I would rather not approve the CSRs manually if we can let the `machine-approver` do it for us. Let's first understand what it would take for us to integrate with the regular workflow.

internal/controller/controllers/agent_controller.go
flaper87: > I will try to see if the CSRs can be associated with the specific agent.

There should be a way to just approve the CSRs related to a specific host, in case we really need to approve these manually.

filanov:
> * Is there a way for us to integrate with the OCP workflows rather than approving them manua

I'm not aware to all the details but as far as i understand CSR is not approved automatically in None platform. 
Some issue with DNS? @romfreiman @avishayt maybe you know the reason?

internal/controller/controllers/agent_controller.go
ori-amizur: @flaper87 do you think the similar code https://github.com/openshift/assisted-installer/blob/e3e0586daac9b07b93f5ca2d220ac0c458f441b3/src/assisted_installer_controller/assisted_installer_controller.go#L368-L384 in assisted-installer should be removed as well ?

cmd/main.go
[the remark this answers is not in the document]
filanov: Not sure i understand this comment, i know that there is a duplication don't we want to resolve it?

internal/controller/controllers/agent_controller.go
tsorya: If we have machine approver in None platform, i agree that better to create relevant machine yamls as we do on BM platform, it will allow to have almost same code for all the platforms

internal/controller/controllers/agent_controller.go
tsorya: one liner 
oc get csr csr-99n79 -o jsonpath='{ .spec.request }' | base64 -d |  openssl req  -in /dev/stdin -text  | grep  remote-worker1
        Subject: O=system:nodes, CN=system:node:remote-worker1

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +209,132 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+func (r *AgentReconciler) isAssociatedWithAgent(csr *certificatesv1.CertificateSigningRequest, agent *aiv1beta1.Agent) (bool, error) {
+	decodedCSR, _ := pem.Decode(csr.Spec.Request)
+	if decodedCSR == nil {
+		return false, errors.Errorf("Failed to decode request of CSR %s", csr.Name)
+	}
+	cr, err := x509.ParseCertificateRequest(decodedCSR.Bytes)
+	if err != nil {
+		return false, errors.Wrapf(err, "Failed to parse CSR %s", csr.Name)
+	}
+	return strings.EqualFold("system:node:"+agent.Spec.Hostname, cr.Subject.CommonName), nil
```
tsorya, line 282: this can be none not? if i didn't set it ?

eranco74:
> > * Is there a way for us to integrate with the OCP workflows rather than approving them manua
> 
> I'm not aware to all the details but as far as i understand CSR is not approved automatically in None platform. Some issue with DNS? @romfreiman @avishayt maybe you know the reason?

So this change is for ZTP installing none platform??
Is this a real scenario?

romfreiman:
@eranco74 why not? This is basically what @tsorya does right now with bf, no?

eranco74:
No, @tsorya isn't doing ZTP.
He is downloading the ISO manually and booting the hosts.
To my understanding in ZTP we have BMAC.

eranco74:
There is no reference in the code that the CSR approval is done only for None platform.
:confused:

ori-amizur:
> There is no reference in the code that the CSR approval is done only for None platform. confused

We decided (@filanov @avishayt and me) that approval will be applied for all, not only none platform

eranco74:
> > There is no reference in the code that the CSR approval is done only for None platform. confused
> 
> We decided (@filanov @avishayt and me) that approval will be applied for all, not only none platform
https://github.com/openshift/assisted-service/pull/3169#issuecomment-1008861585
According to this :point_up: this is specific logic for none.
In any case it seems that this PR is trying to manually approve CRs that should get approved by the machine-approver.
For none platform it might make some sense (Not sure about it) but for bearmetal platform we know that the machine-approver is doing exactly that (approve the CSR) so I don't understand why we need it.

flaper87 (changes requested):
Writing a more general comment here rather than many comments throughout the PR:

This function currently checks the nodes to see if the CSR needs to be approved.

```go
func (r *AgentReconciler) shouldApproveCSRs(agent *aiv1beta1.Agent, clients ClusterKubeClient) (bool, error) {

	nodes, err := clients.ListNodes()
	if err != nil {
		return false, err
	}
	return !isNodeApproved(agent, nodes.Items), nil
}
```

However, there's one CSR that needs to be approved *before* the node exists. This is the CSR that is used to create the Node's CR itself (a.k.a NodeCSR). Then a new CSR is created that also needs to be approved.

here's how the machine approver verifies whether this is a NodeCert or not: 

https://github.com/openshift/cluster-machine-approver/blob/e44b8bd5eaad0e4f2273e6901b5d6202d7c37b15/pkg/controller/node_client.go#L38-L52

And here's where the actual CSR validation happens: 

https://github.com/openshift/cluster-machine-approver/blob/e44b8bd5eaad0e4f2273e6901b5d6202d7c37b15/pkg/controller/csr_check.go#L56

I think, what we may actually want to happen here is: 

1. Only approve CSRs for None platform. As it is right now, it looks like it would approve all the CSRs that match an agent. There may be a case to extend this to the BIY flow.
2. Make sure we only approve CSRs that match an agent with an ongoing installation. Probably better to verify that we are only approving the first two CSRs.
3. Match the CSR content with the agent, and the inventory if needed (match IP addresses too).

Hope this helps :rocket:

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +209,132 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+}
+
+func (r *AgentReconciler) ApproveAllCsrs(clients ClusterKubeClient, agent *aiv1beta1.Agent) error {
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +209,132 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+	namespacedName := types.NamespacedName{
+		Namespace: agent.Spec.ClusterDeploymentName.Namespace,
+		Name:      fmt.Sprintf(adminKubeConfigStringTemplate, agent.Spec.ClusterDeploymentName.Name),
+	}
+
+	secret, err := getSecret(ctx, r.Client, r.APIReader, namespacedName)
+	if err != nil {
+		return false, false, err
+	}
+	clients, err := r.ClusterKubeClientFactory.Create(secret)
```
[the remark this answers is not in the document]
ori-amizur: It is not the same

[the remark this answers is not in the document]
ori-amizur: Besides specialized nodes api client, it has specialized CSR api client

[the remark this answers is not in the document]
tsorya: can't we use same for both? we can take advanced one

internal/controller/controllers/agent_controller.go
```diff
@@ -299,7 +482,18 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
...
 		if h.Progress != nil && h.Progress.CurrentStage != "" {
-			agent.Status.Progress.CurrentStage = h.Progress.CurrentStage
+			var isNoneHostRebooting bool
+
+			if isNoneHostRebooting, err = r.isDay2NonePlatformHostRebooting(ctx, agent, h); err != nil {
+				log.WithError(err).Errorf("Failed to find if agent %s/%s belongs to none platform cluster and is rebooting", agent.Namespace, agent.Name)
+				return ctrl.Result{RequeueAfter: defaultRequeueAfterOnError}, nil
+			}
+
+			if isNoneHostRebooting {
```
tsorya, line 492: what is difference between None and BM stage in rebooting?

tsorya: maybe better to have function like 
setCurrentStage and move all the logic there? it will be mostly same as isDay2NonePlatformHostRebooting ?

internal/controller/controllers/agent_controller.go
```diff
@@ -272,11 +404,62 @@ func (r *AgentReconciler) deregisterHostIfNeeded(ctx context.Context, log logrus
 	return buildReply(nil)
 }
 
+//
+//  In assisted installer, UserManagedNetworking implicates none platform.  This flag is part of AgentClusterInstall spec.
+//
+func (r *AgentReconciler) isNonePlatform(ctx context.Context, agent *aiv1beta1.Agent) (isNone bool, err error) {
```
flaper87, line 410: This same function is being implemented in BMAC. Can you extract it from the controllers and put it in a common place?

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +210,132 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+		if !isCsrApproved(csr) {
+			associated, err := r.isAssociatedWithAgent(csr, agent)
+			if err != nil || !associated {
+				if err != nil {
+					r.Log.WithError(err).Error("Find agent <-> CSR association")
+				}
+				continue
+			}
+			if err = clients.ApproveCsr(csr); err != nil {
+				r.Log.WithError(err).Error("Failed to approve CSR")
```
flaper87, line 303: can you add more info to this error? What agent? what node? what cluster?

internal/controller/controllers/agent_controller.go
flaper87: I think what @tsorya means is that the field could be empty if it's not set. Instead of using `agent.Spec` we should use the data in the Status. 

That said, the Status is set during the discovery phase, which may not be the right one. The way other parts of BMAC do it is to use the value from the Spec if set otherwise the one from the Status:


https://github.com/openshift/assisted-service/blob/1d327257c3cc54485c3cd09b517dc56907873967/internal/controller/controllers/bmh_agent_controller.go#L512-L517

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +210,132 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+func isCsrApproved(csr *certificatesv1.CertificateSigningRequest) bool {
+	for _, c := range csr.Status.Conditions {
+		if c.Type == certificatesv1.CertificateApproved {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *AgentReconciler) isAssociatedWithAgent(csr *certificatesv1.CertificateSigningRequest, agent *aiv1beta1.Agent) (bool, error) {
```
flaper87, line 274: I think we should do other checks on the CSR, as I've suggested elsewhere.

If possible, I think we should get as close as possible to the checks done by the cluster-machine-approver

https://github.com/openshift/cluster-machine-approver/blob/master/pkg/controller/csr_check.go


Maybe we can vendor `cluster-machine-approver` and reuse some of its internal functions? What do you think?

At a high level, we should at least check the IPs/SAN and make sure they match. Also, make sure that we were expecting a new CSR rather than just listing them all and checking them.

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +210,132 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+		for _, ip := range append(intf.IPV4Addresses, intf.IPV6Addresses...) {
+			splitIp := strings.Split(ip, "/")
+			if len(splitIp) == 2 {
+				addresses = append(addresses, splitIp[0])
+			}
+		}
+	}
+	for i := range nodes {
+		n := &nodes[i]
+		if strings.EqualFold(n.Name, agent.Spec.Hostname) {
```
flaper87, line 225: Some comment as below w.r.t the use of `agent.Spec.Hostname`

internal/controller/controllers/agent_controller_test.go
```diff
@@ -1243,6 +1245,259 @@ var _ = Describe("agent reconcile", func() {
 
 })
 
+var _ = Describe("Approve CSRs", func() {
+	pemCSR := `-----BEGIN CERTIFICATE REQUEST-----
```
flaper87, line 1249: Can you write a comment of what the decoded certificate should look like?

internal/controller/controllers/csr_utils.go
```diff
@@ -0,0 +1,80 @@
+package controllers
```
flaper87, line 1: How much of this code can we share/reuse from the cluster-machine-approver? 

https://github.com/openshift/cluster-machine-approver/tree/master/pkg

It looks to me that we may be able to vendor CMA and re-use its utilities

flaper87 (changes requested):

internal/controller/controllers/agent_controller.go
ori-amizur: 2 different PRs.  They are a bit different.  I will merge them afterwards

internal/controller/controllers/agent_controller.go
ori-amizur: OK

internal/controller/controllers/agent_controller.go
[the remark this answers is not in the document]
ori-amizur: What is advanced one ?

getSpokeClient returns general client.  All places that approve CSRs use specialized client for specific CRs.  We can move to specialized clients all over the code, but I am not sure I want to do it now.

internal/controller/controllers/agent_controller.go
ori-amizur: Referring to https://github.com/openshift/assisted-service/pull/3169#discussion_r796648762 - OK
SAN is optional in CSR.  The above file is all lower-cased so it us package private.  It is used just by the controller.  In addition it looks for machines and do some logic on it which is irrelevant for us. 
Can take all checks from the above file.  SAN is not checked there (or maybe I missed it).  Not sure it is required to be there?
We are expecting CSR because we trigger CSR approval only when host moves to Done and the agent is not Done yet which we consider rebooting

internal/controller/controllers/agent_controller.go
ori-amizur: OK

internal/controller/controllers/agent_controller_test.go
ori-amizur: OK

ori-amizur: Certificate or CSR ?

internal/controller/controllers/agent_controller.go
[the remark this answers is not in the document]
ori-amizur: OK. Will be done as part of another PR

internal/controller/controllers/agent_controller.go
```diff
@@ -333,14 +488,29 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 	} else {
 		setConditionsUnknown(agent)
 	}
+	if agent.Status.DebugInfo.State == models.HostStatusAddedToExistingCluster &&
+		agent.Status.Progress.CurrentStage == models.HostStageRebooting {
+		alreadyApproved, requeue, err := r.tryApproveDay2CSRs(ctx, agent)
```
tsorya, line 493: why not to put logic of re queuing inside  tryApproveDay2CSRs ? I mean why to return boolean if you can return relevant object and remove most of the logic from here?

ori-amizur: I preferred the tryApproveDay2CSRs just to manage the approval.  I didn't want it to handle the requeue and managing stage (rebooting/done) which is the flag if we want to continue the approval cycle. Only updateStatus should touch and modify the agent.

ori-amizur: In general I think that we should separate responsibilities: updateStatus updates the agent returns the return value to reconcile, and tryApproveDay2CSRs just tries to approve.

internal/controller/controllers/agent_controller.go
ori-amizur: OK

internal/controller/controllers/agent_controller.go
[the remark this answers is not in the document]
ori-amizur: https://issues.redhat.com/browse/MGMT-9192

tsorya:
looks good as for me, @flaper87 lets push it? 
@ori-amizur please open ticket for uniting noth spoke clients

ori-amizur:
> looks good as for me, @flaper87 lets push it? @ori-amizur please open ticket for uniting noth spoke clients

https://issues.redhat.com/browse/MGMT-9192

internal/controller/controllers/agent_controller.go
```diff
@@ -299,7 +444,18 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 		}
 
 		if h.Progress != nil && h.Progress.CurrentStage != "" {
-			agent.Status.Progress.CurrentStage = h.Progress.CurrentStage
+			var isNoneHostRebooting bool
+
+			if isNoneHostRebooting, err = r.isDay2NonePlatformHostRebooting(ctx, agent, h); err != nil {
```
flaper87, line 449: We can move the variable assignment outside the if statement so that we can reuse it in L492

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +209,95 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+	if err != nil {
+		r.Log.WithError(err).Errorf("Agent %s/%s: Failed to create spoke client", agent.Namespace, agent.Name)
+		return false
+	}
+	hostname := getAgentHostname(agent)
+	var (
+		validateNodeCsr       nodeCsrValidator
+		shouldApproveMoreCSRs bool
+	)
+	node, err := clients.GetNode(hostname)
```
flaper87, line 282: Let's add a `TODO` so that we can check the proper hostname for the IPv6 case. TL;DR: In IPv6, the node will have a full FQDN as name rather than the hostname

internal/controller/controllers/spoke_k8s_client.go
```diff
@@ -0,0 +1,84 @@
...
+	data := spokeK8sClient{
+		csrClient:   config.CertificatesV1().CertificateSigningRequests(),
+		nodesClient: config.CoreV1().Nodes(),
+		log:         cf.log,
+	}
+	return &data, nil
+}
+
+func (c *spokeK8sClient) ListCsrs() (*certificatesv1.CertificateSigningRequestList, error) {
+	return c.csrClient.List(context.TODO(), metav1.ListOptions{})
```
flaper87, line 63: Let's filter only for `Pending` CSRs. Normally, when CSRs are `Approved` they are garbage collected but it takes a bit.

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +209,95 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+
+func (r *AgentReconciler) approveAIHostsCSRs(clients SpokeK8sClient, agent *aiv1beta1.Agent, validateNodeCsr nodeCsrValidator) {
+	csrList, err := clients.ListCsrs()
+	if err != nil {
+		r.Log.WithError(err).Errorf("Failed to get CSRs for agent %s/%s", agent.Namespace, agent.Name)
+		return
+	}
+	for i := range csrList.Items {
+		csr := &csrList.Items[i]
+		if !isCsrApproved(csr) {
```
flaper87, line 241: If we filter for `Pending`, we don't need to check if it's `Approved`

internal/controller/controllers/csr_utils.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func createNodeServerCsrValidator(node *corev1.Node) nodeCsrValidator {
+	return func(agent *aiv1beta1.Agent, csr *certificatesv1.CertificateSigningRequest, x509cr *x509.CertificateRequest) (bool, error) {
+		return validateNodeServerCSR(agent, node, csr, x509cr)
+	}
+}
+
+func isCsrAssociatedWithAgent(x509CSR *x509.CertificateRequest, agent *aiv1beta1.Agent) bool {
+	return strings.EqualFold(nodeUserPrefix+getAgentHostname(agent), x509CSR.Subject.CommonName)
```
flaper87, line 202: TODO: Check if this will be affected by IPv6 full FQDN

internal/controller/controllers/csr_utils.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+	return false
+}
+
+func inTimeSpan(start, end, check time.Time) bool {
+	return check.After(start) && check.Before(end)
+}
+
+func isCsrInTimeSpan(agent *aiv1beta1.Agent, csr *certificatesv1.CertificateSigningRequest) bool {
```
flaper87, line 218: Can you write a small docstring of what this function does and used for, pls

internal/controller/controllers/agent_controller.go
```diff
@@ -204,6 +209,95 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+func (r *AgentReconciler) shouldApproveCSR(csr *certificatesv1.CertificateSigningRequest, agent *aiv1beta1.Agent, validateNodeCsr nodeCsrValidator) (bool, error) {
+	x509CSR, err := getX509ParsedRequest(csr)
+	if err != nil {
+		return false, err
+	}
+	if !isCsrAssociatedWithAgent(x509CSR, agent) {
+		return false, nil
+	}
+
+	if !isCsrInTimeSpan(agent, csr) {
```
flaper87, line 226: Let's ask Joel why the timeSpan check is relevant and document it. The motivation may be security-related and we should document it :D 

Otherwise, we can probably just remove this check

internal/controller/controllers/csr_utils.go
```diff
@@ -0,0 +1,228 @@
...
+		"system:authenticated",
+	)
+
+	return req.Spec.Username == nodeBootstrapperUsername && nodeBootstrapperGroups.Equal(sets.NewString(req.Spec.Groups...))
+}
+
+// Type representing a function that validates CSR for client or server.
+type nodeCsrValidator func(agent *aiv1beta1.Agent, csr *certificatesv1.CertificateSigningRequest, x509cr *x509.CertificateRequest) (bool, error)
+
+// Recognize if the CSR is a client CSR, that is sent to allow the node to join the cluster
```
flaper87, line 111: Could you please document the various checks this function does?

internal/controller/controllers/csr_utils.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	if !funk.ContainsString(x509CSR.Subject.Organization, nodeGroup) {
+		return false, errors.Errorf("CSR %s agent %s/%s: Organization %v doesn't include %s", csr.Name, agent.Namespace, agent.Name,
+			x509CSR.Subject.Organization, nodeGroup)
+	}
+	if x509CSR.Subject.CommonName != csr.Spec.Username {
+		return false, errors.Errorf("CSR %s agent %s/%s: Mismatched CommonName %s != %s for CSR %s", csr.Name, agent.Namespace, agent.Name,
+			x509CSR.Subject.CommonName, csr.Spec.Username, csr.Name)
+	}
+	agentDNSNames := append(getNodeDNSNames(node), getAgentHostname(agent))
```
flaper87, line 178: TODO: check if this is affected by the IPv6 fqdn.

We could remove the `getAgentHostname` from here since the matching should be done on the node itself, which exists in the cluster already.

internal/controller/controllers/csr_utils.go
```diff
@@ -0,0 +1,228 @@
...
+	for _, dnsName := range x509CSR.DNSNames {
+		if !funk.ContainsString(agentDNSNames, dnsName) {
+			return false, errors.Errorf("CSR %s agent %s/%s: DNS name %s missing from available DNS names %v", csr.Name, agent.Namespace, agent.Name,
+				dnsName, agentDNSNames)
+		}
+	}
+	agentIPs := getAgentIPs(agent)
```
flaper87, line 185: As in the previous comment, we should probably using the node's IP here as that's the info that should have been used for the CSR.

There's no agent in the spoke cluster, so probably we should not be matching the CSR with the agent itself

flaper87 (changes requested):
These comments were brought to you thanks to a pair review done between @ori-amizur @tsorya and @flaper87

internal/controller/controllers/agent_controller.go
```diff
@@ -272,11 +366,62 @@ func (r *AgentReconciler) deregisterHostIfNeeded(ctx context.Context, log logrus
...
+func (r *AgentReconciler) isDay2NonePlatformHostRebooting(ctx context.Context, agent *aiv1beta1.Agent, h *models.Host) (bool, error) {
+	if swag.StringValue(h.Status) == models.HostStatusAddedToExistingCluster &&
+		h.Progress.CurrentStage == models.HostStageDone {
+		switch agent.Status.Progress.CurrentStage {
+		case models.HostStageDone:
+			return false, nil
+		case models.HostStageRebooting:
+			return true, nil
+		default:
+			isNone, err := r.isNonePlatform(ctx, agent)
```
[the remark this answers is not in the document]
ori-amizur: Will remove the case for rebooting

internal/controller/controllers/agent_controller.go
ori-amizur: OK

internal/controller/controllers/spoke_k8s_client.go
ori-amizur: To find if CSR is pending, we need to filter the CSR to find if "Approved" condition does not exist.  @flaper87 do you know if it is possible and how to do it?

internal/controller/controllers/agent_controller.go
ori-amizur: Same as above

internal/controller/controllers/agent_controller.go
ori-amizur: He says its purpose is to avoid approving out of date CSRs. https://coreos.slack.com/archives/CUPJTHQ5P/p1644320238911699?thread_ts=1643629029.082179&cid=CUPJTHQ5P
Can be removed

internal/controller/controllers/csr_utils.go
ori-amizur: OK

internal/controller/controllers/agent_controller.go
ori-amizur: Removed

internal/controller/controllers/spoke_k8s_client.go
ori-amizur: Putting it on hold for now

internal/controller/controllers/agent_controller.go
ori-amizur: Putting it on hold

internal/controller/controllers/csr_utils.go
ori-amizur: OK

internal/controller/controllers/csr_utils.go
ori-amizur: OK

flaper87 (commented):
I think thisis good for a first iteration: 

/lgtm

flaper87:
/retest

@tsorya want to give it the final approval?

flaper87 (approved):
@tsorya is on PTO!

PR #3180, opened 2022-01-11
MGMT-8046: BMAC should not create BMHs on the spoke cluster if it sees that platform==none

ori-amizur:
# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @flaper87 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	bmh_v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
 	aiv1beta1 "github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/internal/ignition"
 	"github.com/openshift/assisted-service/models"
@@ -729,6 +730,24 @@ func (r *BMACReconciler) reconcileBMH(ctx context.Context, log logrus.FieldLogge
 	return reconcileComplete{dirty: true, stop: true}
 }
 
+//
+//  In assisted installer, UserManagedNetworking implicates none platform.  This flag is part of AgentClusterInstall spec.
+//
+func (r *BMACReconciler) isNonePlatform(ctx context.Context, cd *hivev1.ClusterDeployment) (isNone, propagateError bool, err error) {
+	if cd.Spec.ClusterInstallRef == nil {
+		return false, false, errors.Errorf("Cluster Install Reference is null for cluster deployment ns=%s name=%s", cd.Namespace, cd.Name)
+	}
+	clusterInstall := hiveext.AgentClusterInstall{}
+	namespacedName := types.NamespacedName{
+		Namespace: cd.Namespace,
+		Name:      cd.Spec.ClusterInstallRef.Name,
+	}
+	if err = r.Get(ctx, namespacedName, &clusterInstall); err != nil {
+		return false, !k8serrors.IsNotFound(err), errors.Wrapf(err, "Could not get AgentClusterInstall %s for ClusterDeployment %s", cd.Spec.ClusterInstallRef.Name, cd.Name)
+	}
+	return clusterInstall.Spec.Networking.UserManagedNetworking, false, nil
+}
+
 // Reconcile the `BareMetalHost` resource on the spoke cluster
 //
 // Baremetal-operator in the hub cluster creates a host using the live-iso feature. To add this host as a worker node
@@ -755,6 +774,21 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		Namespace: agent.Spec.ClusterDeploymentName.Namespace,
 		Name:      fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name),
 	}
+	isNonePlatform, propagateError, err := r.isNonePlatform(ctx, cd)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to determine if platform is none for cluster deployment %s/%s", cd.Namespace, cd.Name)
+		// In case ACI reference doesn't exist or ACI not found do not reconcile again since such a change in these will trigger BMH reconciliation
+		// In all other cases propagate the error
+		if propagateError {
+			return reconcileError{err}
+		}
+		return reconcileComplete{}
+	}
+	if isNonePlatform {
+		// If none platform do not attempt to create spoke BMH
+		return reconcileComplete{}
+	}
+
 	secret, err := getSecret(ctx, r.Client, r.APIReader, key)
 	if err != nil {
 		return reconcileError{err}
--- a/internal/controller/controllers/bmh_agent_controller_test.go
+++ b/internal/controller/controllers/bmh_agent_controller_test.go
@@ -10,6 +10,7 @@ import (
 	bmh_v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
 	"github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/models"
@@ -17,6 +18,7 @@ import (
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	machinev1beta1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -669,6 +671,7 @@ var _ = Describe("bmac reconcile", func() {
 		var host *bmh_v1alpha1.BareMetalHost
 		var agent *v1beta1.Agent
 		var cluster *hivev1.ClusterDeployment
+		var clusterInstall *hiveext.AgentClusterInstall
 		var adminKubeconfigSecret *corev1.Secret
 		var secretName string
 		var bmhName string
@@ -717,7 +720,8 @@ var _ = Describe("bmac reconcile", func() {
 			cluster = newClusterDeployment(clusterName, testNamespace, defaultClusterSpec)
 			cluster.Spec.Installed = true
 			Expect(c.Create(ctx, cluster)).ShouldNot(HaveOccurred())
-
+			clusterInstall = newAgentClusterInstall(cluster.Spec.ClusterInstallRef.Name, testNamespace, hiveext.AgentClusterInstallSpec{}, cluster)
+			Expect(c.Create(ctx, clusterInstall)).ShouldNot(HaveOccurred())
 			secretName = fmt.Sprintf(adminKubeConfigStringTemplate, clusterName)
 			adminKubeconfigSecret = &corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
@@ -833,6 +837,27 @@ var _ = Describe("bmac reconcile", func() {
 				err = spokeClient.Get(ctx, types.NamespacedName{Name: adminKubeconfigSecret.Name, Namespace: OPENSHIFT_MACHINE_API_NAMESPACE}, spokeSecret)
 				Expect(err).To(BeNil())
 				Expect(spokeSecret.Data).To(Equal(adminKubeconfigSecret.Data))
+			})
+			It("should not set spoke BMH - None platform", func() {
+				clusterInstall.Spec.Networking.UserManagedNetworking = true
+				Expect(c.Update(ctx, clusterInstall)).ToNot(HaveOccurred())
+				result, err := bmhr.Reconcile(ctx, newBMHRequest(host))
+				Expect(err).To(BeNil())
+				Expect(result).To(Equal(ctrl.Result{}))
+
+				updatedHost := &bmh_v1alpha1.BareMetalHost{}
+				err = c.Get(ctx, types.NamespacedName{Name: host.Name, Namespace: testNamespace}, updatedHost)
+				Expect(err).To(BeNil())
+
+				spokeBMH := &bmh_v1alpha1.BareMetalHost{}
+				spokeClient := bmhr.spokeClient
+				err = spokeClient.Get(ctx, types.NamespacedName{Name: host.Name, Namespace: OPENSHIFT_MACHINE_API_NAMESPACE}, spokeBMH)
+				Expect(err).ToNot(BeNil())
+				Expect(k8serrors.IsNotFound(err)).To(BeTrue())
+				spokeSecret := &corev1.Secret{}
+				err = spokeClient.Get(ctx, types.NamespacedName{Name: adminKubeconfigSecret.Name, Namespace: OPENSHIFT_MACHINE_API_NAMESPACE}, spokeSecret)
+				Expect(err).ToNot(BeNil())
+				Expect(k8serrors.IsNotFound(err)).To(BeTrue())
 			})
 
 			It("validate label on Secrets", func() {


internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -729,6 +730,22 @@ func (r *BMACReconciler) reconcileBMH(ctx context.Context, log logrus.FieldLogge
...
+	if cd.Spec.ClusterInstallRef == nil {
+		return false, false, errors.Errorf("Cluster Install Reference is null for cluster deployment ns=%s name=%s", cd.Namespace, cd.Name)
+	}
+	clusterInstall := hiveext.AgentClusterInstall{}
+	namespacedName := types.NamespacedName{
+		Namespace: cd.Namespace,
+		Name:      cd.Spec.ClusterInstallRef.Name,
+	}
+	if err = r.Get(ctx, namespacedName, &clusterInstall); err != nil {
+		return false, !k8serrors.IsNotFound(err),
```
filanov, line 743: if it's not found error what will re-trigger the reconcile? 
Is there some watch for cluster reference?

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -755,6 +772,16 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		Namespace: agent.Spec.ClusterDeploymentName.Namespace,
 		Name:      fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name),
 	}
+	isNonePlatform, err := r.isNonePlatform(ctx, cd)
+	if err != nil {
+		log.WithError(err).Error("Is none platform")
+		return reconcileError{err}
```
filanov, line 778: @flaper87 is that reconcile immediately? if so can it create a hot loop?

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -729,6 +730,22 @@ func (r *BMACReconciler) reconcileBMH(ctx context.Context, log logrus.FieldLogge
...
+		return false, errors.Errorf("Cluster Install Reference is null for cluster deployment ns=%s name=%s", cd.Namespace, cd.Name)
+	}
+	clusterInstall := hiveext.AgentClusterInstall{}
+	namespacedName := types.NamespacedName{
+		Namespace: cd.Namespace,
+		Name:      cd.Spec.ClusterInstallRef.Name,
+	}
+	if err = r.Get(ctx, namespacedName, &clusterInstall); err != nil {
+		return false,
+			errors.Wrapf(err, "AgentClusterInstall %s does not exist for ClusterDeployment %s", cd.Spec.ClusterInstallRef.Name, cd.Name)
```
flaper87, line 744: What do you think about propagating `err` as is rather than wrapping it under that message? I fear that this code may be saying that the ACI doesn't exist when the error could be something else. Another option would be to rephrase the error: 

```suggestion
			errors.Wrapf(err, "Could not get AgentClusterInstall %s for ClusterDeployment %s", cd.Spec.ClusterInstallRef.Name, cd.Name)
```

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -729,6 +730,22 @@ func (r *BMACReconciler) reconcileBMH(ctx context.Context, log logrus.FieldLogge
 	return reconcileComplete{dirty: true, stop: true}
 }
 
+func (r *BMACReconciler) isNonePlatform(ctx context.Context, cd *hivev1.ClusterDeployment) (isNone bool, err error) {
```
flaper87, line 733: please, document this function.

It may not be obvious to future readers that `UserNetworkManagedNetworking` means None. For example, some may wonder why not checking the [ControlPlaneAgents](https://github.com/openshift/assisted-service/blob/f140a9ed60f1507e2cdeb71092e486918d50b6a6/api/hiveextension/v1beta1/agentclusterinstall_types.go#L284)  field instead to check if it's an SNO cluster. The answer is that we want to be able to deploy HA clusters that are platform None.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -755,6 +772,16 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		Namespace: agent.Spec.ClusterDeploymentName.Namespace,
 		Name:      fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name),
 	}
+	isNonePlatform, err := r.isNonePlatform(ctx, cd)
+	if err != nil {
+		log.WithError(err).Error("Is none platform")
```
flaper87, line 777: Please, could you use a more descriptive error?

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -729,6 +730,22 @@ func (r *BMACReconciler) reconcileBMH(ctx context.Context, log logrus.FieldLogge
 	return reconcileComplete{dirty: true, stop: true}
 }
 
+func (r *BMACReconciler) isNonePlatform(ctx context.Context, cd *hivev1.ClusterDeployment) (isNone bool, err error) {
+	if cd.Spec.ClusterInstallRef == nil {
+		return false, errors.Errorf("Cluster Install Reference is null for cluster deployment ns=%s name=%s", cd.Namespace, cd.Name)
```
flaper87, line 735: Does this have to be an error? If it's an error, we may end in an infinite reconcile loop. We could also not return an error, call `reconcileComplete` and let a new reconcile happen if there are changes to the Clusterdeployment. [BMAC has a watch on ClusterDeployments](https://github.com/openshift/assisted-service/blob/master/internal/controller/controllers/bmh_agent_controller.go#L1402-L1404)

internal/controller/controllers/bmh_agent_controller.go
flaper87: Yeah, the controller will likely reconcile again as soon as possible. If we want to re-queue, then I think using `reconcileRequeue` may be better. We could rework this so that we react to changes on ACI by adding a watch, but that doesn't save us from errors and it may hide important issues.

internal/controller/controllers/bmh_agent_controller_test.go
```diff
@@ -834,6 +838,32 @@ var _ = Describe("bmac reconcile", func() {
...
+				clusterInstall.Spec.Networking.UserManagedNetworking = true
+				Expect(c.Update(ctx, clusterInstall)).ToNot(HaveOccurred())
+				result, err := bmhr.Reconcile(ctx, newBMHRequest(host))
+				Expect(err).To(BeNil())
+				Expect(result).To(Equal(ctrl.Result{}))
+
+				updatedHost := &bmh_v1alpha1.BareMetalHost{}
+				err = c.Get(ctx, types.NamespacedName{Name: host.Name, Namespace: testNamespace}, updatedHost)
+				Expect(err).To(BeNil())
+				Expect(updatedHost.ObjectMeta.Annotations).To(HaveKey(BMH_HARDWARE_DETAILS_ANNOTATION))
```
flaper87, line 851:
```suggestion				
```

internal/controller/controllers/bmh_agent_controller_test.go
```diff
@@ -834,6 +838,32 @@ var _ = Describe("bmac reconcile", func() {
...
+				Expect(updatedHost.ObjectMeta.Annotations).ToNot(HaveKey(BMH_DETACHED_ANNOTATION))
+				Expect(updatedHost.ObjectMeta.Annotations).To(HaveKey(BMH_AGENT_IGNITION_CONFIG_OVERRIDES))
+				Expect(updatedHost.ObjectMeta.Annotations[BMH_AGENT_IGNITION_CONFIG_OVERRIDES]).NotTo(Equal(""))
+				Expect(updatedHost.ObjectMeta.Annotations[BMH_AGENT_IGNITION_CONFIG_OVERRIDES]).To(ContainSubstring("dGVzdA=="))
```
flaper87, line 855:
```suggestion
```

flaper87 (changes requested):
Great work, @ori-amizur 

Wrote a few comments :pray:

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: OK. Will rephrase

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: In this case we will have to return something to indicate that we do not want to reconcile again

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: OK

internal/controller/controllers/bmh_agent_controller_test.go
ori-amizur: What's the difference ?

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: It should be handled by reconcileError wrapper

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: OK

flaper87 (commented):
I think we can move forward with this PR in the light of the latest findings (we don't need Machines for none platform and CSRs should be approved by the user/automated flow).

@ori-amizur can you address the comments so that we can approve it?

ori-amizur:
> I think we can move forward with this PR in the light of the latest findings (we don't need Machines for none platform and CSRs should be approved by the user/automated flow).
> 
> @ori-amizur can you address the comments so that we can approve it?

@flaper87 addressed them.  Anything else ?

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -755,6 +776,20 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		Namespace: agent.Spec.ClusterDeploymentName.Namespace,
 		Name:      fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name),
 	}
+	isNonePlatform, requeue, err := r.isNonePlatform(ctx, cd)
+	if err != nil {
+		log.WithError(err).Error("Failed to check if platform is none")
+		if requeue {
+			return reconcileRequeue{requeueAfter: DEFAUAT_BMH_REQUEUE_DURATION}
```
flaper87, line 783: I'm trying to think about this a bit more and I don't think we need to requeue. I know we briefly discussed this in the previous version of the PR so, please, don't hate me :hear_no_evil: 

If the ACI doesn't exist (which it should if we consider this is day2), I think it's fine for us to just error out. Using requeue is not recommended unless necessary. 

I think it's fine to log the error if ACI doesn't exist and return `reconcileComplete` This is because we have a watch on `ClusterDeployment` which would trigger a new loop if something changes on that front.

internal/controller/controllers/bmh_agent_controller_test.go
flaper87: it's just not needed for this test. We can skip these `Expect` lines

flaper87 (changes requested):
Thanks for addressing the comments, one more note on the way we are checking things. Otherwise, LGTM

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: If ACI doesn't exist requeue is false so we do not requeue.  Requeue is done if there is an error on Get

internal/controller/controllers/bmh_agent_controller_test.go
ori-amizur: OK

internal/controller/controllers/bmh_agent_controller.go
flaper87: > If ACI doesn't exist requeue is false so we do not requeue. Requeue is done if there is an error on Get

This is exactly what I mean. The get could also fail because the ACI was specified in the CD but it doesn't exist.

So, if we fail to get ACI for whatever reason, I think we should log the error but skip the requeue. We have the watches in place which should trigger a reconcile if needed.

ori-amizur: OK

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -729,6 +731,25 @@ func (r *BMACReconciler) reconcileBMH(ctx context.Context, log logrus.FieldLogge
 	return reconcileComplete{dirty: true, stop: true}
 }
 
+//
+//  In assisted installer, UserManagedNetworking implicates none platform.  This flag is part of AgentClusterInstall spec.
+//
+func (r *BMACReconciler) isNonePlatform(ctx context.Context, cd *hivev1.ClusterDeployment) (isNone, requeue bool, err error) {
+	if cd.Spec.ClusterInstallRef == nil {
+		return false, false, errors.Errorf("Cluster Install Reference is null for cluster deployment ns=%s name=%s", cd.Namespace, cd.Name)
```
flaper87, line 739: I understand that technically this error and the one in L748 (the Get error) are not the same but, logically (for the flow), they mean pretty much the same. We don't have an ACI that we can use to validate if this is None platform or not.

Regardless of None platform, we always need an ACI when deploying a day2 cluster using the KubeAPI flow so, if we hit this case, then there's a bigger issue and we should not proceed with the current reconciliation.

Unless I'm missing something, of course :)

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: So what is wrong now ?

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: OK.  We will always return complete (although it is a change of flow for BM but it is unexpected)

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -755,6 +775,16 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		Namespace: agent.Spec.ClusterDeploymentName.Namespace,
 		Name:      fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name),
 	}
+	isNonePlatform, err := r.isNonePlatform(ctx, cd)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to determine if platform is none for cluster deployment %s/%s", cd.Namespace, cd.Name)
+	}
+	if isNonePlatform || err != nil {
+		// If none platform do not attempt to create spoke BMH
+		// If error occurred no retry will be attempted since we will get same result
```
flaper87, line 784: There will be a retry if the `ClusterDeployment` changes since BMAC has a watch on this CRD.

Could you mention this? (just asking this because we need another push to remove `DEFAUAT_BMH_REQUEUE_DURATION` )

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -77,6 +78,7 @@ const (
 	MACHINE_TYPE                        = "machine.openshift.io/cluster-api-machine-type"
 	MCS_CERT_NAME                       = "ca.crt"
 	OPENSHIFT_MACHINE_API_NAMESPACE     = "openshift-machine-api"
+	DEFAUAT_BMH_REQUEUE_DURATION        = 10 * time.Second
```
flaper87, line 81:
```suggestion
```

flaper87 (changes requested):
tiny nit, otherwise lgtm

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -755,6 +775,16 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
...
 		Name:      fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name),
 	}
+	isNonePlatform, err := r.isNonePlatform(ctx, cd)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to determine if platform is none for cluster deployment %s/%s", cd.Namespace, cd.Name)
+	}
+	if isNonePlatform || err != nil {
+		// If none platform do not attempt to create spoke BMH
+		// If error occurred no retry will be attempted since we will get same result
+		return reconcileComplete{}
```
flaper87, line 785: We chatted about this so I want to bring here what we discussed: 

* This logic should ignore not found errors and mark the reconcile loop as complete
* The rest of the errors can be propagated (return reconcileError{})
* When errors are returned, the reconcile [loop will retry again](https://cluster-api.sigs.k8s.io/developer/providers/implementers-guide/controllers_and_reconciliation.html#reconciliation)

flaper87 (approved):
thanks for working on this, the investigation, and your patience :rocket:

PR #3182, opened 2022-01-11
MGMT-8582: Leader elector to use leases

slaviered:
A leader is currently locked over ConfigMap resource. This kind of
locking mechanism is deprecated in favor of using Leases

Replace the current lock with LeaseLock and clear the current
ConfigMap that is currently being used for the locking mechanism
    
On rolling upgrade, the existing leader will lose the leadership
and one of the newly upgraded instances will take the lead

# Assisted Pull Request

- [ ] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/deploy/olm-catalog/manifests/assisted-service-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/manifests/assisted-service-operator.clusterserviceversion.yaml
@@ -664,18 +664,22 @@ spec:
           - create
         - apiGroups:
           - ""
+          - coordination.k8s.io
           resources:
           - configmaps
+          - leases
           verbs:
           - create
         - apiGroups:
           - ""
+          - coordination.k8s.io
           resourceNames:
           - assisted-service-leader-election-helper
           - assisted-service-migration-helper
           - assisted-service-baseiso-helper
           resources:
           - configmaps
+          - leases
           verbs:
           - get
           - update
--- a/pkg/leader/leaderelector.go
+++ b/pkg/leader/leaderelector.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	corev1 "k8s.io/api/core/v1"
+	coordv1 "k8s.io/api/coordination/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
@@ -51,20 +51,24 @@ func (f *DummyElector) RunWithLeader(ctx context.Context, run func() error) erro
 var _ ElectorInterface = &Elector{}
 
 type Elector struct {
-	log           logrus.FieldLogger
-	config        Config
-	kube          *kubernetes.Clientset
-	isLeader      bool
-	configMapName string
+	log      logrus.FieldLogger
+	config   Config
+	kube     *kubernetes.Clientset
+	lockName string
+	isLeader bool
 }
 
-func NewElector(kubeClient *kubernetes.Clientset, config Config, configMapName string, logger logrus.FieldLogger) *Elector {
-	logger = logger.WithField("configMap", configMapName)
-	return &Elector{log: logger, config: config, kube: kubeClient, configMapName: configMapName, isLeader: false}
+func NewElector(kubeClient *kubernetes.Clientset, config Config, lockName string, logger logrus.FieldLogger) *Elector {
+	logger = logger.WithField("lock", lockName)
+	return &Elector{log: logger, config: config, kube: kubeClient, lockName: lockName, isLeader: false}
 }
 
 func (l *Elector) IsLeader() bool {
 	return l.isLeader
+}
+
+func (l *Elector) setLeader(status bool) {
+	l.isLeader = status
 }
 
 // Wait for leader, run given function, drop leader and exit.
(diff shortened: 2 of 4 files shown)


pkg/leader/leaderelector.go
```diff
@@ -51,22 +52,32 @@ func (f *DummyElector) RunWithLeader(ctx context.Context, run func() error) erro
...
-	logger = logger.WithField("configMap", configMapName)
-	return &Elector{log: logger, config: config, kube: kubeClient, configMapName: configMapName, isLeader: false}
+func NewElector(kubeClient *kubernetes.Clientset, config Config, lockName string, logger logrus.FieldLogger) *Elector {
+	logger = logger.WithField("lock", lockName)
+	mu := &sync.RWMutex{}
+	return &Elector{log: logger, config: config, kube: kubeClient, lockName: lockName, isLeader: false, mu: mu}
 }
 
 func (l *Elector) IsLeader() bool {
+	l.mu.RLock()
```
filanov, line 70: why do you need a mutex? we have only one instance of each thread running locally

slaviered: But the query for IsLeader can come from different threads (like monitor) while the value is set from the elector thread. is it not?

filanov: Why wasn't it a problem until now?

tsorya: to say the the truth i don't understand why we need this mutex. What's the problem of getting boolean value by multiple threads?

pkg/leader/leaderelector.go
```diff
@@ -51,22 +52,32 @@ func (f *DummyElector) RunWithLeader(ctx context.Context, run func() error) erro
...
+	defer l.mu.RUnlock()
 	return l.isLeader
 }
 
+func (l *Elector) setLeader(status bool) {
+	l.mu.Lock()
```
tsorya, line 76: same question , why we need lock? setting boolean is thread safe so why it is needed?

filanov: I agree, lock maybe solve some races on a local pod but you will still have a race between different pods, and in wort case one of them will fail to run. so it's really depend on what you are trying to solve. 
IMO the lock doesn't avoid any issues, in worst case you will get the right reply in the next iteration.

subsystem/leader_test.go
```diff
@@ -227,15 +227,15 @@ var _ = Describe("Leader tests", func() {
 		log.Infof("Verify that previous leader was not changed")
 		waitForPredicate(timeout, test1.isLeader)
 	})
-	It("Deleting configmap in a loop", func() {
-		By("Deleting configmap in a loop (it must be recreated all the time), leader will loose leader and retake it")
+	It("Deleting lock underlying resource in a loop", func() {
```
flaper87, line 230: Should we also add a test for cleaning deprecated resources? If anything, just to test that we are not cleaning more than we should :)

pkg/leader/leaderelector.go
```diff
@@ -134,19 +156,26 @@ func (l *Elector) StartLeaderElection(ctx context.Context) error {
...
-	cm := &corev1.ConfigMap{
+func (l *Elector) clean() {
+	err := l.kube.CoreV1().ConfigMaps(l.config.Namespace).Delete(context.Background(),
+		l.lockName, metav1.DeleteOptions{})
+	if err != nil && !k8serrors.IsNotFound(err) {
+		l.log.WithError(err).Errorf("Failed to delete config map for leader lock %s", l.lockName)
+	}
+}
+
+func (l *Elector) createResource() error {
```
flaper87, line 167:
```suggestion
func (l *Elector) createLease() error {
```

subsystem/leader_test.go
```diff
@@ -227,15 +227,15 @@ var _ = Describe("Leader tests", func() {
...
+		By("Deleting leases in a loop (it must be recreated all the time), leader will loose leader and retake it")
 		leader1 := leader.NewElector(client, cf, configMapName, log)
```
flaper87, line 232:
```suggestion
		leader1 := leader.NewElector(client, cf, lockName, log)
```

flaper87 (changes requested):
Few small comments but, in general, I agree with @tsorya and @filanov with regards to introducing a mutex in this PR.

tsorya:
It looks good but if i remember right you need to update rbac on staging, integration and prod? At least previously it should be done manually or everything will be broken for qe

slaviered:
/hold

waiting for AppSRE configuration change that allow the RBAC changes to go on staging

deploy/olm-catalog/manifests/assisted-service-operator.clusterserviceversion.yaml
```diff
@@ -664,18 +664,22 @@ spec:
           - create
         - apiGroups:
           - ""
+          - coordination.k8s.io
```
filanov, line 667: aren't those generated? https://coreos.slack.com/archives/G01A5NB3S6M/p1642489754024700

slaviered: this change is the product of running generate-all. But following the discussion I am adding also the operator changes

slaviered:
- Permission to add leases from app-sre added and merged
- Manual tests:
   - Test Infra deployment - Success
   - OLM deployment - Success

filanov (approved):

filanov (approved):

PR #3199, opened 2022-01-13
MGMT-8894: Simplify IPv6 Network Manager configuration

mkowalski:
This commit simplifies confguration of the Network Manager in case IPv6
is used. Till now we were

* configuring consistent DHCPv6 client identification
* creating system-connections-merged overlay configuration
* manually configuring keyfile path

We are able to simplify this flow for the following reasons

* with the introduction of InfraEnvs, the static network configuration
  is now a property of InfraEnv and not of a cluster
* RHCOS handles creation of systemConnectionMerged whenever needed and
  if so, it handles configuration of the keyfile paths

Until now, the configuration applied by Assisted Service was not causing
problems as in RHCOS <=4.9 the keyfile paths were overriden whenever a
different overlay directory has been created.

Starting from RHCOS 4.10 this is no longer the case and network
configurations do not need to be stored in any merged directory. With
the removal of the custom keyfile path, we are leveraging the following
properties

* default path is /etc/NetworkManager/system-connections
* aforementioned path contains all the configurations coming from the
  custom NMStateConfig

Closes: [MGMT-8894](https://issues.redhat.com/browse/MGMT-8894)
Closes: Bug-2030289
Relates-to: Bug-2040195

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @yevgeny-shnaidman 
/cc @flaper87 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -59,12 +59,20 @@ ipv6.dhcp-duid=ll
 `
 
 // Configuration to be used by MCO manifest to get consistent IPv6 DHCP client identification.
-const Ipv6DuidRuntimeConf = `
+const Ipv6DuidRuntimeConfPre410 = `
 [connection]
 ipv6.dhcp-iaid=mac
 ipv6.dhcp-duid=ll
 [keyfile]
 path=/etc/NetworkManager/system-connections-merged
+`
+
+// Configuration of consistent IPv6 DHCP without the keyfile path set. This is because starting in
+// RHCOS 4.10 it is breaking the Network Manager configuration.
+const Ipv6DuidRuntimeConf = `
+[connection]
+ipv6.dhcp-iaid=mac
+ipv6.dhcp-duid=ll
 `
 
 // configuration of NM to disable handling of /etc/resolv.conf
--- a/internal/ignition/ignition.go
+++ b/internal/ignition/ignition.go
@@ -58,24 +58,6 @@ const (
 	workerIgn = "worker.ign"
 )
 
-const SystemConnectionsMerged = `
-[Unit]
-After=systemd-tmpfiles-setup.service
-[Mount]
-Where=/etc/NetworkManager/system-connections-merged
-What=overlay
-Type=overlay
-Options=lowerdir=/etc/NetworkManager/system-connections,upperdir=/run/nm-system-connections,workdir=/run/nm-system-connections-work
-[Install]
-WantedBy=multi-user.target
-`
-
-const kniTempFile = `
-D /run/nm-system-connections 0755 root root - -
-D /run/nm-system-connections-work 0755 root root - -
-d /etc/NetworkManager/system-connections-merged 0755 root root -
-`
-
 const agentMessageOfTheDay = `
 **  **  **  **  **  **  **  **  **  **  **  **  **  **  **  **  **  ** **  **  **  **  **  **  **
 This is a host being installed by the OpenShift Assisted Installer.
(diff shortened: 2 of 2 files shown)


mkowalski:
/retest
`[Fail] Cluster [BeforeEach] update cluster name exceed max length (54 characters)`, not relevant to this PR

internal/ignition/ignition.go
```diff
@@ -875,40 +857,29 @@ func (g *installerGenerator) updateDhcpFiles() error {
...
-		// add overlay configuration for NM in case of 4.7 and None platform.
-		// TODO - remove once this configuration is integrated in MCO for None platform (Bugzilla 1928473)
-		setFileInIgnition(config, "/etc/tmpfiles.d/kni.conf", fmt.Sprintf("data:,%s", url.PathEscape(kniTempFile)), false, 420)
-		setUnitInIgnition(config, SystemConnectionsMerged, "etc-NetworkManager-system\\x2dconnections\\x2dmerged.mount", true)
+	var v6config string
+	if is410Version {
+		v6config = common.Ipv6DuidRuntimeConf
+	} else {
+		v6config = common.Ipv6DuidRuntimeConfPre410
 	}
```
filanov, line 881:
```suggestion
	v6config := common.Ipv6DuidRuntimeConfPre410
	if is410Version {
		v6config = common.Ipv6DuidRuntimeConf
	} 
```

filanov (commented):
is `StaticNetworkConfigured` still relevant or can it be removed?

romfreiman:
Is it backwards compatible?

internal/common/common.go
```diff
@@ -59,14 +59,22 @@ ipv6.dhcp-duid=ll
 `
 
 // Configuration to be used by MCO manifest to get consistent IPv6 DHCP client identification.
-const Ipv6DuidRuntimeConf = `
+const Ipv6DuidRuntimeConfPre410 = `
```
yevgeny-shnaidman, line 62: I think that we should rename it Ipv6DuidRuntimeConfPre49 ( if i am not mistaken we don't need to set keyfile Path in 4.9 also

internal/ignition/ignition.go
```diff
@@ -1107,15 +1072,6 @@ func setFileInIgnition(config *config_latest_types.Config, filePath string, file
 	config.Storage.Files = append(config.Storage.Files, file)
 }
 
-func setUnitInIgnition(config *config_latest_types.Config, contents, name string, enabled bool) {
```
yevgeny-shnaidman, line 1110: actually , this is a usefull function. We may not need it now, but maybe later we will need to add some systemd unit to ignition. What do you think about moving this function and setFileInIgnition to a different file and keep them there?

internal/common/common.go
mkowalski: You are correct, but on the other hand (just for the safety) I wanted to keep the current workflow in 4.9 because it works "as is". Indeed we have discovered that using keyfile path in 4.9 is not needed, but if we remove it with this patch, it needs that we need to do additional testing and potentially break stuff that already works (I know that we personally believe it will not break, but there may be always some corner case somewhere out there)

internal/ignition/ignition.go
mkowalski: Not sure I understand. I can't leave this function because linter complains that we have an unused function in the codebase. Any idea how to overcome it? I totally agree it may become useful in the future, but I'm not sure how to embed it into the current flow

mkowalski:
>Is it backwards compatible?

@romfreiman

* it **is not compatible** with API v1; this patch must not be backported to any version that does not use API v2
* it **is compatible** with pre-4.10 because it only changes configuration files for 4.10 and later

internal/ignition/ignition.go
mkowalski: Right, obvious one. Fixed

internal/ignition/ignition.go
yevgeny-shnaidman: we can probably use something like:
//lint:ignore U1000 Ignore unused function 
but it is too much bother, lets remove it

mkowalski: Good, that was fairly easy. I'll keep the function

internal/common/common.go
yevgeny-shnaidman: ok, sure

mkowalski:
/test subsystem-kubeapi-aws
`[Fail] [kube-api]cluster installation [It] deploy CD with ACI and agents with ignitionEndpoint`, I can't see a connection with this PR

mkowalski:
/hold

This may or may be not a valid error. Slack thread - https://coreos.slack.com/archives/CUPJTHQ5P/p1642442514149900

PR #3219, opened 2022-01-20
MGMT-8783: support multiple patch versions of os/release images

danielerez:
Added support to maintain OS and release images with cpu_architecture in major.minor.patch format. This functionality is a means to allow multiple z-stream versions concurrently, instead of the previous behaviour which limited a single z-stream for every y-stream.

The following changes were made in versions package in order to introduce the new behaviour:
* GetReleaseImgae:
  * Returns an image based on the specified x.y.z.
  * If missing, fallback to x.y (as previously).
* GetOsImage:
  * Returns an image based on the specified x.y.z.
  * If missing, fallback to x.y (as previously).
  * If x.y is not available, try to find the latest applicable image for the specified x.y.z. This is done to avoid forcing a direct version match between OS and release image. E.g. for kube-api custom z-stream release versions support.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @filanov 
/cc @osherdp 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -3,6 +3,7 @@ package versions
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/go-openapi/runtime/middleware"
@@ -37,7 +38,7 @@ type Handler interface {
 	GetDefaultReleaseImage(cpuArchitecture string) (*models.ReleaseImage, error)
 	GetOsImage(openshiftVersion, cpuArchitecture string) (*models.OsImage, error)
 	GetLatestOsImage(cpuArchitecture string) (*models.OsImage, error)
-	GetCPUArchitectures(openshiftVersion string) ([]string, error)
+	GetCPUArchitectures(openshiftVersion string) []string
 	GetOpenshiftVersions() []string
 	AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion, cpuArchitecture string) (*models.ReleaseImage, error)
 }
@@ -98,30 +99,32 @@ func (h *handler) ListSupportedOpenshiftVersions(ctx context.Context, params ope
 
 func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params operations.V2ListSupportedOpenshiftVersionsParams) middleware.Responder {
 	openshiftVersions := models.OpenshiftVersions{}
-	availableOcpVersions := h.GetOpenshiftVersions()
-	for _, key := range availableOcpVersions {
-		// Fetch available architectures for openshift version
-		availableArchitectures, err := h.GetCPUArchitectures(key)
-		if err != nil {
-			return operations.NewListSupportedOpenshiftVersionsInternalServerError().
-				WithPayload(common.GenerateInternalFromError(err))
-		}
-		// Fetch release image for openshift version (values should be similar across architectures)
-		releaseImage, err := h.GetReleaseImage(key, common.DefaultCPUArchitecture)
-		if err != nil {
-			return operations.NewListSupportedOpenshiftVersionsInternalServerError().
-				WithPayload(common.GenerateInternalFromError(err))
+	for _, releaseImage := range h.releaseImages {
+		key := *releaseImage.Version
+		if swag.StringValue(releaseImage.CPUArchitecture) == "" {
+			// Empty implies default architecture
+			*releaseImage.CPUArchitecture = common.DefaultCPUArchitecture
 		}
 
-		openshiftVersion := models.OpenshiftVersion{
-			CPUArchitectures: availableArchitectures,
-			Default:          releaseImage.Default,
-			DisplayName:      *releaseImage.Version,
-			SupportLevel:     h.getSupportLevel(*releaseImage),
+		openshiftVersion, exists := openshiftVersions[key]
+		if !exists {
+			openshiftVersion = models.OpenshiftVersion{
+				CPUArchitectures: []string{*releaseImage.CPUArchitecture},
+				Default:          releaseImage.Default,
+				DisplayName:      key,
+				SupportLevel:     h.getSupportLevel(*releaseImage),
+			}
+			openshiftVersions[key] = openshiftVersion
+		} else {
+			openshiftVersion.CPUArchitectures = append(
+				openshiftVersion.CPUArchitectures, *releaseImage.CPUArchitecture)
+			if *releaseImage.CPUArchitecture == common.DefaultCPUArchitecture {
+				// Default flag is specified on x86 image
+				openshiftVersion.Default = releaseImage.Default
+			}
+			openshiftVersions[key] = openshiftVersion
 		}
-		openshiftVersions[key] = openshiftVersion
 	}
-
 	return operations.NewV2ListSupportedOpenshiftVersionsOK().WithPayload(openshiftVersions)
 }
 
@@ -129,9 +132,6 @@ func (h *handler) GetMustGatherImages(openshiftVersion, cpuArchitecture, pullSec
 	versionKey, err := h.getKey(openshiftVersion)
 	if err != nil {
 		return nil, err
-	}
-	if !funk.Contains(h.GetOpenshiftVersions(), versionKey) {
-		return nil, errors.Errorf("No operators must-gather version for unsupported openshift version %s", versionKey)
 	}
 	if h.mustGatherVersions == nil {
 		h.mustGatherVersions = make(MustGatherVersions)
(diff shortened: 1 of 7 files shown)


internal/versions/versions.go
```diff
@@ -175,42 +171,88 @@ func (h *handler) GetDefaultReleaseImage(cpuArchitecture string) (*models.Releas
...
-	for _, osImage := range h.osImages {
+	// Find OS images of relevant CPU architecture
+	osImages := funk.Filter(h.osImages, func(osImage *models.OsImage) bool {
+		// Empty implies default CPU architecture
+		if swag.StringValue(osImage.CPUArchitecture) == "" {
+			return cpuArchitecture == "" || cpuArchitecture == common.DefaultCPUArchitecture
+		}
 		if cpuArchitecture == "" {
-			// Empty implies default CPU architecture
 			cpuArchitecture = common.DefaultCPUArchitecture
```
carbonin, line 181: Can this be done before the `funk.Filter` call?
It seems strange to edit this variable inside the filter function.

carbonin: I think this would also simplify the return in the previous block.

internal/versions/versions.go
```diff
@@ -175,42 +171,88 @@ func (h *handler) GetDefaultReleaseImage(cpuArchitecture string) (*models.Releas
...
-func (h *handler) GetReleaseImage(openshiftVersion, cpuArchitecture string) (*models.ReleaseImage, error) {
 	versionKey, err := h.getKey(openshiftVersion)
 	if err != nil {
 		return nil, err
 	}
 
-	for _, releaseImage := range h.releaseImages {
-		if cpuArchitecture == "" {
+	if osImage == nil {
+		// Fallback to x.y version
```
carbonin, line 200: This isn't necessarily x.y, right? The first pass tries to match whatever is passed in `openshiftVersion`. This tries to match the key for the release image for ocp version `openshiftVersion` right?

internal/versions/versions.go
```diff
@@ -175,42 +171,88 @@ func (h *handler) GetDefaultReleaseImage(cpuArchitecture string) (*models.Releas
...
+	if osImage != nil {
+		return osImage.(*models.OsImage), nil
+	}
+
+	return nil, errors.Errorf("The requested OS image for version (%s) isn't specified in OS images list", openshiftVersion)
+}
+
+// Returns the ReleaseImage entity
+func (h *handler) GetReleaseImage(openshiftVersion, cpuArchitecture string) (*models.ReleaseImage, error) {
+	// Find OS images of relevant CPU architecture
```
carbonin, line 226: We're not looking for an OS image here, right? Not sure I understand this comment.

internal/versions/versions.go
```diff
@@ -175,42 +171,88 @@ func (h *handler) GetDefaultReleaseImage(cpuArchitecture string) (*models.Releas
...
+	releaseImages := funk.Filter(h.releaseImages, func(releaseImage *models.ReleaseImage) bool {
+		if swag.StringValue(releaseImage.CPUArchitecture) == "" {
 			// Empty implies default CPU architecture
-			cpuArchitecture = common.DefaultCPUArchitecture
+			return cpuArchitecture == "" || cpuArchitecture == common.DefaultCPUArchitecture
```
carbonin, line 230: Same as before, I think this would be more clear if we set `cpuArchitecture` to the default outside of the filter function.

internal/versions/versions.go
```diff
@@ -175,42 +171,88 @@ func (h *handler) GetDefaultReleaseImage(cpuArchitecture string) (*models.Releas
...
-			// Empty implies default CPU architecture
 			cpuArchitecture = common.DefaultCPUArchitecture
 		}
-		if *osImage.OpenshiftVersion == versionKey && *osImage.CPUArchitecture == cpuArchitecture {
-			return osImage, nil
-		}
+		return swag.StringValue(osImage.CPUArchitecture) == cpuArchitecture
+	})
+	if funk.IsEmpty(osImages) {
+		return nil, errors.Errorf("The requested CPU architecture (%s) isn't specified in OS images list", cpuArchitecture)
```
carbonin, line 186: Is this particularly important to report?
Could we just check both requirements in the same `Find` instead of filtering first?

internal/versions/versions_test.go
```diff
@@ -232,13 +258,24 @@ var _ = Describe("list versions", func() {
 			Expect(*osImage.CPUArchitecture).Should(Equal(common.DefaultCPUArchitecture))
 		})
 
+		It("fetch OS image by major.minor", func() {
+			osImage, err = h.GetOsImage("4.9", common.DefaultCPUArchitecture)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(*osImage.OpenshiftVersion).Should(Equal("4.9"))
+		})
+
+		It("fetch missing major.minor.patch - find a patch version by major.minor", func() {
```
carbonin, line 267: Should we have another test that asks for `4.10` and returns `4.10.1`?

carbonin (commented):
After a first pass the logic in the versions package looks very complicated.
I'm not sure how much it can be simplified so let's just make sure we have tests covering all of these different possibilities, defaults, and fallbacks.

internal/versions/versions.go
```diff
@@ -101,16 +101,12 @@ func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params o
...
-				WithPayload(common.GenerateInternalFromError(err))
-		}
+		availableArchitectures := h.GetCPUArchitectures(key)
 		// Fetch release image for openshift version (values should be similar across architectures)
 		releaseImage, err := h.GetReleaseImage(key, common.DefaultCPUArchitecture)
 		if err != nil {
-			return operations.NewListSupportedOpenshiftVersionsInternalServerError().
-				WithPayload(common.GenerateInternalFromError(err))
+			// Missing release image - skip version
+			continue
```
osherdp, line 109: If I understand the case here, maybe we'll be better to release a warning log here?
If the list of release images is missing this image

osherdp:
/test e2e-metal-assisted-onprem
(no worries, not a blocker)

internal/versions/versions.go
osherdp: @danielerez can you elaborate on the case here?

danielerez: So this check handles a scenario in which an OS image x.y.z version is specified, but a correlated release image is missing. Refactored this func to go over the release images instead, so it no longer relevant.

internal/versions/versions.go
danielerez: Right, will simplfy.

internal/versions/versions.go
danielerez: Well I've separated the architecture filtering just to avoid checking it again later in the func. I.e. if a specific x.y.z isn't found, we fallback to x.y, etc. So seems simpler to check the architecture only once. We could still avoid the early return here if it would be clearer :/

internal/versions/versions.go
danielerez: So it's actually just matching against x.y version of OS image. I.e. we cover three options here:
1. x.y.z is available
2. only x.y is available
3. if both 1 and 2 are not available - find latest x.y.z

internal/versions/versions.go
danielerez: oops, menat release images here

internal/versions/versions.go
danielerez: yeah, will handle it the same

danielerez:
> After a first pass the logic in the versions package looks very complicated. I'm not sure how much it can be simplified so let's just make sure we have tests covering all of these different possibilities, defaults, and fallbacks.

Yeah sorry for that, didn't tried to avoid a full refactor when introducing this change. But added some more tests and simplified some of the logic. Hope it'll be clearer.

internal/versions/versions.go
carbonin: Ah because `getKey` is really `inMajorMinorFormat`. Got it.

internal/versions/versions.go
```diff
@@ -175,42 +175,103 @@ func (h *handler) GetDefaultReleaseImage(cpuArchitecture string) (*models.Releas
...
+		// Find latest available patch version by x.y version
+		osImages := funk.Filter(osImages, func(osImage *models.OsImage) bool {
+			imageVersionKey, err := h.getKey(*osImage.OpenshiftVersion)
+			if err != nil {
+				return false
+			}
+			return imageVersionKey == versionKey
+		}).([]*models.OsImage)
+		sort.Slice(osImages, func(i, j int) bool {
+			return *osImages[i].OpenshiftVersion > *osImages[j].OpenshiftVersion
```
carbonin, line 220: Does this direct comparison work?
Wouldn't `4.10` end up sorting the opposite of expected compared to `4.9` when using straight string or numerical sort?

carbonin: Can you make sure we have a test for this with `4.9.1`, `4.9.9`, and `4.9.10` versions?

internal/versions/versions.go
```diff
@@ -175,42 +175,103 @@ func (h *handler) GetDefaultReleaseImage(cpuArchitecture string) (*models.Releas
...
+		cpuArchitecture = common.DefaultCPUArchitecture
 	}
-
-	for _, releaseImage := range h.releaseImages {
-		if cpuArchitecture == "" {
-			// Empty implies default CPU architecture
-			cpuArchitecture = common.DefaultCPUArchitecture
+	// Filter Release images by specified CPU architecture
+	releaseImages := funk.Filter(h.releaseImages, func(releaseImage *models.ReleaseImage) bool {
+		if swag.StringValue(releaseImage.CPUArchitecture) == "" {
```
carbonin, line 244: Do we really expect that we would configure the service with a release image with no arch? I don't think this was the case before.

internal/versions/versions.go
```diff
@@ -175,42 +175,103 @@ func (h *handler) GetDefaultReleaseImage(cpuArchitecture string) (*models.Releas
 
 // Returns the OsImage entity
 func (h *handler) GetOsImage(openshiftVersion, cpuArchitecture string) (*models.OsImage, error) {
+	if cpuArchitecture == "" {
+		// Empty implies default CPU architecture
+		cpuArchitecture = common.DefaultCPUArchitecture
+	}
+	// Filter OS images by specified CPU architecture
+	osImages := funk.Filter(h.osImages, func(osImage *models.OsImage) bool {
+		if swag.StringValue(osImage.CPUArchitecture) == "" {
```
carbonin, line 184: This is strange, I think it's safe to assume that we've configured the service correctly, right?

internal/versions/versions.go
```diff
@@ -249,64 +310,51 @@ func (h *handler) AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion
...
+	osImage, err := h.GetOsImage(ocpReleaseVersion, cpuArchitecture)
 	if err != nil || osImage.URL == nil {
-		return nil, errors.Errorf("No OS images are available for version: %s", ocpVersionKey)
+		return nil, errors.Errorf("No OS images are available for version: %s", ocpReleaseVersion)
 	}
 
 	// Fetch ReleaseImage if exists
-	releaseImage, err := h.GetReleaseImage(ocpVersionKey, cpuArchitecture)
-	if err != nil {
+	releaseImage := funk.Find(h.releaseImages, func(releaseImage *models.ReleaseImage) bool {
```
carbonin, line 320: This is probably worth a comment.
Unless you've spent some time looking at this code it's not clear why we can't continue to use `GetReleaseImage` here.

(It's because we don't want the fallback logic in this case, right?)

internal/versions/versions.go
```diff
@@ -249,64 +310,51 @@ func (h *handler) AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion
...
+	if releaseImage == nil {
 		// Create a new ReleaseImage
 		releaseImage = &models.ReleaseImage{
-			OpenshiftVersion: &ocpVersionKey,
+			OpenshiftVersion: &ocpReleaseVersion,
 			CPUArchitecture:  &cpuArchitecture,
+			URL:              &releaseImageUrl,
+			Version:          &ocpReleaseVersion,
 		}
-	} else {
```
carbonin, line 272: This previously allowed for "updating" was that ever useful?
This is just in-memory, right? Do we ever alter the release images we support outside of startup?

If we do allow this do change, do we still want to be able to update the X.Y key here or will that just always get added as X.Y.Z now? Is it important that that behavior changed?

internal/versions/versions.go
```diff
@@ -249,64 +310,51 @@ func (h *handler) AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion
...
-				// Empty or missing property implies default CPU architecture
-				defaultArch := common.DefaultCPUArchitecture
-				osImage.CPUArchitecture = &defaultArch
-			}
-			if !funk.Contains(cpuArchitectures, *osImage.CPUArchitecture) {
-				cpuArchitectures = append(cpuArchitectures, *osImage.CPUArchitecture)
-			}
+func (h *handler) GetCPUArchitectures(openshiftVersion string) []string {
+	// Find OS images of specified openshiftVersion
+	osImages := funk.Filter(h.osImages, func(osImage *models.OsImage) bool {
```
carbonin, line 344: In this case, I don't think two loops through the list helps make this easier to read.
Maybe that made sense in the earlier function where you would have to check in multiple places, but I think the previous version of this function was fine.

internal/versions/versions.go
carbonin: That's fine, this can wait for a proper refactoring.

internal/versions/versions.go
carbonin: Nah this is fine. Things look separated out pretty clearly now.

carbonin (approved):

osherdp:
/retest 
(seems like jenkins doesn't retest by its own for some reason)

osherdp:
@danielerez I think this arm subsystem test is not going to pass any time soon
Can you check that out?
/hold

filanov:
@danielerez  the test failed on 
```
"conditions": [
          {
            "type": "SpecSynced",
            "status": "False",
            "lastProbeTime": "2022-01-26T19:35:23Z",
            "lastTransitionTime": "2022-01-26T19:35:23Z",
            "reason": "InputError",
            "message": "The Spec could not be synced due to an input error: Requested CPU architecture arm64 is not available"
          },
```

danielerez:
> Details



> @danielerez I think this arm subsystem test is not going to pass any time soon Can you check that out? /hold

Yeah fixed the issue, passed in the last run.
/unhold

osherdp (approved):

PR #3222, opened 2022-01-21
MGMT-7896: Add host's log's URL to Agent CR's status.

CrystalChun:
[MGMT-7896](https://issues.redhat.com/browse/MGMT-7896): Added the host's log's URL to the Agent CR's debugInfo when it becomes available so that the user may view and download that host's logs.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Manually tested by deploying nodes using the assisted-test-infra repository. Then once the status was updated with the logs  URL, curling it locally to see if the url is correct and contains the correct logs.

## Assignees

/cc @rollandf 
/cc @carbonin 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -317,6 +317,14 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 				return ctrl.Result{Requeue: true}, nil
 			}
 		}
+		if agent.Status.DebugInfo.LogsURL == "" && !time.Time(h.LogsCollectedAt).Equal(time.Time{}) { // logs collection time is updated means logs are available
+			logsURL, err := generateControllerLogsDownloadURL(r.ServiceBaseURL, clusterId.String(), r.AuthType, agent.Name, "host")
+			if err != nil {
+				log.WithError(err).Error("failed to generate controller logs URL")
+				return ctrl.Result{}, err
+			}
+			agent.Status.DebugInfo.LogsURL = logsURL
+		}
 		connected(agent, status)
 		requirementsMet(agent, status)
 		validated(agent, status, h)
@@ -346,6 +354,31 @@ func (r *AgentReconciler) populateEventsURL(log logrus.FieldLogger, agent *aiv1b
 		agent.Status.DebugInfo.EventsURL = eventUrl
 	}
 	return nil
+}
+
+func generateControllerLogsDownloadURL(baseURL string, clusterID string, authType auth.AuthType, host string, logsType string) (string, error) {
+	hostID := strfmt.UUID(host)
+	builder := &installer.V2DownloadClusterLogsURL{
+		ClusterID: strfmt.UUID(clusterID),
+		HostID:    &hostID,
+		LogsType:  &logsType,
+	}
+	u, err := builder.Build()
+	if err != nil {
+		return "", err
+	}
+
+	downloadURL := fmt.Sprintf("%s%s", baseURL, u.RequestURI())
+	if authType != auth.TypeLocal {
+		return downloadURL, nil
+	}
+
+	downloadURL, err = gencrypto.SignURL(downloadURL, clusterID, gencrypto.ClusterKey)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to sign %s controller logs URL for host %s", logsType, host)
+	}
+
+	return downloadURL, nil
 }
 
 func setConditionsUnknown(agent *aiv1beta1.Agent) {
--- a/internal/controller/controllers/agent_controller_test.go
+++ b/internal/controller/controllers/agent_controller_test.go
@@ -748,6 +748,61 @@ var _ = Describe("agent reconcile", func() {
 		Expect(agent.Status.DebugInfo.EventsURL).To(HavePrefix(expectedEventUrlPrefix))
 	})
 
+	It("validate Logs URL", func() {
+		serviceBaseURL := "http://example.com"
+		hr.ServiceBaseURL = serviceBaseURL
+		hostId := strfmt.UUID(uuid.New().String())
+		expectedLogsUrlPrefix := fmt.Sprintf("%s/api/assisted-install/v2/clusters/%s/logs", serviceBaseURL, sId.String())
+		commonHost := &common.Host{
+			Host: models.Host{
+				ID:         &hostId,
+				ClusterID:  &sId,
+				Status:     swag.String(models.HostStatusKnown),
+				StatusInfo: swag.String("Some status info"),
+			},
+		}
+		backEndCluster = &common.Cluster{Cluster: models.Cluster{
+			ID: &sId,
+			Hosts: []*models.Host{
+				&commonHost.Host,
+			}}}
+		host := newAgent(hostId.String(), testNamespace, v1beta1.AgentSpec{ClusterDeploymentName: &v1beta1.ClusterReference{Name: "clusterDeployment", Namespace: testNamespace}})
+		By("before installation")
+		clusterDeployment := newClusterDeployment("clusterDeployment", testNamespace, getDefaultClusterDeploymentSpec("clusterDeployment-test", "test-cluster-aci", "pull-secret"))
+		Expect(c.Create(ctx, clusterDeployment)).To(Succeed())
+		mockInstallerInternal.EXPECT().GetHostByKubeKey(gomock.Any()).Return(commonHost, nil).AnyTimes()
+		mockInstallerInternal.EXPECT().GetClusterByKubeKey(gomock.Any()).Return(backEndCluster, nil).Times(1)
+		Expect(c.Create(ctx, host)).To(Succeed())
+		request := newHostRequest(host)
+		result, err := hr.Reconcile(ctx, request)
+		Expect(err).To(BeNil())
+		Expect(result).To(Equal(ctrl.Result{}))
+		agent := &v1beta1.Agent{}
+		key := types.NamespacedName{
+			Namespace: testNamespace,
+			Name:      hostId.String(),
+		}
+		Expect(c.Get(ctx, key, agent)).To(Succeed())
+		Expect(agent.Status.DebugInfo.LogsURL).To(Equal(""))
+
+		By("during installation")
+		backEndCluster.Hosts[0].Status = swag.String(models.HostStatusInstalling)
+		backEndCluster.Hosts[0].LogsCollectedAt = strfmt.DateTime(time.Now())
+		mockInstallerInternal.EXPECT().GetClusterByKubeKey(gomock.Any()).Return(backEndCluster, nil)
+		_, err = hr.Reconcile(ctx, request)
+		Expect(err).To(BeNil())
+		Expect(c.Get(ctx, key, agent)).To(Succeed())
+		Expect(agent.Status.DebugInfo.LogsURL).To(HavePrefix(expectedLogsUrlPrefix))
+
+		By("after installation")
+		backEndCluster.Hosts[0].Status = swag.String(models.HostStatusInstalled)
+		mockInstallerInternal.EXPECT().GetClusterByKubeKey(gomock.Any()).Return(backEndCluster, nil)
+		_, err = hr.Reconcile(ctx, request)
+		Expect(err).To(BeNil())
+		Expect(c.Get(ctx, key, agent)).To(Succeed())
+		Expect(agent.Status.DebugInfo.LogsURL).To(HavePrefix(expectedLogsUrlPrefix))
+	})
+
 	It("Agent update ignition override valid cases", func() {
 		hostId := strfmt.UUID(uuid.New().String())
 		commonHost := &common.Host{
--- a/api/v1beta1/agent_types.go
+++ b/api/v1beta1/agent_types.go
@@ -242,6 +242,9 @@ type DebugInfo struct {
 	// EventsURL specifies an HTTP/S URL that contains events which occured during the cluster installation process
 	// +optional
 	EventsURL string `json:"eventsURL,omitempty"`
+	// LogsURL specifies a url for download controller logs tar file.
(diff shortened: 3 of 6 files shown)


internal/controller/controllers/agent_controller.go
```diff
@@ -348,6 +363,35 @@ func (r *AgentReconciler) populateEventsURL(log logrus.FieldLogger, agent *aiv1b
 	return nil
 }
 
+func generateControllerLogsDownloadURL(baseURL string, clusterID string, authType auth.AuthType, filterBy ...string) (string, error) {
```
flaper87, line 366: I imagine some of the code here came from: https://github.com/openshift/assisted-service/blob/69bb1f6dae8c21aa31f00af04d3d2b1c0f1084ac/internal/controller/controllers/common.go#L266-L286


Do you think we can extract the common pieces into a function they both call?


Also, maybe it's worth moving this function into the `common.go` package where the `generateEventsURL` function is.

internal/controller/controllers/agent_controller.go
```diff
@@ -348,6 +363,35 @@ func (r *AgentReconciler) populateEventsURL(log logrus.FieldLogger, agent *aiv1b
...
+	if err != nil {
+		return "", errors.Wrap(err, "failed to sign cluster controller logs URL")
+	}
+
+	return downloadURL, nil
+}
+
+func areLogsCollected(logCollectionTime strfmt.DateTime) bool {
+	return !time.Time(logCollectionTime).Equal(time.Time{}) // timestamp update, meaning logs were collected from a controller
+}
```
flaper87, line 393: We may not need this function, what do you think? We are consuming it from one place only and It's only doing this one thing, which we can likely do inline.

flaper87 (commented):
Great stuff, simple and to the point. Just commented with a few suggestions.

Also, it'd be great if you can add some tests :pray:

internal/controller/controllers/agent_controller.go
```diff
@@ -317,6 +319,19 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 				return ctrl.Result{Requeue: true}, nil
 			}
 		}
+		if agent.Status.DebugInfo.LogsURL == "" {
+			logsCollected := areLogsCollected(h.LogsCollectedAt)
+			if logsCollected {
```
carbonin, line 324: This can all be one line:
`if agent.Status.DebugInfo.LogsURL == "" && areLogsCollected(h.LogsCollectedAt)`

Even with @flaper87 's suggestion (removing the `areLogsCollected` function), I think this would be fine as one line and a comment.

internal/controller/controllers/agent_controller.go
```diff
@@ -348,6 +363,35 @@ func (r *AgentReconciler) populateEventsURL(log logrus.FieldLogger, agent *aiv1b
...
+	u, err := url.Parse(path)
+	if err != nil {
+		return "", err
+	}
+	queryParams := u.Query()
+	for i := 0; i < len(filterBy)-1; i += 2 {
+		queryParams.Set(filterBy[i], filterBy[i+1])
+	}
+	u.RawQuery = queryParams.Encode()
+	downloadURL := u.String()
```
carbonin, line 378: One option is to try to extract this into a separate function to share across events and logs, but I think this could be simplified quite a bit by using the generated url builder functions we already have access to along with simplifying parameters to this function (we don't support arbitrary filtering on logs only by host and type, so just pass those).

Then this can be simplified to something like:

```
import "github.com/openshift/assisted-service/restapi/operations/installer"
...
func generateControllerLogsDownloadURL(baseURL string, clusterID string, authType auth.AuthType, hostID *strfmt.UUID, logsType *string) (string, error) {
  builder := &installer.V2DownloadClusterLogsURL{
    ClusterID: strfmt.UUID(clusterID),
    HostID: hostID,
    LogsType: logsType,
  }

  u, err := builder.Build()
  if err != nil {
    return "", err
  }

  downloadURL := fmt.Sprintf("%s%s", baseURL, u.RequestURI())

  if authType != auth.TypeLocal {
    return downloadURL, nil
  }

  downloadURL, err = gencrypto.SignURL(downloadURL, clusterID, gencrypto.ClusterKey)
  if err != nil {
    return "", errors.Wrap(err, "failed to sign cluster controller logs URL")
  }

  return downloadURL, nil
}
```

This gets rid of all the repeated indexing and parsing logic.

internal/controller/controllers/agent_controller.go
CrystalChun: That's a great idea! I discussed with Nick and we think the best approach might be to create a separate PR to refactor this function to contain the common pieces to limit the scope of this PR to just adding the cluster logs. What do you think?

internal/controller/controllers/agent_controller.go
CrystalChun: Agreed! I'll go ahead and move it to the single line

internal/controller/controllers/agent_controller.go
CrystalChun: Good idea, I've modified the code to reflect this. Thanks!

internal/controller/controllers/agent_controller.go
CrystalChun: Thanks for letting me know about the URL builder! I've modified the code to use this, how does it look?

internal/controller/controllers/agent_controller.go
filanov: If it's a global function then it belongs to the same common file

internal/controller/controllers/agent_controller.go
```diff
@@ -348,6 +358,31 @@ func (r *AgentReconciler) populateEventsURL(log logrus.FieldLogger, agent *aiv1b
...
+	}
+
+	downloadURL := fmt.Sprintf("%s%s", baseURL, u.RequestURI())
+	if authType != auth.TypeLocal {
+		return downloadURL, nil
+	}
+
+	downloadURL, err = gencrypto.SignURL(downloadURL, clusterID, gencrypto.ClusterKey)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to sign cluster controller logs URL")
```
filanov, line 380:
```suggestion
		return "", errors.Wrapf(err, "failed to sign %s controller logs URL", logsType)
```

carbonin (commented):
I like the way this came out.
Just needs some unit tests

internal/controller/controllers/agent_controller.go
CrystalChun: It's not being used as a global function yet, but if it will be later on then is it best to move it to the common file in this PR?

internal/controller/controllers/agent_controller.go
CrystalChun: Added the change! Thank you for the suggestion :)

internal/controller/controllers/agent_controller_test.go
```diff
@@ -748,6 +748,66 @@ var _ = Describe("agent reconcile", func() {
 		Expect(agent.Status.DebugInfo.EventsURL).To(HavePrefix(expectedEventUrlPrefix))
 	})
 
+	It("validate Logs URL", func() {
+		serviceBaseURL := "http://acme.com"
```
carbonin, line 752: Super minor nit, but I usually prefer using `example.com` as it's a domain specifically reserved for being used in test and example code. ref: https://www.iana.org/domains/reserved

internal/controller/controllers/agent_controller_test.go
```diff
@@ -748,6 +748,66 @@ var _ = Describe("agent reconcile", func() {
...
+		backEndCluster = &common.Cluster{Cluster: models.Cluster{
+			ID: &sId,
+			Hosts: []*models.Host{
+				&commonHost.Host,
+			}}}
+		host := newAgent(hostId.String(), testNamespace, v1beta1.AgentSpec{ClusterDeploymentName: &v1beta1.ClusterReference{Name: "clusterDeployment", Namespace: testNamespace}})
+		By("before installation")
+
+		clusterDeployment := newClusterDeployment("clusterDeployment", testNamespace, getDefaultClusterDeploymentSpec("clusterDeployment-test", "test-cluster-aci", "pull-secret"))
+		Expect(c.Create(ctx, clusterDeployment)).To(BeNil())
```
carbonin, line 773: I prefer `To(Succeed())` anywhere the error value is the only thing returned from the call in `Expect`.
I feel like it reads better.

You have this in a few different places so I'll only put the comment once.

internal/controller/controllers/agent_controller_test.go
```diff
@@ -748,6 +748,66 @@ var _ = Describe("agent reconcile", func() {
...
+
+		By("during installation")
+		backEndCluster.Hosts[0].Status = swag.String(models.HostStatusInstalling)
+		backEndCluster.Hosts[0].LogsCollectedAt = strfmt.DateTime(time.Now())
+		mockInstallerInternal.EXPECT().GetClusterByKubeKey(gomock.Any()).Return(backEndCluster, nil)
+		_, err = hr.Reconcile(ctx, request)
+		fmt.Fprintf(GinkgoWriter, "agent logs url: %s\nhost id: %s\nevents url: %s", agent.Status.DebugInfo.LogsURL, hostId.String(), agent.Status.DebugInfo.EventsURL)
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(c.Get(ctx, key, agent)).ShouldNot(HaveOccurred())
+		Expect(agent.Status.DebugInfo.LogsURL).NotTo(BeEmpty())
```
carbonin, line 798: I don't think you need this expectation if you have the one on the following line.

internal/controller/controllers/agent_controller_test.go
```diff
@@ -748,6 +748,66 @@ var _ = Describe("agent reconcile", func() {
...
+		Expect(agent.Status.DebugInfo.LogsURL).To(HavePrefix(expectedLogsUrlPrefix))
+
+		By("after installation")
+		backEndCluster.Hosts[0].Status = swag.String(models.HostStatusInstalled)
+		mockInstallerInternal.EXPECT().GetClusterByKubeKey(gomock.Any()).Return(backEndCluster, nil)
+		_, err = hr.Reconcile(ctx, request)
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(c.Get(ctx, key, agent)).ShouldNot(HaveOccurred())
+		Expect(agent.Status.DebugInfo.LogsURL).NotTo(BeEmpty())
```
carbonin, line 807: Don't need this extra assertion.

internal/controller/controllers/agent_controller_test.go
CrystalChun: Modified it, thank you!

internal/controller/controllers/agent_controller_test.go
CrystalChun: Changed all occurrences to use this instead, thank you!

internal/controller/controllers/agent_controller_test.go
CrystalChun: Gotcha, removed it. Thanks Nick!

internal/controller/controllers/agent_controller_test.go
CrystalChun: Gotcha, removed it. Thanks Nick!

internal/controller/controllers/agent_controller_test.go
```diff
@@ -748,6 +748,61 @@ var _ = Describe("agent reconcile", func() {
...
+		host := newAgent(hostId.String(), testNamespace, v1beta1.AgentSpec{ClusterDeploymentName: &v1beta1.ClusterReference{Name: "clusterDeployment", Namespace: testNamespace}})
+		By("before installation")
+		clusterDeployment := newClusterDeployment("clusterDeployment", testNamespace, getDefaultClusterDeploymentSpec("clusterDeployment-test", "test-cluster-aci", "pull-secret"))
+		Expect(c.Create(ctx, clusterDeployment)).To(Succeed())
+		mockInstallerInternal.EXPECT().GetHostByKubeKey(gomock.Any()).Return(commonHost, nil).AnyTimes()
+		mockInstallerInternal.EXPECT().GetClusterByKubeKey(gomock.Any()).Return(backEndCluster, nil).Times(1)
+		Expect(c.Create(ctx, host)).To(Succeed())
+		request := newHostRequest(host)
+		result, err := hr.Reconcile(ctx, request)
+		Expect(err).To(Succeed())
```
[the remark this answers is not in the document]
CrystalChun: Gotcha, that makes sense! Thanks for clarifying. I'll go ahead and modify the statements.

internal/controller/controllers/agent_controller.go
```diff
@@ -317,6 +317,16 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 				return ctrl.Result{Requeue: true}, nil
 			}
 		}
+		if agent.Status.DebugInfo.LogsURL == "" && !time.Time(h.LogsCollectedAt).Equal(time.Time{}) { // logs collection time is updated means logs are available
+			logsURL, err := generateControllerLogsDownloadURL(r.ServiceBaseURL, clusterId.String(), r.AuthType, agent.Name, "host")
+			if err != nil {
+				log.WithError(err).Error("failed to generate controller logs URL")
+				return ctrl.Result{}, err
+			}
+			if logsURL != "" {
```
carbonin, line 326: Just noticed this. Will this ever return and empty string and no error?
Feels like this shouldn't be necessary.

CrystalChun: True! I've removed it. Thank you!

carbonin (approved):

carbonin (approved):

PR #3237, opened 2022-01-25
Bug 2034686: Expose a way for users to disable TLS verification in image service through the agent service config

CrystalChun:
[Bug 2034686](https://bugzilla.redhat.com/show_bug.cgi?id=2034686): Expose a way for users to disable TLS verification in image service through the agent service config. 

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] Manual (Elaborate on how it was tested): Created an OCP cluster, deployed OLM bundle, created AgentServiceConfig, and described the image service pod to see the environment variable appeared.
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @flaper87 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -74,7 +74,8 @@ const (
 	defaultIngressCertCMName      string = "default-ingress-cert"
 	defaultIngressCertCMNamespace string = "openshift-config-managed"
 
-	configmapAnnotation = "unsupported.agent-install.openshift.io/assisted-service-configmap"
+	configmapAnnotation                 = "unsupported.agent-install.openshift.io/assisted-service-configmap"
+	imageServiceSkipVerifyTLSAnnotation = "unsupported.agent-install.openshift.io/assisted-image-service-skip-verify-tls"
 
 	assistedConfigHashAnnotation = "agent-install.openshift.io/config-hash"
 	mirrorConfigHashAnnotation   = "agent-install.openshift.io/mirror-hash"
@@ -803,6 +804,11 @@ func (r *AgentServiceConfigReconciler) newAssistedCM(ctx context.Context, log lo
 }
 
 func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
+	skipVerifyTLS, ok := instance.ObjectMeta.GetAnnotations()[imageServiceSkipVerifyTLSAnnotation]
+	if !ok {
+		skipVerifyTLS = "false"
+	}
+
 	deploymentLabels := map[string]string{
 		"app": imageServiceName,
 	}
@@ -825,6 +831,7 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
 			{Name: "ASSISTED_SERVICE_SCHEME", Value: "https"},
 			{Name: "ASSISTED_SERVICE_HOST", Value: serviceName + "." + r.Namespace + ".svc:" + servicePort.String()},
 			{Name: "REQUEST_AUTH_TYPE", Value: "param"},
+			{Name: "INSECURE_SKIP_VERIFY", Value: skipVerifyTLS},
 		},
 		VolumeMounts: []corev1.VolumeMount{
 			{Name: "tls-certs", MountPath: "/etc/image-service/certs"},
--- a/docs/operator.md
+++ b/docs/operator.md
@@ -246,6 +246,18 @@ After modifying content of the ConfigMap a new rollout of the Deployment has to 
 oc rollout restart deployment/assisted-service -n assisted-installer
 ```
 
+### Toggle TLS Check on Assisted Image Service
+
+It is possible to toggle TLS checking from the Assisted Image Service by using the annotation `"unsupported.agent-install.openshift.io/assisted-image-service-skip-verify-tls"` on the AgentServiceConfig CR. By default, this is set to `false`, meaning all TLS connections are verified. When this annotation is set to `true`, then the Assisted Image Service skips verifying TLS connections.
+
+It affects the following services: downloading ISO images from a secure server.
+
+Add the annotation to the AgentServiceConfig:
+
+```bash
+oc annotate --overwrite AgentServiceConfig agent unsupported.agent-install.openshift.io/assisted-image-service-skip-verify-tls=true
+```
+
 ### Mirror Registry Configuration
 
 A ConfigMap can be used to configure assisted service to create installations using mirrored content. The ConfigMap contains two keys:


internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -825,6 +825,7 @@ func (r *AgentServiceConfigReconciler) newImageServiceDeployment(ctx context.Con
 			{Name: "ASSISTED_SERVICE_SCHEME", Value: "https"},
 			{Name: "ASSISTED_SERVICE_HOST", Value: serviceName + "." + r.Namespace + ".svc:" + servicePort.String()},
 			{Name: "REQUEST_AUTH_TYPE", Value: "param"},
+			{Name: "INSECURE_SKIP_VERIFY", Value: "false"},
```
carbonin, line 828: I think this needs to either be `true` or configurable in some way.

@filanov what do you think? What are the proportion of connected operator users who would object to us downloading the base isos insecurely?

filanov: Good question, most of the use cases are disconnected environments that probably want to use secured connection. 
So first lets make it configurable, it can be true as default.

CrystalChun: Setting the value to `false` means to verify and use secured connection since it's inverse logic. 

For making it configurable, is it best to do that through the assisted service configmap and then have this environment variable read it through there?

carbonin: I don't think it belongs in the assisted service configmap as it's not configuration for the assisted service.

There are a few ways we could go about this.
1. Put it in the agent service config CRD
    - This is what I have done in the past as it is a property of the installation, but it doesn't seem to be something we're doing here too much
2. Make it an env var in the operator pod
    - We do this with various config currently specifically [here](https://github.com/openshift/assisted-service/blob/master/internal/controller/controllers/images.go)

I'm curious what @flaper87 and @mhrivnak think about these options.
Like I said, I'd generally prefer 1, but the convention in this operator seems to be 2 right now. Is that just to avoid too much churn in the agentserviceconfig API?

mhrivnak: `AgentServiceConfig` is definitely the right way to go. It's an API for configuring and deploying the assisted service and related containers.

Env vars on the operator should be limited to only those things that affect the behavior of the operator itself.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -50,6 +50,13 @@ type MustGatherImage struct {
 	Url string `json:"url"`
 }
 
+type ImageServiceOptions struct {
```
carbonin, line 53: Maybe `ImageServiceConfig`? Naming is hard.
Also I like splitting this out into a separate struct.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -50,6 +50,13 @@ type MustGatherImage struct {
...
+	// InsecureSkipVerify toggles whether or not the image service verifies
+	// certificates when downloading iso images from a server.
+	// +optional
+	InsecureSkipVerify string `json:"insecureSkipVerify"`
```
carbonin, line 57: Any reason this shouldn't be a bool?
Otherwise I feel like we should validate what the user puts here.

mhrivnak, line 57: It would be better to use a type alias and then define constants with that type for each allowed value. The k8s API conventions [describe that](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types).

Then you can use validation markers similar to this:

```
   // +kubebuilder:validation:Optional
   // +kubebuilder:validation:Enum=true;false
   // +kubebuilder:default:=false
```

This will make it obvious to the API user what values can be used and what the default is, while also taking advantage of built-in validation.

mhrivnak: The field name makes sense when it's part of a TLS config struct, but may lack context here. Something like `InsecureSkipTLSVerify` would be more clear IMO.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -50,6 +50,13 @@ type MustGatherImage struct {
 	Url string `json:"url"`
 }
 
+type ImageServiceOptions struct {
+	// InsecureSkipVerify toggles whether or not the image service verifies
+	// certificates when downloading iso images from a server.
```
mhrivnak, line 55: nit: this may not be limited to ISO images.

But it would be good to be explicit about in which circumstances TLS verification is skipped. I assume this does not affect communication with assisted-service for example?

"InsecureSkipVerify toggles whether or not the image service verifies TLS certificates when downloading image artifacts from the sources configured in OSImages..." or something like that.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -82,6 +89,10 @@ type AgentServiceConfigSpec struct {
 	// that are used if one the operators fails to be successfully deployed
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Must-Gather Images"
 	MustGatherImages []MustGatherImage `json:"mustGatherImages,omitempty"`
+
+	// ImageServiceOptions exposes variables for the image service that can be set
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Image Service Options"
+	ImageServiceOptions *ImageServiceOptions `json:"imageServiceOptions,omitempty"`
```
mhrivnak, line 95: nit: IMO we don't need "Options" in the field name. This is a section within the Spec, and I don't think we need each section to re-declare itself as `FooSpec`, `FooOptions`, `FooConfig`, etc. Just `Spec.Foo` will do.

mhrivnak (changes requested):
Sanity-checking the solution:

Plumbing through this setting to the http client is a similar level of work to plumbing through a custom CA certificate. Obviously the latter is more secure; would it be acceptable to the end users? Is there a reason we're not doing that?

carbonin:
@CrystalChun can you associate this with the BZ rather than the JIRA?
If we want to backport it then we'll need it attached to the bug I think.

https://bugzilla.redhat.com/show_bug.cgi?id=2034686

api/v1beta1/agentserviceconfig_types.go
mhrivnak: `bool` is [generally discouraged](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types).

carbonin:
>Plumbing through this setting to the http client is a similar level of work to plumbing through a custom CA certificate. Obviously the latter is more secure; would it be acceptable to the end users? Is there a reason we're not doing that?

Similar, but this is still simpler and we've already verified that this is acceptable to the customer with the issue.
Additionally this is a bit of a temporary fix until we start fetching the images through the release image.
The conversation and decision from @romfreiman and @filanov is in the JIRA

api/v1beta1/agentserviceconfig_types.go
carbonin: That's a great link, thanks!

mhrivnak:
> > Plumbing through this setting to the http client is a similar level of work to plumbing through a custom CA certificate. Obviously the latter is more secure; would it be acceptable to the end users? Is there a reason we're not doing that?
> 
> Similar, but this is still simpler and we've already verified that this is acceptable to the customer with the issue. Additionally this is a bit of a temporary fix until we start fetching the images through the release image. The conversation and decision from @romfreiman and @filanov is in the JIRA

If this is a temporary fix for a single user, how about just using an annotation similar to the ["unsupported configmap"](https://github.com/openshift/assisted-service/blob/19dafda106a632994ba09689b6d7af59b9f32917/docs/operator.md#specifying-environmental-variables-via-configmap) approach? We don't need a whole configmap: how about just an annotation on the AgentServiceConfig:

`unsupported.agent-install.openshift.io/assisted-image-service-tls-verify=false`

That way we can roll out the capability, make it clear that it won't be supported long-term, and then we won't be stuck forever with an API on our CRD that we aren't thrilled with.

CrystalChun:
/retitle Bug 2034686: Expose disable TLS environment variable in image service deployment

carbonin:
> how about just an annotation on the AgentServiceConfig

:+1: yeah I think that makes the most sense. Also because it isn't something we really want most people using.

mhrivnak:
> > how about just an annotation on the AgentServiceConfig
> 
> +1 yeah I think that makes the most sense. Also because it isn't something we really want most people using.

@CrystalChun WDYT? This would make the API minutia irrelevant at least. :smile: In addition to the obvious, I would just suggest logging at process start that the setting is enabled.

CrystalChun:
> > > how about just an annotation on the AgentServiceConfig
> > 
> > 
> > +1 yeah I think that makes the most sense. Also because it isn't something we really want most people using.
> 
> @CrystalChun WDYT? This would make the API minutia irrelevant at least. 😄 In addition to the obvious, I would just suggest logging at process start that the setting is enabled.
That makes sense since it's a temporary fix. I'll go ahead and modify the PR to make it an annotation change and remove the API changes. Thank you both for your input!!

mhrivnak (changes requested):
This is great. If you could just document it [near the other annotation's docs](https://github.com/openshift/assisted-service/blob/19dafda106a632994ba09689b6d7af59b9f32917/docs/operator.md#specifying-environmental-variables-via-configmap), this looks good to me.

mhrivnak (approved):

filanov (approved):

PR #3258, opened 2022-01-28
MGMT-8300: Allow API VIP on cluster creation

mkowalski:
This commit changes the behaviour of cluster creation in order to allow
passing API VIP directly when creating a cluster. This is in order to
remove a need for a subsequent cluster update in order to provide the
value.

This changes the cluster creation API as before it did not allow passing
API VIP. A set of validators is created in order to make sure that API
VIP is allowed for a specific set of scenarios and not unconditionally
for any arbitrary network configuration.

Closes: [MGMT-8300](https://issues.redhat.com/browse/MGMT-8300)

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @ori-amizur 
/cc @filanov 
/cc @flaper87 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/validations/validations.go
+++ b/internal/cluster/validations/validations.go
@@ -393,23 +393,49 @@ func ValidateVipDHCPAllocationWithIPv6(vipDhcpAllocation bool, machineNetworkCID
 	return nil
 }
 
-func DerefString(obj interface{}) *string {
-	switch v := obj.(type) {
-	case string:
-		return swag.String(v)
-	case *string:
-		return v
-	default:
-		return nil
-	}
-}
-
 func ValidateIPAddresses(ipV6Supported bool, obj interface{}) error {
 	var allAddresses []*string
+	var apiVip string
+	var ingressVip string
+	var machineNetworks []*models.MachineNetwork
+	userManagedNetworking := false
+	vipDhcpAllocation := false
+	targetConfiguration := common.Cluster{}
 
-	ingressVip := funk.Get(obj, "IngressVip")
-	apiVip := funk.Get(obj, "APIVip")
-	allAddresses = append(allAddresses, DerefString(ingressVip), DerefString(apiVip))
+	switch c := obj.(type) {
+	case *models.ClusterCreateParams:
+		apiVip = c.APIVip
+		ingressVip = c.IngressVip
+		machineNetworks = c.MachineNetworks
+		if c.UserManagedNetworking != nil {
+			userManagedNetworking = *c.UserManagedNetworking
+		}
+		if c.VipDhcpAllocation != nil {
+			vipDhcpAllocation = *c.VipDhcpAllocation
+		}
+		targetConfiguration.ClusterNetworks = c.ClusterNetworks
+		targetConfiguration.ServiceNetworks = c.ServiceNetworks
+		targetConfiguration.MachineNetworks = c.MachineNetworks
+	case *models.V2ClusterUpdateParams:
+		if c.APIVip != nil {
+			apiVip = *c.APIVip
+		}
+		if c.IngressVip != nil {
+			ingressVip = *c.IngressVip
+		}
+		machineNetworks = c.MachineNetworks
+		if c.UserManagedNetworking != nil {
+			userManagedNetworking = *c.UserManagedNetworking
+		}
+		if c.VipDhcpAllocation != nil {
+			vipDhcpAllocation = *c.VipDhcpAllocation
+		}
+		targetConfiguration.ClusterNetworks = c.ClusterNetworks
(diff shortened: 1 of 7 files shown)


romfreiman:
Are we verifying it's ipv4 only?

mkowalski:
> Are we verifying it's ipv4 only?

Yes we are, in `network.VerifyVips(nil, string(c.MachineNetworks[0].Cidr), api, c.IngressVip, true, nil)` we are comparing against the first Machine Network in the list and there is already another validator for confirming that the first Machine Network is from IPv4.

Just for the sake of completeness I'm adding an explicit unit test for that :)

internal/cluster/validations/validations.go
```diff
@@ -420,6 +420,33 @@ func ValidateIPAddresses(ipV6Supported bool, obj interface{}) error {
...
+
+		dualStack := network.CheckIfClusterIsDualStack(&targetConfiguration)
+		if dualStack {
+			err = network.VerifyVips(nil, string(c.MachineNetworks[0].Cidr), api, c.IngressVip, true, nil)
+			if err != nil {
+				return common.NewApiError(http.StatusBadRequest, err)
+			}
+		} else {
+			if api != "" {
+				err := errors.Errorf("Providing API VIP during cluster creation is allowed only for dual-stack clusters.")
```
ori-amizur, line 444: Why only dual stack ?

internal/cluster/validations/validations.go
```diff
@@ -420,6 +420,33 @@ func ValidateIPAddresses(ipV6Supported bool, obj interface{}) error {
...
+		targetConfiguration.MachineNetworks = c.MachineNetworks
+
+		api := ""
+		if c.APIVip != nil {
+			api = *c.APIVip
+		}
+
+		dualStack := network.CheckIfClusterIsDualStack(&targetConfiguration)
+		if dualStack {
+			err = network.VerifyVips(nil, string(c.MachineNetworks[0].Cidr), api, c.IngressVip, true, nil)
```
ori-amizur, line 438: Are machine networks required during create ?

mkowalski: No, that's a good catch, thanks! We allow creating a cluster object without any Machine Networks, so this check needs to be hardened against it.

internal/cluster/validations/validations.go
mkowalski: My motivations is as follows - currently for single stack there is no logic of providing API VIP during cluster creation. Till now the field was transparently dropped. As I'm adding a scenario for a use case that has been reported, I need to have a flow of "dual-stack + API VIP".

Looks like no one ever complained that they cannot provide API VIP when creating a single-stack cluster. This means that
* either this flow is not supported
* or no one ever used it

Depending on the answer, I may want to keep the error here or again transparently drop the value. What I definitely don't want to do in this PR is to revise all the possible single-stack flows and decide when API VIP may or may not be accepted (because again, today the API VIP is always dropped).

Any thoughts?

ori-amizur: I think for simplicity single and dual stack should behave similarly, except that dual stack has several networks and not one.

mkowalski:
/retitle MGMT-8300: Allow API VIP on cluster creation

internal/cluster/validations/validations.go
mkowalski: Correct. I have added more flows there to not just blindly forbid API VIP for single-stack. At the same time, for single-stack we have a lot of logic around that does e.g. machine network autocalculation and this affects VIPs, so I'd appreciate a review on that here.

I tried to add unit-tests for the most common combinations I could invent. I guess the most "error-prone" scenario will be single-stack where user provides API VIP and at the same time relies on the autocalculation of machine networks.

At the same time for the single-stack we may have a difference in how we treat SNO and multi-node, so yeah, it's all a bit complex.

mkowalski:
The outcome of the architecture forum on February 10th was to proceed with API VIP allowed during cluster creation.

mkowalski:
/retest

Quay problems
```
 Trying to pull quay.io/ocpmetal/oc-image:bug-1823143-multi-arch...
error: build error: error creating build container: initia... received unexpected HTTP status: 500 Internal Server Error 
```

osherdp (approved):

osherdp:
> /retest
> 
> `error: build error: COPY --from=quay.io/edge-infrastructure/swagger-codegen-cli:2.4.18: no stage or image found with that name`

thank you for writing that down, I want to try something that might minimize those flakes

PR #3297, opened 2022-02-07
NO-ISSUE: Add OKD support

### LINKED ISSUE #2580: Make it work with OKD
The docs should contain a hint of how AI can be used to deploy OKD instead of OCP

vrutkovs:
This PR adds assisted-installer support for OKD clusters. Most notable difference between OCP and OKD install is that OKD starts with Fedora CoreOS instead of RHCOS. FCOS doesn't have kubelet/crio/mcd (but has podman), so in order to make bootstrap work on live CD we do the following:
* pull an image with RPMs (`OKD_RPMS_IMAGE` env var)
* create an overlayFS for `/usr` (as its read-only on rpm-ostree systems)
* extract RPMs from the image and use `rpm` to install them
* expand `/var` and `/run` partitions to 6 GB (in FCOS these partitions are 3.2 GB for 16 GB VM which is insufficient for bootstrap to complete)
* hold off agent.service until RPMs are installed
* add fake `systemd-journal-gatewayd.socket` (required from `mcd --once-from` to complete, as FCOS doesn't have journal gateway preinstalled)

This PR also makes easy to deploy OKD-specific version of AI via podman: use `make deploy-onprem OKD=true` to deploy service using `okd-configmap.yml`

TODO:
* [x] Update `generate.sh` to have `okd-configmap.yml` autoupdated when `configmap.yml` is being generated
       Fixed in 3327bcd1f8ab3a8a83b1d84a0161f3804510374e
* [x] Allow `agent.service` to restart on failure. This would make `systemctl start agent` unnecessary.
       Possible unintended consequences?
       Fixed in b84acd808584a68d6478b69121bbe59dd76eaf70 - `Wants/After` is a better solution (agent.service already restarts on failure)

- [x] Fixes #2580 
- [x] Enhancement 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @romfreiman 
/cc @tsorya 
/cc @eranco74 

## Checklist

## Reviewers Checklist

### DIFF
--- a/deploy/podman/okd-configmap.yml
+++ b/deploy/podman/okd-configmap.yml
@@ -1,0 +1,31 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config
+data:
+  ASSISTED_SERVICE_HOST: 127.0.0.1:8090
+  ASSISTED_SERVICE_SCHEME: http
+  AUTH_TYPE: none
+  DB_HOST: 127.0.0.1
+  DB_NAME: installer
+  DB_PASS: admin
+  DB_PORT: "5432"
+  DB_USER: admin
+  DEPLOY_TARGET: onprem
+  DISK_ENCRYPTION_SUPPORT: "false"
+  DUMMY_IGNITION: "false"
+  ENABLE_SINGLE_NODE_DNSMASQ: "false"
+  HW_VALIDATOR_REQUIREMENTS: '[{"version":"default","master":{"cpu_cores":4,"ram_mib":16384,"disk_size_gb":120,"installation_disk_speed_threshold_ms":10,"network_latency_threshold_ms":100,"packet_loss_percentage":0},"worker":{"cpu_cores":2,"ram_mib":8192,"disk_size_gb":120,"installation_disk_speed_threshold_ms":10,"network_latency_threshold_ms":1000,"packet_loss_percentage":10},"sno":{"cpu_cores":8,"ram_mib":16384,"disk_size_gb":120,"installation_disk_speed_threshold_ms":10}}]'
+  IMAGE_SERVICE_BASE_URL: http://127.0.0.1:8888
+  IPV6_SUPPORT: "true"
+  LISTEN_PORT: "8888"
+  NTP_DEFAULT_SERVER: ""
+  POSTGRESQL_DATABASE: installer
+  POSTGRESQL_PASSWORD: admin
+  POSTGRESQL_USER: admin
+  PUBLIC_CONTAINER_REGISTRIES: 'quay.io'
+  SERVICE_BASE_URL: http://127.0.0.1:8090
+  STORAGE: filesystem
+  OS_IMAGES: '[{"openshift_version":"4.9","cpu_architecture":"x86_64","url":"https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/34.20210626.3.1/x86_64/fedora-coreos-34.20210626.3.1-live.x86_64.iso","rootfs_url":"https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/34.20210626.3.1/x86_64/fedora-coreos-34.20210626.3.1-live-rootfs.x86_64.img","version":"34.20210626.3.1"}]'
+  RELEASE_IMAGES: '[{"openshift_version":"4.9","cpu_architecture":"x86_64","url":"quay.io/openshift/okd:4.9.0-0.okd-2022-01-29-035536","version":"4.9.0-0.okd-2022-01-29-035536","default":true}]'
+  OKD_RPMS_IMAGE: quay.io/vrutkovs/okd-rpms:4.9
--- a/hack/generate.sh
+++ b/hack/generate.sh
@@ -88,6 +88,8 @@ function generate_events() {
 function generate_configuration() {
     OS_IMAGES=$(< ${__root}/data/default_os_images.json tr -d "\n\t ")
     RELEASE_IMAGES=$(< ${__root}/data/default_release_images.json tr -d "\n\t ")
+    OKD_OS_IMAGES=$(< ${__root}/data/default_okd_os_images.json tr -d "\n\t ")
+    OKD_RELEASE_IMAGES=$(< ${__root}/data/default_okd_release_images.json tr -d "\n\t ")
     MUST_GATHER_IMAGES=$(< ${__root}/data/default_must_gather_versions.json tr -d "\n\t ")
     OPERATOR_OS_IMAGES=$(< ${__root}/data/default_os_images.json jq -c 'del (.[] | select(.openshift_version == "4.6",.openshift_version == "4.7"))')
     PUBLIC_CONTAINER_REGISTRIES=$(< ${__root}/data/default_public_container_registries.txt)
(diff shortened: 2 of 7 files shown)


deploy/podman/okd-configmap.yml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v1
```
omertuc, line 1: A lot of the values in the original configmap are auto-generated https://github.com/openshift/assisted-service/blob/a9869850e72b002f4e9f89374da2bfd2e60dc43a/hack/generate.sh#L100-L103

omertuc: We also have [automation](https://github.com/openshift-assisted/assisted-installer-deployment/blob/454edbd4fea43c99f6bffc816a3426248a63fa2c/tools/bump_ocp_releases.py#L1) in place to periodically create PRs which bump the z-stream version, you'll probably want to do the same for OKD too

vrutkovs: Good idea, added a TODO

internal/ignition/ignition.go
```diff
@@ -130,6 +130,45 @@ IMAGE=$(echo $1 | sed 's/:.*//')
 podman images | grep $IMAGE || podman rmi --force $1 || true
 `
 
+const okdBinariesOverlayTemplate = `#!/bin/env bash
+set -eux
+# Fetch an image with OKD rpms
+RPMS_IMAGE="%s"
+while ! podman pull --quiet "${RPMS_IMAGE}"
+do
+    echo "Pull failed. Retrying ${RPMS_IMAGE}..."
```
omertuc, line 139: Add a small sleep here, just in-case podman fails on something very early, we probably don't want to spin too quickly

internal/ignition/ignition.go
```diff
@@ -130,6 +130,45 @@ IMAGE=$(echo $1 | sed 's/:.*//')
...
+rpm -Uvh /tmp/rpms/*
+touch /tmp/.okd-rpms-installed
+podman rmi -f "${RPMS_IMAGE}"
+# Expand /var to 6G (3.2G on 16GB VM in insufficient)
+/bin/truncate -s 6G /run/ephemeral.xfsloop
+losetup -c /dev/loop0
+xfs_growfs /var
+mount -o remount,size=6G /run
+# Start agent
+systemctl start agent
```
omertuc, line 161: Isn't the agent already waiting on `/tmp/.okd-rpms-installed`? Why does it need a `start`?

vrutkovs: `agent` is not set to autorestart on failure, so it needs to be kickstarted. I suppose we could update the service, but not sure about unintended effects of this

omertuc: >autorestart on failure
What failure is it hitting? (assuming it's just waiting on your file to show up I wouldn't expect it to fail)

omertuc: >I suppose we could update the service, but not sure about unintended effects of this

That sounds like a good idea, let me verify if this wasn't done on purpose for some good reason

omertuc: Looks (f6403fb153d45bcb3204b61e28c410a6ec029fe2) like it was restarting early on, trying to find the commit where someone removed it

omertuc: >agent is not set to autorestart on failure

Oh wait this doesn't seem true
It is set to auto restart on anything, not just failure https://github.com/openshift/assisted-service/blob/70cd21e2b64f0e28993182c29cae733242e359a9/internal/ignition/ignition.go#L147

vrutkovs: It would be blocked by absence of `/tmp/.okd-rpms-installed` (which systemd treats as not started), but it won't autorestart

omertuc: I'm confused, if it's blocked by the absence of okd-rpms-installed, shouldn't systemd be watching for the existence of this file and start the service when that file shows up?

Or is that not how it works?

vrutkovs: >shouldn't systemd be watching for the existence of this file and start the service when that file shows up, automatically?

Not if you have Restart unset or disabled :/

omertuc: But it is set 
image

vrutkovs: Ah, interesting. I'll play with that some more then

mhrivnak (changes requested):

deploy/podman/okd-configmap.yml
```diff
@@ -0,0 +1,31 @@
...
+  NTP_DEFAULT_SERVER: ""
+  POSTGRESQL_DATABASE: installer
+  POSTGRESQL_PASSWORD: admin
+  POSTGRESQL_USER: admin
+  PUBLIC_CONTAINER_REGISTRIES: 'quay.io'
+  SERVICE_BASE_URL: http://127.0.0.1:8090
+  STORAGE: filesystem
+  OS_IMAGES: '[{"openshift_version": "4.9","cpu_architecture": "x86_64","url": "https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/34.20210626.3.1/x86_64/fedora-coreos-34.20210626.3.1-live.x86_64.iso","rootfs_url": "https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/34.20210626.3.1/x86_64/fedora-coreos-34.20210626.3.1-live-rootfs.x86_64.img", "version": "34.20210626.3.1"}]'
+  RELEASE_IMAGES: '[{"openshift_version": "4.9", "cpu_architecture": "x86_64", "url": "quay.io/openshift/okd:4.9.0-0.okd-2022-01-29-035536", "version": "4.9.0-0.okd-2022-01-29-035536", "default": true}]'
+  OKD_RPMS_IMAGE: quay.io/vrutkovs/okd-rpms:4.9
```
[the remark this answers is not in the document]
vrutkovs: Later on we'll build this image as part of https://github.com/openshift/okd-machine-os and have it included in OKD payload and mirrored to quay.

deploy/podman/okd-configmap.yml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v1
+kind: ConfigMap
```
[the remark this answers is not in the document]
vrutkovs: Updated README in 51e9e91ce34fa0df3ef95aaa49fd488678a0b356

internal/ignition/ignition.go
vrutkovs: Reworked in b84acd808584a68d6478b69121bbe59dd76eaf70 - OKD installs would add `Wants/After` dropin for agent service

internal/ignition/ignition.go
vrutkovs: Fixed in e91b88ece692a523f852439fa2c7fdbf22f88189

deploy/podman/okd-configmap.yml
vrutkovs: Fixed in 3327bcd1f8ab3a8a83b1d84a0161f3804510374e

hack/generate.sh
```diff
@@ -99,8 +99,8 @@ function generate_configuration() {
 
     sed -i "s|OS_IMAGES:.*|OS_IMAGES: '${OS_IMAGES}'|" ${__root}/deploy/podman/configmap.yml
```
omertuc, line 100: Consider creating `assisted-service/data/default_okd_release_images.json` and `assisted-service/data/default_okd_os_images.json`, so the source of truth can be entirely in `data/`

vrutkovs: Ah, good idea. Fixed in f277d54031fa945c805b0c7a5d8e1204b64b49de

vrutkovs:
continuous-integration/jenkins/pr-merge won't pass due to invalid commit messages - I'll squash them in several smaller ones once we're happy with the code changes

hack/generate.sh
omertuc: :ok_hand:

omertuc:
Looks great as far as I'm concerned. @mhrivnak if you don't have any more concerns please /lgtm

/approve

vrutkovs:
Reformatted commits to have `pr-merge` test pass

mhrivnak (changes requested):

deploy/podman/okd-configmap.yml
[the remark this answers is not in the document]
LorbusChris: Also somewhat related to this: https://github.com/openshift/os/issues/498
If that were implemented, it would re-align things nicely.

PR #3310, opened 2022-02-08
NO-ISSUE: Don't wait for CI to update coverage

adriengentil:
Don't wait for all the CI jobs to complete to send the coverage report.
We will then get the coverage as soon as the unit tests have run.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,7 @@
 codecov:
   notify:
     require_ci_to_pass: no
+    wait_for_ci: no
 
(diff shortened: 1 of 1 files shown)


osherdp (approved):

osherdp:
> Can one of the admins verify this patch?

BTW @adriengentil
IIRC, this mechanism of ok-to-test is required because you're not with the assisted-installer-dev role
There's a repo in gitlab that's called app-interface, where you can add this role to your github account

PR #3311, opened 2022-02-08
MGMT-8853: Prevents machine networks conflict errors

CrystalChun:
[MGMT-8853](https://issues.redhat.com/browse/MGMT-8853): Prevents conflict errors with machine networks table in database.

**Context**: Subsystem tests were failing due to the infrequent error
```
time="2022-02-02T00:45:24Z" level=error msg=updateClusterData func="github.com/openshift/assisted-service/internal/bminventory.(*bareMetalInventory).updateClusterInternal" file="/go/src/github.com/openshift/origin/internal/bminventory/inventory.go:2403" cluster_id=fbf88971-6338-4be3-b6e1-0c4a6ccd3e14 \
error="failed to update machine network {1.2.3.0/24 fbf88971-6338-4be3-b6e1-0c4a6ccd3e14} of cluster fbf88971-6338-4be3-b6e1-0c4a6ccd3e14: ERROR: duplicate key value violates unique constraint \"machine_networks_pkey\" (SQLSTATE 23505)" go-id=127982 pkg=Inventory request_id=4ddeda08-b8fd-4dc7-adcf-b8771099e49e
```

**Root cause**: Both the monitoring and the updateCluster were inserting into the machine networks table in different gorm (DB) transactions. Occasionally when monitoring inserted just before the updateCluster inserted into the machine networks table, the unique pkey error above occurs for the subsystem test (since it uses the update cluster) since they both insert the exact same data into the table.

**Solution**: Use the `Create` function instead of the original `Save` function so that way any conflicts can be ignored. In this scenario, nothing is done when there's a conflict since there's no changes to what's being inserted/updated. 

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Testing Details:

1. Run assisted-service on minikube locally
2. Reran a failing subsystem test 1000 times and see if the error occurs
```bash
for i in {1..1001} 
    do make test FOCUS="localhost is not valid" VERBOSE=true SKIP="metrics"
done
```
3. Check the logs to see if the error occurs
```bash
kubectl logs -n assisted-installer $(kubectl get po --no-headers -n assisted-installer | grep assisted-service | awk '{print $1}') | grep "failed to update machine network"
```

Previously, the error would occur once every ~100 trials. After the fix, it does not occur.

## Assignees

/cc @carbonin 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -69,6 +69,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
 )
@@ -2963,7 +2964,8 @@ func (b *bareMetalInventory) updateNetworkTables(db *gorm.DB, cluster *common.Cl
 		}
 		for _, machineNetwork := range cluster.MachineNetworks {
 			machineNetwork.ClusterID = *cluster.ID
-			if err = db.Save(machineNetwork).Error; err != nil {
+			// MGMT-8853: Nothing is done when there's a conflict since there's no change to what's being inserted/updated.
+			if err = db.Clauses(clause.OnConflict{DoNothing: true}).Create(machineNetwork).Error; err != nil {
 				err = errors.Wrapf(err, "failed to update machine network %v of cluster %s", *machineNetwork, params.ClusterID)
 				return common.NewApiError(http.StatusInternalServerError, err)
 			}
--- a/internal/cluster/common.go
+++ b/internal/cluster/common.go
@@ -18,6 +18,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


carbonin:
Do you think it's worth making the same change to the other networks in the same function? Or are those different somehow (maybe they don't have separate tables ...)?

filanov:
i think that the description in the pr belongs to the commit message as well.

filanov:
> Do you think it's worth making the same change to the other networks in the same function? Or are those different somehow (maybe they don't have separate tables ...)?

Is the monitor setting them?

internal/bminventory/inventory.go
```diff
@@ -2963,7 +2964,7 @@ func (b *bareMetalInventory) updateNetworkTables(db *gorm.DB, cluster *common.Cl
 		}
 		for _, machineNetwork := range cluster.MachineNetworks {
 			machineNetwork.ClusterID = *cluster.ID
-			if err = db.Save(machineNetwork).Error; err != nil {
+			if err = db.Clauses(clause.OnConflict{DoNothing: true}).Create(machineNetwork).Error; err != nil {
```
filanov, line 2967: consider adding a comment explaining why it was done.

CrystalChun:
> Do you think it's worth making the same change to the other networks in the same function? Or are those different somehow (maybe they don't have separate tables ...)?

I see in postgres
```
installer=# select * from public.cluster_networks;
 cidr | cluster_id | host_prefix 
------+------------+-------------
(0 rows)

installer=# select * from public.service_networks;
 cidr | cluster_id 
------+------------
(0 rows)
```
The service_networks table probably uses both the cidr and cluster_id as pkeys similar to the machine_networks, but I'm unsure of the cluster_networks. Would it be appropriate to do nothing on conflict when creating them? 

Also I think it's a good idea to check this out and potentially change them if they have the same issues, but I think for this PR it should be limited to the scope of the error (machine networks). What do you think?

carbonin:
>but I think for this PR it should be limited to the scope of the error (machine networks). What do you think?

That's fine. If it's not completely obvious that they would have the same problem (and especially if we're not seeing the same problem) then this can stay targeted to just machine networks.

carbonin (approved):

PR #3322, opened 2022-02-10
MGMT-9233: disk validation fails for day2 hosts

slaviered:
The root cause of the problem was that a day2 cluster had an empty disk encryption configuration while a day2 host was created with an auto-assign role.

The empty disk encryption configuration caused the disk encryption validation to execute. At the same time, a mix of day1 and day2 logic in the validation's code failed the validation due to an absent role.

The fix is to separate day2 from day1's logic in the validation  (day2 logic does not need role assertion) and specifically assign a worker role to day2 hosts.

The latter will also prevent future problems stemming from validations that pends on role assignment.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -6199,6 +6199,11 @@ func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installe
 		c = &cluster.Cluster
 	}
 
+	//day2 host is always a worker
+	if hostutil.IsDay2Host(host) {
+		host.Role = models.HostRoleWorker
+	}
+
 	if err = b.hostApi.RegisterHost(ctx, host, tx); err != nil {
 		log.WithError(err).Errorf("failed to register host <%s> infra-env <%s>",
 			params.NewHostParams.HostID.String(), params.InfraEnvID.String())
--- a/internal/host/validations_test.go
+++ b/internal/host/validations_test.go
@@ -120,12 +120,13 @@ var _ = Describe("Validations test", func() {
 
 		getDay2Host := func() models.Host {
 			h := hostutil.GenerateTestHostByKind(hostID, infraEnvID, &clusterID, models.HostStatusDiscovering, models.HostKindHost, models.HostRoleWorker)
-			*h.Kind = models.HostKindAddToExistingClusterHost
+			h.Kind = swag.String(models.HostKindAddToExistingClusterHost)
 			return h
 		}
 
 		createDay2Cluster := func() {
 			c := hostutil.GenerateTestCluster(clusterID, common.TestIPv4Networking.MachineNetworks)
+			c.Kind = swag.String(models.ClusterKindAddHostsCluster)
 			c.DiskEncryption = &models.DiskEncryption{}
 			Expect(db.Create(&c).Error).ToNot(HaveOccurred())
 		}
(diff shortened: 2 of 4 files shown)


internal/host/validations_test.go
```diff
@@ -414,6 +414,23 @@ var _ = Describe("Validations test", func() {
 			Expect(validationMessage).To(Equal("TPM version could not be found, make sure TPM is enabled in host's BIOS"))
 		})
 
+		It("day2 host - disk encryption is not available", func() {
+			createDay2Cluster()
```
nmagnezi, line 418: looking at createDay2Cluster(), but I suspect it does not create a day2 function, can you please double-check? that might also explain why this is passing despite my comment above.

This function calls GenerateTestCluster(), and does not change anything with cluster Kind
https://github.com/openshift/assisted-service/blob/b2907295956224410f3572858eb2e72a84586c0f/internal/host/validations_test.go#L127-L131

GenerateTestCluster() is using the day1 cluster kind ClusterKindCluster
https://github.com/openshift/assisted-service/blob/b2907295956224410f3572858eb2e72a84586c0f/internal/host/hostutil/test_utils.go#L25-L38

nmagnezi (changes requested):

nmagnezi:
Also, can we have coverage for this in subsystem test(s) as well?

internal/bminventory/inventory.go
```diff
@@ -6199,6 +6199,11 @@ func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installe
 		c = &cluster.Cluster
 	}
 
+	//day2 host is always a worker
+	if hostutil.IsDay2Host(host) {
```
nmagnezi, line 6203: per my earlier comments - why is this in an if statement? it will always be true.

nmagnezi (changes requested):
Please add a test to:
https://github.com/openshift/assisted-service/blob/b2907295956224410f3572858eb2e72a84586c0f/subsystem/day2_cluster_test.go#L90

internal/bminventory/inventory.go
nmagnezi: Actually, it won't always be true. disregard this comment.

internal/host/validator.go
```diff
@@ -393,11 +393,9 @@ func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) Va
 		return ValidationPending
 	}
 
-	role := common.GetEffectiveRole(c.host)
```
danielerez, line 396: Was EnableOn was explicitly set to nil? Otherwise it would skip the flow in line 388.
It should set to default value ('none') here: https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory.go#L407
Maybe we should had 'nil' check to the validation as well then.

slaviered: This is day1 scenario. On day2 the EnableOn is nil because the entire structure is not initialized. With this, you want the validation to run still

PR #3326, opened 2022-02-10
improve docs: mention minikube tunnel

rccrdpccl:
Local minikube environment won't boot without `minikube tunnel` active, as some services type `LoadBalancer` are deployed.

Added documentation to hint at starting `minikube tunnel`

- [ X] Documentation

## Assignees

/cc @michaellevy101 

## Checklist

- [ X] Title and description added to both, commit and PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,13 @@ S3 service (minio), DB and will use the image generator to create the images in 
 skipper make deploy-all
 ```
 
+Note: when deployed in minikube, we need to be running [minikube tunnel](https://minikube.sigs.k8s.io/docs/handbook/accessing/#using-minikube-tunnel), as some services are type `LoadBalancer` and need to be exposed.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,6 +103,9 @@ S3 service (minio), DB and will use the image generator to create the images in
 skipper make deploy-all
 ```
 
+Note: when deployed in minikube, we need to be running [minikube tunnel](https://minikube.sigs.k8s.io/docs/handbook/accessing/#using-minikube-tunnel), as some services are type `LoadBalancer` and need to be exposed.
+
```
nmagnezi, line 107: might be useful to mention this command for readers to know it to run it locally in the background:
```sh
nohup minikube tunnel &>/dev/null
```

osherdp (approved):

PR #3361, opened 2022-02-15
NO-ISSUE: validate api/api-int when ENABLE_SINGLE_NODE_DNSMASQ is not set

vrutkovs:
If SNO DNSMasq manifest is not enabled, validator should check that external api/api-int domain exists. This is mostly useful for OKD SNO, where dnsmasq hack won't work - it uses systemd-resolved and dnsmasq is not included.

- [x] Bug fix

- [x] None

## How was this code tested?

- [x] No tests needed

## Assignees

/cc @omertuc 
/cc @eranco74 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -130,11 +130,10 @@ type PrepareConfig struct {
 }
 
 type Config struct {
-	PrepareConfig           PrepareConfig
-	InstallationTimeout     time.Duration `envconfig:"INSTALLATION_TIMEOUT" default:"24h"`
-	FinalizingTimeout       time.Duration `envconfig:"FINALIZING_TIMEOUT" default:"5h"`
-	MonitorBatchSize        int           `envconfig:"CLUSTER_MONITOR_BATCH_SIZE" default:"100"`
-	EnableSingleNodeDnsmasq bool          `envconfig:"ENABLE_SINGLE_NODE_DNSMASQ" default:"false"`
+	PrepareConfig       PrepareConfig
+	InstallationTimeout time.Duration `envconfig:"INSTALLATION_TIMEOUT" default:"24h"`
+	FinalizingTimeout   time.Duration `envconfig:"FINALIZING_TIMEOUT" default:"5h"`
+	MonitorBatchSize    int           `envconfig:"CLUSTER_MONITOR_BATCH_SIZE" default:"100"`
 }
 
 type Manager struct {
@@ -1151,7 +1150,7 @@ func (m *Manager) GenerateAdditionalManifests(ctx context.Context, cluster *comm
 		return errors.Wrap(err, "failed to add chrony manifest")
 	}
 
-	if common.IsSingleNodeCluster(cluster) && m.EnableSingleNodeDnsmasq {
+	if common.IsSingleNodeCluster(cluster) && m.manifestsGeneratorAPI.IsSNODNSMasqEnabled() {
 		if err := m.manifestsGeneratorAPI.AddDnsmasqForSingleNode(ctx, log, cluster); err != nil {
 			return errors.Wrap(err, "failed to add dnsmasq manifest")
 		}
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -1172,8 +1172,18 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
 	}
 }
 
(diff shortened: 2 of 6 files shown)


internal/host/validator.go
```diff
@@ -1172,8 +1172,9 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
 	}
 }
 
-func isNonePlatformMultiNodeInstallation(c *validationContext) bool {
-	return swag.BoolValue(c.cluster.UserManagedNetworking) && !common.IsSingleNodeCluster(c.cluster)
+func shouldValidateAPIDomainName(c *validationContext) bool {
```
omertuc, line 1175: Name is a bit misleading since it's also called in `isAppsDomainNameResolvedCorrectly` which doesn't have anything to do with the (k8s) API

Should probably be something like `shouldValidateDnsResolution`

internal/cluster/cluster.go
```diff
@@ -1151,7 +1150,7 @@ func (m *Manager) GenerateAdditionalManifests(ctx context.Context, cluster *comm
 		return errors.Wrap(err, "failed to add chrony manifest")
 	}
 
-	if common.IsSingleNodeCluster(cluster) && m.EnableSingleNodeDnsmasq {
+	if common.IsSingleNodeCluster(cluster) {
```
omertuc, line 1153: Forgot a call to `!m.isSNODNSMasqEnabled()` here?

vrutkovs: Fixed by exposing `IsSNODNSMasqEnabled` in `ManifestsGenerator`

internal/host/validator.go
omertuc: But it's not just API, it's also for `isAppsDomainNameResolvedCorrectly`

vrutkovs: Oh, I see, `shouldValidateDnsResolution` is a good choice then

internal/host/validator.go
```diff
@@ -1172,8 +1172,9 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
 	}
 }
 
-func isNonePlatformMultiNodeInstallation(c *validationContext) bool {
-	return swag.BoolValue(c.cluster.UserManagedNetworking) && !common.IsSingleNodeCluster(c.cluster)
+func shouldValidateAPIDnsResolution(c *validationContext) bool {
+	networkCfg := network.Config{}
+	return swag.BoolValue(c.cluster.UserManagedNetworking) && !common.IsSingleNodeCluster(c.cluster) && !networkCfg.EnableSingleNodeDnsmasq
```
omertuc, line 1177: I think I can come up with situations in which this condition doesn't make sense, can you break it up like this:

```go
if !swag.BoolValue(c.cluster.UserManagedNetworking) {
    // No need to perform DNS validations if network is not user managed
    return false;
}

// User managed networking requires DNS validation, unless it's SNO with the dnsmasq configuration
isSnoWithDnsmasq := common.IsSingleNodeCluster(c.cluster) && networkCfg.EnableSingleNodeDnsmasq

return !isSnoWithDnsmasq
```

internal/host/validator.go
```diff
@@ -1172,8 +1172,14 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
...
 }
 
-func isNonePlatformMultiNodeInstallation(c *validationContext) bool {
-	return swag.BoolValue(c.cluster.UserManagedNetworking) && !common.IsSingleNodeCluster(c.cluster)
+func shouldValidateDnsResolution(c *validationContext) bool {
+	// Skip DNS resolution checks in IPI network mode
+	if !swag.BoolValue(c.cluster.UserManagedNetworking) {
+		return false
+	}
+	// Only SNO clusters with DNSMasq manifests enabled need to be checked
```
omertuc, line 1180: The comment should say the opposite (sno with dnsmasq enabled *don't* need to be checked)

vrutkovs: :face_with_spiral_eyes: fixed

vrutkovs:
Unit tests passed
/hold cancel

filanov:
/hold 
Please add tests to manifests generator

eranco74 (commented):
/lgtm
/unhold

filanov:
@adriengentil looks like the coverage report have some issue with the rebase?

PR #3365, opened 2022-02-16
MGMTBUGSM-98: Prevent nil pointer error when cluster ID is nil.

CrystalChun:
[MGMTBUGSM-98](https://issues.redhat.com/browse/MGMTBUGSM-98)
Originally when populating the logs URL in the agent CR, there
was no check for the clusterID so if there was a nil clusterID
then assisted service would go into a crash loop.

This change moves the logic for populating the logs URL under
the check for the cluster ID.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

To be tested in @jelkosz's environment

## Assignees

/cc @carbonin 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -437,14 +437,14 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 			if err != nil {
 				return ctrl.Result{Requeue: true}, nil
 			}
-		}
-		if agent.Status.DebugInfo.LogsURL == "" && !time.Time(h.LogsCollectedAt).Equal(time.Time{}) { // logs collection time is updated means logs are available
-			logsURL, err := generateControllerLogsDownloadURL(r.ServiceBaseURL, clusterId.String(), r.AuthType, agent.Name, "host")
-			if err != nil {
-				log.WithError(err).Error("failed to generate controller logs URL")
-				return ctrl.Result{}, err
+			if agent.Status.DebugInfo.LogsURL == "" && !time.Time(h.LogsCollectedAt).Equal(time.Time{}) { // logs collection time is updated means logs are available
+				logsURL, err := generateControllerLogsDownloadURL(r.ServiceBaseURL, clusterId.String(), r.AuthType, agent.Name, "host")
+				if err != nil {
+					log.WithError(err).Error("failed to generate controller logs URL")
+					return ctrl.Result{}, err
+				}
+				agent.Status.DebugInfo.LogsURL = logsURL
 			}
-			agent.Status.DebugInfo.LogsURL = logsURL
 		}
 		connected(agent, status)
 		requirementsMet(agent, status)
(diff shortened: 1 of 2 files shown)


internal/controller/controllers/agent_controller.go
```diff
@@ -437,14 +437,14 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 			if err != nil {
 				return ctrl.Result{Requeue: true}, nil
 			}
-		}
-		if agent.Status.DebugInfo.LogsURL == "" && !time.Time(h.LogsCollectedAt).Equal(time.Time{}) { // logs collection time is updated means logs are available
-			logsURL, err := generateControllerLogsDownloadURL(r.ServiceBaseURL, clusterId.String(), r.AuthType, agent.Name, "host")
-			if err != nil {
-				log.WithError(err).Error("failed to generate controller logs URL")
-				return ctrl.Result{}, err
+			if agent.Status.DebugInfo.LogsURL == "" && !time.Time(h.LogsCollectedAt).Equal(time.Time{}) { // logs collection time is updated means logs are available
```
filanov, line 440: What will happen to the url in case the host is bound to a new cluster? 
It looks like we will keep the same URL without the new cluster no? 
it may involve UX as well, i see two options 
1. keep the old url until the host get bound to a new cluster - can help debug issues in case host was replaced for some reason but may be confusing to uses that we will see logs on an unbound host
2. Delete the url when host is unbound - we lose the logs but it will not confuse the user
Crystal what do you think?

filanov (commented):
suggesting to cover a case where `UnbindHostInternal` fails

internal/controller/controllers/agent_controller.go
carbonin: I would expect that whatever is responsible for binding (or maybe unbinding) a host would clear these fields.
There should be enough time to grab the url for a failed host before it gets reassigned.

That all said, I think this PR can fix the bug first as we were not clearing the logs (or events for that matter) before this change, right?

filanov: Sounds good to me. 
Regardless need to cover it with a unit tests https://github.com/openshift/assisted-service/pull/3365#pullrequestreview-884949401

CrystalChun: Good point about when a host becomes unbound! 

I agree, I think we can create a separate PR/issue for investigating the log removal when a host is unbound. I'll take a look at the current unbound logic and see if we can include it there.

filanov: > A question I have is: internally, when a host is unbound, do the logs get removed from storage?
I don't think so 
The cleanup usually done when deleting the cluster.

internal/controller/controllers/agent_controller_test.go
```diff
@@ -578,6 +578,33 @@ var _ = Describe("agent reconcile", func() {
 		Expect(conditionsv1.FindStatusCondition(agent.Status.Conditions, v1beta1.SpecSyncedCondition).Status).To(Equal(corev1.ConditionTrue))
 	})
 
+	It("Agent update status with no cluster id ", func() {
```
filanov, line 581: it doesn't look like the coverage have changed. 
I suggest to check a specific flow and fail `UnbindHostInternal` similar to https://github.com/openshift/assisted-service/blob/67e0bc1275f07acf46a2f4531f6984e6f7edb7f3/internal/controller/controllers/agent_controller_test.go#L542

PR #3367, opened 2022-02-16
MGMT-9134: convert update hostname to v2 in subsystem

slaviered:
This is a pre-requisite code for removing V1 RegisterCluster.
Updating hostname is now done via V2UpdateHost rather than via UpdateCluster

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -6502,6 +6502,7 @@ func (b *bareMetalInventory) V2UpdateHostInternal(ctx context.Context, params in
 	log := logutil.FromContext(ctx, b.log)
 	var c *models.Cluster
 	var cluster *common.Cluster
+	var usages usage.FeatureUsage = make(usage.FeatureUsage)
 
 	txSuccess := false
 	tx := b.db.Begin()
@@ -6527,7 +6528,7 @@ func (b *bareMetalInventory) V2UpdateHostInternal(ctx context.Context, params in
 	if err != nil {
 		return nil, err
 	}
-	err = b.updateHostName(ctx, host, params.HostUpdateParams.HostName, tx)
+	err = b.updateHostName(ctx, host, params.HostUpdateParams.HostName, usages, tx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


carbonin (approved):

carbonin:
Looks like the test is failing because we don't set usage when we update the host directly I guess?
Sounds like something we should actually fix.

/hold

internal/bminventory/inventory.go
```diff
@@ -3182,16 +3182,15 @@ func (b *bareMetalInventory) updateHostNames(ctx context.Context, params install
 			log.WithError(err).Errorf("invalid hostname format: %s", err)
 			return err
 		}
+		localhost = localhost || host.RequestedHostname == "localhost"
```
carbonin, line 3185: What is the logic supposed to be here?

It looks like we're going to get `localhost == true` here for every host in the loop after the first one that is set to `localhost`.
That doesn't seem right.

slaviered: This is exactly what I aimed for: to have at least one host that updated its name from localhost to something else

carbonin: > to have at least one host that updated its name from localhost to something else

That isn't what this code is doing. I think you need to change the condition to `host.RequestedHostname != "localhost"`. And also ideally change the variable name to something more descriptive like `haveCustomHostname`.

Right now this variable is effectively `haveHostWithLocalhost` which I don't think is what you want.

internal/bminventory/inventory.go
```diff
@@ -6596,12 +6608,13 @@ func (b *bareMetalInventory) updateHostRole(ctx context.Context, host *common.Ho
...
 }
 
-func (b *bareMetalInventory) updateHostName(ctx context.Context, host *common.Host, hostname *string, db *gorm.DB) error {
+func (b *bareMetalInventory) updateHostName(ctx context.Context, host *common.Host, hostname *string, usages usage.FeatureUsage, db *gorm.DB) error {
 	log := logutil.FromContext(ctx, b.log)
 	if hostname == nil {
 		log.Infof("No request for hostname update for host %s", host.ID)
 		return nil
 	}
+	localhost := host.Host.RequestedHostname == "localhost"
```
carbonin, line 6617: Same here, so we're setting `from_localhost` to `true` when the hostname _is_ `localhost`?
That seems backwards.

internal/bminventory/inventory.go
```diff
@@ -6613,6 +6626,7 @@ func (b *bareMetalInventory) updateHostName(ctx context.Context, host *common.Ho
 			*hostname, host.ID, host.InfraEnvID)
 		return common.NewApiError(http.StatusConflict, err)
 	}
+	b.setUsage(true, usage.RequestedHostnameUsage, &map[string]interface{}{"from_localhost": localhost}, usages)
```
carbonin, line 6629: What is `from_localhost` supposed to mean?
Do we even really need extra data in this case?

If the usage is set to true then a host has requested a hostname other than `localhost`, right?

carbonin:
Talked with @slaviered and we figured that it was best to not change the extra data sent with the metric in a PR that is mainly just trying to move subsystem over to v2.

For now this will use the same metric as before and we can open a new issue and PR for changing the meaning of the metric as a follow up.

PR #3385, opened 2022-02-23
MGMT-9356: Add go mod to models

filanov:
it's a first step toward spliting kube-api api dir into a sub module.
kube-api crd definition is using definitions from the models directly
to avoid cycle dependency first need to split the model into a sub module
from the main go mod

require specific gorm version because i had some issue with timezone
translation

go.mod is created once, if there are chnagese in the swagger file that
reuqire to update go.mod it will be done manually it's not common
process but if we call go mod every time we will have conflicts on the
pr checks every time package is updated event none of the code changes
so it's fair trade

- [ ] New Feature 
- [X] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @flaper87 
/cc @eranco74 
/cc @omertuc 

## Checklist

## Reviewers Checklist


filanov:
/hold
I'm not sure about the automatic run of go mod, it may cause some conflicts, i wasn't to see how the CI react first.

osherdp:
I think dependabot will try to update all the ``go.mod|sum`` files when it bumps versions
At least it used to do that in Python: https://github.com/openshift/assisted-test-infra/pull/1294

hack/generate.sh
```diff
@@ -56,6 +56,10 @@ function remove_dashes_and_dots() {
     done
 }
 
+function init_models_submodule() {
+    cd models && go mod init github.com/openshift/assisted-service/models && go mod tidy
+}
```
osherdp, line 61: To make ``cd`` scope only for this function, use:
```suggestion
function init_models_submodule() (
    cd models && go mod init github.com/openshift/assisted-service/models && go mod tidy
)
```

Or alternatively wrap ``cd models && ...`` with parenthesis

go.mod
```diff
@@ -86,9 +88,11 @@ require (
 replace (
 	github.com/googleapis/gnostic => github.com/googleapis/gnostic v0.4.1
 	github.com/metal3-io/baremetal-operator => github.com/openshift/baremetal-operator v0.0.0-20210409032903-31b989a197eb // Use OpenShift fork
+	github.com/openshift/assisted-service/models => ./models
 	go.etcd.io/bbolt => go.etcd.io/bbolt v1.3.5
 	go.etcd.io/etcd => go.etcd.io/etcd v0.5.0-alpha.5.0.20200910180754-dd1b699fc489 // ae9734ed278b is the SHA for git tag v3.4.13
 	google.golang.org/grpc => google.golang.org/grpc v1.27.1
+	gorm.io/gorm => gorm.io/gorm v1.22.3
```
osherdp, line 95: isn't it inconsistent with the models' version?

osherdp (changes requested):

go.mod
filanov: yes but in the models it have a new version that caused me some issues with unit tests, i wrote in the comment 
```
require specific gorm version because i had some issue with timezone
translation
```

hack/generate.sh
filanov: I think that i will remove the automatic part. it may cause issues if we run generate from swagger in the CI and it will automatically update a module and complain about it. 
So it will be static and if the complication fails then the author of the PR will need to fix it, similar to what we have in the root directory, what do you think?

go.mod
osherdp: Ah, I thought you meant to use a specific gorm version, but the same version on both modules

filanov: The gorm version in under model is newer `gorm.io/gorm v1.23.1` 
because the rood model import from it so it takes the same version, but it cause some issue in the code, so i enforce specific version in the root mod file.

hack/generate.sh
osherdp: that solution you did is actually a great idea. just removing ``*.go`` files

eranco74 (commented):
/lgtm

osherdp:
@filanov unhold when you're ready

flaper87 (approved):
I'd be lying if I said I checked every single line. Structure looks good.

thanks for working on this

PR #3389, opened 2022-02-23
MGMT-9383: Add enhancement doc for iPXE boot

carbonin:
Not all user infrastructure is able to boot isos directly, and iPXE is a commonly
used alternative for booting machines. We should support this by making the
application directly serve iPXE artifacts so that systems can boot with no
other iPXE infrastructure required.

https://issues.redhat.com/browse/MGMT-9383

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @mhrivnak 
/cc @avishayt 
/cc @filanov 
/cc @flaper87 
/cc @dtantsur 
/cc @CrystalChun 

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/ipxe-host-boot.md
+++ b/docs/enhancements/ipxe-host-boot.md
@@ -1,0 +1,154 @@
+---
+title: ipxe-host-boot
+authors:
+  - "@carbonin"
+  - “@mhrivnak”
+creation-date: 2022-02-22
+last-updated: 2022-02-24
+---
+
+# iPXE Host Boot
+
+## Summary
+
+Not all user infrastructure is able to boot isos directly, and iPXE is a commonly 
+used alternative for booting machines. We should support this by making the 
+Assisted Installer directly serve iPXE artifacts so that systems can boot with no
+other iPXE infrastructure required.
+
+## Motivation
+
+Booting a host using iPXE should be a first-class operation when using
+assisted-service. This has been found to be at least as common a method 
+as providing the live ISO via virtual media and is also required in some 
+environments. Users are already unpacking the ISO we provide into PXE
+artifacts and serving those in their infrastructure. We can make this much
+easier for them by serving those artifacts as well as the iPXE boot script
+directly from our application.
+
+### Goals
+
+- Serve iPXE artifacts (kernel, rootfs, and initrd)
+- Serve an iPXE boot script
+- Include the ignition in the provided initrd
+
+### Non-Goals
+
+- Support PXE (tftp based)
+- Integration with ironic (separate effort)
+- iPXE boot script customization (maybe in a follow-up if it’s useful)
+- Configure the minimal ISO to pull the rootfs from the image service
+
+## Proposal
+
+The image-service will serve the kernel, rootfs, and initrd with ignition 
+appended (see Implementation Details), while the assisted service will 
+provide an API to download the boot script. The boot script created and served 
+by the assisted service will contain URLs to the required artifacts hosted by 
+the image service.
+
+### User Stories
+
+#### Story 1
+
+As a user of assisted installer I want to boot my hosts using iPXE without 
+setting up my own http server to host artifacts.
+
+### Implementation Details/Notes/Constraints
+
+#### Ironic Integration
+
+To support integration with Ironic’s existing iPXE boot methods, the 
+discovery ignition must be appended to the initrd rather than specified as 
+a separate kernel parameter. The image service will also handle this initrd 
+manipulation when an initrd is requested for a particular infra-env.
+
+#### New Image Service APIs
+
+Image service APIs will be added to download the required boot artifacts. The 
+rootfs and kernel will be served directly from the template image. The initrd 
+endpoint will be specific to an infra-env in order to embed the ignition.
+
+- `GET /boot-artifacts/{artifact-name}?{params}`
+  - artifact-name can be either “rootfs” or “kernel”
+  - params must include “version” and “arch” which are used to identify a template image
+  - No authentication is required as we’re serving these directly from the uncustomized image
+  - Examples
+    - `/boot-artifacts/rootfs?version=4.9&arch=x86_64`
+    - `/boot-artifacts/kernel?version=4.10&arch=arm64`
+- `GET /images/{image-id}/pxe-initrd?{params}`
+  - image-id is the uuid of an infra-env in assisted-service
+  - params must include “version” and “arch”
+  - params may include “api_key” or “image_token” depending on assisted-service authentication configuration
+  - Example
+    - `/images/afe293b9-e3a0-47ed-8952-f26721626497/pxe-initrd?version=4.10&arch=x86_64&image_token={token}`
+
+#### New Assisted Service REST APIs
+
+Assisted service APIs will be added to download the ipxe boot script and to 
+retrieve a presigned url for downloading the boot script.
+
+- `GET /v2/infra-envs/{infra-env-id}/downloads/ipxe-script`
+  - Will return an iPXE boot script with artifact URLs pointing to the image service
+  - Initrd URL will be presigned
+- `GET  /v2/infra-envs/{infra-env-id}/downloads/ipxe-script-url`
+  - Will return a presigned URL for the iPXE script
+
+#### Assisted Service Kubernetes APIs
+
+A presigned download url for the iPXE boot script and all relevant artifacts will be added
+to the InfraEnv resource status. This URL will be set as a part of the InfraEnv controller
+reconcile loop.
+
+Proposed API addition example:
+
+```yaml
+status:
+  bootArtifacts:
+    initrd: https://image-service.example.com/images/afe293b9-e3a0-47ed-8952-f26721626497/pxe-initrd?version=4.10&arch=x86_64
+    rootfs: https://image-service.example.com/boot-artifacts/rootfs?version=4.10&arch=x86_64
+    kernel: https://image-service.example.com/boot-artifacts/kernel?version=4.10&arch=x86_64
+    ipxeScript: https://assisted-service.example.com/v2/infra-envs/afe293b9-e3a0-47ed-8952-f26721626497/downloads/ipxe-script
+```
+
+#### Authentication Tokens
+
+Both the iPXE boot script and the initrd download URLs will be presigned with an 
+authentication token. In the cloud offering (console.redhat.com) the same mechanism 
+described in image-service-cloud-authentication will be used which means the tokens 
+will include an expiration.
+
+This expiration will be set to four hours for both the iPXE boot script and the initrd.
+The expected use case will be for users to point their machines directly at the boot 
+script URL which can be refreshed by the UI easily.
+
+This only becomes a potential problem if a user downloads and hosts the iPXE script, but 
+not the initrd. This is because the URL in the boot script for the initrd will expire after 
+four hours and machines will no longer be able to boot. In this case the expectation is that 
+such a user would download all the artifacts, host them in their own infrastructure, and edit 
+the boot script accordingly.
+
+#### Proxy Support
+
+It should be noted explicitly that iPXE does not have support for proxy servers.
+This means that if a host needs a proxy to access the boot artifacts the artifacts
+will need to be mirrored and hosted internally.
+
+This should be called out in the UI as the assisted installer otherwise works well
+by just specifying proxy information.
+
+### Risks and Mitigations
+
+## Design Details [optional]
+
+### Open Questions
+
+### UI Impact
+
+Is the UI going to display the boot script or just provide the URL?
+
+### Test Plan
+
+## Drawbacks
+
+## Alternatives


docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+### Risks and Mitigations
+
+## Design Details [optional]
+
+### Open Questions
+
+When running in the cloud how do we set expiration times for the tokens we use for 
+auth when downloading the boot script and the initrd?
+The boot script can probably be treated similarly to the image as it is only downloaded once.
+The initrd will be downloaded every time a machine boots and the url needs to remain valid.
```
dtantsur, line 111: Will it? I would assume you'll switch to local boot after the initial deployment.

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+  - params may include “api_key” or “image_token” depending on assisted-service authentication configuration
+  - Example
+    - `/images/afe293b9-e3a0-47ed-8952-f26721626497/pxe-initrd?version=4.10&arch=x86_64&image_token={token}`
+
+#### New Assisted Service REST APIs
+
+Assisted service APIs will be added to download the ipxe boot script and to 
+retrieve a presigned url for downloading the boot script.
+
+- `GET /v2/infra-envs/{infra-env-id}/downloads/ipxe-script`
```
dtantsur, line 91: You could convert all URL params to kernel params.

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+  - Will return an iPXE boot script with artifact URLs pointing to the image service
+  - Initrd URL will be presigned
+- `GET  /v2/infra-envs/{infra-env-id}/downloads/ipxe-script-url`
+  - Will return a presigned URL for the iPXE script
+
+#### Assisted Service Kubernetes APIs
+
+A presigned download url for the iPXE boot script will be added to the infra-env 
+resource status. This URL will be set as a part of the infra-env controller reconcile loop.
+
```
dtantsur, line 101: For the Ironic integration purposes, it would be somewhat easier to have the initramfs link directly on the `InfraEnv` resource since that's the resource we'll be interacting with (and that's how ISO is handled).

Otherwise we'll probably start parsing the iPXE script :)

docs/enhancements/ipxe-host-boot.md
carbonin: Sorry, I don't mean if the same single machine reboots, but rather if/when additional machines boot using the same script.

docs/enhancements/ipxe-host-boot.md
carbonin: That's interesting. That's assuming users are manipulating the original url we give them though which probably isn't the UX we're looking for (at least in the cloud).

FWIW we don't have a way to alter kernel parameters for the discovery iso today so it wouldn't be a regression if we just provide the same parameters that currently exist in the iso boot config.

docs/enhancements/ipxe-host-boot.md
carbonin: This is a good point. I don't want to force parsing the iPXE script on anyone.

Does it make sense to do both to ease the ZTP automation? At that point should we also provide urls to the rootfs and kernel for completeness? Any thoughts on this @filanov @avishayt @mhrivnak ?

carbonin: Do we know when we're using ZTP in the infra-env controller? Maybe we put all these options in the API as optional and provide the initramfs if we're in ZTP "mode" and the script url otherwise?

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+last-updated: 2022-02-22
+---
+
+# iPXE Host Boot
+
+## Summary
+
+Not all user infrastructure is able to boot isos directly, and iPXE is a commonly 
+used alternative for booting machines. We should support this by making the 
+application directly serve iPXE artifacts so that systems can boot with no
```
avishayt, line 16: Which application?
```suggestion
Assisted Installer directly serve iPXE artifacts so that systems can boot with no 
```

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+- Serve iPXE artifacts (kernel, rootfs, and initrd)
+- Serve an iPXE boot script
+- Include the ignition in the provided initrd
+
+### Non-Goals
+
+- Support PXE (tftp based)
+- Integration with ironic (separate effort)
+- iPXE boot script customization (maybe in a follow-up if it’s useful)
+- Configure the minimal ISO to pull the rootfs from the image service
```
avishayt, line 40: What does minimal ISO have to do with this proposal?

docs/enhancements/ipxe-host-boot.md
avishayt: We may as well provide URLs for all artifacts.  We could be explicit and have a property for each, or provide one URL where the user will need to specify the artifact as a query param.  Not sure which is better.

docs/enhancements/ipxe-host-boot.md
carbonin: Currently when we create the minimal ISO template image we embed a url to the rootfs from the OsImages structure.
Now that we're serving the rootfs we could point back to our own service rather than requiring the user mirror the rootfs in addition to the full iso.

I'm saying doing that is out of scope here but could be something we do in the future, but I can remove it from this list if it's causing confusion.

docs/enhancements/ipxe-host-boot.md
mhrivnak: I think it's fine to have a URL for each one. I think we should not expect users to compose URLs as part of this API.
```
status:
  ipxe:
    rootfs: https://....
    initramfs: https://....
    ...
```

mhrivnak: Or maybe some of these artifacts would be used outside an ipxe context? Then just call the section "artifacts" or something similar instead of "ipxe".

filanov: providing parameters sounds complicated, if it's another filed in the status i don't think that it's a big deal

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+### Goals
+
+- Serve iPXE artifacts (kernel, rootfs, and initrd)
+- Serve an iPXE boot script
+- Include the ignition in the provided initrd
+
+### Non-Goals
+
+- Support PXE (tftp based)
+- Integration with ironic (separate effort)
```
romfreiman, line 38: metal3? :)

docs/enhancements/ipxe-host-boot.md
romfreiman: worth having a task on that? It would be more of a self contained solution, right?

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+#### Story 1
+
+As a user of assisted installer I want to boot my hosts using iPXE without 
+setting up my own http server to host artifacts.
+
+### Implementation Details/Notes/Constraints
+
+#### Ironic Integration
+
+To support integration with Ironic’s existing iPXE boot methods, the
```
romfreiman, line 61: will it work for other platforms as well? or something specific to ironic?

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+a separate kernel parameter. The image service will also handle this initrd 
+manipulation when an initrd is requested for a particular infra-env.
+
+#### New Image Service APIs
+
+Image service APIs will be added to download the required boot artifacts. The 
+rootfs and kernel will be served directly from the template image. The initrd 
+endpoint will be specific to an infra-env in order to embed the ignition.
+
+- `GET /boot-artifacts/{artifact-name}?{params}`
```
romfreiman, line 72: is it under the same assisted swagger? how would it be exposed from c.rh.com?

docs/enhancements/ipxe-host-boot.md
romfreiman: yea, better than ipxe... then we can serve the rootfs to the minimal iso in the same manner, no?

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Design Details [optional]
+
+### Open Questions
+
+When running in the cloud how do we set expiration times for the tokens we use for 
+auth when downloading the boot script and the initrd?
+The boot script can probably be treated similarly to the image as it is only downloaded once.
+The initrd will be downloaded every time a machine boots and the url needs to remain valid.
+This raises UX-type questions. Will people point their machines at the boot script endpoint
```
romfreiman, line 112: hmm looking into equinix option, they let u provide script endpotint. I would assume it can be pointed directly to the SaaS. Otherwise, someone will have to host it - more resource, extra machine, networking. I would like to avoid it. @javipolo wdyt?

docs/enhancements/ipxe-host-boot.md
filanov: as far as i understand it's under the image service not assisted

docs/enhancements/ipxe-host-boot.md
filanov: what other platforms? 
as far as i know this is the only platform that running another agent no? or is there another use case?

docs/enhancements/ipxe-host-boot.md
romfreiman: that's exactly my quesiton. How would it be served from c.rh.com if it's not under AI

carbonin: The image service is already exposed at a different path than assisted service.
ISO downloads from the production environment look like this today:
`https://api.openshift.com/api/assisted-images/images/{uuid}?{params}`

So this endpoint would be:
`https://api.openshift.com/api/assisted-images/boot-artifacts/{artifact-name}?{params}`

docs/enhancements/ipxe-host-boot.md
carbonin: This just means that we're forced to include the ignition as a part of the initrd as opposed to a separate kernel parameter (as is done in the [fcos docs](https://docs.fedoraproject.org/en-US/fedora-coreos/live-booting/#_booting_via_ipxe), for example).

I don't think it makes much sense for us to do this in the ZTP case and not in others so we'd do it the same no matter what. This is just mentioning how the ironic use-case is forcing the design a particular way.

docs/enhancements/ipxe-host-boot.md
carbonin: :shrug: I don't know what the difference is honestly.
What I meant here is that this is in support of the ZTP convergence flow rather than implementing any of it.

docs/enhancements/ipxe-host-boot.md
carbonin: > yea, better than ipxe... then we can serve the rootfs to the minimal iso in the same manner, no?

Nothing about the rootfs (or this url) is specific to ipxe, this is just about how we present the URL to the user in the infra-env.

docs/enhancements/ipxe-host-boot.md
romfreiman: what I meant is if we provide this kind of initrd to equinix server - will it boot? Or it needs ignition seperately

carbonin: Should boot fine.

docs/enhancements/ipxe-host-boot.md
avishayt: Can leave it here - let's open a task

docs/enhancements/ipxe-host-boot.md
avishayt: But it doesn't matter - these URLs aren't really part of the API.  The URLs appear as properties in the infraenv.  Nobody should really be constructing them by hand.

docs/enhancements/ipxe-host-boot.md
avishayt: OK, so maybe let's stick with boot artifacts:
```
status:
  bootArtifacts:
    rootfs: https://....
    initramfs: https://....
    ...
```

docs/enhancements/ipxe-host-boot.md
javipolo: I'd totally point directly to the SaaS, it makes all the process way easier and cleaner from the user point of view

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,125 @@
...
+authors:
+  - "@carbonin"
+  - “@mhrivnak”
+creation-date: 2022-02-22
+last-updated: 2022-02-22
+---
+
+# iPXE Host Boot
+
+## Summary
```
omertuc, line 12: - Think about proxies

docs/enhancements/ipxe-host-boot.md
ohadlevy: there could be an equiv grubby script or other items that can use these artifacts.

docs/enhancements/ipxe-host-boot.md
ohadlevy: > @ohadlevy if I remember correctly iPXE doesn't support proxies right?

right. same for UEFI http boot afaik

docs/enhancements/ipxe-host-boot.md
carbonin: Depends on the scope. If "assisted installer" means the whole stack (including the assisted service, image service, agent, controller, etc) then yeah, this is fine.

As stated later, technically the image service will host the artifacts and assisted service will host the script so I guess "Assisted Installer" makes sense here?

docs/enhancements/ipxe-host-boot.md
filanov: then we will probably need a warning in the UI right? 
Users still can download the artifacts and host them locally but it will not work directly.

docs/enhancements/ipxe-host-boot.md
carbonin: Created MGMT-9406 which is marked as blocking on MGMT-9383

docs/enhancements/ipxe-host-boot.md
carbonin: Is there somewhere in this doc we want to call this out specifically?
Maybe in "Risks and Mitigations"?

omertuc: Probably `### Implementation Details/Notes/Constraints` under a UI/UX section

docs/enhancements/ipxe-host-boot.md
carbonin: I think this may be a good direction for specifying kernel arguments in the future, but not really in the scope of this enhancement (in fact I specifically have "iPXE boot script customization" in the non-goals section already).

docs/enhancements/ipxe-host-boot.md
carbonin: Added an example for the kube api addition.

docs/enhancements/ipxe-host-boot.md
carbonin: Added a section about proxy.

docs/enhancements/ipxe-host-boot.md
```diff
@@ -0,0 +1,154 @@
...
+- `GET /v2/infra-envs/{infra-env-id}/downloads/ipxe-script`
+  - Will return an iPXE boot script with artifact URLs pointing to the image service
+  - Initrd URL will be presigned
+- `GET  /v2/infra-envs/{infra-env-id}/downloads/ipxe-script-url`
+  - Will return a presigned URL for the iPXE script
+
+#### Assisted Service Kubernetes APIs
+
+A presigned download url for the iPXE boot script and all relevant artifacts will be added 
+to the infra-env resource status. This URL will be set as a part of the infra-env controller
```
mhrivnak, line 100: Suggest calling this `InfraEnv`, which is the actual API name.

carbonin: Sure, do you have an opinion on how it's written in the rest of the doc? Do we have a standard for how to reference it if we're not specifically talking about the kube API?

mhrivnak: I don't know if we have a convention for the general concept. I think the rest of the doc is fine where it's not referring to a specific API.

flaper87:
@avishayt should we merge this enhancement?

avishayt:
Looks good to me
/unhold

PR #3417, opened 2022-02-28
MGMT-9115: Remove V1 GetDiscoveryIgnition UpdateDiscoveryIgnition

nmagnezi:
This change does several things:
- Removes `GetDiscoveryIgnition`; unit tests will use `V2DownloadInfraEnvFiles`.

- in the absence of a helper function: `pkg/middleware` is now exportable and
  includes a `GetNext` to assert against types in unit tests. `NewResponder()` left in place.

- Removes `UpdateDiscoveryIgnition`; unit tests will use `UpdateInfraEnv`.

- A couple[1] of `UpdateDiscoveryIgnition` unit tests are no longer applicable.
  The reason: `objectHandler.DeleteObject` is not being used by `UpdateInfraEnv`.

- Removed a couple of events that were used exclusively in `UpdateDiscoveryIgnitionInternal()`:
  `discovery_ignition_config_applied` and `ignition_updated_therefore_image_deleted`.

[1] https://github.com/openshift/assisted-service/blob/e32309591c61d6de29db0255ae56a2f419085eac/internal/bminventory/inventory_test.go#L8749-L8780

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @carbonin 

## Checklist

## Reviewers Checklist


docs/events.yaml
```diff
@@ -408,20 +408,6 @@ x-events:
   properties:
     infra_env_id: UUID
 
-- name: discovery_ignition_config_applied
-  message: "Custom discovery ignition config was applied to the cluster"
```
carbonin, line 412: Do we have an analogous event for infra-env?
I think this would be good to see in events.

internal/bminventory/inventory_test.go
```diff
@@ -8887,100 +8902,71 @@ var _ = Describe("UpdateDiscoveryIgnition", func() {
...
-			eventstest.WithClusterIdMatcher(clusterID.String())))
-		response := bm.UpdateDiscoveryIgnition(ctx, params)
-		Expect(response).To(BeAssignableToTypeOf(&installer.UpdateDiscoveryIgnitionCreated{}))
+		params := installer.UpdateInfraEnvParams{
+			InfraEnvID:           *infraEnv.ID,
+			InfraEnvUpdateParams: &models.InfraEnvUpdateParams{IgnitionConfigOverride: override},
+		}
+		mockIgnitionBuilder.EXPECT().FormatDiscoveryIgnitionFile(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(discovery_ignition_3_1, nil).Times(2)
+		mockS3Client.EXPECT().GetBaseIsoObject(gomock.Any(), gomock.Any()).Return("rhcos", nil).Times(1)
+		mockS3Client.EXPECT().UploadISO(gomock.Any(), gomock.Any(), "rhcos", gomock.Any()).Return(nil).Times(1)
```
carbonin, line 8913: Not for this PR, but this isn't how this works in any environment any more.
After the v1 cleanup we should change the unit tests to use the image service everywhere.

docs/events.yaml
filanov: Good question, now that it's part of infra env user will probably get an even when updating it no? 
@avishayt maybe it's something that we need to consider, do we want to publish another event for specific updates? 
Or is the message in update event should include additional information about things that have been updated?

filanov (commented):
we don't have any subsystem test for the ignition part?

docs/events.yaml
avishayt: I think that for the regular update API it would be messy to list all the changes.  If there is something out of the ordinary (like customizations) then it's important for support and we should have events, but changing network or hostname doesn't need it.

nmagnezi: Currently, `UpdateInfraEnv()` (which replaced `UpdateDiscoveryIgnition`) produce `ignition_config_image_generated` event for success: 
https://github.com/openshift/assisted-service/blob/483d52644b50f652f6c96e6f614734e24f3353ad/docs/events.yaml#L584-L590

and `upload_image_failed` in a case of failure:
https://github.com/openshift/assisted-service/blob/483d52644b50f652f6c96e6f614734e24f3353ad/docs/events.yaml#L577-L582

The `ignition_config_image_generated` event is different from `discovery_ignition_config_applied` and `ignition_updated_therefore_image_deleted` has no replacement here, probably because of the changes we had to work with the image-service.

Whould we add more events? and if yes, should it be a part of this PR?

internal/bminventory/inventory_test.go
nmagnezi: Agreed. I just opened [MGMT-9460](https://issues.redhat.com/browse/MGMT-9460) to capture this.

nmagnezi:
> we don't have any subsystem test for the ignition part?

Unfortunately no.

nmagnezi:
ci/prow/e2e-metal-assisted-operator-ztp fails multiple times  with `Timed out waiting for csv to become ready!`

https://search.ci.openshift.org/?search=Timed+out+waiting+for+csv+to+become+ready%21&maxAge=48h&context=1&type=bug%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

Retriggered while checking why.

carbonin:
> we don't have any subsystem test for the ignition part?

We have that "dummy" ignition generator still from back when we were using a kube job.
Maybe it's time to get rid of that? Is there any reason we should still be using it?

docs/events.yaml
carbonin: I don't think we're actually ever sending either of those events because of the early return for the image service here: https://github.com/openshift/assisted-service/blob/01aef1f8361609863f4e97cf7b77d0b18447150c/internal/bminventory/inventory.go#L1166-L1172

There's a LOT of dead code in there that I want to get to cleaning up after the main v1 removal is done.

pkg/filemiddleware/middleware.go
```diff
@@ -10,23 +10,27 @@ import (
...
-	return &fileMiddlewareResponder{
+	return &FileMiddlewareResponder{
 		next:     next,
 		fileName: fname,
 		length:   length,
 	}
 }
 
-type fileMiddlewareResponder struct {
+type FileMiddlewareResponder struct {
```
carbonin, line 20: I'll put it here just for future us, but it would be really nice to be able to validate responses in tests with `middleware.Responder` objects by just using the existing API (i.e. `WriteResponse`) rather than having the do type assertions every where in tests.

carbonin (approved):

PR #3419, opened 2022-03-01
MGMT-9177: OVNKubernetes as network type with SNO

empovit:
- Do not allow OpenShiftSDN as a network type for SNO
- Select OVNKubernetes automatically if no network type was specified

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/validator.go
+++ b/internal/cluster/validator.go
@@ -240,7 +240,11 @@ func (v *clusterValidator) isNetworkTypeValid(c *clusterPreprocessContext) Valid
 		return ValidationFailure
 	}
 
-	if doesClusterNetworksMatchNetworkType(c.cluster) {
+	if hasClusterNetworksUnsupportedByNetworkType(c.cluster) {
+		return ValidationFailure
+	}
+
+	if isHighAvailabilityModeUnsupportedByNetworkType(c.cluster) {
 		return ValidationFailure
 	}
 
@@ -260,8 +264,10 @@ func (v *clusterValidator) printIsNetworkTypeValid(context *clusterPreprocessCon
 		if !funk.ContainsString(validNetworkTypes, swag.StringValue(context.cluster.NetworkType)) && context.cluster.NetworkType != nil {
 			return "The network type is not valid; the valid network types are OpenShiftSDN or OVNKubernetes"
 		}
-		if doesClusterNetworksMatchNetworkType(context.cluster) {
+		if hasClusterNetworksUnsupportedByNetworkType(context.cluster) {
 			return "The cluster is configured with IPv6 which is not supported by OpenShiftSDN; use OVNKubernetes instead"
+		} else if isHighAvailabilityModeUnsupportedByNetworkType(context.cluster) {
+			return "High-availability mode 'None' (SNO) is not supported by OpenShiftSDN; use another network type instead"
 		} else if isVipDhcpAllocationAndOVN(context.cluster) {
 			return "VIP DHCP allocation is not supported when the cluster is configured to use OVNKubernetes."
 		} else {
@@ -272,13 +278,18 @@ func (v *clusterValidator) printIsNetworkTypeValid(context *clusterPreprocessCon
 	}
 }
 
-func doesClusterNetworksMatchNetworkType(cluster *common.Cluster) bool {
+func hasClusterNetworksUnsupportedByNetworkType(cluster *common.Cluster) bool {
 	return funk.Any(funk.Filter(common.GetNetworksCidrs(cluster), func(ip *string) bool {
 		if ip == nil {
 			return false
 		}
 		return network.IsIPv6CIDR(*ip)
 	})) && cluster.NetworkType != nil && swag.StringValue(cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes
+}
+
+func isHighAvailabilityModeUnsupportedByNetworkType(cluster *common.Cluster) bool {
+	return swag.StringValue(cluster.HighAvailabilityMode) == models.ClusterHighAvailabilityModeNone &&
+		cluster.NetworkType != nil && swag.StringValue(cluster.NetworkType) == models.ClusterNetworkTypeOpenShiftSDN
 }
 
 func isVipDhcpAllocationAndOVN(cluster *common.Cluster) bool {
--- a/internal/controller/controllers/clusterdeployments_controller.go
+++ b/internal/controller/controllers/clusterdeployments_controller.go
@@ -909,7 +909,7 @@ func selectClusterNetworkType(params *models.V2ClusterUpdateParams, cluster *com
 			return false
 		}
 		return network.IsIPv6CIDR(*ip)
-	})) {
+	})) || common.IsSingleNodeCluster(cluster) {
 		return models.ClusterNetworkTypeOVNKubernetes
 	} else {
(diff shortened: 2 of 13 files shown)


internal/cluster/validator.go
```diff
@@ -281,6 +287,11 @@ func doesClusterNetworksMatchNetworkType(cluster *common.Cluster) bool {
 	})) && cluster.NetworkType != nil && swag.StringValue(cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes
 }
 
+func doesAvailabilityModeMatchNetworkType(cluster *common.Cluster) bool {
+	return swag.StringValue(cluster.HighAvailabilityMode) == models.ClusterHighAvailabilityModeNone &&
```
romfreiman, line 291: @empovit the user will still should be able to specify their own network type. We should not allow only SDN

internal/cluster/validator.go
```diff
@@ -281,6 +287,11 @@ func doesClusterNetworksMatchNetworkType(cluster *common.Cluster) bool {
 	})) && cluster.NetworkType != nil && swag.StringValue(cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes
 }
 
+func doesAvailabilityModeMatchNetworkType(cluster *common.Cluster) bool {
```
eranco74, line 290: I don't think we should block, we just want the default for SNO to be OVNKuberntes

internal/cluster/validator.go
```diff
@@ -262,6 +266,8 @@ func (v *clusterValidator) printIsNetworkTypeValid(context *clusterPreprocessCon
 		}
 		if doesClusterNetworksMatchNetworkType(context.cluster) {
 			return "The cluster is configured with IPv6 which is not supported by OpenShiftSDN; use OVNKubernetes instead"
+		} else if doesAvailabilityModeMatchNetworkType(context.cluster) {
```
eranco74, line 269: I don't think we should block it, in case some reason we do block it, 
note that users can use the API to set whatever network type they like e.g. Calico, Cilium, Flannel etc.
it doesn't even have to be `OpenShiftSDN`

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -909,7 +909,7 @@ func selectClusterNetworkType(params *models.V2ClusterUpdateParams, cluster *com
 			return false
 		}
 		return network.IsIPv6CIDR(*ip)
-	})) {
+	})) || common.IsSingleNodeCluster(cluster) {
```
eranco74, line 912: Seems strange that this is the only place this change is required, what about the REST API?

internal/cluster/validator.go
romfreiman: the decision with PMs was to block

internal/cluster/validator.go
romfreiman: we should block OpenShiftSDN, default to OVN, all the rest are allowed

internal/cluster/validator.go
empovit: @romfreiman Interesting. Can I have more background on the issue? Where does the requirement come from? The ticket says "_Allow OVN only as NetworkType_". Am I misinterpreting it?

internal/cluster/validator.go
empovit: @eranco74 The ticket says "_Allow OVN only as NetworkType_". What am I missing?

internal/controller/controllers/clusterdeployments_controller.go
empovit: @eranco74 Doesn't the validator https://github.com/openshift/assisted-service/pull/3419/files/965c7c9a42d13b3f501bb9392572b4d2cfd26c02#diff-a712b20e2e375b0286d437906cfa73f4423a506449c19e71ba14633cd89dd01b serve the REST API? According to the unit tests it does. In any case, it's still a WIP, I will keep looking for other places.

internal/cluster/validator.go
romfreiman: yea, sorry for the confusion. Look at the epic

internal/cluster/validator.go
eranco74: OK, I thought we just want it as default.
But if the PMs want to block it so be it.

internal/controller/controllers/clusterdeployments_controller.go
eranco74: So after going over the code it seems that if no network type was provided by the user the assisted-installer defaults to the  network type returned by the Openshift installer (decided when generating the install config) this means that different OCP versions will return different network type

internal/cluster/validator.go
omertuc, line 290:
```suggestion
func doesHighAvailabilityModeMatchNetworkType(cluster *common.Cluster) bool {
```

omertuc: or alternatively
```suggestion
func doesHAModeMatchNetworkType(cluster *common.Cluster) bool {
```

internal/cluster/validator.go
```diff
@@ -281,6 +287,11 @@ func doesClusterNetworksMatchNetworkType(cluster *common.Cluster) bool {
 	})) && cluster.NetworkType != nil && swag.StringValue(cluster.NetworkType) != models.ClusterNetworkTypeOVNKubernetes
 }
 
+func doesAvailabilityModeMatchNetworkType(cluster *common.Cluster) bool {
+	return swag.StringValue(cluster.HighAvailabilityMode) == models.ClusterHighAvailabilityModeNone &&
+		cluster.NetworkType != nil && swag.StringValue(cluster.NetworkType) == models.ClusterNetworkTypeOpenShiftSDN
```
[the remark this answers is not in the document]
empovit: We want to block SNO+SDN, don't we?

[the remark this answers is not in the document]
romfreiman: yes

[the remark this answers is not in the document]
empovit: I think the function name is confusing you, but I had to follow a pattern here :/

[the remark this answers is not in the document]
empovit: I renamed the functions to better show their meaning

internal/controller/controllers/clusterdeployments_controller.go
empovit: @eranco74 @romfreiman IIUC it means an SNO installation may end up having OpenShiftSDN, which is exactly what we want to avoid. Do you think we should default to OVNKubernetes in the API, too? Or make network type mandatory in case of SNO?

eranco74: I think we should default in the API.
While there is a similar task for the OpenShift installer, I don't think we want to depend on it or on the version it will get delivered to.
Adding a default in the assisted-service API also adds visibility of the network type prior to the installation.

eranco74 (commented):
/lgtm
I think we should add default network type on the create API (can be added in another PR)

PR #3426, opened 2022-03-02
MGMT-9221 Automatic user-defined agent labels enhancement

avishayt:
Enhancement for automatic user-defined agent labels

### DIFF
--- a/docs/enhancements/automatic-agent-classification-labels.md
+++ b/docs/enhancements/automatic-agent-classification-labels.md
@@ -1,0 +1,129 @@
+---
+title: automatic-agent-classification-labels
+authors:
+  - "@avishay"
+creation-date: 2022-02-27
+last-updated: 2022-02-27
+---
+
+# Automatic Agent Classification Labels
+
+## Summary
+
+When using late binding, users select which Agents should be part of a cluster based on their inventory.  The selection can be manual or by using a label selector.
+Two kinds of labels are helpful:
+* Inventory labels: Simply represents the hardware, such as architecture, virtualization support, etc.  The agent controller already adds these labels.
+* User-defined labels: These are labels that a user would add to characterize their hardware.  Examples include “t-shirt sizes” or hosts suitable for hosting storage or machine learning workloads.  There is currently no automation for adding these labels.
+
+This proposal is to automate user-defined labels.  Users would define the desired labels and conditions, and the system would automate the process.
+
+## Motivation
+
+Having users define labels manually is time-consuming and error-prone.  Defining conditions is more reliable and scalable for users.
+
+### Goals
+
+- Provide a flexible and easy-to-use API for defining desired labels and conditions.
+- Extend the Infrastructure Operator to maintain the desired labels based on the specified conditions with minimal impact on scale.
+
+### Non-Goals
+
+- Ship pre-defined labels and conditions.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a Cluster Creator, I want the system to automatically label Agents based on specific conditions so that I can more easily select Agents for cluster membership.	
+
+## Design Details [optional]
+
+We will create a new AgentClassification namespace-scoped CRD that defines the desired label key-value and a query to be evaluated according to the Agent’s inventory.
+
+### AgentClassification CRD Definition
+
+```
+const (
+	QueryValidCondition conditionsv1.ConditionType = "QueryValid"
+	QueryValidReason    string                     = "QueryIsValid"
+	QueryNotValidReason string                     = "QueryIsNotValid"
+
+	QueryErrorsCondition conditionsv1.ConditionType = "QueryErrors"
+	QueryErrorsOK        string                     = "NoQueryErrors"
+	QueryHasErrors       string                     = "HasQueryErrors"
+)
+
+// AgentClassificationSpec defines the desired state of AgentClassification
+type AgentClassificationSpec struct {
+	// LabelKey specifies the label key to apply to matched Agents
+	//
+	// +immutable
+	LabelKey string `json:"labelKey"`
+
+	// LabelValue specifies the label value to apply to matched Agents
+	//
+	// +immutable
+	LabelValue string `json:"labelValue"`
+
+	// Query is in gojq format (https://github.com/itchyny/gojq#difference-to-jq)
+	// and will be invoked on each Agent's inventory. The query should return a
+	// boolean. The operator will apply the label to any Agent for which "true"
+	// is returned.
+	Query string `json:"query"`
+}
+
+// AgentClassificationStatus defines the observed state of AgentClassification
+type AgentClassificationStatus struct {
+	// MatchedCount shows how many Agents currently match the classification
+	MatchedCount int `json:"matchedCount,omitempty"`
+
+	// ErrorCount shows how many Agents encountered errors when matching the classification
+	ErrorCount int `json:"errorCount,omitempty"`
+
+	Conditions []conditionsv1.Condition `json:"conditions,omitempty"`
+}
+```
+
+We will base the expression definition on the [gojq](https://github.com/itchyny/gojq) library, which supports jq queries in Go.  The query will be run on each Agent's inventory.  If the query returns true, then the specified label will be applied.
+
+For example,
+The query for label "size:medium" might be: ".cpu.count == 2 and .memory.PhysicalBytes >= 4294967296 and .memory.PhysicalBytes < 8589934592"
+The query for label "storage:large" might be: "[.disks[] | select(.sizeBytes > 1073741824000)] | length > 5"
+
+### Operator changes
+
+A new agent-label-controller will reconcile Agent CRs and update labels on Agents as necessary.  It will be triggered by updates to both Agents and AgentClassifications via watch mapping.  The controller lists the AgentClassifications in the Agent's namespace and for each:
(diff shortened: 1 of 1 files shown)


docs/enhancements/automatic-agent-classification-labels.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Operator changes
+
+When an Agent's inventory is updated, the Agent controller will now need to list all AgentClassification CRs in the namespace and update the Agent's labels accordingly.
+
+A new controller will watch AgentClassification CRs.  Upon creation/update, it will list all Agents in the namespace and apply the relevant labels.  It will also ensure that the specified label key is unique in that namespace.  Upon deletion, it will delete the label from all Agents.
+
+### Risks and Mitigations
+
+With a large number of Agents and AgentClassifications, the scale can be adversely affected.  To mitigate this, we scope the effects of an AgentClassifications to its namespace and evaluate the expressions only if the Agent's inventory was updated or an AgentClassification Spec was updated.  Both should not happen often.
```
eranco74, line 84: We keep adding logic and calculations around inventory changes.
The inventory gets updated on every inventory response from the agent (some optimization exists but doesn't seem to work)

eranco74 (commented):
Overall looks like a great idea

docs/enhancements/automatic-agent-classification-labels.md
```diff
@@ -0,0 +1,101 @@
...
+type ExpressionDefinition struct {
+	// Expression is the expression that will be evaluated, in govaluate format
+	Expression string `json:"expression"`
+
+	// Parameters maps parameters from the above expression to paths in the Agent inventory
+	Parameters map[string]string `json:"parameters"`
+}
+```
+
+We will base the expression definition on the [govaluate](https://github.com/Knetic/govaluate) library, which supports a language for expressions with parameters.  Along with each expression (e.g., “cpu > 8”), the user will define a list of variables and map them to inventory items (e.g., “cpu”: “Cpu.Count”).  The controller will need to parse the inventory path and translate it into the corresponding value.  In addition, the inventory contains arrays for disks and network interfaces - the user will pass "Disks(N).SizeBytes" for at least N disks having the specified capacity, or "Disks().SizeBytes" for all disks.  The controller will need to parse this and create additional govaluate expressions for each disk, and then check for at least N returning true.
```
omertuc, line 74: >"Disks(N).SizeBytes" for at least N disks having the specified capacity, or "Disks().SizeBytes" for all disks

This sounds pretty limiting.

For example, we have disks in the inventory which are not fit for installation. Usually these disks you'd never want to use at all, not just for installation. Users would probably want to say things like "all disks that are eligible for installation" or "all SSD disks", i.e. do some sort of filtering on the network/disk arrays, and this DSL would not allow such flexibility.

Could we somehow make use of an off-the-shelf JSON query language, something like the `jq` query language, or jsonpath?

docs/enhancements/automatic-agent-classification-labels.md
avishayt: Yes, definitely.  I recently opened [MGMT-9343](https://issues.redhat.com/browse/MGMT-9343) to track this.

docs/enhancements/automatic-agent-classification-labels.md
filanov: i see that there something like that https://github.com/itchyny/gojq

docs/enhancements/automatic-agent-classification-labels.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Story 1
+
+As a Cluster Creator, I want the system to automatically label Agents based on specific conditions so that I can more easily select Agents for cluster membership.	
+
+## Design Details [optional]
+
+We will create a new AgentClassification namespace-scoped CRD that defines the desired label key-value and a list of expressions to be evaluated according to the Agent’s inventory.  Each expression is generic such that it can refer to any properties in the inventory.
+
+### AgentClassification CRD Definition
```
filanov, line 45: will it have a status indicating the validity of the expressions?

docs/enhancements/automatic-agent-classification-labels.md
```diff
@@ -0,0 +1,101 @@
...
+	// Parameters maps parameters from the above expression to paths in the Agent inventory
+	Parameters map[string]string `json:"parameters"`
+}
+```
+
+We will base the expression definition on the [govaluate](https://github.com/Knetic/govaluate) library, which supports a language for expressions with parameters.  Along with each expression (e.g., “cpu > 8”), the user will define a list of variables and map them to inventory items (e.g., “cpu”: “Cpu.Count”).  The controller will need to parse the inventory path and translate it into the corresponding value.  In addition, the inventory contains arrays for disks and network interfaces - the user will pass "Disks(N).SizeBytes" for at least N disks having the specified capacity, or "Disks().SizeBytes" for all disks.  The controller will need to parse this and create additional govaluate expressions for each disk, and then check for at least N returning true.
+
+### Operator changes
+
+When an Agent's inventory is updated, the Agent controller will now need to list all AgentClassification CRs in the namespace and update the Agent's labels accordingly.
```
filanov, line 78: does it have to be the same controller? 
If i remember correct when updating inventory agent controller doesn't look at inventory changes, maybe it need to be a different controller that will watch the agents?

docs/enhancements/automatic-agent-classification-labels.md
avishayt: Yes, forgot to add, thanks

docs/enhancements/automatic-agent-classification-labels.md
avishayt: Great suggestion, I will look into using this instead of govaluate.

docs/enhancements/automatic-agent-classification-labels.md
avishayt: Why do we need another controller to watch agents?  We already have a method that updates the inventory in the Status, so why not update the labels there?

filanov: Actually it depend on the scale as well, if we have the agent to do all the work then it will mean that the reconcile loop will be mo complicated and will probably run longer, if it's a different controller then we can split the work, after all it is just adding labels. 
I'm not sure if it's really the right way, but it looks like it's doing very specific work.

docs/enhancements/automatic-agent-classification-labels.md
avishayt: Added, please take a look

docs/enhancements/automatic-agent-classification-labels.md
avishayt: Good question.  I'm not sure how much parallelism there is between controllers, so even if we move it to another controller they might execute serially.  I will check that.

docs/enhancements/automatic-agent-classification-labels.md
```diff
@@ -0,0 +1,106 @@
...
+	//
+	// +immutable
+	LabelKey string `json:"labelKey"`
+
+	// LabelValue specifies the label value to apply to matched Agents
+	//
+	// +immutable
+	LabelValue string `json:"labelValue"`
+
+	// A query in jq format that will be invoked on each Agent's inventory.
```
omertuc, line 66:
```suggestion
	// A query in gojq format (https://github.com/itchyny/gojq#difference-to-jq) that will be invoked on each Agent's inventory.
```

docs/enhancements/automatic-agent-classification-labels.md
filanov: Is there a condition that will validate the expressions? the status just show matched count. 
How will the user debug it?

avishayt: Added Conditions in the Status and defined QueryValidCondition const

docs/enhancements/automatic-agent-classification-labels.md
```diff
@@ -0,0 +1,112 @@
...
+The query for label "size:medium" might be: ".cpu.count == 2 and .memory.PhysicalBytes >= 4294967296 and .memory.PhysicalBytes < 8589934592"
+The query for label "storage:large" might be: "[.disks[] | select(.sizeBytes > 1073741824000)] | length > 5"
+
+### Operator changes
+
+When an Agent's inventory is updated, the Agent controller will now need to list all AgentClassification CRs in the namespace and update the Agent's labels accordingly.
+
+A new controller will watch AgentClassification CRs.  Upon creation/update, it will list all Agents in the namespace and apply the relevant labels.  It will also ensure that the specified label key is unique in that namespace.  Upon deletion, it will delete the label from all Agents.
+
+### Risks and Mitigations
```
omertuc, line 93: As a user looking at a particular agent, how can I tell apart these two situations :
1. Agent didn't match any labels
2. The controller didn't label the agent yet

Would it be possible to add a meta label/annotation to all agents indicating when they were last processed by the labeling controller? It would eliminate this ambiguity and provide consumers with a way to ensure no races occur when processing those labels (after creating a new AgentClassification, a consumer wouldn't have to wait an arbitrary amount of time while the controller is processing possibly thousands of agents - they could use this timestamp indicator to ensure a particular agent has been classified)

omertuc: Alternatively, if it's possible from a technical perspective (I'm not sure it is), it might be better to give each AgentClassification a "done processing" status.

Not sure how to define such thing though, since agents can be created / deleted all the time

filanov (approved):

PR #3436, opened 2022-03-06
MGMT-9527 fix generate-bundle after moving api to sub modules

filanov:
operator-sdk doesn't know how to work with sub-modules so it require to
enter into api directory and run the commands from there, in addition
the PROJECT file need to be located in the same place, according to
operator-sdk code i saw that it is not configurable to set the path to
this file.
For some reason even after setting input and output dir it still creates
a local copy of config/manifests/kustomization.yaml so i had to do a
cleanup and remove it.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @osherdp 
/cc @djzager 

## Checklist

## Reviewers Checklist

### DIFF
--- a/hack/generate.sh
+++ b/hack/generate.sh
@@ -140,7 +140,7 @@ with open("'${__root}/config/samples/agent-install.openshift.io_v1beta1_agentser
 }
 
 # Generate manifests e.g. CRD, RBAC etc.
-function generate_manifests() {
+function generate_manifests() (
     if [ "${ENABLE_KUBE_API:-}" != "true" ]; then exit 0; fi
 
     local crd_options=${CRD_OPTIONS:-"crd:trivialVersions=true"}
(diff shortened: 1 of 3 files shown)


filanov:
/hold
@mhrivnak, @djzager let me know what you think. 
The context is that we split the api directory into a separate sub module, but it doesn't play nice with operator-sdk and to make it work i had to make some hacks. 
Let me know if you have a suggestion to cleaner solution.

hack/generate.sh
```diff
@@ -158,14 +158,20 @@ function generate_manifests() {
...
+        cd ..
     fi
 
     cp ${controller_crd_path}/resources.yaml ${BUILD_FOLDER}/resources.yaml
 }
 
 function generate_bundle() {
     ENABLE_KUBE_API=true generate_manifests
-    operator-sdk generate kustomize manifests --apis-dir api -q
+    # temp copy for operator-sdk that doesn't know how to handle sub-modules
```
osherdp, line 169: Maybe replace with a TODO so it will be explicit that we want it as temporary solution

hack/generate.sh
```diff
@@ -158,14 +158,20 @@ function generate_manifests() {
         kustomize build ${controller_crd_path} > ${controller_crd_path}/resources.yaml
         controller-gen object:headerFile=${hack_boilerplate} paths="./..."
         goimports -w  ${controller_path}
+        cd ..
```
osherdp, line 161: you can also replace function parenthesis to have regular brackets instead of curly brackets
it will make ``cd ./api`` an action relevant only to the function scope
```
function generate_manifests() (
    ...
)

osherdp (changes requested):
as long as it works, I think it will be okay (as a temporary fix)

hack/generate.sh
osherdp: that comment is only an optional suggestion, not required ^

hack/generate.sh
filanov: Unless someone have a better solution it will not be temporary, the load function that is called by operator-sdk is not configurable https://github.com/kubernetes-sigs/kubebuilder/blob/cf1575e08dff6c2b63d442b8fa6ec397fba64363/pkg/config/store/yaml/store.go#L65-L67

hack/generate.sh
filanov: Done

hack/generate.sh
osherdp: maybe we can submit an issue for ``operator-sdk`` to handle that (including handling their dependencies)

osherdp (approved):

hack/generate.sh
filanov: good idea https://github.com/operator-framework/operator-sdk/issues/5583

filanov:
/unhold
unholding because we are blocked without it. 
If anything have suggestions we will fix them in another PR

PR #3460, opened 2022-03-09
MGMT-9539: Add presigned iPXE script URL endpoint

carbonin:
### Add an endpoint which generates a pre-signed URL to the iPXE script for an infra-env.

The new endpoint itself is rather simple. It uses the URL builder to create the url and signs it using the same method we use for ISO downloads.

The more complicated parts are the two initial refactoring commits:

#### Deduplicate presigned URL swagger objects
Previously there were two url types defined in the swagger spec `presigned` and `infra-env-image-url` both served the same purpose and had a `url` attribute.
This commit combines these implementations into a new type, `presigned-url` which will also be used for the upcoming endpoint to get a url to download the ipxe boot script.

#### Allow image auth in a query param
This commit adds a new security definition `imageURLAuth` to handle query parameter based auth for downloading infra-env files.
This is how a pre-signed URL for the ipxe-script will be implemented.
The authentication for this definition will be implemented using the same functions as imageAuth. The only differences here are how the token is provided (query param rather than header) and to which endpoints it applies (only the infra-env files download for url-based).

https://issues.redhat.com/browse/MGMT-9539

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @avishayt 
/cc @CrystalChun 

## Checklist

## Reviewers Checklist


swagger.yaml
```diff
@@ -9129,8 +9173,10 @@ definitions:
         type: string
         format: uuid
 
-  infra-env-image-url:
```
filanov, line 9132: does it affect any of the clients?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -643,3 +643,37 @@ func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *c
 
 	return fmt.Sprintf(ipxeScriptFormat, kernelURL, rootfsURL, initrdURL), nil
 }
+
+func (b *bareMetalInventory) GetInfraEnvIPXEScriptDownloadURL(ctx context.Context, params installer.GetInfraEnvIPXEScriptDownloadURLParams) middleware.Responder {
```
filanov, line 647: Besides getting the URL is it going to be stored? 
I don't remember this part and i didn't find it in the enhancement

swagger.yaml
```diff
@@ -4470,6 +4475,53 @@ paths:
           schema:
             $ref: '#/definitions/error'
 
+  /v2/infra-envs/{infra_env_id}/downloads/ipxe-script-url:
```
avishayt, line 4478: We already have:
/v2/clusters/{cluster_id}/downloads/files-presigned
So maybe create
/v2/infra-envs/{infra_env_id}/downloads/files-presigned
?

internal/bminventory/inventory_v2_handlers.go
carbonin: > it going to be stored?

No, only in the infraenv kube-api.

swagger.yaml
carbonin: I'm not sure, good question. I hope not as the API isn't actually changing, but maybe it would change the type name that was generated or something.

I don't think the agent is getting pre-signed URLs and given that the tests passed I guess it's not a problem for test-infra. Can we ask someone on the UI team?

swagger.yaml
carbonin: :shrug: Sure.

Do we care about making it work for the discovery ignition as well as the ipxe script?

swagger.yaml
jkilzi: **It will change the name of the generated TypeScript interface**

TL;DR: no big deal because the object remains the same (same members) it won't affect the UI code during runtime. In our dev-environments, we'll have to rename the interface in a few places.

PS: thanks for marking the `url` as required :)

filanov (approved):

filanov:
was merged to 2.5 no need to backport.

PR #3469, opened 2022-03-13
NO-ISSUE: sporadic ci failure to get epel metadata

slaviered:
building assisted-service-build in ci fails sporadically
with the message:
Cannot retrieve metalink for repository: epel/x86_64

cleaning the cache and updating certificates was recommended as
the course of action to reduce the probability of such an error

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/hack/setup_env.sh
+++ b/hack/setup_env.sh
@@ -53,10 +53,10 @@ function assisted_service() {
     install -o root -g root -m 0755 /tmp/kubectl /usr/local/bin/kubectl && \
     rm -f /tmp/kubectl
 
+  yum clean all && yum makecache
+  yum --disablerepo=epel -y update ca-certificates
   yum install -y --setopt=skip_missing_names_on_install=False \
     docker podman awscli python3-pip genisoimage skopeo p7zip
-
-  yum clean all
 
   butane
 


slaviered:
@osherdp is this the right place for the update of certificates and cache cleaning or should I transfer it out to the start of the assisted-service-build dockerfile just after fetching the epel package?

hack/setup_env.sh
```diff
@@ -53,11 +53,11 @@ function assisted_service() {
     install -o root -g root -m 0755 /tmp/kubectl /usr/local/bin/kubectl && \
     rm -f /tmp/kubectl
 
+  yum clean all && yum makecache
+  yum --disablerepo=epel -y update ca-certificates
```
osherdp, line 57: what ``--disablerepo=epel`` means in the context of this command?

osherdp:
> @osherdp is this the right place for the update of certificates and cache cleaning or should I transfer it out to the start of the assisted-service-build dockerfile just after fetching the epel package?

depending on if it's related only to docker or in general

hack/setup_env.sh
danielerez: iinm, it's just a common workaround to update certs, as epel repo is using https. I.e. update the certificates while avoiding failure with fetching epel.

osherdp: so we update certs for all repositories but epel, and afterwards it will not try to do the same for epel?
any link to this kind of suggestion? (from stackoverflow I assume)

danielerez: I think it should update all certs, the disable of epel is just to prevent failure.
here's some reference: https://techglimpse.com/yum-install-epel-repo-error-solution/

osherdp: can we see if we can just remove epel usage?
that would be better if possible

osherdp (changes requested):

hack/setup_env.sh
danielerez: epel is required for p7zip: https://linuxconfig.org/how-to-install-p7zip-on-redhat-8
do u suggest install it manually instead? is it worth the effort:/

osherdp: I couldn't find any 7za usage
So our best option is to remove installation of epel and 7zip

danielerez: no luck. jq is also in epel, not sure if it's used in the image though.

osherdp: you're sure ``jq`` requires epel?
I did manage to remove epel installation and still install ``jq``

danielerez: tried 'skipper build assisted-service-build' without jq:
#5 9.056 No package jq available.

maybe you got jq from another repo?
try 'yum search jq'?

slaviered: jq is needed for the scripts

osherdp: Ah oops, you're right. It's required
Either way we can proceed with your solution if it will do no harm (and it seems to not harm anything)

osherdp (approved):

hack/setup_env.sh
slaviered: we can move the certificate update to the place in the Dockerfile where we bring in epel release. This way it will also affect previous calls to yum update

osherdp: The reason it's so confusing is that we're based on centos **7** in our dockerfile
We should figure out a way to solve it too, but that's another issue

PR #3474, opened 2022-03-13
NO-ISSUE: generate support rootless podman

celebdor:
When doing things like `make generate-python-client`, we instructed
docker to map the local user into the container, so that the files will
have the right ownership after the Docker part of the build.

In podman, when running it rootless (there's no reason why we'd ever
need to run those steps with root podman), root inside the container is
already the right user outside of the container, so we need to prevent
this setting from applying.

The commit also makes sure that if you have the docker command supported
with podman, it will detect rightly that it is podman you are using.

Yes, preferably by somebody using either docker and/or podman in Linux and Mac OS 
Yes

- [x] Enhancement 
- [x] CI/CD 

## How was this code tested?

It was tested on Fedora 35 with rootless podman with the `make generate-python-client`

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @osherdp 
/cc @eliorerz 
/cc @mkowalski 

## Checklist

I don't think it needs doc changes nor unit tests.

## Reviewers Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,9 @@ PWD = $(shell pwd)
 BUILD_FOLDER = $(PWD)/build/$(NAMESPACE)
 ROOT_DIR := $(or ${ROOT_DIR},$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST)))))
 CONTAINER_COMMAND := $(or ${CONTAINER_COMMAND},docker)
+ifeq ($(CONTAINER_COMMAND), docker)
+	CONTAINER_COMMAND = $(shell docker -v | cut -f1 -d' ' | tr '[:upper:]' '[:lower:]')
+endif
 TARGET := $(or ${TARGET},local)
 KUBECTL=kubectl -n $(NAMESPACE)
 
@@ -17,6 +20,9 @@ endif
 
 ifeq ($(CONTAINER_COMMAND), podman)
 	PUSH_FLAGS = --tls-verify=false
+	UID_FLAGS =
+else
+	UID_FLAGS = -u $(shell id -u):$(shell id -u)
 endif
 
 ASSISTED_ORG := $(or ${ASSISTED_ORG},quay.io/edge-infrastructure)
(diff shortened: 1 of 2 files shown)


hack/generate.sh
```diff
@@ -13,21 +13,21 @@ function lint_swagger() {
 
 function generate_go_server() {
     rm -rf restapi
-    docker run -u $(id -u):$(id -u) -v ${__root}:${__root}:rw,Z -v /etc/passwd:/etc/passwd -w ${__root} \
+    docker run $UID_FLAGS -v ${__root}:${__root}:rw,Z -v /etc/passwd:/etc/passwd -w ${__root} \
```
osherdp, line 16: I think we have another usage of this file outside of this makefile, so it will be a good idea to provide some default values in case ``UID_FLAGS`` is not defined

Makefile
```diff
@@ -3,6 +3,9 @@ PWD = $(shell pwd)
 BUILD_FOLDER = $(PWD)/build/$(NAMESPACE)
 ROOT_DIR := $(or ${ROOT_DIR},$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST)))))
 CONTAINER_COMMAND := $(or ${CONTAINER_COMMAND},docker)
+ifeq ($(CONTAINER_COMMAND), docker)
+	CONTAINER_COMMAND = $(shell docker -v | cut -f1 -d' ')
```
osherdp, line 7: hmm, doesn't work for me
```
make: Docker: Command not found
```

osherdp: also, why do we need it?

osherdp (changes requested):

hack/generate.sh
celebdor: that's a very good point! Thanks, @osherdp !

Makefile
celebdor: If you have `docker` pointing to podman, it otherwise thinks you are using docker, not podman. So this is a way to detect when you have podman posing as docker.

osherdp (approved):

osherdp:
/hold
/test e2e-metal-assisted-onprem
just to be extra-safe

PR #3482, opened 2022-03-15
MGMT-9443: tenancy based auth

slaviered:
The current authorization scheme in the SaaS grants read and write
access only to the cluster's owner. To align with the OCM authorization
scheme, we would like to allow the following:

- A tenancy-based read-only access to resources.
- Users should be allowed to grant editing capabilities to their
  resources to other users in their organization.

To establish a tenancy-based authorization, the auth middleware now
performs the following:

- 'Ownership' is now defined on an organization level (a user is allowed
   access if they belong to the same organization as the resource's creator).

- For Update and Delete of cluster-related APIs, the authorizer shall issue
   an additional query to AMS to check if the user is allowed to change the
   resource. This change supports the 'Cluster Editor' role assignment.

- If the API is related to a cluster-bound InfraEnv, the access permission is
  established by the bound cluster as described above.

- If the API is related to a non-bound InfraEnv, ownership for reading access
  is established on the organization level. Update and Delete operations are
  allowed only to the resource creator.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/auth/auth_utils.go
+++ b/pkg/auth/auth_utils.go
@@ -1,6 +1,8 @@
 package auth
 
 import (
+	"net/http"
+
 	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/strfmt"
 	"github.com/openshift/assisted-service/internal/common"
@@ -33,4 +35,14 @@ func shouldStorePayloadInCache(err error) bool {
 		return serr.StatusCode() < 500
 	}
 	return false
+}
+
+func isUpdateRequest(request *http.Request) bool {
+	return request != nil && (request.Method == http.MethodPost ||
+		request.Method == http.MethodPut ||
+		request.Method == http.MethodPatch)
+}
+
+func isDeleteRequest(request *http.Request) bool {
+	return request != nil && request.Method == http.MethodDelete
 }
--- a/pkg/auth/authz_handler.go
+++ b/pkg/auth/authz_handler.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/golang-jwt/jwt/v4"
@@ -19,6 +20,7 @@ import (
 )
 
 type AuthzHandler struct {
+	cfg     *Config
 	Enabled bool
 	log     logrus.FieldLogger
 	client  *ocm.Client
@@ -27,6 +29,7 @@ type AuthzHandler struct {
 
 func NewAuthzHandler(cfg *Config, ocmCLient *ocm.Client, log logrus.FieldLogger, db *gorm.DB) *AuthzHandler {
 	a := &AuthzHandler{
+		cfg:     cfg,
 		Enabled: cfg.AuthType == TypeRHSSO,
 		client:  ocmCLient,
 		log:     log,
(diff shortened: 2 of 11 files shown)


config/crd/bases/agent-install.openshift.io_agents.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.6.2
```
danielerez, line 7: no need for these changes in this pr...

pkg/auth/auth_utils.go
```diff
@@ -34,3 +36,10 @@ func shouldStorePayloadInCache(err error) bool {
 	}
 	return false
 }
+
+func isUpdateRequest(request *http.Request) bool {
+	return request != nil && (request.Method == http.MethodPost ||
+		request.Method == http.MethodPut ||
+		request.Method == http.MethodPatch ||
+		request.Method == http.MethodDelete)
```
danielerez, line 44: maybe we should handle delete separately. i.e. we can check against AMS whether delete action is permitted instead of update. just in case the role model would change.

pkg/auth/authz_handler.go
```diff
@@ -46,20 +47,124 @@ func (a *AuthzHandler) CreateAuthorizer() func(*http.Request) error {
...
+		return handleOwnershipQueryError(err)
+	}
+
+	return true, nil
+}
+
+func (a *AuthzHandler) isObjectOwnedByOrg(id string, obj interface{}, payload *ocm.AuthPayload) (bool, error) {
 	if a.db != nil {
-		err := a.db.First(obj, "id = ? and user_name = ?", id, username).Error
+		err := a.db.First(obj, "id = ? and org_id = ?", id, payload.Organization).Error
```
danielerez, line 78: maybe worth caching this as well. to avoid redundant calls to db.

pkg/auth/authz_handler.go
```diff
@@ -46,20 +47,124 @@ func (a *AuthzHandler) CreateAuthorizer() func(*http.Request) error {
...
+	} else {
+		return a.isObjectOwnedByUser(id, obj, payload)
+	}
+}
+
+func (a *AuthzHandler) hasSubscriptionAccess(clusterId string, payload *ocm.AuthPayload) (bool, error) {
+	if ownedBy, err := a.isObjectOwnedByUser(clusterId, &common.Cluster{}, payload); ownedBy {
+		return true, nil
+
+	} else if a.isTenancyEnabled() {
```
danielerez, line 97: no need for else here (returned in the prev if)

pkg/ocm/authorization.go
```diff
@@ -21,18 +22,34 @@ type authorization struct {
 }
 
 func (a authorization) AccessReview(ctx context.Context, username, action, resourceType string) (allowed bool, err error) {
+	return a.accessReview(ctx, username, "", action, resourceType, "")
+}
+
+func (a authorization) SubscriptionAccess(ctx context.Context, username, org, action, subscriptionId string) (allowed bool, err error) {
```
danielerez, line 28: can we keep this helper in authz_handler? as this is just a wrapper of AccessReview, it would probably be cleaner to keep this correlated to the OCM auth external API.
I.e. just add to current AccessReview func organizationId and subscriptionId as done in the API: https://api.openshift.com/?urls.primaryName=Authorization%20service#/default/post_api_authorizations_v1_access_review

pkg/auth/authz_handler.go
```diff
@@ -162,6 +269,11 @@ func (a *AuthzHandler) allowedToUseAssistedInstaller(username string) (bool, err
 		context.Background(), username, ocm.AMSActionCreate, ocm.BareMetalClusterResource)
 }
 
+func (a *AuthzHandler) hasClusterEditRole(payload *ocm.AuthPayload, subscriptionID string) (bool, error) {
```
danielerez, line 272: can we pass subscriptionID as UUID instead of string? should be nicer to get the string value when invoking the AccessReview API.

pkg/ocm/utils.go
```diff
@@ -13,6 +13,7 @@ import (
 const (
 	BareMetalClusterResource string = "BareMetalCluster"
 	AMSActionCreate          string = "create"
+	AMSActionUpdate          string = "update"
```
danielerez, line 16: we should consider also AMSActionDelete

pkg/auth/auth_utils.go
slaviered: I thought we agreed to have the update action for delete?

pkg/ocm/authorization.go
slaviered: But then I will have to change all the places for access review calls because the API will change (it needs more parameters).

config/crd/bases/agent-install.openshift.io_agents.yaml
slaviered: it's temporary problem in master that will be solved soon

subsystem/authz_test.go
```diff
@@ -100,7 +100,12 @@ var _ = Describe("test authorization", func() {
 		})
 	})
 
-	Context("regular user", func() {
+	Context("regular user - tenancy disabled", func() {
+		if Options.EnableOrgTenancy {
```
slaviered, line 104: These tests should succeed even under tenancy access rules. We should see what is wrong here

pkg/auth/authz_handler.go
```diff
@@ -46,20 +49,123 @@ func (a *AuthzHandler) CreateAuthorizer() func(*http.Request) error {
...
-		err := a.db.First(obj, "id = ? and user_name = ?", id, username).Error
+		err := a.db.First(obj, "id = ? and org_id = ?", id, payload.Organization).Error
+		return handleOwnershipQueryError(err)
+	}
+
+	return true, nil
+}
+
+func (a *AuthzHandler) hasOwnerAccess(id string, obj interface{}, payload *ocm.AuthPayload) (bool, error) {
+	if a.isTenancyEnabled() {
```
slaviered, line 87: perhaps here the best way is to ask first if object owned by user and only then own by org for backward comparability reasons

romfreiman:
Are we aligned with the UI on the implementation and will it play nicely with unified cluster list in OCM UI?

slaviered:
> Are we aligned with the UI on the implementation and will it play nicely with unified cluster list in OCM UI?

Yes on both accounts

danielerez:
> > Are we aligned with the UI on the implementation and will it play nicely with unified cluster list in OCM UI?
> 
> Yes on both accounts

Right, the OCM UI already has an 'Access control' tab for every cluster in the unified cluster list. So it allows to grant a ClusterEditor role for another user in the org. I.e. we can rely on that when querying AMS.

pkg/auth/authz_handler.go
```diff
@@ -46,20 +49,123 @@ func (a *AuthzHandler) CreateAuthorizer() func(*http.Request) error {
...
+	return true, nil
+}
+
+func (a *AuthzHandler) isObjectOwnedByUser(id string, obj interface{}, payload *ocm.AuthPayload) (bool, error) {
+	if a.db != nil {
+		err := a.db.First(obj, "id = ? and user_name = ?", id, payload.Username).Error
+		return handleOwnershipQueryError(err)
+	}
+
+	return true, nil
```
danielerez, line 74: If db is null we should return false and some error, maybe an early return first?

pkg/auth/authz_handler.go
```diff
@@ -46,20 +49,123 @@ func (a *AuthzHandler) CreateAuthorizer() func(*http.Request) error {
...
+}
+
+func (a *AuthzHandler) isObjectOwnedByOrg(id string, obj interface{}, payload *ocm.AuthPayload) (bool, error) {
 	if a.db != nil {
-		err := a.db.First(obj, "id = ? and user_name = ?", id, username).Error
+		err := a.db.First(obj, "id = ? and org_id = ?", id, payload.Organization).Error
+		return handleOwnershipQueryError(err)
+	}
+
+	return true, nil
```
danielerez, line 83: same

pkg/auth/authz_handler.go
danielerez: yeah, could be cleaner as tenancy is a specific use-case.

pkg/ocm/authorization.go
danielerez: We use it in only one place iinm, worth it for a small change:)

subsystem/authz_test.go
danielerez: Yes but there are a few tests that should fail in case of tenancy enabled. E.g. "can't get not owned cluster" - when tenancy is enabled a user can get not owned cluster if in same org. So need to go over these and change/add ones specifically for tenancy support.

pkg/auth/authz_handler.go
```diff
@@ -162,6 +270,11 @@ func (a *AuthzHandler) allowedToUseAssistedInstaller(username string) (bool, err
 		context.Background(), username, ocm.AMSActionCreate, ocm.BareMetalClusterResource)
 }
 
+func (a *AuthzHandler) hasClusterEditRole(payload *ocm.AuthPayload, subscriptionID string) (bool, error) {
+	return a.client.Authorization.SubscriptionAccess(
+		context.Background(), payload.Username, payload.Organization, ocm.AMSActionUpdate, subscriptionID)
```
danielerez, line 275: So after checking in AMS, seems the AccessReview API accepts an 'AMS id' for the organization (i.e. as for subscription). However, for each user we get in the payload the external_id of the organization, which is different. Any way, the organization ID is optional in this API as it's already available AMS database. Both accounts and subscriptions tables as org_id column, so it's redundant for checking perms. In short, we simply don't need to pass payload.Organization :)

pkg/auth/authz_handler.go
slaviered: No. This is a deliberate behavior of this layer to support some unit tests

pkg/auth/auth_utils.go
slaviered: fixed

pkg/ocm/authorization.go
slaviered: changed.

pkg/auth/authz_handler.go
slaviered: fixed

pkg/auth/authz_handler.go
slaviered: on a separate PR

pkg/auth/authz_handler.go
slaviered: All the APIs in this file are using strings.

danielerez (commented):
looks good. minor comments.

subsystem/authz_test.go
```diff
@@ -117,6 +117,11 @@ var _ = Describe("test authorization", func() {
 			Expect(err).ShouldNot(HaveOccurred())
 		})
 
+		It("with cluster editor role can get cluster", func() {
+			_, err := editclusterUserBMClient.Installer.V2DeregisterCluster(ctx, &installer.V2DeregisterClusterParams{ClusterID: userClusterID})
```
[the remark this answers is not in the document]
slaviered: it should. Will fix

subsystem/wiremock_stubs.go
```diff
@@ -81,6 +82,10 @@ func (w *WireMock) CreateWiremockStubsForOCM() error {
 		return err
 	}
 
+	if err := w.createStubsForClusterEditor(); err != nil {
```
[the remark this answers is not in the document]
slaviered: Yes. With the next version I am uploading you will not see that part

PR #3489, opened 2022-03-16
MGMTBUGSM-191: Race between agent and clusterDeployment reconciler causes day2 host installation to ignore requested hostname

ori-amizur:
In kube API, in agent controller, there is a function updateIfNeeded
that updates host record in DB.  There are severeal updates in this
function.  The first is to synchronize the approved flag with the spec.

When the clusterDeployment controller attempts to install day2 host, it
checks if it ready and if it is approved.  If the answer answer to both
is yes, then it starts installation.

There might be a situation that agent flag is set to true, but not all
host updates have been performed yet (hostname in our case).  In this
case the install-config will not reflect completely the host data after
update.

The solution is to move the update of te approved flag to be the last
operation in that function (after all the other updates).

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @yevgeny-shnaidman 
/cc @flaper87 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -1086,18 +1086,6 @@ func (r *AgentReconciler) updateIfNeeded(ctx context.Context, log logrus.FieldLo
 	spec := agent.Spec
 	var err error
 
-	if internalHost.Approved != spec.Approved {
-		err = r.Installer.UpdateHostApprovedInternal(ctx, internalHost.InfraEnvID.String(), agent.Name, spec.Approved)
-		if err != nil {
-			if errors.Is(err, gorm.ErrRecordNotFound) {
-				err = common.NewApiError(http.StatusNotFound, err)
-			}
-			log.WithError(err).Errorf("Failed to approve Agent")
-			return err
-		}
-		log.Infof("Updated Agent Approve %s %s", agent.Name, agent.Namespace)
-	}
-
 	err = r.updateInstallerArgs(ctx, log, internalHost, agent)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/controller/controllers/agent_controller.go
```diff
@@ -1086,18 +1086,6 @@ func (r *AgentReconciler) updateIfNeeded(ctx context.Context, log logrus.FieldLo
 	spec := agent.Spec
 	var err error
 
-	if internalHost.Approved != spec.Approved {
-		err = r.Installer.UpdateHostApprovedInternal(ctx, internalHost.InfraEnvID.String(), agent.Name, spec.Approved)
```
filanov, line 1090: I'm suggesting to return an finish the reconcile after this call. 
This way the reconcile will be short and no races will be done no?

ori-amizur: When should the host update be done ?

ori-amizur: The fix should verify that the approved flag is set not before all other host updates

filanov (approved):

filanov:
@ori-amizur please retitle the PR to bz so we can backport it later

ori-amizur:
/retitle Bug 2064622: Race between agent and clusterDeployment reconciler causes day2 host installation to ignore requested hostname

filanov:
@ori-amizur it was pushed to 2.5 so you don't need to backport

PR #3502, opened 2022-03-17
Bug 2065288: Use ignition 3.2 and allow for day2 overrides

mkowalski:
Currently there is a bug that causes day-2 worker ignition override to
be ignored without any error message.

This PR changes the behaviour of ignition generator so that the function
handling day-2 nodes takes the override into account. Given that there
are separate functions for day-1 and day-2 nodes, this PR also renames
the latter so that it's clear that it's used only for day-2.

This PR changes the ignition version from 3.1.0 to 3.2.0 and modifies
all the handlers to merge with backwards compatiblity. By doing so, we
allow customers to provide overrides in any version they wish as long as
it is compatible with the major ignition version used. Previously the
override version had to strictly match 3.1.0 used by us; with this
change, it is no longer the case.

Closes: [MGMT-9064](https://issues.redhat.com/browse/MGMT-9064)
Closes: Bug-2065288
Closes: [MGMTBUGSM-201](https://issues.redhat.com/browse/MGMTBUGSM-201)

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -737,9 +737,9 @@ func (b *bareMetalInventory) V2ImportClusterInternal(ctx context.Context, kubeKe
 	return &newCluster, nil
 }
 
-func (b *bareMetalInventory) createAndUploadNodeIgnition(ctx context.Context, cluster *common.Cluster, host *models.Host, ignitionEndpointToken string) error {
+func (b *bareMetalInventory) createAndUploadDay2NodeIgnition(ctx context.Context, cluster *common.Cluster, host *models.Host, ignitionEndpointToken string) error {
 	log := logutil.FromContext(ctx, b.log)
-	log.Infof("Starting createAndUploadNodeIgnition for cluster %s, host %s", cluster.ID, host.ID)
+	log.Infof("Starting createAndUploadDay2NodeIgnition for cluster %s, host %s", cluster.ID, host.ID)
 
 	// Specify ignition endpoint based on cluster configuration:
 	address := cluster.APIVip
@@ -762,16 +762,11 @@ func (b *bareMetalInventory) createAndUploadNodeIgnition(ctx context.Context, cl
 		caCert = cluster.IgnitionEndpoint.CaCertificate
 	}
 
-	ignitionBytes, err := b.IgnitionBuilder.FormatSecondDayWorkerIgnitionFile(ignitionEndpointUrl, caCert, ignitionEndpointToken)
+	fullIgnition, err := b.IgnitionBuilder.FormatSecondDayWorkerIgnitionFile(ignitionEndpointUrl, caCert, ignitionEndpointToken, host)
 	if err != nil {
-		return errors.Errorf("Failed to create ignition string for cluster %s", cluster.ID)
+		return errors.Wrapf(err, "Failed to create ignition string for cluster %s, host %s", cluster.ID, host.ID)
 	}
 
-	// Update host ignition hostname:
-	fullIgnition, err := ignition.SetHostnameForNodeIgnition(ignitionBytes, host)
-	if err != nil {
-		return errors.Errorf("Failed to create ignition string for cluster %s, host %s", cluster.ID, host.ID)
-	}
 	fileName := fmt.Sprintf("%s/worker-%s.ign", cluster.ID, host.ID)
 	log.Infof("Uploading ignition file <%s>", fileName)
 	err = b.objectHandler.Upload(ctx, fullIgnition, fileName)
@@ -1406,7 +1401,7 @@ func (b *bareMetalInventory) InstallSingleDay2HostInternal(ctx context.Context, 
 	}
 
 	// move host to installing
-	err = b.createAndUploadNodeIgnition(ctx, cluster, &h.Host, h.IgnitionEndpointToken)
+	err = b.createAndUploadDay2NodeIgnition(ctx, cluster, &h.Host, h.IgnitionEndpointToken)
 	if err != nil {
 		log.Errorf("Failed to upload ignition for host %s", h.RequestedHostname)
 		return err
@@ -1478,7 +1473,7 @@ func (b *bareMetalInventory) V2InstallHost(ctx context.Context, params installer
 	if cluster, err = common.GetClusterFromDB(b.db, *h.ClusterID, common.SkipEagerLoading); err != nil {
 		return common.GenerateErrorResponder(err)
 	}
-	err = b.createAndUploadNodeIgnition(ctx, cluster, h, host.IgnitionEndpointToken)
+	err = b.createAndUploadDay2NodeIgnition(ctx, cluster, h, host.IgnitionEndpointToken)
 	if err != nil {
 		log.Errorf("Failed to upload ignition for host %s", h.RequestedHostname)
 		return common.GenerateErrorResponder(err)
@@ -5066,6 +5061,8 @@ func (b *bareMetalInventory) V2UpdateHostIgnitionInternal(ctx context.Context, p
 			log.WithError(err).Errorf("Failed to parse host ignition config patch %s", params.HostIgnitionParams)
 			return nil, common.NewApiError(http.StatusBadRequest, err)
 		}
+	} else {
+		log.Infof("Trying to apply empty host ignition config to host %s in infra-env %s", params.HostID, params.InfraEnvID)
 	}
 
 	err = b.db.Model(&common.Host{}).Where(identity.AddUserFilter(ctx, "id = ? and infra_env_id = ?"), params.HostID, params.InfraEnvID).Update("ignition_config_overrides", params.HostIgnitionParams.Config).Error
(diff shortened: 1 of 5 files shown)


mkowalski:
/retitle [WIP] Bug 2065288: Don't ignore day-2 worker ignition override

mkowalski:
/retitle [WIP] Bug 2065288: Use ignition 3.2 and allow for day2 overrides

internal/bminventory/inventory.go
```diff
@@ -767,6 +767,17 @@ func (b *bareMetalInventory) createAndUploadNodeIgnition(ctx context.Context, cl
 		return errors.Errorf("Failed to create ignition string for cluster %s", cluster.ID)
 	}
 
+	// Apply overrides to the ignition
```
filanov, line 770: Minor comment but i don't think that this comment is needed, the code here is clear, personally i prefer comments that explain why and not how or what

internal/bminventory/inventory.go
```diff
@@ -5066,6 +5077,8 @@ func (b *bareMetalInventory) V2UpdateHostIgnitionInternal(ctx context.Context, p
 			log.WithError(err).Errorf("Failed to parse host ignition config patch %s", params.HostIgnitionParams)
 			return nil, common.NewApiError(http.StatusBadRequest, err)
 		}
+	} else {
+		log.Warnf("Trying to apply empty host ignition config to host %s in infra-env %s", params.HostID, params.InfraEnvID)
```
filanov, line 5081: What is the side affect here? 
What will happen in this case?

mkowalski: So what happens here is the following - as soon as you provide non-empty `params.HostIgnitionParams`, you enter this function. However, what we parse and apply is `params.HostIgnitionParams.Config`, meaning you have a potential field for confusion (that I hit when I was testing stuff). Imagine two scenarios

1) You have a host with ignition already overriden. You send a new override as `params.HostIgnitionParams` instead of `params.HostIgnitionParams.Config`. In reality, you have just overriden your previous ignition with an empty one. Not what was desired (and no warning on the way)

2) You have a host with no ignition. You send override as `params.HostIgnitionParams` instead of `params.HostIgnitionParams.Config`, as above. Once again, your override will not be applied.

The problem is that currently in both scenarios above, you will see the log message about successful update - `Custom discovery ignition config was applied to host %s in infra-env %s"`. This is confusing, because

* you see message saying override has been applied successfuly
* your host does not have any override

For this reason I decided to add a warning for a scenario when `params.HostIgnitionParams` exists but `params.HostIgnitionParams.Config` is empty.

eranco74: Seems like a valid use case (deleting an override) consider changing the log level to info.

internal/bminventory/inventory.go
```diff
@@ -767,6 +767,17 @@ func (b *bareMetalInventory) createAndUploadNodeIgnition(ctx context.Context, cl
 		return errors.Errorf("Failed to create ignition string for cluster %s", cluster.ID)
 	}
 
+	// Apply overrides to the ignition
+	if host.IgnitionConfigOverrides != "" {
+		var res string
+		res, err = ignition.MergeIgnitionConfig(ignitionBytes, []byte(host.IgnitionConfigOverrides))
+		if err != nil {
+			return errors.Errorf("Failed to apply ignition override for host %s", host.ID)
```
eranco74, line 775: return the actual error
```suggestion
			return errors.Wrap(err, "Failed to apply ignition override for host %s", host.ID)
```

internal/bminventory/inventory_test.go
```diff
@@ -8938,9 +8938,9 @@ var _ = Describe("UpdateInfraEnv - Ignition", func() {
 		Expect(response).To(BeAssignableToTypeOf(common.NewApiError(http.StatusBadRequest, errors.Errorf("error"))))
 	})
 
-	It("returns bad request when provided an old version", func() {
+	It("returns bad request when provided too recent version", func() {
```
eranco74, line 8941: Currently, the test validates that an override with an old ignition version will return a bad request, what is the purpose of this change?

internal/ignition/ignition.go
```diff
@@ -348,7 +346,7 @@ const discoveryIgnitionConfigFormat = `{
 
 const secondDayWorkerIgnitionFormat = `{
 	"ignition": {
-	  "version": "3.1.0",
+	  "version": "3.2.0",
```
eranco74, line 349: Will it work with 3.1.0 as well?
If not we are breaking API here.

eranco74:
Note that different RHCOS versions support diffrent ignition versions, since the assisted-service installs different OCP versions (that use different RHCOS) the ignition version it generates should match the ignition version generated by the installer of that OCP release.

internal/bminventory/inventory_test.go
mkowalski: What we do currently is we require a strict match for the ignition override, i.e. we hardcoded 3.1.0 in Assisted and this is the only override that works. So indeed, override with 3.0.0 is a non-working combination.

However, `ignition` library provides also a function "parse/merge with backwards-compatibility" which will merge any compatible version, so in such a scenario it can merge a template with 3.2.0 with override in 3.0.0 and result in a correct 3.2.0; for this reason this test is no longer valid.

internal/ignition/ignition.go
mkowalski: I did manually test it on a spoke with OCP 4.8; as in the other comment - providing an override with 3.1.0 does work because I'm not requiring a strict match, but allow for backwards-compatible ignition override

mkowalski:
> Note that different RHCOS versions support diffrent ignition versions, since the assisted-service installs different OCP versions (that use different RHCOS) the ignition version it generates should match the ignition version generated by the installer of that OCP release.

Right, this is still a bit of an unknown for me. Can you point me where and how it's done currently? I see everywhere that we just hardcode 3.1.0 and roll like this; there is no any check for the OCP version nor stuff like this

internal/bminventory/inventory.go
mkowalski: ACK

internal/bminventory/inventory.go
mkowalski: >Seems like a valid use case (deleting an override) consider changing the log level to info.

ACK

eranco74:
> I see everywhere that we just hardcode 3.1.0 and roll like this

Actually, it seems that we are assuming it's 3.2 and if we fail to parse the config we fall back to 3.1
https://github.com/openshift/assisted-service/blob/7587c104f0f6f378f20b8af082a926527778ab33/internal/ignition/ignition.go#L1081

eranco74:
> I see everywhere that we just hardcode 3.1.0 and roll like this
Seems like there is a bug [here](https://github.com/openshift/assisted-service/blob/7587c104f0f6f378f20b8af082a926527778ab33/internal/ignition/ignition.go#L1089) I don't know why we are setting the version to 3.1.0 after the config was translated to 3.2 :man_shrugging:

mkowalski:
Ah right... So if you look what I did with this function
```
config, _, err := config_latest.ParseCompatibleVersion(content)
```
it's that instead doing strict parsing I always try to merge those up to 3.2.0. The part with "if we fail to parse the config we fall back to 3.1" isn't about the OCP version but about the Ignition syntax and at the same time in the Ignition docs we can read

```
Any valid 3.1.0 configuration can be updated to a 3.2.0 configuration by simply changing the version string in the config
```

I just can't find a reason why we were all the time using `.Parse()` instead of `.ParseCompatibleVersion()` which simplifies a lot of flows we have. Any idea about that?

eranco74:
> there is no any check for the OCP version nor stuff like this
I think there is a test metrics that will check all versions the assisted-service support
/test ?

mkowalski:
/test e2e-metal-assisted-4-6

Infra error

mkowalski:
/retitle Bug 2065288: Use ignition 3.2 and allow for day2 overrides

mkowalski:
/retest e2e-metal-assisted-4-6

Cluster was indefinitely in `status pending-for-input`; no explicit errors anywhere in gather and this seems to be intermittently succeeding 🤷🏻

mkowalski:
I did a cross-check for ignition support in OCP 4.6; our job is using OCP [4.6.16](https://openshift-release.apps.ci.l2s4.p1.openshiftapps.com/releasestream/4-stable/release/4.6.16) which uses RHCOS [46.82.202101301821-0](https://releases-rhcos-art.cloud.privileged.psi.redhat.com/contents.html?stream=releases%2Frhcos-4.6&release=46.82.202203142206-0) that has Ignition [2.6.0](https://github.com/coreos/ignition/blob/main/NEWS#L162-L172). That mentions features present in 3.2.0, therefore I claim that it's safe to bump the version we use.

mkowalski:
/hold

OCP 4.6 does not support ignition 3.2.0; I will add a code patch that checks cluster version and uses 3.1.0 if OCP 4.6 is detected.

mkowalski:
/close

Superseded by #3524

PR #3508, opened 2022-03-18
NO-ISSUE: tweak unit tests to allow running PostgreSQL on k8s instead of docker

vrutkovs:
Instead of exposing docker socket when running unit-tests locally we could make use of k8s (deployed via cluster-bot or minikube). `kind` could work as well, but it doesn't seem to set External IPs for LB svc here.

- [x] Enhancement 

- [x] None

## How was this code tested?

- [x] No tests needed

## Assignees

/cc @flaper87 
/cc @paul-maidment 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/common_unitest_db.go
+++ b/internal/common/common_unitest_db.go
@@ -1,81 +1,341 @@
 package common
 
 import (
+	"context"
 	"fmt"
+	"log"
 	"os"
+	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/google/uuid"
 	. "github.com/onsi/gomega"
 	"github.com/ory/dockertest/v3"
 	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
+	k8s "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
 )
 
+/*
+  Database connection handling funcs. This supports running Postgres in:
+	* k8s cluster
+	* if k8s connection is not available try to run it as docker container
+
+	When SKIP_UT_DB env var is set localhost:5432 is being used.
+*/
 const (
 	dbDockerName  = "ut-postgres"
 	dbDefaultPort = "5432"
+
+	k8sNamespace = "assisted-installer"
 )
 
-type DBContext struct {
+/// DBContext is an interface for various DB implementations
+type DBContext interface {
+	GetHostPort() (string, string)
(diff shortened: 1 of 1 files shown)


internal/common/common_unitest_db.go
```diff
@@ -119,13 +342,33 @@ func DeleteTestDB(db *gorm.DB, dbName string) {
 	Expect(db.Error).ShouldNot(HaveOccurred())
 }
 
-func openTestDB(dbName string) (*gorm.DB, error) {
-	dsn := fmt.Sprintf("host=127.0.0.1 port=%s user=postgres password=admin sslmode=disable", gDbCtx.GetPort())
+func getDBDSN(dbName string) string {
+	var host, port string
+	if os.Getenv("SKIP_UT_DB") != "" {
```
omertuc, line 347: Could be nice to put this logic inside getDBContext()

internal/common/common_unitest_db.go
```diff
@@ -1,81 +1,304 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/wait"
+	k8s "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
 )
```
omertuc, line 29: A comment at the top of this file explaining the fallback hierarchy would be nice

deploy/assisted-image-service.yaml
```diff
@@ -1,6 +1,11 @@
 apiVersion: v1
 kind: List
 items:
+- kind: ServiceAccount
```
omertuc, line 4: Is this committed here by mistake?

vrutkovs: No, this SA is required by other Deployments in `deploy/`, so while its not required to pass unit tests, its required to pass `make deploy-test` on k8s

omertuc: Fair enough, but seems a bit unrelated to this PR. It's important we keep things distinct in case they need to be backported.

internal/common/common_unitest_db.go
vrutkovs: Ah, good idea. Created `NoDBContext`

internal/common/common_unitest_db.go
vrutkovs: done

deploy/assisted-image-service.yaml
omertuc: This is especially important because it's really annoying backporting things that already exist in master - you have to create a fake BZ for master without any PR, and you can't use the convenient cherry-pick robot for the process

vrutkovs: I don't think the whole PR is worth backporting - this is a dev-only feature, so I'd rather have all things in one PR for this.

However I don't mind breaking this up

omertuc: I meant it the other way around, that small change may be worth backporting and it will be a pain to do it if it's merged here

vrutkovs: good point - spinned it off to https://github.com/openshift/assisted-service/pull/3629

PR #3510, opened 2022-03-19
MGMT-9672: align manifest authz

slaviered:
The manifest handler implements its own access checks within the application layer. Those verifications are redundant and with the new tenancy policies also incorrect.

Since manifests are sub-objects of clusters, their access should is governed by the cluster access rules in the auth layer which should be the sole responsible for authorization checks.

Tests were changed accordingly and the case of missing cluster (which is handled by the auth middleware) was moved to the subsystem tests.

- [x] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/subsystem/manifests_test.go
+++ b/subsystem/manifests_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
+	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/openshift/assisted-service/client/installer"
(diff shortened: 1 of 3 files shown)


slaviered:
cc: @danielerez 
cc: @carbonin

subsystem/manifests_test.go
```diff
@@ -138,6 +139,44 @@ spec:
 		})
 	})
 
+	It("cluster doesn't exist", func() {
+		non_exiting_id := strfmt.UUID(uuid.New().String())
+		By("create manifest returns not found")
```
danielerez, line 144: for clarity, maybe better to put each of these in a separate By func. E.g as done in https://github.com/openshift/assisted-service/pull/3510/files#diff-6173872f98a7065d377afcb41777af74922ed6d692ac80254da515c6d3477d88L153

slaviered: It has the same effect

subsystem/manifests_test.go
```diff
@@ -138,6 +139,44 @@ spec:
...
+		_, err := userBMClient.Manifests.V2CreateClusterManifest(ctx, &manifests.V2CreateClusterManifestParams{
+			ClusterID: non_exiting_id,
+			CreateManifestParams: &models.CreateManifestParams{
+				Content:  &base64Content,
+				FileName: &manifestFile.FileName,
+				Folder:   &manifestFile.Folder,
+			},
+		})
+		Expect(err).To(BeAssignableToTypeOf(installer.NewGetInfraEnvNotFound()))
```
carbonin, line 153: I don't understand this. Why would we return an infraEnv error for a cluster-related API?

subsystem/manifests_test.go
carbonin: Yes, but the test output is more clear when they are separate. Also then you'll know if one or multiple are failing.
Currently if any expectation here fails then the rest of the tests don't get run.

carbonin (changes requested):
Please separate out the test cases into their own `It` blocks for separate endpoints.

subsystem/manifests_test.go
danielerez: oh right, it should probably be NewV2GetClusterNotFound

internal/manifests/manifests_test.go
```diff
@@ -198,20 +198,6 @@ var _ = Describe("ClusterManifestTests", func() {
 			Expect(responsePayload.Payload.Folder).To(Equal(defaultFolder))
 		})
 
-		It("cluster doesn't exist", func() {
```
carbonin, line 201: Why remove these tests? Shouldn't we still error in these cases?

slaviered: As I wrote in the PR description, the logic is not in the app layer anymore but in the middleware so those tests were moved to the subsystem test to make sure the system behaves as we expect when we depends on the authz handler

subsystem/manifests_test.go
slaviered: Yes. It was an copy paste mistake. Fixing

subsystem/manifests_test.go
slaviered: > Please separate out the test cases into their own `It` blocks for separate endpoints.
Each steps in the flow depends on the other (we first create, then list, and only at the end delete). This is why it has to be done within a single It (the order in a context is not guaranteed)

carbonin: If the calls succeeded then sure, but there is no dependency when every call errors.
And either way I would still suggest they be separate.

internal/manifests/manifests_test.go
slaviered: we have but not here but in the subsystem. These function will not be called if the cluster does not exist

carbonin: > These function will not be called if the cluster does not exist

What about cases where we're running without auth or in kube-api?
The authz middleware doesn't check ownership in those scenarios.
We should have a sane response from these functions even if we're not using rhsso auth.

carbonin: From @avishayt in a slack conversation:

> If the question is "do we need to make sure that the cluster exists?", I would say "nice to have but no".  In SaaS, if someone deleted the cluster then the authz won't allow them to view the manifests.  In k8s API, manifests are read from ConfigMaps and not from the backend anyway (the controller lists manifests just to check which ones it already created in the backend).

I guess this means that whether the cluster exists in the database is actually irrelevant (before and after this change). I suppose we could have a test for when the manifest doesn't exist in s3, but that doesn't seem super important.

subsystem/manifests_test.go
carbonin: FWIW I would still like these to be split.
There's no reason for them to be in one test and it should be pretty low effort to separate them

carbonin: I don't understand why this keeps getting resolved.
Am I missing something with how this test works?

It really feels like there's no reason that these should be in the same test case.

slaviered: test case are now separated to individual tests

danielerez:
> link

/hold

PR #3512, opened 2022-03-20
MGMT-9371: added owner query param to clusters/infraenvs requests

danielerez:
Added owner query param to the following GET list: clusters and infra-evns.
If the owner param is provided, the returned list includes only clusters/infra-envs that are owned by the specified user.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @slaviered 
/cc @avishayt 

## Checklist

## Reviewers Checklist

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -130,6 +130,11 @@ paths:
           type: boolean
           allowEmptyValue: true
           default: false
+        - in: query
+          name: owner
+          description: If provided, returns only clusters that are owned by the specified user.
+          type: string
+          required: false
       responses:
         "200":
           description: Success.
@@ -408,6 +413,11 @@ paths:
           description: If provided, returns only infra-envs which directly reference this cluster.
           type: string
           format: uuid
+          required: false
+        - in: query
+          name: owner
+          description: If provided, returns only infra-envs that are owned by the specified user.
+          type: string
           required: false
       responses:
         "200":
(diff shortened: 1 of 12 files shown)


swagger.yaml
```diff
@@ -130,6 +130,12 @@ paths:
           type: boolean
           allowEmptyValue: true
           default: false
+        - in: query
+          name: filter_by_owner
+          description: Whether to return only clusters that are owned by the user.
+          type: boolean
```
avishayt, line 136: Maybe it would be better to make the filter generic so that a user can query someone else's resources (in the same org of course)?
```suggestion
          name: owner
          description: Whether to return only clusters that are owned by the user.
          type: string
```

danielerez: good idea, added.

slaviered: Also from the UI point of view it will have to fetch the current user ID which I am not sure it has

danielerez: I've added also organization to the filter, so it's enough as it affects only GET requests.

slaviered: We will not be able to query other organization anyhow so no need to have organization as well.

danielerez: Yeah the org in db request is just for filtering the list.

PR #3547, opened 2022-03-24
MGMT-9404: remove user filter from UpdateHostApproved

danielerez:
UpdateHostApprovedInternal is currently filtering hosts by user_name when the entity is fetched.
Since this func is used exclusively in kube-api [flow](https://github.com/openshift/assisted-service/blob/f28bddc8d95a89a4804b77b362dda721e3e14214/internal/controller/controllers/agent_controller.go#L1175), the filter has no effect (kube-api uses kube_key_namespace to [retrieve](https://github.com/openshift/assisted-service/blob/f28bddc8d95a89a4804b77b362dda721e3e14214/internal/controller/controllers/agent_controller.go#L144) the host). Hence, we can safely remove AddUserFilter from the func.

Note: perhaps we should consider marking UpdateHostApprovedInternal as relevant only for kube-api. I.e. to avoid missing authorization if ever exposed to the external API.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @slaviered 
/cc @carbonin 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -3867,6 +3867,8 @@ func (b *bareMetalInventory) GetCommonHostInternal(_ context.Context, infraEnvId
 	return common.GetHostFromDB(b.db, infraEnvId, hostId)
 }
 
+// Updates host's approved field by a specified flag.
+// Used execlusively by kube-api.
 func (b *bareMetalInventory) UpdateHostApprovedInternal(ctx context.Context, infraEnvId, hostId string, approved bool) error {
 	log := logutil.FromContext(ctx, b.log)
 	log.Infof("Updating Approved to %t Host %s InfraEnv %s", approved, hostId, infraEnvId)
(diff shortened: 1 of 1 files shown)


carbonin:
I mentioned this in our meeting, but now thinking about it more seriously.

Do you want to take this opportunity to put just this function in a new file? Something like `internal/bminventory/controller_helpers.go`? That would hopefully do enough to communicate that it's only being used by the controllers.

Also you could add a function header comment.

danielerez:
> I mentioned this in our meeting, but now thinking about it more seriously.
> 
> Do you want to take this opportunity to put just this function in a new file? Something like `internal/bminventory/controller_helpers.go`? That would hopefully do enough to communicate that it's only being used by the controllers.

Yeah good idea, just need to see how much effort it'd take.

> 
> Also you could add a function header comment.

I'll fallback to that option if needed:)

carbonin:
A new file in the same package should be no effort at all.

danielerez:
> A new file in the same package should be no effort at all.

Right, but seems there are quite a few functions that worth moving. Created a ticket to tackle separately: https://issues.redhat.com/browse/MGMT-9735

carbonin:
Let's add a comment then for now.

PR #3550, opened 2022-03-24
MGMT-9673: Ownerhsip API to authz

slaviered:
Authorization becomes more complex with the tenancy based feature and more authorization scheme added for other projects.

In the current implementation the application layer is using stateless utilities to check for access permission to object on top of the existing mechanism. These checks are sometimes redundant and or not quite aligned with new authorization schemes.

Using stateless utilities make us expose feature flags and states in the auth API where they do not belong. For all those reasons an API for authorization that will be shared by all authorization handlers and expose a coherent access control algorithms to the application is required.

This is the first installment of the enhancement. It defines The `OwnedBy` interface that wraps the db object with authorization-related conditions based on the active authorization layer.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -179,6 +179,7 @@ type bareMetalInventory struct {
 	operatorManagerApi   operators.API
 	generator            generator.ISOInstallConfigGenerator
 	authHandler          auth.Authenticator
+	authzHandler         auth.Authorizer
 	k8sClient            k8sclient.K8SClient
 	ocmClient            *ocm.Client
 	leaderElector        leader.Leader
@@ -208,6 +209,7 @@ func NewBareMetalInventory(
 	usageApi usage.API,
 	operatorManagerApi operators.API,
 	authHandler auth.Authenticator,
+	authzHandler auth.Authorizer,
 	k8sClient k8sclient.K8SClient,
 	ocmClient *ocm.Client,
 	leaderElector leader.Leader,
@@ -238,6 +240,7 @@ func NewBareMetalInventory(
 		usageApi:             usageApi,
 		operatorManagerApi:   operatorManagerApi,
 		authHandler:          authHandler,
+		authzHandler:         authzHandler,
 		k8sClient:            k8sClient,
 		ocmClient:            ocmClient,
 		leaderElector:        leaderElector,
@@ -2652,31 +2655,29 @@ func (b *bareMetalInventory) calculateHostNetworks(log logrus.FieldLogger, clust
 func (b *bareMetalInventory) listClustersInternal(ctx context.Context, params installer.V2ListClustersParams) ([]*models.Cluster, error) {
 	log := logutil.FromContext(ctx, b.log)
 	db := b.db
+
+	var dbClusters []*common.Cluster
+	var clusters []*models.Cluster
+
 	if swag.BoolValue(params.GetUnregisteredClusters) {
 		if !identity.IsAdmin(ctx) {
 			return nil, common.NewApiError(http.StatusForbidden, errors.New("only admin users are allowed to get unregistered clusters"))
 		}
 		db = db.Unscoped()
 	}
-	var dbClusters []*common.Cluster
-	var clusters []*models.Cluster
-	whereCondition := make([]string, 0)
 
-	filterByOrg := b.authHandler.EnableOrgTenancy()
-	if user := identity.AddOwnerFilter(ctx, "", filterByOrg, swag.StringValue(params.Owner)); user != "" {
-		whereCondition = append(whereCondition, user)
-	}
+	db = b.authzHandler.OwnedByUser(ctx, db, swag.StringValue(params.Owner))
 
 	if params.OpenshiftClusterID != nil {
-		whereCondition = append(whereCondition, fmt.Sprintf("openshift_cluster_id = '%s'", *params.OpenshiftClusterID))
+		db = db.Where("openshift_cluster_id = ?", *params.OpenshiftClusterID)
 	}
 
 	if len(params.AmsSubscriptionIds) > 0 {
-		whereCondition = append(whereCondition, fmt.Sprintf("ams_subscription_id IN %s", common.ToSqlList(params.AmsSubscriptionIds)))
+		db = db.Where("ams_subscription_id IN (?)", params.AmsSubscriptionIds)
 	}
 
 	dbClusters, err := common.GetClustersFromDBWhere(db, common.UseEagerLoading,
-		common.DeleteRecordsState(swag.BoolValue(params.GetUnregisteredClusters)), strings.Join(whereCondition, " AND "))
+		common.DeleteRecordsState(swag.BoolValue(params.GetUnregisteredClusters)))
 	if err != nil {
 		log.WithError(err).Error("Failed to list clusters in db")
 		return nil, common.NewApiError(http.StatusInternalServerError, err)
@@ -3950,14 +3951,13 @@ func (b *bareMetalInventory) ListInfraEnvs(ctx context.Context, params installer
 	var dbInfraEnvs []*common.InfraEnv
 	var infraEnvs []*models.InfraEnv
 
-	condition := ""
-	if params.ClusterID != nil {
-		condition = fmt.Sprintf("cluster_id = '%s'", params.ClusterID)
-	}
-	filterByOrg := b.authHandler.EnableOrgTenancy()
-	whereCondition := identity.AddOwnerFilter(ctx, condition, filterByOrg, swag.StringValue(params.Owner))
+	db = b.authzHandler.OwnedByUser(ctx, db, swag.StringValue(params.Owner))
 
-	dbInfraEnvs, err := common.GetInfraEnvsFromDBWhere(db, whereCondition)
+	if params.ClusterID != nil {
+		db = db.Where("cluster_id = ?", params.ClusterID)
+	}
+
+	dbInfraEnvs, err := common.GetInfraEnvsFromDBWhere(db)
 	if err != nil {
 		log.WithError(err).Error("Failed to list infraEnvs in db")
 		return common.NewApiError(http.StatusInternalServerError, err)
--- a/pkg/auth/authz.go
+++ b/pkg/auth/authz.go
@@ -1,0 +1,38 @@
+package auth
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/openshift/assisted-service/pkg/ocm"
+	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+type Authorizer interface {
+	/* Limits the database query to access records that are owned by the current user,
+	 * according to the configured access policy.
+	 */
+	OwnedBy(ctx context.Context, db *gorm.DB) *gorm.DB
+
+	/* Limits the database query to access records owned only by the input user,
+	 * regardless of the configured access policy. If user-based authentication
+	 * is not supported, the function effectively will not limit access.
+	 */
+	OwnedByUser(ctx context.Context, db *gorm.DB, username string) *gorm.DB
+
+	/* Provides the middleware authorization algorithm */
+	CreateAuthorizer() func(*http.Request) error
+}
+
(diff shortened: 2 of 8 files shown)


internal/bminventory/inventory.go
```diff
@@ -2787,31 +2790,35 @@ func (b *bareMetalInventory) calculateHostNetworks(log logrus.FieldLogger, clust
 func (b *bareMetalInventory) listClustersInternal(ctx context.Context, params installer.V2ListClustersParams) ([]*models.Cluster, error) {
 	log := logutil.FromContext(ctx, b.log)
 	db := b.db
+
+	var dbClusters []*common.Cluster
+	var clusters []*models.Cluster
+
+	//this call should be before OwnedBy
+	if params.Owner != nil {
```
danielerez, line 2798: maybe worth adding it to the OwnedBy API? i.e. could be cleaner as a simple arg in signature.

internal/bminventory/inventory.go
```diff
@@ -2787,31 +2790,35 @@ func (b *bareMetalInventory) calculateHostNetworks(log logrus.FieldLogger, clust
...
+		ctxparams.SetParam(ctx, auth.X_USER_SCOPE, params.Owner)
+	}
+
 	if swag.BoolValue(params.GetUnregisteredClusters) {
+		//bypass for the GC mechanism
```
danielerez, line 2803: GUC?:)

pkg/auth/none_authz_handler.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func (*NoneHandler) CreateAuthorizer() func(*http.Request) error {
+	return func(*http.Request) error {
+		return nil
+	}
+}
+
+func (*NoneHandler) OwnedBy(ctx context.Context, db *gorm.DB) *gorm.DB {
+	if identity.IsAdmin(ctx) {
```
danielerez, line 27: the tests for this logic is in authz_handler_test, right?

pkg/auth/authz_handler.go
```diff
@@ -19,34 +20,19 @@ import (
 	"gorm.io/gorm"
 )
 
+/* AuthzHandler is the authorizer middleware that is being used for
+ * RHSSO authentication cases and work with AMS to ensure tenancy
+ * and user based access policies
+ */
 type AuthzHandler struct {
```
danielerez, line 27: maybe worth renaming to RHSSOAuthorizer, i.e. to be consistent with RHSSOAuthenticator (also the file)

internal/bminventory/inventory.go
slaviered: Garbage Collection (This is part of the api that servers only the GC server)

pkg/auth/none_authz_handler.go
slaviered: Yes

internal/bminventory/inventory.go
slaviered: We should keep the API as generic as possible. I am removing this assignment based on the design discussion with @avishayt

danielerez:
/hold
waiting for tests

internal/bminventory/inventory.go
```diff
@@ -2787,31 +2790,33 @@ func (b *bareMetalInventory) calculateHostNetworks(log logrus.FieldLogger, clust
...
-	whereCondition := make([]string, 0)
 
-	filterByOrg := b.authHandler.EnableOrgTenancy()
-	if user := identity.AddOwnerFilter(ctx, "", filterByOrg, swag.StringValue(params.Owner)); user != "" {
-		whereCondition = append(whereCondition, user)
+	db = b.authzHandler.OwnedBy(ctx, db)
+
+	if params.Owner != nil {
+		db = db.Where("user_name = ?", params.Owner)
 	}
```
carbonin, line 2808: This is probably worth a comment.
Just reading these lines makes it feel like we don't need to include the additional `params.Owner` bit.
But in this case, `OwnedBy` is not actually looking for a specific owner, right?

pkg/auth/authz.go
```diff
@@ -0,0 +1,30 @@
...
+	"gorm.io/gorm"
+)
+
+type Ownership interface {
+	OwnedBy(ctx context.Context, db *gorm.DB) *gorm.DB
+}
+type Authorizer interface {
+	Ownership
+	CreateAuthorizer() func(*http.Request) error
+}
```
carbonin, line 18: Why separate these?

pkg/auth/rhsso_authz_handler.go
```diff
@@ -179,17 +177,17 @@ func (a *AuthzHandler) checkInfraEnvBasedAccess(id string, request *http.Request
 	return a.isObjectOwnedByUser(id, &common.InfraEnv{}, payload)
 }
 
-func (a *AuthzHandler) Authorizer(request *http.Request) error {
+func (a *AuthzHandler) AuthorizerFunc(request *http.Request) error {
```
carbonin, line 180: Why change these names?

pkg/auth/rhsso_authz_handler_test.go
```diff
@@ -93,6 +97,129 @@ var _ = Describe("Authz email domain", func() {
 	}
 })
 
+var _ = Describe("autz_ownership_api", func() {
```
carbonin, line 100: Typo, should be `authz_ownership_api`, but also I think it would be more correct to do `Describe("OwnedBy")` here as that's the function under test.

pkg/auth/rhsso_authz_handler_test.go
```diff
@@ -93,6 +97,129 @@ var _ = Describe("Authz email domain", func() {
...
+				payload := &ocm.AuthPayload{}
+				payload.Role = ocm.AdminRole
+				ctx = context.WithValue(ctx, restapi.AuthKey, payload)
+
+				sql := db.ToSQL(func(tx *gorm.DB) *gorm.DB {
+					return handler.OwnedBy(ctx, tx).First(&models.Cluster{})
+				})
+				Expect(sql).NotTo(ContainSubstring("id = "))
+				Expect(sql).NotTo(ContainSubstring("user_name"))
+				Expect(sql).NotTo(ContainSubstring("org_id"))
```
carbonin, line 132: We have a connection to the db. Why not actually create the rows and ensure you get the right ones back rather than doing string comparisons on the sql?

carbonin (changes requested):

internal/bminventory/inventory.go
slaviered: it is. We had a lengthy discussion about this implementation today with @avishayt. I will summarize it in a comment

pkg/auth/authz.go
slaviered: For clarity

pkg/auth/none_authz_handler.go
```diff
@@ -0,0 +1,32 @@
...
+		return db
+	}
+	username := ocm.UserNameFromContext(ctx)
+	return db.Where("user_name = ?", username)
```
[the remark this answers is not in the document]
slaviered: It actually the same behavior we had beforehand with AddUserFilter but you are right. Now that we have a brand new implementation it is totally redundant as the user_name column would be empty

pkg/auth/rhsso_authz_handler.go
slaviered: Authorizer is now the interface for the entire service so this name is overloaded. After the change I found the old name confusing, so I wanted to convey that this is just the authorizer callback function for the resapi middleware. If we you have a better name please suggest it.

pkg/auth/authz.go
carbonin: I don't think it helps anything unless there's an actual use for the second interface, just keep one please.

internal/bminventory/inventory.go
carbonin: Maybe put another way, it _may_ not be looking for a specific owner.
I'm guessing the reason this exists is that because even if tenancy is enabled we might only want to see clusters owned by a specific user, right?

I would definitely add a comment or maybe add another paramter to `OwnedBy` to indicate that we want to filter by a specific user even if tenancy is enabled.

pkg/auth/rhsso_authz_handler.go
carbonin: I would only use `Func` in a function name if we're returning a function type, which this isn't.

I think `Authorize` would be better since this is actually doing the work of authorizing the request.
Also please remove `Func` from the other functions as well.

pkg/auth/authz.go
slaviered: Fixed

pkg/auth/rhsso_authz_handler_test.go
slaviered: Replaced by real test data and assertions on the retrieved records

pkg/auth/rhsso_authz_handler.go
```diff
@@ -65,6 +51,30 @@ func handleOwnershipQueryError(err error) (bool, error) {
...
 }
 
+func (a *AuthzHandler) OwnedBy(ctx context.Context, db *gorm.DB) *gorm.DB {
+	if identity.IsAdmin(ctx) {
+		return db
+	}
+	if a.isTenancyEnabled() {
+		return db.Where("org_id = ?", ocm.OrgIDFromContext(ctx))
+	} else {
+		username := ocm.UserNameFromContext(ctx)
```
danielerez, line 61: nit, no need for the username var

pkg/auth/rhsso_authz_handler.go
```diff
@@ -65,6 +51,30 @@ func handleOwnershipQueryError(err error) (bool, error) {
 	return true, nil
 }
 
+func (a *AuthzHandler) OwnedBy(ctx context.Context, db *gorm.DB) *gorm.DB {
+	if identity.IsAdmin(ctx) {
+		return db
+	}
+	if a.isTenancyEnabled() {
+		return db.Where("org_id = ?", ocm.OrgIDFromContext(ctx))
+	} else {
```
danielerez, line 60: no need for the 'else'

slaviered: It was actually written like that on purpose so it is clear that the user filtering is not a common code

pkg/auth/rhsso_authz_handler.go
slaviered: OK. removing the assignment

pkg/auth/rhsso_authz_handler.go
```diff
@@ -179,7 +188,7 @@ func (a *AuthzHandler) checkInfraEnvBasedAccess(id string, request *http.Request
 	return a.isObjectOwnedByUser(id, &common.InfraEnv{}, payload)
 }
 
-func (a *AuthzHandler) Authorizer(request *http.Request) error {
+func (a *AuthzHandler) authorizerMiddleware(request *http.Request) error {
```
carbonin, line 191: This isn't a middleware, but it's fine for just deduplicating the name for now.

pkg/auth/rhsso_authz_handler_test.go
```diff
@@ -93,6 +97,187 @@ var _ = Describe("Authz email domain", func() {
 	}
 })
 
+var _ = Describe("OwnedBy", func() {
+	var (
+		db      *gorm.DB
+		dbName  string
+		ctx     context.Context
+		handler Authorizer
+		records []common.Cluster
```
carbonin, line 106: I think passing the records to validate to the `AllRecords*` functions would be better.
Sharing this variable across tests could lead to contamination.

You can do this in a followup or when it becomes a problem.

carbonin (approved):

pkg/auth/rhsso_authz_handler_test.go
slaviered: very well. I localized it.

PR #3556, opened 2022-03-24
MGMT-7365: Remove single network configuration fields

mkowalski:
During the dual-stack implementation the new structures have been
introduced - machine_networks, cluster_networks and service_networks
(all of them of type []string). In order to preserve backwards
compatibility, there were wrappers created that translate e.g.
machine_network_cidr (of type string) to machine_networks (of type
[]string).

Usage of the old parameters makes it impossible to pass multiple
networks (because of type mismatch). Given that in API v2 the new
structures ([]string) are fully implemented, there is no need to support
old ones any more.

Closes: [MGMT-7365](https://issues.redhat.com/browse/MGMT-7365)

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @nmagnezi 

## Checklist

## Reviewers Checklist


internal/migrations/20210822134659_multiple_networks.go
```diff
@@ -1,66 +0,0 @@
-package migrations
-
-import (
-	gormigrate "github.com/go-gormigrate/gormigrate/v2"
-	"github.com/openshift/assisted-service/internal/common"
-	"github.com/openshift/assisted-service/models"
-	"gorm.io/gorm"
-)
-
-func multipleNetworks() *gormigrate.Migration {
```
carbonin, line 10: So theoretically we should keep this migration.
Could there be some installation out there that still hasn't run this migration, but will still want to move the cluster data from the old columns to the new tables when they get around to upgrading?

Specifically I wonder if ACM 2.3 is still supported and if people can upgrade directly from 2.3 to 2.6 or something.
I don't know if this migration pre-dates 2.3, but the idea is that it should work effectively forever.

internal/migrations/20220324203547_multiple_networks_cleanup.go
```diff
@@ -0,0 +1,46 @@
...
+		err = tx.Migrator().DropColumn(&clusters, "service_network_cidr")
+		if err != nil {
+			log.Printf("Non-critical error: %s", err)
+		}
+
+		return nil
+	}
+
+	rollback := func(tx *gorm.DB) error {
+		// As the migration never returns any errors, we are not implementing rollback.
```
carbonin, line 37: So again, theoretically this should add the columns back, but if we need to rollback we'd probably deploy an image earlier than this change which should automigrate the columns back into existence, right?

mkowalski:
/hold
Will redesign so that we don't remove columns from schema

PR #3562, opened 2022-03-25
Bug 2065034: Correct ignition validation message for day 2 host

CrystalChun:
[Bug 2065034](https://bugzilla.redhat.com/show_bug.cgi?id=2065034)
Originally when a day 2 host is added to a cluster, the ignition
validation will fail and show the message that the ignition is not
downloadable, which is not true and will confuse the user.

This change will show validation as pending when the API VIP connectivity
in newly added day 2 hosts is empty and only shows the failure
when the connection is not successful. It also modifies the pending
message to reflect this.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Manual test:
1. Used the assisted-test-infra to set up the environment and deploy day 2 node
    ````
        TEST_TEARDOWN=no TEST=./src/tests/test_day2.py TEST_FUNC=test_deploy_day2_nodes_cloud $(MAKE) test;
    ````
3. Added debug logging to see when API VIP connectivity failed while adding day 2 node
    - All hosts eventually move [past this error](https://github.com/openshift/assisted-service/blob/27ed29f80418d11952c8862ce5005771c11a308f/internal/host/validator.go#L710)
    - With correct api vip: the response is success
    - With incorrect api vip (set it to `111.111.111.111` [here](https://github.com/openshift/assisted-test-infra/blob/67e4a3f125b78b18245d90930c3f8ae26fb5e8a7/src/assisted_test_infra/test_infra/helper_classes/day2_cluster.py#L35)): the response is fail

## Assignees

/cc @carbonin 
/cc @danielerez 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -700,16 +700,17 @@ func (v *validator) isIgnitionDownloadable(c *validationContext) ValidationStatu
 	if c.infraEnv != nil {
 		return ValidationSuccessSuppressOutput
 	}
-	if c.inventory == nil {
-		return ValidationPending
-	}
 	if !hostutil.IsDay2Host(c.host) || swag.BoolValue(c.cluster.UserManagedNetworking) {
 		return ValidationSuccessSuppressOutput
+	}
+	if c.host.APIVipConnectivity == "" {
+		return ValidationPending
 	}
 	var response models.APIVipConnectivityResponse
 	if err := json.Unmarshal([]byte(c.host.APIVipConnectivity), &response); err != nil {
 		return ValidationFailure
 	}
+
 	return boolValue(response.IsSuccess)
 }
 
@@ -721,9 +722,9 @@ func (v *validator) printIgnitionDownloadable(c *validationContext, status Valid
 		}
 		return "Ignition is downloadable"
 	case ValidationFailure:
-		return "Ignition is not downloadable"
+		return "Ignition is not downloadable. Please ensure host connectivity to the cluster's API VIP."
 	case ValidationPending:
-		return "Missing inventory"
+		return "Ignition is not ready, pending API VIP connectivity."
 	default:
 		return fmt.Sprintf("Unexpected status %s", status)
 	}
--- a/internal/host/validations_test.go
+++ b/internal/host/validations_test.go
@@ -87,50 +87,134 @@ var _ = Describe("Validations test", func() {
 		mockAndRefreshStatusWithoutEvents(h)
 	}
 
-	Context("Disk encryption validation", func() {
-		getDiskEncryptionValidationResult := func(validationsInfo string) (ValidationStatus, string, bool) {
+	getDay2Host := func() models.Host {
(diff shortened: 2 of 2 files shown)


internal/host/validator.go
```diff
@@ -708,8 +708,9 @@ func (v *validator) isIgnitionDownloadable(c *validationContext) ValidationStatu
 	}
 	var response models.APIVipConnectivityResponse
 	if err := json.Unmarshal([]byte(c.host.APIVipConnectivity), &response); err != nil {
```
carbonin, line 710: What is actually in the field at this point?
It seems strange that this would fail to unmarshal rather than just having the validation fail.

Or even better, shouldn't we be able to detect a pending state in some better way than failing to parse the json?

CrystalChun: The `err` is `unexpected end of JSON input` and in the `response` object is `{ false}`

Do you think it's best not to assume that this is pending validation?

carbonin: `{ false}` looks like the default value of the response struct.
What is `c.host.APIVipConnectivity` in that case?

filanov (commented):
no tests?

internal/host/validator.go
carbonin: Okay, maybe it makes sense then to explicitly check for empty?
That way we don't confuse an expected negative with an error (malformed json).

internal/host/validator.go
```diff
@@ -723,7 +724,7 @@ func (v *validator) printIgnitionDownloadable(c *validationContext, status Valid
 	case ValidationFailure:
 		return "Ignition is not downloadable"
 	case ValidationPending:
-		return "Missing inventory"
+		return "Ignition is not ready, pending API VIP connectivity or missing inventory"
```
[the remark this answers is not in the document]
CrystalChun: Sure! For my own knowledge, what was the original reason the inventory check was added in this function?

internal/host/validator.go
CrystalChun: Good idea! Modified, how does it look?

internal/host/validator.go
```diff
@@ -700,16 +700,17 @@ func (v *validator) isIgnitionDownloadable(c *validationContext) ValidationStatu
 	if c.infraEnv != nil {
 		return ValidationSuccessSuppressOutput
 	}
-	if c.inventory == nil {
-		return ValidationPending
-	}
 	if !hostutil.IsDay2Host(c.host) || swag.BoolValue(c.cluster.UserManagedNetworking) {
 		return ValidationSuccessSuppressOutput
 	}
+	if c.host.APIVipConnectivity == "" {
```
filanov, line 706: Can you please add tests?

CrystalChun: Sorry missed the earlier comment! Will do!

filanov (approved):

PR #3589, opened 2022-03-29
AGENT-34: Set local host to be bootstrap

zaneb:
In the ephemeral agent-based installer image, the host running
assisted-service must become the bootstrap host, so that the
assisted-service remains up until all of the other hosts are installed.

By writing the agent ID to a local file (in openshift/assisted-installer-agent#314), we allow assisted-service to
determine when a host registration is for the local host.

- [ ] New Feature 
- [X] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [X] Ephemeral installer
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @avishayt
/cc @filanov

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -89,6 +89,7 @@ type Config struct {
 	AgentDockerImg                  string            `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/edge-infrastructure/assisted-installer-agent:latest"`
 	ServiceBaseURL                  string            `envconfig:"SERVICE_BASE_URL"`
 	ImageServiceBaseURL             string            `envconfig:"IMAGE_SERVICE_BASE_URL"`
+	InEphemeralInstaller            bool              `envconfig:"EPHEMERAL_INSTALLER" default:"false"`
 	ServiceCACertPath               string            `envconfig:"SERVICE_CA_CERT_PATH" default:""`
 	S3EndpointURL                   string            `envconfig:"S3_ENDPOINT_URL" default:"http://10.35.59.36:30925"`
 	S3Bucket                        string            `envconfig:"S3_BUCKET" default:"test"`
@@ -4518,7 +4519,7 @@ func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installe
 
 	// When the assisted service is running on the same host as the agent, this
 	// host must become the bootstrap so that it installs last
-	if hostutil.IsAssistedServiceHost(host) {
+	if b.InEphemeralInstaller && hostutil.IsAssistedServiceHost(host) {
 		log.Infof("host <%s> also hosts assisted-service", params.NewHostParams.HostID.String())
 		host.Role = models.HostRoleMaster
 		host.Bootstrap = true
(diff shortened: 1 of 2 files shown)


eranco74 (commented):
Looks good, 
Where is the agent-based deployment?
If it's in the assisted-service repo please update it to mount the  localHostIDFile to the assisted-service container.

celebdor:
@eranco74 the self-hosting agent based deployment is currently at https://github.com/openshift-agent-team/fleeting . The accompanying PR is: https://github.com/openshift-agent-team/fleeting/pull/16

internal/bminventory/inventory.go
```diff
@@ -4516,6 +4517,14 @@ func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installe
 		c = &cluster.Cluster
 	}
 
+	// When the assisted service is running on the same host as the agent, this
+	// host must become the bootstrap so that it installs last
+	if b.InEphemeralInstaller && hostutil.IsAssistedServiceHost(host) {
+		log.Infof("host <%s> also hosts assisted-service", params.NewHostParams.HostID.String())
+		host.Role = models.HostRoleMaster
+		host.Bootstrap = true
+	}
```
mkowalski, line 4526: Could we get here some handling for a scenario where more than one host tries to become the bootstrap? I can see that this is determined by the existence of a file in a local filesystem (`/var/run/assisted-agent/host-id`), so this looks like a potential for 2 hosts in the infra to get in a race. Or not? If not, then maybe a comment in the function below `IsAssistedServiceHost()`

internal/host/hostutil/host_utils.go
```diff
@@ -154,6 +156,11 @@ func IgnitionFileName(host *models.Host) string {
 	return fmt.Sprintf("%s-%s.ign", common.GetEffectiveRole(host), host.ID)
 }
 
+func IsAssistedServiceHost(h *models.Host) bool {
```
mkowalski, line 159: Can we get here some comment on what creates the file mentioned here and how this function is meant to be used? It will live in utils, so there is a potential that someone in the future will try to use it for another reason; if that's the case, some note that there is another entity creating the `localHostIDFile` would be helpful

flaper87:
Brought this up in one of the threads where we are discussing this feature. We could switch from this process to using the host's mac-address and match the agent and the service the same way BMAC matches a BMH and an Agent today.

This would require passing the host's mac address to the assisted service process when it starts. If that info is available, then the service will find a matching agent and set it as bootstrap. Otherwise, the regular flow will continue as it does today. 

No need to add new APIs or dependencies on files

avishayt:
> Brought this up in one of the threads where we are discussing this feature. We could switch from this process to using the host's mac-address and match the agent and the service the same way BMAC matches a BMH and an Agent today.
> 
> This would require passing the host's mac address to the assisted service process when it starts. If that info is available, then the service will find a matching agent and set it as bootstrap. Otherwise, the regular flow will continue as it does today.
> 
> No need to add new APIs or dependencies on files

This.  Plus wrap it in a feature flag, disabled by default.

zaneb:
/close
replaced by #3627

PR #3597, opened 2022-03-30
MGMT-9726: Wait for assisted-image-service when deploying the operator

adriengentil:
In the late binding job, we try to download the ISO image from the
assisted-image-service but this one is not available yet.

This change waits for the assisted-image-service to be up before exiting the
deployment script.


osherdp (approved):

omertuc:
It worked, I think :scream:

adriengentil:
> It worked, I think scream

I'll unhold this PR because it looks like it fixes the CI immediately, and track the outcome of the slack discussion!

adriengentil:
/retile MGMT-9726: Wait for assisted-image-service when deploying the operator

adriengentil:
/retitle MGMT-9726: Wait for assisted-image-service when deploying the operator

PR #3621, opened 2022-04-05
Bug 2061317: when using nmstateconfigs ISO is re-generated for each nmstateconfig

ori-amizur:
When updating infra-env, static network configuration is saved if it is
present in the update parameters and if it is different from the
previous value already in the database. Updating infra-env causes
download URL regeneration which is undesirable.
Since this configuration is a slice of elements, that each of them
contains a slice, changing the order of items in any of the slices, may
cause the static network configuration to be different than the previous
value, but equivalent in terms of configuration when applied.
To prevent this, sort was added to verify that equivalent configurations
will generate same string value to be stored in the database.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] Unit tests

## Assignees

/cc @filanov 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/staticnetworkconfig/generator.go
+++ b/pkg/staticnetworkconfig/generator.go
@@ -220,10 +220,49 @@ func (s *StaticNetworkConfigGenerator) validateNMStateYaml(ctx context.Context, 
 	return err
 }
 
+func compareMapInterfaces(intf1, intf2 *models.MacInterfaceMapItems0) bool {
+	if intf1.LogicalNicName != intf2.LogicalNicName {
+		return intf1.LogicalNicName < intf2.LogicalNicName
+	}
+	return intf1.MacAddress < intf2.MacAddress
+}
+
+func compareMacInterfaceMaps(map1, map2 models.MacInterfaceMap) bool {
+	if len(map1) != len(map2) {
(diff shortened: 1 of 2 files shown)


mkowalski (commented):
I like the change 👍🏻 
Can we just make the commit more descriptive? Currently if I do `git log` and see it, I will think that what you introduce is regeneration of the ISO for each entry; i.e. the message is not clear that it in fact attempts to fix the issue (only after reading the change and bugzilla I see what was the issue and solution).

Something like e.g. `Bug xxx: Sort nmstateconfigs before saving in the DB`

filanov:
+1 please explain what it fix and why it's sorted

mkowalski:
/lgtm

I'm not 100% sure if this change only will solve all the issues with regenerating ISO, but it's definitely good improvement

ori-amizur:
> /lgtm
> 
> I'm not 100% sure if this change only will solve all the issues with regenerating ISO, but it's definitely good improvement

Verified by @nshidlin

PR #3622, opened 2022-04-05
MGMT-9673: Add HasAccessTo API to authz

slaviered:
Authorization becomes more complex with the tenancy-based feature and more authorization schemes added for other projects.
In the current implementation, the application layer is using stateless utilities to check for access permission to objects on top of the existing mechanism. These checks are sometimes redundant and or not quite aligned with the new authorization scheme.

Using stateless utilities makes us expose feature flags and states in the auth API where they do not belong. For all those reasons, an API for authorization that will be shared by all authorization handlers and expose coherent access control algorithms to the application is required.

This is the second installment of the enhancement. It defines The HasAccessTo interface that queries whether the current user can access an object, based on the active authorization layer.

The objects that participate in the authorization are cluster, infra-env, and host.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/auth/rhsso_authz_handler.go
+++ b/pkg/auth/rhsso_authz_handler.go
@@ -121,7 +121,7 @@ func (a *AuthzHandler) hasSubscriptionAccess(clusterId string, action string, pa
 			return handleOwnershipQueryError(err)
 		}
 
-		cacheKey := fmt.Sprintf("%s_%s_%s", payload.Username, payload.Organization, cluster.AmsSubscriptionID)
+		cacheKey := fmt.Sprintf("%s_%s_%s_%s", payload.Username, payload.Organization, cluster.AmsSubscriptionID, action)
 		if cacheData, existInCache := a.client.Cache.Get(cacheKey); existInCache {
 			var ok bool
 			isAllowed, ok = cacheData.(bool)
@@ -132,7 +132,6 @@ func (a *AuthzHandler) hasSubscriptionAccess(clusterId string, action string, pa
 			}
 			return isAllowed, nil
 		}
-
 		isAllowed, err = a.hasClusterEditRole(payload, action, cluster.AmsSubscriptionID.String())
 		if shouldStorePayloadInCache(err) {
 			a.client.Cache.Set(cacheKey, isAllowed, 10*time.Minute)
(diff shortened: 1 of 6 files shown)


pkg/auth/rhsso_authz_handler.go
```diff
@@ -178,14 +196,18 @@ func (a *AuthzHandler) checkInfraEnvBasedAccess(id string, request *http.Request
 	}
 
 	if infraEnv.ClusterID != "" {
-		if isAllowed, _ := a.checkClusterBasedAccess(infraEnv.ClusterID.String(), request, payload); isAllowed {
+		if isAllowed, _ := a.checkClusterBasedAccess(infraEnv.ClusterID.String(), action, payload); isAllowed {
```
danielerez, line 199: We're invoking isObjectOwnedByUser here (as part of hasSubscriptionAccess), so in case isAllowed is false - we'll call isObjectOwnedByUser again in line 209. So to avoid that we can maybe do the 'action == ReadAction' first.
i.e. move lines 206-208 to 197. Then we won't need a fallback to checkClusterBasedAccess - just return isAllowed

pkg/auth/rhsso_authz_handler_test.go
```diff
@@ -290,6 +291,230 @@ var _ = Describe("OwnedBy", func() {
...
+		It("ClusterEditorRole on cluster-bound host can read and write according to cluster rules", func() {
+			payload := &ocm.AuthPayload{Username: "user2", Organization: "org1"}
+			ctx = context.WithValue(ctx, restapi.AuthKey, payload)
+
+			//host[id2] bound to cluster[id1]
+			host, _ := common.GetHostFromDBbyHostId(db, id2)
+			mockOcmAuthorization.EXPECT().AccessReview(
+				gomock.Any(), "user2", gomock.Any(), "other", gomock.Any()).
+				Return(true, nil).Times(2)
+			Expect(authzHandler.HasAccessTo(ctx, host, ReadAction)).To(BeFalse())
```
danielerez, line 488: If update/delete action are allowed, shouldn't read action also be allowed here?

slaviered: It is actually a contrived example where the user is not the object owner but the access review allow them a write access. It will not actually happen in real life but it covers a code branch

pkg/auth/rhsso_authz_handler.go
slaviered: fixed

PR #3627, opened 2022-04-06
AGENT-34: Allow preconfiguration of bootstrap host

zaneb:
If the environment variable BOOTSTRAP_HOST_MAC is set, select a host
with that MAC address as the bootstrap and ensure it is designated to be
a control plane host.

This will be used in the ephemeral installer to ensure that the host
running assisted-service is also used as the bootstrap.

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] Ephemeral Installer
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @flaper87 
/cc @avishayt 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -86,6 +86,7 @@ type Config struct {
 	ResetTimeout            time.Duration           `envconfig:"RESET_CLUSTER_TIMEOUT" default:"3m"`
 	MonitorBatchSize        int                     `envconfig:"HOST_MONITOR_BATCH_SIZE" default:"100"`
 	DisabledHostvalidations DisabledHostValidations `envconfig:"DISABLED_HOST_VALIDATIONS" default:""` // Which host validations to disable (should not run in preprocess)
+	BootstrapHostMAC        string                  `envconfig:"BOOTSTRAP_HOST_MAC" default:""`        // For ephemeral installer to ensure the bootstrap for the (single) cluster lands on the same host as assisted-service
 }
 
 //go:generate mockgen -package=host -aux_files=github.com/openshift/assisted-service/internal/host/hostcommands=instruction_manager.go -destination=mock_host_api.go . API
@@ -319,6 +320,25 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster, 
 	if err != nil {
 		log.WithError(err).Errorf("not updating inventory - failed to find infra env %s", h.InfraEnvID.String())
 		return common.NewApiError(http.StatusNotFound, err)
+	}
+
+	if m.Config.BootstrapHostMAC != "" && !h.Bootstrap {
+		for _, iface := range inventory.Interfaces {
+			if iface.MacAddress == m.Config.BootstrapHostMAC {
+				log.Infof("selected local bootstrap host %s for cluster %s", h.ID, cluster.ID)
+				err = updateRole(log, h, models.HostRoleMaster, models.HostRoleMaster, db, string(h.Role))
+				if err != nil {
+					log.WithError(err).Errorf("failed to set master role on bootstrap host for cluster %s", cluster.ID)
+					return errors.Wrapf(err, "Failed to set master role on bootstrap host for cluster %s", cluster.ID)
+				}
+				err = m.SetBootstrap(ctx, h, true, db)
+				if err != nil {
+					log.WithError(err).Errorf("failed to update bootstrap host for cluster %s", cluster.ID)
+					return errors.Wrapf(err, "Failed to update bootstrap host for cluster %s", cluster.ID)
+				}
+				break
+			}
+		}
 	}
 
 	err = m.populateDisksEligibility(ctx, inventory, infraEnv, cluster, h)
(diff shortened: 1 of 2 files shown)


internal/host/host.go
```diff
@@ -86,6 +86,7 @@ type Config struct {
 	ResetTimeout            time.Duration           `envconfig:"RESET_CLUSTER_TIMEOUT" default:"3m"`
 	MonitorBatchSize        int                     `envconfig:"HOST_MONITOR_BATCH_SIZE" default:"100"`
 	DisabledHostvalidations DisabledHostValidations `envconfig:"DISABLED_HOST_VALIDATIONS" default:""` // Which host validations to disable (should not run in preprocess)
+	BootstrapHostMAC        string                  `envconfig:"BOOTSTRAP_HOST_MAC" default:""`
```
avishayt, line 89: A comment would be helpful here

internal/host/host.go
```diff
@@ -321,6 +322,24 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
 		return common.NewApiError(http.StatusNotFound, err)
 	}
 
+	if m.Config.BootstrapHostMAC != "" {
```
avishayt, line 325: This is OK and should work.  Would you rather have the logic here where it is executed approximately once a minute upon inventory update, or perhaps in setBootstrapHost() in inventory.go?

eranco74: Moving this logic to `setBootstrapHost` makes sense.

The problem with that is that there is a situation where the node that should be bootstrap might not have a master role.
Doing the bootstrap assignment here reduces the probability of that (the user might still change the node role and click install before the host sends inventory again).

I guess we should also add code that blocks setting the host role as worker in case the host has the `BootstrapHostMAC`

internal/host/host.go
romfreiman, line 89: Lets add a comment here that this param should be used only in case of a single cluster deployment. Because in podman it won't really work

flaper87 (commented):
Other than @avishayt and @eranco74 comments, I don't have much else to add. 

Looks good

internal/host/host.go
zaneb: I added a condition to check that the host is not already set as the bootstrap, so that we don't execute the DB calls more than once.

I initially tried it in setBootstrapHost(), but then moved it here for the reason Eran mentioned - at that point it is too late to change the Role because installation has already started.

> I guess we should also add code that blocks setting the host role as worker in case the host has the BootstrapHostMAC

I think this could be more generic: don't allow setting the role to Worker if the host is marked as the bootstrap.

internal/host/host.go
```diff
@@ -321,6 +322,24 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
 		return common.NewApiError(http.StatusNotFound, err)
 	}
 
+	if m.Config.BootstrapHostMAC != "" && !h.Bootstrap {
+		for _, iface := range inventory.Interfaces {
+			if iface.MacAddress == m.Config.BootstrapHostMAC {
+				log.Infof("selected local bootstrap host %s for cluster %s", h.ID, cluster.ID)
+				err = m.UpdateRole(ctx, h, models.HostRoleMaster, db)
```
omertuc, line 329: nit: This line kinda makes [this](https://github.com/openshift/assisted-service/blob/0cb3eb4601060c9a4111a3bc99da3c9cc086d115/internal/host/host.go#L615-L616) code comment slightly incorrect, might want to adjust it

omertuc:
When `BootstrapHostMAC` is set, role selection should be completely disabled for the user, both in the API and the UI, otherwise the user could do silly things like setting a different bootstrap host, a decision which would automatically get overridden when a new inventory update comes

omertuc:
When `BootstrapHostMAC` is set, we should not allow the user to change the role for that particular host, otherwise the user could set the host's role to worker and that would put the database in an inconsistent state with unpredictable results

zaneb:
> When `BootstrapHostMAC` is set, we should not allow the user to change the role for that particular host to "auto-assign" or "worker" (both in the UI and the API), otherwise the user could accidentally set that host's role to worker and that would put the database in an inconsistent state with unpredictable results

We're not going to have a UI/API at first, so this isn't an issue yet. I just created [AGENT-113](https://issues.redhat.com/browse/AGENT-113) to keep track of this for when we're implementing the interactive flow.

internal/host/host.go
zaneb: What would be a correct version of that comment?

omertuc: It says it's always invoked from an API (referring to the host role update api) which is kinda untrue because now it's invoked internally from the inventory API. Also it says that the role is user selected but it's not, it's selected automatically in this case with the mac mechanism.

I'm not sure exactly what's the point of that comment, so I'm not sure what would be the correct phrasing, but I can tell after this change it's at least inaccurate, so something should be done. Either delete it or figure out its purpose and rephrase it somehow

zaneb: I replaced the call to `m.UpdateRole()` with a direct call to `updateRole()` so that none of us need to decide what the comment means :smile:

eranco74 (commented):
Looks good, just a nit

flaper87 (approved):

zaneb:
Addressed Eran's comment.
/hold cancel

PR #3630, opened 2022-04-06
MGMT-9732: add user authorization to BindHost

danielerez:
Added proper authz to BindHostInternal func by using HasAccessTo API.
The specified cluster (the should be bounded to host) is checked for ensuring user's update permissions. I.e. cluster is owned by the user, or, if tenancy is enabled, we query AMS with the associated subscription.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @slaviered 
/cc @avishayt 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4792,6 +4792,12 @@ func (b *bareMetalInventory) BindHostInternal(ctx context.Context, params instal
 	if err != nil {
 		return nil, common.NewApiError(http.StatusBadRequest, errors.Errorf("Failed to find cluster %s", params.BindHostParams.ClusterID))
 	}
+	allowed, err := b.authzHandler.HasAccessTo(ctx, cluster, auth.UpdateAction)
+	if !allowed {
+		msg := fmt.Sprintf("Failed to find cluster %s", params.BindHostParams.ClusterID)
+		b.log.WithError(err).Error(msg)
+		return nil, common.NewApiError(http.StatusNotFound, errors.New(msg))
+	}
 	infraEnv, err := common.GetInfraEnvFromDB(b.db, params.InfraEnvID)
 	if err != nil {
 		b.log.WithError(err).Errorf("Failed to get infra env %s", params.InfraEnvID)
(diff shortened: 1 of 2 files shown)


internal/bminventory/inventory.go
```diff
@@ -4792,6 +4792,10 @@ func (b *bareMetalInventory) BindHostInternal(ctx context.Context, params instal
 	if err != nil {
 		return nil, common.NewApiError(http.StatusBadRequest, errors.Errorf("Failed to find cluster %s", params.BindHostParams.ClusterID))
 	}
+	allowed, err := b.authzHandler.HasAccessTo(ctx, cluster, auth.UpdateAction)
+	if err != nil || !allowed {
```
slaviered, line 4796: It's enough to check !allowed because we return false in case of an error. This will make the code more readable. The error could be wrapped in application error here providing it can handle nil gracefully

internal/bminventory/inventory.go
```diff
@@ -4792,6 +4792,10 @@ func (b *bareMetalInventory) BindHostInternal(ctx context.Context, params instal
...
+		return nil, common.NewApiError(http.StatusBadRequest, errors.Errorf("Failed to find cluster %s", params.BindHostParams.ClusterID))
```
slaviered, line 4797: Why are we returning BadRequest which is a code designed for malformed inputs? I think that we should return 403 or 404. This will help the client to understand what the problem is

internal/bminventory/inventory.go
danielerez: Fixed by checking 'allowed' only and added err to the log (as we don't want the expose the error message in the API response). We could consider logging only in the HasAccessTo func if it looks too cumbersome.

internal/bminventory/inventory.go
danielerez: Changed to 404 as we don't want to expose any security ino in the api response.

PR #3639, opened 2022-04-09
AGENT-56 Client to read ZTP manifests and register cluster and infraenv

rwsu:
The client is used by the ephemeral-installer to register the
initial cluster and infraenv for cluster0 based off of the information
from the ZTP manifests.

The ZTP manifests are not directly applied to a kube-api because
a kube-api is not available in the environment. Assisted-service
runs as a container on one of the hosts in the cluster.

The purpose of the client in the ephemeral-installer is to minimize
user interaction and to enable automation use cases.

Briefly, the ephemeral-installer workflow follows these steps:
1. The user builds a single ISO that contains the ZTP manifests and
both assisted-service and agent.
2. The ISO is booted on the hosts that will form the cluster.
3. On one of the hosts, node0, the assisted-service runs and after
assisted-service has started up, this client is executed.
4. The agent runs on all of the hosts and registers them with
assisted-service after the infraenv is created on node0.
5. Once host validation completes, a service on node0 initiates the
cluster installation.

The creation of the cluster and infra params have been refactored
out of the createNewCluster and createInfraEnv functions and into
new functions so that their logic may be reused by this client.

Co-authored-by: Bob Fournier <bfournie@redhat.com>
Co-authored-by: Pawan Pinjarkar <ppinjark@redhat.com>
Ref: https://issues.redhat.com/browse/AGENT-56

- [ x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

The code refactors how the cluster and infra env params are created so it could impact all environments.

## How was this code tested?

- [ ] dev-scripts environment
- [] No tests needed

I built a custom assisted-service image and configured the ISO builder tool in https://github.com/openshift-agent-team/fleeting to use the image in create-cluster-and-infraenv.service. The ISO was booted in a VM and I checked that the client executed and a cluster and a infraenv was registered in assisted-service.

## Assignees

/cc @danielerez
/cc @nmagnezi  
/cc @filanov 

## Checklist

## Reviewers Checklist


lranjbar:
It looks like we need to update some of the subsystem tests or check on the referencing of the imageVersion and cpuArchitecture. Subsystem tests are coming back nil. I think it is related to that.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
```diff
@@ -0,0 +1,248 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+var Options struct {
+	ClusterDeploymentFile   string `envconfig:"CLUSTER_DEPLOYMENT_FILE" default:"/manifests/cluster-deployment.yaml"`
+	AgentClusterInstallFile string `envconfig:"AGENT_CLUSTER_INSTALL_FILE" default:"/manifests/agent-cluster-install.yaml"`
+	InfraEnvFile            string `envconfig:"INFRA_ENV_FILE" default:"/manifests/infraenv.yaml"`
+	PullSecretFile          string `envconfig:"PULL_SECRET_FILE" default:"/manifests/pull-secret.yaml"`
+	ClusterImageSetFile     string `envconfig:"CLUSTER_IMAGE_SET_FILE" default:"/manifests/cluster-image-set.yaml"`
+	NMStateConfigFile       string `envconfig:"NMSTATE_CONFIG_FILE" default:"/manifests/nmstateconfig.yaml"`
```
filanov, line 84: Is that expected to be a single file?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1078,41 +1078,20 @@ func (r *ClusterDeploymentsReconciler) isSNO(clusterInstall *hiveext.AgentCluste
...
 
-func (r *ClusterDeploymentsReconciler) createNewCluster(
-	ctx context.Context,
-	log logrus.FieldLogger,
-	key types.NamespacedName,
-	clusterDeployment *hivev1.ClusterDeployment,
-	clusterInstall *hiveext.AgentClusterInstall) (ctrl.Result, error) {
-
-	log.Infof("Creating a new cluster %s %s", clusterDeployment.Name, clusterDeployment.Namespace)
+func CreateClusterParams(clusterDeployment *hivev1.ClusterDeployment, clusterInstall *hiveext.AgentClusterInstall, pullSecret string, releaseImageVersion string, releaseImageCPUArch string) *models.ClusterCreateParams {
```
filanov, line 1081: please break long lines

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1180,6 +1148,46 @@ func (r *ClusterDeploymentsReconciler) createNewCluster(
...
+	if err != nil {
+		log.WithError(err)
+		_, _ = r.updateStatus(ctx, log, clusterInstall, nil, err)
+		// The controller will requeue after one minute, giving the user a chance to fix releaseImage
+		return ctrl.Result{Requeue: true, RequeueAfter: longerRequeueAfterOnError}, nil
+	}
+
+	clusterParams := CreateClusterParams(clusterDeployment, clusterInstall, pullSecret, *releaseImage.Version, *releaseImage.CPUArchitecture)
+
+	if clusterInstall.Spec.IgnitionEndpoint != nil {
```
filanov, line 1180: a bit strange to leave this part outside of CreateClusterParams 
Maybe it should be another parameter? you can do the parsing outside and pass `ignitionEndpoint` as a parameter as well

internal/controller/controllers/infraenv_controller.go
```diff
@@ -369,6 +362,20 @@ func (r *InfraEnvReconciler) createInfraEnv(ctx context.Context, log logrus.Fiel
 		createParams.InfraenvCreateParams.ClusterID = cluster.ID
 		createParams.InfraenvCreateParams.OpenshiftVersion = cluster.OpenshiftVersion
 	}
+
+	return createParams
+}
+
+func (r *InfraEnvReconciler) createInfraEnv(ctx context.Context, log logrus.FieldLogger, key *types.NamespacedName, infraEnv *aiv1beta1.InfraEnv, cluster *models.Cluster) (*common.InfraEnv, error) {
```
filanov, line 369: please break long lines

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
rwsu: It is a single file right now. We will likely support multiple NMStateConfig files soon.

internal/controller/controllers/clusterdeployments_controller.go
rwsu: Ok.

internal/controller/controllers/clusterdeployments_controller.go
rwsu: Good idea. I will parse outside and pass it in as a parameter.

internal/controller/controllers/infraenv_controller.go
rwsu: Ok.

rwsu:
> It looks like we need to update some of the subsystem tests or check on the referencing of the imageVersion and cpuArchitecture. Subsystem tests erroring and it seems those are coming back nil. I think it is related to that.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
filanov: So lets change it to a list?

rwsu: Correction, it will be a single file containing multiple yaml definitions. @bfournie has a PR in the fleeting repo to add multiple nmstateconfigs: https://github.com/openshift-agent-team/fleeting/pull/25. We will need to move his changes over here too.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
```diff
@@ -0,0 +1,248 @@
...
+A kube-api is not available in the ephemeral-installer environment. This
+is why the ZTP manifests are not applied directly to kubernetes. Instead,
+the manifests are translated into create cluster and infraenv params, and
+then posted to the REST-API to create those resources.
+
+This client is needed to register the cluster and infraenv to minimize
+user interaction and to enable automation use cases.
+
+A service running on the image, waits for the infraenv to be registered
+to assisted-service. Once registered, an infraenv id is available and
```
omertuc, line 42: `an infraenv id is available` is that still true? I thought the client forces a specific infra env ID to be used

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
```diff
@@ -0,0 +1,248 @@
...
+	var clusterImageSet hivev1.ClusterImageSet
+	if err := getFileData(Options.ClusterImageSetFile, &clusterImageSet); err != nil {
+		log.Fatal(err.Error())
+	}
+	return clusterImageSet.Spec.ReleaseImage
+}
+
+func getReleaseVersionAndCpuArch(releaseImage string) (version string, cpuArch string) {
+	// releaseImage is in the form: quay.io/openshift-release-dev/ocp-release:4.9.17-x86_64
+	versionCpuArch := strings.Split(releaseImage, ":")[1]
```
omertuc, line 191: This can break when using port numbers, e.g:
`quay.io:443/openshift-release-dev/ocp-release:4.9.17-x86_64`

omertuc: Or when using exact sha:
`quay.io/openshift-release-dev/ocp-release@sha256:7b67b0cb5ab016528b8efdb6130c000398efc58f55e2226f3cf4e3be59c78443`

But admittedly the latter one is less of a problem because you wouldn't be able to parse the architecture anyway

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
```diff
@@ -0,0 +1,248 @@
...
+	versionCpuArchSplit := strings.Split(versionCpuArch, "-")
+	if len(versionCpuArchSplit) > 2 {
+		// case: 4.10.0-rc1-x86_64
+		version = strings.Join(versionCpuArchSplit[:2], "-")
+		cpuArch = versionCpuArchSplit[2]
+	} else {
+		// case: 4.10.1-x86_64
+		version = versionCpuArchSplit[0]
+		cpuArch = versionCpuArchSplit[1]
+	}
```
omertuc, line 202:
```suggestion
	versionCpuArchSplit := strings.Split(versionCpuArch, "-")
	// Assume the architecture is always the last component
	lastIndex := len(versionCpuArchSplit) - 1
	version = strings.Join(versionCpuArchSplit[:lastIndex], "-")
	cpuArch = versionCpuArchSplit[lastIndex]
```

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
```diff
@@ -0,0 +1,248 @@
...
+	} else if err = yaml.Unmarshal(contents, output); err != nil {
+		err = fmt.Errorf("error unmarshalling contents of %s: %w", filePath, err)
+	}
+
+	return err
+}
+
+func validateNMStateConfigAndInfraEnv(nmStateConfig aiv1beta1.NMStateConfig, infraEnv aiv1beta1.InfraEnv) error {
+	if len(nmStateConfig.ObjectMeta.Labels) == 0 {
+		return errors.Errorf("nmstateconfig does not have any labels set")
```
omertuc, line 221: Why is this a problem? Maybe the error should elaborate?

nmagnezi: I agree with Omer here.
An NMStateConfig with no labels will just not be included in any label matching. Why is this an error?

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	if len(infraEnv.Spec.NMStateConfigLabelSelector.MatchLabels) == 0 {
```
nmagnezi, line 224: Same question here. Must an InfraEnv have a matching NMState?

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
rwsu: The ability to configure the client to use a pre-generated infra-env-id is a future enhancement.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
rwsu: Good catch.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
rwsu: Thanks. I'll add your suggestion.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
rwsu: Yes. We are requiring there to be at least one NMState for node0.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
rwsu: Any NMStateConfigs this client is reading should be targeting the infra-env being created. So it should match the label specified in the infra-env. I'll update the error message to be more specific.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
celebdor: This is because in a fully DHCPed environment we would not have a way to know which address assisted-service is serving at, otherwise.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
```diff
@@ -0,0 +1,244 @@
...
+
+func getReleaseVersion() string {
+	var clusterImageSet hivev1.ClusterImageSet
+	if err := getFileData(Options.ClusterImageSetFile, &clusterImageSet); err != nil {
+		log.Fatal(err.Error())
+	}
+	return clusterImageSet.Spec.ReleaseImage
+}
+
+func getReleaseVersionAndCpuArch(releaseImage string) (version string, cpuArch string) {
```
omertuc, line 189: Would be nice to have explicit failure when you're given a release image that doesn't contain the architecture in the tag - or alternatively, default to x86

Dockerfile.assisted-service
```diff
@@ -47,6 +48,7 @@ RUN curl -s $OC_URL | tar -xzC /usr/local/bin/ oc
 COPY --from=builder /build/assisted-service /assisted-service
 COPY --from=builder /build/assisted-service-operator /assisted-service-operator
 COPY --from=builder /build/assisted-service-admission /assisted-service-admission
+COPY --from=builder /build/ephemeral-installer-register-cluster-and-infraenv /ephemeral-installer-register-cluster-and-infraenv
```
andfasano, line 51: As soon as https://github.com/openshift/assisted-service/pull/3660 will land it will be necessary to add the ephemeral client in `Dockerfile.ocp` as well. 
Not sure if in future (when the images will be shipped in the ocp payload) will make sense to keep that binary in the `Dockerfile.assisted-service` as well

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
rwsu: We will need to validate the release image URL at a higher level before it reaches this client. Like you said, by the time it reaches this client it will be too late. I'm not sure if defaulting to x86_64 is a safe assumption. Let me ask the team.

danielerez: We avoid parsing the ocp release image uri by extracting the version/arch from the image.
See: https://github.com/openshift/assisted-service/blob/f74ba9bdd99924146140c448982c3b925c2669d3/internal/versions/versions.go#L297-L307
This allows us to support a release image uri with a commit hash, i.e. we return the ocp version and cpu arch defined in the image metadata.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
nmagnezi: Ack.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
nmagnezi: Ack.

nmagnezi (changes requested):
Overall, looks good. 
I have some concerns about maintainability:
We need to make sure this is being tested in CI with some good sample manifests, which will prevent future changes of our kubeAPI from breaking this client without taking it into account.
I know this is based on code your guys refactored from CD and InfraEnv controller (which is excellent!), but there could still be corner cases. Also, this will help keep those sample manifests up to date and in good condition - which is beneficial for users.

Lastly, some sample manifests (with reference from a `.md` doc would also be great.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	var secret corev1.Secret
+	if secretErr := getFileData(Options.PullSecretFile, &secret); secretErr != nil {
+		log.Fatal(secretErr.Error())
+	}
+	pullSecret := secret.StringData[".dockerconfigjson"]
+
+	releaseImageVersion, releaseImageCPUArch := getReleaseVersionAndCpuArch(getReleaseVersion())
+	log.Info("releaseImage: " + getReleaseVersion())
```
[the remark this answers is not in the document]
rwsu: Ok, I will save the result to eliminate the extra call.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
```diff
@@ -0,0 +1,244 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+The ephemeral-installer-register-cluster-and-infraenv is a client that runs
```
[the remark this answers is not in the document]
rwsu: Sure. I'll move it to a doc and reference it from the code.

cmd/ephemeral-installer-register-cluster-and-infraenv/main.go
rwsu: Good point about commit hash. I'll switch to using the release package to determine the image version and architecture.

rwsu:
@nmagnezi Thanks for your review. I've added a subsystem test to register the cluster and infraenv. I used the existing manifests in [hive-integration](https://github.com/openshift/assisted-service/tree/master/docs/hive-integration/crds). The exact contents in the manifests don't matter much, but certain values need to transfer from the ZTP manifests into the models, and there are checks for those now. I wanted to avoid creating another set of manifests so I hope using the existing manifests under docs/hive-integration/crds is ok.

docs/agent-based-installer.md
```diff
@@ -0,0 +1,42 @@
+# Agent-based installer
+
+The assisted-service repo contains a client used by the agent-based installer. 
+The client code is located at [cmd/agentbasedinstaller](cmd/agentbasedinstaller/).
```
nmagnezi, line 4: this location URL is broken. should probably start with `../cmd`

nmagnezi (commented):
One minor comment about a doc URL, otherwise LGTM.
I'll let other commenters to ack their comments before I vote.

docs/agent-based-installer.md
rwsu: Thanks. Fixed the location URL.

rwsu:
The new ci failures are happening because we are now using the vendor directory (PR#3660 merged) and vendor/modules.txt needs to be updated to include the replace for agentbasedinstaller in go.mod:

> [3/5] STEP 11/14: RUN CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /build/assisted-service cmd/main.go
> go: inconsistent vendoring in /go/src/github.com/openshift/origin:
> 	github.com/openshift/assisted-service/cmd/agentbasedinstaller: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
> 	To ignore the vendor directory, use -mod=readonly or -mod=mod.
> 	To sync the vendor directory, run:
> 		go mod vendor

Will need to update vendor/modules.txt:

`+# github.com/openshift/assisted-service/cmd/agentbasedinstaller => ./cmd/agentbasedinstaller`

omertuc:
Forgot to update `Dockerfile.assisted-service.ocp`

PR #3669, opened 2022-04-14
MGMT-8987: 4 nodes cluster deployment

flaper87:
Enhancement to define the support and implementation of a 4-nodes
cluster deployment flow through assisted service.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @omertuc 
/cc @avishayt 
/cc @mhrivnak 

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/4-nodes-cluster-deployment.md
+++ b/docs/enhancements/4-nodes-cluster-deployment.md
@@ -56,7 +56,7 @@ The biggest changes anticipated for this enhancement are:
 - Expose the manifests API through the InfraEnv CRD so that custom manifests can be added through
 the ZTP flow.
 
-- Expose the schedulable mastes through the AgentClusterInstall to allow for explicitly creating
+- Expose the schedulable masters through the AgentClusterInstall to allow for explicitly creating
 schedulable masters.
 
 - Modify the `enoughMastersAndWorkers` validation so that it allows 1 worker when masters are
--- a/docs/enhancements/4-nodes-cluster-deployment.md
+++ b/docs/enhancements/4-nodes-cluster-deployment.md
@@ -18,17 +18,17 @@ enclosure that only had 4 nodes. The goal was to use all of the blades rather th
 
 For now, this is achieved with a 2 step deployment, first 3 masters and then add a worker node. It
 would be more time efficient as well as easier and more reliable to be able to do this deployment in
+
 one step, day0.
 
 # Motivation
 
 Assisted service has built-in validations that prevent the deployment of a 4-nodes cluster. This is
-to prevent unhealthy clusters to be deployed. A 4-nodes cluster ends up in an unhealthy state
-because openshift-installer doesn't take schedulable masters into consideration when there are
-workers being deployed. The problem with schedulable masters not being counted as workers causes the
-installer to chose an `InfrastructureTopology` that is SingleReplica. This means that infrastructure
-workloads will run as single replicas, only on workers. This setting cannot be changed once the
-cluster is deployed.
+to prevent the deployment of topologies that are not part of the OCP standard test matrix. A 4-nodes
+cluster doesn't take schedulable masters into consideration when there are workers being deployed,
+which causes the installer to chose an `InfrastructureTopology` that is SingleReplica. This means
+that infrastructure workloads will run as single replicas, only on workers. This setting cannot be
+changed once the cluster is deployed.
 
 The above was brought up to the OpenShift team as [an enhancement](https://github.com/openshift/enhancements/pull/1057/). The feedback so far is that
 it's OK for users (or other installers) to overwrite the openshift manifests to achieve the
@@ -53,9 +53,6 @@ that may not worth the effort at this stage.
 
 The biggest changes anticipated for this enhancement are:
 
-- Expose the manifests API through the InfraEnv CRD so that custom manifests can be added through
-the ZTP flow.
-
 - Expose the schedulable masters through the AgentClusterInstall to allow for explicitly creating
 schedulable masters.
 
@@ -63,10 +60,19 @@ schedulable masters.
 schedulable. [PoC here](https://github.com/flaper87/assisted-service/commit/f07fc6589e4f2d316266ac5533d76bd2faf471d7)
 and [here](https://github.com/flaper87/assisted-service/commit/bb2d82eaacaee7790d2f18711e2ba3a442a4967d)
 
-- Modify Assisted Service's manifests logic to allow for partial manifests to be provided. Openshift
-installer applies the provided manifests in sequence. If 2 manifests for the same CR are provided,
-then the one applied last will be the one used. This is a problem when the user only wants to modify
-1 field. Here's an [example from a PoC](https://github.com/flaper87/assisted-service/commit/c6d72827b43786f95e68e052623e8aef0d4b4c0f).
+- Modify Assisted Service's manifests logic to allow for partial manifests to be provided. The need
+for this change is twofold:
+    1. This is needed because Openshift installer generates whole manifests, some of which have values that are computed at run time, which
+are unknown to the user when triggering the deployment.
+    2. When providing manifests overrides, OpenShift installer applies them in sequence and it
+    expects entire manifests to be provided. You can think of it as the installer running `oc apply
+    -f` for every yaml that exists in the manifests dir.
+
+    Because of the above, users that want to customize manifests don't have a way to provide a
+    partial manifest. Providing a full manifest may cause the override of fields computed at runtime
+    by the installer.
+
+    Here's an [example from a PoC](https://github.com/flaper87/assisted-service/commit/c6d72827b43786f95e68e052623e8aef0d4b4c0f).
 
 ### User Stories
 
--- a/docs/enhancements/4-nodes-cluster-deployment.md
+++ b/docs/enhancements/4-nodes-cluster-deployment.md
@@ -10,7 +10,7 @@ last-updated: yyyy-mm-dd
 
 # Summary
 
-**Note:** Throughout this enhancement, a 4-nodes cluster refers to a 3 schedulable masters and 1 worker.
+**Note:** Throughout this enhancement, a 4-nodes cluster refers to a 3 schedulable masters and 1 worker cluster.
 
 There are cases where customers may just have 4 nodes available and no chance to add more. This was
 one of the cases in the ZTP for Factory Workflow where it was required to deploy a cluster in an
@@ -18,7 +18,6 @@ enclosure that only had 4 nodes. The goal was to use all of the blades rather th
 
 For now, this is achieved with a 2 step deployment, first 3 masters and then add a worker node. It
 would be more time efficient as well as easier and more reliable to be able to do this deployment in
-
 one step, day0.
 
 # Motivation
--- a/docs/enhancements/4-nodes-cluster-deployment.md
+++ b/docs/enhancements/4-nodes-cluster-deployment.md
@@ -1,0 +1,154 @@
+---
+title: 4-nodes-cluster-creation
+authors:
+  - "@flaper87"
+creation-date: 2022-04-14
+last-updated: yyyy-mm-dd
+---
+
+# Allow for a 4-nodes cluster to be created
+
+# Summary
+
+**Note:** Throughout this enhancement, a 4-nodes cluster refers to a 3 schedulable masters and 1 worker.
+
+There are cases where customers may just have 4 nodes available and no chance to add more. This was
+one of the cases in the ZTP for Factory Workflow where it was required to deploy a cluster in an
+enclosure that only had 4 nodes. The goal was to use all of the blades rather than just 3 of them.
+
+For now, this is achieved with a 2 step deployment, first 3 masters and then add a worker node. It
+would be more time efficient as well as easier and more reliable to be able to do this deployment in
+one step, day0.
+
+# Motivation
+
+Assisted service has built-in validations that prevent the deployment of a 4-nodes cluster. This is
+to prevent unhealthy clusters to be deployed. A 4-nodes cluster ends up in an unhealthy state
+because openshift-installer doesn't take schedulable masters into consideration when there are
+workers being deployed. The problem with schedulable masters not being counted as workers causes the
+installer to chose an `InfrastructureTopology` that is SingleReplica. This means that infrastructure
+workloads will run as single replicas, only on workers. This setting cannot be changed once the
+cluster is deployed.
+
+The above was brought up to the OpenShift team as [an enhancement](https://github.com/openshift/enhancements/pull/1057/). The feedback so far is that
+it's OK for users (or other installers) to overwrite the openshift manifests to achieve the
+deployment topologies required. The conversation is still in progress as far as understanding if
+some changes could be done to the openshift installer to improve the user experience here but we
+were encourage to pursue manifests' overwrite as a way to provide this through assisted service.
+
+### Goals
+
+- Allow for a 4-nodes cluster to be created using the KubeAPI flow
+
+### Non-Goals
+
+- Implement logic that will decide, for the user, what the best architecture is. Having a 3+1
+cluster will require that masters are explicitly set to schedulable and the manifests to patch the
+Infrastructure CR are provided.
+
+- Expose this through SaaS at this stage. This would require changes in the UI and other validations
+that may not worth the effort at this stage.
+
+## Proposal
+
+The biggest changes anticipated for this enhancement are:
+
+- Expose the manifests API through the InfraEnv CRD so that custom manifests can be added through
+the ZTP flow.
+
+- Expose the schedulable mastes through the AgentClusterInstall to allow for explicitly creating
+schedulable masters.
+
+- Modify the `enoughMastersAndWorkers` validation so that it allows 1 worker when masters are
+scheulable. [PoC here](https://github.com/flaper87/assisted-service/commit/f07fc6589e4f2d316266ac5533d76bd2faf471d7)
+and [here](https://github.com/flaper87/assisted-service/commit/bb2d82eaacaee7790d2f18711e2ba3a442a4967d)
+
+- Modify Assisted Service's manifests logic to allow for partial manifests to be provided. Openshift
+installer applies the provided manifests in sequence. If 2 manifests for the same CR are provided,
+then the one applied last will be the one used. This is a problem when the user only wants to modify
+1 field. Here's an [example from a PoC](https://github.com/flaper87/assisted-service/commit/c6d72827b43786f95e68e052623e8aef0d4b4c0f).
+
+### User Stories
+
+#### Story 1
+
+As an Infrastructure Admin, I want to be able to deploy a 4 nodes cluster as a single operation the
+same way I deploy a full cluster. Aware of the risks and downsides, I want to be able to still
+pursue this deployment as it fits best the hardware I have available.
+
+### Implementation Details/Notes/Constraints [optional]
+
+The current HTTP API allows to push full manifests. That is, a manifest that can be fed into `oc
+apply -f`. As part of this work, we need to have a way to add partial manifests and, for this, we
+need to decide whether we want to accept yaml-patch files or just regular yaml files. Here are a
+couple of examples:
+
+Partially populated manifest to overwrite the `status.infrastructureTopology` field:
+
+```yaml
+status:
+  infrastructureTopology: HighlyAvailable
+```
+
+yaml-patch version for the above:
+
+```yaml
+- op: add
+  path: /status/infrastructureTopology
+  value: HighlyAvailable
+```
+
+In the first case, the current API can be re-used. Assisted service would be changed to not overrite
+files but rather merge the manifests. This is how [this PoC was
+implemented](https://github.com/flaper87/assisted-service/commit/c6d72827b43786f95e68e052623e8aef0d4b4c0f).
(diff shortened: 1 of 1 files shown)


docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+enclosure that only had 4 nodes. The goal was to use all of the blades rather than just 3 of them.
+
+For now, this is achieved with a 2 step deployment, first 3 masters and then add a worker node. It
+would be more time efficient as well as easier and more reliable to be able to do this deployment in
+one step, day0.
+
+# Motivation
+
+Assisted service has built-in validations that prevent the deployment of a 4-nodes cluster. This is
+to prevent unhealthy clusters to be deployed. A 4-nodes cluster ends up in an unhealthy state
```
omertuc, line 26: Calling them "unhealthy" challenges my understanding - I expect the clusters to be completely healthy and finish installation successfully. 

They just have single replicas infrastructure topology, which has pros and cons, but I wouldn't call them unhealthy.

Is this correct, or am I missing something?

If I'm correct, then it means the current assisted-service validations for preventing 4-node clusters are not justified and should be removed, regardless of this enhancement

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+that may not worth the effort at this stage.
+
+## Proposal
+
+The biggest changes anticipated for this enhancement are:
+
+- Expose the manifests API through the InfraEnv CRD so that custom manifests can be added through
+the ZTP flow.
+
+- Expose the schedulable mastes through the AgentClusterInstall to allow for explicitly creating
```
omertuc, line 59:
```suggestion
- Expose the schedulable masters through the AgentClusterInstall to allow for explicitly creating
```

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+Infrastructure CR are provided.
+
+- Expose this through SaaS at this stage. This would require changes in the UI and other validations
+that may not worth the effort at this stage.
+
+## Proposal
+
+The biggest changes anticipated for this enhancement are:
+
+- Expose the manifests API through the InfraEnv CRD so that custom manifests can be added through
```
omertuc, line 56: A single infraenv could be used to discover hosts for many different clusters, so I think AgentClusterInstall is a better fit

omertuc: Also I think we already have such API: manifestsConfigMapRefs

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+workers being deployed. The problem with schedulable masters not being counted as workers causes the
+installer to chose an `InfrastructureTopology` that is SingleReplica. This means that infrastructure
+workloads will run as single replicas, only on workers. This setting cannot be changed once the
+cluster is deployed.
+
+The above was brought up to the OpenShift team as [an enhancement](https://github.com/openshift/enhancements/pull/1057/). The feedback so far is that
+it's OK for users (or other installers) to overwrite the openshift manifests to achieve the
+deployment topologies required. The conversation is still in progress as far as understanding if
+some changes could be done to the openshift installer to improve the user experience here but we
+were encourage to pursue manifests' overwrite as a way to provide this through assisted service.
```
avishayt, line 37:
```suggestion
were encouraged to pursue manifests' overwrite as a way to provide this through assisted service.
```

docs/enhancements/4-nodes-cluster-deployment.md
omertuc: https://github.com/openshift/assisted-service/blob/e574bc2de2fd3153c7454604e32d7ee7664c1d78/api/hiveextension/v1beta1/agentclusterinstall_types.go#L120-L125

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+The biggest changes anticipated for this enhancement are:
+
+- Expose the manifests API through the InfraEnv CRD so that custom manifests can be added through
+the ZTP flow.
+
+- Expose the schedulable mastes through the AgentClusterInstall to allow for explicitly creating
+schedulable masters.
+
+- Modify the `enoughMastersAndWorkers` validation so that it allows 1 worker when masters are
+scheulable. [PoC here](https://github.com/flaper87/assisted-service/commit/f07fc6589e4f2d316266ac5533d76bd2faf471d7)
```
omertuc, line 63:
```suggestion
schedulable. [PoC here](https://github.com/flaper87/assisted-service/commit/f07fc6589e4f2d316266ac5533d76bd2faf471d7)
```

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+some changes could be done to the openshift installer to improve the user experience here but we
+were encourage to pursue manifests' overwrite as a way to provide this through assisted service.
+
+### Goals
+
+- Allow for a 4-nodes cluster to be created using the KubeAPI flow
+
+### Non-Goals
+
+- Implement logic that will decide, for the user, what the best architecture is. Having a 3+1
```
avishayt, line 45: Do you mean 3+1 vs. SNO+3 workers?  So the user will need to specify the roles explicitly instead of using auto-assign?

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+cluster will require that masters are explicitly set to schedulable and the manifests to patch the
+Infrastructure CR are provided.
+
+- Expose this through SaaS at this stage. This would require changes in the UI and other validations
+that may not worth the effort at this stage.
```
avishayt, line 50: The plan is to check the deployment type in the validation code?  I'd prefer to have fewer differences between deployment types as possible - maybe we could expose it in the SaaS, or just disable it in the UI?

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+- Add a basic ZTP job that deploys a 4-nodes cluster. Alternatively, we could modify the existing
+job and have it always deploy 4 nodes, which would cover multiple scenarios at once.
+
+## Drawbacks
+
+N/A
+
+## Alternatives
+
+- Don't expose the manifests and just make the decision for the user. Openshift Installer, as well
```
omertuc, line 150: Another middleground alternative:
- Give the user a new, simple assisted API (true/false) that will let the user control the behavior here, and assisted will do the heavy lifting of making the correct manifest adjustments

docs/enhancements/4-nodes-cluster-deployment.md
avishayt, line 59:
```suggestion
- Expose the schedulable masters through the AgentClusterInstall to allow for explicitly creating
```

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+
+- Expose this through SaaS at this stage. This would require changes in the UI and other validations
+that may not worth the effort at this stage.
+
+## Proposal
+
+The biggest changes anticipated for this enhancement are:
+
+- Expose the manifests API through the InfraEnv CRD so that custom manifests can be added through
+the ZTP flow.
```
avishayt, line 57: Why is this necessary?

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+
+N/A
+
+## Alternatives
+
+- Don't expose the manifests and just make the decision for the user. Openshift Installer, as well
+as assisted service, make similar decisions when there are just 3 masters. Masters, when there are
+no workers, are automatically set to schedulable. Similar behavior could be implemented for a 3+1
+architecture where, in addition to setting masters as schedulable, assisted would also set the
+`infrastructureTopology` to `HighlyAvailable`.
```
avishayt, line 154: I think this alternative is better (simpler development and more straightforward UX).  And I would also expose it in the SaaS, marking the cluster as "unsupported" in the UI.

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+- Expose the schedulable mastes through the AgentClusterInstall to allow for explicitly creating
+schedulable masters.
+
+- Modify the `enoughMastersAndWorkers` validation so that it allows 1 worker when masters are
+scheulable. [PoC here](https://github.com/flaper87/assisted-service/commit/f07fc6589e4f2d316266ac5533d76bd2faf471d7)
+and [here](https://github.com/flaper87/assisted-service/commit/bb2d82eaacaee7790d2f18711e2ba3a442a4967d)
+
+- Modify Assisted Service's manifests logic to allow for partial manifests to be provided. Openshift
+installer applies the provided manifests in sequence. If 2 manifests for the same CR are provided,
+then the one applied last will be the one used. This is a problem when the user only wants to modify
```
omertuc, line 68: Could you elaborate on what you mean exactly by "applied last"? Usually manifest apply ordering is done with `xx_` numerical prefixes to the manifest file names, maybe we can have a similar scheme here as well?

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+the ZTP flow.
+
+- Expose the schedulable mastes through the AgentClusterInstall to allow for explicitly creating
+schedulable masters.
+
+- Modify the `enoughMastersAndWorkers` validation so that it allows 1 worker when masters are
+scheulable. [PoC here](https://github.com/flaper87/assisted-service/commit/f07fc6589e4f2d316266ac5533d76bd2faf471d7)
+and [here](https://github.com/flaper87/assisted-service/commit/bb2d82eaacaee7790d2f18711e2ba3a442a4967d)
+
+- Modify Assisted Service's manifests logic to allow for partial manifests to be provided. Openshift
```
omertuc, line 66: I personally understand the motivation for partial manifests because I've seen the discussions on the previous enhancement, but could be nice if you laid out the motivations for this in this enhancement, as I don't feel they're obvious

docs/enhancements/4-nodes-cluster-deployment.md
omertuc: No, `SNO + any non-zero amount of workers` on day1 is not something we currently support at all, not even after this enhancement

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+
+yaml-patch version for the above:
+
+```yaml
+- op: add
+  path: /status/infrastructureTopology
+  value: HighlyAvailable
+```
+
+In the first case, the current API can be re-used. Assisted service would be changed to not overrite
```
omertuc, line 101:
```suggestion
In the first case, the current API can be re-used. Assisted service would be changed to not overwrite
```

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+files but rather merge the manifests. This is how [this PoC was
```
omertuc, line 102: Another interesting idea worth exploring: Let the user control whether they want a full overwrite vs just simple merge via annotations on the configmap.


No annotation and we do full overwrite as we always did, to ensure backwards compatibility.
With annotation, the user can choose between overwrite merge.

omertuc: Since the API discoverability of annotations is rather poor, we should maybe consider moving from using primitive configmaps to our own specialized CRD - this way users will be able to control the merge-vs-overwrite behavior with proper CRD fields with good descriptions, rather than some obscure annotation that they only find out about through external docs

omertuc: :man_facepalming: I see the lines right below this comment mention something similar, but with yaml-patch

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+
+The 3+1 topology is not part of OCP's test matrix. This, to some extent, makes it a
+non-supported topology. However, it has been accepted as a valid topology for the ZTPFW use-case.
+
+The reason for the above is that a 4-nodes cluster that uses masters for infrastructure workload
+would be sending infrastructure traffic to the master nodes. This, depending on the usage of the
+cluster, may result in master nodes being overloaded and, in some cases, even unavailable if the
+load is too high.
+
+Using a 4-node cluster must be a concious decision and it requires an understanding of what
```
omertuc, line 119:
```suggestion
Using a 4-node cluster must be a conscious decision and it requires an understanding of what
```

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+workloads will be run on the cluster. This topology is not much different to a compact cluster where
+there are only 3 masters. By exposing these options through the KubeAPI we will be allowing users to
+create such topologies.
+
+Finally, by exposing these options, we would also allow users to overwrite any manifest through the
```
omertuc, line 124: This is a good point, but this is not a new risk - we already give users the ability to overwrite / add complete manifests. If anything, patching support will make it more surgical with a smaller blast radius

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,154 @@
...
+job and have it always deploy 4 nodes, which would cover multiple scenarios at once.
+
+## Drawbacks
+
+N/A
+
+## Alternatives
+
+- Don't expose the manifests and just make the decision for the user. Openshift Installer, as well
+as assisted service, make similar decisions when there are just 3 masters. Masters, when there are
```
omertuc, line 151: >make similar decisions when there are just 3 masters

I don't think it's fair to call this a decision. Making masters schedulable when there are just 3 nodes overall is a no-brainer - without it you will not be able to schedule user workloads at all. There is no alternative to making them schedulable.

But with 3+1 it's not so obvious we want schedulable masters. There is another worker, it can handle user / infra workloads on its own, and the decision of whether the masters should be schedulable is not so obvious in that case and is opinionated

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,160 @@
...
+  - "@flaper87"
+creation-date: 2022-04-14
+last-updated: yyyy-mm-dd
+---
+
+# Allow for a 4-nodes cluster to be created
+
+# Summary
+
+**Note:** Throughout this enhancement, a 4-nodes cluster refers to a 3 schedulable masters and 1 worker.
```
omertuc, line 13:
```suggestion
**Note:** Throughout this enhancement, a 4-nodes cluster refers to a 3 schedulable masters and 1 worker cluster.
```

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,160 @@
...
+
+There are cases where customers may just have 4 nodes available and no chance to add more. This was
+one of the cases in the ZTP for Factory Workflow where it was required to deploy a cluster in an
+enclosure that only had 4 nodes. The goal was to use all of the blades rather than just 3 of them.
+
+For now, this is achieved with a 2 step deployment, first 3 masters and then add a worker node. It
+would be more time efficient as well as easier and more reliable to be able to do this deployment in
+
```
omertuc, line 21:
```suggestion
```

docs/enhancements/4-nodes-cluster-deployment.md
omertuc: Alright, so please add a small remark that emphasizes that this is not a new risk

docs/enhancements/4-nodes-cluster-deployment.md
avishayt: How about:
- If masters are schedulable, automatically set InfrastructureTopology to HighlyAvailable.
- If masters are not schedulable, don't allow it as today (just need to reword the validation to let the user know they can proceed by setting schedulable masters).

It looks like we're jumping through hoops with merging manifests for a use case that might not be interesting.  Using 3 hosts for a control plane just to run containers on one host seems like a huge waste that few would be interested in (better to just use microshift or even podman).

flaper87: > It looks like we're jumping through hoops with merging manifests for a use case that might not be interesting.

I think the merging of manifests is something that we should do regardless. The way custom manifests work right now is certainly very limited and prone to issues given that users have to provide full blown manifests. The UX here is poor, IMHO.

> If masters are schedulable, automatically set InfrastructureTopology to HighlyAvailable.

This was my [original proposal here](https://github.com/openshift/enhancements/pull/1057) but after more thoughts and conversations, I think setting `InfrastructureTopology` to `HighlyAvailable` may be too opinionated and we will do a disservice to our users.

I think it would be better for us to remove the single worker validation (which will align AI with the way the OCP installer works) and then allow the rest to be customized through manifests. Note that AI sets masters as schedulable automatically when there are no workers.

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,159 @@
...
+cluster will require that masters are explicitly set to schedulable and the manifests to patch the
+Infrastructure CR are provided.
+
+- Expose this through SaaS at this stage. This would require changes in the UI and other validations
+that may not worth the effort at this stage.
+
+## Proposal
+
+The biggest changes anticipated for this enhancement are:
+
```
eranco74, line 54: You probably want to change the logic [here](https://github.com/openshift/installer/blob/2d195fb9a0fc7d291a519772a51cf150b82107ad/pkg/asset/manifests/topologies.go#L23)

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,159 @@
...
+For now, this is achieved with a 2 step deployment, first 3 masters and then add a worker node. It
+would be more time efficient as well as easier and more reliable to be able to do this deployment in
+one step, day0.
+
+# Motivation
+
+Assisted service has built-in validations that prevent the deployment of a 4-nodes cluster. This is
+to prevent the deployment of topologies that are not part of the OCP standard test matrix. A 4-nodes
+cluster doesn't take schedulable masters into consideration when there are workers being deployed,
+which causes the installer to chose an `InfrastructureTopology` that is SingleReplica. This means
```
eranco74, line 28: This isn't accurate, for example with the current way we install 4 nodes cluster (3 control-plane nodes, add a worker) the ingress will get scheduled on the control-plane (@omertuc correct me if I'm wrong here)

eranco74: @omertuc do you know what will happen to the ingress deployment in case the masters are schedulable and the InfrastructureTopology is highly available but there is a single worker node?

omertuc: >@omertuc do you know what will happen to the ingress deployment in case the masters are schedulable and the InfrastructureTopology is highly available but there is a single worker node?

Yes, Ingress will have 2 replicas and they will be randomly scheduled on 2 of the 4 nodes (regardless of being control-plane or true worker nodes)

docs/enhancements/4-nodes-cluster-deployment.md
flaper87: @jelkosz and I chatted about this and his feedback is that the work needed in the UI is not much, it should be possible to do it in the required timeframe.

I will update the current PR and ask the QE team to run tests using that PR to make sure it doesn't break the UI. This will allow us to do a release, have the API available, and give the UI team enough time to do the changes on their end.

docs/enhancements/4-nodes-cluster-deployment.md
flaper87: That was the original plan but, instead, we are modifying the manifest as suggested by the OCP team.

flaper87 (commented):
Can we move this enhancement forward if there are no other comments/concerns? 

I will be updating the PR in the following days.

docs/enhancements/4-nodes-cluster-deployment.md
```diff
@@ -0,0 +1,117 @@
...
+
+- Expose this through SaaS at this stage. This would require changes in the UI and other validations
+that may not worth the effort at this stage.
+
+## Proposal
+
+The biggest changes anticipated for this enhancement are:
+
+- Modify Assisted Service so that 3 nodes will be set as schedulable masters and 1 as worker, whenever a 4-node cluster
+is created.
```
avishayt, line 52: (as long as the cluster isn't SNO)

docs/enhancements/4-nodes-cluster-deployment.md
avishayt: We need to warn in the UI that it's unsupported.  We'll need a design for that.

avishayt (approved):

PR #3677, opened 2022-04-18
MGMT-9787: Add the built-in ipxe script endpoint to baremetal deploy docs

carbonin:

https://issues.redhat.com/browse/MGMT-9787

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @javipolo 
/cc @CrystalChun 

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/user-guide/deploy-on-bare-metal.md
+++ b/docs/user-guide/deploy-on-bare-metal.md
@@ -38,91 +38,26 @@ _*Note*: This method could be used also in Virtual environment_
 
 ## iPXE
 
-iPXE deployment method
+*NOTE*: We use a sample URL, please change to fit your use case accordingly
 
-*NOTE1*: We use a sample URL, please change to fit your use case accordingly
-*NOTE2*: We've set the live_url as the node hostname on 8080 port , please change to fit your use case accordingly
+### Accessing the iPXE boot script and artifacts
 
-### Automatic
-
-The automatic way is done using podman, just follow this steps:
-
-```shell
-IPXE_DIR=/tmp/ipxe/ai
-mkdir -p ${IPXE_DIR}
-
-# This command will download the ISO, extract the Images and create the ignition config files
-podman run -e BASE_URL=http://devscripts2ipv6.e2e.bos.redhat.com:8080 -e ISO_URL=$(curl http://devscripts2ipv6.e2e.bos.redhat.com:6008/api/assisted-install/v2/infra-envs/29b516fd-8f3a-42bf-8a59-cb20ef2630e0/downloads/image-url | jq ".url") -v /tmp/ipxe/ai:/data:Z --net=host -it --rm quay.io/ohadlevy/ai-ipxe
-
-# This command will host the iPXE files on an podman container
-podman run  -v ${IPXE_DIR}:/app:ro -p 8080:8080 -d --rm docker.io/bitnami/nginx:latest
+The service serves an iPXE boot script for each infra-env
+The script can be downloaded using:
 ```
-
-To ensure if your container is working fine, check the url with a `curl` command
-
-```shell
-curl http://$(hostname):8080/ipxe
+GET /api/assisted-install/v2/infra-envs/{infra_env_id}/downloads/files?file_name=ipxe-script
 ```
+This URL can either be used directly or the script can be downloaded and hosted separately.
 
-### Manual
-
-The manual way is explained here. You need at least to have the Discovery ISO already generated
-
-Now let's download that ISO in the provisioning machine, where the iPXE files will be hosted (use the _Command to download the ISO_ button from the Assisted Service website
-
-```shell
-IPXE_DIR=/tmp/ipxe/ai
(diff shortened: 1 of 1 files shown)


docs/user-guide/deploy-on-bare-metal.md
```diff
@@ -38,91 +38,21 @@ _*Note*: This method could be used also in Virtual environment_
...
-```
-
-- To ensure if your container is working fine, check the url with a `curl` command
-
-```shell
-curl http://$(hostname):8080/ipxe
-```
+This URL can either be used directly or the script can be downloaded and hosted separately
 
 ### Booting the nodes from iPXE
```
carbonin, line 57: I'm not sure if it's worth changing this here.
It's good information about how to set the script url, but the actual url doesn't match with the example I added.

@javipolo @CrystalChun do you think it's worth re-doing the demo/recording for a few screenshots?

CrystalChun: If we were to redo it then would it be the same flow displayed in the screenshots or would we show a different flow? If it's the same flow, then maybe a note at the top about using correct urls would be sufficient?

docs/user-guide/deploy-on-bare-metal.md
```diff
@@ -38,91 +38,21 @@ _*Note*: This method could be used also in Virtual environment_
...
 
-iPXE deployment method
+*NOTE*: We use a sample URL, please change to fit your use case accordingly
 
-*NOTE1*: We use a sample URL, please change to fit your use case accordingly
-*NOTE2*: We've set the live_url as the node hostname on 8080 port , please change to fit your use case accordingly
+### Accessing the iPXE boot script and artifacts
 
-### Automatic
+The service serves an iPXE boot script for each infra-env
```
CrystalChun, line 45: It might be nice to link to more info about iPXE - maybe the [enhancement](https://github.com/openshift/assisted-service/blob/master/docs/enhancements/ipxe-host-boot.md) doc?

docs/user-guide/deploy-on-bare-metal.md
```diff
@@ -38,91 +38,21 @@ _*Note*: This method could be used also in Virtual environment_
 
 ## iPXE
 
-iPXE deployment method
+*NOTE*: We use a sample URL, please change to fit your use case accordingly
```
CrystalChun, line 41: We can link to the [enhancement](https://github.com/openshift/assisted-service/blob/master/docs/enhancements/ipxe-host-boot.md) doc to show the URL format

docs/user-guide/deploy-on-bare-metal.md
carbonin: :man_shrugging: there's a lot of things we could document.
This shows adding the url to the console directly rather than configuring DHCP, but I thought people would generally be using DHCP.

The steps after entering the script address don't seem specific to ipxe though

docs/user-guide/deploy-on-bare-metal.md
carbonin: I think I'd rather just put the important details in here. Maybe how to get the ipxe download url since it's a separate endpoint.

docs/user-guide/deploy-on-bare-metal.md
carbonin: I don't know if we want to link to enhancements from user docs. They have a lot of extra information that a user really wouldn't need and would have to sift through.

What exactly do you think we should add here? Maybe what endpoint the script is served on?

CrystalChun: That makes sense that it would be too much information. 

Yes, maybe the format of the endpoint like the one in the enhancement `GET /v2/infra-envs/{infra-env-id}/downloads/files?file_name=ipxe-script` so that way the user knows which parts of the example URL they need to modify

avishayt:
Have all comments been addressed?

docs/user-guide/deploy-on-bare-metal.md
javipolo: I think that just adding a note should be enough. Something saying that now it's not needed to download the ISO and extract the artifacts, since you can just copy-paste the ipxe script we provide and that's all

carbonin: I think I handled that.
Ideally a user wouldn't even need to take the script, just point their environment to the script endpoint we expose, right?

carbonin:
> Have all comments been addressed?

I think so, would like @javipolo to confirm I handled what he's talking about though.

docs/user-guide/deploy-on-bare-metal.md
javipolo: That's correct :)

PR #3696, opened 2022-04-22
NO-ISSUE: create build directory for generate targets

adriengentil:
The target that creates the build directory is not available from the
included Makefile in `hack/` directory. As a result, the some generate*
commands fail to execute.

### DIFF
--- a/hack/Makefile
+++ b/hack/Makefile
@@ -10,6 +10,7 @@ generate-go-server:
 	goswagger generate server --template=stratoscale -f swagger.yaml
 
 generate-python-client:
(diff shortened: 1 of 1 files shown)


adriengentil:
For the context, `make generate-bundle` failed in this build: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-assisted-assisted-installer-deployment-master-bump-ocp/1517397814663974912

There is probablity a better solution, but I think we would need to reorganize the Makefiles.

osherdp (approved):

PR #3698, opened 2022-04-23
Bug 2077921: Operator deployment script doesn't wait after patching the Provisioning CR

omertuc:
# Change 1:

Patching the Provisioning CR triggers a rollout of the metal3
deployment. Since the metal3 deployment has just 1 replica, during the
rollout the metal admission webhook is unavailable. As a result any
interaction during that time with BMH resources causes an error (as the
webhook is down).

This is why after patching the Provisioning CR we have to wait for that
process to complete. This includes waiting a few seconds for the
baremetal operator to notice our patch (there's no simple robust way to
do this. It usually happens almost immediately, but we wait 10 seconds
just in case). Then we wait for the deployment to finish rolling out,
and finally we wait for the metal3 pod to be ready.

# Change 2:

When applying the spoke cluster ZTP manifests generated by the ansible
playbook, we used this loop:

```bash
for manifest in $(find ${__dir}/generated -type f); do
    tee < "${manifest}" >(oc apply -f -)
done
```

The issue is that due to `>(oc apply -f -)` running in a subshell, `oc apply`
errors went unnoticed by the bash `errexit` / `pipefail` options and the
script quietly went on as if nothing happened. This made the issue
harder to notice and diagnose.

To fix this, the loop has been modified to use this more readable and
`errexit` / `pipefail` friendly form:

```bash
for manifest in $(find ${__dir}/generated -type f); do
    echo "Applying $manifest"
    cat "${manifest}"
    oc apply -f "${manifest}"
done
```

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [x] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @osherdp

## Checklist

## Reviewers Checklist

### DIFF
--- a/deploy/operator/setup_assisted_operator.sh
+++ b/deploy/operator/setup_assisted_operator.sh
@@ -211,6 +211,21 @@ EOCR
   echo "Enabling configuration of BMH resources outside of openshift-machine-api namespace"
   oc patch provisioning provisioning-configuration --type merge -p '{"spec":{"watchAllNamespaces": true}}'
 
+  # Patching the Provisioning CR triggers a rollout of the metal3 deployment.
+  # Since the metal3 deployment has just 1 replica, during the rollout the
+  # metal admission webhook is unavailable. As a result any interaction during
+  # that time with BMH resources causes an error (as the webhook is down).
+  #
+  # This is why before proceeding we have to wait for that process to complete.
+  # This includes waiting a few seconds for the baremetal operator to notice
+  # our patch (there's no simple robust way to do this. It usually happens
+  # almost immediately, but we wait 10 seconds just in case). Then we wait for
+  # the deployment to finish rolling out, and finally we wait for the metal3
+  # pod to be ready.
+  sleep 10 # Wait for the operator to notice our patch
+  timeout 15m oc rollout status -n openshift-machine-api deployment/metal3
+  oc wait --timeout=5m pod -n openshift-machine-api -l baremetal.openshift.io/cluster-baremetal-operator=metal3-state --for=condition=Ready
+
   echo "Installation of Assisted Installer operator passed successfully!"
 }
 
--- a/deploy/operator/ztp/deploy_spoke_cluster.sh
+++ b/deploy/operator/ztp/deploy_spoke_cluster.sh
@@ -80,7 +80,9 @@ oc get secret "${ASSISTED_PRIVATEKEY_NAME}" -n "${SPOKE_NAMESPACE}" || \
     oc create secret generic "${ASSISTED_PRIVATEKEY_NAME}" --from-file=ssh-privatekey=/root/.ssh/id_rsa --type=kubernetes.io/ssh-auth -n "${SPOKE_NAMESPACE}"
 
 for manifest in $(find ${__dir}/generated -type f); do
-    tee < "${manifest}" >(oc apply -f -)
+    echo "Applying $manifest"
+    cat "${manifest}"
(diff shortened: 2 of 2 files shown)


omertuc:
Waiting seems to work:
```

++ oc patch provisioning provisioning-configuration --type merge -p '{"spec":{"watchAllNamespaces": true}}'
provisioning.metal3.io/provisioning-configuration patched
++ sleep 10
++ oc rollout status -n openshift-machine-api deployment/metal3
Waiting for deployment "metal3" rollout to finish: 0 out of 1 new replicas have been updated...
Waiting for deployment "metal3" rollout to finish: 0 out of 1 new replicas have been updated...
Waiting for deployment "metal3" rollout to finish: 0 out of 1 new replicas have been updated...
Waiting for deployment "metal3" rollout to finish: 0 of 1 updated replicas are available...
deployment "metal3" successfully rolled out
++ oc wait pod -n openshift-machine-api -l baremetal.openshift.io/cluster-baremetal-operator=metal3-state --for=condition=Ready
pod/metal3-74dfd4c95c-2wbtw condition met
```

deploy/operator/setup_assisted_operator.sh
```diff
@@ -211,6 +211,21 @@ EOCR
...
+  # that time with BMH resources causes an error (as the webhook is down).
+  #
+  # This is why before proceeding we have to wait for that process to complete.
+  # This includes waiting a few seconds for the baremetal operator to notice
+  # our patch (there's no simple robust way to do this. It usually happens
+  # almost immediately, but we wait 10 seconds just in case). Then we wait for
+  # the deployment to finish rolling out, and finally we wait for the metal3
+  # pod to be ready.
+  sleep 10 # Wait for the operator to notice our patch
+  oc rollout status -n openshift-machine-api deployment/metal3
```
osherdp, line 226: Can we add some timeout for this action?
I guess it should at least finish in let's say a couple of minutes at most

It's both for the user experience of running this test, for having shorter times in cases of errors, and so that we'll fast-fail and have artifacts truncated with the exact logs responsible for the problem

deploy/operator/setup_assisted_operator.sh
```diff
@@ -211,6 +211,21 @@ EOCR
...
+  oc wait pod -n openshift-machine-api -l baremetal.openshift.io/cluster-baremetal-operator=metal3-state --for=condition=Ready
```
osherdp, line 227: same question as above

osherdp (changes requested):

osherdp:
The static sleeping is not ideal I guess, but I get how it's not that easy doing anything else with options that ``kubectl wait`` currently provides

deploy/operator/setup_assisted_operator.sh
omertuc: `oc wait` defaults to a 30 second timeout

deploy/operator/setup_assisted_operator.sh
omertuc: Good point

deploy/operator/setup_assisted_operator.sh
omertuc: I'll make it explicit

osherdp (approved):

omertuc:
Comentse fixed & CI seems to be passing

PR #3705, opened 2022-04-25
Bug 2078531: Expose ipxe boot artifacts via HTTP

vrutkovs:
This adds a new HTTP route to operator to serve kernel/rootfs artifacts for IPXE deployments. iPXE deployments can't access HTTPS endpoints, so we should have a plain HTTP endpoint to serve these artifacts.

Note that it doesn't apply to console.redhat.com deployment - users are expected to copy those internally and host them via http themselves.

TODO:
* [x] Create or find a matching issue
* [x] Expose `pxe-initrd` in `/boot-artifacts` (needs assisted-image-service)
* [x] (assisted-image-service) expose `boot-artifacts` via insecure port
* [x] Fix tests, add new tests
* [x] Add a wrapper around http.Server and add unit tests
* [x] Reuse `servers` from image-service (hopefully implementation is correct)
* [x] Use env var instead of cli arg
* [x] Use middleware to limit access instead of Route path
* [x] Operator should expose http route only when ASC annotation is set
* [x] Re-import upstream assisted-image-service
* [x] Rename annotation 
* [ ] Update InfraEnv status URL scheme when annotation changes (minor bug?)

- [x] Enhancement 

## How was this code tested?

## Assignees

/cc @carbonin 
/cc @CrystalChun 

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/v1beta1/agentserviceconfig_types.go
+++ b/api/v1beta1/agentserviceconfig_types.go
@@ -88,6 +88,18 @@ type AgentServiceConfigSpec struct {
 	// that are used if one the operators fails to be successfully deployed
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Must-Gather Images"
 	MustGatherImages []MustGatherImage `json:"mustGatherImages,omitempty"`
+
+	// IPXEHTTPRoute is controlling whether the operator is creating plain HTTP routes
+	// iPXE hosts may not work with router cyphers and may access artifacts via HTTP only
+	// This setting accepts "enabled,disabled", defaults to disabled. Empty value defaults to disabled
+	// The following endpoints would be exposed via http:
+	// * api/assisted-installer/v2/infra-envs/<id>/downloads/files?file_name=ipxe-script in assisted-service
+	// * boot-artifacts/ and images/<infra-enf id>/pxe-initrd in -image-service
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Expose IPXE HTTP route"
+	// +kubebuilder:validation:Enum=enabled;disabled
+	// +kubebuilder:validation:default:=disabled
+	// +optional
+	IPXEHTTPRoute string `json:"iPXEHTTPRoute,omitempty"`
 }
 
 // ConditionType related to our reconcile loop in addition to all the reasons
@@ -140,6 +152,11 @@ const (
 	ReasonWebHookServiceAccountFailure string = "ReasonWebHookServiceAccountFailure"
 	// ReasonWebHookAPIServiceFailure when there was a failure related to the webhook's API service.
 	ReasonWebHookAPIServiceFailure string = "ReasonWebHookAPIServiceFailure"
+
+	// IPXEHTTPRouteEnabled is expected value in IPXEHTTPRoute to enable the route
+	IPXEHTTPRouteEnabled string = "enabled"
+	// IPXEHTTPRouteEnabled is expected value in IPXEHTTPRoute to disable the route
+	IPXEHTTPRouteDisabled string = "disabled"
 )
 
 // AgentServiceConfigStatus defines the observed state of AgentServiceConfig
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,13 +8,16 @@ import (
 	"net/http"
 	_ "net/http/pprof"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
 	"github.com/kelseyhightower/envconfig"
+	"github.com/openshift/assisted-image-service/pkg/servers"
 	"github.com/openshift/assisted-service/internal/bminventory"
 	"github.com/openshift/assisted-service/internal/cluster"
 	"github.com/openshift/assisted-service/internal/cluster/validations"
@@ -137,6 +140,7 @@ var Options struct {
 	WorkDir                        string        `envconfig:"WORK_DIR" default:"/data/"`
 	LivenessValidationTimeout      time.Duration `envconfig:"LIVENESS_VALIDATION_TIMEOUT" default:"5m"`
 	ApproveCsrsRequeueDuration     time.Duration `envconfig:"APPROVE_CSRS_REQUEUE_DURATION" default:"1m"`
+	HTTPListenPort                 string        `envconfig:"HTTP_LISTEN_PORT" default:""`
 }
 
 func InitLogs() *logrus.Entry {
@@ -555,12 +559,21 @@ func main() {
 		}
 	}()
 
-	address := fmt.Sprintf(":%s", swag.StringValue(port))
-	if Options.ServeHTTPS {
-		log.Fatal(http.ListenAndServeTLS(address, Options.HTTPSCertFile, Options.HTTPSKeyFile, h))
-	} else {
-		log.Fatal(http.ListenAndServe(address, h))
+	// Interrupt servers on SIGINT/SIGTERM
+	stop := make(chan os.Signal, 1)
+	signal.Notify(stop, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+
+	// Run listen on http and https ports if HTTPSCertFile/HTTPSKeyFile set
+	serverInfo := servers.New(Options.HTTPListenPort, swag.StringValue(port), Options.HTTPSKeyFile, Options.HTTPSCertFile)
+	if serverInfo.HTTP != nil {
+		serverInfo.HTTP.Handler = h
 	}
+	if serverInfo.HTTPS != nil {
+		serverInfo.HTTPS.Handler = h
+	}
+	serverInfo.ListenAndServe()
+	<-stop
+	serverInfo.Shutdown()
 }
 
 func generateAPMTransactionName(request *http.Request) string {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -565,6 +565,9 @@ func main() {
 
 	// Run listen on http and https ports if HTTPSCertFile/HTTPSKeyFile set
 	serverInfo := servers.New(Options.HTTPListenPort, swag.StringValue(port), Options.HTTPSKeyFile, Options.HTTPSCertFile)
+	if serverInfo.HasBothHandlers {
+		h = app.WithIPXEScriptMiddleware(h)
+	}
 	if serverInfo.HTTP != nil {
 		serverInfo.HTTP.Handler = h
 	}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -423,10 +423,14 @@ func main() {
 		}
 	}
 
+	// Determine if IPXE artifact URLs need to be http
+	serverInfo := servers.New(Options.HTTPListenPort, swag.StringValue(port), Options.HTTPSKeyFile, Options.HTTPSCertFile)
+	generateInsecureIPXEURLs := serverInfo.HasBothHandlers
+
 	bm := bminventory.NewBareMetalInventory(db, log.WithField("pkg", "Inventory"), hostApi, clusterApi, infraEnvApi, Options.BMConfig,
 		generator, eventsHandler, objectHandler, metricsManager, usageManager, operatorsManager, authHandler, authzHandler, ocpClient, ocmClient,
 		lead, pullSecretValidator, versionHandler, crdUtils, ignitionBuilder, hwValidator, dnsApi, installConfigBuilder, staticNetworkConfig,
-		Options.GCConfig, providerRegistry)
+		Options.GCConfig, providerRegistry, generateInsecureIPXEURLs)
 
 	events := events.NewApi(eventsHandler, logrus.WithField("pkg", "eventsApi"))
 
@@ -501,6 +505,7 @@ func main() {
 				ImageServiceBaseURL: Options.BMConfig.ImageServiceBaseURL,
 				AuthType:            Options.Auth.AuthType,
 				VersionsHandler:     versionHandler,
+				InsecureIPXEURLs:    generateInsecureIPXEURLs,
 			}).SetupWithManager(ctrlMgr), "unable to create controller InfraEnv")
 
 			failOnError((&controllers.ClusterDeploymentsReconciler{
@@ -564,8 +569,7 @@ func main() {
 	signal.Notify(stop, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 
 	// Run listen on http and https ports if HTTPSCertFile/HTTPSKeyFile set
-	serverInfo := servers.New(Options.HTTPListenPort, swag.StringValue(port), Options.HTTPSKeyFile, Options.HTTPSCertFile)
-	if serverInfo.HasBothHandlers {
+	if generateInsecureIPXEURLs {
 		h = app.WithIPXEScriptMiddleware(h)
 	}
 	if serverInfo.HTTP != nil {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -425,7 +425,7 @@ func main() {
 
 	// Determine if IPXE artifact URLs need to be http
 	serverInfo := servers.New(Options.HTTPListenPort, swag.StringValue(port), Options.HTTPSKeyFile, Options.HTTPSCertFile)
-	generateInsecureIPXEURLs := serverInfo.HasBothHandlers
+	generateInsecureIPXEURLs := serverInfo.HTTP != nil
 
 	bm := bminventory.NewBareMetalInventory(db, log.WithField("pkg", "Inventory"), hostApi, clusterApi, infraEnvApi, Options.BMConfig,
 		generator, eventsHandler, objectHandler, metricsManager, usageManager, operatorsManager, authHandler, authzHandler, ocpClient, ocmClient,
@@ -568,8 +568,8 @@ func main() {
 	stop := make(chan os.Signal, 1)
 	signal.Notify(stop, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 
-	// Run listen on http and https ports if HTTPSCertFile/HTTPSKeyFile set
-	if generateInsecureIPXEURLs {
+	// Run listen on http and https ports if iPXE artifacts need to be exposed via HTTP
+	if serverInfo.HasBothHandlers {
 		h = app.WithIPXEScriptMiddleware(h)
 	}
 	if serverInfo.HTTP != nil {
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -222,6 +222,7 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
 		{"AgentService", aiv1beta1.ReasonAgentServiceFailure, r.newAgentService},
 		{"ServiceMonitor", aiv1beta1.ReasonAgentServiceMonitorFailure, r.newServiceMonitor},
 		{"ImageServiceRoute", aiv1beta1.ReasonImageHandlerRouteFailure, r.newImageServiceRoute},
+		{"ImageServiceIPXERoute", aiv1beta1.ReasonImageHandlerRouteFailure, r.newImageServiceIPXERoute},
 		{"AgentRoute", aiv1beta1.ReasonAgentRouteFailure, r.newAgentRoute},
 		{"AgentLocalAuthSecret", aiv1beta1.ReasonAgentLocalAuthSecretFailure, r.newAgentLocalAuthSecret},
 		{"DatabaseSecret", aiv1beta1.ReasonPostgresSecretFailure, r.newPostgresSecret},
@@ -508,8 +509,9 @@ func (r *AgentServiceConfigReconciler) newImageServiceService(ctx context.Contex
 			svc.ObjectMeta.Annotations = make(map[string]string)
 		}
 		svc.ObjectMeta.Annotations[servingCertAnnotation] = imageServiceName
-		if len(svc.Spec.Ports) == 0 {
-			svc.Spec.Ports = append(svc.Spec.Ports, corev1.ServicePort{})
+		if len(svc.Spec.Ports) != 2 {
+			svc.Spec.Ports = []corev1.ServicePort{}
+			svc.Spec.Ports = append(svc.Spec.Ports, corev1.ServicePort{}, corev1.ServicePort{})
 		}
 		svc.Spec.Ports[0].Name = imageServiceName
 		svc.Spec.Ports[0].Port = int32(imageHandlerPort.IntValue())
@@ -644,6 +646,61 @@ func (r *AgentServiceConfigReconciler) newImageServiceRoute(ctx context.Context,
 		route.Spec.Port = routeSpec.Port
 		route.Spec.WildcardPolicy = routeSpec.WildcardPolicy
 		route.Spec.TLS = routeSpec.TLS
+		return nil
+	}
+
+	return route, mutateFn, nil
+}
+
+func (r *AgentServiceConfigReconciler) newImageServiceIPXERoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
+	// Wait for https route to be created first
+	httpsRoute := &routev1.Route{}
+	if err := r.Get(ctx, types.NamespacedName{Name: imageServiceName, Namespace: r.Namespace}, httpsRoute); err != nil {
+		log.WithError(err).Error("Failed to get https route for image service")
+		return nil, nil, err
+	}
+
+	if httpsRoute.Spec.Host == "" {
+		log.Info("Image service https route found, but host not yet set")
+		return nil, nil, nil
+	}
+
+	weight := int32(100)
+	route := &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-ipxe", imageServiceName),
+			Namespace: r.Namespace,
+		},
+		Spec: routev1.RouteSpec{
+			Host: httpsRoute.Spec.Host,
+		},
+	}
+	routeSpec := routev1.RouteSpec{
+		Host: httpsRoute.Spec.Host,
+		To: routev1.RouteTargetReference{
+			Kind:   "Service",
+			Name:   imageServiceName,
+			Weight: &weight,
+		},
+		Port: &routev1.RoutePort{
+			TargetPort: intstr.FromString(fmt.Sprintf("%s-http", imageServiceName)),
+		},
+		WildcardPolicy: routev1.WildcardPolicyNone,
+	}
+
+	mutateFn := func() error {
+		if err := controllerutil.SetControllerReference(instance, route, r.Scheme); err != nil {
+			return err
+		}
+		// Only update what is specified above in routeSpec.
+		// If we update the entire route.Spec with
+		// route.Spec = routeSpec
+		// it would overwrite any existing values for route.Spec.Host
+		route.Spec.To = routeSpec.To
+		route.Spec.Port = routeSpec.Port
+		route.Spec.WildcardPolicy = routeSpec.WildcardPolicy
+		route.Spec.TLS = routeSpec.TLS
+		route.Spec.Path = routeSpec.Path
 		return nil
 	}
 
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -224,6 +224,7 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
 		{"ImageServiceRoute", aiv1beta1.ReasonImageHandlerRouteFailure, r.newImageServiceRoute},
 		{"ImageServiceIPXERoute", aiv1beta1.ReasonImageHandlerRouteFailure, r.newImageServiceIPXERoute},
 		{"AgentRoute", aiv1beta1.ReasonAgentRouteFailure, r.newAgentRoute},
+		{"AgentIPXERoute", aiv1beta1.ReasonAgentRouteFailure, r.newAgentIPXERoute},
 		{"AgentLocalAuthSecret", aiv1beta1.ReasonAgentLocalAuthSecretFailure, r.newAgentLocalAuthSecret},
 		{"DatabaseSecret", aiv1beta1.ReasonPostgresSecretFailure, r.newPostgresSecret},
 		{"ImageServiceServiceAccount", aiv1beta1.ReasonImageHandlerServiceAccountFailure, r.newImageServiceServiceAccount},
@@ -477,13 +478,18 @@ func (r *AgentServiceConfigReconciler) newAgentService(ctx context.Context, log 
 			svc.ObjectMeta.Annotations = make(map[string]string)
 		}
 		svc.ObjectMeta.Annotations[servingCertAnnotation] = serviceName
-		if len(svc.Spec.Ports) == 0 {
-			svc.Spec.Ports = append(svc.Spec.Ports, corev1.ServicePort{})
+		if len(svc.Spec.Ports) != 2 {
+			svc.Spec.Ports = []corev1.ServicePort{}
+			svc.Spec.Ports = append(svc.Spec.Ports, corev1.ServicePort{}, corev1.ServicePort{})
 		}
 		svc.Spec.Ports[0].Name = serviceName
 		svc.Spec.Ports[0].Port = int32(servicePort.IntValue())
 		svc.Spec.Ports[0].TargetPort = servicePort
 		svc.Spec.Ports[0].Protocol = corev1.ProtocolTCP
+		svc.Spec.Ports[1].Name = fmt.Sprintf("%s-http", serviceName)
+		svc.Spec.Ports[1].Port = int32(serviceHTTPPort.IntValue())
+		svc.Spec.Ports[1].TargetPort = serviceHTTPPort
+		svc.Spec.Ports[1].Protocol = corev1.ProtocolTCP
 		svc.Spec.Selector = map[string]string{"app": serviceName}
 		svc.Spec.Type = corev1.ServiceTypeClusterIP
 		return nil
@@ -593,6 +599,59 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
 		},
 		WildcardPolicy: routev1.WildcardPolicyNone,
 		TLS:            &routev1.TLSConfig{Termination: routev1.TLSTerminationReencrypt},
+	}
+
+	mutateFn := func() error {
+		if err := controllerutil.SetControllerReference(instance, route, r.Scheme); err != nil {
+			return err
+		}
+		// Only update what is specified above in routeSpec.
+		// If we update the entire route.Spec with
+		// route.Spec = routeSpec
+		// it would overwrite any existing values for route.Spec.Host
+		route.Spec.To = routeSpec.To
+		route.Spec.Port = routeSpec.Port
+		route.Spec.WildcardPolicy = routeSpec.WildcardPolicy
+		route.Spec.TLS = routeSpec.TLS
+		return nil
+	}
+
+	return route, mutateFn, nil
+}
+
+func (r *AgentServiceConfigReconciler) newAgentIPXERoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
+
+	// Wait for https route to be created first
+	httpsRoute := &routev1.Route{}
+	if err := r.Get(ctx, types.NamespacedName{Name: serviceName, Namespace: r.Namespace}, httpsRoute); err != nil {
+		log.WithError(err).Error("Failed to get https route for agent service")
+		return nil, nil, err
+	}
+	if httpsRoute.Spec.Host == "" {
+		log.Info("Agent https route found, but host not yet set")
+		return nil, nil, nil
+	}
+
+	weight := int32(100)
+	route := &routev1.Route{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-ipxe", serviceName),
+			Namespace: r.Namespace,
+		},
+		Spec: routev1.RouteSpec{
+			Host: httpsRoute.Spec.Host,
+		},
+	}
+	routeSpec := routev1.RouteSpec{
+		To: routev1.RouteTargetReference{
+			Kind:   "Service",
+			Name:   serviceName,
+			Weight: &weight,
+		},
+		Port: &routev1.RoutePort{
+			TargetPort: intstr.FromString(fmt.Sprintf("%s-http", serviceName)),
+		},
+		WildcardPolicy: routev1.WildcardPolicyNone,
 	}
 
 	mutateFn := func() error {
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -619,23 +619,22 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
 	return route, mutateFn, nil
 }
 
-func (r *AgentServiceConfigReconciler) newAgentIPXERoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
-
-	// Wait for https route to be created first
+func (r *AgentServiceConfigReconciler) newHTTPRoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig, serviceToExpose string) (client.Object, controllerutil.MutateFn, error) {
+	// In order to create plain http route we need https route to be created first to copy its host
 	httpsRoute := &routev1.Route{}
-	if err := r.Get(ctx, types.NamespacedName{Name: serviceName, Namespace: r.Namespace}, httpsRoute); err != nil {
-		log.WithError(err).Error("Failed to get https route for agent service")
+	if err := r.Get(ctx, types.NamespacedName{Name: serviceToExpose, Namespace: r.Namespace}, httpsRoute); err != nil {
+		log.WithError(err).Errorf("Failed to get https route for %s service", serviceToExpose)
 		return nil, nil, err
 	}
 	if httpsRoute.Spec.Host == "" {
-		log.Info("Agent https route found, but host not yet set")
+		log.Infof("https route for %s service found, but host not yet set", serviceToExpose)
 		return nil, nil, nil
 	}
 
 	weight := int32(100)
 	route := &routev1.Route{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      fmt.Sprintf("%s-ipxe", serviceName),
+			Name:      fmt.Sprintf("%s-ipxe", serviceToExpose),
 			Namespace: r.Namespace,
 		},
 		Spec: routev1.RouteSpec{
@@ -645,11 +644,11 @@ func (r *AgentServiceConfigReconciler) newAgentIPXERoute(ctx context.Context, lo
 	routeSpec := routev1.RouteSpec{
 		To: routev1.RouteTargetReference{
 			Kind:   "Service",
-			Name:   serviceName,
+			Name:   serviceToExpose,
 			Weight: &weight,
 		},
 		Port: &routev1.RoutePort{
-			TargetPort: intstr.FromString(fmt.Sprintf("%s-http", serviceName)),
+			TargetPort: intstr.FromString(fmt.Sprintf("%s-http", serviceToExpose)),
 		},
 		WildcardPolicy: routev1.WildcardPolicyNone,
 	}
@@ -670,6 +669,10 @@ func (r *AgentServiceConfigReconciler) newAgentIPXERoute(ctx context.Context, lo
 	}
 
 	return route, mutateFn, nil
+}
+
+func (r *AgentServiceConfigReconciler) newAgentIPXERoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
+	return r.newHTTPRoute(ctx, log, instance, serviceName)
 }
 
 func (r *AgentServiceConfigReconciler) newImageServiceRoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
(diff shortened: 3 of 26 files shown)


internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -644,6 +646,47 @@ func (r *AgentServiceConfigReconciler) newImageServiceRoute(ctx context.Context,
...
+		To: routev1.RouteTargetReference{
+			Kind:   "Service",
+			Name:   imageServiceName,
+			Weight: &weight,
+		},
+		Port: &routev1.RoutePort{
+			TargetPort: intstr.FromString(imageServiceName),
+		},
+		WildcardPolicy: routev1.WildcardPolicyNone,
+		TLS:            &routev1.TLSConfig{InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyAllow},
```
carbonin, line 668: Does this work?

The pod is serving https, so we've been using reencrypt for the other routes and I couldn't get it to work without changing the service in my tests.

carbonin: Okay, I think this needs to be WIP then, because we'd also need to add the insecure port to the image-service service, right?

carbonin:
>  (assisted-image-service) expose boot-artifacts via insecure port

I think we need all the endpoints because the initrd is served via `/images`

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -605,6 +621,62 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-ipxe", serviceName),
+			Namespace: r.Namespace,
+		},
+		Spec: routev1.RouteSpec{
+			Host: httpsRoute.Spec.Host,
+		},
+	}
+	routeSpec := routev1.RouteSpec{
+		Path: "/api/assisted-install/v2",
```
vrutkovs, line 649: Two issues with that:
* possibly too broad
* hardcoded, not sure if we define this anywhere so that it could be constructed

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -644,6 +716,118 @@ func (r *AgentServiceConfigReconciler) newImageServiceRoute(ctx context.Context,
...
+			Name:      fmt.Sprintf("%s-ipxe-initrd", imageServiceName),
+			Namespace: r.Namespace,
+		},
+		Spec: routev1.RouteSpec{
+			Host: httpsRoute.Spec.Host,
+		},
+	}
+	routeSpec := routev1.RouteSpec{
+		Host: httpsRoute.Spec.Host,
+		Path: "/images",
```
vrutkovs, line 800: Also possibly too broad. Ideally we'd serve initrd via `/boot-artifacts` maybe?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1230,6 +1419,16 @@ func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(ctx context.
 				ContainerPort: int32(servicePort.IntValue()),
 				Protocol:      corev1.ProtocolTCP,
 			},
+			{
+				ContainerPort: int32(serviceHTTPPort.IntValue()),
+				Protocol:      corev1.ProtocolTCP,
+			},
+		},
+		Command: []string{
+			"/assisted-service",
```
vrutkovs, line 1428: Perhaps we should convert Dockerfile to use `ENTRYPOINT` so that we would not have to define it here?

cmd/main.go
```diff
@@ -194,6 +197,7 @@ func main() {
 	}
 
 	port := flag.String("port", "8090", "define port that the service will listen to")
+	httpPort := flag.String("http-port", "8091", "define http port that the service will listen to")
 	flag.Parse()
```
carbonin, line 201: :confused: why are these cli args instead of env vars

internal/servers/servers.go
```diff
@@ -0,0 +1,70 @@
+package servers
+
```
carbonin, line 2: Maybe change the image service to have this in `pkg` and import it?

carbonin (commented):
@mhrivnak what do you think about the route approach here?
Should we strive to keep the same host for http and https or just go for two hosts and drop the extra role and exchange complexity in the operator for a few more env vars in the service?

mhrivnak:
> @mhrivnak what do you think about the route approach here? Should we strive to keep the same host for http and https or just go for two hosts and drop the extra role and exchange complexity in the operator for a few more env vars in the service?

Can you lay out the two options in a bit more detail? I'm not quite following the choices.

carbonin:
> Can you lay out the two options in a bit more detail? I'm not quite following the choices.

1. Watch the https route to check for the hostname then create a second route with the same hostname, but pointing to the http port of the service (what the PR currently does).
  - This makes the operator a bit more complex and requires an additional permission in our role
2. Create an entirely separate route for the http service port without setting a hostname and have the service decide which host should be used for what endpoints.
  - This adds two extra env vars to the service for the http hostnames

carbonin:
Ideally we could create a single route that allowed for both http and https traffic to a single service, but that doesn't look possible without removing the service side https

mhrivnak:
Thanks for that. Using a single hostname is more user-friendly. It's what people generally expect when both http and https are options, and it's one less DNS record to make for people who can't use wildcard records. If the extra permission isn't something concerning or significant, I'd go with this option.

cmd/main.go
vrutkovs: `port` is being set via cli-args, so following along was my first choice. 

However it appears that we'd have to set "command" and pass "http-port" in the operator, so maybe we should move this to env vars?

cmd/main.go
```diff
@@ -555,12 +559,14 @@ func main() {
 		}
 	}()
 
-	address := fmt.Sprintf(":%s", swag.StringValue(port))
-	if Options.ServeHTTPS {
-		log.Fatal(http.ListenAndServeTLS(address, Options.HTTPSCertFile, Options.HTTPSKeyFile, h))
-	} else {
-		log.Fatal(http.ListenAndServe(address, h))
```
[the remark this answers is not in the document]
vrutkovs: Updated this. Now the code is:
* Run HTTPS handler if httpsPort and certs are set
* If not and httpPort unset - run http handler using httpsPort
* If httpPort is set, run http handler there

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -474,13 +480,18 @@ func (r *AgentServiceConfigReconciler) newAgentService(ctx context.Context, log
 			svc.ObjectMeta.Annotations = make(map[string]string)
 		}
 		svc.ObjectMeta.Annotations[servingCertAnnotation] = serviceName
-		if len(svc.Spec.Ports) == 0 {
-			svc.Spec.Ports = append(svc.Spec.Ports, corev1.ServicePort{})
+		if len(svc.Spec.Ports) != 2 {
```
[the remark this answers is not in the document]
vrutkovs: We're dropping all known ports if the number doesn't match
```
svc.Spec.Ports = []corev1.ServicePort{}
```

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -605,6 +621,62 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
...
+	}
+
+	mutateFn := func() error {
+		if err := controllerutil.SetControllerReference(instance, route, r.Scheme); err != nil {
+			return err
+		}
+		// Only update what is specified above in routeSpec.
+		// If we update the entire route.Spec with
+		// route.Spec = routeSpec
+		// it would overwrite any existing values for route.Spec.Host
```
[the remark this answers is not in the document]
vrutkovs: `spec.Host` cannot be changed once set, so we can't sync those.

`Host` is being set when a route needs to be created:
```
	route := &routev1.Route{
		ObjectMeta: metav1.ObjectMeta{
			Name:      fmt.Sprintf("%s-ipxe", serviceName),
			Namespace: r.Namespace,
		},
		Spec: routev1.RouteSpec{
			Host: httpsRoute.Spec.Host,
		},
	}
```

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -605,6 +621,62 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
 	return route, mutateFn, nil
 }
 
+func (r *AgentServiceConfigReconciler) newAgentIPXERoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
+
+	// Wait for https route to be created first
+	httpsRoute := &routev1.Route{}
+	if err := r.Get(ctx, types.NamespacedName{Name: serviceName, Namespace: r.Namespace}, httpsRoute); err != nil {
```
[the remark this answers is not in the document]
vrutkovs: Ideally we'd set `spec.subdomain` for both routes, but this feature is targeted for 4.12 afaik

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -605,6 +621,62 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
...
+		log.WithError(err).Error("Failed to get https route for agent service")
```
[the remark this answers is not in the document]
vrutkovs: The controller needs to wait for original route to be created, yes. in order to queue the reconcile we have to throw an error here

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -320,6 +308,56 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
 	return ctrl.Result{}, r.Status().Update(ctx, instance)
 }
 
+func (r *AgentServiceConfigReconciler) exposeIPXERoute(instance *aiv1beta1.AgentServiceConfig) bool {
+	var exposeIPXERouteString string
+	exposeIPXERouteString, ok := instance.ObjectMeta.GetAnnotations()[exposeIPXERouteAnnotation]
+	if !ok {
+		exposeIPXERouteString = "false"
+	}
```
carbonin, line 316: I would think we'd want this logic reversed.
iPXE as a feature won't work if we don't have these routes, right?

@filanov @avishayt @mhrivnak do we want http exposed by default so iPXE can work or do we want people to have to opt-in?

internal/controller/controllers/agentserviceconfig_controller.go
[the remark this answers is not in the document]
carbonin: Maybe change this log to an info since we expect to be in this state at some point and it's not necessarily a problem?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -644,6 +716,118 @@ func (r *AgentServiceConfigReconciler) newImageServiceRoute(ctx context.Context,
 	return route, mutateFn, nil
 }
 
+func (r *AgentServiceConfigReconciler) newImageServiceIPXEBootArtifactsRoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
+	// Wait for https route to be created first
+	httpsRoute := &routev1.Route{}
+	if err := r.Get(ctx, types.NamespacedName{Name: imageServiceName, Namespace: r.Namespace}, httpsRoute); err != nil {
+		log.WithError(err).Error("Failed to get https route for image service")
```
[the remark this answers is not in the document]
carbonin: Same comment as before, maybe change this to info if we're expecting to get into this state.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -605,6 +653,59 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
...
+			return err
+		}
+		// Only update what is specified above in routeSpec.
+		// If we update the entire route.Spec with
+		// route.Spec = routeSpec
+		// it would overwrite any existing values for route.Spec.Host
+		route.Spec.To = routeSpec.To
+		route.Spec.Port = routeSpec.Port
+		route.Spec.WildcardPolicy = routeSpec.WildcardPolicy
+		route.Spec.TLS = routeSpec.TLS
```
carbonin, line 702: `route` doesn't have TLS in this case, right? Maybe just nil this out or set it to whatever the zero value is?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1189,6 +1350,10 @@ func (r *AgentServiceConfigReconciler) newAssistedServiceDeployment(ctx context.
 		newSecretEnvVar("EC_PRIVATE_KEY_PEM", "ec-private-key.pem", agentLocalAuthSecretName),
 	}
 
+	if r.exposeIPXERoute(instance) {
+		envSecrets = append(envSecrets, corev1.EnvVar{Name: "HTTP_LISTEN_PORT", Value: serviceHTTPPort.String()})
```
carbonin, line 1354: This is strange, this isn't a secret so I think I'd expect this to be in the configmap rather than on the deployment.

internal/controller/controllers/agentserviceconfig_controller_test.go
```diff
@@ -1156,6 +1264,21 @@ var _ = Describe("ensureAssistedServiceDeployment", func() {
 			})
 		})
 	})
+
+	Describe("ipxe settings", func() {
```
carbonin, line 1268: I don't think you need an extra `Describe` here.
Since it's one test you can just include the context in the name `should expose two ports for ipxe`, would work.

Also, semantically this would be a `Context` rather than a `Describe` but since you're not actually adding any context here I would also suggest just changing the test name in that case.

internal/controller/controllers/infraenv_controller.go
```diff
@@ -525,6 +525,8 @@ func (r *InfraEnvReconciler) setSignedBootArtifactURLs(infraEnv *aiv1beta1.Infra
 		return err
 	}
 	baseURL, err := url.Parse(r.ServiceBaseURL)
+	// ipxeScript
```
carbonin, line 528: I think this should either be removed or expanded to make it a bit more informative.

pkg/app/middleware.go
```diff
@@ -2,14 +2,23 @@ package app
...
 	"github.com/rs/cors"
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	ipxeScriptQueryKey   = "file_name"
+	ipxeScriptQueryValue = "ipxe-script"
+)
+
+var ipxeScriptPattern = regexp.MustCompile(`^/api/assisted-install/v2/infra-envs/(?P<uuid>.*)/downloads/files`)
```
carbonin, line 20: I wish we could do something better than this but I don't see a good way :worried:

pkg/app/middleware.go
```diff
@@ -64,3 +73,34 @@ func SetupCORSMiddleware(handler http.Handler, domains []string) http.Handler {
...
+		if matches == nil {
+			// Path doesn't match the regexp
+			http.NotFound(w, r)
+			return
+		}
+		if _, err := uuid.Parse(matches[1]); err != nil {
+			// Invalid UUID format
+			http.NotFound(w, r)
+			return
+		}
```
carbonin, line 93: Same comment as the code in the image service, for the purpose of this middleware, we don't care about this so I don't think we need to check it here.

carbonin:
/hold

Putting this on hold because it requires https://github.com/openshift/assisted-image-service/pull/74

internal/controller/controllers/agentserviceconfig_controller.go
vrutkovs: Yes - and we need to ensure the reconciled route doesn't have it either

internal/controller/controllers/agentserviceconfig_controller.go
vrutkovs: `envSecrets` is just an unfortunate name - we inject plain `env` using this object. Usually its references from the secret (like DB), but nothing is preventing us from injecting plain envs there

pkg/app/middleware.go
vrutkovs: Right ,synced with proposed image-service change

internal/controller/controllers/agentserviceconfig_controller.go
vrutkovs: Reworked this to inject the value in the configmap (replacing now defunct `SERVE_HTTPS`)

pkg/app/middleware.go
```diff
@@ -64,3 +72,32 @@ func SetupCORSMiddleware(handler http.Handler, domains []string) http.Handler {
...
+
+func WithIPXEScriptMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Check plain HTTP requests only
+		if r.TLS == nil {
+			if r.Method != http.MethodGet {
+				http.NotFound(w, r)
+				return
+			}
+			matches := ipxeScriptPattern.FindStringSubmatch(r.URL.Path)
```
carbonin, line 84: Do you think we could somehow use https://github.com/openshift/assisted-service/blob/4780c78da601e3cee35a7488ec6a260f9aa16b88/restapi/operations/installer/v2_download_infra_env_files_parameters.go#L51 to save us some of this work?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -77,6 +78,7 @@ const (
 
 	configmapAnnotation                 = "unsupported.agent-install.openshift.io/assisted-service-configmap"
 	imageServiceSkipVerifyTLSAnnotation = "unsupported.agent-install.openshift.io/assisted-image-service-skip-verify-tls"
+	exposeIPXERouteAnnotation           = "unsupported.agent-install.openshift.io/assisted-image-expose-plain-http-route-for-ipxe"
```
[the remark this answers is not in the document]
carbonin: Also (as you mentioned previously in chat) this will be supported as far as I know.

internal/imageservice/urls.go
```diff
@@ -8,38 +8,40 @@ import (
...
+	return buildURL(baseURL, fmt.Sprintf("%s/rootfs", BootArtifactsPath), true, map[string]string{
 		"version": version,
 		"arch":    arch,
 	})
 }
 
 func InitrdURL(baseURL, imageID, version, arch string) (string, error) {
 	path := fmt.Sprintf("/images/%s/pxe-initrd", imageID)
-	return buildURL(baseURL, path, map[string]string{
+	return buildURL(baseURL, path, true, map[string]string{
```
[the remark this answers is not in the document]
vrutkovs: Ah, correct, we'd need to pass the "plain HTTP enabled" flag here

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -320,6 +308,56 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
 	return ctrl.Result{}, r.Status().Update(ctx, instance)
 }
 
+func (r *AgentServiceConfigReconciler) exposeIPXERoute(instance *aiv1beta1.AgentServiceConfig) bool {
+	var exposeIPXERouteString string
+	exposeIPXERouteString, ok := instance.ObjectMeta.GetAnnotations()[exposeIPXERouteAnnotation]
+	if !ok {
+		exposeIPXERouteString = "false"
+	}
+	exposeIPXERoute, err := strconv.ParseBool(exposeIPXERouteString)
```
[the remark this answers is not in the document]
vrutkovs: I don't think we had a similar story anywhere - `unsupported.agent-install.openshift.io/assisted-image-service-skip-verify-tls` is passed verbatim as `INSECURE_SKIP_VERIFY` env var.

I'm also leaning towards not checking the value

internal/imageservice/urls.go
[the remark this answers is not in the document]
vrutkovs: Ah, good catch

cmd/main.go
```diff
@@ -555,12 +559,24 @@ func main() {
...
+	// Run listen on http and https ports if HTTPSCertFile/HTTPSKeyFile set
+	serverInfo := servers.New(Options.HTTPListenPort, swag.StringValue(port), Options.HTTPSKeyFile, Options.HTTPSCertFile)
+	if serverInfo.HasBothHandlers {
+		h = app.WithIPXEScriptMiddleware(h)
+	}
+	if serverInfo.HTTP != nil {
+		serverInfo.HTTP.Handler = h
+	}
+	if serverInfo.HTTPS != nil {
+		serverInfo.HTTPS.Handler = h
```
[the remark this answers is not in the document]
vrutkovs: I think we could do
```
if serverInfo.HTTPS != nil {
	serverInfo.HTTPS.Handler = h
}
if serverInfo.HasBothHandlers {
	h = app.WithIPXEScriptMiddleware(h)
}
if serverInfo.HTTP != nil {
	serverInfo.HTTP.Handler = h
}
```
instead, but that would be a bit more confusing - `h` would be in fact different for HTTP and HTTPS handlers.

internal/controller/controllers/agentserviceconfig_controller.go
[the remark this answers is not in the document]
vrutkovs: Right, we need a better name, maybe `agent-install.openshift.io/assisted-image-expose-http-ipxe-route`? Doesn't really roll off the tongue either

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -238,38 +244,20 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
...
-			conditionsv1.SetStatusConditionNoHeartbeat(&instance.Status.Conditions, conditionsv1.Condition{
-				Type:    aiv1beta1.ConditionReconcileCompleted,
-				Status:  corev1.ConditionFalse,
-				Reason:  component.reason,
-				Message: msg,
-			})
-			if statusErr := r.Status().Update(ctx, instance); statusErr != nil {
-				log.WithError(err).Error("Failed to update status")
-				return ctrl.Result{Requeue: true}, statusErr
+	// If IPXE HTTP Routes are exposed sync routes
```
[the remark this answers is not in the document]
vrutkovs: Rephrased into `// Additional routes need to be synced if HTTP iPXE routes are exposed`

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -605,6 +653,61 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
 	return route, mutateFn, nil
 }
 
+func (r *AgentServiceConfigReconciler) newAgentIPXERoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
+
+	// Wait for https route to be created first
```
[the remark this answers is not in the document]
vrutkovs: `// In order to create plain http route we need https route to be created first to copy its host`?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -644,6 +747,61 @@ func (r *AgentServiceConfigReconciler) newImageServiceRoute(ctx context.Context,
 	return route, mutateFn, nil
 }
 
+func (r *AgentServiceConfigReconciler) newImageServiceIPXERoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig) (client.Object, controllerutil.MutateFn, error) {
```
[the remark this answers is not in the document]
vrutkovs: Right, moved shared part into `newHTTPRoute`

pkg/app/middleware.go
```diff
@@ -10,6 +11,13 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	ipxeScriptQueryKey   = "file_name"
+	ipxeScriptQueryValue = "ipxe-script"
+)
+
+var ipxeScriptPattern = regexp.MustCompile(`^/api/assisted-install/v2/infra-envs/([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/downloads/files`)
```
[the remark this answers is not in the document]
vrutkovs: Still tinkering with this - seem `go-openapi` needs middleware context which I can't seem to generate in unit tests.

Meanwhile patched this to a least use `DefaultBasePath` and ensures this middleware doesn't break readiness/liveness checks

cmd/main.go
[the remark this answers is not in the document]
carbonin: You could define a new variable for the handler with the middleware.
That should clear it up.

internal/controller/controllers/agentserviceconfig_controller.go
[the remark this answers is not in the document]
carbonin: `agent-install.openshift.io/expose-http-ipxe-route` maybe?

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: Right and in this case, since it's a http route I figured we didn't need to take TLS from the `routeSpec` object, we would just set it to `nil` or whatever the zero value is.
Especially because this makes the comment inaccurate (specifically the "Only update what is specified above in routeSpec", bit).

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -605,6 +653,64 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
 	return route, mutateFn, nil
 }
 
+func (r *AgentServiceConfigReconciler) newHTTPRoute(ctx context.Context, log logrus.FieldLogger, instance *aiv1beta1.AgentServiceConfig, serviceToExpose string, description string) (client.Object, controllerutil.MutateFn, error) {
```
carbonin, line 656: Is it really necessary to have a separate `description` param here? Any reason not to use `serviceToExpose` in logs? I'd imagine it would be more meaningful, especially given messages like `Failed to get https route for %s service` where I would expect the `%s` to be a service name.

pkg/app/middleware_test.go
```diff
@@ -73,3 +77,61 @@ var _ = Describe("WithHealthMiddleware", func() {
 		Expect(resp.StatusCode).To(Equal(http.StatusOK))
 	})
 })
+
+var _ = Describe("WithInitrdViaHTTPMiddleware", func() {
```
carbonin, line 81: Should we add a test here for the TLS condition?

vrutkovs: Good idea, added

pkg/app/middleware_test.go
```diff
@@ -73,3 +77,82 @@ var _ = Describe("WithHealthMiddleware", func() {
...
+		httpsClient *http.Client
+	)
+
+	BeforeEach(func() {
+		mux := http.NewServeMux()
+		mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			fmt.Fprintln(w, "Hello!")
+		}))
+		httpServer = httptest.NewServer(WithIPXEScriptMiddleware(mux))
+		httpsServer = httptest.NewServer(WithIPXEScriptMiddleware(mux))
```
carbonin, line 95: Why is this any different from the http server?
Shouldn't this be `NewTLSServer`?

vrutkovs: Ah, right, fixed

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: This is going to stay opt-in.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -88,6 +88,12 @@ type AgentServiceConfigSpec struct {
 	// that are used if one the operators fails to be successfully deployed
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Must-Gather Images"
 	MustGatherImages []MustGatherImage `json:"mustGatherImages,omitempty"`
+
+	// ExposeIPXEHTTPRoute is controlling whether the operator is creating plain HTTP routes
+	// iPXE hosts may not work with router cyphers and may access artifacts via HTTP only
+	// This setting defaults to false
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Expose IPXE HTTP route"
+	ExposeIPXEHTTPRoute bool `json:"exposeIPXEHTTPRoute,omitempty"`
```
carbonin, line 96: Even if we make it very targeted I think we should still try for a string here.

Maybe `HTTPIPXERoute` with `enabled` or `disabled` as possibly values?
Does that really gain us anything over a bool though? I don't see a whole lot of room for expansion.

@mhrivnak @avishayt do you want to put some suggestions here?

vrutkovs: That'd make us parse the string instead of golang's json.Marshal, I don't see any benefit tbh

carbonin: > That'd make us parse the string instead of golang's json.Marshal, I don't see any benefit tbh

Not sure I understand this. The string key and value will get into the structure just fine from json.
The motivation is the same that we referenced in slack which is the [kubernetes API conventions around primitive types](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types)

vrutkovs: Ah, okay, I guess `Enabled` could be a good choice for a start

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -88,6 +88,12 @@ type AgentServiceConfigSpec struct {
 	// that are used if one the operators fails to be successfully deployed
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Must-Gather Images"
 	MustGatherImages []MustGatherImage `json:"mustGatherImages,omitempty"`
+
+	// ExposeIPXEHTTPRoute is controlling whether the operator is creating plain HTTP routes
+	// iPXE hosts may not work with router cyphers and may access artifacts via HTTP only
+	// This setting defaults to false, accepts "true,false"
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Expose IPXE HTTP route"
+	ExposeIPXEHTTPRoute string `json:"exposeIPXEHTTPRoute,omitempty"`
```
mhrivnak, line 96: Moving the "expose" part into the value will give more flexibility and fit the intent of keeping this as a string. For example, call the field `ipxeHTTPRoute` with values `enabled` and `disabled`. That leaves the door open for some other mode later, like perhaps `authenticated` or `ipfiltered` or whatever else we might make up.

vrutkovs: Reworked to use `enabled`/`disabled`

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -88,6 +88,12 @@ type AgentServiceConfigSpec struct {
 	// that are used if one the operators fails to be successfully deployed
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Must-Gather Images"
 	MustGatherImages []MustGatherImage `json:"mustGatherImages,omitempty"`
+
+	// ExposeIPXEHTTPRoute is controlling whether the operator is creating plain HTTP routes
```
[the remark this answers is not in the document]
vrutkovs: We expose all endpoints, but then filter these in the middleware:
* `api/assisted-installer/v2/infra-envs/<id>/downloads/files?file_name=ipxe-script`
* `boot-artifacts/` and `images/<infra-enf id>/pxe-initrd` in -image-service

Do we want to list these in the comments? I'd rather have it listed in the documentation

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -88,6 +88,12 @@ type AgentServiceConfigSpec struct {
 	// that are used if one the operators fails to be successfully deployed
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Must-Gather Images"
 	MustGatherImages []MustGatherImage `json:"mustGatherImages,omitempty"`
+
+	// IPXEHTTPRoute is controlling whether the operator is creating plain HTTP routes
```
carbonin, line 92: I think a comment from @mhrivnak about documenting which endpoints are exposed over http got lost here.

api/v1beta1/agentserviceconfig_types.go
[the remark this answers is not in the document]
vrutkovs: Added this to the comment

api/v1beta1/agentserviceconfig_types.go
vrutkovs: Updated the comment

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -88,6 +88,15 @@ type AgentServiceConfigSpec struct {
...
+	// iPXE hosts may not work with router cyphers and may access artifacts via HTTP only
+	// This setting defaults to false, accepts "enabled,disabled". Empty value defaults to disabled
+	// The following endpoints would be exposed via http:
+	// * api/assisted-installer/v2/infra-envs/<id>/downloads/files?file_name=ipxe-script in assisted-service
+	// * boot-artifacts/ and images/<infra-enf id>/pxe-initrd in -image-service
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Expose IPXE HTTP route"
```
mhrivnak, line 98: I think we also want:
```
// +kubebuilder:validation:Enum=enabled;disabled
```

https://sdk.operatorframework.io/docs/building-operators/golang/references/openapi-validation/

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -238,38 +242,20 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
...
-			conditionsv1.SetStatusConditionNoHeartbeat(&instance.Status.Conditions, conditionsv1.Condition{
-				Type:    aiv1beta1.ConditionReconcileCompleted,
-				Status:  corev1.ConditionFalse,
-				Reason:  component.reason,
-				Message: msg,
-			})
-			if statusErr := r.Status().Update(ctx, instance); statusErr != nil {
-				log.WithError(err).Error("Failed to update status")
-				return ctrl.Result{Requeue: true}, statusErr
+	// Additional routes need to be synced if HTTP iPXE routes are exposed
```
mhrivnak, line 250: Is there reconcile logic to ensure that these do not exist in case it's set to disabled? I may have missed that, but didn't see it on my first read-through.

carbonin:
Looks good to me.
I'll let @mhrivnak have the last review though

mhrivnak (changes requested):

internal/controller/controllers/agentserviceconfig_controller.go
vrutkovs: Yes, we have unit tests:
* "should not create plain http route by default"
* "should not create plain http route if ExposeIPXEHTTPRoute is explicitly disabled"
* "should not create plain http route if ExposeIPXEHTTPRoute is not unknown"

mhrivnak: I am looking for:

- should remove plain http route if ExposeIPXEHTTPRoute changes from "enabled" to "disabled"

Do we have that?

vrutkovs: Ah, good idea, added that - `should remove http route after IPXEHTTPRouteEnabled changed to disabled`.

However that won't update links in InfraEnv - this would need to be fixed in a separate PR I suppose

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -605,6 +646,79 @@ func (r *AgentServiceConfigReconciler) newAgentRoute(ctx context.Context, log lo
...
+
+	return route, mutateFn, nil
+}
+
+func (r *AgentServiceConfigReconciler) removeHTTPIPXERoute(ctx context.Context, instance *aiv1beta1.AgentServiceConfig, serviceToExpose string) error {
+	route := &routev1.Route{}
+	routeName := fmt.Sprintf("%s-ipxe", serviceToExpose)
+	namespacedName := types.NamespacedName{Name: routeName, Namespace: r.Namespace}
+	if err := r.Get(ctx, namespacedName, route); err == nil {
+		return r.Client.Delete(ctx, &routev1.Route{
```
mhrivnak, line 708: It's a good idea to detect a NotFound error here and return `nil` in that case.

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: Why defer fixing the infraenv links?

Speaking of which: Don't we want a separate set of non-TLS links on each infraenv, so we can keep the TLS links as well? If a user wants to do non-TLS ipxe in one infranev, that doesn't necessarily mean they want to turn of TLS for all boot artifacts everywhere, right? Or are we assuming that `InfraEnv.status.bootArtifacts` is only ever used for ipxe, and thus we think it's reasonable to treat TLS as an all-or-nothing setting?

mhrivnak:
It looks like the CRD will need to be re-generated when the API changes are done and settled.

internal/controller/controllers/agentserviceconfig_controller.go
vrutkovs: >Why defer fixing the infraenv links?

There are two scenarios when iPXE is being used:
* http links are used directly - dhcp server is configured to consume ipxeScript straight from ACM
* https links are used indirectly - customer fetches the artifacts, places them on a web server manually etc.
This might be the case for installations where ACM is not reachable by DHCP server directly, ipxe scripts need modifications (i.e. kernel args) etc.

As a result this setting is used to simplify ipxe script access in customer environment, not as a preference or customization. Its not something people want to change throughout the hub's lifecycle - if a more complicated system is required, ipxeScript would need to be adjusted manually anyway, so http/https access is orthogonal

>If a user wants to do non-TLS ipxe in one infranev, that doesn't necessarily mean they want to turn of TLS for all boot artifacts everywhere, right? Or are we assuming that InfraEnv.status.bootArtifacts is only ever used for ipxe, and thus we think it's reasonable to treat TLS as an all-or-nothing setting?

`boot-artifacts` are standard RHCOS kernel/rootfs, so these are

internal/controller/controllers/agentserviceconfig_controller.go
vrutkovs: Previous `r.Get` would ensure that `r.Client.Delete` cannot return NotFound, isn't it? There might a race there however

vrutkovs:
/unhold

Unholding this as image-service PR merged and the PR is ready for review

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: The Get is just from the client's cache, which exacerbates the potential race that would exist anyway.

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: Maybe it would help to clarify what exactly you meant by "that won't update links in InfraEnv - this would need to be fixed in a separate PR". What update won't happen? Is it that the links in `bootArtifacts` would all be http, and remain that way without https?

> As a result this setting is used to simplify ipxe script access in customer environment, not as a preference or customization.

"Setting", "preference", "customization", I'm not sure what the significance is here of how they are different?

> Its not something people want to change throughout the hub's lifecycle

Given hardware refresh cycles, I fully expect management clusters to out-live the hardware they are managing. Or to be managing multiple generations of hardware at the same time. Some that use ipxe, some that use redfish virtualmedia perhaps, etc.

Not to mention, this is a declarative API. Whatever desired state is specified should be achievable regardless of current real-world state.

> boot-artifacts are standard RHCOS kernel/rootfs, so these are

Are there words missing at the end?

carbonin: > What update won't happen? Is it that the links in bootArtifacts would all be http, and remain that way without https?

As far as I understand this will generally work as we expect it to. The only issue being that we don't update any of the boot artifact urls unless the iso url changes. This was done to avoid constant churn on the auth token portion of the URL and I feel like changing this behavior is out of scope for this PR.

If the user changes the `ExposeIPXEHTTPRoute` we will rollout the assisted service deployment and whenever the existing logic says to update an infraEnv it will be updated with the correct URL scheme.

> Don't we want a separate set of non-TLS links on each infraenv, so we can keep the TLS links as well? 

I'm not sure this is important. The individual boot artifact URLs were only exposed to allow ironic to fetch the artifacts individually rather than requiring them to pull the artifacts out of our iPXE boot script (they provide their own iPXE script to boot hosts in the converged ZTP flow). If a user has mixed deployment methods then the ISO link will always be https, and the ipxe script url scheme will depend on the env var and whether it was updated as a described above.

That all said I do think we need to improve the infraEnv boot artifact handling eventually. I could see a bug where someone had http exposed, removed it, then ZTP starts failing because the scheme wasn't updated in the infraEnv, but I think that's an uncommon enough case (not to mention that ZTP isn't even using ipxe yet) that we can handle it in a followup.

internal/controller/controllers/agentserviceconfig_controller.go
vrutkovs: Right, added an additional check

vrutkovs:
Regenerated CRDs

internal/controller/controllers/agentserviceconfig_controller.go
vrutkovs: Addressed

cmd/main.go
```diff
@@ -419,10 +423,17 @@ func main() {
 		}
 	}
 
+	// Determine if IPXE artifact URLs need to be http
+	generateInsecureIPXEURLs := false
+	serverInfo := servers.New(Options.HTTPListenPort, swag.StringValue(port), Options.HTTPSKeyFile, Options.HTTPSCertFile)
+	if serverInfo.HasBothHandlers {
+		generateInsecureIPXEURLs = true
+	}
```
CrystalChun, line 431: Nit: can't we set `generateInsecureIPXEURLs := serverInfo.HasBothHandlers`

internal/controller/controllers/agentserviceconfig_controller.go
vrutkovs: updated the logic (initrd URL is updated and so are other links) and added a unittest

vrutkovs:
Not sure if flake

internal/controller/controllers/infraenv_controller.go
```diff
@@ -573,8 +578,18 @@ func (r *InfraEnvReconciler) updateInfraEnvStatus(
...
 		imageCreatedAt := metav1.NewTime(time.Time(internalInfraEnv.GeneratedAt))
 		infraEnv.Status.CreatedTime = &imageCreatedAt
+	}
 
-		// set initrd and script endpoint here so we're not changing the auth token constantly
+	existingInitrdURL, err := url.Parse(infraEnv.Status.BootArtifacts.InitrdURL)
+	if err != nil {
+		return r.handleEnsureISOErrors(ctx, log, infraEnv, err, internalInfraEnv)
+	}
+	desiredInitrdURL, err := url.Parse(infraEnv.Status.BootArtifacts.KernelURL)
```
mhrivnak, line 587: Why is the kernel URL the reference? Does it get updated to the correct scheme through some other process?

mhrivnak:
@carbonin I'm happy if you're happy.

internal/controller/controllers/infraenv_controller.go
```diff
@@ -573,8 +578,19 @@ func (r *InfraEnvReconciler) updateInfraEnvStatus(
...
+	existingInitrdURL, err := url.Parse(infraEnv.Status.BootArtifacts.InitrdURL)
+	if err != nil {
+		return r.handleEnsureISOErrors(ctx, log, infraEnv, err, internalInfraEnv)
+	}
+	// KernelURL is not signed so it gets updated on every reconciliation, thus it has correct URL schema.
+	desiredKernelURL, err := url.Parse(infraEnv.Status.BootArtifacts.KernelURL)
+	if err != nil {
+		return r.handleEnsureISOErrors(ctx, log, infraEnv, err, internalInfraEnv)
+	}
+	if existingInitrdURL.Scheme != desiredKernelURL.Scheme {
```
carbonin, line 592: Would it be more clear to use `r.InsecureIPXEURLs` rather than parsing out the previously set artifact?
So something like:
```
if r.InsecureIPXEURLs && existingInitrdURL == "https" || !r.InsecureIPXEURLs && existingInitrdURL == "http" {
```

Alternatively you could break this out into a helper like `func isHttpsURL(url string) bool` if the condition is too ugly.

carbonin (commented):
A few minor things. Looking good @vrutkovs

internal/controller/controllers/infraenv_controller.go
vrutkovs: It just feels weird to update just the initrd URL separately. It feels a bit better when we keep kernelURL and InitrdURL schemas in sync

carbonin: > It just feels weird to update just the initrd URL separately.

Isn't this what we're already doing though? The value of `r.InsecureIPXEURLs` is the source of truth for what all the scheme's should be, why introduce an additional dependency between urls?

carbonin (approved):

carbonin:
Looks like there's some lint and unit-test failures that need to be addressed.
Content looks good though

internal/controller/controllers/infraenv_controller.go
vrutkovs: Reverted back to d61b506fec1751e2debdc4bbd70ff6500bc6ef74 - breaks edge-subsystem-kubeapi-aws:
```
* Subsystem Suite: [kube-api]cluster installation verify InfraEnv ISODownloadURL and image CreatedTime are not changing - update Annotations
* Subsystem Suite: [kube-api]cluster installation verify InfraEnv ISODownloadURL and image CreatedTime are changing - update IgnitionConfigOverride
```

carbonin (changes requested):

cmd/main.go
```diff
@@ -419,10 +423,14 @@ func main() {
 		}
 	}
 
+	// Determine if IPXE artifact URLs need to be http
+	serverInfo := servers.New(Options.HTTPListenPort, swag.StringValue(port), Options.HTTPSKeyFile, Options.HTTPSCertFile)
+	generateInsecureIPXEURLs := serverInfo.HasBothHandlers
```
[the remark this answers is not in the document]
carbonin: I think the correct logic here would be:

`generateSecureIPXEURLs := serverInfo.HTTPS != nil && serverInfo.HTTP == nil`

Here's a table of the relevant combinations of servers and how they affect how we serve the artifacts.
Probably overkill, but it helped me realize that we actually serve the artifacts over http in two cases and we're only accounting for one here.

|           | HTTPS | NO HTTPS|
|-------|----|----|
| HTTP | http ipxe | http ipxe |
| NO HTTP | https ipxe | X |

carbonin (approved):

vrutkovs:
/hold

subsystem still fails

vrutkovs:
/retest

Hopefully these are flakes

carbonin:
Hmmm looks like some strange behavior in the capi and e2e tests.

capi is complaining about 404s when trying to download the iso and the e2e test is getting failures trying to verify that the API is running correctly.

Maybe something is still strange with the middleware? I'll take a closer look.

cmd/main.go
```diff
@@ -555,12 +564,23 @@ func main() {
...
-	} else {
-		log.Fatal(http.ListenAndServe(address, h))
+	// Interrupt servers on SIGINT/SIGTERM
+	stop := make(chan os.Signal, 1)
+	signal.Notify(stop, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
+
+	// Run listen on http and https ports if iPXE artifacts need to be exposed via HTTP
+	if generateInsecureIPXEURLs {
+		h = app.WithIPXEScriptMiddleware(h)
+	}
```
carbonin, line 574: Yeah, this is now being used incorrectly.

In the http only case we don't want this middleware.
This condition should be the old one where we start the middleware only if we're running both servers.

carbonin: Looks like we don't have the same problem in the image service because it's using  the `HasBothHandlers` attribute directly. Not sure why the CAPI tests were complaining then ... will look closer at those.

carbonin:
In the CAPI case it looks like the image service is propegating the 404 from the assisted service to the client:

`{"file":"/go/src/github.com/openshift/origin/internal/handlers/iso.go:49","func":"github.com/openshift/assisted-image-service/internal/handlers.(*isoHandler).ServeHTTP","level":"error","msg":"Error retrieving ignition content: ignition request to http://assisted-service:8090/api/assisted-install/v2/infra-envs/d3cc2d56-aed9-463c-9b8e-afae01b628bf/downloads/files?api_key=eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJpbmZyYV9lbnZfaWQiOiJkM2NjMmQ1Ni1hZWQ5LTQ2M2MtOWI4ZS1hZmFlMDFiNjI4YmYifQ.1MTb7Q8EEnqRoOcOaqv1689ucS_g0eW9mNByy1HJIRaTd0aZWSKBdGiKtFqVcsIRK6_GvsvOLWnuv_uQgYyXXQ\u0026file_name=discovery.ign returned status 404\n","time":"2022-05-13T18:55:09Z"}`

vrutkovs:
edge-subsystem-aws now passing \o/

/unhold

PR #3728, opened 2022-04-28
MGMT-9986: Assisted-service should evaluate if it should use the converged flow

eranco74:
# Background
[Converge ZTP and Metal3 flows](https://github.com/openshift/enhancements/pull/1059/files)

# Description
This PR is phase 1 of metal3 and assisted-service convergence.
Converged flow is available when kubeAPI is enabled and the HUB cluster
is running baremetal-operator with version 4.11.0 and above
The converged flow is determined per InfraEnv by an annotation infraenv.agent-install.openshift.io/enable-ironic-agent, default is false, this should allow the boot it yourself flow to keep working without any changes

When the converged flow is enabled and the InfraEnv has the `enable-ironic-agent` annotation set to true the assisted-service will do the following:
1. Add the Ironic Agent to the discoveryISO ignition config  
2. Disable the assisted-agent in the discoveryISO ignition config

# Future work:
1. Add PreprovisioningImage controller to assisted-service
2. Refactor BMAC to allow BMH inspection
Note: the implementation in cbohelper.go will get updated to use cluster-baremetal-operator code once it exposes this functionality - https://github.com/openshift/cluster-baremetal-operator/pull/261

https://issues.redhat.com/browse/MGMT-9986

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

WIP, since this is the first phase of the integration with metal3 there isn't much to test 

- [x] dev-scripts environment
- [x] No tests needed

## Assignees
/cc @flaper87
/cc @avishayt

## Checklist

## Reviewers Checklist


internal/cbohelper/cbohelper.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+//go:generate mockgen -source=cbohelper.go -package=cbohelper -destination=mock_cbouhelper.go
+type CBOHelperApi interface {
+	GenerateIronicConfig() (ignition_config_types_32.Config, error)
+	ConvergedFlowAvailable() bool
+}
+
+type Config struct {
+	BaremetalIronicAgentImage      string `envconfig:"IRONIC_AGENT_IMAGE" default:"registry.ci.openshift.org/openshift:ironic-agent:latest"`
```
tsorya, line 28: lets at least have as json per cpu arch

tsorya: I think default value should be from upstream registry

internal/cbohelper/cbohelper.go
```diff
@@ -0,0 +1,175 @@
...
+type CBOHelper struct {
+	client.Client
+	log                    logrus.FieldLogger
+	kubeAPIEnabled         bool
+	config                 Config
+	convergedFlowAvailable bool
+}
+
+func New(c client.Client, log logrus.FieldLogger, kubeAPIEnabled bool, config Config) CBOHelperApi {
+	ch := CBOHelper{Client: c, log: log, kubeAPIEnabled: kubeAPIEnabled, config: config}
```
tsorya, line 41: ch looks very similar to channel , maybe you can change var name?

internal/cbohelper/cbohelper.go
```diff
@@ -0,0 +1,175 @@
...
+		err = fmt.Errorf("there should be only one pod listed for the given label")
+	}
+	return hostIP, err
+}
+
+func (ch *CBOHelper) ConvergedFlowAvailable() bool {
+	return ch.convergedFlowAvailable
+}
+
+func (ch *CBOHelper) setConvergedFlowAvailability() {
```
tsorya, line 148: if i understand right we can't have hybrid option right? I mean we will run only in converged flow or not. Right? i can't remove label from infraenv to allow using not converged flow for example?

tsorya: I think will be nice to have such an option though

internal/ignition/ignition.go
```diff
@@ -1403,6 +1406,17 @@ func (ib *ignitionBuilder) FormatDiscoveryIgnitionFile(ctx context.Context, infr
 		ib.log.WithError(err).Errorln("Unable to build user SSH public key JSON")
 		return "", err
 	}
+	// handle ZTP converged flow
+	UseConvergedFlow := false
+	if infraEnv.EnableIronicAgent {
```
tsorya, line 1411: there is no validation in infraenv controller for it? not sure it is very obvious place to verify infra env  validity

internal/ignition/ignition.go
```diff
@@ -1481,9 +1495,18 @@ func (ib *ignitionBuilder) FormatDiscoveryIgnitionFile(ctx context.Context, infr
 		return "", err
 	}
 
-	res := buf.String()
+	res := ""
+	if UseConvergedFlow {
+		res, err = ib.getConvergedDiscoveryTemplate(buf.Bytes())
```
tsorya, line 1500: i wonder why it should be different from , for example OKD flow? everything is in the same template and in case you need you provide relevant flags? Here if i understand right you create additional configs and append them. It is ok though creates another flow. Maybe we can stick to existing one?

internal/ignition/ignition.go
avishayt: I'm assuming that if the user requested converged but converged isn't available, this error would be propagated to the InfraEnv Status in some way?

internal/controller/controllers/infraenv_controller.go
```diff
@@ -341,6 +343,19 @@ func (r *InfraEnvReconciler) ensureISO(ctx context.Context, log logrus.FieldLogg
...
 
+func ShouldEnableIronicAgent(log logrus.FieldLogger, infraEnv *aiv1beta1.InfraEnv) bool {
+	value, ok := infraEnv.GetAnnotations()[EnableIronicAgentAnnotation]
+	if !ok {
+		return false
+	}
+	log.Debugf("InfraEnv annotation %s value %s", EnableIronicAgentAnnotation, value)
+	enable, err := strconv.ParseBool(value)
+	if err != nil {
+		log.WithError(err).Errorf("faild to parse %s to bool value", value)
```
tsorya, line 354: typo -> faild -> failed

internal/controller/controllers/infraenv_controller.go
```diff
@@ -341,6 +343,19 @@ func (r *InfraEnvReconciler) ensureISO(ctx context.Context, log logrus.FieldLogg
 	return r.updateInfraEnvStatus(ctx, log, infraEnv, updatedInfraEnv)
 }
 
+func ShouldEnableIronicAgent(log logrus.FieldLogger, infraEnv *aiv1beta1.InfraEnv) bool {
```
tsorya, line 346: if it using ironic agent depends on service configuration and we verify it in ignition process , why it even matter which label we have for infraenv? I assume that we should allow both option and decide by looking on infraenv label and not service configuration

internal/cbohelper/cbohelper.go
eranco74: You are correct, we still need to figure out how we get the ironic agent image.

internal/cbohelper/cbohelper.go
eranco74: We have an hybrid option (per infra env).
This is about the availability of the converged flow (4.11 and above)

internal/ignition/ignition.go
eranco74: It's different because we wrote the OKD services ourselves and we get the Ironic ignition bits from CBO.
in order to generate the Ironic ignition, CBO requires a few parameters that we need to fetch during runtime so it makes more sense to  get the Ironic ignition from CBO and merge it with ours

internal/controller/controllers/infraenv_controller.go
eranco74: As explained above, it's both.
There is a service wide configuration for enabling the converged flow, and there is an infraEnv configuration whether to enable the Ironic Agent (use the converge flow) in that InfraEnv ISO

eranco74:
This seems wrong, trying something else

PR #3744, opened 2022-05-02
MGMT-10196: feature enablement enhancement doc

nmagnezi:

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @eliorerz 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/saas-per-organization-feature-access.md
+++ b/docs/enhancements/saas-per-organization-feature-access.md
@@ -1,0 +1,90 @@
+---
+title: saas-per-organization-feature-access
+authors:
+  - "@nmagnezi"
+creation-date: 2022-05-02
+last-updated: 2022-05-02
+---
+
+# SaaS Per-Organization Feature Access
+
+The current OCM-based authorization scheme used by assisted service as a part of the SaaS operation does not support exposing features to certain users.
+We wish to allow the following:
+* An ability to expose selected group(s) of users to assisted-service functionality (in OCM, the term is `capability`).
+* The user grouping should be based on the organization ID.
+
+
+## Motivation
+
+Some features might be of interest to partners or early adopters.
+Other features might not be ready for public consumption: tech-preview or an internal-only functionality.
+Exposing such features to a pre-determined group of users can help us to gather feedback and improve while keeping the other users with a production-level user experience.
+
+
+### Goals
+* Add a generic method to assisted service authorization middleware to assert against an organization's capability. The method should query OCM based on the payload and asserted capability name.
+* Implement a functionality-specific assertion for `ARM`.
+
+
+## Proposal
+
+
+Initially, the authorization middleware should have a method to check if the user belongs to an organization with a particular capability.
+Adding such a method would be the first step, both to assert organization capabilities, such as ARM support, and any other capability of interest in the future.
+
+The API endpoints that need ARM support feature enablement:
+
+### V2ListSupportedOpenshiftVersions API
+
+* [ReleaseImage](../../models/release_image.go) has a `CPUArchitecture` attribute.
+* Expected to filter out ARM version if that capability is not included for the user organization.
+
+
+### V2RegisterCluster API
+
+* [ClusterCreateParams](../../models/cluster_create_params.go) has a `CPUArchitecture` `attribute.
+* Expected to return `HTTP 400` (BadRequest) if `ARM` got selected and that capability is not included for the user organization.
+
+
+### RegisterInfraEnv API
+
+* [InfraEnvCreateParams](../../models/infra_env_create_params.go) has a `CPUArchitecture` attribute.
+* Expected to return `HTTP 400` (BadRequest) if `ARM` got selected and that capability is not included for the user organization.
+
+
+Each of the APIs mentioned above should make use of the new authorization middleware method to make a decision based on the organization's capability existence/absence.
+
+An AMS API call to check for capability will only happen when relevant. In this case, only if the selected CPU architecture (both for cluster and infraEnv) is ARM.
+
+
+### Risks and Mitigations
+
+* The lack of organization capabilities caching might have a SaaS performance impact. See [Open Questions](#Open_Questions).
+* This enhancement is relevant for assisted SaaS, but should not change anything for on-prem. For that, we will include a feature flag to disable this when needed.
+
+### Open Questions
+
+Based on internal discussions, we understand that organization capabilities are not cached in AMS. we should:
+1. Understand the performance impact, if any.
+2. Find an alternative, if needed. We can, for example, cache it locally as done in `isReadOnlyAdmin()` [here](../../pkg/auth/rhsso_authenticator.go).
+
+**Note:** The API endpoints involved in ARM CPU architecture are not expected to be called for many times by each user, but this might not be the case for other future capabilities.
+
+
+### UI/UX Impact
+
+Unlike the UI, the backend will adapt its response to a change in organization capabilities on each API call.
+For ARM specifically, this is less relevant as the user cannot change the CPU architecture post cluster and infra_env creation, yet it might play a role in other future features.
+The UX for those cases will be discussed when it becomes relevant.
+
+The UI should determine which CPU architectures are supported and accessible to the user by querying The AMS API for the user organization capabilities.
+It sends a request to AMS subscriptions (inherited from the organization) via `api.openshift.com/api/accounts_mgmt/v1/subscriptions/`.
+This request gets sent on each UI refresh.
+
+There's a field named `items`, and in that field, a list of `capabilities` in the response payload.
+this request gets sent on each UI refresh.
+
+
+## Alternatives
+
+1. Use OCM roles instead, which will group users based on their email domains.


docs/enhancements/feature-enablement.md
```diff
@@ -0,0 +1,78 @@
...
+### V2ListSupportedOpenshiftVersions API
+
+* [ReleaseImage](../../models/release_image.go) has a `CPUArchitecture` attribute.
+* Expected to filter out ARM version if that capability is not included for the user organization.
+
+
+### V2RegisterCluster API
+
+* [ClusterCreateParams](../../models/cluster_create_params.go) has a `CPUArchitecture` `attribute.
+* Expected to return `HTTP 404` if `ARM` got selected and that capability is not included for the user organization.
```
nmagnezi, line 45: @avishayt should we return 404 in this case? or something else?

docs/enhancements/feature-enablement.md
```diff
@@ -0,0 +1,78 @@
+---
+title: feature-enablement
+authors:
+  - "@nmagnezi"
+creation-date: 2022-05-02
+last-updated: 2022-05-02
+---
+
+# Feature Enablement
```
avishayt, line 9: Let's make the title more descriptive.  How about "SaaS Per-Organization Feature Access"?

docs/enhancements/feature-enablement.md
```diff
@@ -0,0 +1,78 @@
...
+
+The current OCM-based authorization scheme used by assisted service as a part of the SaaS operation does not support a rollout of certain features to a select number of users.
```
avishayt, line 11:
```suggestion
The current OCM-based authorization scheme used by assisted service as a part of the SaaS operation does not support exposing features to certain users.
```

docs/enhancements/feature-enablement.md
```diff
@@ -0,0 +1,78 @@
...
+We wish to allow the following:
+* An ability to expose selected group(s) of users to assisted-service functionality (in OCM, the term is `capability`).
+* The user grouping should be based on the organization ID.
+
+
+## Motivation
+
+Some features might be of interest to partners or early adopters.
```
avishayt, line 19: Some features also may not be ready for public consumption.

docs/enhancements/feature-enablement.md
avishayt: 400, just like if a user would select any other string that's not in the enum (e.g., "foo")

docs/enhancements/feature-enablement.md
```diff
@@ -0,0 +1,78 @@
...
+### V2RegisterCluster API
+
+* [ClusterCreateParams](../../models/cluster_create_params.go) has a `CPUArchitecture` `attribute.
+* Expected to return `HTTP 404` if `ARM` got selected and that capability is not included for the user organization.
+
+
+### RegisterInfraEnv API
+
+* [InfraEnvCreateParams](../../models/infra_env_create_params.go) has a `CPUArchitecture` attribute.
+* Expected to return `HTTP 404` if `ARM` got selected and that capability is not included for the user organization.
```
avishayt, line 51:
```suggestion
* Expected to return `HTTP 400` if `ARM` got selected and that capability is not included for the user organization.
```

docs/enhancements/feature-enablement.md
```diff
@@ -0,0 +1,78 @@
...
+Based on internal discussions, we understand that organization capabilities are not cached. we should:
+1. Understand the performance impact, if any.
+2. Find an alternative, if needed.
+
+**Note:** The API endpoints involved in ARM CPU architecture are not expected to be called for many times by each user, but this might not be the case for other future capabilities.
+
+
+### UI Impact
+
+The UI should be able to determine which CPU is supported and accessible to the user by querying The V2ListSupportedOpenshiftVersions API endpoint.
```
avishayt, line 73: What about features where it may be less obvious?  Would the UI call OCM, or would the backend need to expose it?

docs/enhancements/feature-enablement.md
```diff
@@ -0,0 +1,78 @@
...
+
+
+## Alternatives
+
+1. Use OCM roles instead, which will group users based on their email domains.
```
avishayt, line 78: Are there any benefits to doing this?  The drawback is that it doesn't exactly give us what we want.

docs/enhancements/feature-enablement.md
oourfali: As I see it, the UI should call AMS to check for that capability.

docs/enhancements/feature-enablement.md
```diff
@@ -0,0 +1,78 @@
...
+
+Initially, the authorization middleware should have a method to check if the user belongs to an organization with a particular capability.
+Adding such a method would be the first step, both to assert organization capabilities, such as ARM support, and any other capability of interest in the future.
+
+The API endpoints that need ARM support feature enablement:
+
+### V2ListSupportedOpenshiftVersions API
+
+* [ReleaseImage](../../models/release_image.go) has a `CPUArchitecture` attribute.
+* Expected to filter out ARM version if that capability is not included for the user organization.
```
oourfali, line 39: iiuc, you're using V2ListSupportedOpenshiftVersions anyhow, and you just want to add the capability check in addition to it. That makes sense.

However, consider that it might make sense for V2ListSupportedOpenshiftVersions to return ARM if the openshift version supports it, and add another check for the capability.
Both would work. It is just a matter of how you want your API to behave like.

nmagnezi: Worth a discussion.
We need to think if it makes sense to have 2 APIs and what would be the use-case of each API in that case.
From what I understand, we aim to "hide" this functionality in case it is not allowed (thus the HTTP 400 response that Avishay mentioned above). Let's follow up on this.

docs/enhancements/feature-enablement.md
nmagnezi: The drawback is the domain-based user grouping, which is less granular.
I brought this up because it came up in discussions.

docs/enhancements/feature-enablement.md
nmagnezi: does it currently do that? for other purposes?

docs/enhancements/feature-enablement.md
nmagnezi: Done

docs/enhancements/feature-enablement.md
nmagnezi: Done

docs/enhancements/feature-enablement.md
nmagnezi: Done

docs/enhancements/feature-enablement.md
nmagnezi: Done

docs/enhancements/feature-enablement.md
avishayt: Yes, the UI already calls OCM/AMS

gamli75: we will have 2 flows for the UI for enable/ disable features: 1. call OCM 2. use /v2/feature-support-levels from the BE, right?

gamli75:
few remarks:
1. What will happen if the user created a cluster with permission to use feature and then we changed it in OCM - what will happen in the UI?
2. How do we configure the feature enablement in OCM? Is it a configuration file?
3. I thought this should be a generic solution - how we will support the following combination: OCP 4.11 without ODF for nokia.com?

oourfali:
1. The UI should show/hide all operations related to the capability based on the settings. So if the user had the capability and now no longer has it, then the operations he was able to do before, should now be hidden. For the ARM examples I'd imagine that cluster creation is the only thing we block. So if the user no longer has this capability then he won't be able to create more ARM clusters
2. Adding a capability is in the AMS repo. Applying it to an org is done via opening an OHSS (SD support) ticket (usually gets quickly resolved). We can discuss with SD maybe the option to allow us to set that as well.

docs/enhancements/saas-per-organization-feature-access.md
```diff
@@ -0,0 +1,79 @@
...
+Initially, the authorization middleware should have a method to check if the user belongs to an organization with a particular capability.
+Adding such a method would be the first step, both to assert organization capabilities, such as ARM support, and any other capability of interest in the future.
+
+The API endpoints that need ARM support feature enablement:
+
+### V2ListSupportedOpenshiftVersions API
+
+* [ReleaseImage](../../models/release_image.go) has a `CPUArchitecture` attribute.
+* Expected to filter out ARM version if that capability is not included for the user organization.
+
```
danielerez, line 41: Maybe we should add an API like /v2/feature-gates. I.e. returns a list of available features for the current user. That way, both API user and UI could invoke a single query to get all capabilities. We could also cache the list in the service if makes sense.

docs/enhancements/feature-enablement.md
nmagnezi: done

docs/enhancements/saas-per-organization-feature-access.md
```diff
@@ -0,0 +1,79 @@
...
+
+### V2RegisterCluster API
+
+* [ClusterCreateParams](../../models/cluster_create_params.go) has a `CPUArchitecture` `attribute.
+* Expected to return `HTTP 400` (BadRequest) if `ARM` got selected and that capability is not included for the user organization.
```
gamli75, line 46: @avishayt why the response is 400 - bad request and not 403 - unauthorized? the user will try to update the request. What will be the reason? Will it include the forbidden capability?

docs/enhancements/saas-per-organization-feature-access.md
nmagnezi: Isn't this duplicating a check of user organization capabilities in AMS/OCM API? both assisted-service and the UI should query that API.
IF we implement `/v2/feature-gate` in assisted-service, won't we just proxy that AMS/OCM API?

docs/enhancements/feature-enablement.md
nmagnezi: @oourfali 
Based on the other comments and the discussion we had: I understand that the expected behavior is to "hide" the CPU architectures that the user has no access rights to. Let me know if you think the same.

docs/enhancements/saas-per-organization-feature-access.md
gamli75: @avishayt based on the meeting that we had and decided that there are 2 different APIs: one for the supported features per OCP version and the other for the capabilities, do we need to filter the list as Nir wrote here?

gamli75 (commented):
We need to add a remark that since we have 2 different APIs for features, we should mention that the capabilities have a higher priority

docs/enhancements/saas-per-organization-feature-access.md
avishayt: I believe we said we would have a single API as we have today (feature support levels).  Capabilities should be queried from AMS.
I assumed that users without capabilities should see no trace of a feature that they can't use.  So for example, if a user isn't authorized to use ARM then the behavior should be as if the service doesn't even recognize it as an option (i.e., the same behavior as specifying "cpu_architecture=foobar" - "foobar" doesn't show up in the list of architectures and APIs return 400 if that's specified).  Alternatively, we can decide to expose the existence of the ARM feature to users but not allow setting it, but this seems less user-friendly to me.

docs/enhancements/saas-per-organization-feature-access.md
avishayt: Please see my answer in the comment above.

docs/enhancements/saas-per-organization-feature-access.md
```diff
@@ -0,0 +1,91 @@
...
+### Open Questions
+
+Based on internal discussions, we understand that organization capabilities are not cached. we should:
+1. Understand the performance impact, if any.
+2. Find an alternative, if needed.
+
+**Note:** The API endpoints involved in ARM CPU architecture are not expected to be called for many times by each user, but this might not be the case for other future capabilities.
+
+
+### UI Impact
```
jkilzi, line 72: LGTM
But let's see what will be the outcome after we discuss [the remaining open questions](https://docs.google.com/document/d/1lW3JdDUaTV-RwU2X1sJfVL3wEhMoRw6xZpycUzAvqMw/edit#heading=h.1a6jw0zd1ic7) during our follow up meeting tomorrow

nmagnezi: anything we should add to this doc?

docs/enhancements/saas-per-organization-feature-access.md
```diff
@@ -0,0 +1,91 @@
...
+* Expected to return `HTTP 400` (BadRequest) if `ARM` got selected and that capability is not included for the user organization.
+
+
+### RegisterInfraEnv API
+
+* [InfraEnvCreateParams](../../models/infra_env_create_params.go) has a `CPUArchitecture` attribute.
+* Expected to return `HTTP 400` (BadRequest) if `ARM` got selected and that capability is not included for the user organization.
+
+
+Each of the APIs mentioned above should make use of the new authorization middleware method to make a decision based on the org capability existence/absence.
```
eranco74, line 55: It will use the new authorization middleware if the request requires the feature (ARM) or it will get the org capabilities regardless of the request?

docs/enhancements/saas-per-organization-feature-access.md
jkilzi: Good to go from my POV, I would merge the UI with the UX section.

docs/enhancements/saas-per-organization-feature-access.md
```diff
@@ -0,0 +1,91 @@
...
+Based on internal discussions, we understand that organization capabilities are not cached. we should:
+1. Understand the performance impact, if any.
+2. Find an alternative, if needed.
+
+**Note:** The API endpoints involved in ARM CPU architecture are not expected to be called for many times by each user, but this might not be the case for other future capabilities.
+
+
+### UI Impact
+
+#### UX
```
jkilzi, line 74:
```suggestion
#### UI/UX
```

docs/enhancements/saas-per-organization-feature-access.md
jkilzi, line 72:
```suggestion
```

docs/enhancements/saas-per-organization-feature-access.md
nmagnezi: It will be used only if the selected `CPUArchitecture` (both for cluster and infraEnv) is ARM.

docs/enhancements/saas-per-organization-feature-access.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+### Risks and Mitigations
+
+* The lack of organization capabilities caching might have a SaaS performance impact. See [Open Questions](#Open_Questions).
+* This enhancement is relevant for assisted SaaS, but should not change anything for on-prem. For that, we will include a feature flag to disable this when needed.
+
+### Open Questions
+
+Based on internal discussions, we understand that organization capabilities are not cached. we should:
```
mhrivnak, line 65: not cached where? Can you elaborate?

nmagnezi: Not cached in AMS. Therefore we should probably cache this as we do in:
https://github.com/openshift/assisted-service/blob/fcbe615c34c6397ca11dedc42d521a05f3c936c4/pkg/auth/rhsso_authenticator.go#L246-L261

docs/enhancements/saas-per-organization-feature-access.md
nmagnezi: done

docs/enhancements/saas-per-organization-feature-access.md
nmagnezi: I included this in the doc.

filanov (approved):

PR #3761, opened 2022-05-03
MGMT-8738: Create event with status info even if status doesn't change

michaellevy101:
At present, no media disconnection event is created during cluster installation (cluster status is installing) because the AI get notification from the installer before getting the media disconnection special error code  from the agent.
so the status is already Error and no the status remains the same.
This commit creates the status change event for media disconnection to provide the a friendly message even when the status remains the same.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -2674,24 +2674,42 @@ func (b *bareMetalInventory) handleReplyError(params installer.V2PostStepReplyPa
 }
 
 func (b *bareMetalInventory) handleMediaDisconnection(params installer.V2PostStepReplyParams, ctx context.Context, log logrus.FieldLogger, h *models.Host) error {
-	status := models.HostStatusError
 	statusInfo := fmt.Sprintf("%s - %s", string(models.HostStageFailed), mediaDisconnectionMessage)
 
-	// Install command reports its status with a different API, directly from the assisted-installer.
-	// Just adding our diagnose to the existing error message.
-	if swag.StringValue(h.Status) == status && h.StatusInfo != nil {
-		// Add the message only once
-		if strings.Contains(*h.StatusInfo, statusInfo) {
-			return nil
-		}
-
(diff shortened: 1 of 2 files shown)


internal/bminventory/inventory.go
```diff
@@ -2674,24 +2674,42 @@ func (b *bareMetalInventory) handleReplyError(params installer.V2PostStepReplyPa
...
+		// Update status info
+		err := b.db.Model(h).Updates(map[string]interface{}{
+			"status_info": statusInfo,
+		}).Error
+
+		if err != nil {
+			return err
+		}
+
+		// Create event even for the status info change even when the status remain the same
```
gamli75, line 2704: it means that we will get 2 events:
one event - host status was changed to error without status info and in another one, the host status was changed from error to error with status info, right?
Do we use this event in our data dashboards in elastic? want to be sure that it won't affect our reports
Another question - I hope that this behavior is only for the short term and we won't have a duplicate event in the better solution.

michaellevy101: In the Elastic side we aggregate the events by cluster id.
Un the next version we should update the event.
Maybe even for the current version let me explore it

michaellevy101: The solution should be creating a dedicated media disconnection event. Having fake event isn't good idea (having a status change event without changing the status) 
Updating the existing event isn't good solution too since it will break every consumer. Event should be unmutable

gamli75:
we want to have the full & correct fix for this issue and not work around so I'm closing this issue.

PR #3767, opened 2022-05-04
MGMT-9574: Improve error message when updating static IP

CrystalChun:
[MGMT-9574](https://issues.redhat.com/browse/MGMT-9574)
Previously when a user updates an infra env with an invalid
network yaml, the error message included a traceback given
from the stderr which is confusing and unclear.

This changes the message returned so that it's clear what
went wrong.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

1. Deploy assisted-service with changes
2. Run `make deploy_nodes` to deploy a cluster
3. Curl the infra-envs patch API
    ````bash
    curl -XPATCH  --header "Content-Type: application/json" $SERVICE_URL/api/assisted-install/v2/infra-envs/$INFRA_ENV_ID -d '{"static_network_config":[{"network_yaml":"aaaa","mac_interface_map":[{"mac_address":"52:54:00:ee:42:e1","logical_nic_name":"eth0"}]}]}'
    ````
5. Check the error message returned
## Assignees

/cc @carbonin 
/cc @filanov 

## Checklist

    - I added a subsystem test

## Reviewers Checklist

### DIFF
--- a/pkg/staticnetworkconfig/generator.go
+++ b/pkg/staticnetworkconfig/generator.go
@@ -120,9 +120,8 @@ func (s *StaticNetworkConfigGenerator) executeNMStatectl(ctx context.Context, ho
 	}
 	stdout, stderr, retCode := executer.ExecuteWithContext(ctx, "nmstatectl", "gc", f.Name())
 	if retCode != 0 {
-		msg := fmt.Sprintf("<nmstatectl gc> failed, errorCode %d, stderr %s, input yaml <%s>", retCode, stderr, hostYAML)
-		s.log.Errorf("%s", msg)
-		return "", fmt.Errorf("%s", msg)
+		s.log.Errorf("<nmstatectl gc> failed, errorCode %d, stderr %s, input yaml <%s>", retCode, stderr, hostYAML)
+		return "", fmt.Errorf("failed to execute nmstatectl gc, check the value provided for 'network_yaml': %s", hostYAML)
 	}
 	return stdout, nil
 }
--- a/subsystem/infra_env_test.go
+++ b/subsystem/infra_env_test.go
@@ -204,4 +204,18 @@ var _ = Describe("Infra_Env", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(script).To(HavePrefix("#!ipxe"))
 	})
+	It("fails when given invalid static network config", func() {
+		staticNetworkConfig := models.HostStaticNetworkConfig{
+			NetworkYaml: "aaaaa",
+		}
+		staticNetworkConfigs := []*models.HostStaticNetworkConfig{&staticNetworkConfig}
+		updateParams := &installer.UpdateInfraEnvParams{
+			InfraEnvID: infraEnvID,
+			InfraEnvUpdateParams: &models.InfraEnvUpdateParams{
+				StaticNetworkConfig: staticNetworkConfigs,
+			},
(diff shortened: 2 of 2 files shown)


subsystem/infra_env_test.go
```diff
@@ -204,4 +204,18 @@ var _ = Describe("Infra_Env", func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(script).To(HavePrefix("#!ipxe"))
 	})
+	It("fails to update static network config", func() {
```
carbonin, line 207: Minor, but this should say when it fails. For example:
`fails when given invalid static network config`

pkg/staticnetworkconfig/generator.go
```diff
@@ -122,7 +122,8 @@ func (s *StaticNetworkConfigGenerator) executeNMStatectl(ctx context.Context, ho
 	if retCode != 0 {
 		msg := fmt.Sprintf("<nmstatectl gc> failed, errorCode %d, stderr %s, input yaml <%s>", retCode, stderr, hostYAML)
 		s.log.Errorf("%s", msg)
-		return "", fmt.Errorf("%s", msg)
+		returnMsg := fmt.Sprintf("Failed to execute nmstatectl gc, check the value provided for 'network_yaml': %s", hostYAML)
+		return "", fmt.Errorf("%s", returnMsg)
```
carbonin, line 126: I think this can be simplified now that we're not using the same value for the return and log. For example we could do:

```
s.log.Errorf("<nmstatectl gc> failed, errorCode %d, stderr %s, input yaml <%s>", retCode, stderr, hostYAML)
return "", fmt.Errorf("Failed to execute nmstatectl gc, check the value provided for 'network_yaml': %s", hostYAML)
```

Also, are we pretty sure this will be caused by an invalid input?
If so, we should say that more clearly. Maybe something like "Failed to execute nmstatectl gc, invalid input yaml: %s".

Lastly, the convention for errors is to not capitalize the first word of the message so that it can be wrapped and look reasonable.

So eventually this would be something like `Update infraEnv failed: failed to execute...` rather than `Update infraEnv failed: Failed to exectute...`

CrystalChun: Thank you for pointing out the simplification and the note about the convention! I'll keep that in mind for next time too.

As far as errors, for the most part, I've seen the invalid input error. However, I added a unit test for this and there was an error that says the `nmstatectl` command was missing. I'm not sure if this would be relevant in production, what do you think?

subsystem/infra_env_test.go
CrystalChun: Thank you!

pkg/staticnetworkconfig/generator.go
carbonin: > error that says the nmstatectl command was missing. I'm not sure if this would be relevant in production

We definitely need the command to be present for any of this to work, so I don't think we'd run into that in prod.
That said, I don't have strong opinions on the wording so what you have is probably fine.

CrystalChun: Great that's good to know! Thanks Nick!

PR #3768, opened 2022-05-05
MGMT-9379: Auto approve CSR's for Day2 SNO ZTP "None platform" hosts.

paul-maidment:
Presently, the AgentClusterInstall CR does not have the field
"clusterInstall.Spec.Networking.UserManagedNetworking" set for SNO
clusters. This is however, correctly registered in the database.

This leads to an issue where a rebooting none platform day 2 host will not have it's CSRs approved if the cluster is an SNO cluster and the ZTP install method is being used.

To address this, we now determine whether or not the cluster is a user
managed networking cluster by reading this directly from the database.

In the agent, the variable name "isNoneDay2Rebooting" has been changed
to "shouldAutoApproveCSRs" to make the intent much clearer.

The purpose of this setting is to ensure that the host is placed into a
special "HostStageRebooting" stage when a 2nd day host is rebooting. When the agent CR is set to this
stage, the agent reconciler will check to see if there are any unsigned
CR's and attempt to sign them.

- [ ] New Feature 
- [x ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @ omertuc
/cc @ ori-amizur

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -77,6 +77,37 @@ type AgentReconciler struct {
 	ApproveCsrsRequeueDuration time.Duration
 }
 
+func (r *AgentReconciler) getClusterDeployment(agent *aiv1beta1.Agent, context *context.Context) (*hivev1.ClusterDeployment, error) {
+	if agent.Spec.ClusterDeploymentName != nil {
+		kubeKey := types.NamespacedName{
+			Namespace: agent.Spec.ClusterDeploymentName.Namespace,
+			Name:      agent.Spec.ClusterDeploymentName.Name,
+		}
+		clusterDeployment := &hivev1.ClusterDeployment{}
+		err := r.Get(*context, kubeKey, clusterDeployment)
+		if err == nil {
+			return clusterDeployment, nil
+		}
+		return nil, err
+	}
+	return nil, nil
+}
+
+func (r *AgentReconciler) getCluster(agent *aiv1beta1.Agent) (*common.Cluster, error) {
+	if agent.Spec.ClusterDeploymentName != nil {
+		kubeKey := types.NamespacedName{
+			Namespace: agent.Spec.ClusterDeploymentName.Namespace,
+			Name:      agent.Spec.ClusterDeploymentName.Name,
+		}
+		cluster, err := r.Installer.GetClusterByKubeKey(kubeKey)
+		if err != nil {
+			return nil, err
+		}
+		return cluster, nil
+	}
+	return nil, nil
+}
+
 // +kubebuilder:rbac:groups=agent-install.openshift.io,resources=agents,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=agent-install.openshift.io,resources=agents/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=agent-install.openshift.io,resources=agents/ai-deprovision,verbs=update
@@ -157,29 +188,23 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 	}
 
 	if agent.Spec.ClusterDeploymentName != nil {
-		kubeKey := types.NamespacedName{
-			Namespace: agent.Spec.ClusterDeploymentName.Namespace,
-			Name:      agent.Spec.ClusterDeploymentName.Name,
-		}
-		clusterDeployment := &hivev1.ClusterDeployment{}
-
-		// Retrieve clusterDeployment
-		if err = r.Get(ctx, kubeKey, clusterDeployment); err != nil {
+		_, err = r.getClusterDeployment(agent, &ctx)
+		if err != nil {
 			errMsg := fmt.Sprintf("failed to get clusterDeployment with name %s in namespace %s",
 				agent.Spec.ClusterDeploymentName.Name, agent.Spec.ClusterDeploymentName.Namespace)
 			log.WithError(err).Error(errMsg)
 			// Update that we failed to retrieve the clusterDeployment
 			//TODO MGMT-7844 add mapping CD-ACI to rnot requeue always
-			return r.updateStatus(ctx, log, agent, origAgent, &h.Host, nil, errors.Wrapf(err, errMsg), true)
+			return r.updateStatus(ctx, log, agent, origAgent, &h.Host, nil, nil, errors.Wrapf(err, errMsg), true)
 		}
 
 		// Retrieve cluster by ClusterDeploymentName from the database
-		cluster, err2 := r.Installer.GetClusterByKubeKey(kubeKey)
+		cluster, err2 := r.getCluster(agent)
 		if err2 != nil {
 			log.WithError(err2).Errorf("Fail to get cluster name: %s namespace: %s in backend",
 				agent.Spec.ClusterDeploymentName.Name, agent.Spec.ClusterDeploymentName.Namespace)
 			// Update that we failed to retrieve the cluster from the database
-			return r.updateStatus(ctx, log, agent, origAgent, &h.Host, nil, err2, true)
+			return r.updateStatus(ctx, log, agent, origAgent, &h.Host, nil, nil, err2, true)
 		}
 
 		if h.ClusterID == nil {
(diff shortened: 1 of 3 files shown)


omertuc:
Looking great. Some remarks about the title/description:

- The title makes it sound as if we will auto approve day-2 hosts all the time instead of just on ZTP none-platform

- It makes it sound like "auto approving CSRs for day-2 hosts" is a new feature even though that's already implemented. What's new / getting fixed in this PR is just a bug with SNO and I think the title should communicate that

internal/controller/controllers/agent_controller.go
```diff
@@ -440,13 +467,25 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
...
-			if isNoneDay2Rebooting, err = r.isDay2NonePlatformHostRebooting(ctx, agent, h); err != nil {
-				log.WithError(err).Errorf("Failed to find if agent %s/%s belongs to none platform cluster and is rebooting", agent.Namespace, agent.Name)
-				return ctrl.Result{RequeueAfter: defaultRequeueAfterOnError}, nil
+		// Determine whether or not CSRs should be automatically signed.
+		if cluster != nil {
+			if cluster.UserManagedNetworking == nil {
+				log.Error("UserManagedNetworking was not set, cannot determine whether or not to auto approve CSRs")
+				shouldAutoApproveCSRs = false
+			} else {
+				isNonePlatform := *(cluster.UserManagedNetworking)
```
omertuc, line 476: Can you open a Jira ticket that about the fact that `UserManagedNetworking` should be somehow reflected in the AgentClusterInstall object?

Also add a comment above this line that says "We look at the database cluster object to check for UserManagedNetworking rather than look at the equivalent AgentClusterInstall field because that field the latter does not get properly set in Single Node clusters"

romfreiman: > Can you open a Jira ticket that about the fact that UserManagedNetworking should be somehow correctly reflected in the AgentClusterInstall object


hmm dont we have it? Isnt it something that @ori-amizur worked on? @filanov

omertuc: We have a UserManagedNetworking field on ACI but it's set by users and doesn't sync when the backend implicitly sets it in the database for SNOs

paul-maidment: https://issues.redhat.com/browse/MGMT-10311

filanov: Wrote in the ticket that we can use webhooks to set default values in the spec.

internal/controller/controllers/agent_controller.go
```diff
@@ -193,32 +218,42 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+
+			return r.updateStatus(ctx, log, agent, origAgent, &host.Host, cluster.ID, cluster, nil, true)
 		} else if *h.ClusterID != *cluster.ID {
 			log.Infof("ClusterDeploymentName is changed in Agent %s. unbind first", agent.Name)
 			return r.unbindHost(ctx, log, agent, origAgent, h)
 		}
 	}
 
+	// Retrieve cluster by ClusterDeploymentName from the database
+	cluster, err2 := r.getCluster(agent)
```
ori-amizur, line 232: Why is it needed again ?

internal/controller/controllers/agent_controller.go
```diff
@@ -393,17 +428,10 @@ func (r *AgentReconciler) deregisterHostIfNeeded(ctx context.Context, log logrus
...
-		h.Progress.CurrentStage == models.HostStageDone {
-		if agent.Status.Progress.CurrentStage == models.HostStageDone {
-			return false, nil
-		} else {
-			isNone, err := isAgentInNonePlatformCluster(ctx, r.Client, agent)
-			if err != nil {
-				return false, err
-			}
-			return isNone, nil
+func (r *AgentReconciler) isDay2RebootingHost(ctx context.Context, agent *aiv1beta1.Agent, h *models.Host) (bool, error) {
```
ori-amizur, line 431: No need for error now

internal/controller/controllers/agent_controller.go
```diff
@@ -440,13 +467,28 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
...
-			if isNoneDay2Rebooting, err = r.isDay2NonePlatformHostRebooting(ctx, agent, h); err != nil {
-				log.WithError(err).Errorf("Failed to find if agent %s/%s belongs to none platform cluster and is rebooting", agent.Namespace, agent.Name)
-				return ctrl.Result{RequeueAfter: defaultRequeueAfterOnError}, nil
+		// Determine whether or not CSRs should be automatically signed.
+		if cluster != nil {
+			if cluster.UserManagedNetworking == nil {
+				// We look at the database cluster object to check for UserManagedNetworking rather than look at
+				// the equivalent AgentClusterInstall field because that field the latter does not get
+				// properly set in Single Node clusters
+				log.Error("UserManagedNetworking was not set, cannot determine whether or not to auto approve CSRs")
```
ori-amizur, line 476: nil UserManagedNetworking should be assumed as false.  Anyway, it is not an error

internal/controller/controllers/agent_controller.go
```diff
@@ -440,13 +467,28 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
...
 		}
 
-		if h.Progress != nil && h.Progress.CurrentStage != "" {
-			if isNoneDay2Rebooting, err = r.isDay2NonePlatformHostRebooting(ctx, agent, h); err != nil {
-				log.WithError(err).Errorf("Failed to find if agent %s/%s belongs to none platform cluster and is rebooting", agent.Namespace, agent.Name)
-				return ctrl.Result{RequeueAfter: defaultRequeueAfterOnError}, nil
+		// Determine whether or not CSRs should be automatically signed.
+		if cluster != nil {
+			if cluster.UserManagedNetworking == nil {
+				// We look at the database cluster object to check for UserManagedNetworking rather than look at
```
ori-amizur, line 473: We have swag.BoolValue for that

internal/controller/controllers/agent_controller.go
```diff
@@ -440,13 +467,28 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 			agent.Status.ValidationsInfo = newValidationsInfo
 		}
 
-		if h.Progress != nil && h.Progress.CurrentStage != "" {
-			if isNoneDay2Rebooting, err = r.isDay2NonePlatformHostRebooting(ctx, agent, h); err != nil {
-				log.WithError(err).Errorf("Failed to find if agent %s/%s belongs to none platform cluster and is rebooting", agent.Namespace, agent.Name)
-				return ctrl.Result{RequeueAfter: defaultRequeueAfterOnError}, nil
+		// Determine whether or not CSRs should be automatically signed.
+		if cluster != nil {
```
ori-amizur, line 471: We have already many places checking for SNO:
https://github.com/openshift/assisted-service/blob/6382aa70347ac0e31c1dfbd6ed957be88d357acc/internal/controller/controllers/clusterdeployments_controller.go#L1074

And

https://github.com/openshift/assisted-service/blob/6382aa70347ac0e31c1dfbd6ed957be88d357acc/internal/common/common.go#L113

Why another one ?

internal/controller/controllers/agent_controller.go
```diff
@@ -440,13 +467,28 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 			agent.Status.ValidationsInfo = newValidationsInfo
 		}
 
-		if h.Progress != nil && h.Progress.CurrentStage != "" {
-			if isNoneDay2Rebooting, err = r.isDay2NonePlatformHostRebooting(ctx, agent, h); err != nil {
-				log.WithError(err).Errorf("Failed to find if agent %s/%s belongs to none platform cluster and is rebooting", agent.Namespace, agent.Name)
-				return ctrl.Result{RequeueAfter: defaultRequeueAfterOnError}, nil
+		// Determine whether or not CSRs should be automatically signed.
```
ori-amizur, line 470: Please move the logic that identifies if CSRs should be approved to a function

internal/controller/controllers/agent_controller.go
paul-maidment: The cluster that we retrieve on line 202 is out of scope by this point, or at the very least, you have no guarantee that cluster would be populated, so you have to reload it.

The reason cluster is being retrieved here is so that it can be passed to updateStatus where the cluster database object is interrogated for whether or not UserManagedNetworking is enabled.

filanov:
There is a helper function that can check if we are using user managed network on not https://github.com/openshift/assisted-service/blob/b50070f9c64138afeef7301a996f32854d767cc8/internal/controller/controllers/clusterdeployments_controller.go#L604-L607 
In addition i added a comment to https://issues.redhat.com/browse/MGMT-10311 that we maybe can change the code and then look at the CR. 
Currently this change looks a bit complicated and we can probably make it simple
@omertuc @paul-maidment what do you think?

paul-maidment:
Closing this PR as there are simpler ways this can be achieved.

PR #3776, opened 2022-05-08
removing OLM operator must-gather from docs

oourfali:
As the OLM operators must gather isn't something
we'd like to be widely used, yet, we are removing
it from the operator docs.

### DIFF
--- a/docs/operator.md
+++ b/docs/operator.md
@@ -196,31 +196,6 @@ spec:
       cpuArchitecture: "x86_64"
 ```
 
-### Available Olm Operators Must Gather Images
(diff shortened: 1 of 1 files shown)


ybettan:
/approve
@oourfali You should add yourself as an `approver` and probably also as a `reviewer` under https://github.com/openshift/assisted-service/blob/master/OWNERS_ALIASES

gamli75 (approved):

oourfali:
/retitle Bug 2082908: removing OLM operator must-gather from docs

PR #3791, opened 2022-05-09
MGMT-9021: Set feature usage for ignition config overrides

CrystalChun:
[MGMT-9021](https://issues.redhat.com/browse/MGMT-9021)
Sets the feature usage for ignition config overrides when
it's specified by the user.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Creating an infra-env
1. Deploy service with changes
2. Curl the api to create a cluster
3. Curl the api to create an infra-env with an ignition config override 
4. Get the cluster to see that the feature usage flag was added

Updating an infra-env
1. Deploy service with changes
2. Curl the api to create a cluster
3. Curl the api to create an infra-env without an ignition config override 
4. Get the cluster to see that the feature usage flag was not added
5. Patch the infra-env with an ignition config override
4. Get the cluster to see that the feature usage flag was added

## Assignees

/cc @carbonin 
/cc @omertuc 

## Checklist

## Reviewers Checklist


omertuc:
IIUC ignition_config_overrides can be set in 3 levels:
- Host level
- Infra-env level
- Cluster level

I feel we might possibly need 3 different feature usage flags here, to indicate where the ignition config overrides are coming from. It could also be just a single flag, that gets set from the OR of those 3 levels.

internal/bminventory/inventory.go
```diff
@@ -4048,6 +4054,19 @@ func (b *bareMetalInventory) validateClusterInfraEnvRegister(ctx context.Context
 	return nil
 }
 
+func (b *bareMetalInventory) setIgnitionConfigOverrideUsage(clusterId strfmt.UUID, ignitionConfigOverride string, db *gorm.DB) error {
+	var cluster *common.Cluster
+	cluster, err := common.GetClusterFromDB(db, clusterId, common.SkipEagerLoading)
```
omertuc, line 4059: We'll need FOR UPDATE here as well

CrystalChun:
> IIUC ignition_config_overrides can be set in 3 levels:
> ...

Thanks for explaining the different levels the ignition config override is set!

For the `cluster` level, I noticed the [cluster model](https://github.com/openshift/assisted-service/blob/master/models/cluster.go#L125) has the property `IgnitionConfigOverrides` but I don't see where it is set in the code and I don't see it specified as a [create](https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L4162) or [update](https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L4337) parameter in the swagger. 

Do you happen to know if it's set anywhere or do you know someone who might know?

Would the ignition endpoint count as part of this? I wouldn't think so, but what do you think?

internal/bminventory/inventory.go
CrystalChun: Thanks for catching this! I switched to the correct function.

carbonin:
> For the cluster level, I noticed the [cluster model](https://github.com/openshift/assisted-service/blob/master/models/cluster.go#L125) has the property IgnitionConfigOverrides but I don't see where it is set in the code and I don't see it specified as a [create](https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L4162) or [update](https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L4337) parameter in the swagger.

This was from before infraEnv existed. This should only be relevant on infraEnv and host now.

> Would the ignition endpoint count as part of this? I wouldn't think so, but what do you think?

Not sure what this means

CrystalChun:
> This was from before infraEnv existed. This should only be relevant on infraEnv and host now.



Thanks for letting me know the history! Should that parameter be removed from the cluster object?

> > Would the ignition endpoint count as part of this? I wouldn't think so, but what do you think?
> 
> Not sure what this means

Sorry I meant would setting the [`ignition_endpoint`](https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L4303) parameter count for setting the ignition config override feature usage? I don't think it would

carbonin:
> Sorry I meant would setting the [ignition_endpoint](https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L4303) parameter count for setting the ignition config override feature usage? I don't think it would

No that's something else I think.
Unfortunately the [original PR](https://github.com/openshift/assisted-service/pull/2849) doesn't explain why it was added. @eranco74 @avishayt ?

carbonin:
> Should that parameter be removed from the cluster object?

Removing fields needs to be done carefully as it would be a breaking API change.
We probably should have caught it when moving to v2, but now that it's still there we probably will have to just leave it for now.

internal/bminventory/inventory.go
```diff
@@ -4048,6 +4054,35 @@ func (b *bareMetalInventory) validateClusterInfraEnvRegister(ctx context.Context
 	return nil
 }
 
+func (b *bareMetalInventory) setIgnitionConfigOverrideUsage(clusterId strfmt.UUID, ignitionConfigOverride, level string, db *gorm.DB) error {
+	var cluster *common.Cluster
+	cluster, err := common.GetClusterFromDBForUpdate(db, clusterId, common.SkipEagerLoading)
```
omertuc, line 4059: FOR UPDATE has to happen as part of a transaction, but from a quick scan of the code I don't think this instance of `db` is a transaction.

CrystalChun: Oh I see, right it isn't!

internal/bminventory/inventory.go
```diff
@@ -3853,6 +3855,10 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
 				errMsg = fmt.Sprintf("%s. Error: %s", errMsg, errString)
 			}
 			log.Errorf(errMsg)
+			tx.Rollback()
+			if r := recover(); r != nil {
+				tx.Rollback()
+			}
```
CrystalChun, line 3861: I'm not sure if this is needed

internal/bminventory/inventory.go
CrystalChun: It is now! All calls to this function are now transactions

internal/bminventory/inventory.go
```diff
@@ -4089,10 +4136,14 @@ func (b *bareMetalInventory) UpdateInfraEnvInternal(ctx context.Context, params
 				errWrapperLog = log.WithError(err)
 			}
 			errWrapperLog.Errorf("Failed to update InfraEnv with id %s", params.InfraEnvID)
+			tx.Rollback()
+			if r := recover(); r != nil {
+				tx.Rollback()
+			}
```
CrystalChun, line 4142: Not sure if this is needed

carbonin (commented):
Do we care about this being correct for non-SaaS use cases?

Specifically the discovery ignition overrides are a property of an infraEnv, but we're only setting the usage on the cluster if the infraEnv cluster ID is set. This doesn't really work in a late binding scenario.

@avishayt ?

internal/bminventory/inventory.go
```diff
@@ -4196,6 +4245,11 @@ func (b *bareMetalInventory) updateInfraEnvData(ctx context.Context, infraEnv *c
 
 	if params.InfraEnvUpdateParams.IgnitionConfigOverride != "" && params.InfraEnvUpdateParams.IgnitionConfigOverride != infraEnv.IgnitionConfigOverride {
 		updates["ignition_config_override"] = params.InfraEnvUpdateParams.IgnitionConfigOverride
+		// Set the feature usage for ignition config overrides since it changed
+		if err := b.setIgnitionConfigOverrideUsage(infraEnv.ClusterID, params.InfraEnvUpdateParams.IgnitionConfigOverride, "infra-env", db); err != nil {
```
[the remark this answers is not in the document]
avishayt: Correct.  That's important since ignition config override is available only via REST API and not via UI.  So the most foolproof way to do it would be to iterate over the cluster's hosts to find the set of InfraEnvs that they booted from (host.infra_env_id), then check those InfraEnvs to see if this was set.

internal/bminventory/inventory.go
```diff
@@ -4123,10 +4167,14 @@ func (b *bareMetalInventory) UpdateInfraEnvInternal(ctx context.Context, params
...
 			errWrapperLog.Errorf("Failed to update InfraEnv with id %s", params.InfraEnvID)
+			tx.Rollback()
+			if r := recover(); r != nil {
+				tx.Rollback()
+			}
 		}
 	}()
 
-	if infraEnv, err = common.GetInfraEnvFromDB(b.db, params.InfraEnvID); err != nil {
+	if infraEnv, err = common.GetInfraEnvFromDB(tx, params.InfraEnvID); err != nil {
```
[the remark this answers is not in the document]
CrystalChun: I don't think so since this object is [passed](https://github.com/openshift/assisted-service/pull/3791/files#diff-747a429a0e12307ef65d754c83b6898ce3858b621005e1e7af6e420da1f1657cR4201) to the [`updateInfraEnvData`](https://github.com/openshift/assisted-service/blob/7b0200234511254c2315eed7490a0c39fbbd676b/internal/bminventory/inventory.go#L4222) function which only updates certain columns instead of the full object

internal/bminventory/inventory.go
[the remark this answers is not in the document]
CrystalChun: The function [`setIgnitionConfigOverrideUsage`](https://github.com/CrystalChun/assisted-service/blob/2e41ac4d0bd0dad16c8783aa72468e633a0f5d01/internal/bminventory/inventory.go#L4096) checks for the existence of the cluster and if the `ClusterID` isn't set (from my understanding) then it'll just return.

internal/bminventory/inventory.go
```diff
@@ -4079,6 +4093,35 @@ func (b *bareMetalInventory) validateClusterInfraEnvRegister(ctx context.Context
...
+		props[level] = ignitionOverriden
+
+		if !ignitionOverriden {
+			// Check to see if at least one level of ignition config override is set
+			for _, val := range props {
+				if val.(bool) {
+					ignitionOverriden = true
+				}
+			}
+		}
```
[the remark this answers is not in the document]
CrystalChun: I modified it! It was supposed to determine whether the feature usage was actually used by iterating through all the levels since this function could be called for a level where the usage isn't set.

Example: if at the host level, the ignition override was set but this function was called from the infra-env level where the override wasn't set. We don't want to set the overall feature usage to false.

Does this seem like a reasonable check?

internal/bminventory/inventory.go
```diff
@@ -4079,6 +4093,35 @@ func (b *bareMetalInventory) validateClusterInfraEnvRegister(ctx context.Context
...
+	cluster, err := common.GetClusterFromDBForUpdate(db, clusterId, common.SkipEagerLoading)
+	if err != nil {
+		return err
+	}
+	if usages, uerr := usage.Unmarshal(cluster.Cluster.FeatureUsage); uerr == nil {
+		props := usages[usage.IgnitionConfigOverrideUsage].Data
+		if props == nil {
+			props = make(map[string]interface{})
+		}
+		ignitionOverriden := ignitionConfigOverride != ""
```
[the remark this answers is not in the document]
CrystalChun: Fixed, thank you!

internal/bminventory/inventory.go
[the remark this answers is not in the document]
carbonin: I think what we're getting at here is that there are situations where hosts in a cluster could be using a discovery ignition override, but the infraEnv doesn't have cluster id set. The current implementation would miss those cases.

The more I think about this though the more I think it will be very difficult to track this correctly.
The problem is that the override is set independently of an action on a host. So it's impossible for us to know if a host was registered using an altered discovery ignition or not.

Consider this chain of events:
1. InfraEnv is created with no override
2. User downloads discovery iso
3. Ignition override is set
4. Host is booted with previously downloaded iso

In this case the host would be discovered and associated with an infraEnv which has an ignition override but hasn't used that override. I don't know if there's a way to handle this case correctly 100% of the time.

The problem is that with late binding discovery ignition is not a cluster-level property. It sounds obvious because the override is set on an infraEnv, but trying to turn an infraEnv attribute into a cluster usage is problematic. I'm starting to question if it's worth trying to track this in late binding cases.

What do you think @avishayt? Is this something we care about for late binding? What value are we getting out of tracking this even in the SaaS? Would it be easier to do our reporting just based on if there's an override set in infraEnvs or is that too difficult to normalize with the other "usage" reports?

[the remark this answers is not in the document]
avishayt: Excellent point. Currently all usage is on clusters. I think the best we can do is assume that all hosts are booted from the latest image and document this assumption. I am also OK with adding the assumption that clusterID is set if that helps significantly, as this will be the typical case in SaaS.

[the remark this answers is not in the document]
carbonin: If we're really just targeting SaaS for this then the current PR is probably fine.
The worst that happens is we don't get the usage recorded in late-binding cases :man_shrugging:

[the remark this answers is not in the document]
avishayt: Yes, it's fine as long as it doesn't barf when ClusterID isn't set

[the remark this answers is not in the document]
CrystalChun: I tested it by creating an InfraEnv and setting the ignition config override! It prints an error that the cluster doesn't exist and returns

`time="2022-06-01T00:09:11Z" level=error msg="failed to set ignition config override usage for cluster " func="github.com/openshift/assisted-service/internal/bminventory.(*bareMetalInventory).updateInfraEnvData" file="/go/src/github.com/openshift/origin/internal/bminventory/inventory.go:4256" error="Failed to get cluster : record not found" go-id=2551 infra_env_id=12815435-4150-462e-96a6-8a3e8d8a6f55 pkg=Inventory request_id=0bad7499-cf0b-458e-8df7-a61421e69dfe`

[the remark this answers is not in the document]
carbonin: Great. Lets make these warnings though if they're not failing the operation.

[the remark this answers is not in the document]
CrystalChun: Converted them to warnings!

internal/bminventory/inventory.go
```diff
@@ -3882,58 +3884,37 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
...
 
-	if err = auth.ValidateAccessToCPUArchitecture(ctx, b.authzHandler, params.InfraenvCreateParams.CPUArchitecture); err != nil {
-		return nil, err
-	}
+	var cluster *common.Cluster
+	clusterId := params.InfraenvCreateParams.ClusterID
+	if clusterId != nil {
+		cluster, err = common.GetClusterFromDB(b.db, *clusterId, common.SkipEagerLoading)
+		if err != nil {
+			err = errors.Errorf("Cluster ID %s does not exists", clusterId.String())
```
carbonin, line 3899: Should this continue to be a `BadRequest` response as it was in https://github.com/openshift/assisted-service/pull/3791/files#diff-747a429a0e12307ef65d754c83b6898ce3858b621005e1e7af6e420da1f1657cL4068

internal/bminventory/inventory.go
```diff
@@ -4011,18 +3993,76 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
...
 		return nil, err
 	}
 
-	success = true
 	return b.GetInfraEnvInternal(ctx, installer.GetInfraEnvParams{InfraEnvID: *infraEnv.ID})
 }
 
+func (b *bareMetalInventory) validateInfraEnvCreateParams(ctx context.Context, params installer.RegisterInfraEnvParams, cluster *common.Cluster) error {
+	var err error
+	if cluster != nil && cluster.CPUArchitecture != "" && cluster.CPUArchitecture != params.InfraenvCreateParams.CPUArchitecture {
```
carbonin, line 4016: This looks new to me.
Why add it now?

internal/bminventory/inventory.go
```diff
@@ -4011,18 +3993,76 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
...
+	if err = auth.ValidateAccessToCPUArchitecture(ctx, b.authzHandler, params.InfraenvCreateParams.CPUArchitecture); err != nil {
+		b.log.WithError(err).Errorf("error validaitng access to cpu arch")
+		return err
+	}
+
+	if params.InfraenvCreateParams.Proxy != nil {
+		if err = validateProxySettings(params.InfraenvCreateParams.Proxy.HTTPProxy,
+			params.InfraenvCreateParams.Proxy.HTTPSProxy,
+			params.InfraenvCreateParams.Proxy.NoProxy, &params.InfraenvCreateParams.OpenshiftVersion); err != nil {
+			b.log.WithError(err).Errorf("error validaitngproxy ")
```
carbonin, line 4031: Typo, should be `"error validating proxy"`

internal/bminventory/inventory.go
```diff
@@ -4011,18 +3993,76 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
...
+func (b *bareMetalInventory) validateInfraEnvCreateParams(ctx context.Context, params installer.RegisterInfraEnvParams, cluster *common.Cluster) error {
+	var err error
+	if cluster != nil && cluster.CPUArchitecture != "" && cluster.CPUArchitecture != params.InfraenvCreateParams.CPUArchitecture {
+		err = errors.Errorf("Specified CPU architecture doesn't match the cluster (%s)",
+			cluster.CPUArchitecture)
+		return err
+	}
+
+	if err = auth.ValidateAccessToCPUArchitecture(ctx, b.authzHandler, params.InfraenvCreateParams.CPUArchitecture); err != nil {
+		b.log.WithError(err).Errorf("error validaitng access to cpu arch")
```
carbonin, line 4023: Typo `validating`

internal/bminventory/inventory.go
carbonin: Also I don't think we need to add more context than we had previously it should be fine to just copy the error handling as it was.

internal/bminventory/inventory.go
```diff
@@ -4011,18 +3993,76 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
...
+	if ntpSource != b.Config.DefaultNTPSource && !pkgvalidations.ValidateAdditionalNTPSource(ntpSource) {
+		err = errors.Errorf("Invalid NTP source: %s", ntpSource)
+		return err
+	}
+
+	if params.InfraenvCreateParams.SSHAuthorizedKey != nil && *params.InfraenvCreateParams.SSHAuthorizedKey != "" {
+		if err = validations.ValidateSSHPublicKey(*params.InfraenvCreateParams.SSHAuthorizedKey); err != nil {
+
+			err = errors.Errorf("SSH key is not valid")
+			b.log.WithError(err).Errorf("error validating ssh key")
```
carbonin, line 4047: Why add a log here? We didn't log before, right?

internal/bminventory/inventory.go
```diff
@@ -4011,18 +3993,76 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
...
+
+	ntpSource := swag.StringValue(params.InfraenvCreateParams.AdditionalNtpSources)
+	if ntpSource != b.Config.DefaultNTPSource && !pkgvalidations.ValidateAdditionalNTPSource(ntpSource) {
+		err = errors.Errorf("Invalid NTP source: %s", ntpSource)
+		return err
+	}
+
+	if params.InfraenvCreateParams.SSHAuthorizedKey != nil && *params.InfraenvCreateParams.SSHAuthorizedKey != "" {
+		if err = validations.ValidateSSHPublicKey(*params.InfraenvCreateParams.SSHAuthorizedKey); err != nil {
+
```
carbonin, line 4045: Lots of bonus whitespace.

internal/bminventory/inventory.go
CrystalChun: It was originally here: https://github.com/openshift/assisted-service/pull/3791/files/43514d7cfceae56afb0bd6ea465c5ecd7c37af16#diff-747a429a0e12307ef65d754c83b6898ce3858b621005e1e7af6e420da1f1657cL4075-L4077

(Does this link work?)

internal/bminventory/inventory.go
CrystalChun: Oops thank you! This was a debug log I accidentally left in

internal/bminventory/inventory.go
carbonin: No, but I found where it was. Makes sense.

internal/bminventory/inventory_test.go
```diff
@@ -6588,9 +6588,59 @@ var _ = Describe("infraEnvs", func() {
...
+				CPUArchitecture:  "x86_64",
+			}}
+			err := db.Create(&cluster).Error
+			Expect(err).ShouldNot(HaveOccurred())
+			mockInfraEnvRegisterSuccess()
+			mockEvents.EXPECT().SendInfraEnvEvent(ctx, eventstest.NewEventMatcher(
+				eventstest.WithNameMatcher(eventgen.InfraEnvRegisteredEventName))).Times(1)
+			mockUsage.EXPECT().Add(gomock.Any(), usage.IgnitionConfigOverrideUsage, gomock.Any()).Times(1)
+			mockUsage.EXPECT().Save(gomock.Any(), gomock.Any(), gomock.Any()).Times(1)
+			mockUsage.EXPECT().Remove(gomock.Any(), gomock.Any()).Times(0)
```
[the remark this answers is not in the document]
carbonin: This still applies, but I don't expect you to change the usage being mocked in this PR. Resolving.

internal/bminventory/inventory.go
carbonin: I think it is because you want to make sure you get the cluster update to include the usage and the actual infraEnv update or neither, right? (i.e. we don't want the infraEnv update to succeed but then fail and not get the usage update).

internal/bminventory/inventory.go
CrystalChun: Right, good catch! Modified it back

avishayt: Why isn't it NotFound?

avishayt (commented):
One comment , otherwise looks good

internal/bminventory/inventory.go
carbonin: I suppose it's because the cluster ID is provided as input vs being the resource you're operating on.
That said I was just comparing to the previous implementation.

I guess NotFound would work if the cluster actually wasn't found, but to be super pedantic, I'd expect BadRequest if the ID was something other than a uuid and I'd expect InternalServerError if some other database error happened (connection error, perhaps).

internal/bminventory/inventory.go
```diff
@@ -4198,6 +4261,11 @@ func (b *bareMetalInventory) updateInfraEnvData(ctx context.Context, infraEnv *c
 
 	if params.InfraEnvUpdateParams.IgnitionConfigOverride != "" && params.InfraEnvUpdateParams.IgnitionConfigOverride != infraEnv.IgnitionConfigOverride {
```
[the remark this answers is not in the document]
carbonin: EDIT: yup, the next few lines look the same.

You cant unset the image type, static network config, or pull secret either.

internal/bminventory/inventory.go
avishayt: Agree with Nick

CrystalChun: Went ahead and took a closer look

If the ID is not a UUID then the error is caught earlier on, but it doesn't return a `BadRequest` (at least not from curling the API):
```
$ curl ${ASSISTED_SERVICE_URL}/api/assisted-install/v2/infra-envs -X POST -H "Content-Type: application/json" --data '{"name":"test2","cluster_id": "test"}'

{"code":601,"message":"cluster_id in body must be of type uuid: \"test\""}
```
We might be able to rule out the cluster_id not being type UUID as an error.

This leaves:
1. Cluster_id not found --> `NotFound` error
2. DB error --> `InternalServerError`

I've made changes to reflect this. How does it look to you both?

PR #3796, opened 2022-05-10
MGMT-10263: Do not call refresh status from bm-inventory if called from Kube API

ori-amizur:
Calling refresh-status not from monitor produces excessive load.  To
avoid this load, refresh status was removed in bm-inventory if
called from Kube API.
The refresh was kept for REST API since UI expects immediate affect to
update operation.
In addition, the refresh status was removed from install if no role was
updated as part of auto-assign operation.
This change also verifies that refresh will be triggered if any of the
networks were actually changed.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @filanov 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -151,7 +151,7 @@ type InstallerInternals interface {
 	TransformClusterToDay2Internal(ctx context.Context, clusterID strfmt.UUID) (*common.Cluster, error)
 	AddReleaseImage(ctx context.Context, releaseImageUrl, pullSecret, ocpReleaseVersion, cpuArchitecture string) (*models.ReleaseImage, error)
 	GetClusterSupportedPlatformsInternal(ctx context.Context, params installer.GetClusterSupportedPlatformsParams) (*[]models.PlatformType, error)
-	V2UpdateHostInternal(ctx context.Context, params installer.V2UpdateHostParams) (*common.Host, error)
+	V2UpdateHostInternal(ctx context.Context, params installer.V2UpdateHostParams, interactivity Interactivity) (*common.Host, error)
 	GetInfraEnvByKubeKey(key types.NamespacedName) (*common.InfraEnv, error)
 	UpdateInfraEnvInternal(ctx context.Context, params installer.UpdateInfraEnvParams) (*common.InfraEnv, error)
 	RegisterInfraEnvInternal(ctx context.Context, kubeKey *types.NamespacedName, params installer.RegisterInfraEnvParams) (*common.InfraEnv, error)
@@ -1050,9 +1050,11 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params 
 	if cluster, err = common.GetClusterFromDBWithHosts(b.db, params.ClusterID); err != nil {
 		return nil, common.NewApiError(http.StatusNotFound, err)
 	}
+
+	var autoAssigned bool
+
 	// auto select hosts roles if not selected yet.
 	err = b.db.Transaction(func(tx *gorm.DB) error {
-		var autoAssigned bool
 		var selected bool
 		for i := range cluster.Hosts {
 			if selected, err = b.hostApi.AutoAssignRole(ctx, cluster.Hosts[i], tx); err != nil {
@@ -1074,17 +1076,20 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params 
 		return nil, err
 	}
 
-	if err = b.refreshAllHostsOnInstall(ctx, cluster); err != nil {
-		return nil, err
-	}
-	if _, err = b.clusterApi.RefreshStatus(ctx, cluster, b.db); err != nil {
-		return nil, err
+	if autoAssigned {
+		if err = b.refreshAllHostsOnInstall(ctx, cluster); err != nil {
+			return nil, err
+		}
+		if _, err = b.clusterApi.RefreshStatus(ctx, cluster, b.db); err != nil {
+			return nil, err
+		}
+
+		// Reload again after refresh
+		if cluster, err = common.GetClusterFromDBWithHosts(b.db, params.ClusterID); err != nil {
+			return nil, common.NewApiError(http.StatusNotFound, err)
+		}
 	}
 
-	// Reload again after refresh
-	if cluster, err = common.GetClusterFromDBWithHosts(b.db, params.ClusterID); err != nil {
-		return nil, common.NewApiError(http.StatusNotFound, err)
-	}
 	// Verify cluster is ready to install
 	if ok, reason := b.clusterApi.IsReadyForInstallation(cluster); !ok {
 		return nil, common.NewApiError(http.StatusConflict,
@@ -1704,10 +1709,12 @@ func (b *bareMetalInventory) v2UpdateClusterInternal(ctx context.Context, params
 		return nil, err
 	}
 
-	err = b.updateHostsAndClusterStatus(ctx, cluster, tx, log)
-	if err != nil {
-		log.WithError(err).Errorf("failed to validate or update cluster %s state or its hosts", cluster.ID)
-		return nil, common.NewApiError(http.StatusInternalServerError, err)
+	if interactivity == Interactive {
+		err = b.updateHostsAndClusterStatus(ctx, cluster, tx, log)
+		if err != nil {
+			log.WithError(err).Errorf("failed to validate or update cluster %s state or its hosts", cluster.ID)
+			return nil, common.NewApiError(http.StatusInternalServerError, err)
+		}
 	}
 
 	b.updateClusterNetworkVMUsage(cluster, params.ClusterUpdateParams, usages, log)
@@ -1832,16 +1839,12 @@ func (b *bareMetalInventory) updateNonDhcpNetworkParams(updates map[string]inter
 				return common.NewApiError(http.StatusBadRequest, errors.Wrap(err, "Calculate machine network CIDR"))
 			}
 			if primaryMachineNetworkCidr != "" {
-				if network.IsMachineCidrAvailable(cluster) {
-					cluster.MachineNetworks[0].Cidr = models.Subnet(primaryMachineNetworkCidr)
-				} else {
-					cluster.MachineNetworks = []*models.MachineNetwork{{Cidr: models.Subnet(primaryMachineNetworkCidr)}}
-				}
-				// In case we calculated Machine Network automatically (because conditions for
-				// doing so were met), we update the `params` payload so that from the
-				// perspective of the DB write/delete actions it looks like a valid update
-				// operation. Without that we don't have other way to mark this field of the
-				// cluster as dirty.
+				// We set the machine networks in the ClusterUpdateParams, so they will be viewed as part of the request
+				// to update the cluster
+
+				// Earlier in this function, if reqDualStack was false and the MachineNetworks was non-empty, the function
+				// returned with an error.  Therefore, params.ClusterUpdateParams.MachineNetworks is empty here before
+				// the assignment below.
 				params.ClusterUpdateParams.MachineNetworks = []*models.MachineNetwork{{Cidr: models.Subnet(primaryMachineNetworkCidr)}}
 			}
 		}
@@ -2014,8 +2017,9 @@ func (b *bareMetalInventory) updateClusterData(_ context.Context, cluster *commo
 func (b *bareMetalInventory) updateNetworks(db *gorm.DB, params installer.V2UpdateClusterParams, updates map[string]interface{},
 	cluster *common.Cluster, userManagedNetworking, vipDhcpAllocation bool) error {
 	var err error
+	var updated bool
 
-	if params.ClusterUpdateParams.ClusterNetworks != nil {
+	if params.ClusterUpdateParams.ClusterNetworks != nil && !network.AreClusterNetworksIdentical(params.ClusterUpdateParams.ClusterNetworks, cluster.ClusterNetworks) {
 		for _, clusterNetwork := range params.ClusterUpdateParams.ClusterNetworks {
 			if err = network.VerifyClusterOrServiceCIDR(string(clusterNetwork.Cidr)); err != nil {
 				return common.NewApiError(http.StatusBadRequest, errors.Wrapf(err, "Cluster network CIDR %s", string(clusterNetwork.Cidr)))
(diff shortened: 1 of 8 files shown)


internal/bminventory/inventory.go
```diff
@@ -1031,9 +1031,17 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params
 	if cluster, err = common.GetClusterFromDBWithHosts(b.db, params.ClusterID); err != nil {
 		return nil, common.NewApiError(http.StatusNotFound, err)
 	}
+
+	// Verify cluster is ready to install
+	if ok, reason := b.clusterApi.IsReadyForInstallation(cluster); !ok {
```
carbonin, line 1036: Why was this added? Seems like this would prevent the cluster from installing in a case where it wasn't ready because roles needed to be assigned, but the code added below looks like it would handle exactly that situation.

internal/network/utils.go
```diff
@@ -239,3 +239,51 @@ func GetIPv4Networks(inventory *models.Inventory) (ret []string, err error) {
...
+next:
+	for _, e1 := range n1 {
+		for _, e2 := range n2 {
+			if e1.Cidr == e2.Cidr {
+				continue next
+			}
+		}
+		return false
+	}
+	return true
```
carbonin, line 256: Could this just be:
```
return funk.Every(n1, n2...)
```
ref https://pkg.go.dev/github.com/thoas/go-funk#Every

carbonin: Or are we specifically dealing with a situation where the cluster id field could be different?

carbonin (commented):
> Calling refresh-status not from monitor produces excessive load ...
> refresh status was removed in bm-inventory if not called from REST API (non-interactive)

I think there's an extra negative in here somewhere. Can you rephrase this?
If refresh is causing extra load when called "_not_ from monitor", why was it removed when it _is_ called from the monitor.

The code look like were running the refresh status for the interactive case only, now (so not running it from the controller case).

Does that sound right?

internal/network/utils.go
ori-amizur: Doesn't work if n2 is not []interface{} (Try it).
In addition, we ignore cluster id - it is not part of the payload and added later before save to the database.

internal/bminventory/inventory.go
ori-amizur: This should not be called if cluster is not ready.  When cluster is not ready we don't know if it can be fixed by calling auto-assign.
Anyway, roles should be set by the refresh-role. Here https://github.com/openshift/assisted-service/blob/424cea808c11a43eee184437bac03e851849f894/internal/host/monitor.go#L145 , so the roles in this stage should have already set.
We might be able to remove the code to auto-assign here, but I was a bit careful...

ori-amizur: In addition, auto assign role does not prevent the cluster to become ready. Here https://github.com/openshift/assisted-service/blob/424cea808c11a43eee184437bac03e851849f894/internal/cluster/validator.go#L419-L436

internal/network/utils.go
carbonin: Well that's unfortunate. If anyone's interested I tried it out here https://go.dev/play/p/hsJmfNHRUO4.
I thought this would work with array parameter expansion given that it would work if we pass all the individual values in the array, but no such luck.

internal/bminventory/inventory.go
```diff
@@ -1973,8 +1976,9 @@ func (b *bareMetalInventory) updateClusterData(_ context.Context, cluster *commo
 func (b *bareMetalInventory) updateNetworks(db *gorm.DB, params installer.V2UpdateClusterParams, updates map[string]interface{},
 	cluster *common.Cluster, userManagedNetworking, vipDhcpAllocation bool) error {
 	var err error
+	var updated bool
 
-	if params.ClusterUpdateParams.ClusterNetworks != nil {
+	if params.ClusterUpdateParams.ClusterNetworks != nil && !network.AreClusterNetworksIdentical(params.ClusterUpdateParams.ClusterNetworks, cluster.ClusterNetworks) {
```
carbonin, line 1981: Would it actually be better to do some kind of set difference here so that when the networks are updated we only run the verification on the ones that are new?

internal/bminventory/inventory.go
```diff
@@ -1990,24 +1994,32 @@ func (b *bareMetalInventory) updateNetworks(db *gorm.DB, params installer.V2Upda
...
 			if err = network.VerifyMachineCIDR(string(machineNetwork.Cidr), common.IsSingleNodeCluster(cluster)); err != nil {
 				return common.NewApiError(http.StatusBadRequest, errors.Wrapf(err, "Machine network CIDR '%s'", string(machineNetwork.Cidr)))
 			}
 		}
 		cluster.MachineNetworks = params.ClusterUpdateParams.MachineNetworks
+		updated = true
+	}
+
+	if swag.BoolValue(params.ClusterUpdateParams.VipDhcpAllocation) != swag.BoolValue(cluster.VipDhcpAllocation) ||
+		swag.BoolValue(params.ClusterUpdateParams.UserManagedNetworking) != swag.BoolValue(cluster.UserManagedNetworking) {
```
carbonin, line 2021: Where are these updated?
Did we just need to add this because the `updated` variable was also added?

carbonin (commented):
> Calling refresh-status not from monitor produces excessive load

This is still not accurate, right? (from the PR description)
I'm assuming you mean the controller by "monitor" here, right?
I think you mean "Calling refresh-status from the controller produces excessive load" it would also be helpful to reference which controller is causing this.

---------------------------------------------------------------------------------------------

Also sorry for the extra questions, but this PR feels like it's doing a good bit more than the description states.
In the future it would be helpful to break unrelated changes into separate PRs or at least separate commits so that you have somewhere to fully explain the reasoning behind each of the changes.

For example this PR feels like it's doing all the following changes which don't really relate to each other.

1. Only refresh status on cluster install when roles were auto-assiged
2. Only force a status refresh on cluster/host update when it was triggered by the REST API
3. Don't re-verify cluster networks on update if the networks were not changed
4. Set `trigger_monitor_timestamp` when cluster networks are updated

As someone who isn't _super_ familiar with this code it would help me when I'm reviewing to have each of these changes described fully, including why they're needed and what behavior might change because we're making them.

ori-amizur:
> > Calling refresh-status not from monitor produces excessive load
> 
> This is still not accurate, right? (from the PR description) I'm assuming you mean the controller by "monitor" here, right? I think you mean "Calling refresh-status from the controller produces excessive load" it would also be helpful to reference which controller is causing this.

I meant monitor. https://github.com/openshift/assisted-service/blob/ff812ad654f743c9c9a044c8b1e8cb6aca042d44/internal/host/monitor.go#L109
 And
 https://github.com/openshift/assisted-service/blob/ff812ad654f743c9c9a044c8b1e8cb6aca042d44/internal/cluster/cluster.go#L529
 In general these are the places that refresh-status should be called from.  All other places should be avoided - in our case API calls in bminventory.

> 
> For example this PR feels like it's doing all the following changes which don't really relate to each other.
> 
> 1. Only refresh status on cluster install when roles were auto-assiged

We try to rely on the monitoring.  So if the cluster was already ready, we want to avoid calling refresh cluster and hosts.
If the hosts were auto-assigned, we suspect that it might cause a need to refresh (although from what I saw in the code it doesn't).  So this is a precaution not to make a bigger change.  There is a possibility to avoid the refresh after the auto-assign, only do DB reload.

> 2. Only force a status refresh on cluster/host update when it was triggered by the REST API

I would prefer to avoid calling refresh-status altogether (not only from Kube API), but the UI expects immediate effect to update calls.

Since calling refresh-status is avoided if called from Kube API, I wanted to verify that the refresh-status is triggered when it needs to be.  For this I fixed some logic that currently exists:
> 3. Don't re-verify cluster networks on update if the networks were not changed
If it is already in the DB it is already verified.  If it is changed it has to be verified, and also trigger refresh.

> 4. Set `trigger_monitor_timestamp` when cluster networks are updated
I wanted to trigger refresh when it needs to be.  Since refresh-status was always called in this call, even if the flag was not set the information would have been taken in account in refresh.  When making such change, I verified that refresh is triggered only when there is actual change in the networks data.

> 
> As someone who isn't _super_ familiar with this code it would help me when I'm reviewing to have each of these changes described fully, including why they're needed and what behavior might change because we're making them.

internal/bminventory/inventory.go
```diff
@@ -1809,17 +1816,13 @@ func (b *bareMetalInventory) updateNonDhcpNetworkParams(updates map[string]inter
 				return common.NewApiError(http.StatusBadRequest, errors.Wrap(err, "Calculate machine network CIDR"))
 			}
 			if primaryMachineNetworkCidr != "" {
-				if network.IsMachineCidrAvailable(cluster) {
```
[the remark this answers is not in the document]
ori-amizur: It checks the same thing.
But the change checks it on the ClusterUpdateParams while before we checked it on cluster.
Since later we compare the ClusterUpdateParams and the cluster before updating, and then copy the parameters from ClusterUpdateParams to cluster, then here we just modify the ClusterUpdateParams and not the cluster.
Otherwise, if we modify here the cluster, then later when we check the difference (if to update), it will indicate that there was no change.

internal/bminventory/inventory.go
ori-amizur: Isn't that what we do here?

internal/bminventory/inventory.go
ori-amizur: Because of this https://github.com/openshift/assisted-service/blob/ff812ad654f743c9c9a044c8b1e8cb6aca042d44/internal/bminventory/inventory.go#L2161

We need to verify that the machine networks are updated in the database

internal/bminventory/inventory.go
```diff
@@ -2039,8 +2051,11 @@ func (b *bareMetalInventory) updateNetworks(db *gorm.DB, params installer.V2Upda
 			}
 		}
 	}
-
-	return b.updateNetworkTables(db, cluster, params)
+	if updated {
+		updates["trigger_monitor_timestamp"] = time.Now()
```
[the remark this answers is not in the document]
ori-amizur: Was not, and it was a bug.   Since now we do not always refresh, we need to verify that it is set

internal/bminventory/inventory.go
```diff
@@ -1809,17 +1816,13 @@ func (b *bareMetalInventory) updateNonDhcpNetworkParams(updates map[string]inter
...
+					params.ClusterUpdateParams.MachineNetworks = []*models.MachineNetwork{{Cidr: models.Subnet(primaryMachineNetworkCidr)}}
 				}
-				// In case we calculated Machine Network automatically (because conditions for
-				// doing so were met), we update the `params` payload so that from the
-				// perspective of the DB write/delete actions it looks like a valid update
-				// operation. Without that we don't have other way to mark this field of the
-				// cluster as dirty.
-				params.ClusterUpdateParams.MachineNetworks = []*models.MachineNetwork{{Cidr: models.Subnet(primaryMachineNetworkCidr)}}
 			}
 		}
```
mkowalski, line 1827: I'm just thinking, in this part of code... Previously we were updating both `cluster.MachineNetworks` and `params.ClusterUpdateParams.MachineNetworks`. With this change we will be updating only `params.ClusterUpdateParams.MachineNetworks`. Is it clear who will take care of updating `cluster.MachineNetworks` and when?

I recall from the past that there were some problems with the value being updated, but at a wrong time. So random places in the code were referring to the value post-update and some to the value pre-update. Just trying to understand how is this move going to affect the flow.

I trust that the change here is not breaking stuff, but can't prove it by just looking at this piece

internal/bminventory/inventory_test.go
```diff
@@ -13409,31 +13422,31 @@ func validateNetworkConfiguration(cluster *models.Cluster, clusterNetworks *[]*m
...
+		ExpectWithOffset(1, cluster.MachineNetworks).To(HaveLen(len(*machineNetworks)))
 		for index := range *machineNetworks {
-			Expect(cluster.MachineNetworks[index].ClusterID).To(Equal(*cluster.ID))
-			Expect(cluster.MachineNetworks[index].Cidr).To(Equal((*machineNetworks)[index].Cidr))
+			ExpectWithOffset(1, cluster.MachineNetworks[index].ClusterID).To(Equal(*cluster.ID))
+			ExpectWithOffset(1, cluster.MachineNetworks[index].Cidr).To(Equal((*machineNetworks)[index].Cidr))
 		}
 		// TODO(MGMT-9751-remove-single-network)
-		Expect(cluster.MachineNetworkCidr).To(Equal(""))
+		ExpectWithOffset(1, cluster.MachineNetworkCidr).To(Equal(""))
```
mkowalski, line 13449: I don't understand the core of this offset appearing here. Could you please explain a bit?

internal/network/utils.go
```diff
@@ -239,3 +239,31 @@ func GetIPv4Networks(inventory *models.Inventory) (ret []string, err error) {
 func GetIPv6Networks(inventory *models.Inventory) (ret []string, err error) {
 	return getHostNetworks(inventory, func(i *models.Interface) []string { return i.IPV6Addresses })
 }
+
+func areNetworksIdentical(len1, len2 int, f func(int, int) bool) bool {
+	if len1 != len2 {
+		return false
+	}
+next:
```
mkowalski, line 247: Is this `next:` desired here?

mkowalski: Ok, after googling a bit I'm not sure there is more elegant way to escape from double for-loop (another option being extracting to a yet separate function)

ori-amizur: I will be happy for suggestions.
The only way I am aware of is to put the inner loop in a function

internal/bminventory/inventory.go
ori-amizur: I tried to make it more consistent.  ClusterUpdateParams is what we want to update, and the cluster itself is the current status.
The networks are updated in this function https://github.com/openshift/assisted-service/blob/1df12c082e044f862dcfe847d028918b76a16393/internal/bminventory/inventory.go#L2056

And are verified here

https://github.com/openshift/assisted-service/blob/1df12c082e044f862dcfe847d028918b76a16393/internal/bminventory/inventory.go#L1983

internal/network/utils.go
```diff
@@ -239,3 +239,31 @@ func GetIPv4Networks(inventory *models.Inventory) (ret []string, err error) {
 func GetIPv6Networks(inventory *models.Inventory) (ret []string, err error) {
 	return getHostNetworks(inventory, func(i *models.Interface) []string { return i.IPV6Addresses })
 }
+
+func areNetworksIdentical(len1, len2 int, f func(int, int) bool) bool {
```
carbonin, line 243: Can you give `f` a more meaningful name?
Maybe change this this function to `areNetworkListsIdentical` then `f` could be `areNetworksEqual` or something?

internal/network/utils.go
carbonin: I think putting the inner loop in a named (local-to-this-function) function would be helpful. I don't think the labels are used frequently enough for people to be super familiar with them.

carbonin (commented):
Just a few minor comments. Looking good @ori-amizur

internal/network/utils.go
ori-amizur: Yes

PR #3801, opened 2022-05-12
Bug 2084612: Add a retry in setup lso

adriengentil:
From time to time, the `oc annotate` fails with the following error:
```
Created project openshift-local-storage
Error from server (Conflict): Operation cannot be fulfilled on namespaces "openshift-local-storage": the object has been modified; please apply your changes to
 the latest version and try again
```

This change works around this issue with a retry.

### DIFF
--- a/deploy/operator/setup_lso.sh
+++ b/deploy/operator/setup_lso.sh
@@ -18,7 +18,7 @@ function print_help() {
 function install_lso() {
   oc adm new-project openshift-local-storage || true
 
-  oc annotate project openshift-local-storage openshift.io/node-selector='' --overwrite=true
+  retry -- oc annotate project openshift-local-storage openshift.io/node-selector='' --overwrite=true
 
   catalog_source_name="redhat-operators"
   if [ "${DISCONNECTED}" = true ]; then
--- a/deploy/operator/utils.sh
+++ b/deploy/operator/utils.sh
@@ -155,3 +155,40 @@ function nth_ip() {
 
   python -c "from ansible_collections.ansible.netcommon.plugins.filter import ipaddr; print(ipaddr.nthhost('"$network"', $idx))"
 }
+
+function retry() {
+    attempts=5
+    interval=1
+
+    local OPTIND
+    while getopts "a:i:" opt ; do
+      case "${opt}" in
+          a )
+              attempts="${OPTARG}"
+              ;;
+          i )
+              interval="${OPTARG}"
+              ;;
+          * )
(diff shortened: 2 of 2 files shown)


deploy/operator/utils.sh
```diff
@@ -155,3 +155,26 @@ function nth_ip() {
 
   python -c "from ansible_collections.ansible.netcommon.plugins.filter import ipaddr; print(ipaddr.nthhost('"$network"', $idx))"
 }
+
+function retry() {
+    attempts="${1:-5}"
+    interval="${2:-1}"
```
osherdp, line 161: iiuc, usage of optional positional arguments might be weird seeing, because you cannot easily understand what this means:
```
retry 5 oc annotate project openshift-local-storage openshift.io/node-selector='' --overwrite=true  # will it take "oc" as the interval?
```

osherdp (changes requested):
looks good, just one comment

deploy/operator/utils.sh
adriengentil: yeah, I was wondering too. I'll use getopts instead. Thanks!

adriengentil: Done!

deploy/operator/setup_lso.sh
```diff
@@ -18,7 +18,7 @@ function print_help() {
 function install_lso() {
   oc adm new-project openshift-local-storage || true
 
-  oc annotate project openshift-local-storage openshift.io/node-selector='' --overwrite=true
+  retry -- annotate project openshift-local-storage openshift.io/node-selector='' --overwrite=true
```
osherdp, line 21: shouldn't it be:
```suggestion
  retry -- oc annotate project openshift-local-storage openshift.io/node-selector='' --overwrite=true
```
?

osherdp (changes requested):

deploy/operator/setup_lso.sh
adriengentil: :man_facepalming:

adriengentil: Fixed, I shouldn't touch anything today :sweat_smile:

osherdp (approved):

omertuc:
We'll want to backport this, can you do the bugzilla dance?

osherdp:
/test images
I should it should be fixed now

And for ``e2e-metal-assisted``, maybe we should override as it has no connection to the chages

PR #3804, opened 2022-05-12
MGMT-10275: When getting release image, let a single thread query it, and other threads wait for it

ori-amizur:
When installation starts, many threads attempt to get the image
concurrently.  Each creates a process that attempts to get it, but all
need the same information.
Added cache that verifies that an image is retrieved only once, and all
other threads are waiting for the result.

[MGMT-10271](https://issues.redhat.com//browse/MGMT-10271): Get only id, status, and inventory when querying hosts for connectivity check
Only these fields are needed to construct the connectivity-check step.

[MGMT-10267](https://issues.redhat.com//browse/MGMT-10267): Create cache for host valid interfaces
In order to construct the connectivity check step, each host valid
interfaces are needed.
Since the Unmarshal operation is costly, the valid interfaces of each
host are kept in cache to speed up the valid interfaces retrieval.

[MGMT-10270](https://issues.redhat.com//browse/MGMT-10270): Create queue to throttle connectivity requests
Since the connectivity check takes a lot of bandwidth, (both the request
and the reply size are proportional to the number of hosts), a mechanism
to throttle the number of requests per minute was created.  This
mechanism, limits the number of connectivity requests per minute for
clusters with more than 150 hosts.  The mechanism, allows to generate
up to 150 connectivity checks per minute.  The connectivty checks
requests that were not admitted to run, are kept in queue for next time
such step creation will be attempted.
The queue keeps the order according to "first come first serve" policy.

- [ ] New Feature 
- [X] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/expiring_cache.go
+++ b/internal/common/expiring_cache.go
@@ -1,0 +1,50 @@
+package common
+
+import (
+	"sync"
+	"time"
+
+	"github.com/patrickmn/go-cache"
+)
+
+type ExpiringCache interface {
+	GetOrInsert(key string, value interface{}) (interface{}, bool)
+	Get(key string) (interface{}, bool)
+	Set(key string, value interface{})
+}
+
+type expiringCache struct {
+	cache *cache.Cache
+	mutex sync.Mutex
+}
+
+func NewExpiringCache(defaultExpiration, cleanupInterval time.Duration) ExpiringCache {
+	return &expiringCache{cache: cache.New(defaultExpiration, cleanupInterval)}
+}
+
+func (e *expiringCache) GetOrInsert(key string, value interface{}) (actualIntf interface{}, loaded bool) {
+	actualIntf, loaded = e.Get(key)
+	if loaded {
+		return
+	}
+	e.mutex.Lock()
+	defer e.mutex.Unlock()
+	actualIntf, loaded = e.Get(key)
+	if loaded {
+		return
+	}
+	e.cache.SetDefault(key, value)
+	return value, false
+}
+
+func (e *expiringCache) Get(key string) (interface{}, bool) {
+	actualIntf, loaded := e.cache.Get(key)
+	if loaded {
+		e.cache.SetDefault(key, actualIntf)
+	}
+	return actualIntf, loaded
+}
+
+func (e *expiringCache) Set(key string, value interface{}) {
+	e.cache.SetDefault(key, value)
+}
--- a/internal/host/hostcommands/connectivity_check_cmd.go
+++ b/internal/host/hostcommands/connectivity_check_cmd.go
@@ -2,18 +2,25 @@ package hostcommands
 
 import (
 	"context"
+	"sync"
+	"time"
 
+	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/connectivity"
 	"github.com/openshift/assisted-service/models"
 	"github.com/sirupsen/logrus"
+	"github.com/thoas/go-funk"
 	"gorm.io/gorm"
 )
+
+const maxConnectivityRequestsPerMinute = 150
 
 type connectivityCheckCmd struct {
 	baseCmd
 	db                     *gorm.DB
 	connectivityValidator  connectivity.Validator
 	connectivityCheckImage string
+	queue                  common.ExpiringCache
 }
 
 func NewConnectivityCheckCmd(log logrus.FieldLogger, db *gorm.DB, connectivityValidator connectivity.Validator, connectivityCheckImage string) *connectivityCheckCmd {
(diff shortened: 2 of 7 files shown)


internal/common/expiring_cache.go
```diff
@@ -0,0 +1,46 @@
+package common
+
```
tsorya, line 2: why not to use https://github.com/jellydator/ttlcache , it has everything we need

ori-amizur: We already use this one, https://github.com/patrickmn/go-cache
I wanted to add GetOrInsert

internal/host/hostcommands/connectivity_check_cmd.go
```diff
@@ -22,33 +29,67 @@ func NewConnectivityCheckCmd(log logrus.FieldLogger, db *gorm.DB, connectivityVa
...
 	}
 }
 
+type clusterQueue struct {
+	mutex         sync.Mutex
+	admittedQueue []time.Time
+	waitingQueue  []string
+}
+
+func getHostKey(host *models.Host) string {
```
tsorya, line 42: can you add it to common?

internal/common/expiring_cache.go
```diff
@@ -0,0 +1,46 @@
...
+type expiringCache struct {
+	cache *cache.Cache
+	mutex sync.Mutex
+}
+
+func NewExpiringCache(defaultExpiration, cleanupInterval time.Duration) ExpiringCache {
+	return &expiringCache{cache: cache.New(defaultExpiration, cleanupInterval)}
+}
+
+func (e *expiringCache) GetOrInsert(key string, value interface{}) (interface{}, bool) {
```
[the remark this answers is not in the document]
ori-amizur: The GetOrInsert is done in atomic way.  If the value already exists it returns the existing value.  Otherwise it adds and returns the new one.
We do not want a case that 2 threads at the same time will attempt to insert a value for a specific key and one will overwrite the other.

internal/host/hostcommands/connectivity_check_cmd.go
ori-amizur: Yes

internal/host/hostcommands/connectivity_check_cmd.go
```diff
@@ -22,33 +29,67 @@ func NewConnectivityCheckCmd(log logrus.FieldLogger, db *gorm.DB, connectivityVa
...
+	return host.ID.String() + "@" + host.InfraEnvID.String()
+}
+
+func (c *connectivityCheckCmd) isAdmitted(host *models.Host) bool {
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/oc/release.go
```diff
@@ -154,10 +160,32 @@ func (r *release) GetReleaseArchitecture(log logrus.FieldLogger, releaseImage st
...
+
 func (r *release) getImageFromRelease(log logrus.FieldLogger, imageName, releaseImage, pullSecret string, insecure bool) (string, error) {
 	// Fetch image URL from cache
-	if image, ok := r.imagesMap[imageName][releaseImage]; ok {
-		return image, nil
+	actualImageValue, err := r.getImageValue(imageName, releaseImage)
+	if err != nil {
+		return "", err
+	}
+	if actualImageValue.value != "" {
```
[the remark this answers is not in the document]
ori-amizur: I have no problem using https://github.com/jellydator/ttlcache .  We just already use https://github.com/openshift/assisted-service/blob/dbe941db78105d7cbdf827a6b1400686f4b0fe9a/pkg/auth/local_authenticator.go#L15 .  I don't think we need two of them

internal/common/expiring_cache.go
ori-amizur: BTW, https://github.com/jellydator/ttlcache uses generics (which is nice), but was added in go 1.18.  We are in 1.17.

internal/oc/release.go
```diff
@@ -39,16 +40,21 @@ type Release interface {
...
 	config   Config
 
 	// A map for caching images (image name > release image URL > image)
-	imagesMap map[string]map[string]string
+	imagesMap common.ExpiringCache
 }
 
 func NewRelease(executer executer.Executer, config Config) Release {
-	return &release{executer, config, make(map[string]map[string]string)}
+	return &release{executer: executer, config: config, imagesMap: common.NewExpiringCache(time.Hour*24, time.Hour*24)}
```
tsorya, line 57: in case of release it should not expire

ori-amizur: Yes

internal/common/expiring_cache.go
```diff
@@ -0,0 +1,46 @@
...
+	e.mutex.Lock()
+	defer e.mutex.Unlock()
+	actualIntf, loaded := e.cache.Get(key)
```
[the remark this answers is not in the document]
ori-amizur: Yes

internal/common/expiring_cache.go
```diff
@@ -0,0 +1,50 @@
...
+		return
+	}
+	e.cache.SetDefault(key, value)
+	return value, false
+}
+
+func (e *expiringCache) Get(key string) (interface{}, bool) {
+	actualIntf, loaded := e.cache.Get(key)
+	if loaded {
+		e.cache.SetDefault(key, actualIntf)
```
tsorya, line 43: why we need to set after get?

tsorya: that's why i like ttlcache more :)

PR #3806, opened 2022-05-13
MGMT-9574: Improve error message when updating static network

CrystalChun:
[MGMT9574](https://issues.redhat.com/browse/MGMT-9574
)
Improvement PR for https://github.com/openshift/assisted-service/pull/3767

The error message returned is now more clear and includes
the actual error when nmstatectl is run and which host
the error occurred on.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

1. Deploy assisted-service with changes
2. Run make deploy_nodes to deploy a cluster
3. Curl the infra-envs patch API
    ````
    curl -XPATCH  --header "Content-Type: application/json" $SERVICE_URL/api/assisted-install/v2/infra-envs/$INFRA_ENV_ID -d '{"static_network_config":[{"network_yaml":"aaaa","mac_interface_map":[{"mac_address":"52:54:00:ee:42:e1","logical_nic_name":"eth0"}]}]}'
    ````
5. Check the error message returned
    ````
    {"code":"400","href":"","id":400,"kind":"Error","reason":"1 error occurred:\n\t* failed to validate network yaml for host 0, invalid yaml, error: Route {'table-id': 254, 'destination': '0.0.0.0/0', 'next-hop-address': '192.168.126.1', 'next-hop-interface': 'eth1'} next hop to unknown interface\n\n"}
    ````

## Assignees

/cc @carbonin 
/cc @filanov 

## Checklist

- [x] Does this change include unit-tests (note that code changes require unit-tests) - subsystem test was added in original pr

## Reviewers Checklist

### DIFF
--- a/pkg/staticnetworkconfig/generator.go
+++ b/pkg/staticnetworkconfig/generator.go
@@ -61,7 +61,8 @@ func (s *StaticNetworkConfigGenerator) GenerateStaticNetworkConfigData(ctx conte
 	for i, hostConfig := range staticNetworkConfig {
 		hostFileList, err := s.generateHostStaticNetworkConfigData(ctx, hostConfig, fmt.Sprintf("host%d", i))
 		if err != nil {
-			s.log.WithError(err).Errorf("Failed to create static config for host")
+			err = errors.Wrapf(err, "failed to create static config for host %d", i)
+			s.log.Error(err)
 			return nil, err
 		}
 		filesList = append(filesList, hostFileList...)
@@ -121,7 +122,8 @@ func (s *StaticNetworkConfigGenerator) executeNMStatectl(ctx context.Context, ho
 	stdout, stderr, retCode := executer.ExecuteWithContext(ctx, "nmstatectl", "gc", f.Name())
 	if retCode != 0 {
 		s.log.Errorf("<nmstatectl gc> failed, errorCode %d, stderr %s, input yaml <%s>", retCode, stderr, hostYAML)
-		return "", fmt.Errorf("failed to execute nmstatectl gc, check the value provided for 'network_yaml': %s", hostYAML)
+		errMsg := strings.Split(stderr, "Error:")
+		return "", fmt.Errorf("invalid yaml, error: %s", strings.TrimSpace(errMsg[len(errMsg)-1]))
 	}
 	return stdout, nil
 }
@@ -189,7 +191,9 @@ func (s *StaticNetworkConfigGenerator) ValidateStaticConfigParams(ctx context.Co
 	var err *multierror.Error
 	for i, hostConfig := range staticNetworkConfig {
 		err = multierror.Append(err, s.validateMacInterfaceName(i, hostConfig.MacInterfaceMap))
-		err = multierror.Append(err, s.validateNMStateYaml(ctx, hostConfig.NetworkYaml))
+		if validateErr := s.validateNMStateYaml(ctx, hostConfig.NetworkYaml); validateErr != nil {
+			err = multierror.Append(err, fmt.Errorf("failed to validate network yaml for host %d, %s", i, validateErr))
+		}
 	}
 	return err.ErrorOrNil()
 }


pkg/staticnetworkconfig/generator.go
```diff
@@ -121,7 +122,15 @@ func (s *StaticNetworkConfigGenerator) executeNMStatectl(ctx context.Context, ho
...
 		s.log.Errorf("<nmstatectl gc> failed, errorCode %d, stderr %s, input yaml <%s>", retCode, stderr, hostYAML)
-		return "", fmt.Errorf("failed to execute nmstatectl gc, check the value provided for 'network_yaml': %s", hostYAML)
+		errMsg := stderr
+		errRegex := "Error: (?P<message>.*)"
+		re := regexp.MustCompile(errRegex)
+		matches := re.FindStringSubmatch(stderr)
+		msgIndex := re.SubexpIndex("message")
+		if len(matches) > msgIndex {
+			errMsg = matches[msgIndex]
+		}
```
carbonin, line 132: This is all rather complicated for error handling.
Can you give an example of the message we're matching against?

Also will the host index we have here be useful to the user? In the kubeapi, for example aren't nmstate yamls provided as individual resources? How would they figure out which one was index 3, for example?

CrystalChun: The type of message we would match against looks like this:
```
  File "/usr/bin/nmstatectl", line 33, in <module>
    sys.exit(load_entry_point('nmstate==1.2.1', 'console_scripts', 'nmstatectl')())
  File "/usr/lib/python3.10/site-packages/nmstatectl/nmstatectl.py", line 74, in main
    return args.func(args)
  File "/usr/lib/python3.10/site-packages/nmstatectl/nmstatectl.py", line 397, in run_gen_config
    libnmstate.generate_configurations(state), use_yaml=use_yaml
  File "/usr/lib/python3.10/site-packages/libnmstate/nmstate.py", line 419, in generate_configurations
    net_state = NetState(desire_state, gen_conf_mode=True)
  File "/usr/lib/python3.10/site-packages/libnmstate/net_state.py", line 55, in __init__
    self._route = RouteState(
  File "/usr/lib/python3.10/site-packages/libnmstate/route.py", line 204, in __init__
    self._merge_routes(des_route_state, ifaces)
  File "/usr/lib/python3.10/site-packages/libnmstate/route.py", line 223, in _merge_routes
    raise NmstateValueError(rt.invalid_reason)
libnmstate.error.NmstateValueError: Route {'table-id': 254, 'destination': '0.0.0.0/0', 'next-hop-address': '192.168.126.1', 'next-hop-interface': 'eth1'} next hop to unknown interface
```

So we would match that last line where it contains the actual error message
```
libnmstate.error.NmstateValueError: Route {'table-id': 254, 'destination': '0.0.0.0/0', 'next-hop-address': '192.168.126.1', 'next-hop-interface': 'eth1'} next hop to unknown interface
```

This is based on feedback received in https://issues.redhat.com/browse/MGMT-9574

The host seemed like it would be useful based on the feedback in the issue, but what do you think? Is it better to send back the host name instead?

carbonin: Is there any reason this can't just split on new line and take the last entry?
Maybe there's something we can do before to catch some simple validations like yaml format?

To be honest I think this is too much processing for an error string.

CrystalChun: Gotcha, splitting on a new line showed up a little strangely so I went ahead and just split on the `Error:` string instead of using the full regex. 

That's a good idea to add a yaml format validation. I'll look into adding that tomorrow! 

There are instances where it's not a yaml format error, though.

pkg/staticnetworkconfig/generator.go
```diff
@@ -59,7 +59,7 @@ func (s *StaticNetworkConfigGenerator) GenerateStaticNetworkConfigData(ctx conte
 	s.log.Infof("Start configuring static network for %d hosts", len(staticNetworkConfig))
 	filesList := []StaticNetworkConfigData{}
 	for i, hostConfig := range staticNetworkConfig {
-		hostFileList, err := s.generateHostStaticNetworkConfigData(ctx, hostConfig, fmt.Sprintf("host%d", i))
+		hostFileList, err := s.generateHostStaticNetworkConfigData(ctx, hostConfig, fmt.Sprintf("host%d", i), i)
 		if err != nil {
 			s.log.WithError(err).Errorf("Failed to create static config for host")
 			return nil, err
```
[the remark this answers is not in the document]
CrystalChun: Gotcha, I've attempted to do this, but from what I see, the way these functions are called doesn't expose the host id or other host information [[src](https://github.com/openshift/assisted-service/blob/95758658dfe1ba70eeb1bbf68740646602568970/pkg/staticnetworkconfig/generator.go#L54)]. 

One option would be to add a `Host ID` field to the [HostStaticNetworkConfig](https://github.com/openshift/assisted-service/blob/8f7bc60886cbab86104063d54d2a0e99fd538734/api/vendor/github.com/openshift/assisted-service/models/host_static_network_config.go#L19). Wdyt?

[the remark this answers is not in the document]
carbonin: I don't think that's a good solution for this.
That data is stored in the database, so we'd either need to accept that it wouldn't be present in some cases or somehow create a database migration to add it (not sure if we would even be able to do this).

Either of those options sound like more work than they're worth.

I think just stop passing the index around and add it to a wrapped error here.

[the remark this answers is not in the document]
CrystalChun: Ok I see, thank you for the explanations!

I went ahead and stopped passing the index and wrapped the error with the host index. How does it look now?

CrystalChun:
New error message now looks like this:
```
{"code":"400","href":"","id":400,"kind":"Error","reason":"1 error occurred:\n\t* failed to validate network yaml for host 0, invalid yaml, error: Route {'table-id': 254, 'destination': '0.0.0.0/0', 'next-hop-address': '192.168.126.1', 'next-hop-interface': 'eth1'} next hop to unknown interface\n\n"}
```

pkg/staticnetworkconfig/generator.go
```diff
@@ -61,7 +61,8 @@ func (s *StaticNetworkConfigGenerator) GenerateStaticNetworkConfigData(ctx conte
 	for i, hostConfig := range staticNetworkConfig {
 		hostFileList, err := s.generateHostStaticNetworkConfigData(ctx, hostConfig, fmt.Sprintf("host%d", i))
 		if err != nil {
-			s.log.WithError(err).Errorf("Failed to create static config for host")
+			err = fmt.Errorf("failed to create static config for host %d, %s", i, err)
```
carbonin, line 64: I think using `errors.Wrapf` would be the better choice for this.

CrystalChun: Good idea, thank you! Modified it.

carbonin (approved):

PR #3813, opened 2022-05-15
MGMT-9315: Create inventory cache during refresh status preprocessing, to avoid unnecessary inventory unmarshalling

ori-amizur:
When running refresh-status on cluster hosts, hostname-unique validation
checks that the current hostname is unique.  In order to get a host's
hostname, there is a need to unmarshall its inventory.  For cluster with
many hosts, the number of unmarshalls is (# of hosts)^2 (squared), since
each host has to unmarshall the inventory of the other hosts.
To change that, an inventory cache was added, so each inventory is
unmarshalled once per monitoring cycle.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @filanov 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -57,12 +57,33 @@ func (v ValidationStatus) String() string {
 	return string(v)
 }
 
+type InventoryCache map[string]*models.Inventory
+
+func (ic InventoryCache) GetOrUnmarshal(host *models.Host) (inventory *models.Inventory, err error) {
+	if host.Inventory == "" {
+		return nil, nil
+	}
+	var ok bool
+	key := host.ID.String() + "@" + host.InfraEnvID.String()
+	inventory, ok = ic[key]
+	if !ok {
+		inventory, err = common.UnmarshalInventory(host.Inventory)
+		if err != nil {
+			return
+		}
+		ic[key] = inventory
+	}
+	return
+
+}
+
 type validationContext struct {
 	host                    *models.Host
 	cluster                 *common.Cluster
 	infraEnv                *common.InfraEnv
 	inventory               *models.Inventory
 	db                      *gorm.DB
+	inventoryCache          InventoryCache
 	clusterHostRequirements *models.ClusterHostRequirements
 	minCPUCoresRequirement  int64
 	minRAMMibRequirement    int64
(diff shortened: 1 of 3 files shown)


internal/host/validator.go
```diff
@@ -94,18 +95,32 @@ func (c *validationContext) loadInfraEnv() error {
 	return err
 }
 
-func (c *validationContext) loadInventory() error {
-	if c.host.Inventory != "" {
-		var inventory models.Inventory
-		err := json.Unmarshal([]byte(c.host.Inventory), &inventory)
+func (c *validationContext) getInventory(host *models.Host) (inventory *models.Inventory, err error) {
```
carbonin, line 98: I feel like this would be better as a `Get` function on the InventoryCache type.

It doesn't look like this uses anything from the validation context other than the cache.

internal/host/validator.go
```diff
@@ -1133,11 +1149,10 @@ func (v *validator) hasDefaultRoute(c *validationContext) ValidationStatus {
 		return ValidationPending
 	}
 
-	inv, err := common.UnmarshalInventory(c.host.Inventory)
-	if err != nil || len(inv.Routes) == 0 {
+	if len(c.inventory.Routes) == 0 {
```
carbonin, line 1152: When does `c.inventory` get set? Are we sure it's set before this function is called?
Should this be using `c.getInventory(c.host)`?

It's a bit confusing to have an inventory cache and also a separate cache for the specific host's inventory. Maybe just remove the `c.inventory` field everywhere in favor of a helper that does `c.getInventory(c.host)`?

internal/host/validator.go
ori-amizur: OK

internal/host/validator.go
ori-amizur: It is loaded Here 
https://github.com/openshift/assisted-service/blob/dbe941db78105d7cbdf827a6b1400686f4b0fe9a/internal/host/validator.go#L192
or here
https://github.com/openshift/assisted-service/blob/dbe941db78105d7cbdf827a6b1400686f4b0fe9a/internal/host/validator.go#L217

Fixed to the format that exists in other places in the code like:
https://github.com/openshift/assisted-service/blob/dbe941db78105d7cbdf827a6b1400686f4b0fe9a/internal/host/validator.go#L566

carbonin (approved):

PR #3826, opened 2022-05-17
Bug 2076333: Add clusterDeploymentRef as label to the Agent

mkowalski:
This PR extends the Agent CR by creating labels with a reference to the
clusterDeployment. This is so that in the future it will be possible to
list agents belonging to a particular clusterDeployment only by
filtering Agents by label and not by iterating over all the objects and
querying their spec.

Due to the fact that maximum length for a label value is 63 characters 
but the resource name can be up to 253, we are labeling only the 
namespace as this one has the limit of 63 characters.

This leverages the fact that api server can filter objects by label, but
it cannot filter by the arbitrary field in spec.

The PR is part of a bigger change that is going to extend error
reporting in ClusterDeployment controller and AgentClusterInstall
controller.

Contributes-to: [MGMTBUGSM-307](https://issues.redhat.com//browse/MGMTBUGSM-307)

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @avishayt 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -60,7 +60,8 @@ import (
 
 const (
 	AgentFinalizerName   = "agent." + aiv1beta1.Group + "/ai-deprovision"
-	InventoryLabelPrefix = "inventory." + aiv1beta1.Group + "/"
+	BaseLabelPrefix      = aiv1beta1.Group + "/"
+	InventoryLabelPrefix = "inventory." + BaseLabelPrefix
 )
 
 // AgentReconciler reconciles a Agent object
@@ -208,7 +209,7 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 		return r.updateStatus(ctx, log, agent, origAgent, &h.Host, h.ClusterID, err, !IsUserError(err))
 	}
 
-	err = r.updateInventory(log, ctx, &h.Host, agent)
+	err = r.updateInventoryAndLabels(log, ctx, &h.Host, agent)
 	if err != nil {
 		return r.updateStatus(ctx, log, agent, origAgent, &h.Host, h.ClusterID, err, true)
 	}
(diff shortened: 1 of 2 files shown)


internal/controller/controllers/agent_controller.go
```diff
@@ -213,6 +213,11 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 		return r.updateStatus(ctx, log, agent, origAgent, &h.Host, h.ClusterID, err, true)
 	}
 
+	err = r.updateClusterDeploymentRefLabels(log, ctx, agent)
```
avishayt, line 216: I would move the call to updateInventoryLabels() from inside updateInventory() to here, and then unite updateInventoryLabels() and updateClusterDeploymentRefLabels() to just be updateLabels().  I think that makes the code more organized and we also update the resource only once.

internal/controller/controllers/agent_controller.go
```diff
@@ -1022,6 +1027,35 @@ func (r *AgentReconciler) updateInventoryLabels(log logrus.FieldLogger, ctx cont
...
+	name := ""
+	namespace := ""
+
+	if agent.Spec.ClusterDeploymentName != nil {
+		name = agent.Spec.ClusterDeploymentName.Name
+		namespace = agent.Spec.ClusterDeploymentName.Namespace
+	}
+
+	changed := false
+	changed = setAgentLabel(log, agent, "agent-install.openshift.io/clusterdeployment-name", name) || changed
```
avishayt, line 1040: Just realized that the maximum length for a label value is 63 characters, but a resource name can be up to 253.  Namespace can be up to 63 so I suggest we add a label only for the namespace - at least it will help us filter the clusterdeployments to down a namespace.  Unless you have other ideas?

mkowalski: Good catch, would have caused us some trouble sooner or later... Let's just label by namespace, that should be good enough for start.

internal/controller/controllers/agent_controller.go
mkowalski: Sure thing, just refactored it and put it all into a single function call

internal/controller/controllers/agent_controller.go
```diff
@@ -1006,6 +1008,12 @@ func (r *AgentReconciler) updateInventoryLabels(log logrus.FieldLogger, ctx cont
 	changed = setAgentLabel(log, agent, InventoryLabelPrefix+"host-productname", inventory.SystemVendor.ProductName) || changed
 	changed = setAgentLabel(log, agent, InventoryLabelPrefix+"host-isvirtual", strconv.FormatBool(inventory.SystemVendor.Virtual)) || changed
 
+	namespace := ""
+	if agent.Spec.ClusterDeploymentName != nil {
+		namespace = agent.Spec.ClusterDeploymentName.Namespace
+	}
+	changed = setAgentLabel(log, agent, "agent-install.openshift.io/clusterdeployment-namespace", namespace) || changed
```
avishayt, line 1015: Don't hard-code the key prefix

mkowalski: Fixed

PR #3839, opened 2022-05-23
AGENT-205: Fix OCP image build

vfreex:

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/Dockerfile.assisted-service.ocp
+++ b/Dockerfile.assisted-service.ocp
@@ -1,6 +1,6 @@
 # Build binaries
 FROM registry.ci.openshift.org/openshift/release:golang-1.17 as builder
-
+WORKDIR /src
 COPY . .
 RUN CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /build/assisted-service cmd/main.go
 RUN CGO_ENABLED=0 GOFLAGS="" GO111MODULE=on go build -o /build/assisted-service-operator cmd/operator/main.go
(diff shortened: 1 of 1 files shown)


osherdp:
I think usage of [quay.io/ocpmetal/oc-image:bug-1823143-multi-arch-ai-bug-2069976](http://quay.io/ocpmetal/oc-image:bug-1823143-multi-arch-ai-bug-2069976)  should be preserved, as the changes didn't get into oc 4.11
/hold
(just until we figure out the status of this patch, everything else looks good :) )

eranco74 (commented):
/hold
We need the custom oc for supporting installation with mirror registry

lranjbar:
/approve

Please note that this change is only to `Dockerfile.assisted-service.ocp` which is the assisted-service image landing in the OCP payload. From what I understand the Openshift policy is: "All the images in the OCP payload have to use the other official images in the OCP payload." We will have to get this fix in the official `oc` image in the OCP payload to support disconnected environments.

In the meantime, we need to have this image built by ART so that we can start testing the OCP payload official image for assisted-service properly. I will let others from the Openshift Agent Team weigh in on this but while this is unfortunate that this bug is still not fixed... this is likely the path forward for us.

@vfreex Thank you for opening this PR for us. 😃 

Agent team: @dhellmann @zaneb @rwsu @pawanpinjarkar @andfasano @bfournie @celebdor

zaneb:
If this just prevents us from doing disconnected installs then I think it is the right way to move forward for now. We may have to re-evaluate if the bug doesn't get a timely fix, but anything is better than nothing.
If the way that assisted-service interacts with the CLI does not work with the standard binary (e.g. passes a command-line argument that will be rejected?) then we might need to find another approach even in the short term. That's not going to involve using a base image from quay though, so I think we should go ahead with this change in either case.

/hold cancel

lranjbar:
JIRA issue tracker: AGENT-205

PR #3842, opened 2022-05-23
MGMT-10003: save-partlabel and save-partindex coreos-installer args are not honored because the partition is formatted previously

tsorya:
[MGMT-10003](https://issues.redhat.com//browse/MGMT-10003): save-partlabel and save-partindex coreos-installer args are not honored because the partition is formatted previously
If save-partlabel or save-partindex are added to installer args, skip
disk speed test and set installer flag to skip cleaning installation
disk

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
- [] No tests needed

## Assignees

/cc @avishayt 
/cc @

## Checklist

## Reviewers Checklist


pkg/commonutils/common_utils.go
```diff
@@ -18,3 +19,13 @@ func MeasureOperation(operation string, log logrus.FieldLogger, metricsApi metri
 		}
 	}
 }
+
+func SaveDiskPartitionsIsSet(installerArgs string) bool {
```
avishayt, line 23: Why is this redefined in internal/host/hostcommands/cmd_common.go?

tsorya: actually by mistake, leftover

internal/host/conditions.go
```diff
@@ -32,6 +33,11 @@ func (v *validator) isInstallationDiskSpeedCheckSuccessful(c *validationContext)
 	if c.infraEnv != nil {
 		return false
 	}
+
+	if hostutil.SaveDiskPartitionsIsSet(c.host.InstallerArgs) {
```
carbonin, line 37: Is there any way we could skip this in the agent and return a successful result?
Then we wouldn't have to check this everywhere the speed check matters, right?

I assume we can't because we don't have the installer args in the speed check command?

internal/host/host_test.go
```diff
@@ -2371,7 +2371,8 @@ var _ = Describe("UpdateImageStatus", func() {
 	}
 
 	tests := []struct {
-		name                  string
+		name string
```
carbonin, line 2374: :confused: why change this here?

internal/host/conditions.go
avishayt: Another way is to add a fake result into the DB, then the rest of the code will see that it was already checked and will skip the check.

internal/host/host_test.go
tsorya: actually it was make format that did it

internal/host/conditions.go
tsorya: @carbonin we can't as you said. 
@avishayt pushing fake values will require removing them in case install args will be changed. Not sure we want to enter this logic

tsorya: we can provide installer args to agent and lets it decide what to do with installer command and which result to return

avishayt: I'm OK with the code as it is, just trying to address @carbonin 's concern

internal/host/host_test.go
carbonin: Strange that it would add a space in a test structure, but :man_shrugging:

internal/host/conditions.go
carbonin: Not a big concern. If it's difficult I wouldn't worry too much about it.
It doesn't look like we're checking for this in too many other places so not a problem.

carbonin (approved):

tsorya:
/retitle Bug 2119308: save-partlabel and save-partindex coreos-installer args are not honored because the partition is formatted previously

PR #3849, opened 2022-05-23
MGMT-10445: Add enhancement for automating returning an agent to the infra-env

carbonin:
Without a BareMetalHost there is currently no way to automate the process
of unbinding a host from a cluster. When a user wants to reuse a host for a new 
cluster they have to use the assisted-service unbind API, then manually boot
the host back into the discovery image.

To support automating this procedure, the assisted-service will configure nodes
in installed clusters to run an agent. When appropriate this agent will
reconfigure the host to boot into a discovery image, allowing the host to
be made available for installation without the user manually interacting with
a management system to reboot the host into the discovery image.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @mhrivnak 
/cc @avishayt 
/cc @filanov 
/cc @oourfali 

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/auto-return-agent-to-infra-env.md
+++ b/docs/enhancements/auto-return-agent-to-infra-env.md
@@ -1,0 +1,251 @@
+---
+title: auto-return-agent-to-infra-env
+authors:
+  - "@carbonin"
+creation-date: 2022-05-17
+last-updated: 2022-07-18
+---
+
+# Automatically Return Agents to InfraEnv
+
+## Summary
+
+Without a BareMetalHost there is currently no way to automate the process
+of unbinding a host from a cluster. When a user wants to reuse a host for a new
+cluster they have to use the assisted-service unbind API, then manually boot
+the host back into the discovery image.
+
+To support automating this procedure, the assisted-service will configure nodes
+in installed clusters to run an agent. When appropriate this agent will
+reconfigure the host to boot into a discovery image, allowing the host to
+be made available for installation without the user manually interacting with
+a management system to reboot the host into the discovery image.
+
+## Motivation
+
+A user creating and managing clusters using the assisted-service may not have
+access to the systems used to configure and boot hosts. This is especially
+true in bare-metal use cases when a BareMetalHost is not present. The
+assisted-service can make this use case easier by handling the reprovisioning
+process.
+
+### Goals
+
+* Reconfigure a host to boot the discovery image as a part of unbinding from a
+  cluster without a BareMetalHost
+* Do not run an agent on a node if it is not being unbound
+
+### Non-Goals
+
+* Automate graceful removal of a node from an existing cluster
+* Booting a node that is unresponsive or not currently part of an installed
+  cluster
+* Registering an agent to an InfraEnv other than the one it is currently
+  associated with
+* Wiping a host's storage when it is removed from a cluster
+
+## Proposal
+
+When a host should be returned to the InfraEnv use the kubeconfig created when
+the cluster was installed to run an agent on the host as a privileged DaemonSet
+pod. This agent will handle downloading the required artifacts and rebooting
+the host into the discovery process.
+
+This process will only be used in situations where the assisted-service is sure
+to have connectivity to the installed cluster's API endpoint. In cases such as
+the SaaS this feature will not be used.
+
+### Timeline
+
+1. User unbinds the host from the cluster
+2. Assisted service creates a DaemonSet to run the assisted-installer-agent
+   on the host as a privileged pod
+  * If there is no BareMetalHost associated with the agent CR, the service
+    executes a new host state transition (`TransitionTypeReclaimHost`)
+  * Host moves to state `HostStatusReclaiming`
+3. The agent downloads the InfraEnv's kernel and initrd (including discovery
+   ignition) to the host's `/boot` partition
+4. The agent creates a boot entry to boot into the discovery kernel and initrd
+   and posts a successful step response to the service
+  * Host moves to state `HostStatusReclaimingPendingReboot`
+5. The agent reboots the host into the discovery image
+6. Discovery agent registers host
+  * Service moves host to `HostStatusDiscoveringUnbound`
+
+### User Stories
+
+#### Story 1
+
+As a cluster creator I want to use the assisted service to repurpose hosts
+from existing clusters without requiring access to the host management
+infrastructure (BMC, hypervisor, etc).
+
+### Implementation Details/Notes/Constraints [optional]
+
+* The boot artifact URLs (initrd, kernel, and rootfs) will be provided to the
+  agent directly as a part of the pod spec that runs the agent.
+* The agent will require two new commands; one for each of the new states
+  1. Download boot artifacts and write boot entry (`HostStatusReclaiming`)
+  2. Reboot the node (`HostStatusReclaimingPendingReboot`)
+* The existing unbind API will be used to trigger this process
+  * This means removing the cluster ref from the agent CR when using the kubeAPI
+
+#### Example boot entry
+
+```
+$ cat /boot/loader/entries/00-assisted-discovery.conf
+title Assisted Installer Discovery
+version 999
+options random.trust_cpu=on ignition.firstboot ignition.platform.id=metal 'coreos.live.rootfs_url=http://10.1.155.24:6016/boot-artifacts/rootfs?arch=x86_64&version=4.10'
+linux /boot/discovery/vmlinuz
+initrd /boot/discovery/c4bd9751-f3ae-49ac-a36b-d3ffebef4a39-initrd.img
+```
+
+The name of this file and the version number ensure that it should always be the
+first boot entry and will be selected upon reboot.
+
+#### State Machine Changes
+Currently when a host is installed, all unbind APIs will call the `UnbindHost`
+transition. This will always move the host to
+`HostStatusUnbindingPendingUserAction`. If the agent has a BMH associated with
+it, the bare metal host agent controller (BMAC) will make the changes required
+to boot the host back into discovery. If there is no BMH associated with the
+agent (or the kube API is not in use), the host will remain in
+`HostStatusUnbindingPendingUserAction`.
+
+The following diagram illustrates the new proposed transition to indicate that a
+running agent should execute the reclaim procedure.
+
+```
+                   -----------------------
+                   | HostStatusInstalled |
+                   -----------------------
+                             |
+                      User unbinds agent
+                             |
+     Kube API and no BMH     |   REST API, Kube API with BMH, or other machine management
+          -------------------+-------------------
+          |                                     |
+          V                                     V
+-----------------------------            ----------------------------
+| TransitionTypeReclaimHost |            | TransitionTypeUnbindHost |
+-----------------------------            ----------------------------
+            |                                                      |
+            V                                                      |
+  ------------------------                                         |
+  | HostStatusReclaiming |                                         |
+  ------------------------                                         |
+                |                                                  |
+       Agent posts step reply                                      |
+                |                                                  |
+       ---------+----------------------------------                |
+       |                                          |                |
+    Success                                    Failure             |
+       |                                          |                |
+       V                                          |                |
+---------------------------------                 |                |
+| HostStatusReclaimingRebooting |----------       |                |
+---------------------------------      Timeout    |                |
+                |                         |       |                |
+------------------------------   -------------------------------   |
+| TransitionTypeRegisterHost |   | TransitionTypeReclaimFailed |   |
+------------------------------   -------------------------------   |
+                |                         |       |                |
+                V                         V       V                V
+--------------------------------  ----------------------------------------
+| HostStatusDiscoveringUnbound |  | HostStatusUnbindingPendingUserAction |
+--------------------------------  ----------------------------------------
+              ^                                       |
+              |       ------------------------------  |
+              --------| TransitionTypeRegisterHost |---
+                      ------------------------------
+```
+
+#### Assumptions
+
+* This approach only works if the host is currently part of a working cluster
+  - This cluster needs to have been installed by assisted service so that
+    it has access to the kubeconfig
+* This reclaim procedure will only be triggered if assisted service is running
+  the kubernetes API controllers
+  - Assisted service needs to ensure the host isn't being managed by a bare
+    metal host (only possible with the kubernetes API)
+  - When using the REST API there is no way for a user to indicate if a host
+    should be unbound normally or if it should be reclaimed
+* There must be enough free space on the `/boot` partition
+  - Space is required to store the initrd and kernel (~90MB)
+  - The agent can check for this
+* A node in a spoke cluster can be identified by the hostname in the agent
+  - In a non-CAPI scenario there are no other reliable means to identify which
+    node maps to which agent
+  - With CAPI a string of references exist that could be used to identify the
+    node, but the hostname should be reliable in this case as well
+
+### Risks and Mitigations
+
+It's possible that something could go wrong during the boot process. A user
+that doesn't have access to the host infrastructure wouldn't be able to debug
+or solve this type of problem.
+
+A timeout will be implemented when the service expects the host to reregister
+(in `HostStatusReclaimingPendingReboot`). After this time the host will move to
+the existing `HostStatusUnbindingPendingUserAction` state.
+
+If the node hasn't been removed from the cluster when the host reboots the host
+could boot back into the node and the pod running the agent would attempt to
+register. The assisted service would not know the difference between this
+scenario and if the node had booted into discovery. To deal with this, the
+next-step-runner process will be started in the pod rather than the full agent
+as the only difference is the registration step.
+
+If the node being reclaimed is being managed by a machine API provider in the
+installed cluster it doesn't make much sense for assisted to be attempting to
+manage it as well. If a node is unexpectedly removed from a cluster with
+machine management it's likely that the underlying provider will remove the
+infrastructure behind that node rendering it unusable by assisted. Assisted
+service should attempt to determine if the node is under some other management
+before starting the agent. If it is, proceed as before and move the node to
+`HostStatusUnbindingPendingUserAction`.
+
+## Design Details [optional]
+
+### Open Questions
+
+### UI Impact
+
+No UI changes will be required as this procedure will be triggered using the
+existing unbind action.
+
+### Test Plan
+
+Any existing end to end late binding test could be expanded to include
+returning the agents to the infraEnv after a successful install.
+
+## Drawbacks
+
+This will only work in environments where the service has access to the
+kubernetes API on the cluster and where we using the agent kubernetes API.
+This specifically means that it won't be usable in the SaaS.
+
+## Alternatives
+
+1. Install an agent when the cluster is first installed
+  - This would be a full polling approach where the agent is always polling the
+    service to determine if it should reprovision.
+  - This is likely not worth the resources and clusters don't exist long enough in
+    the cloud for this to work currently.
+2. Implement starting the agent in MAPI and/or CAPI
+  - These components are already the starting point for removing a node from a
+    cluster and would likely be the ones triggering the unbind
+  - In this case the assisted-service wouldn't require the spoke cluster
+    kubeconfig as MAPI/CAPI should already have this access.
+  - A lot of information will be required to properly start the agent
+    - Agent image pull spec
+    - Hub cluster CA cert
+    - http(s) proxy information
+    - Agent authentication information (pull secret)
+    - Boot artifact URLs
+  - There's not much relevant difference between doing any of this in the CAPI
+    provider vs assisted
+  - Adding a MAPI provider would involve creating our own platform type which
+    is not something we want to take on


docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+downloading the required artifacts and rebooting the host into the discovery
+process.
+
+### Timeline
+
+1. User triggers an action which will prepare a host to be unbound
+  * Assisted service will create a MachineConfig to run our agent on the host as
+    a systemd service.
+  * The agent will download the InfraEnv's kernel and initrd (including
+    discovery ignition) to the host's `/boot` partition.
```
omertuc, line 57: Is that the EFI partition mounted on `/boot`?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+  * The agent will create a GRUB boot entry to boot into the discovery kernel
```
omertuc, line 58: worth mentioning that the entry will also make use of the discovery ignition

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+    and initrd.
+2. User (or other automation) removes the node from the cluster
+  * Drain existing workloads and delete the node
```
omertuc, line 61:
```suggestion
  * Cordon the node
  * Drain existing workloads and delete the node
```

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+## Proposal
+
+When a host should be returned to the InfraEnv use the kubeconfig created when
+the cluster was installed to run an agent on the host. This agent will handle
+downloading the required artifacts and rebooting the host into the discovery
+process.
+
+### Timeline
+
+1. User triggers an action which will prepare a host to be unbound
```
omertuc, line 53: What happens if the host is a master node? Assisted doesn't know how to add master nodes to a cluster, so it kinda leaves you with a broken cluster that you cannot fix using just assisted

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+  * A step to reboot the host into the new boot entry
+3. The service will need a new API to prepare a cluster for hosts to be reclaimed
+  * This needs to be separate from unbind so that the user has time to cleanly
+    remove a host from the cluster
+  * `/v2/clusters/{cluster_id}/prepare-unbind`?
+
+#### Assumptions
+
+This approach only works if the host is currently part of a working cluster
+which the assisted service can access using the kubeconfig generated during
```
omertuc, line 95: We garbage collect the kubeconfig on the SaaS deployment. Is this enhancement about our kube-api deployments only?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Non-Goals
+
+* Automate graceful removal of a node from an existing cluster
+* Booting a node that is unresponsive or not currently part of an installed
+  cluster
+
+## Proposal
+
+When a host should be returned to the InfraEnv use the kubeconfig created when
```
omertuc, line 46: We garbage collect the kubeconfig on the SaaS deployment. Is this enhancement about our kube-api deployments only?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### Story 1
+
+As a cluster creator I want to use the assisted service to repurpose hosts 
+from existing clusters without requiring access to the host management 
+infrastructure (BMC, hypervisor, etc).
+
+### Implementation Details/Notes/Constraints [optional]
+
+1. A way to query for URLs to the iPXE boot artifacts over the REST API should be
```
omertuc, line 77:
```suggestion
1. A way to query for URLs to the iPXE boot artifacts over the REST API should be
```

The artifacts are used in ipxe, but this enhancement has nothing to do with iPXE, right? Could you clarify this in the enhancement?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+implemented. Currently these URLs are only exposed through the iPXE script and
```
omertuc, line 78: exposed through == one could technically download the iPXE script, parse it and find out those URLs?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+    remove a host from the cluster
+  * `/v2/clusters/{cluster_id}/prepare-unbind`?
+
+#### Assumptions
+
+This approach only works if the host is currently part of a working cluster
+which the assisted service can access using the kubeconfig generated during
+install.
+
+There will also need to be enough space on the `/boot` partition to download the
```
omertuc, line 98: Is this something we can guarantee? Looking at a random host on my cluster it looks like the boot partition only has 384MB left, which IIRC is not enough for initrd

docs/enhancements/auto-return-agent-to-infra-env.md
omertuc: EDIT: I see you "addressed" this under open questions

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+## Design Details [optional]
+
+### Open Questions
+
+1. Will users want to reclaim individual hosts in an existing cluster or will
+   the whole cluster always be decommissioned at the same time?
+  * This will determine how we target the MachineConfig
+  * Ideally we don't want to be running an agent on nodes we know we don't want
+    to reclaim
+2. How long is the kubeconfig for a particular cluster saved?
```
omertuc, line 117: For SaaS, I think it's on the order of 1 month or something of that sort

For kubeapi, it's saved forever on the hub

For on-prem REST, I think we do garbage collection but we should probably disable that

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+kubernetes API on the cluster. This specifically means that it won't be useful
+in the SaaS.
+
+## Alternatives
+
+The only strategy which would work in the cloud would be a full polling 
+approach where we set up an agent on the node at install that constantly asks
+the service if it should reprovision.
+
+This is likely not worth the resources and clusters don't exist long enough in
```
omertuc, line 139: It also sounds like a security nightmare. I wouldn't want my AI installed clusters to be potentially controlled by the installation service that helped me install it, it feels wrong

docs/enhancements/auto-return-agent-to-infra-env.md
romfreiman: I think we can make it as a non goal in the first step (does not mean that we dont have to think about it of course)

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+2. User (or other automation) removes the node from the cluster
+  * Drain existing workloads and delete the node
+3. User unbinds the host from the cluster using the assisted-service API
+  * This triggers the agent to reboot the host into the discovery image
+4. Host is registered as `known-unbound` in the InfraEnv and is ready to be used
+   in new clusters
+
+### User Stories
+
+#### Story 1
```
romfreiman, line 69: what about auto scale (downscale in this case)? I assume the autoscaler runs within the spoke cluster, right? then what? how the hub knows?

romfreiman, line 69: lets think about hypershift/capi as well here
@eranco74 @avishayt

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+
+When a host should be returned to the InfraEnv use the kubeconfig created when
+the cluster was installed to run an agent on the host. This agent will handle
+downloading the required artifacts and rebooting the host into the discovery
+process.
+
+### Timeline
+
+1. User triggers an action which will prepare a host to be unbound
+  * Assisted service will create a MachineConfig to run our agent on the host as
```
eranco74, line 54: Will this new state get reflected in the host status/ agent conditions?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+It's possible that something could go wrong during the boot process. A user 
+that doesn't have access to the host infrastructure wouldn't be able to debug 
+or solve this type of problem. In this case the user is no worse off than they
+currently would be and will have to boot the discovery ISO manually.
+
+## Design Details [optional]
+
+### Open Questions
+
+1. Will users want to reclaim individual hosts in an existing cluster or will
```
eranco74, line 112: I think the agent will have different logic when removing a single node (cordon, wait for drain and reboot) vs decommissioning the cluster - it shouldn't care about the cordon but should wait with the reboot until all nodes are running the agent (so no one is left behind)

docs/enhancements/auto-return-agent-to-infra-env.md
romfreiman: I think we can assume it's there - otherwise ACM wont work neither, right?

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: I figured it would be the same image.
It would still be querying the service for commands to run on a host. Why make it a different thing?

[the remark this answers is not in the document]
carbonin: The `unbind` API currently is a sub-resource of the infraEnv so the original infraEnv must still exist otherwise the host would not be accessible.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Yeah, I struggled with how to reference the discovery runtime without using `iso` or `image` everywhere.
Do you have a good suggestion for how to reference "running the same stuff as when we boot the discovery iso"?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+### Timeline
+
+1. User triggers an action which will prepare a host to be unbound
+  * Assisted service will create a MachineConfig to run our agent on the host as
+    a systemd service.
+  * The agent will download the InfraEnv's kernel and initrd (including
+    discovery ignition) to the host's `/boot` partition.
+  * The agent will create a GRUB boot entry to boot into the discovery kernel
+    and initrd.
+2. User (or other automation) removes the node from the cluster
```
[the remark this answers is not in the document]
carbonin: Talking to @filanov it sounded like this wasn't an area we really wanted to be involved in.
We're an installer, not a full management application. It feels like this is the job of some higher-level automation.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: :man_shrugging: I'll make this correct, but it's not something we're implementing as a part of this enhancement.
I just added this as an example of what some other user or system would need to do.

Really I thought this was needed to explain why we couldn't do the entire process as a part of the `unbind` API.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: I don't really understand this question.
Is this related to the question about hypershift and capi below?

carbonin: Right, this is the assisted-specific portion of that process.
Presumably hypershift or CAPI would utilize this new capability in assisted to make the machine removal process better.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Technically, yes one could do that, but lets not.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+
+As a cluster creator I want to use the assisted service to repurpose hosts 
+from existing clusters without requiring access to the host management 
+infrastructure (BMC, hypervisor, etc).
+
+### Implementation Details/Notes/Constraints [optional]
+
+1. A way to query for URLs to the iPXE boot artifacts over the REST API should be
+implemented. Currently these URLs are only exposed through the iPXE script and
+in the InfraEnv CR, but the agent running on a node will need access to these
```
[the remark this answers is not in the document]
carbonin: That's fair. We could specify the initrd and kernel urls as a part of the systemd service config in the machine config.
Then the agent wouldn't need to query for them at all.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+artifacts.
+  * This can be added to the existing payload returned when querying an InfraEnv
+  * Alternatively this would need a new API endpoint
+    * `/v2/infra-envs/{infra_env_id}/boot-artifacts`?
+2. The agent will likely require two changes/additions
+  * A step to query the boot artifacts and write the boot entry
+  * A step to reboot the host into the new boot entry
```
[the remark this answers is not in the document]
carbonin: The agent won't know. The service will tell it when it should do the reboot.
In turn the service knows because the user said it was safe to unbind the host.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: That's more than enough. The initrd and kernel together are only about 80-90MB

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Yeah, that's why this was split into two steps.
1. Run the agent on the hosts
** User or other automation ensures the cluster is ready for hosts to be removed here **
2. Reboot into discovery

I think I described this in the `Timeline` section of `Proposal`

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: I mean that's what we're doing anyway though, right?
We're still using the kubeconfig that we generated on install to execute privileged operations on your machines from the installer.

We're just only running the agent when you tell us to rather than running it the entire time.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
avishayt: In the case of SaaS I agree that we're an installer, but in the case of CIM/hypershift we're doing full lifecycle.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
avishayt: I can't think of a compelling use case for switching InfraEnvs, but we don't have much feedback from users yet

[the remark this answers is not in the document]
oourfali: In my view, we think about the assisted-installer-agent as our agent for commands, whether those are in install time, day2, late binding, unbinding and etc, then we should rely on it for that as well.
Is this indeed the way we see it? Is our agent generic enough?

If, however, we think that the current agent fits mostly for installation, and we should have another agent running at all times for such operations (current ones and also ones we think we'll need to add), then we can consider adding another one.

docs/enhancements/auto-return-agent-to-infra-env.md
oourfali: our main assumption, iiuc, is that the removal itself is done out of the assisted installer, and the automated part is triggering the unbounding and reuse of the host in another provisioning.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Yeah I'll make it clear earlier on that this specifically will never be relevant for SaaS.
Among other reasons (like this one) there's no reason to assume we can access the API endpoint of clusters installed with the SaaS.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: As mentioned in other threads this is never going to be relevant in the SaaS as there's no reason to assume we can access the kubernetes API server.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: It's possible that a user would want to reclaim all the nodes in a cluster at once though, right?
I don't think we want to disable this totally for control plane nodes.

I mention this in the "Open Questions" section. Specifically I mention how we target the MachineConfig.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: I'm going to resolve this as the agents are always directly associated with a single infraEnv. If that infraEnv was removed then the agents are either gone or already associated with another infraEnv.

I'll add "moving an agent to a different infraEnv" as a non-goal.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: I also mentioned that the initrd includes the ignition in the previous bullet. Is that clear enough?

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: Added this as an option below.

docs/enhancements/auto-return-agent-to-infra-env.md
filanov: Very good point, in case of hypershift it will not be a problem because we don't manage those  and it will mean that user deleted the cluster, we just need to make sure that he is doing it after we unbound the hosts.
In case of full bare metal i think that we can start without it as rom suggested and then improve the process.

docs/enhancements/auto-return-agent-to-infra-env.md
filanov: i don't think that we will be able to run the agent before the host is unbound, if i remember correctly once it will try to  register the sate machine will reject it because the host is installed, we will probably need to change that, or change the order.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
filanov: something is missing here for me as well, currently unbind mean that user removed the cluster reference, i'm not sure i understand the flow here.

docs/enhancements/auto-return-agent-to-infra-env.md
eranco74: So currently, what happens is (1 & 4).
Once CAPI sees that a node should be removed from a cluster it removes it from the cluster. We need to figure out how to block  CAPI from removing the node until 2 & 3 happen.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,140 @@
...
+This approach only works if the host is currently part of a working cluster
+which the assisted service can access using the kubeconfig generated during
+install.
+
+There will also need to be enough space on the `/boot` partition to download the
+initrd and kernel, but the agent can check for this during the preparation step.
+
+### Risks and Mitigations
+
+It's possible that something could go wrong during the boot process. A user
```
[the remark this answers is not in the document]
eranco74: It can, this is a new agent functionality that requires new steps.
The server can parse the step response and update something on the agent so that the agent controller will be able to set the info on some condition.
Currently we update the agent conditions by looking at the validations and installation status/progress

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,161 @@
...
+
+* Automate graceful removal of a node from an existing cluster
+* Booting a node that is unresponsive or not currently part of an installed
+  cluster
+* Registering an agent to an InfraEnv other than the one it is currently
+  associated with
+
+## Proposal
+
+When a host should be returned to the InfraEnv use the kubeconfig created when
```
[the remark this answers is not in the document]
avishayt: Not sure what you mean.  Do you mean to run the agent on the existing openshift node?  It's possible, but once the agent starts writing to disk the host will be unresponsive because it's running from the same disk (assuming you want to install on the same disk).

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
avishayt: I didn't mean Assisted needs to do everything, but it could trigger the node maintenance operator, for example
https://github.com/medik8s/node-maintenance-operator

docs/enhancements/auto-return-agent-to-infra-env.md
avishayt: In case of hypershift/capi, autoscaler will run on the hub

avishayt: The CAPI provider also does 5 (unbind)

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,161 @@
...
+1. User triggers an action which will prepare a host to be unbound
+  * Assisted service will create a MachineConfig to run the 
+    assisted-instller-agent on the host as a systemd service.
+  * The agent will download the InfraEnv's kernel and initrd (including
+    discovery ignition) to the host's `/boot` partition.
+  * The agent will create a GRUB boot entry to boot into the discovery kernel
+    and initrd.
+2. User (or other automation) removes the node from the cluster
+  * Cordon the node
+  * Drain existing workloads
```
nmagnezi, line 68: is this a part of the Epic? 
How does it work? by using some existing k8s features or by implementing something else?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,161 @@
...
+the host state machine to indicate when a host has run through the preparation
+steps and when it is ready to be unbound and moved back to the InfraEnv pool.
+
+#### Assumptions
+
+This approach only works if the host is currently part of a working cluster
+which the assisted service can access using the kubeconfig generated during
+install.
+
+There will also need to be enough space on the `/boot` partition to download the
```
nmagnezi, line 105: should we run some pre-flight validation for this?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,161 @@
...
+## Proposal
+
+When a host should be returned to the InfraEnv use the kubeconfig created when
+the cluster was installed to run an agent on the host. This agent will handle
+downloading the required artifacts and rebooting the host into the discovery
+process.
+
+This process will only be used in situations where the assisted-service is sure
+to have connectivity to the installed cluster's API endpoint. In cases such as
+the SaaS this feature will not be used.
```
nmagnezi, line 55: I think I understand why it won't be used in SaaS, but still might be worth mentioning it here.

carbonin: I did, right?

>This process will only be used in situations where the assisted-service is sure to have connectivity to the installed cluster's API endpoint

carbonin: I did, right?

>This process will only be used in situations where the assisted-service is sure to have connectivity to the installed cluster's API endpoint

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: >if i remember correctly once it will try to register the sate machine will reject it because the host is installed, we will probably need to change that, or change the order.

Yeah, I think the state machine will need some changes as a part of this (we talked about this today).
Perhaps the agent will need to query the service before attempting to register to allow the service to know the difference between an agent running on an installed node and one running as a part of discovery.

carbonin: >Once CAPI sees that a node should be removed from a cluster it removes it from the cluster. We need to figure out how to block CAPI from removing the node until 2 & 3 happen.

>The CAPI provider also does 5 (unbind)

It sounds like for this to work we will need to break some of this up.
If CAPI is going to trigger us to move a host back into the infraenv using this process, then the host needs to still be a part of the cluster in order for us to get the agent running on it.

At the very least we need CAPI to unbind(or whatever this new process gets called) before it removes the node from the cluster.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: Talked to @filanov about this.
I'm working on adding some details about the changes to the host state machine that I have in mind.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: >but the agent can check for this during the preparation step

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: > Currently we update the agent conditions by looking at the validations and installation status/progress

Yeah, I think we would have some new status for this. Definitely for the time between when we tell the node to reboot and when it re-registers. I'll add details about the statemachine changes and new agent steps soon.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,161 @@
...
+
+* The boot artifact URLs will be provided to the agent directly as a part of the
+  machine config that runs the agent.
+* The agent will require two new commands
+  * A step to download the boot artifacts and write the boot entry
+  * A step to reboot the host into the new boot entry
+* The service will need a new API to prepare a cluster for hosts to be reclaimed
+  * This needs to be separate from unbind so that the user has time to cleanly
+    remove a host from the cluster
+  * `/v2/clusters/{cluster_id}/prepare-unbind`?
```
mhrivnak: This is preparing to reclaim an entire cluster at once? And is separate from just preparing a single host for removal from a cluster?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,187 @@
...
+Any existing end to end late binding test could be expanded to include
+returning the agents to the infraEnv after a successful install.
+
+## Drawbacks
+
+This will only work in environments where the service has access to the
+kubernetes API on the cluster. This specifically means that it won't be useful
+in the SaaS.
+
+## Alternatives
```
mhrivnak, line 180: Another alternative is to require use of a mapi or capi provider for agents. It could implement the whole workflow when a Machine gets deleted of running a new agent and draining the node. And it would work with the SaaS in theory since the source of truth on whether a node should still be in a cluster is whether its Machine resource has been deleted.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,187 @@
...
+  * Host moves to state `HostStatusReclaiming`
+3. The agent downloads the InfraEnv's kernel and initrd (including discovery 
+   ignition) to the host's `/boot` partition
+4. The agent creates a GRUB boot entry to boot into the discovery kernel and 
+   initrd
+5. The agent reboots the host into the discovery image
+6. Discovery agent registers host
+  * Service moves host to `HostStatusDiscoveringUnbound`
+
+### User Stories
```
mhrivnak, line 72: Story 2: "As a cluster creator, I want data wiped from hosts when they are removed from my clusters to avoid data leakage to other people who create and use clusters in the same environment."

If we're providing a complete host-recycling workflow, this is an important piece.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: With the last changes I made, the only thing we'll do is move to `UnbindingPendingUserAction` if the agent has some problem.

When the agent responds after running into an issue that response should have the error info, but currently it's not a validation so we'd have to find somewhere else to put it.

Do you think this is worth adding a new column to host for?
I guess the alternative would be to add a "preparing" state of some kind and some validation steps. Artifact access, and disk space come to mind.

What do you think?

mhrivnak (changes requested):
Generally I think it would help to draw a brighter line between the use cases of removing one host vs. removing all the hosts in a cluster. If the latter is out of scope, let's state that explicitly.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: With the current proposal I don't think there's really any way for us to know the state of the other nodes in the cluster from the service.

It would just rely on the user doing everything they could before actually telling assisted-service to reclaim the node which would then directly cause it to reboot.

I think it would be enough to document this.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,161 @@
...
+* Automate graceful removal of a node from an existing cluster
+* Booting a node that is unresponsive or not currently part of an installed
+  cluster
+* Registering an agent to an InfraEnv other than the one it is currently
+  associated with
+
+## Proposal
+
+When a host should be returned to the InfraEnv use the kubeconfig created when
+the cluster was installed to run an agent on the host. This agent will handle
```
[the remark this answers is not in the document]
carbonin: Yes

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,161 @@
...
+
+This process will only be used in situations where the assisted-service is sure
+to have connectivity to the installed cluster's API endpoint. In cases such as
+the SaaS this feature will not be used.
+
+### Timeline
+
+1. User triggers an action which will prepare a host to be unbound
+  * Assisted service will create a MachineConfig to run the 
+    assisted-instller-agent on the host as a systemd service.
```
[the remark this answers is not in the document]
carbonin: Would a `DaemonSet` have permissions to do what we need (remount `/boot` as read/write, add a boot entry, and reboot)?
Also do `DaemonSet`s get removed if someone drains a node?

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
mhrivnak: I re-posted this comment. Github failed to keep it with the text during a revision.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
mhrivnak: I re-posted this comment. Github failed to keep it with the text during a revision.

docs/enhancements/auto-return-agent-to-infra-env.md
mhrivnak: Looks like this no longer appears in the proposal, so you can resolve/ignore.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Risks and Mitigations
+
+It's possible that something could go wrong during the boot process. A user 
+that doesn't have access to the host infrastructure wouldn't be able to debug 
+or solve this type of problem. In this case the user is no worse off than they
+currently would be and will have to boot the discovery ISO manually.
+
+A timeout will be implemented when the service expects the host to reregister.
+After this time the host will move to the existing unbinding-pending-user-action
```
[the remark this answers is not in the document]
carbonin: There is an existing transition to move a host from this state to `DiscoveringUnbound` when an agent registers.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,187 @@
...
+artifacts and rebooting the host into the discovery process.
+
+This process will only be used in situations where the assisted-service is sure
+to have connectivity to the installed cluster's API endpoint. In cases such as
+the SaaS this feature will not be used.
+
+### Timeline
+
+1. User unbinds the host from the cluster
+2. Assisted service creates a MachineConfig to run the
```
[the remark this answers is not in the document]
carbonin: Would a DaemonSet have permissions to do what we need (remount /boot as read/write, add a boot entry, and reboot)?
Also do DaemonSets get removed if someone drains a node?

docs/enhancements/auto-return-agent-to-infra-env.md
mhrivnak: If/when capi or mapi is involved, I think it makes more sense for it to start the agent on the node being removed. The mapi/capi provider is guaranteed to always have access to the k8s API.

carbonin: As far as I know CAPI is the main use case here, right? @avishayt @filanov ?

The only problem with that is that the CAPI provider wouldn't (and shouldn't IMO) know how to correctly start our agent for a particular infraenv. We could change this proposal to expose an endpoint that would return the ignition that we would put in a machine config for a specific host though.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: I could add to the diagram to show more of what already exists if that would be helpful.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: I like this idea, but as I mentioned in the other thread I think the service should still be providing whatever information we use to define how the agent runs (ignition, pod spec, whatever else it could be).

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: Added.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,187 @@
...
+* Automate graceful removal of a node from an existing cluster
+* Booting a node that is unresponsive or not currently part of an installed
+  cluster
+* Registering an agent to an InfraEnv other than the one it is currently
+  associated with
+
+## Proposal
+
+When a host should be returned to the InfraEnv use the kubeconfig created when
+the cluster was installed to run an agent on the host as a systemd service
```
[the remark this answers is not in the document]
carbonin: I mean, this sentence says pretty much that, right? "When a host should be returned ... run an agent on the host ...".
I suppose we should add more detail about how we're going to target this to individual nodes so we're not causing a bunch of reboots with machineconfigs.

This ties a bit into the open question about if someone is going to do this for an entire cluster or not, but I think the more useful question is how to target the configs since the entire entrypoint is unbinding a single host. I someone wants to do this for all hosts in a cluster it will just be multiple rounds of the same logic.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: I'm genuinely curious about this.
I did a look around the documentation, but don't see anything indicating that any special permissions are granted to DaemonSet pods.

The only thing I did find was some information about how they're scheduled slightly differently, but nothing that would answer my question about draining. (Although I'll probably just test this in a bit after I get a cluster installed).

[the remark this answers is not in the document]
mhrivnak: The `oc adm drain` command has a `--ignore-daemonsets` flag that's important to use, because otherwise the drain won't remove daemonset pods.

It just runs a Pod; can we give a Pod the needed permissions? I would think so unless there's a gotcha.

docs/enhancements/auto-return-agent-to-infra-env.md
mhrivnak: I would think a MachineSet would be associated with an InfraEnv.

docs/enhancements/auto-return-agent-to-infra-env.md
mhrivnak: Yep. I'd think a MachineSet maps to an InfraEnv, and somehow the InfraEnv would provide the info needed to run the agent, though I'm not sure it needs to provide the pod spec necessarily.

To be clear, I don't necessarily think this is a better idea, but it is definitely worth listing as an alternative and at least explaining why we're not going this direction.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Added

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
tsorya: We will push machine config with some label? and add this label to node?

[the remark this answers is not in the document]
tsorya: Will MCO reboot the host to apply the change?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,200 @@
...
+          -------------------+-------------------
+          |                                     |
+          V                                     V
+-----------------------------            ----------------------------
+| TransitionTypeReclaimHost |            | TransitionTypeUnbindHost |
+-----------------------------            ----------------------------
+            |                                               |
+            V                                               |
+  ------------------------                                  |
+  | HostStatusReclaiming |                                  |
```
carbonin, line 135: Do we do anything when the agent registers here? Or just no-op that transition?

carbonin: @filanov?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,200 @@
...
+
+#### Story 1
+
+As a cluster creator I want to use the assisted service to repurpose hosts
+from existing clusters without requiring access to the host management
+infrastructure (BMC, hypervisor, etc).
+
+#### Story 2
+
+As a cluster creator, I want the host's storage wiped when it is removed from a
```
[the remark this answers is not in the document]
mhrivnak: Data also for cases where the users of a cluster-as-a-service want to prevent data leakage to other users. Or even just ensure other drives are in a consistent clean state for discovery and consumption by thing like ODF.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: I think it's probably best to add another status after this one.

We'll move to `HostStatusReclaiming` when a user unbinds. The agent will be given a step to set up the host for booting back into discovery.
Move from `HostStatusReclaiming` to `HostStatusReclaimReboot` when the agent reports that the artifacts and boot entry have been set up properly (posted step response)
Move from `HostStatusReclaimReboot` to `HostStatusDiscoveringUnbound` when the host registers.

If the host fails in Reclaiming or times out in ReclaimReboot we'll move to `HostStatusUnbindingPendingUserAction`

This means we can avoid creating a reclaim-specific status column in the host and will more clearly indicate that the host is rebooting.
This also reserves UnbindingPendingUserAction for cases where user (or bmac) action is actually required. We can also continue to use this state for our error case.

What do you think @filanov ?

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: Also after looking more into machineconfigs it looks like the pool concept will make this difficult to target something to run only on a single node.

Specifically these bits of documentation:
> MCO applies changes to operating systems in pools of machines.
> A node can have multiple labels applied that indicate its type, such as master or worker, however it can be a member of only a single machine config pool

Maybe it won't be too hard to get the agent to run as a very privileged daemonset pod :pray:

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: > I would think a MachineSet would be associated with an InfraEnv.

I don't think this is the case currently, but I also don't think it matters.
As far as I can tell the CAPI provider just maps AgentMachines to Agents and has no knowledge of the infraEnv.
That said it could presumably find the infraEnv based on the label on the agent.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Added another state below to track reboot.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: I added an alternative here for MAPI/CAPI, but I don't think it's something we can do because we also want this to be available for a fully boot-it-yourself user.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,234 @@
...
+  - This would be a full polling approach where the agent is always polling the 
+    service to determine if it should reprovision.
+  - This is likely not worth the resources and clusters don't exist long enough in
+    the cloud for this to work currently.
+2. Implement starting the agent in MAPI and/or CAPI
+  - These components are already the starting point for removing a node from a
+    cluster and would likely be the ones triggering the unbind
+  - In this case the assisted-service wouldn't require the spoke cluster
+    kubeconfig as MAPI/CAPI should already have this access.
+  - This would mean that users not using MAPI/CAPI would not have access to this
```
mhrivnak, line 230: This alternative is predicated on these assumptions:
* If you're using an existing MAPI or CAPI provider (aws, gcp, ...), you don't need this feature at all.
* This alternative would be implemented as a MAPI/CAPI provider (maybe as part of capi-agent).
* boot-it-yourself would still be the way to bring hardware under initial management.

Does that clarify things? I think the statement on this line then basically amounts to "users who choose not to use it won't be using it."

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,234 @@
...
+environment.
+
+## Design Details [optional]
+
+### Open Questions
+
+1. How does assisted service differentiate between an agent registering from the
+   discovery live environment from the DaemonSet pod?
+  - Could be a problem if the host reboots into the node OS rather than the
+    discovery boot entry (described in "Risks and Mitigations")
```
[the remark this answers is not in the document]
mhrivnak: Did you look any further into the option of just running next-step-runner?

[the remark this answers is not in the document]
carbonin: It doesn't look like it will be a problem and will probably be the best way around this issue.
I can remove this question.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Yes, this makes sense assuming we have no issues implementing and shipping this feature as a part of MAPI _and_ CAPI.

I think I should probably split this into two different conversations about MAPI and CAPI and focus more on the difficulties of implementing/delivering this as a part of those components.

I'll update this in a bit.

mhrivnak: I'll send you info. Anyone else who is interested, just ping me.

docs/enhancements/auto-return-agent-to-infra-env.md
avishayt: Today's implementation:
1. CAPI cordons the node and then deletes the AgentMachine.
2. The CAPI provider reconciles when the AgentMachine is being deleted. It has a finalizer where it unbinds the agent and removes the finalizer.

Proposed implementation:
1. Upon creating an AgentMachine, the CAPI provider will also add a [lifecycle hook](https://github.com/kubernetes-sigs/cluster-api/blob/1ba14a07c47b87a7f2e3bef00cd592e96bb9e8e4/internal/controllers/machine/machine_controller.go#L305) so that it has a chance to boot the agent.  This means that the CAPI provider will need to watch Machines.
2. When the CAPI provider sees that a Machine is waiting on the PreDrainDeleteHook, it will unbind the agent, causing it to reboot, and remove the PreDrainDeleteHook.
3. There is no need for the AgentMachine finalizer.

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
avishayt: I don't know if I would include this story in the enhancement.  It's a valid story regardless of how the host was re-used.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,229 @@
...
+
+* Automate graceful removal of a node from an existing cluster
+* Booting a node that is unresponsive or not currently part of an installed
+  cluster
+* Registering an agent to an InfraEnv other than the one it is currently
+  associated with
+
+## Proposal
+
+When a host should be returned to the InfraEnv use the kubeconfig created when
```
CrystalChun, line 48: Is there any scenario where the kubeconfig is not valid and the host is unreachable?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,229 @@
...
+1. User unbinds the host from the cluster
+2. Assisted service creates a DaemonSet to run the assisted-installer-agent
+   on the host as a privileged pod
+  * If there is no BareMetalHost associated with the agent CR, the service
+    executes a new host state transition (`TransitionTypeReclaimHost`)
+  * Host moves to state `HostStatusReclaiming`
+3. The agent downloads the InfraEnv's kernel and initrd (including discovery
+   ignition) to the host's `/boot` partition
+4. The agent creates a boot entry to boot into the discovery kernel and initrd
+  * Host moves to state `HostStatusReclaimingPendingReboot`
```
CrystalChun, line 68: Does each host reboot individually or do they all wait for each other to reach this state and reboot all at once?

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
mhrivnak: The important part is that for some use cases, this will be a hard requirement, and it could be a serious breech if data is leaked. "automating returning an agent to the infra-env" for those use cases would have to include wiping data. If we ship this feature without that, it needs to be loud and clear that data is not wiped.

docs/enhancements/auto-return-agent-to-infra-env.md
mhrivnak: If the API server TLS certificate were [replaced](https://docs.openshift.com/container-platform/4.10/security/certificates/api-server.html), that would presumably make the kubeconfig invalid.

carbonin: Yeah I think we definitely will account for the kubeconfig being invalid.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Opened https://issues.redhat.com/browse/MGMT-10968 to handle the CAPI provider changes.

docs/enhancements/auto-return-agent-to-infra-env.md
CrystalChun: Yeah my concern was if they were rebooted separately, would that cause any issues? 

That's a good idea to keep the uncoordinated approach and see if any issues crop up. Thanks for the details!

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,247 @@
...
+  - When using the REST API there is no way for a user to indicate if a host
+    should be unbound normally or if it should be reclaimed
+* There must be enough free space on the `/boot` partition
+  - Space is required to store the initrd and kernel (~90MB)
+  - The agent can check for this
+* A node in a spoke cluster can be identified by the hostname in the agent
+  - In a non-CAPI scenario there are no other reliable means to identify which
+    node maps to which agent
+  - With CAPI a string of references exist that could be used to identify the
+    node, but the hostname should be reliable in this case as well
```
carbonin, line 187: We could also have some tiered logic where we first check for the CAPI machine resources, then check for the system uid then fall back to hostname, but hostname should just work all the time, right? Does anyone know of a reason it wouldn't?

@mhrivnak @avishayt @tsorya @eranco74 ?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,247 @@
...
+A timeout will be implemented when the service expects the host to reregister
+(in `HostStatusReclaimingPendingReboot`). After this time the host will move to
+the existing `HostStatusUnbindingPendingUserAction` state.
+
+If the node hasn't been removed from the cluster when the host reboots the host
+could boot back into the node and the pod running the agent would attempt to
+register. The assisted service would not know the difference between this
+scenario and if the node had booted into discovery. To deal with this, the
+next-step-runner process will be started in the pod rather than the full agent
+as the only difference is the registration step.
```
carbonin, line 204: @mhrivnak I know we talked about this not being an issue yesterday, but I think we were focused on CAPI where the node would have been removed from the cluster.

If we're not using CAPI and the user hasn't removed the node from the cluster and it reboots, it would get its workloads back, right? If so, that would definitely cause a problem.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,247 @@
...
+  - A lot of information will be required to properly start the agent
+    - Agent image pull spec
+    - Hub cluster CA cert
+    - http(s) proxy information
+    - Agent authentication information (pull secret)
+    - Boot artifact URLs
+  - There's not much relevant difference between doing any of this in the CAPI
+    provider vs assisted
+  - Adding a MAPI provider would involve creating our own platform type which
+    is not something we want to take on
```
carbonin, line 247: @mhrivnak @filanov does this about cover what we talked about?

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Does this seem resolved to all of you?
Don't want to close it out if there are still questions.

docs/enhancements/auto-return-agent-to-infra-env.md
CrystalChun: I thought the [first step](https://github.com/openshift/assisted-service/pull/3849/files#diff-064572287176be5e9df4e402623ab95cce4ab3061a918d14093b7d6d891d1317R59) was that the user has to remove the host from the cluster, unless is there an instance where they wouldn't and the agent starts rebooting? Like if there were a timing issue where the service hasn't saved that the host was removed from the cluster, but the host has started/finished rebooting?

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
CrystalChun: Maybe (for the next iteration) there could be an exposed option to the user where they could choose whether or not to wipe the data on certain/all hosts?

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,247 @@
...
+--------------------------------  ----------------------------------------
+| HostStatusDiscoveringUnbound |  | HostStatusUnbindingPendingUserAction |
+--------------------------------  ----------------------------------------
+              ^                                       |
+              |       ------------------------------  |
+              --------| TransitionTypeRegisterHost |---
+                      ------------------------------
+```
+
+#### Assumptions
```
JoelSpeed, line 169: From a conversation with @carbonin last week, to add to this list, we must make sure that the hosts aren't being managed by Machine API/Cluster API within the guest cluster itself. Nick mentioned something about configuring machinesets in the guest cluster for vSphere, if that's happening, then MAPI now thinks it owns the host and reclaiming it could cause issues.
We discussed a way to get around this issue using lifecycle hooks

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,247 @@
...
+* There must be enough free space on the `/boot` partition
+  - Space is required to store the initrd and kernel (~90MB)
+  - The agent can check for this
+* A node in a spoke cluster can be identified by the hostname in the agent
+  - In a non-CAPI scenario there are no other reliable means to identify which
+    node maps to which agent
+  - With CAPI a string of references exist that could be used to identify the
+    node, but the hostname should be reliable in this case as well
+
+### Risks and Mitigations
```
JoelSpeed, line 189: We should add some stuff here with regards to "who owns the Machine", it's not clear to me that in a centrally managed agent install that there isn't going to be machine management within the guest cluster itself, technically a user could, if they have the right credentials for say vSphere, start owning/managing the VMs

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: "unbind" doesn't actually mean "remove node from cluster". And in-fact this can't be the case because if the node was removed from the cluster then after that point we wouldn't be able to start the pod that is going to actually do the work.

CrystalChun: Oh gotcha, thanks for the explanation!

docs/enhancements/auto-return-agent-to-infra-env.md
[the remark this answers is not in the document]
carbonin: Maybe, but we don't really have a way for users to provide us with any additional information right now. Especially when unbinding in the kube API.

I think that something like this might require a bit more design.

docs/enhancements/auto-return-agent-to-infra-env.md
carbonin: Added a paragraph to the end of this section.

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,256 @@
...
+same environment.
+
+### Implementation Details/Notes/Constraints [optional]
+
+* The boot artifact URLs (initrd, kernel, and rootfs) will be provided to the
+  agent directly as a part of the pod spec that runs the agent.
+* The agent will require two new commands; one for each of the new states
+  1. Download boot artifacts and write boot entry (`HostStatusReclaiming`)
+  2. Reboot the node (`HostStatusReclaimingPendingReboot`)
+* The exiting unbind API will be used to trigger this process
```
CrystalChun, line 95: Is this a misspelling? `existing`?

docs/enhancements/auto-return-agent-to-infra-env.md
CrystalChun: Looping back, does that mean that the user would need to unbind, wait for the hosts to move to state `HostStatusReclaimingPendingReboot`, then remove them from the cluster via another API?

Would it be a good idea to introduce another state before this state if it's not CAPI that prompts the user to use the remove host from cluster API? A state like `HostStatusReclaimingPendingUserAction`

docs/enhancements/auto-return-agent-to-infra-env.md
CrystalChun: I could see the argument for it being an `exciting` API 😆

docs/enhancements/auto-return-agent-to-infra-env.md
```diff
@@ -0,0 +1,251 @@
...
+                |                                           |
+       Agent posts step reply                               |
+                |                                           |
+       ---------+----------------------------------         |
+       |                                          |         |
+    Success                                    Failure      |
+       |                                          |         |
+       V                                          |         |
+-------------------------------------             |         |
+| HostStatusReclaimingPendingReboot |------       |         |
```
avishayt, line 147: I think we changed this to ReclaimingRebooting?

carbonin: Yup, let me update.

PR #3870, opened 2022-05-29
MGMT-8568: Use NMState GenerateConfiguration Go binding

nmagnezi:
This change replaces the code nmstate configuration validation from
using a binary named nmstatectl to a dynamic nmstate Go binding.

Using the Go binding, the code makes use of the GenerateConfiguration function
to validate nmstate configurations, which is a lot faster and consumes
less resources compared to a use of an external binary.

In addition, changes to how we compile the code were made to support
the above-mentioned Go binding.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


nmagnezi:
will resume work in this as soon as https://github.com/nmstate/nmstate/issues/1914 gets fixed.

hack/setup_env.sh
```diff
@@ -60,7 +60,7 @@ function assisted_service() {
     rm -f /tmp/kubectl
 
   yum install -y --setopt=skip_missing_names_on_install=False \
-    docker podman awscli python3-pip genisoimage skopeo
```
nmagnezi, line 63: I'm checking where to get `awscli` from. it is not available in centos stream 9

filanov:
@oshercc first need to change the container in the CI?

osherdp:
> @osherdp first need to change the container in the CI?

what container?

nmagnezi:
Code should be compiled on CentOS 9 stream.
For runtime, it also needs an up to date version of `nmstate-libs` available on CentOS 9 stream, otherwise it fails with:
`/assisted-service: error while loading shared libraries: libnmstate.so.2.1.0: cannot open shared object file: No such file or directory`

Another issue found today: https://github.com/nmstate/nmstate/issues/1970

Dockerfile.assisted-service
```diff
@@ -34,15 +34,15 @@ RUN go mod download
...
+RUN CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /build/assisted-service-admission cmd/webadmission/main.go
+RUN CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /build/agent-installer-client cmd/agentbasedinstaller/client/main.go
 
 # Create final image
-FROM quay.io/centos/centos:stream8
+FROM quay.io/centos/centos:stream9
 
 # openshift-install requires this
-RUN dnf install -y libvirt-libs nmstate &&\
+RUN dnf install -y libvirt-libs nmstate nmstate-libs && \
```
nmagnezi, line 45: TODO: check if we can avoid installing `nmstate`

nmagnezi:
/unhold
ready for reviews

pkg/staticnetworkconfig/generator.go
```diff
@@ -90,40 +92,16 @@ func (s *StaticNetworkConfigGenerator) generateHostStaticNetworkConfigData(ctx c
 	return filesList, nil
 }
 
-func (s *StaticNetworkConfigGenerator) executeNMStatectl(ctx context.Context, hostYAML string) (string, error) {
-	err := s.sem.Acquire(ctx, 1)
+func (s *StaticNetworkConfigGenerator) validateNMStateConfig(hostYAML string) (string, error) {
```
carbonin, line 95: This is not just for validating the config it also actually generates the resulting config files.

I think we should name it better or maybe, because it's so small, we just move the contents to where it's used.
It's only two calls and it is only used in two places.

pkg/staticnetworkconfig/generator.go
```diff
@@ -90,40 +92,16 @@ func (s *StaticNetworkConfigGenerator) generateHostStaticNetworkConfigData(ctx c
...
+	hostJSON, err := yamlconvertor.YAMLToJSON([]byte(hostYAML))
 	if err != nil {
-		s.log.WithError(err).Errorf("Failed to lock semaphore for nmstatectl execution")
-		return "", err
+		fmt.Printf("err: %v\n", err)
```
carbonin, line 98: I think you should handle this error.

Also remove debugging printf statements.

Dockerfile.assisted-service
```diff
@@ -16,7 +16,7 @@ RUN cd /assisted-service/build && python3 ../tools/client_package_initializer.py
 # Build binaries
 FROM quay.io/centos/centos:stream9 as builder
 
-RUN dnf install -y gcc && dnf clean all
+RUN dnf install --enablerepo=crb -y gcc nmstate-devel && dnf clean all
```
celebdor, line 19: why not use the nmstate-static so you don't need to have nmstate.so at runtime?

nmagnezi: I tried to make it work with `nmstate-static`, with no success.
It kept on failing with:
`/assisted-service: error while loading shared libraries: libnmstate.so.2.1.0: cannot open shared object file: No such file or directory`

Since the dynamic Go binding work, I proceeded with that.

pkg/staticnetworkconfig/generator.go
```diff
@@ -33,20 +33,22 @@ type StaticNetworkConfigData struct {
...
-	sem *semaphore.Weighted
+	log     logrus.FieldLogger
+	nmstate Nmstate
 }
 
-func New(log logrus.FieldLogger, cfg Config) StaticNetworkConfig {
+func New(log logrus.FieldLogger, nm *nmstate.Nmstate) StaticNetworkConfig {
+	if nm == nil {
+		nm = nmstate.New()
+	}
```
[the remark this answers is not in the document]
nmagnezi: Done

pkg/staticnetworkconfig/generator.go
nmagnezi: Removed.
returning to error, so it is handled by the calling function

pkg/staticnetworkconfig/generator.go
nmagnezi: I think there's enough here for it to remain as a function. I changed the function name to  better reflect what it is doing.

carbonin (approved):

PR #3880, opened 2022-05-31
MGMT-9175: Use an interface with a Mac address

CrystalChun:
[MGMT-9175](https://issues.redhat.com/browse/MGMT-9175)
Part of epic https://issues.redhat.com/browse/MGMT-10087
The assisted installer agent has changes to
pass all interfaces to the assisted service.
PR for the agent changes:
https://github.com/openshift/assisted-installer-agent/pull/360
Originally, the assisted service selected the first interface
in the list, which might not have a MAC address.
This change now looks for an interface with a MAC
address and uses it for the install config.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @ori-amizur 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/provider/baremetal/installConfig.go
+++ b/internal/provider/baremetal/installConfig.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/assisted-service/internal/host/hostutil"
 	"github.com/openshift/assisted-service/internal/installcfg"
 	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
 )
 
 func (p baremetalProvider) AddPlatformToInstallConfig(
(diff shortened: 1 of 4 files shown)


internal/provider/baremetal/installConfig.go
```diff
@@ -40,7 +41,18 @@ func (p baremetalProvider) AddPlatformToInstallConfig(
 			p.Log.Warnf("Failed to unmarshall Host %s inventory", hostutil.GetHostnameForMsg(host))
 			return err
 		}
-		hosts[yamlHostIdx].BootMACAddress = inventory.Interfaces[0].MacAddress
+
+		for _, iface := range inventory.Interfaces {
+			if iface.MacAddress != "" {
```
filanov, line 46: Maybe you could check the type? empty string or physical to be on the safe side

internal/provider/baremetal/installConfig.go
```diff
@@ -40,7 +41,18 @@ func (p baremetalProvider) AddPlatformToInstallConfig(
...
+				hosts[yamlHostIdx].BootMACAddress = iface.MacAddress
+			}
+		}
+		if hosts[yamlHostIdx].BootMACAddress == "" {
```
filanov, line 50: I'm not sure if this can happen, after all the agent was able to connect to the service somehow no?

filanov (commented):
Consider adding a TODO in this code so you will change it later according to the env var for the deployment

internal/provider/baremetal/installConfig.go
carbonin: I suppose if you look for empty string it would work, but the problem we're trying to solve with this PR is that we can't get the agent PR that sends the type data to pass e2e tests without a fix here first.

Plus I don't think we'd want to filter here. We should eventually be able to use whatever interface the user chooses, even if it's virtual, right? That said, this wasn't driven by a user selection before, which is why I asked why it was okay to just use the first in the list.

internal/provider/baremetal/installConfig.go
carbonin: Reading the code it looks like something we should check for even if we don't expect it to happen in reality. I'm fine with it.

internal/provider/baremetal/installConfig.go
CrystalChun: I think we'd want to revisit this code when adding the environment variable to enable virtual interfaces ([MGMT-10411](https://issues.redhat.com/browse/MGMT-10411))

filanov: i want to keep the same behavior as we have now, the empty string is for the type and it's targeting old agents. i think that the code looks good now.

internal/provider/baremetal/installConfig.go
filanov: sounds reasonable

PR #3904, opened 2022-06-06
MGMT-9561: VSphere platform lacking setting disk.EnableUUID guidance

michaellevy101:
The parameter disk.EnableUUID is necessary so that the VMDK always presents a consistent UUID to the VM, thus allowing the disk to be mounted properly.
Please read https://access.redhat.com/solutions/4606201

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/hardware/validator.go
+++ b/internal/hardware/validator.go
@@ -35,6 +35,7 @@ type Validator interface {
 	GetInfraEnvHostRequirements(ctx context.Context, infraEnv *common.InfraEnv) (*models.ClusterHostRequirements, error)
 	DiskIsEligible(ctx context.Context, disk *models.Disk, infraEnv *common.InfraEnv, cluster *common.Cluster, host *models.Host, allDisks []*models.Disk) ([]string, error)
 	ListEligibleDisks(inventory *models.Inventory) []*models.Disk
+	IsValidStorageDeviceType(disk *models.Disk) bool
 	GetInstallationDiskSpeedThresholdMs(ctx context.Context, cluster *common.Cluster, host *models.Host) (int64, error)
 	// GetPreflightHardwareRequirements provides hardware (host) requirements that can be calculated only using cluster information.
 	// Returned information describe requirements coming from OCP and OLM operators.
@@ -115,10 +116,9 @@ func (v *validator) DiskIsEligible(ctx context.Context, disk *models.Disk, infra
 				humanize.Bytes(uint64(disk.SizeBytes)), humanize.Bytes(uint64(minSizeBytes))))
 	}
 
-	allowedDriveTypes := []string{string(models.DriveTypeHDD), string(models.DriveTypeSSD), string(models.DriveTypeMultipath)}
-	if !funk.ContainsString(allowedDriveTypes, string(disk.DriveType)) {
+	if !v.IsValidStorageDeviceType(disk) {
 		notEligibleReasons = append(notEligibleReasons,
-			fmt.Sprintf(wrongDriveTypeTemplate, disk.DriveType, strings.Join(allowedDriveTypes, ", ")))
+			fmt.Sprintf(wrongDriveTypeTemplate, disk.DriveType, strings.Join(v.getValidDeviceStorageTypes(), ", ")))
 	}
 
 	// We only allow multipath if all paths are FC
@@ -135,6 +135,10 @@ func (v *validator) DiskIsEligible(ctx context.Context, disk *models.Disk, infra
 	}
 
 	return notEligibleReasons, nil
+}
+
+func (v *validator) IsValidStorageDeviceType(disk *models.Disk) bool {
+	return funk.ContainsString(v.getValidDeviceStorageTypes(), string(disk.DriveType))
 }
 
 func (v *validator) purgeServiceReasons(reasons []string) []string {
@@ -377,6 +381,10 @@ func (v *validator) getMasterRequirements(cluster *common.Cluster, requirements 
 
 func (v *validator) getOCPRequirementsForVersion(openshiftVersion string) (*models.VersionedHostRequirements, error) {
 	return v.VersionedRequirements.GetVersionedHostRequirements(openshiftVersion)
+}
+
+func (v *validator) getValidDeviceStorageTypes() []string {
+	return []string{string(models.DriveTypeHDD), string(models.DriveTypeSSD), string(models.DriveTypeMultipath)}
 }
 
 func compileDiskReasonTemplate(template string, wildcards ...interface{}) *regexp.Regexp {
--- a/internal/host/validations_test.go
+++ b/internal/host/validations_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/openshift/assisted-service/internal/operators/api"
 	"github.com/openshift/assisted-service/internal/provider/registry"
 	"github.com/openshift/assisted-service/models"
+	"github.com/openshift/assisted-service/pkg/conversions"
 	"gorm.io/gorm"
 )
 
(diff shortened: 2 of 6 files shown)


internal/host/validator.go
```diff
@@ -1427,3 +1427,36 @@ func (v *validator) printNonOverlappingSubnets(c *validationContext, status Vali
...
+		return !v.hwValidator.IsHDDevice(disk) || disk.HasUUID
+	})
+
+	return boolValue(allHDValid)
+}
+
+func (v *validator) printVSphereUUIDEnabled(context *validationContext, status ValidationStatus) string {
+	switch status {
+	case ValidationSuccess:
+		return "VSphere disk.UUID is enabled for this host"
```
romfreiman, line 1454: host? vm?
What do ocp docs saying? Lets use same terminology

romfreiman: @michaellevy101

romfreiman:
Lets align on the return message please

omertuc:
This will have to wait for the new agent to be in prod for a few weeks, or for the agent upgrade feature, otherwise we're gonna block vsphere installation that are using old agents

omertuc:
This could've been solved if the has uuid boolean was a pointer, but it's not, so we're stuck

internal/hardware/validator.go
```diff
@@ -378,6 +384,10 @@ func (v *validator) getOCPRequirementsForVersion(openshiftVersion string) (*mode
 	return v.VersionedRequirements.GetVersionedHostRequirements(openshiftVersion)
 }
 
+func (v *validator) getHdDeviceTypes() []string {
```
omertuc, line 387: What is "Hd"?

internal/host/validator.go
```diff
@@ -1432,3 +1432,40 @@ func (v *validator) printNonOverlappingSubnets(c *validationContext, status Vali
...
+	if c.cluster.Platform == nil || c.cluster.Platform.Type == nil || *c.cluster.Platform.Type != models.PlatformTypeVsphere {
+		return ValidationSuccess
+	}
+
+	if c.inventory.Disks == nil {
+		return ValidationPending
+	}
+
+	for _, disk := range c.inventory.Disks {
+		if v.hwValidator.IsHDDevice(disk) && !disk.HasUUID {
```
omertuc, line 1454: Why do we only look at disks that are "HD" (whatever that means, I'm not sure yet)?

Also you should probably ignore all non-eligible disks in this loop, and you'll get this "HD" check for free

internal/host/validator.go
```diff
@@ -1432,3 +1432,40 @@ func (v *validator) printNonOverlappingSubnets(c *validationContext, status Vali
...
+
+	return ValidationSuccess
+}
+
+func (v *validator) printVSphereUUIDEnabled(context *validationContext, status ValidationStatus) string {
+	switch status {
+	case ValidationSuccess:
+		return "VSphere disk.UUID is enabled for this host"
+	case ValidationFailure:
+		return "VSphere disk.UUID isn't enabled for this vm, it's necessary for disks to be mounted properly"
```
omertuc, line 1467: In the commit message you call it disk.EnableUUID but here it's disk.UUID? which one is it?

internal/hardware/validator.go
michaellevy101: installable drive like Hard Drive

michaellevy101: not only installation but a drive that can be use for storage for example CDROM isn't suite for that

omertuc: Hard drive means HDD but this function returns also SSD as type

omertuc: > that can be use for storage

Please explain this in the function documentation, and pick a name that reflects that

internal/host/validator.go
michaellevy101: eligibility has more criterias because it used for installation but here i check not only installation device but storage devices.

internal/host/validator.go
michaellevy101: disk.EnableUUID is the vsphere property name set in the vm level

michaellevy101: but even if it set for a VM its disks might not have a UUID depend on the disk type so the check should be for every disk

internal/host/validator.go
omertuc: Yeah I could guess, but please explain this in the commit message / code / documentation - it's not obvious

michaellevy101:
@gamli75 https://github.com/openshift/assisted-service/pull/3904#issuecomment-1150343228

internal/host/validator.go
michaellevy101: misunderstood you concern. It's the same as @romfreiman's

omertuc: Mine was more about host vs disk
Rom seems to be about "Host" vs "VM"

internal/host/validator.go
```diff
@@ -1432,3 +1432,40 @@ func (v *validator) printNonOverlappingSubnets(c *validationContext, status Vali
 	}
 	return fmt.Sprintf("Unexpected status %s", status)
 }
+
+func (v *validator) isVSphereDiskUIDEnabled(c *validationContext) ValidationStatus {
```
omertuc, line 1436:
```suggestion
func (v *validator) isVSphereDiskUUIDEnabled(c *validationContext) ValidationStatus {
```

internal/host/validator.go
```diff
@@ -1432,3 +1432,40 @@ func (v *validator) printNonOverlappingSubnets(c *validationContext, status Vali
...
+	if c.cluster.Platform == nil || c.cluster.Platform.Type == nil || *c.cluster.Platform.Type != models.PlatformTypeVsphere {
+		return ValidationSuccess
+	}
+
+	if c.inventory.Disks == nil {
+		return ValidationPending
+	}
+
+	for _, disk := range c.inventory.Disks {
+		if v.hwValidator.IsValidStorageDeviceType(disk) && !disk.HasUUID {
```
omertuc, line 1454:
```suggestion
                // vSphere only adds a UUID to disks which can potentially be used for storage,
                // if any of them doesn't have that flag, it's likely because the user has forgot to
                // enable `disk.EnableUUID` for this virtual machine
                // See https://access.redhat.com/solutions/4606201
		if v.hwValidator.IsValidStorageDeviceType(disk) && !disk.HasUUID {
```
This is not so obvious

gamli75:
@michaellevy101 please change the title and add the ticket number in Jira

gamli75:
> This will have to wait for the new agent to be in prod for a few weeks, or for the agent upgrade feature, otherwise we're gonna block vsphere installations that are using old agents, as we have no way to tell apart "no UUID" from "this is an old agent that doesn't know how to report "has UUID" yet"

@omertuc we can merge this code and disable this validation on production but I want QE to test it on staging.

internal/host/validations_test.go
```diff
@@ -565,4 +567,110 @@ var _ = Describe("Validations test", func() {
...
+				HasUUID:   false,
+			}
+
+			HDD := &models.Disk{
+				SizeBytes: conversions.GibToBytes(120),
+				DriveType: "HDD",
+				HasUUID:   true,
+			}
+
+			updateHostInventory(func(inventory *models.Inventory) {
```
gamli75, line 662: please also add a test with a disk that has an invalid storage type

internal/hardware/validator.go
```diff
@@ -24,7 +24,7 @@ import (
 const (
 	tooSmallDiskTemplate       = "Disk is too small (disk only has %s, but %s are required)"
 	wrongDriveTypeTemplate     = "Drive type is %s, it must be one of %s."
-	wrongMultipathTypeTemplate = "Multipath device has path of type %s, it must be %s"
+	wrongMultipathTypeTemplate = "Multipath device has path of type %s, it must be one of %s"
```
omertuc, line 27: Bad rebase?

PR #3905, opened 2022-06-07
MGMT-10411: Allow enabling virtual interfaces

CrystalChun:
[MGMT-10411](https://issues.redhat.com/browse/MGMT-10411)
As part of epic https://issues.redhat.com/browse/MGMT-10087
Exposes an environment variable in the service
that allows virtual interfaces to be
used.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [x] assisted-test-infra environment: deployed and installed nodes after using this service image and adding `ENABLE_VIRTUAL_INTERFACES` to the configmap
- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @filanov 
/cc @ori-amizur 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/test_configuration.go
+++ b/internal/common/test_configuration.go
@@ -239,6 +239,41 @@ func GenerateTestDefaultInventory() string {
 	return string(b)
 }
 
+func GenerateTestInventoryWithVirtualInterface() string {
+	inventory := &models.Inventory{
+		Interfaces: []*models.Interface{
+			{
+				Name: "eth0",
+				IPV4Addresses: []string{
+					"192.0.2.0/24",
+				},
+				IPV6Addresses: []string{
+					"2001:db8::/32",
+				},
+				Type: "physical",
+			},
+			{
+				Name: "cni1",
+				IPV4Addresses: []string{
+					"198.51.100.0/24",
+				},
+				IPV6Addresses: []string{
+					"2001:db8::/32",
+				},
+				Type: "device",
+			},
+		},
+		Disks: []*models.Disk{
+			TestDefaultConfig.Disks,
+		},
+		Routes: TestDefaultRouteConfiguration,
+	}
+
+	b, err := json.Marshal(inventory)
+	Expect(err).To(Not(HaveOccurred()))
+	return string(b)
+}
+
 func GenerateTest2IPv4AddressesInventory() string {
 	inventory := &models.Inventory{
 		Interfaces: []*models.Interface{
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -81,6 +81,7 @@ type Config struct {
 	DisabledHostvalidations  DisabledHostValidations `envconfig:"DISABLED_HOST_VALIDATIONS" default:""` // Which host validations to disable (should not run in preprocess)
 	BootstrapHostMAC         string                  `envconfig:"BOOTSTRAP_HOST_MAC" default:""`        // For ephemeral installer to ensure the bootstrap for the (single) cluster lands on the same host as assisted-service
 	MaxHostDisconnectionTime time.Duration           `envconfig:"HOST_MAX_DISCONNECTION_TIME" default:"3m"`
+	EnableVirtualInterfaces  bool                    `envconfig:"ENABLE_VIRTUAL_INTERFACES" default:"false"`
 }
 
 //go:generate mockgen --build_flags=--mod=mod -package=host -aux_files=github.com/openshift/assisted-service/internal/host/hostcommands=instruction_manager.go -destination=mock_host_api.go . API
@@ -349,6 +350,19 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster, 
 		log.WithError(err).Errorf("not updating inventory - failed to check disks eligibility for host %s", h.ID)
 		return common.NewApiError(http.StatusInternalServerError, err)
 	}
+
+	// Remove virtual interfaces if the feature is not enabled
+	if !m.Config.EnableVirtualInterfaces {
+		var physicalInterfaces []*models.Interface = make([]*models.Interface, 0)
+		for _, intf := range inventory.Interfaces {
+			if intf.Type == "" || intf.Type == "physical" || intf.Type == "vlan" || intf.Type == "bond" {
+				// Empty interface type indicates an older agent, which only passes physical interfaces
+				physicalInterfaces = append(physicalInterfaces, intf)
+			}
+		}
+		inventory.Interfaces = physicalInterfaces
+	}
+
 	m.populateDisksId(inventory)
 	inventoryStr, err = common.MarshalInventory(inventory)
 	if err != nil {
--- a/internal/host/host_test.go
+++ b/internal/host/host_test.go
@@ -1323,6 +1323,84 @@ var _ = Describe("UpdateInventory", func() {
 			Expect(h.InstallationDiskID).To(Equal(diskId))
 		})
 	})
+	Context("Interfaces", func() {
+		const (
+			diskName = "FirstDisk"
+			diskId   = "/dev/disk/by-id/FirstDisk"
+		)
+		var (
+			dummy = &leader.DummyElector{}
+		)
+
+		BeforeEach(func() {
+			host = hostutil.GenerateTestHost(hostId, infraEnvId, "", models.HostStatusDiscovering)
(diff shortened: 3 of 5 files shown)


carbonin:
Is this allowing the interfaces to be used or just indicating that the agent should collect and send them?

CrystalChun:
> Is this allowing the interfaces to be used or just indicating that the agent should collect and send them?

It's intended for both, should they be separate?

cmd/main.go
```diff
@@ -146,6 +146,7 @@ var Options struct {
 	HTTPListenPort                 string        `envconfig:"HTTP_LISTEN_PORT" default:""`
 	AllowConvergedFlow             bool          `envconfig:"ALLOW_COVERGED_FLOW" default:"true"`
 	IronicIgnitionBuilderConfig    ignition.IronicIgniotionBuilderConfig
+	EnableVirtualInterfaces        bool `envconfig:"ENABLE_VIRTUAL_INTERFACES" default:"false"`
```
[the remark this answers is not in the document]
CrystalChun: There was [discussion in the epic](https://issues.redhat.com/browse/MGMT-10087?focusedCommentId=20333260&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20333260) that it should be globally enabled/disabled as opposed to at the cluster/infra-env level.

internal/host/hostcommands/inventory_cmd.go
```diff
@@ -26,6 +29,9 @@ func (h *inventoryCmd) GetSteps(ctx context.Context, host *models.Host) ([]*mode
 			host.ID.String(),
 		},
 	}
+	if h.enableVirtualInterfaces {
```
filanov, line 32: why do we need to send it to the agent? 
I was thinking that the agent will send everything and the service will decide if it's need to be used or not.

CrystalChun: For backwards compatibility - if the agent is newer than the service and sends all the interfaces, the service will use the first interface (hard-coded) which is not always a useable one. 

Therefore we needed a way for the new agent to tell when the service is new enough to handle all the interfaces being sent.

Slack discussion for context: https://coreos.slack.com/archives/C014N2VLTQE/p1654481131218899

filanov: i suspect that this is not backward compatible, we can have a case of new service old agent not the other way around

internal/host/host.go
```diff
@@ -349,6 +350,25 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
 		log.WithError(err).Errorf("not updating inventory - failed to check disks eligibility for host %s", h.ID)
 		return common.NewApiError(http.StatusInternalServerError, err)
 	}
+
+	// MGMT-10087: All interfaces will be sent to the service from the agent
```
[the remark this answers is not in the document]
CrystalChun: Gotcha, removed it, thank you!

internal/common/test_configuration.go
```diff
@@ -239,6 +239,41 @@ func GenerateTestDefaultInventory() string {
 	return string(b)
 }
 
+func GenerateTestInventoryWithVirtualInterface() string {
+	inventory := &models.Inventory{
+		Interfaces: []*models.Interface{
+			{
+				Name: "eth0",
+				IPV4Addresses: []string{
+					"1.2.3.4/24",
```
carbonin, line 248: nit, but you could use addresses from the subnets specifically designated for use in tests and documentation here.

They are `192.0.2.0/24`, `198.51.100.0/24`, and `203.0.113.0/24`
https://en.wikipedia.org/wiki/Reserved_IP_addresses

internal/host/host_test.go
```diff
@@ -1323,6 +1323,72 @@ var _ = Describe("UpdateInventory", func() {
...
+				nil, createValidatorCfg(), nil, defaultConfig, dummy, nil, nil)
+			Expect(enabled_hapi.UpdateInventory(ctx, &host, host.Inventory)).ToNot(HaveOccurred())
+
+			h := hostutil.GetHostFromDB(hostId, infraEnvId, db)
+			Expect(h.Inventory).ToNot(BeEmpty())
+			inventory, err := common.UnmarshalInventory(h.Inventory)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(inventory).To(Not(BeNil()))
+			Expect(inventory.Interfaces).Should(HaveLen(2))
+			Expect(inventory.Interfaces[0].Type).To(Equal(physicalInterface.Type))
```
carbonin, line 1388: Can we rely on the ordering here? Might cause a flake if we can't. Maybe it's better to range through them and ensure you find the virtual interface?

internal/host/host_test.go
```diff
@@ -1323,6 +1323,72 @@ var _ = Describe("UpdateInventory", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(inventory).To(Not(BeNil()))
+			Expect(inventory.Interfaces).Should(HaveLen(1))
+			Expect(inventory.Interfaces[0].Type).To(Equal(physicalInterface.Type))
+		})
+		It("Saves virtual interfaces when virtual interface flag is enabled", func() {
+			defaultConfig.EnableVirtualInterfaces = true
+			enabled_hapi := NewManager(common.GetTestLog(), db, mockEvents, mockValidator,
+				nil, createValidatorCfg(), nil, defaultConfig, dummy, nil, nil)
+			Expect(enabled_hapi.UpdateInventory(ctx, &host, host.Inventory)).ToNot(HaveOccurred())
```
carbonin, line 1380: nit, but when I see `ToNot(HaveOccurred())` I always think we're expecting a method not to be called.

Personally I prefer `Expect(enabled_hapi.UpdateInventory(ctx, &host, host.Inventory)).To(Succeed())` in cases where we're asserting on the error response of a function that returns just an error value.

internal/host/host.go
```diff
@@ -349,6 +350,22 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
...
 		return common.NewApiError(http.StatusInternalServerError, err)
 	}
+
+	// Remove virtual interfaces if the feature is not enabled
+	if !m.Config.EnableVirtualInterfaces {
+		var physicalInterfaces []*models.Interface = make([]*models.Interface, 0)
+		for _, intf := range inventory.Interfaces {
+			if intf.Type == "" {
+				// Indicates an older agent, so all interfaces are non-virtual
+				break
```
carbonin, line 360: If we break the loop here, when the (presumably) first interface doesn't have a type, won't we set `inventory.Interfaces` to the empty `physicalInterfaces` slice?

I think it would be better if you just treat the empty case the same as the other physical ones below and add it.

Also can you add a test case for an empty type?

internal/host/host_test.go
```diff
@@ -1323,6 +1323,72 @@ var _ = Describe("UpdateInventory", func() {
 			Expect(h.InstallationDiskID).To(Equal(diskId))
 		})
 	})
+	Context("Interfaces", func() {
+		const (
+			diskName = "FirstDisk"
+			diskId   = "/dev/disk/by-id/FirstDisk"
+		)
+		var (
+			physicalInterface = models.Interface{
```
carbonin, line 1332: This doesn't look like it is used anywhere other than where we assert on the type.
Couldn't we just hardcode "physical" in the test like we're doing everywhere else? Why define the whole structure?

internal/host/host_test.go
CrystalChun: Gotcha, modified it, thank you!

internal/host/host_test.go
CrystalChun: True, modified it to loop through the interfaces instead. Thank you!

internal/host/host_test.go
CrystalChun: Removed it, thank you!

internal/host/host.go
CrystalChun: That's right, removed this break and added tests for the empty type. Thank you!

internal/common/test_configuration.go
CrystalChun: Modified it! Thank you!

carbonin (approved):

PR #3914, opened 2022-06-07
MGMT-10569: Remove cluster ignition_config_overrides column

carbonin:
Remove the ignition_config_overrides column from the database and remove the corresponding attribute from the model.

Automigrate doesn't remove columns so we need to do this manually
It's required to `Exec` here because the gormigrate `DropColumn` function requires that the model contain the field being dropped, but that is also being removed with this change. Additionally it doesn't support the `IF EXISTS` portion so it would require more logic to handle this migration correctly.

The migration that populates an infraEnv for each cluster is also removed as a part of this change.
It should not be needed in any environment at this point and keeping it would mean trying to maintain this migration without using the cluster model and that's not worth the effort when this is never expected to be needed again.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

1. Deployed latest `master` branch locally to minikube.
2. Built this branch and deployed the resulting image.
3. Verified that the column was correctly removed.

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @avishayt 
/cc @filanov 

## Checklist

## Reviewers Checklist


avishayt:
Looks good, thanks!  Please fix lint.

carbonin:
CAPI job is failing everywhere it seems.

carbonin:
/unhold

Nevermind, didn't realize it wasn't a required test.

PR #3925, opened 2022-06-10
MGMT-9259: Enhance the schedulable masters feature (simplified)

mmartinv:
The current API always returns the value of the 'schedulable_masters'
property saved in the database at cluster creation time (defaults to
"false" if omitted). However the UI needs the API to return the default
correct value depending on the number of hosts in the cluster.

The proposed changes return the expected default values in case the
'schedulable_masters' property was not actively set.

- [x] New Feature
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @avishayt 

## Checklist

## Reviewers Checklist

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -4286,10 +4286,6 @@ definitions:
         description: "The desired network type used."
         enum: ['OpenShiftSDN', 'OVNKubernetes']
         default: 'OpenShiftSDN'
-      schedulable_masters:
-        type: boolean
-        description: Schedule workloads on masters
-        default: false
       cluster_networks:
         type: array
         description: Cluster networks that are associated with this cluster.
@@ -4463,7 +4459,11 @@ definitions:
       schedulable_masters:
         type: boolean
         description: Schedule workloads on masters
-        default: false
+        default: true
+      use_scheduling_defaults:
+        type: boolean
+        description: False if the scheduling of workloads on masters has been set by the user through the API.
+        default: true
       cluster_networks:
         type: array
         description: Cluster networks that are associated with this cluster.
@@ -4674,7 +4674,11 @@ definitions:
       schedulable_masters:
         type: boolean
         description: Schedule workloads on masters
-        default: false
+        default: true
+      use_scheduling_defaults:
+        type: boolean
+        description: False if the scheduling of workloads on masters has been set by the user through the API.
+        default: true
       updated_at:
         type: string
         format: date-time
--- a/api/vendor/github.com/openshift/assisted-service/models/cluster.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/cluster.go
@@ -231,6 +231,9 @@ type Cluster struct {
 	// Format: date-time
 	UpdatedAt timeext.Time `json:"updated_at,omitempty" gorm:"type:timestamp with time zone"`
 
+	// False if the scheduling of workloads on masters has been set by the user through the API.
+	UseSchedulingDefaults *bool `json:"use_scheduling_defaults,omitempty"`
+
 	// Indicate if the networking is managed by the user.
 	UserManagedNetworking *bool `json:"user_managed_networking,omitempty"`
 
--- a/api/vendor/github.com/openshift/assisted-service/models/cluster_create_params.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/cluster_create_params.go
@@ -109,9 +109,6 @@ type ClusterCreateParams struct {
 	// Required: true
 	PullSecret *string `json:"pull_secret"`
 
-	// Schedule workloads on masters
-	SchedulableMasters *bool `json:"schedulable_masters,omitempty"`
-
 	// The IP address pool to use for service IP addresses. You can enter only one IP address pool. If you need to access the services from an external network, configure load balancers and routers to manage the traffic.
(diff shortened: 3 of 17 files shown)


swagger.yaml
```diff
@@ -4463,7 +4459,11 @@ definitions:
       schedulable_masters:
         type: boolean
         description: Schedule workloads on masters
-        default: false
+        default: true
+      use_scheduling_defaults:
+        type: boolean
+        description: False if the scheduling of workloads on masters has been set by the user through the API.
```
carbonin, line 4465: When would a user set this?

Based on the description it sounds like this value is not something that is provided by the user, but then why is it in the update API?

swagger.yaml
```diff
@@ -4463,7 +4459,11 @@ definitions:
       schedulable_masters:
```
carbonin, line 4459: This was removed from create, but left in update?
Seems strange.

carbonin (changes requested):
I'm going to start with just an API review because I think there are some problems that need to be addressed with the changes to the swagger before it's worth looking at the code changes.

carbonin:
> The current API always returns the value of the 'schedulable_masters'
property saved in the database at cluster creation time (defaults to
"false" if omitted).

So is the actual issue here that we're not updating the value in the database even when we're installing a compact (3-node) cluster? It feels like we should address that rather than changing APIs.

carbonin:
Also, given that some form of this patch was merged and reverted previously I feel like I'm missing some context.
Can you add some more description to the PR about why you're solving this issue in this way and how this PR addresses the issues with the previous approach?

gamli75:
@avishayt can you please review the API changes?

swagger.yaml
mmartinv: The user must set `use_scheduling_defaults` to `True` to return to the expected defaults which are calculated depending on the number of the hosts in the cluster.

swagger.yaml
```diff
@@ -4286,10 +4286,6 @@ definitions:
         description: "The desired network type used."
         enum: ['OpenShiftSDN', 'OVNKubernetes']
         default: 'OpenShiftSDN'
-      schedulable_masters:
```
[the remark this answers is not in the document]
mmartinv: Yes, this is because we don't allow the user to change the scheduling of masters when the number of hosts in the cluster is less than 5. At cluster creation time the number of hosts are 0 and the mandatory value for the `schedulable_masters` property is `true`

swagger.yaml
mmartinv: The user will be able to change the scheduling on master's side (update the cluster object) when the number of hosts is 5 or more.

mmartinv:
> > The current API always returns the value of the 'schedulable_masters'
> > property saved in the database at cluster creation time (defaults to
> > "false" if omitted).
> 
> So is the actual issue here that we're not updating the value in the database even when we're installing a compact (3-node) cluster? It feels like we should address that rather than changing APIs.

Yes, the current implementation does not update the database when a host is added/removed to the cluster.

swagger.yaml
carbonin: This comment doesn't make sense to me.
How can something be a "default" if it's "calculated depending on the number of hosts"?

swagger.yaml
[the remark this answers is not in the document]
carbonin: Also I don't see anything in the code base that indicates that we can't set `schedulable_masters` to any value at cluster creation time.

I don't see any reason to remove this.

gamli75:
this PR should fix the following:
[MGMT-8089](https://issues.redhat.com/browse/MGMT-8089) schedulableMasters field in clusters api returns wrong default values
[MGMT-10591](https://issues.redhat.com/browse/MGMT-10591) 3 masters cluster can start install with scheduled_masters=false
and cover the tests cases that are describe in the doc: https://docs.google.com/document/d/1UCCgxREudnV9frZvON7ZMUG1dy5nBfZmTgrMgleU0U4/edit

swagger.yaml
mmartinv: With "expected defaults" I mean the expected values which are set by the _openshift-install_ command depending on the number of hosts. For less than 5 hosts or more the default expected value would be `true` and `false` otherwise.

avishayt:
IIRC installing 3 masters will cause openshift-installer to set them to be schedulable automatically.  It's just that our API will return misleading information, right?  I think we should have a meeting to discuss this PR.

carbonin:
As far as I can tell this value is only set by the user currently. I don't see anywhere that the backend is enforcing this for SNO or compact clusters.

Is this patch trying to automate some of the logic around this field? It would be good to describe the intended behavior in the PR description if that is indeed what you're doing.

If we're trying to differentiate between when the field is set by a user vs when we set a default I'd think removing [this logic](https://github.com/openshift/assisted-service/blob/9d50fb89dd33b119f6242c2b08c5235b36909842/internal/bminventory/inventory.go#L298-L300) and then inferring the value at install time would be enough. Why change the API?

carbonin:
> I think we should have a meeting to discuss this PR.

Agree

carbonin:
We (@avishayt @gamli75 @lalon4 @osherdp @mmartinv @carbonin) had a discussion and I summarized the result in https://issues.redhat.com/browse/MGMT-8089?focusedCommentId=20438991&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-20438991

PR #3928, opened 2022-06-13
MGMT-10832: Add document for agent upgrade feature

jhernand:
This patch adds a document describing the feature that will allow
upgrading agents without rebooting the hosts.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

To be done.

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/agent-upgrade.md
+++ b/docs/enhancements/agent-upgrade.md
@@ -1,0 +1,299 @@
+---
+title: agent-upgrade
+authors:
+  - "@jhernand"
+creation-date: 2022-06-13
+last-updated: 2022-06-13
+---
+
+# Agent upgrade
+
+## Summary
+
+Usually the time since the agent is started and the cluster is installed is
+short. But in some situations it may be long, even many days, and during that
+long time the service may be upgraded and may require a newer version of the
+agent. There is currently no mechanism to upgrade the agent, the service
+silently works with the older agent assuming that nothing will go wrong. The
+only way to upgrade the agent would be to reboot the host, and we don't ask
+users to do that. Even if we did, it would be inconvenient, specially if there
+are many hosts or if rebooting a host requires manual intervention. This is
+particularly common in the late binding scenario. This is a proposal to
+implement a mechanism to upgrade the agent without having to reboot the host.
+
+Note that all new agent versions require a rollout of the service pods (as the
+agent image name is delivered as an environment variable to the service). Not
+all rollouts of service pods come with a newer agent version.
+
+## Motivation
+
+Reduce compatibility concerns when designing and developing new features.
+
+Reduce the need to reboot hosts when there are changes in the service that
+require new versions of the agent.
+
+### Goals
+
+The agent image will be automatically upgraded to the version expected by the
+service before proceeding to the installation of the cluster.
+
+Note that in this context _version_ is really a container image name, like
+`registry-proxy.engineering.redhat.com/rh-osbs/openshift4-assisted-installer-agent-rhel8:v1.0.0-141`.
+
+### Non-Goals
+
+The upgrade feature will not work for changes in the part of the agent that
+performs the initial registration since the upgrade part relies on the
+registration mechanism.
+
+Upgrading the agent after installation has started is out of scope for this
+enhancement.
+
+Upgrading other images like the image of the service itself of the image of the
+assisted installer or installer controller is completely out of the scope of
+this enhancement.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a developer I want the agent to be automatically upgraded to what the service
+expects, so that I can assume that the service will never proceed to install a
+cluster with a host that is running an incompatible version of the agent.
+
+### Implementation Details/Notes/Constraints
+
+A new `compatible-agent` validation will be added to the service to check if
+the agent is compatible.
+
+The logic for that validation will be a comparison of the value of the
+`AGENT_DOCKER_IMAGE` environment variable passed to the service, and the
+`discovery_agent_version` header sent by the agent and stored in the `hosts`
+table.
+
+The `discovery_agent_version` currently reported by the agent is only the image
+tag. For example, if the full image reference is
+`registry-proxy.engineering.redhat.com/rh-osbs/openshift4-assisted-installer-agent-rhel8:v1.0.0-141`
+the the agent will send only `v1.0.0-141` in the `discovery_agent_version`
+header. That will need to be changed in the agent, so that it sends the full
+image reference. The server will be changed to extract from that full image
+reference the tag where required, in particular for generation of metrics.
+
+In order to support both old and new agents the validation will not be a simple
+string comparison. Instead it will check if the `discovery_agent_version` sent
+by the agent is a full image reference or only a tag. If it is a full image
+reference then it will be a exact string comparison with the value of the
+`AGENT_DOCKER_IMAGE` environment variable. If it is only a tag, then it will be
+a comparison with the tag extracted from the value of the `AGENT_DOCKER_IMAGE`
+environment variable. In this last case a warning will be written to the log
+explaining that the comparison may be unreliable because only the tags were
+compared.
+
+To make this comparison useful it is important that the `AGENT_DOCKER_IMAGE`
+environment variable passed to the service doesn't use the `:latest` tag, or in
+general any moving tag, as that renders the comparison useless. In the _SaaS_
+stage environment that variable is already using a tag, so that shouldn't be a
+problem:
+
+```
+$ oc get deployment assisted-service -n assisted-installer-stage -o json | \
+jq '.spec.template.spec.containers[].env[] | select(.name == "AGENT_DOCKER_IMAGE")'
+{
+  "name": "AGENT_DOCKER_IMAGE",
+  "value": "registry-proxy.engineering.redhat.com/rh-osbs/openshift4-assisted-installer-agent-rhel8:v1.0.0-141"
+}
+```
+
+In the integration and development environments we are using the `:latest` tag,
+which effectively disables this feature.
+
+The service will check during startup the value of the `AGENT_DOCKER_IMAGE` and
+it will write a warning to the log if it is using the `:latest` tag.
+
+A new `upgrade-agent` step will be added to the service and to the agent.
+
+When the service receives a request for next steps from a host it will check if
+that host is in one of the following states:
+
+- `binding`
+- `disabled-unbound`
+- `disabled`
+- `disconnected-unbound`
+- `disconnected`
+- `discovering-unbound`
+- `discovering`
+- `insufficient-unbound`
+- `insufficient`
+- `known-unbound`
+- `known`
+- `pending-for-input`
+- `unbinding-pending-user-action`
+- `unbinding`
+
+If the host is in one of the above states then the service will check the result
+of the `compatible-agent` validation. If the validation passed then the service
+will return the usual next steps. If the validation didn't pass then the service
+will return only the `upgrade-agent` step, replacing all the usual steps.
+
+If on the other hand the host is in one of the following states:
+
+- `added-to-existing-cluster`
+- `cancelled`
+- `error`
+- `installed`
+- `installing-in-progress`
+- `installing`,
+- `preparing-failed`
+- `preparing-for-installation`
+- `preparing-successful`
+- `resetting-pending-user-action`
+- `resetting`
+
+Then the service will return the usual next steps regardless of the result of
+the validation. The reason for this is that we don't want to interrupt ongoing
+installations or log gathering.
+
+The `upgrade-agent` request will have a `agent_image` field containing the full
+reference of the image that the agent should upgrade to.
+
+The `upgrade-agent` response will have a `agent_image` field containing the
+full reference of the image that the agent has upgraded to, and a `result`
+field to indicate if downloaded the image succeeded or failed.
+
+The service will generate a `upgrade-agent-started` event when it sends the
+`upgrade-agent` command to the agent.
+
+The service will generate a `upgrade-agent-finished` event when the agent
+successfully downloads the image, and a `upgrade-agent-failed` event when it
+fails.
+
+If the agent fails to download the image the service will keep sending it the
+`upgrade-agent` command till it succeeds.
+
+The agent will be changed so that it sends the full image reference in the
+`discovery_agent_version` header.
+
+When the agent receives the `upgrade-agent` step it will do the following:
+
+1. It will check if the new image is already downloaded with the `podman image
+   exists ...` command. If the image has already been downloaded it will return
+   to the service a response indicating that the image has been successfully
+   downloaded.
+
+2. If the image hasn't been downloaded it will try to download it using `podman
+   pull ...`. If that succeeds it will exit the next step runner process, so
+   that the main process will start it again with the new image.
+
+3. If downloading the image fails the agent will return to the service a
+   response indicating it and will not exit. This means that the next step
+   runner will continue to run, and it will eventually receive again from the
+   service the instruction to upgrade, and therefore it will try again to download
+   the image.
+
+The `upgrade-agent` step is also (similar to register) something we must not
+break in order to allow agent upgrade.
+
+### Risks and Mitigations
+
+The main risk is introducing interference with the state machine that controls
+the hosts. To reduce the chances of that the upgrade will be performed only when
+the host is not in the process of installing.
+
+The feature will also be behind a feature flag controlled by the
+`ENABLE_UPGRADE_AGENT` environment variable.
+
+## Design Details
+
+None, see the implementation details above.
+
+### Open Questions
+
+- Do the _on premise_ installations use values of `AGENT_DOCKER_IMAGE` with the
+  `:latest` tag? If so that will need to be changed to use a unique version tag or
+  a SHA, otherwise the feature will be effectively disabled.
+
+- Do we want an different handling for the new `compatible-agent` validation in
+  the UI?
+
+- Should we add a new `upgrade_status` text field to the host type to store the
+  current status of the upgrade? This would be with information sent by the
+  agent, something like "downloading new agent image", "failed to download new
+  agent image", etc. It could be useful to have this in the Kubernetes API
+  because events aren't available for users there.
+
+### UI Impact
+
+The UI will at the very minimum need to display correctly the new
+`compatible-agent` validation. Without changes it will display the raw
+`compatible-agent` string as the name of the validation. We will probably want
+to change that to display a more user friendly name, like "Agent compatibility".
+
+We will probably want to change the way that the result of this validation is
+presented to the user. Without change it will be presented like any other failed
+validation, and that means that the user will assume that she has to perform
+some corrective action. In this case the corrective action is automatic, and
+only explained in the description of the validation.
+
+The description of the `compatible-agent` validation will be the following:
+
+> The installation cannot start just yet because this host's agent is in the
+> process of being upgraded to a newer version, please wait, this could take
+> a few minutes.
+
+When the validation fails it will be presented with an hourglass icon, or
+something else less frightening than the regular error icon.
+
+### Test Plan
+
+We will need the following test cases for the service:
+
+- Send an initial registration request to the service containing an incompatible
+  agent image and verify that the validation fails.
+
+- Send an initial registration request to the service containing a compatible
+  agent image and verify that the validation passes.
+
+- Send an initial registration request with a compatible agent image, then send
+  a request for next steps with an incompatible agent and verify that the
+  validation fails.
+
+- Send an initial registration request with a compatible agent image, then send
+  a request for next steps with the same compatible agent image and verify that
+  the validation passes.
+
+- Send an initial registration request with a compatible agent image. Move the
+  host to the `preparing-for-install`, `installing` and `installing-in-progress`
+  states. For each of those states send a request for next steps with an
+  incompatible agent image and verify that the new `upgrade-agent` step is not
+  returned.
+
+- Send an initial registration request with a compatible agent image. Move the
+  host to any non-installing state. For each of those states send a request for
+  next steps with an incompatible agent image and verify that the new
+  `upgrade-agent` step is returned.
+
+We will need the following test cases for the agent:
+
+- Send a request for next steps, make sure that the result is the
+  `upgrade-agent` step and verify that the agent is restarted with the new agent
+  image.
+
+We will need the following end to end test cases:
+
+- Start the service as usual, with the default `AGENT_DOCKER_IMAGE` environment
+  variable. Register a host. Restart the service with a different value of the
+  `AGENT_DOCKER_IMAGE`. Doesn't need to be a different image, just a different
+  value, for example the SHA of that image. Verify that the agent service is
+  restarted and registers again with the new image name.
+
+## Drawbacks
+
+Using a validation for this feature may be confusing for users if we don't
+change the UI to clearly indicate that when an agent isn't compatible with the
+service there is no need for manual corrective actions.
+
+## Alternatives
+
+Introduce a new `upgrading` state in the agent state machine.


docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+last-updated: 2022-06-13
+---
+
+# Agent upgrade
+
+## Summary
+
+Usually the time since the agent is started and the cluster is installed is
+short. But in some situations it may be long, even many days, and during that
+long time the service may be upgraded and require a newer version of the agent.
```
omertuc, line 15:
```suggestion
short. But in some situations it may be long, even many days, and during that
long time the service may be upgraded and may require a newer version of the agent.
```

All new agent versions require a rollout of the service pods (as the agent version is delivered as an environment variable). Not all rollouts of service pods come with a newer agent version.

This should probably be clarified

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+The current mechanism to perform that agent upgrade is to reboot the host. That
```
omertuc, line 16: There is no current mechanism. We just work with older agent versions and pray nothing goes wrong

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+## Motivation
+
+Reduce the need to reboot hosts when there are changes in the service that
+require new versions of the agent.
+
+### Goals
+
+A host that has been booted and registered with matching versions of the agent
+and the service will not need to be rebooted in order to install the cluster
+with a newer version of the service.
```
omertuc, line 30:
```suggestion
with a newer version of the service.
```

This is not really about newer versions of the service, but more about service pods which expect a newer version of the agent. The service version itself is not a concern here

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+
+Note that in this context _version_ is really a container image name, like
+`registry-proxy.engineering.redhat.com/rh-osbs/openshift4-assisted-installer-agent-rhel8:v1.0.0-141`.
+
+### Non-Goals
+
```
omertuc, line 36: Worth mentioning that a change of [INSTALLER_IMAGE or CONTROLLER_IMAGE ](https://github.com/openshift/assisted-service/blob/c1eccd19b6f7a6bdd94639f074a675f6b4a43fcc/internal/host/hostcommands/instruction_manager.go#L54-L55) during installation will not reconciled like the AGENT_IMAGE will be after this enhancement.

The implication of this is that service<->installer and service<->controller communication (which is mostly one directional, aimed at the service) has to remain backwards compatible

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a cluster creator I want to be able to generate an ISO, boot and register
+hosts with a version of the service, then upgrade the service to a version that
+requires a newer version of the agent, and finally install the cluster without
+having to reboot the hosts.
```
omertuc, line 49: Also here, there was never any expectations from users to reboot their hosts, we just upgraded and hoped no one would get hurt in the process

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+
+A new `compatible-agent` validation will be added to the service to check if the
+agent is compatible.
+
+The logic for that validation will be a simple string comparison of the value
+of the `AGENT_DOCKER_IMAGE` environment variable with the value of the
+`discovery_agent_version` stored in the `hosts` table.
+
+To make this comparison useful it is important that the `AGENT_DOCKER_IMAGE`
+environment variable passed to the service doesn't use the `:latest` tag, as
```
omertuc, line 68: Any "moving" tag (that is, a tag that gets pushed to more than once) is bad really, not just `:latest`

docs/enhancements/agent-upgrade.md
omertuc: e.g. `quay.io/edge-infrastructure/assisted-installer-agent:ocm-2.4` is also just as bad as `:latest`, but

docs/enhancements/agent-upgrade.md
omertuc: e.g. `quay.io/edge-infrastructure/assisted-installer-agent:ocm-2.4` is also just as bad as `:latest`

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+```
+$ oc get deployment assisted-service -n assisted-installer-stage -o json | \
+jq '.spec.template.spec.containers[].env[] | select(.name == "AGENT_DOCKER_IMAGE")'
+{
+  "name": "AGENT_DOCKER_IMAGE",
+  "value": "registry-proxy.engineering.redhat.com/rh-osbs/openshift4-assisted-installer-agent-rhel8:v1.0.0-141"
+}
+```
+
+In the integration and development environments we are using the `:latest` tag,
```
omertuc, line 81: We should probably think about fixing this

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+
+A new `upgrade-agent` step will be added to the service and to the agent.
+
+When the service receives a request for next steps from a host it will check if
+that host is in one of the `preparing-for-installation`, `installing` or
+`installing-in-progress` states. In it is then the service will return the next
+steps as usual. If it isn't then the service will retrieve from the database the
+result of the `compatible-agent` validation. If the validation passed then the
+service will return the next steps as usual, otherwise it will replace all the
+usual steps the `upgrade-agent` step.
```
omertuc, line 98: This sounds good

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+the host is not in the process of installing, i.e. not in one of the
+`preparing-for-installation`, `installing` or `installing-in-progress` states.
+
+## Design Details
+
+None, see the implementation details above.
+
+### Open Questions
+
+- Do the _on premise_ installations use values of `AGENT_DOCKER_IMAGE` with the
```
omertuc, line 118: I think there's nothing we can do about this other than warn users in the documentation about "don't use moving image tags". There's no way we could possibly know whether a tag is moving or not so no way for us to completely warn / block users doing this.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+short. But in some situations it may be long, even many days, and during that
+long time the service may be upgraded and require a newer version of the agent.
+The current mechanism to perform that agent upgrade is to reboot the host. That
+is inconvenient, specially if there are many hosts or if rebooting a host
+requires manual intervention. This is a proposal to implement a mechanism to
+upgrade the agent without having to reboot the host.
+
+## Motivation
+
+Reduce the need to reboot hosts when there are changes in the service that
```
eranco74, line 23: Rebooting the hosts isn't a thing, The motivation is to allow comparability between long-running nodes that run old agent version with newer service version that might require newer agent version.

docs/enhancements/agent-upgrade.md
eranco74: I don't think the installer image and controller image have something to do with this enhancement.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+and the service will not need to be rebooted in order to install the cluster
+with a newer version of the service.
+
+Note that in this context _version_ is really a container image name, like
+`registry-proxy.engineering.redhat.com/rh-osbs/openshift4-assisted-installer-agent-rhel8:v1.0.0-141`.
+
+### Non-Goals
+
+Hosts will still need to be rebooted with a new version of the ISO if there
+are changes in the part of the agent that performs the initial registration.
```
eranco74, line 38:
```suggestion
The upgrade feature will not work for changes in the part of the agent that performs the initial registration since the upgrade part relies on the registration mechanism.
```

docs/enhancements/agent-upgrade.md
eranco74: @osherdp thoughts?

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+In the integration and development environments we are using the `:latest` tag,
+which effectively disables this feature.
+
+The description of the `compatible-agent` validation will be the following:
+
+> The installation cannot start just yet because this host's agent is in the
+> process of being upgraded to a newer version, please wait, this could take
+> a few minutes.
+
+A new `upgrade-agent` step will be added to the service and to the agent.
```
eranco74, line 90: The `upgrade-agent` step is also (similar to register) something we must not break in order to allow agent upgrade.

docs/enhancements/agent-upgrade.md
eranco74, line 98: I think that if the `compatible-agent` validation failed the host status should be `insufficient`.
So in essence we can check the `compatible-agent` validation for sending the `agent-upgrade` step only if the host state is `insufficient`.

docs/enhancements/agent-upgrade.md
eranco74: The user won't typically  change the definitions for the agent, so I think we can expect sha-based image on-prem.
We can add a warning saying we might not successfully validate the agent version comparability in case the agent image is tag-based.

eranco74 (commented):
Looks good

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+hosts with a version of the service, then upgrade the service to a version that
+requires a newer version of the agent, and finally install the cluster without
+having to reboot the hosts.
+
+#### Story 2
+
+As a cluster Creator I want to be able to generate an ISO, boot and register
+hosts with a version of the service, then downgrade the service to a version
+that requires an older version of the agent, and finally install the cluster
+without having to reboot the hosts.
```
oourfali, line 56: I guess the main point here is that I'd like to be able to successfully install the cluster even if cases the agent image has to be changed due to the service requirements, whether if it is an "upgrade" change, or a "downgrade" one.

docs/enhancements/agent-upgrade.md
oourfali: s/ "In it is"/ "If it is,"

oourfali: @eranco74 I got confused. the first sentence implies you do the validation first, and then move the status to insufficient.
The second one implies that you do the test only if the status is insufficient.
Or there are two kind of validations - one which just moves the status to insufficient, and the other is actually requesting for an upgrade to take place?

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+`installing-in-progress` states. In it is then the service will return the next
+steps as usual. If it isn't then the service will retrieve from the database the
+result of the `compatible-agent` validation. If the validation passed then the
+service will return the next steps as usual, otherwise it will replace all the
+usual steps the `upgrade-agent` step.
+
+When the agent receives the `agent-upgrade` step it will run the equivalent of
+`systemctl restart agent`. That will restart the agent service, which will
+perform again the initial host registration process, get the new agent image
+name and start the agent container using it.
```
oourfali, line 103: I was under the impression that there is a registration taking place after the new image is used. Is it needed? Or is it as described, registration + get the new agent image name + fetch it + use it?

docs/enhancements/agent-upgrade.md
osherdp: integration and development envs usually have short-termed testing either way so I think it's low risk
we can change how integration env gets updated (automatically). lately we have discussed to automatically update some values in ``saas.yaml`` file, those agent/installer/reporter images in integration env are good candidates as well

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+Reduce the need to reboot hosts when there are changes in the service that
+require new versions of the agent.
+
+### Goals
+
+A host that has been booted and registered with matching versions of the agent
+and the service will not need to be rebooted in order to install the cluster
+with a newer version of the service.
+
+Note that in this context _version_ is really a container image name, like
```
vrutkovs, line 32: I don't think pullspec would be a great choice for this. In CI we may a different registry and on customer installs images may be mirrored.

Perhaps container labelset (or at least vcs url/commit) would suit better

docs/enhancements/agent-upgrade.md
tsorya: Validation should run on all states except "installation ones". We should be able to move from Ready to insufficient if agent image was changed, we have pretty much examples of it already

docs/enhancements/agent-upgrade.md
tsorya: Small issue is that agent is running all steps in parallel, so all other step can fail cause of reboot (not sure if it is a big deal)

tsorya: > I was under the impression that there is a registration taking place after the new image is used. Is it needed? Or is it as described, registration + get the new agent image name + fetch it + use it?

You will get new image upon registration.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+When the service receives a request for next steps from a host it will check if
+that host is in one of the `preparing-for-installation`, `installing` or
+`installing-in-progress` states. In it is then the service will return the next
+steps as usual. If it isn't then the service will retrieve from the database the
+result of the `compatible-agent` validation. If the validation passed then the
+service will return the next steps as usual, otherwise it will replace all the
+usual steps the `upgrade-agent` step.
+
+When the agent receives the `agent-upgrade` step it will run the equivalent of
+`systemctl restart agent`. That will restart the agent service, which will
```
tsorya, line 101: We remove agent image on service restart and it will require re-pulling old image and then pulling new one. I assume they will have many common labels but we already saw systems where it took more then 3 minutes to pull image. Maybe we should consider retry policy on [RunAgent](https://github.com/openshift/assisted-installer-agent/blob/89405454e4f080235a8dd06374bab49be1424231/src/agent/agent.go#L40) itself? I mean if next_step_runner got exit_code "upgrade_required" we should not restart agent, we should only restart register process. It will allow us to pull only one image and mostly will not set host to disconnected as we will be back quicker than 3 minutes

docs/enhancements/agent-upgrade.md
vrutkovs: >We set agent version in configmap as full image path

Ah, okay, that sounds suitable for disconnected usecases

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
jhernand: Replaced both stories with one about the expectations of the developer.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
jhernand: I have opened https://issues.redhat.com/browse/MGMT-10828 to track that. Please comment there.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
jhernand: @eranco74 where will this warning be? In the log? In the UI?

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
jhernand: I am keeping the current description, if you disagree please reopen this conversation.

docs/enhancements/agent-upgrade.md
jhernand: I removed these two stories and replaced them one a single one explaining that the developer expects the agent to be upgraded automatically so that she can do breaking changes to both the service and the agent. Hope that is clearer.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,189 @@
...
+perform again the initial host registration process, get the new agent image
+name and start the agent container using it.
+
+### Risks and Mitigations
+
+The main risk is introducing interference with the state machine that controls
+the hosts. To reduce the chances of that the upgrade will be performed only when
+the host is not in the process of installing, i.e. not in one of the
+`preparing-for-installation`, `installing` or `installing-in-progress` states.
```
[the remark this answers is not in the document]
jhernand: Done.

jhernand:
The new version of the patch addresses the comments received so far, please take another look.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,196 @@
...
+# Agent upgrade
+
+## Summary
+
+Usually the time since the agent is started and the cluster is installed is
+short. But in some situations it may be long, even many days, and during that
+long time the service may be upgraded and may require a newer version of the
+agent. There is currently no mechanism to upgrade the agent, the service
+silently works with the older agent assuming that nothing will go wrong. The
+only way to upgrade the ageint would be to reboot the host, and we don't ask
```
romfreiman, line 18: agent

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,196 @@
...
+creation-date: 2022-06-13
+last-updated: 2022-06-13
+---
+
+# Agent upgrade
+
+## Summary
+
+Usually the time since the agent is started and the cluster is installed is
+short. But in some situations it may be long, even many days, and during that
```
romfreiman, line 14: I would state latebinding as the most common challenge

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,196 @@
...
+  the UI?
+
+### UI Impact
+
+The UI will at the very minimum need to display correctly the new
+`compatible-agent` validation. Without changes it will display the raw
+`compatible-agent` string as the name of the validation. We will probably want
+to change that to display a more user friendly name, like "Agent compatibility".
+
+We will probably want to change the way that the result of this validation is
```
romfreiman, line 139: I think this is a significant ux change and we should talk with @nirfarkas

docs/enhancements/agent-upgrade.md
tsorya: To send step alone we are going to add new state? another way it will require to implement logic in each step per all states we will allow upgrade flow

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,196 @@
...
+
+The `upgrade-agent` step is also (similar to register) something we must not
+break in order to allow agent upgrade.
+
+### Risks and Mitigations
+
+The main risk is introducing interference with the state machine that controls
+the hosts. To reduce the chances of that the upgrade will be performed only when
+the host is not in the process of installing, i.e. not in one of the
+`preparing-for-installation`, `installing`, `installing-in-progress`, `error` or
```
tsorya, line 116: should we send agent upgrade on discovery state? I think we should write down which state upgrade will run with and not which states will be excluded

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,196 @@
...
+The description of the `compatible-agent` validation will be the following:
+
+> The installation cannot start just yet because this host's agent is in the
+> process of being upgraded to a newer version, please wait, this could take
+> a few minutes.
+
+A new `upgrade-agent` step will be added to the service and to the agent.
+
+When the service receives a request for next steps from a host it will check if
+that host is in one of the `preparing-for-installation`, `installing`,
```
tsorya, line 95: should we send agent upgrade on discovery state? I think we should write down which state upgrade will run with and not which states will be excluded

docs/enhancements/agent-upgrade.md
jhernand: @tsorya I was thinking that in the "GetNextSteps" method of the instruction manager we could get the result of the `compatible-agent` validation, and then if the state is one that allows upgrading the agent replace the calculated steps with just the new `upgrade-agent` step.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,196 @@
...
+`installing-in-progress`, `error` or `cancel` states. If it is then the service
+will return the next steps as usual. If it isn't then the service will retrieve
+from the database the result of the `compatible-agent` validation. If the
```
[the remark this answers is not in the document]
jhernand: The "GetNextSteps" method of the instruction manager where I am thinking this logic could go already receives a host object as parameter. The new code would just get the validation results from that object, so conceptually we are retrieving them from the database, but in practice there is no additional database access.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
tsorya: Actually yes, it is possible, sorry for confusion

docs/enhancements/agent-upgrade.md
jhernand: Added a note explaining that upgrading those images is not related and out of scope.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,228 @@
...
+As a developer I want the agent to be automatically upgraded to what the service
+expects, so that I can assume that the service will never proceed to install a
+cluster with a host that is running an incompatible version of the agent.
+
+### Implementation Details/Notes/Constraints
+
+A new `compatible-agent` validation will be added to the service to check if the
+agent is compatible.
+
+The logic for that validation will be a simple string comparison of the value
```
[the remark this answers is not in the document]
jhernand: According to my tests the latest version of the agent sends in the `discovery_agent_version` header only the tag of the image reference. Are we open to change that to the full image reference? You mention it has been changed at least twice in the past? What was the reason to change it to only the tag?

I am updating the document with your suggestion.

docs/enhancements/agent-upgrade.md
jhernand: Done.

omertuc:
Avoid force pushing enhancements, especially when rebasing, it's very hard to know what changed

docs/enhancements/agent-upgrade.md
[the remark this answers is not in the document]
empovit: That's exactly the problem :( we don't really know why it was changed, and then changed back (after a fix for the first change, so the second change broke the fix). Take a look at that PR's Jira ticket.

jhernand:
Sorry for that @omertuc, these are the changes:

image

docs/enhancements/agent-upgrade.md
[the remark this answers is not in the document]
eranco74: Don't we just want to comapre the tag (sha256) anyway?
I don't think the registry or image name matters.
Moreover in case of disconnected installation, the agent fqdn (registry+name+tag) can pint to the internet and relay on the ICSP config on the node to resolve the image from the mirror registry.
So I think that we are better off if we just compare the tag.

docs/enhancements/agent-upgrade.md
eranco74: Actually, we control the validation text in the backend.
This is the current message format, I think it's clear enough, but perhaps @nirfarkas can think of a better phrasing to reflect that this validation should resolve itself. 
```
The installation cannot start just yet because this host's agent is in the process of being upgraded to a newer version, please wait, this could take a few minutes.
``` 
In any case, I don't think this validation requires a different UX than any other validation.
The NTP sync validation is another validation that gets resolved automatically and presented to the user just the same.

docs/enhancements/agent-upgrade.md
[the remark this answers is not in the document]
tsorya: To say the truth i think better to compare full string. Not sure why we should match only tag, it doesn't save anything for us

docs/enhancements/agent-upgrade.md
jhernand: Without changes to the UI it will use the enum value without changes as the title of the validation, something like this:

**compatible-agent**: The installation ...

But we probably want to present it in a more human readable format. We already do it for other validations. Something like this:

**Agent compatibility**: The installation ...

That requires a small change in the UI.

We may also want to present it in a different category. Without changes it will be in the "Hardware" category. We could want to add a new "Software" category. That would require some extra changes in the UI.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+`registry-proxy.engineering.redhat.com/rh-osbs/openshift4-assisted-installer-agent-rhel8:v1.0.0-141`.
+
+### Non-Goals
+
+The upgrade feature will not work for changes in the part of the agent that
+performs the initial registration since the upgrade part relies on the
+registration mechanism.
+
+Upgrading other images like the image of the service itself of the image of the
+Kubernetes controller is completely out of the scope of this enhancement.
```
omertuc, line 50:
```suggestion
assisted installer or installer controller is completely out of the scope of this enhancement.
```

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+Note that in this context _version_ is really a container image name, like
+`registry-proxy.engineering.redhat.com/rh-osbs/openshift4-assisted-installer-agent-rhel8:v1.0.0-141`.
+
+### Non-Goals
+
+The upgrade feature will not work for changes in the part of the agent that
+performs the initial registration since the upgrade part relies on the
+registration mechanism.
+
+Upgrading other images like the image of the service itself of the image of the
```
omertuc, line 49:
```suggestion
Upgrading other images like the image of the service itself or the image of the
```

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+- `preparing-failed`
+- `preparing-successful`
+- `resetting-pending-user-action`
+- `resetting`
+- `unbinding-pending-user-action`
+- `unbinding`
+
+If the host is in one of the above states then the service will check the result
+of the `compatible-agent` validation. If the validation passed then the service
+will return the usal next steps. If the validaton didn't pass then the service
```
omertuc, line 146:
```suggestion
will return the usual next steps. If the validaton didn't pass then the service
```

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+
+- `preparing-for-installation`
+- `installing`,
+- `installing-in-progress`
+- `error`
+- `cancelled`
+
+Then the service will return the usual next steps regardless of the result of
+the validation. The reason for this is that we don't want to interrupt ongoing
+installations of log gathering.
```
omertuc, line 159:
```suggestion
installations or log gathering.
```

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+As a developer I want the agent to be automatically upgraded to what the service
+expects, so that I can assume that the service will never proceed to install a
+cluster with a host that is running an incompatible version of the agent.
+
+### Implementation Details/Notes/Constraints
+
+A new `compatible-agent` validation will be added to the service to check if the
+agent is compatible.
+
+The logic for that validation will be a comparison of the value of the
```
omertuc, line 67: It's important to note that the validation should automatically pass when the host is not in the relevant states (e.g. installing)

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+
+In the integration and development environments we are using the `:latest` tag,
+which effectively disables this feature.
+
+The service will check during startup the value of the `AGENT_DOCKER_IMAGE` and
+it will write a warning to the log if it is using the `:latest` tag.
+
+The description of the `compatible-agent` validation will be the following:
+
+> The installation cannot start just yet because this host's agent is in the
```
omertuc, line 113: Consider changing ⚠️ to ⌛ (or something less frightening) for this validation

docs/enhancements/agent-upgrade.md
omertuc: That's a good point

docs/enhancements/agent-upgrade.md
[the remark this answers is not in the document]
eranco74: @tsorya what version the agent sends when installing in a disconnected env?
If we configure the service with the agent image from the mirror and that is what the agent is sending - great.
But if we rely on the ICSP to translate the agent fqdn in the service to the image from the mirror the agent might report something else.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+- `unbinding`
+
+If the host is in one of the above states then the service will check the result
+of the `compatible-agent` validation. If the validation passed then the service
+will return the usal next steps. If the validaton didn't pass then the service
+will return only the `upgrade-agent` step, replacing all the usual steps.
+
+If on the other hand the host is in one of the followin states:
+
+- `preparing-for-installation`
```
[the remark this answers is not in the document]
eranco74: But the host might have already got the install command (the host moves to installing status when the installer start running on the node)
Since the `preparing-for-installation` should be fairly short (few seconds) I think it's safer not to upgrade once the user clicked install.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Goals
+
+The agent image will be automatically upgraded to the version expected by the
+service before proceeding to the installation of the cluster.
+
+Note that in this context _version_ is really a container image name, like
+`registry-proxy.engineering.redhat.com/rh-osbs/openshift4-assisted-installer-agent-rhel8:v1.0.0-141`.
+
+### Non-Goals
```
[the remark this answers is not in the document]
jhernand: Done.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Implementation Details/Notes/Constraints
+
+A new `compatible-agent` validation will be added to the service to check if the
+agent is compatible.
+
+The logic for that validation will be a comparison of the value of the
+`AGENT_DOCKER_IMAGE` environment variable passed to the service, and the
+`discovery_agent_version` header sent by the agent and stored in the `hosts`
+table.
```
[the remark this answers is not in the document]
jhernand: @avishayt we already have the `disconvery_agent_version` column in the `hosts` table. Can't we use that? Why do you think we need a new column?

docs/enhancements/agent-upgrade.md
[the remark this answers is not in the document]
jhernand: Done.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,256 @@
...
+When the agent receives the `agent-upgrade` step it will do nothing and return
+an exit code indicating that the agent needs to be upgraded. When receiving that
+exit code the next step runner will also exit, with the same code. The main
+program will react to that exit code starting the next step runner with the new
+image.
+
+The `upgrade-agent` step is also (similar to register) something we must not
+break in order to allow agent upgrade.
+
+### Risks and Mitigations
```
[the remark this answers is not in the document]
jhernand: @avishayt is your suggestion to have another mini state machine for upgrades, parallel to the regular host state machine? Or just a free text field just for display purposes?

docs/enhancements/agent-upgrade.md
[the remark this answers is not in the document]
jhernand: @eranco74 the agent extracts the tag from the image reference that it received from the service during host registration, and that is what it sends. Note that this is a pure string manipulation, no inspection of the image is involved.

docs/enhancements/agent-upgrade.md
jhernand: Done.

docs/enhancements/agent-upgrade.md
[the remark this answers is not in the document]
avishayt: a free text field just for display purposes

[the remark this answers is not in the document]
jhernand: @avishayt there will be events generated by the server when the upgrade starts, when it finishes, and when it fails, something like this:

image

So there will already be information available for debugging. That will not be included in the validation message, only in the events. Do you think it is necessary to also have that information in the validation message?

docs/enhancements/agent-upgrade.md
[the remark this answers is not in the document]
avishayt: You are correct, I forgot that we save it.

docs/enhancements/agent-upgrade.md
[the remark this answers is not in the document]
avishayt: Let's leave this as an open question and start without it.  The problem is that when using the k8s API events aren't really accessible to the user.  It would be best to either make the validation stand-alone or start sending k8s events.  Let's discuss with a wider forum offline.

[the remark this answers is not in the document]
jhernand: Added it to the _Open Questions_ section of the document.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,294 @@
...
+
+The service will check during startup the value of the `AGENT_DOCKER_IMAGE` and
+it will write a warning to the log if it is using the `:latest` tag.
+
+A new `upgrade-agent` step will be added to the service and to the agent.
+
+When the service receives a request for next steps from a host it will check if
+that host is in one of the following states:
+
+- `added-to-existing-cluster`
```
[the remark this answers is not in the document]
jhernand: Done.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,294 @@
...
+- `discovering`
+- `installed`
+- `installing-in-progress`
+- `insufficient-unbound`
+- `insufficient`
+- `known-unbound`
+- `known`
+- `pending-for-input`
+- `preparing-failed`
+- `preparing-successful`
```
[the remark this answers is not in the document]
jhernand: Done.

docs/enhancements/agent-upgrade.md
```diff
@@ -0,0 +1,294 @@
...
+- `resetting-pending-user-action`
+- `resetting`
```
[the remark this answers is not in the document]
jhernand: Done.

PR #3586, opened 2022-03-29
MGMT-9710: Provide extended networking defaults in default-config API

mkowalski:
This PR extends the default-config API to contain more defaults
regarding network configuration. This is driven by the fact that
Assisted UI allows now to create single- as well as dual-stack clusters
with multiple combinations of IPv4 and IPv6, therefore it's useful to
have an endpoint that will provide suggested values.

Closes: [MGMT-9710](https://issues.redhat.com/browse/MGMT-9710)

## List all the issues related to this PR

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [x] Cloud
- [ ] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @batzionb

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] Reviewers have been listed
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #3946, opened 2022-06-15
[cloud_hotfix_releases] MGMT-9710: Provide extended networking defaults in default-config API…

osherdp:
… (#3586)

This PR extends the default-config API to contain more defaults
regarding network configuration. This is driven by the fact that
Assisted UI allows now to create single- as well as dual-stack clusters
with multiple combinations of IPv4 and IPv6, therefore it's useful to
have an endpoint that will provide suggested values.

Backports #3586


osherdp:
/hold
(waiting for verification from QE)

osherdp:
Bug verified on integration
/unhold

gamli75 (approved):

PR #3949, opened 2022-06-15
NO-ISSUE: Move host validation models to `internal/models`

jhernand:
Currently the types used to represent the host validations are part of
the `internal/host` package. But those types will be needed also in the
`internal/host/hostcommands` package in order to check the result of the
`compatible-agent` validation that will be introduced as part of
MGMT-10678. It isn't possible to just use the existing types because
that would introduce a circular dependency between the `internal/host`
and `internal/host/hostcommands` packages. To avoid that this patch
moves the types to a new `internal/models` package.

Note that this is a very large patch in terms of the lines that it
modifies, but most of then are just to introduce the new
`internalmodels` package selector in the places where the relocated
types are currently being used.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


gamli75:
I think it's a little bit confusing that we have https://github.com/openshift/assisted-service/tree/master/models folder where all the content was generated by the swagger and another folder internal/models 
@avishayt WDYT?

internal/models/validations.go
```diff
@@ -0,0 +1,116 @@
+package models
```
tsorya, line 1: IMO models is not the best name as we have models package from swagger

jhernand: @tsorya actually initially I thought that these types belong into that "models" package generated fro the OpenAPI specification. But the "ValidationsInfo" of the "Host" type is a string containing a JSON document instead of a typed object. I don't know that is the reason for that, but I assume there is one. That makes me think that we consider that representation of validation results an "internal model", thus the name of the package. Note that in the package selector used in the rest of the code is "internalmodels" to avoid the conflict.

What is your suggestion for the name of this package?

jhernand:
@ori-amizur the reason for needing the `HostValidationID` type in the `hostcommands` package is that I will be adding the code to check the result of the `compatible-agent` validation to the instruction manager, which lives in that `hostcommands` package. To do that I need all the types required to deserialize the `ValidationsInfo`, including the `HostValidationID`.

Note that the `HostValidationID` type already existed, only that it was private and named `validationID`. I assume that the type generated from the OpenAPI specification isn't used directly because we wanted to add the `category` method.

ori-amizur:
@jhernand where is this validation used? Is it part of host state machine ? Is it invoked by refresh-status ?
Why is the validation in hostcommands and not in host ?  hostcommands should not contain validations.  In case conditions are needed to create commands, they are not pre-flight validations that are checked periodically by the monitor.

ori-amizur:
> 

I suggest not to use validations.  There should be conditional function in common, that can be called from host package (for validation), and from hostcommands.  So hostcommands should use this function instead of validation.

eranco74:
> > 
> 
> I suggest not to use validations. There should be conditional function in common, that can be called from host package (for validation), and from hostcommands. So hostcommands should use this function instead of validation.

I think this is a nice idea, but the issue is that we want to set the host status to insufficient in case the agent version on the host isn't the agent version the service requires.
This is why we decided to have a validation for the agent version.
Since this validation is somewhat special, there isn't an action item for the user, instead the service should resolve it by upgrading the agent on the host, we need the service to check this validation status and send the upgrade agent host command in case it's failing.
@ori-amizur @tsorya is there another way we can check the validation status without this refactoring?
@jhernand moving just the validation name to a common module won't do the trick?

jhernand:
@eranco74 we can create a `IsAgentCompatible` function in the `common` package, as suggested by @ori-amizur. Then that can be called when doing the validations, and again from `GetNextSteps`. I this particular case that is perfectly possible because the validation is just an string comparison. I will do that unless there are strong objections.

Regardless I would like to understand what was the reason to make that `validations_info` field of the cluster model a string that contains a serialized JSON object. I guess there will be some historical reasons for that. But for a newcomer like me it looks strange: it should instead be an structured type, and that the corresponding type definition and documentation should be part of the OpenAPI specification, and therefore part of the `models` package.

I am closing this pull request.

PR #3964, opened 2022-06-16
MGMT-10848: Fix double agent unbind when deregistering a cluster

carbonin:
When a cluster is removed in the kubeAPI the clusterdeployments
controller removes all the cluster references from all the agents. This
should cause the agent controller to unbind each of those agents.

The clusterdeployment controller also calls `DeregisterClusterInternal`
which also unbinds the hosts from the cluster.

This isn't a problem currently because both operations should be doing
the same thing, but it will become a problem when hosts should be moved
back to the infraEnv on unbind only if they are not associated with an
BMH, and the kubeAPI is being used. In this case the agent controller
will know if the host should be moved to the infraEnv or not, but the
unbind call from `DeregisterClusterInternal` would have to assume that
the hosts should _not_ be moved back to the infraEnv as it would also be
used in the SaaS REST API.

This difference in implementation would cause a race and needs to be
resolved.

This PR moves all the SaaS specific logic out of
`DeregisterClusterInternal` and into `V2DeregisterCluster`. This will
ensure that this logic is still run when it is needed, but avoid the
race with the controllers when using the kubeAPI.

https://issues.redhat.com/browse/MGMT-10848

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

#### Manual test steps:
- Deployed assisted with test-infra with kube-api enabled
- Created 3 unbound agents (created infra-env, booted hosts with discovery image)
- Created the cluster resources
- Waited for hosts to be bound and ready to install (did not approve hosts)
- Deleted clusterdeployment
- Observed hosts move to known-unbound

I also did the same test in a non-late-binding scenario with an infra-env that referenced a specific cluster deployment.
In this case when I deleted the clusterdeployment the agents were also deleted without issue.

## Assignees

/cc @filanov 
/cc @danielerez 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -136,7 +136,7 @@ type InstallerInternals interface {
 	GetClusterByKubeKey(key types.NamespacedName) (*common.Cluster, error)
 	GetHostByKubeKey(key types.NamespacedName) (*common.Host, error)
 	InstallClusterInternal(ctx context.Context, params installer.V2InstallClusterParams) (*common.Cluster, error)
-	DeregisterClusterInternal(ctx context.Context, params installer.V2DeregisterClusterParams) error
+	DeregisterClusterInternal(ctx context.Context, cluster *common.Cluster) error
 	V2DeregisterHostInternal(ctx context.Context, params installer.V2DeregisterHostParams) error
 	GetCommonHostInternal(ctx context.Context, infraEnvId string, hostId string) (*common.Host, error)
 	UpdateHostApprovedInternal(ctx context.Context, infraEnvId string, hostId string, approved bool) error
@@ -781,35 +781,13 @@ func (b *bareMetalInventory) integrateWithAMSClusterDeregistration(ctx context.C
 	return nil
 }
 
-func (b *bareMetalInventory) DeregisterClusterInternal(ctx context.Context, params installer.V2DeregisterClusterParams) error {
+// DeregisterClusterInternal contains only what is required for the cluster deployment controller to deregister the cluster
+func (b *bareMetalInventory) DeregisterClusterInternal(ctx context.Context, cluster *common.Cluster) error {
 	log := logutil.FromContext(ctx, b.log)
-	var cluster *common.Cluster
-	var err error
-	log.Infof("Deregister cluster id %s", params.ClusterID)
+	log.Infof("Deregister cluster id %s", cluster.ID)
 
-	if cluster, err = common.GetClusterFromDB(b.db, params.ClusterID, common.UseEagerLoading); err != nil {
-		return common.NewApiError(http.StatusNotFound, err)
-	}
-
-	if b.ocmClient != nil {
-		if err = b.integrateWithAMSClusterDeregistration(ctx, cluster); err != nil {
-			log.WithError(err).Errorf("Cluster %s failed to integrate with AMS on cluster deregistration", params.ClusterID)
-			return common.NewApiError(http.StatusInternalServerError, err)
-		}
-	}
-
-	if err = b.deleteDNSRecordSets(ctx, *cluster); err != nil {
-		log.Warnf("failed to delete DNS record sets for base domain: %s", cluster.BaseDNSDomain)
-	}
-
-	if err = b.deleteOrUnbindHosts(ctx, cluster); err != nil {
-		log.WithError(err).Errorf("failed delete or unbind hosts when deregistering cluster: %s", params.ClusterID)
-		return common.NewApiError(http.StatusInternalServerError, err)
-	}
-
-	err = b.clusterApi.DeregisterCluster(ctx, cluster)
-	if err != nil {
-		log.WithError(err).Errorf("failed to deregister cluster %s", params.ClusterID)
+	if err := b.clusterApi.DeregisterCluster(ctx, cluster); err != nil {
+		log.WithError(err).Errorf("failed to deregister cluster %s", cluster.ID)
 		return common.NewApiError(http.StatusNotFound, err)
 	}
 	return nil
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -3437,8 +3437,7 @@ func (b *bareMetalInventory) V2ResetHost(ctx context.Context, params installer.V
 	}
 
 	if host.ClusterID == nil {
-		err = fmt.Errorf("host %s is not bound to any cluster, cannot reset host", params.HostID.String())
-		log.Errorf("ResetHost for host %s is forbidden: not a Day2 hosts", params.HostID.String())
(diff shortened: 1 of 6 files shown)


carbonin:
/hold

This seems like it makes hosts fail to unbind because the unbind process requires the cluster to still be present for some reason.
I'll try to resolve that.

carbonin:
/test edge-e2e-metal-assisted-capi

Some 404 accessing minikube API, seems unrelated.

internal/bminventory/inventory.go
```diff
@@ -3719,13 +3687,14 @@ func (b *bareMetalInventory) GetHostByKubeKey(key types.NamespacedName) (*common
 	if h.ClusterID != nil {
 		cluster, err = common.GetClusterFromDB(b.db, *h.ClusterID, common.SkipEagerLoading)
 		if err != nil {
```
filanov, line 3689: consider handling not found error and return an error for all the other cases

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -63,7 +63,29 @@ func (b *bareMetalInventory) V2GetCluster(ctx context.Context, params installer.
...
+			log.WithError(err).Errorf("Cluster %s failed to integrate with AMS on cluster deregistration", params.ClusterID)
+			return common.NewApiError(http.StatusInternalServerError, err)
+		}
+	}
+
+	if err = b.deleteDNSRecordSets(ctx, *cluster); err != nil {
+		log.Warnf("failed to delete DNS record sets for base domain: %s", cluster.BaseDNSDomain)
+	}
+
+	if err = b.deleteOrUnbindHosts(ctx, cluster); err != nil {
```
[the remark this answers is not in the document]
carbonin: I don't think we supply dns or AMS integration when using the kubeAPI, right?

My intention was to make the `Internal` function only what is needed for kubeAPI as REST should always be a superset.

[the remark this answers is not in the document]
carbonin: > Maybe worth a note somewhere that kube-api is using only the Internal func :/

Isn't this the entire purpose of the `Internal*` functions? I said this in the commit message, but I can add some comment to the internal function definition.

```
This PR moves all the SaaS specific logic out of
`DeregisterClusterInternal` and into `V2DeregisterCluster`. This will
ensure that this logic is still run when it is needed, but avoid the
race with the controllers when using the kubeAPI.
```

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1361,7 +1359,28 @@ func (r *ClusterDeploymentsReconciler) deleteClusterInstall(ctx context.Context,
 	return buildReply(err)
 }
 
-func (r *ClusterDeploymentsReconciler) UnbindAgents(ctx context.Context, log logrus.FieldLogger, clusterDeployment types.NamespacedName) error {
+func (r *ClusterDeploymentsReconciler) shouldDeleteAgentOnUnbind(ctx context.Context, agent aiv1beta1.Agent, clusterDeployment types.NamespacedName) bool {
+	infraEnvName, ok := agent.Labels[aiv1beta1.InfraEnvNameLabel]
```
carbonin, line 1363: I don't see the agent controller setting this anywhere, but we do set it on create.

Should I also add something to the agent controller to ensure this is always set in this PR?

internal/bminventory/inventory.go
```diff
@@ -3437,8 +3437,7 @@ func (b *bareMetalInventory) V2ResetHost(ctx context.Context, params installer.V
 	}
 
 	if host.ClusterID == nil {
-		err = fmt.Errorf("host %s is not bound to any cluster, cannot reset host", params.HostID.String())
-		log.Errorf("ResetHost for host %s is forbidden: not a Day2 hosts", params.HostID.String())
+		log.Errorf("host %s is not bound to any cluster, cannot reset host", params.HostID.String())
```
nmagnezi, line 3440: good catch.

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1359,7 +1359,28 @@ func (r *ClusterDeploymentsReconciler) deleteClusterInstall(ctx context.Context,
...
+	if !ok {
+		return false
```
nmagnezi, line 1365: please add a warning log, as this is a scenario we do not usually expect.

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1359,7 +1359,28 @@ func (r *ClusterDeploymentsReconciler) deleteClusterInstall(ctx context.Context,
...
+	}
+
+	infraEnv := &aiv1beta1.InfraEnv{}
+	if err := r.Get(ctx, types.NamespacedName{Name: infraEnvName, Namespace: agent.Namespace}, infraEnv); err != nil {
+		return false
```
nmagnezi, line 1370: ditto

nmagnezi (changes requested):
I had a conversation with Nick about this PR.
Added some comments, but looks good overall.

Please also update [docs/hive-integration/README.md#clusterdeployment](https://github.com/openshift/assisted-service/blob/master/docs/hive-integration/README.md#clusterdeployment) to reflect the changes.

carbonin:
> Do you think we have to consider stale hosts (bound without cluster) as valid cases?

I'm not sure I understand what would be different.
The agent controller would continue trying to unbind with a reconcile requeue if the unbind operation fails.
What else would we do?

carbonin:
> I think stale hosts should not be a valid case.

What is the consequence of a "stale" host? If this PR stays the way it is and an agent is trying to unbind and failing to update the DB for some reason what kind of issues are we going to run into?

> The flow might be the following: ...

This looks like a good approach if it's needed, but I don't see why it is yet.

carbonin:
/hold

Going to run this through some more manual tests to make sure it's all working like we want it to.

filanov:
@carbonin will SAAS/UI/ACM UX change because of it?
Making sure if UI need to change because of it

internal/controller/controllers/clusterdeployments_controller.go
carbonin: Opened as https://issues.redhat.com/browse/MGMT-10969

carbonin:
I was able to test deleting the cluster deployment both when the infraenv had a cluster reference and when it didn't in a test infra environment and saw everything behave as expected.

@ori-amizur how strongly do you feel that we should hold removing the cluster until the hosts are unbound or deleted in the kube-api case? If you think it can wait for if/when it actually is an issue then this should be ready to go.

/unhold

PR #3970, opened 2022-06-17
MGMT-10850: Message for Day2 reboot is misleading

paul-maidment:
The install progress message in for a node is misleading when a Day2
host is rebooting as it states that "No further updates will be posted"

This is inaccurate as there are certain activities that could lead to an
update to the agent status. For example CSR auto signature operations.

This commit rectifies this problem by removing the confusing part of the
message.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @omertuc 
/cc @ori-amizur 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -155,10 +155,11 @@ type Manager struct {
 	leaderElector                 leader.Leader
 	monitorClusterQueryGenerator  *common.MonitorClusterQueryGenerator
 	monitorInfraEnvQueryGenerator *common.MonitorInfraEnvQueryGenerator
+	kubeApiEnabled                bool
 }
 
 func NewManager(log logrus.FieldLogger, db *gorm.DB, eventsHandler eventsapi.Handler, hwValidator hardware.Validator, instructionApi hostcommands.InstructionApi,
-	hwValidatorCfg *hardware.ValidatorCfg, metricApi metrics.API, config *Config, leaderElector leader.ElectorInterface, operatorsApi operators.API, providerRegistry registry.ProviderRegistry) *Manager {
+	hwValidatorCfg *hardware.ValidatorCfg, metricApi metrics.API, config *Config, leaderElector leader.ElectorInterface, operatorsApi operators.API, providerRegistry registry.ProviderRegistry, kubeApiEnabled bool) *Manager {
 	th := &transitionHandler{
 		db:            db,
 		log:           log,
@@ -178,6 +179,7 @@ func NewManager(log logrus.FieldLogger, db *gorm.DB, eventsHandler eventsapi.Han
 		metricApi:      metricApi,
 		Config:         *config,
 		leaderElector:  leaderElector,
+		kubeApiEnabled: kubeApiEnabled,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


omertuc:
The message was good in the context of REST AI, it's not good in the context of ZTP. The solution is not to change the message, the solution is to remove it altogether in the context of ZTP

internal/host/host.go
```diff
@@ -586,8 +588,13 @@ func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, pro
 			swag.StringValue(h.Status), models.HostStatusError, statusInfo)
 	case models.HostStageRebooting:
 		if swag.StringValue(h.Kind) == models.HostKindAddToExistingClusterHost {
+			statusInfoMessage := statusInfoRebootingDay2
+			if m.kubeApiEnabled {
+				statusInfoMessage = statusInfoRebootingDay2ManagedInstallation
```
omertuc, line 593: We don't need any message here, the user doesn't need to do anything, just like they don't need to have a message / do anything for a regular rebooting stage during a normal day-1 installation

internal/host/host.go
```diff
@@ -586,8 +588,13 @@ func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, pro
 			swag.StringValue(h.Status), models.HostStatusError, statusInfo)
 	case models.HostStageRebooting:
 		if swag.StringValue(h.Kind) == models.HostKindAddToExistingClusterHost {
+			statusInfoMessage := statusInfoRebootingDay2
+			if m.kubeApiEnabled {
+				statusInfoMessage = ""
```
omertuc, line 593: Why is this needed? Just add an `&&` to the `if swag.StringValue(h.Kind) == models.HostKindAddToExistingClusterHost` condition above

paul-maidment: A few lines below we update the host with a status, this update does not depend on KubeAPI being enabled. 

I don't think it's appropriate to add the condition to the `if` statement for this reason.

internal/host/host.go
```diff
@@ -585,9 +587,11 @@ func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, pro
 		_, err = hostutil.UpdateHostStatus(ctx, logutil.FromContext(ctx, m.log), m.db, m.eventsHandler, h.InfraEnvID, *h.ID,
 			swag.StringValue(h.Status), models.HostStatusError, statusInfo)
 	case models.HostStageRebooting:
-		if swag.StringValue(h.Kind) == models.HostKindAddToExistingClusterHost {
+		if swag.StringValue(h.Kind) == models.HostKindAddToExistingClusterHost && !m.kubeApiEnabled{
+			statusInfoMessage := statusInfoRebootingDay2
+
```
omertuc, line 592:
```suggestion
```

internal/host/host.go
```diff
@@ -585,9 +587,11 @@ func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, pro
...
 			_, err = hostutil.UpdateHostProgress(ctx, logutil.FromContext(ctx, m.log), m.db, m.eventsHandler, h.InfraEnvID, *h.ID,
-				swag.StringValue(h.Status), models.HostStatusAddedToExistingCluster, statusInfoRebootingDay2,
+				swag.StringValue(h.Status), models.HostStatusAddedToExistingCluster, statusInfoMessage,
```
omertuc, line 594:
```suggestion
				swag.StringValue(h.Status), models.HostStatusAddedToExistingCluster, statusInfoRebootingDay2,
```

internal/host/host.go
```diff
@@ -585,9 +587,11 @@ func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, pro
 		_, err = hostutil.UpdateHostStatus(ctx, logutil.FromContext(ctx, m.log), m.db, m.eventsHandler, h.InfraEnvID, *h.ID,
 			swag.StringValue(h.Status), models.HostStatusError, statusInfo)
 	case models.HostStageRebooting:
-		if swag.StringValue(h.Kind) == models.HostKindAddToExistingClusterHost {
+		if swag.StringValue(h.Kind) == models.HostKindAddToExistingClusterHost && !m.kubeApiEnabled{
```
filanov, line 590: Maybe instead of passing another parameter you can check if the host have a kube-key? https://github.com/openshift/assisted-service/blob/7f0a41b671297e987ae07a6e55d32f82a5c0c974/internal/common/db.go#L78

paul-maidment: I think that `h` here is a models.Host, whereas the KubeKeyNamespace field is an attribute of common.Host. 
Not sure we can use this here?

omertuc:
Feel free to unhold once @paul-maidment and @filanov agree

internal/host/host.go
filanov: Make sense.

PR #3988, opened 2022-06-20
AGENT-215: Set release mirror for agent-based-installer

bfournie:
For the agent-based-installer the release mirror needs to be set when configured in the agent. Use the assisted-service environment variable to accomplish this.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/agentbasedinstaller/agent_based_installer.go
+++ b/cmd/agentbasedinstaller/agent_based_installer.go
@@ -21,7 +21,7 @@ import (
 )
 
 func RegisterCluster(ctx context.Context, log *log.Logger, bmInventory *client.AssistedInstall, pullSecret string, clusterDeploymentPath string,
-	agentClusterInstallPath string, clusterImageSetPath string) (*models.Cluster, error) {
+	agentClusterInstallPath string, clusterImageSetPath string, releaseImageMirror string) (*models.Cluster, error) {
 	var cd hivev1.ClusterDeployment
 	if cdErr := GetFileData(clusterDeploymentPath, &cd); cdErr != nil {
 		return nil, cdErr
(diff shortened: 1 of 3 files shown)


bfournie:
> I'm not sure what this `releaseMirror` param is doing. (seems obsolete) Seems that it's never passed to the `getReleaseVersionAndCpuArch` function. From looking at the code it seems that you can just use `releaseImage` with the argument you want to pass as `releaseMirror` and get the same behavior.

So the releaseMirror is passed from getReleaseVersionAndCpuArch() to releaseHandler.GetOpenshiftVersion() and there is a slight difference in behavior - https://github.com/openshift/assisted-service/blob/master/internal/oc/release.go#L118. The last parameter to getImageFromRelease() "insecure" is always true in this case. So just setting releaseImage to releaseMirror won't give the same behavior.

eranco74:
> > I'm not sure what this `releaseMirror` param is doing. (seems obsolete) Seems that it's never passed to the `getReleaseVersionAndCpuArch` function. From looking at the code it seems that you can just use `releaseImage` with the argument you want to pass as `releaseMirror` and get the same behavior.
> 
> So the releaseMirror is passed from getReleaseVersionAndCpuArch() to releaseHandler.GetOpenshiftVersion() and there is a slight difference in behavior - https://github.com/openshift/assisted-service/blob/master/internal/oc/release.go#L118. The last parameter to getImageFromRelease() "insecure" is always true in this case. So just setting releaseImage to releaseMirror won't give the same behavior.

You are right,
Since your code is going to be the only place that uses the releaseMirror option, consider changing GetOpenshiftVersion signature to get `insecure bool`  instead of `releaseMirror string`.
e.g.
`func (r *release) GetOpenshiftVersion(log logrus.FieldLogger, releaseImage string, insecure bool, pullSecret string) (string, error)`
It should simplify the code, and actually represent what it's actually mean.

PR #4004, opened 2022-06-22
MGMT-10912: Add anonymized user identifier to assisted daatabase

eliorerz:
As part of anonymizing elastic data this PR adds anonymized unique user ID alongside the user name/email stored in the DB.
This user-id will replace the user name/email on assisted events.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @gamli75

### DIFF
--- a/api/vendor/github.com/openshift/assisted-service/models/cluster.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/cluster.go
@@ -231,6 +231,9 @@ type Cluster struct {
 	// Format: date-time
 	UpdatedAt timeext.Time `json:"updated_at,omitempty" gorm:"type:timestamp with time zone"`
 
+	// An unique user ID that is generated from the given user-name
+	UserID string `json:"user_id,omitempty"`
+
 	// Indicate if the networking is managed by the user.
 	UserManagedNetworking *bool `json:"user_managed_networking,omitempty"`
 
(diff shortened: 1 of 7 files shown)


gamli75:
one remark - please add a test that the same user creates 2 clusters and verify that we will have the same user Id for both clusters

nmagnezi (commented):
A couple of questions:
1. Is that hashing method reversible?
2. If yes, how do we make sure external viewers won't de-anonymize user names? I did not see any secret key used to hash the names.

eliorerz:
> A couple of questions:
> 
> 1. Is that hashing method reversible?
> 2. If yes, how do we make sure external viewers won't de-anonymize user names? I did not see any secret key used to hash the names.

No, except from brute-force you cannot reverse this hash function.
This hash function is no cryptographic so it doesn't use any key to encrypt the user-name.

eliorerz:
@nmagnezi The only "weakness" here, is that we can allegedly get single user-id for few different username

romfreiman:
why do we add it to the DB? rather then doing the anonymazation as part of the pipline

eliorerz:
The decision was made to do it only on `assisted-events-scrape` https://github.com/openshift-assisted/assisted-events-scrape/pull/119

PR #4005, opened 2022-06-23
Use releaseImageMirror in GetReleaseArchitecture

bfournie:
Use the mirror when running 'oc image info', similar to how all of the other 'oc' commands are used to get release info.
This functionality is only used by the agent based installer.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/oc/release.go
+++ b/internal/oc/release.go
@@ -39,7 +39,7 @@ type Release interface {
 	GetMustGatherImage(log logrus.FieldLogger, releaseImage string, releaseImageMirror string, pullSecret string) (string, error)
 	GetOpenshiftVersion(log logrus.FieldLogger, releaseImage string, releaseImageMirror string, pullSecret string) (string, error)
 	GetMajorMinorVersion(log logrus.FieldLogger, releaseImage string, releaseImageMirror string, pullSecret string) (string, error)
-	GetReleaseArchitecture(log logrus.FieldLogger, releaseImage string, pullSecret string) (string, error)
+	GetReleaseArchitecture(log logrus.FieldLogger, releaseImage string, releaseImageMirror string, pullSecret string) (string, error)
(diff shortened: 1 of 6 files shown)


internal/oc/release.go
```diff
@@ -145,11 +145,16 @@ func (r *release) GetMajorMinorVersion(log logrus.FieldLogger, releaseImage stri
 	return fmt.Sprintf("%d.%d", v.Segments()[0], v.Segments()[1]), nil
 }
 
-func (r *release) GetReleaseArchitecture(log logrus.FieldLogger, releaseImage string, pullSecret string) (string, error) {
-	if releaseImage == "" {
-		return "", errors.New("no releaseImage is provided")
+func (r *release) GetReleaseArchitecture(log logrus.FieldLogger, releaseImage string, releaseImageMirror string, pullSecret string) (string, error) {
```
mhrivnak, line 148: Why does this function need to change? Looks like it would do the same thing if you just passed the different value as the `releaseImage` argument.

I see there's some logic here that allows passing both values, and this function will choose to prefer the mirror in that case. But it seems like that logic should happen in one place at a higher level than here.

bfournie: Yes, the check for the releaseImageMirror could definitely be done at the higher level and pass in one value for the image, however all the other public functions in this file take both the releaseImage and the mirror - https://github.com/openshift/assisted-service/blob/master/internal/oc/release.go#L36-L41 so this change makes this function consistent with the rest.

omertuc: sgtm, @mhrivnak unhold if it's fine by you

PR #4012, opened 2022-06-23
MGMT-9982: Alter ipxe-script endpoint to handle boot order

ori-amizur:
When using iPXE would like to handle any boot order.  The options are
[network, hd], and [hd, network].
Currently we are only handling the second case, where we attempt to boot
from hd first.  If we fail, then we attempt to boot from network.
If network is first, we need to handle the case that hd is already
installed and host is rebooting, or installed.  In this case we would
like to avoid using the iPXE booting and boot from hd instead.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @carbonin 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -5040,6 +5040,9 @@ func (b *bareMetalInventory) V2UpdateHostIgnitionInternal(ctx context.Context, p
 }
 
 func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params installer.V2DownloadInfraEnvFilesParams) middleware.Responder {
+	if swag.BoolValue(params.BootControl) && params.FileName != "ipxe-script" {
+		return common.NewApiError(http.StatusBadRequest, errors.New(`boot_control can be set only for "ipxe-script"`))
+	}
 	infraEnv, err := common.GetInfraEnvFromDB(b.db, params.InfraEnvID)
 	if err != nil {
 		b.log.WithError(err).Errorf("Failed to get infra env %s", params.InfraEnvID)
@@ -5056,7 +5059,7 @@ func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params
 		}
 		filename = params.FileName
 	case "ipxe-script":
-		content, err = b.infraEnvIPXEScript(ctx, infraEnv)
+		content, err = b.infraEnvIPXEScript(ctx, infraEnv, params.Mac, swag.BoolValue(params.BootControl))
 		if err != nil {
 			b.log.WithError(err).Error("Failed to create ipxe script")
 			return common.GenerateErrorResponder(err)
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -120,6 +120,11 @@ var (
 	imageServiceBaseURL = fmt.Sprintf("https://%s%s", imageServiceHost, imageServicePath)
 )
 
+func toMac(macStr string) *strfmt.MAC {
+	mac := strfmt.MAC(macStr)
+	return &mac
+}
+
 func mockClusterRegisterSteps() {
 	mockSecretValidator.EXPECT().ValidatePullSecret(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil).Times(1)
 	mockVersions.EXPECT().GetReleaseImage(gomock.Any(), gomock.Any()).Return(common.TestDefaultConfig.ReleaseImage, nil).Times(1)
@@ -9119,6 +9124,7 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
 
 	BeforeEach(func() {
 		db, dbName = common.PrepareTestDB()
+		cfg.ServiceBaseURL = "https://test.com"
 		bm = createInventory(db, cfg)
 		var err error
 		bm.ImageExpirationTime, err = time.ParseDuration("4h")
@@ -9143,10 +9149,19 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
 		ctrl.Finish()
 	})
 
-	getResponseData := func(fileName string) []byte {
+	getResponse := func(fileName string, withMac, bootControl bool) middleware.Responder {
 		params := installer.V2DownloadInfraEnvFilesParams{InfraEnvID: infraEnvID, FileName: fileName}
-		response := bm.V2DownloadInfraEnvFiles(ctx, params)
+		if withMac {
+			params.Mac = toMac("f8:75:a4:a4:00:fe")
+		}
+		if bootControl {
+			params.BootControl = swag.Bool(true)
+		}
+		return bm.V2DownloadInfraEnvFiles(ctx, params)
+	}
 
+	getResponseData := func(fileName string, withMac, bootControl bool) []byte {
+		response := getResponse(fileName, withMac, bootControl)
 		fileMw, ok := response.(*filemiddleware.FileMiddlewareResponder)
 		Expect(ok).To(BeTrue())
 		innerType, ok := fileMw.GetNext().(*installer.V2DownloadInfraEnvFilesOK)
@@ -9159,7 +9174,7 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
 
 	It("returns discovery.ign successfully", func() {
 		mockIgnitionBuilder.EXPECT().FormatDiscoveryIgnitionFile(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(discovery_ignition_3_1, nil).Times(1)
-		body := getResponseData("discovery.ign")
+		body := getResponseData("discovery.ign", false, false)
 		config, report, err := ign_3_1.Parse(body)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(report.IsFatal()).To(BeFalse())
@@ -9180,7 +9195,53 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
 
 	It("returns ipxe-script successfully", func() {
 		mockVersions.EXPECT().GetOsImage(common.TestDefaultConfig.OpenShiftVersion, gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).Times(1)
-		content := getResponseData("ipxe-script")
+		content := getResponseData("ipxe-script", false, false)
+		lines := strings.Split(string(content), "\n")
+
+		Expect(lines[0]).To(Equal("#!ipxe"))
+
+		By("validating the kernel line")
+		kernelRegex := regexp.MustCompile(`^kernel (\S+) initrd=initrd coreos.live.rootfs_url=(\S+) (.+)`)
+		match := kernelRegex.FindStringSubmatch(lines[2])
+		fmt.Fprintf(GinkgoWriter, "lines[2]: %s\n", lines[2])
+		fmt.Fprintf(GinkgoWriter, "match: %s\n", match)
+		Expect(match).NotTo(BeNil())
+
+		kernelURL, err := url.Parse(match[1])
+		Expect(err).NotTo(HaveOccurred())
+		Expect(kernelURL.Host).To(Equal(imageServiceHost))
+		Expect(kernelURL.Path).To(Equal(imageServicePath + "/boot-artifacts/kernel"))
+		Expect(kernelURL.Query().Get("version")).To(Equal(*common.TestDefaultConfig.OsImage.OpenshiftVersion))
+		Expect(kernelURL.Query().Get("arch")).To(Equal(*common.TestDefaultConfig.OsImage.CPUArchitecture))
+
+		rootfsURL, err := url.Parse(match[2])
+		Expect(err).NotTo(HaveOccurred())
+		Expect(rootfsURL.Host).To(Equal(imageServiceHost))
+		Expect(rootfsURL.Path).To(Equal(imageServicePath + "/boot-artifacts/rootfs"))
+		Expect(rootfsURL.Query().Get("version")).To(Equal(*common.TestDefaultConfig.OsImage.OpenshiftVersion))
+		Expect(rootfsURL.Query().Get("arch")).To(Equal(*common.TestDefaultConfig.OsImage.CPUArchitecture))
+
+		Expect(match[3]).To(Equal(`random.trust_cpu=on rd.luks.options=discard ignition.firstboot ignition.platform.id=metal console=tty1 console=ttyS1,115200n8 coreos.inst.persistent-kargs="console=tty1 console=ttyS1,115200n8"`))
+
+		By("validating the initrd line")
+		initrdRegex := regexp.MustCompile(`^initrd --name initrd (.+)`)
+		match = initrdRegex.FindStringSubmatch(lines[1])
+		Expect(match).NotTo(BeNil())
+
+		initrdURL, err := url.Parse(match[1])
+		Expect(err).NotTo(HaveOccurred())
+		Expect(initrdURL.Scheme).To(Equal("http"))
+		Expect(initrdURL.Host).To(Equal(imageServiceHost))
+		Expect(initrdURL.Path).To(Equal(fmt.Sprintf("%s/images/%s/pxe-initrd", imageServicePath, infraEnvID)))
+		Expect(initrdURL.Query().Get("version")).To(Equal(*common.TestDefaultConfig.OsImage.OpenshiftVersion))
+		Expect(initrdURL.Query().Get("arch")).To(Equal(*common.TestDefaultConfig.OsImage.CPUArchitecture))
+
+		Expect(lines[3]).To(Equal("boot"))
+	})
+
+	It("returns ipxe-script successfully with mac", func() {
+		mockVersions.EXPECT().GetOsImage(common.TestDefaultConfig.OpenShiftVersion, gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).Times(1)
+		content := getResponseData("ipxe-script", true, false)
 		lines := strings.Split(string(content), "\n")
 
 		Expect(lines[0]).To(Equal("#!ipxe"))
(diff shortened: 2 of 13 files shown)


internal/bminventory/inventory_v2_handlers.go
```diff
@@ -605,13 +607,66 @@ func (b *bareMetalInventory) signURL(ctx context.Context, infraEnvID, urlString,
...
+	serviceURL.Scheme = "http"
+
+	return fmt.Sprintf(ipxeFirstScriptFormat, serviceURL.String(), urlPath), nil
+}
+
+func (b *bareMetalInventory) shouldServScript(infraEnv *common.InfraEnv, mac *strfmt.MAC) (bool, error) {
+	var hosts []*models.Host
+	macStr := mac.String()
+	if err := b.db.Where("infra_env_id = ? and (inventory like ? or inventory like ?)", infraEnv.ID.String(), fmt.Sprintf("%%%s%%", strings.ToUpper(macStr)),
+		fmt.Sprintf("%%%s%%", strings.ToLower(macStr))).Find(&hosts).Error; err != nil {
```
carbonin, line 641: This is quite the query.

I wonder if this is worth its own function.
Also is this specifically better than doing `LOWER(inventory) like ...`?
Then you wouldn't need the second case.

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -605,13 +607,66 @@ func (b *bareMetalInventory) signURL(ctx context.Context, infraEnvID, urlString,
...
+		return false, err
+	}
+	if len(hosts) != 1 {
+		return true, nil
+	}
```
carbonin, line 646: I think we should check separately for `0` vs `>1`.

The zero case should be this logic, but if we somehow we get more than one I think we should return an error.

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -638,7 +693,14 @@ func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *c
...
 	}
 
-	return fmt.Sprintf(ipxeScriptFormat, initrdURL, kernelURL, rootfsURL), nil
+	return fmt.Sprintf(ipxeSecondScriptFormat, initrdURL, kernelURL, rootfsURL), nil
+}
+
+func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *common.InfraEnv, mac *strfmt.MAC, urlPath string) (string, error) {
+	if mac == nil || len(*mac) == 0 {
+		return b.firstIPXEScript(urlPath)
+	}
```
carbonin, line 702: This means that we *have* to pass a host mac to get the script at all.
I thought we talked about retaining a way to query for the script without requiring a host.

internal/bminventory/inventory.go
```diff
@@ -198,6 +199,7 @@ type bareMetalInventory struct {
 	gcConfig             garbagecollector.Config
 	providerRegistry     registry.ProviderRegistry
 	insecureIPXEURLs     bool
+	httpListenPortStr    string
```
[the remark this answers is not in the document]
ori-amizur: Other artifacts access the image service.

internal/bminventory/inventory_v2_handlers.go
ori-amizur: The inventory is the field we query. mac is the variable we pass to the query.  
I don't think to perform LOWER on such a big field is a good idea.

internal/bminventory/inventory_v2_handlers.go
ori-amizur: OK

internal/bminventory/inventory.go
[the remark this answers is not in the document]
carbonin: Not the ipxe script which we are currently exposing conditionally as http from the assisted service [here](https://github.com/openshift/assisted-service/blob/6f2f295b170c39550e59e30b1d49356d128f5a67/internal/controller/controllers/infraenv_controller.go#L533-L555)

Currently it looks like we're relying on this working with a route which would not require a port and would assume the defaults for http(s). This works in the route case, but wouldn't if we were exposing http and https simultaneously and also not using an openshift route.

Fortunately I don't think we have such a use case currently. If we do, I think we should tackle it as a separate effort since, like I said, we'd need more than just this PR to address it fully (the image service would also need two ports).

internal/bminventory/inventory_v2_handlers.go
ori-amizur: We can perform the query with dummy mac.  It will provide the script.

internal/bminventory/inventory.go
```diff
@@ -4925,7 +4929,11 @@ func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params
 		}
 		filename = params.FileName
 	case "ipxe-script":
-		content, err = b.infraEnvIPXEScript(ctx, infraEnv)
+		urlPath := fmt.Sprintf("%s/v2/infra-envs/%s/downloads/files", client.DefaultBasePath, infraEnv.ID.String())
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/bminventory/inventory.go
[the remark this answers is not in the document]
ori-amizur: When we use two ports, we just get a configuration of http port.
We do not have any insecure service URL, like we have The service URL.  We currently have the http port as configuration parameter, which we listen to.
How can we know what are the host and port of the insecure service URL? Maybe we should have another configuration parameter for this?  (InsecureServiceBaseURL) ?

internal/bminventory/inventory_v2_handlers.go
ori-amizur: OK. Changed

internal/bminventory/inventory_v2_handlers.go
ori-amizur: Return error means that we do not serve IPXE script

internal/bminventory/inventory_test.go
```diff
@@ -9177,9 +9192,98 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
 			os.Unsetenv("EC_PRIVATE_KEY_PEM")
 		})
 
+		Context("Skipped IPXE", func() {
+			scenarios := []struct {
```
carbonin, line 9196: I very much don't like this style of test.
I think it ruins one of the best parts of BDD frameworks which is how the test clearly describes the behavior that is expected.

Also this becomes very difficult to maintain if only one of these cases needs some kind of different handling in the future.

I think I'd prefer to have a few different examples like this intead of this scenario struct and a loop:
```go
It("doesn't serve the ipxe script when the host is installing or installed", func() {
  // Test all the InstallingInProgress/Installed cases here
})

It("serves the ipxe script when the host is not installing", func() {
  // Test the other cases here
})
```

internal/bminventory/inventory_test.go
```diff
@@ -9177,9 +9192,98 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
...
+						match := initrdRegex.FindStringSubmatch(strings.Split(string(content), "\n")[1])
+						Expect(match).NotTo(BeNil())
+					} else {
+						response := getResponse("ipxe-script", true, true)
+						verifyApiErrorString(response, http.StatusNotFound, "IPXE booting skipped")
+					}
+				})
+			}
+		})
+		It("IPXE without mac with boot control", func() {
```
carbonin, line 9277: Same here, why is this in the local auth context?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -605,13 +606,63 @@ func (b *bareMetalInventory) signURL(ctx context.Context, infraEnvID, urlString,
...
+		return "", err
+	}
+	builder := installer.V2DownloadInfraEnvFilesURL{
+		InfraEnvID: *infraEnv.ID,
+		FileName:   "ipxe-script",
+	}
+	return fmt.Sprintf(ipxeFirstScriptFormat, builder.StringFull(parsedURL.Scheme, parsedURL.Host)), nil
+}
+
+func (b *bareMetalInventory) shouldServeScript(infraEnv *common.InfraEnv, mac *strfmt.MAC) (bool, error) {
```
carbonin, line 635: I think this could use a better name. Maybe `shouldServeHostIPXEScript`?
That way someone immediately knows it's specific to IPXE and to an individual host.

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -605,13 +606,63 @@ func (b *bareMetalInventory) signURL(ctx context.Context, infraEnvID, urlString,
...
+`
+
+const ipxeSecondScriptFormat = `#!ipxe
 initrd --name initrd %s
 kernel %s initrd=initrd coreos.live.rootfs_url=%s random.trust_cpu=on rd.luks.options=discard ignition.firstboot ignition.platform.id=metal console=tty1 console=ttyS1,115200n8 coreos.inst.persistent-kargs="console=tty1 console=ttyS1,115200n8"
 boot
 `
 
-func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *common.InfraEnv) (string, error) {
+func (b *bareMetalInventory) firstIPXEScript(infraEnv *common.InfraEnv) (string, error) {
```
carbonin, line 619: I think this should have a better name. Maybe `hostRedirectIPXEScript` that way we're describing what the script does in contrast to the one that boots the host rather than just referencing it by order.

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -605,13 +606,63 @@ func (b *bareMetalInventory) signURL(ctx context.Context, infraEnvID, urlString,
...
+	}
+	h := hosts[0]
+	switch swag.StringValue(h.Status) {
+	case models.HostStatusInstalled:
+		return false, nil
+	case models.HostStatusInstallingInProgress:
+		if h.Progress != nil {
+			switch h.Progress.CurrentStage {
+			case models.HostStageDone, models.HostStageConfiguring, models.HostStageJoined, models.HostStageRebooting, models.HostStageWaitingForIgnition:
+				return false, nil
```
carbonin, line 658: Since a `false` here will cause an API error I think it would be great to get some information about why we're erroring in the response and in the log.

Do you think we could refactor this function so that it only returns an error type and we serve the script only when it succeeds?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -638,7 +689,29 @@ func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *c
...
+func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *common.InfraEnv, mac *strfmt.MAC, bootControl bool) (string, error) {
+	if mac == nil || len(*mac) == 0 {
+		if bootControl {
+			return b.firstIPXEScript(infraEnv)
+		} else {
+			return b.bootIPXEScript(ctx, infraEnv)
+		}
+	}
+	return b.secondIPXEScript(ctx, infraEnv, mac)
 }
```
carbonin, line 715: I think if you combined these the logic would read better:

```go
func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *common.InfraEnv, mac *strfmt.MAC, bootControl bool) (string, error) {
    if mac != nil && len(*mac) != 0 {
        if err := b.shouldServeHostIPXEScript(infraEnv, mac); err != nil {
            return "", err
        }
    }

    if bootControl {
        return b.hostRedirectIPXEScript(infraEnv)
    }
    return b.bootIPXEScript(ctx, infraEnv)
```

Also this example assumes some functions are renamed/refactored as mentioned in the other comments.

internal/bminventory/inventory_v2_handlers.go
carbonin: If we do this refactoring the function might need another rename too.

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -605,13 +606,63 @@ func (b *bareMetalInventory) signURL(ctx context.Context, infraEnvID, urlString,
...
+	}
+	parsedURL, err := url.Parse(serviceBaseURL)
+	if err != nil {
+		return "", err
+	}
+	builder := installer.V2DownloadInfraEnvFilesURL{
+		InfraEnvID: *infraEnv.ID,
+		FileName:   "ipxe-script",
+	}
+	return fmt.Sprintf(ipxeFirstScriptFormat, builder.StringFull(parsedURL.Scheme, parsedURL.Host)), nil
```
carbonin, line 632: This URL should be signed if we're using local auth, right?

internal/bminventory/inventory_test.go
```diff
@@ -9177,9 +9192,98 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
 			os.Unsetenv("EC_PRIVATE_KEY_PEM")
 		})
 
+		Context("Skipped IPXE", func() {
```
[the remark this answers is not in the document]
carbonin: Maybe we should keep it here and add a test that the redirect URL is signed.

internal/bminventory/inventory_test.go
ori-amizur: I previously received comments that I am not running similar tests in loop.  We need to have single approach for such tests

internal/bminventory/inventory.go
```diff
@@ -89,6 +89,7 @@ type Config struct {
 	ignition.IgnitionConfig
 	AgentDockerImg                      string            `envconfig:"AGENT_DOCKER_IMAGE" default:"quay.io/edge-infrastructure/assisted-installer-agent:latest"`
 	ServiceBaseURL                      string            `envconfig:"SERVICE_BASE_URL"`
+	InsecureServiceBaseURL              string            `envconfig:"INSECURE_SERVICE_BASE_URL"`
```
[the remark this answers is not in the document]
carbonin: We don't have an insecure port. If we needed it we would already have it for the existing ipxe script endpoint that uses `http`

If we do need it then that's a separate task.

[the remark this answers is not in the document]
ori-amizur: So what is your suggestion ?
Do you suggest a solution to take the host from the ServiceBaseURL environment variable and change the scheme to http and change the port to 80 ?
Should it be hard coded or dependent on some flag ?
We cannot use the ServiceBaseURL as is.  Especially when we have http-port specified

internal/bminventory/inventory_v2_handlers.go
ori-amizur: OK

internal/bminventory/inventory_test.go
```diff
@@ -9177,9 +9192,98 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
...
+				},
+			}
+			for i := range scenarios {
+				s := scenarios[i]
+				It(s.name, func() {
+					id := strfmt.UUID(uuid.New().String())
+					host := models.Host{
+						ID:         &id,
+						InfraEnvID: infraEnvID,
+						Inventory:  "abc f8:75:a4:a4:00:fe def",
```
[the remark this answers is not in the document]
carbonin: > I would rather not parse all the inventories just to locate the host with the mac address.

I agree, I'm just illustrating how this test is fragile.

internal/bminventory/inventory_v2_handlers.go
ori-amizur: I will do the refactor. Look at line 701.  An error is generated with specific text.  Anything else is needed ?

internal/bminventory/inventory.go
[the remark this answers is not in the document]
carbonin: And if we do decide there's a problem with this then it needs to handle both cases and therefore doesn't belong in this PR.

internal/bminventory/inventory_v2_handlers.go
carbonin: Yeah, I think it would be good to create error text that includes _exactly_ why we're skipping the script.
Like the current status/stage if that's the reason or the mac we're searching for if we couldn't find the host.

I'm imagining a scenario where a user isn't getting the script served to them and doesn't know why. The easiest way to debug this would be to have this information both in the log and the response text.

internal/bminventory/inventory_v2_handlers.go
ori-amizur: Don't know.  Who verifies it ?

ori-amizur: What token key should be used ?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -605,13 +606,68 @@ func (b *bareMetalInventory) signURL(ctx context.Context, infraEnvID, urlString,
...
+	macStr := mac.String()
+	if err := b.db.Where("infra_env_id = ? and (inventory like ? or inventory like ?)", infraEnv.ID.String(), fmt.Sprintf("%%%s%%", strings.ToUpper(macStr)),
+		fmt.Sprintf("%%%s%%", strings.ToLower(macStr))).Find(&hosts).Error; err != nil {
+		return common.NewApiError(http.StatusInternalServerError, errors.Wrapf(err, "IPXE booting skipped. InfraEnv %s: Host with mac %s", infraEnv.ID.String(), macStr))
+	}
+	switch len(hosts) {
+	case 0:
+		return nil
+	case 1:
+		break
```
carbonin, line 650: I don't think we need this `break`

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -605,13 +606,68 @@ func (b *bareMetalInventory) signURL(ctx context.Context, infraEnvID, urlString,
...
+	default:
+		return common.NewApiError(http.StatusInternalServerError, errors.Errorf("IPXE booting skipped. Unexpected number of hosts %d with mac %s", len(hosts), macStr))
```
carbonin, line 652: Can we add a test for when more than one host matches?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -638,7 +694,18 @@ func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *c
 		return "", errors.Wrap(err, "failed to sign initrd URL")
 	}
 
-	return fmt.Sprintf(ipxeScriptFormat, initrdURL, kernelURL, rootfsURL), nil
+	return fmt.Sprintf(ipxeBootScriptFormat, initrdURL, kernelURL, rootfsURL), nil
+}
+
+func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *common.InfraEnv, mac *strfmt.MAC, bootControl bool) (string, error) {
+	if mac != nil && len(*mac) != 0 {
```
carbonin, line 701: Can the second condition be written `*mac != ""`?

internal/bminventory/inventory_test.go
```diff
@@ -9226,18 +9241,101 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
 
 	It("fails to return ipxe-script when openshift version is nil", func() {
 		mockVersions.EXPECT().GetOsImage(common.TestDefaultConfig.OpenShiftVersion, gomock.Any()).Return(&models.OsImage{}, nil).Times(1)
-		params := installer.V2DownloadInfraEnvFilesParams{InfraEnvID: infraEnvID, FileName: "ipxe-script"}
+		params := installer.V2DownloadInfraEnvFilesParams{InfraEnvID: infraEnvID, FileName: "ipxe-script", Mac: toMac("f8:75:a4:a4:00:fe")}
```
carbonin, line 9244: Why add a mac here? I don't think it should matter for this test.

internal/bminventory/inventory_test.go
```diff
@@ -9180,7 +9195,7 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
 
 	It("returns ipxe-script successfully", func() {
 		mockVersions.EXPECT().GetOsImage(common.TestDefaultConfig.OpenShiftVersion, gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).Times(1)
-		content := getResponseData("ipxe-script")
+		content := getResponseData("ipxe-script", true, false)
```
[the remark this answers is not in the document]
carbonin: If we need a test specifically for when the script should be served and a mac address is present we should add a new test for that case.

internal/bminventory/inventory_test.go
carbonin: Looks like this is testing for auth now :+1:

internal/bminventory/inventory_test.go
```diff
@@ -9256,9 +9354,22 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			tok := chainUrl.Query().Get("api_key")
+			_, err = bm.authHandler.AuthURLAuth(tok)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
 		It("signs the initrd ipxe-script url correctly", func() {
 			mockVersions.EXPECT().GetOsImage(common.TestDefaultConfig.OpenShiftVersion, gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).Times(1)
-			content := getResponseData("ipxe-script")
+			content := getResponseData("ipxe-script", true, false)
```
carbonin, line 9372: None of these existing tests should have a mac passed. We need to ensure they work the same as they did before.

carbonin (commented):
Maybe we should also add a subsystem test.
It looks like we already have one for downloading the ipxe script [here](https://github.com/openshift/assisted-service/blob/18e826f97bb72ce7b73f6eb5e8767f22c1902e9d/subsystem/infra_env_test.go#L184-L191)

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -638,18 +693,33 @@ func (b *bareMetalInventory) infraEnvIPXEScript(ctx context.Context, infraEnv *c
...
 	if err != nil {
 		return common.GenerateErrorResponder(err)
 	}
 
 	builder := &installer.V2DownloadInfraEnvFilesURL{
-		InfraEnvID: params.InfraEnvID,
-		FileName:   params.FileName,
+		InfraEnvID:  params.InfraEnvID,
+		FileName:    params.FileName,
+		BootControl: params.BootControl,
```
carbonin, line 722: Will this lead to a query parameter like `boot_control=` with no value?

internal/bminventory/inventory_test.go
```diff
@@ -14038,8 +14246,14 @@ var _ = Describe("GetInfraEnvPresignedFileURL", func() {
 			Expect(u.Query().Get("image_token")).To(Equal(""))
 			Expect(u.Query().Get("api_key")).To(Equal(""))
 			Expect(u.Query().Get("file_name")).To(Equal("discovery.ign"))
+			Expect(u.Query().Get("boot_control")).To(BeEmpty())
 			Expect(u.Path).To(Equal(fmt.Sprintf("/api/assisted-install/v2/infra-envs/%s/downloads/files", infraEnvID.String())))
 		})
+
+		It("generates a url with no token for discovery.ign with boot_control", func() {
```
carbonin, line 14253: This isn't what this test is testing for.
This should be something like `returns bad request when boot_control is used with discovery.ign`

internal/bminventory/inventory_test.go
```diff
@@ -14038,8 +14246,14 @@ var _ = Describe("GetInfraEnvPresignedFileURL", func() {
 			Expect(u.Query().Get("image_token")).To(Equal(""))
 			Expect(u.Query().Get("api_key")).To(Equal(""))
 			Expect(u.Query().Get("file_name")).To(Equal("discovery.ign"))
+			Expect(u.Query().Get("boot_control")).To(BeEmpty())
```
carbonin, line 14249: Can we instead test that there is no `boot_control` key rather than the value being empty?
I just want to make sure the discovery ignition URL doesn't contain that parameter, it would look pretty strange.

subsystem/infra_env_test.go
```diff
@@ -204,6 +206,86 @@ var _ = Describe("Infra_Env", func() {
...
+		res, err := userBMClient.Installer.GetInfraEnvPresignedFileURL(ctx, &installer.GetInfraEnvPresignedFileURLParams{
+			InfraEnvID:  infraEnvID,
+			FileName:    "ipxe-script",
+			BootControl: swag.Bool(true)})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(res.Payload).ToNot(BeNil())
+		url := swag.StringValue(res.Payload.URL)
+		Expect(url).NotTo(BeEmpty())
+
+		By("Serv redirect script")
```
carbonin, line 219: Minor typo should be `Serve`

carbonin (commented):
Looking good, just some minor comments.

internal/bminventory/inventory_v2_handlers.go
ori-amizur: Checking it in unit test

internal/bminventory/inventory_test.go
ori-amizur: Tested

subsystem/infra_env_test.go
```diff
@@ -197,13 +199,94 @@ var _ = Describe("Infra_Env", func() {
 		u := res.Payload.URL
 		Expect(u).NotTo(BeNil())
 
+		Expect(*u).ToNot(ContainSubstring("boot_control"))
```
carbonin, line 202: This seems to be failing the subsystem test.

I'm not sure how important it is to ensure this isn't present in the ipxe-script case though. (my original concern was just about other files at this endpoint).

@avishayt do you think it would be particularly bad for this url to change after an upgrade even though the contents would be the same?

It's not the iso url so we won't cause a host reboot or anything. :man_shrugging:

ori-amizur: Now it doesn't fail

avishayt:
Nice!
1. What is "boot_control"?  Is this something standard?  When is it set?
2. Please update docs/

ori-amizur:
> 1. What is "boot_control"?  Is this something standard?  When is it set?
@avishayt please look at https://issues.redhat.com/browse/MGMT-9982

> 2. Please update docs/
OK

ori-amizur:
> The JIRA mentions it, but doesn't explain the iPXE client part. When does the client set it, and according to what?

If there is a decision to use boot_control, it should be set as part of the URL that is set in the DHCP server.  This is an indication to the assisted service that the serving policy of the iPXE script should depend on the state the host is in.  There are several cases that are handled:

1. The host is not yet in the assisted service.  In this case iPXE script should be served which will cause the host to boot from it.
2. The host is part of the assisted service, but it should still boot from the discovery ISO.  In this case, the iPXE script should be served.
3. The host is in an installation stage after (and including) rebooting stage.  In this case the iPXE script should not be served, which will cause the host to boot from HD.

The host is recognized by the mac address. The iPXE macro ${net0/mac} is replaced by the mac address.

avishayt:
OK I understand now, thanks!

avishayt:
The code looks OK to me.  Waiting on Nick's review and docs.

ori-amizur:
@avishayt added to existing doc.

PR #4014, opened 2022-06-23
AGENT-227: Add host configuration support to ephemeral installer client

zaneb:
* [AGENT-238](https://issues.redhat.com/browse/AGENT-238) Add subcommands to the ephemeral installer client
* [AGENT-227](https://issues.redhat.com/browse/AGENT-227) Add support for configuring the install disk based on root device hints
* [AGENT-229](https://issues.redhat.com/browse/AGENT-229) Add support for configuring the role for each host

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] Ephemeral installer
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @flaper87
/cc @lranjbar

## Checklist

## Reviewers Checklist


mkowalski:
I reviewed the part covering configuring the install disk based on root device hints and it looks okay for me. Unfortunately at this moment I don't have enough knowledge about the rest of the code touched by this PR to safely LGTM it as a whole

celebdor:
This PR should get /lgtm and /approval from approvers in ZTP side (due to the rootdevicehints). Let's put a hold until we have /lgtm from agent side and /lgtm from ZTP side and /approve from somebody on the approvers list.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -882,85 +883,37 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
...
+			Smart:                   dev.Smart,
+			Vendor:                  dev.Vendor,
+			Wwn:                     dev.Wwn,
 		}
+	}
 
-		return disk.ID
+	acceptable := hostutil.GetAcceptableDisksWithHints(disks, hints)
+	if len(acceptable) > 0 {
+		return acceptable[0].ID
```
paul-maidment, line 916: Seems reasonable to pick the first acceptable disk though I do question whether or not there might need to be consideration of which of the acceptable is "best"

Are there any other priorities under which we would like to select the acceptable disk, is the first always the best choice?
Should the list of acceptable disks be sorted any way to place the "most acceptable" first?

internal/host/hostutil/host_utils.go
```diff
@@ -151,6 +152,85 @@ func GetDiskByInstallationPath(disks []*models.Disk, installationPath string) *m
...
+//
+// The logic is quite straightforward and the checks done match what is in the aforementioned link.
+// Some string checks require equality, others partial equality, whereas the int checks require numeric comparison.
+//
+// Ironic's internal filter process requires that all the disks have to fully match the RootDeviceHints (and operation),
+// which is what this function does.
+//
+// This function also filters out disks that are not elegible for installation, as we already know those cannot be used.
+func GetAcceptableDisksWithHints(disks []*models.Disk, hints *bmh_v1alpha1.RootDeviceHints) []*models.Disk {
+	if hints == nil {
```
paul-maidment, line 171: As it stands in the current implementation, hints will never be nil due to the check in `findInstallationDiskID` is the desired behaviour that findInstallationDiskID in the absence of hints should return an arbitrary disk ID or should it return an empty ID.

If the former, then should the check be removed from `findInstallationDiskID` and if the latter should this check be removed here?

paul-maidment: Actually, noted that a few lines below there is an eligibility check for the disk, which will only run if this condition is not triggered. As it stands, this function could return ineligible disks in the collection if there are no hints specified.

zaneb: > Is the desired behaviour that findInstallationDiskID in the absence of hints should return an arbitrary disk ID or should it return an empty ID?

For the existing BMAC use, it should return nothing. For the new use in this PR, it's easier if it returns the list of all disks. Hence the checks in both places.

> Actually, noted that a few lines below there is an eligibility check for the disk, which will only run if this condition is not triggered. As it stands, this function could return ineligible disks in the collection if there are no hints specified.

That's a good point, thanks.

paul-maidment (approved):
/lgtm

flaper87:
/lgtm
/approve

The rootdevicehints piece looks sane!

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -882,85 +883,37 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
...
-				continue
-			}
+	disks := make([]*models.Disk, len(devices))
+
+	for i, dev := range devices {
+		disks[i] = &models.Disk{
+			Bootable:                dev.Bootable,
+			ByID:                    dev.ByID,
+			ByPath:                  dev.ByPath,
+			DriveType:               models.DriveType(strings.ToUpper(dev.DriveType)),
```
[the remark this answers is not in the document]
zaneb: I had that originally, but it seemed wrong to cast a string that was possibly in the wrong case to a `DriveType`. Didn't notice the multipath one though tbh. In practice we're only checking for `"HDD"` so it will work. Not sure what the least wrong thing is.

PR #4018, opened 2022-06-24
 MGMT-10790: Always use APIVipDNSName to obtain API address

paul-maidment:
Presently we fetch the API address for ignition download and validation
using two independent pieces of code, these also prioritise cluster.APIVip and
cluster.APIVipDNSName in different ways, which leads to incorrect validation.

This commit addresses that by extracting this logic into a function that
may be called from both locations.

During the development of this feature, it was decided that the only
valid address during installation is cluster.APIVipDNSName and so this
function has been created to only return that.

At least for future modification, this logic is in a single location.

Additionally, the validation of whether or not an ignition is downloadable for Day2 hosts was skipped for hosts with user managed networking enabled. We have removed this so that the validation will be conducted for all day 2 hosts now.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -779,13 +779,7 @@ func (b *bareMetalInventory) createAndUploadDay2NodeIgnition(ctx context.Context
 	log := logutil.FromContext(ctx, b.log)
 	log.Infof("Starting createAndUploadDay2NodeIgnition for cluster %s, host %s", cluster.ID, host.ID)
 
-	// Specify ignition endpoint based on cluster configuration:
-	address := cluster.APIVip
-	if address == "" {
-		address = swag.StringValue(cluster.APIVipDNSName)
-	}
-
-	ignitionEndpointUrl := fmt.Sprintf("http://%s:22624/config/%s", address, host.MachineConfigPoolName)
+	ignitionEndpointUrl := fmt.Sprintf("http://%s:22624/config/%s", common.GetAPIHostname(cluster), host.MachineConfigPoolName)
 	if cluster.IgnitionEndpoint != nil && cluster.IgnitionEndpoint.URL != nil {
 		url, err := url.Parse(*cluster.IgnitionEndpoint.URL)
 		if err != nil {
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -392,3 +392,16 @@ func GetTagFromImageRef(ref string) string {
 		return ""
 	}
 }
+
+func GetAPIHostname(c *Cluster) string {
+	// Despite the confusing name of this parameter, in day-2 scenarios where
+	// this function is used it could either be a DNS domain name that points
+	// at the API's IP address (this is the default) or it could also not be a
+	// DNS domain name at all - such as when the user chooses to override this
+	// parameter with an IP address. Such override is commonly done by users in
+	// day-2 SaaS imported clusters where the user never bothered to set up DNS
+	// for their day-1 cluster in the first place and just wants the worker to
+	// connect to the API directly. The UI even has a special dialog to help users
+	// do that.
+	return swag.StringValue(c.APIVipDNSName)
+}
--- a/internal/host/hostcommands/api_vip_connectivity_check_cmd.go
+++ b/internal/host/hostcommands/api_vip_connectivity_check_cmd.go
@@ -7,7 +7,6 @@ import (
 	"net/url"
 	"path"
 
-	"github.com/go-openapi/swag"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/models"
 	"github.com/sirupsen/logrus"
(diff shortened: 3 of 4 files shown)


tsorya:
@paul-maidment please right description of the PR , honestly idk what we are fixing here

romfreiman:
/hold
I think we should reverse the order. First API then DNS. Why did you choose the opposite order?

internal/common/common.go
```diff
@@ -371,3 +371,12 @@ func CanonizeStrings(slice []string) (ret []string) {
 func GetHostKey(host *models.Host) string {
 	return host.ID.String() + "@" + host.InfraEnvID.String()
 }
+
+// GetAPIHost returns either the domain name or IP address for the cluster.
+func GetAPIHostname(c *Cluster) string {
+	host := swag.StringValue(c.APIVipDNSName)
+	if host == "" {
```
tsorya, line 378: i think order is not right. If we have vip better to use it

tsorya: i am actually mistaken, we should even fallback to api vip.

omertuc:
/hold

Please see JIRA [comment](https://issues.redhat.com/browse/MGMT-10790?focusedCommentId=20524991&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20524991)

internal/common/common.go
```diff
@@ -392,3 +392,7 @@ func GetTagFromImageRef(ref string) string {
 		return ""
 	}
 }
+
+func GetAPIHostname(c *Cluster) string {
+	return swag.StringValue(c.APIVipDNSName)
```
omertuc, line 397:
```suggestion
    // Despite the confusing name of this parameter, in day-2 scenarios where
    // this function is used it could either be a DNS domain name that points
    // at the API's IP address (this is the default) or it could also not be a
    // DNS domain name at all - such as when the user chooses to override this
    // parameter with an IP address. Such override is commonly done by users in
    // day-2 SaaS imported clusters where the user never bothered to set up DNS
    // for their day-1 cluster in the first place and just wants the worker to
    // connect to the API directly. The UI even has a special dialog to help users
    // do that.
    return swag.StringValue(c.APIVipDNSName)
```

internal/bminventory/inventory.go
```diff
@@ -780,12 +780,9 @@ func (b *bareMetalInventory) createAndUploadDay2NodeIgnition(ctx context.Context
 	log.Infof("Starting createAndUploadDay2NodeIgnition for cluster %s, host %s", cluster.ID, host.ID)
 
 	// Specify ignition endpoint based on cluster configuration:
-	address := cluster.APIVip
-	if address == "" {
-		address = swag.StringValue(cluster.APIVipDNSName)
-	}
+	apiHost := common.GetAPIHostname(cluster)
```
omertuc, line 783: unnecessary variable

internal/host/hostcommands/api_vip_connectivity_check_cmd.go
```diff
@@ -73,15 +72,12 @@ func (c *apivipConnectivityCheckCmd) GetSteps(ctx context.Context, host *models.
 }
 
 func getIngnitionEndPoint(cluster *common.Cluster, host *models.Host) (string, error) {
-	addressPart := swag.StringValue(cluster.APIVipDNSName)
-	if addressPart == "" {
-		addressPart = cluster.APIVip
-	}
+	apiHost := common.GetAPIHostname(cluster)
```
omertuc, line 75: unnecessary variable

paul-maidment:
> The title of the PR doesn't really apply any more as well

I'd say that the title is still representative of what is happening, we want to ensure that when obtaining the address of the API for ignition or validation that this is done in the same way (i.e. consistently) for both use cases.

with some added detail in the description about why we got rid of the "prioritisation" code and chose to return ApiVipDNS instead

does that make sense or did you think the title should be different?

omertuc:
> > The title of the PR doesn't really apply any more as well
> 
> I'd say that the title is still representative of what is happening, we want to ensure that when obtaining the address of the API for ignition or validation that this is done in the same way (i.e. consistently) for both use cases.
> 
> with some added detail in the description about why we got rid of the "prioritisation" code and chose to return ApiVipDNS instead
> 
> does that make sense or did you think the title should be different?

While it's technically correct in "achieving consistency" I feel like achieving consistency is a bit beside the point now. The point is that we now simply always use apivipdns everywhere, never the apivip. But that's just my subjective opinion, either way is fine, up to you

omertuc:
Commit doesn't mention the second part of the Jira ticket which this PR addresses

omertuc:
/hold

Let's run some heavy tests

PR #4041, opened 2022-06-29
MGMT-10734: Add host states and transitions for reclaim

carbonin:
Adds the statemachine transitions and new states in order to implement
the following diagram:

```
                   -----------------------
                   | HostStatusInstalled |
                   -----------------------
                             |
                      User unbinds agent
                             |
     Kube API and no BMH     |   REST API, Kube API with BMH, or other machine management
          -------------------+-------------------
          |                                     |
          V                                     V
-----------------------------            ----------------------------
| TransitionTypeReclaimHost |            | TransitionTypeUnbindHost |
-----------------------------            ----------------------------
            |                                               |
            V                                               |
  ------------------------                                  |
  | HostStatusReclaiming |                                  |
  ------------------------                                  |
                |                                           |
       Agent posts step reply                               |
                |                                           |
       ---------+----------------------------------         |
       |                                          |         |
    Success                                    Failure      |
       |                                          |         |
       V                                          |         |
-------------------------------------             |         |
| HostStatusReclaimingPendingReboot |------       |         |
-------------------------------------     |       |         |
                |                         |       |         |
  ------------------------------       Timeout    |         |
  | TransitionTypeRegisterHost |          |       |         |
  ------------------------------          |       |         |
                |                         |       |         |
                V                         V       V         V
--------------------------------  ----------------------------------------
| HostStatusDiscoveringUnbound |  | HostStatusUnbindingPendingUserAction |
--------------------------------  ----------------------------------------
              ^                                       |
              |       ------------------------------  |
              --------| TransitionTypeRegisterHost |---
                      ------------------------------

```

Specifically I'm looking for input on the names of these states in this patch. This doesn't actually make any of the transitions happen yet.

Related to the enhancement in https://github.com/openshift/assisted-service/pull/3849
Part 1 of https://issues.redhat.com/browse/MGMT-10734

The next PR will add the code to push the hosts through these new states when required.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

This code is just adding states and transitions, no functional changes were added.

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @avishayt 
/cc @filanov 
/cc @nmagnezi 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/pool_host_statemachine.go
+++ b/internal/host/pool_host_statemachine.go
@@ -27,6 +27,7 @@ func NewPoolHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler)
 			stateswitch.State(models.HostStatusKnownUnbound),
 			stateswitch.State(models.HostStatusUnbinding),
 			stateswitch.State(models.HostStatusUnbindingPendingUserAction),
+			stateswitch.State(models.HostStatusReclaimingRebooting),
 		},
 		DestinationState: stateswitch.State(models.HostStatusDiscoveringUnbound),
 		PostTransition:   th.PostRegisterHost,
--- a/api/vendor/github.com/openshift/assisted-service/models/host.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/host.go
@@ -151,7 +151,7 @@ type Host struct {
 
 	// status
 	// Required: true
-	// Enum: [discovering known disconnected insufficient disabled preparing-for-installation preparing-failed preparing-successful pending-for-input installing installing-in-progress installing-pending-user-action resetting-pending-user-action installed error resetting added-to-existing-cluster cancelled binding unbinding unbinding-pending-user-action known-unbound disconnected-unbound insufficient-unbound disabled-unbound discovering-unbound]
+	// Enum: [discovering known disconnected insufficient disabled preparing-for-installation preparing-failed preparing-successful pending-for-input installing installing-in-progress installing-pending-user-action resetting-pending-user-action installed error resetting added-to-existing-cluster cancelled binding unbinding unbinding-pending-user-action known-unbound disconnected-unbound insufficient-unbound disabled-unbound discovering-unbound reclaiming reclaiming-rebooting]
 	Status *string `json:"status"`
 
 	// status info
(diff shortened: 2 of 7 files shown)


avishayt:
1. What is HostStatusReclaimingPendingReboot?  Do we move there just before the reboot to understand the progress?
2. For Kube API with BMH, do we want to skip the generic method as in the diagram, or keep the same flow regardless and use BMH as a fallback?

carbonin:
>What is HostStatusReclaimingPendingReboot? Do we move there just before the reboot to understand the progress?

Effectively, yeah.
I wanted to split the two so that we could get a step response after things like the image downloads and writing out the boot entry. Then `HostStatusReclaimingPendingReboot` will tell the agent to just reboot and we'll progress when the agent registers.

internal/host/pool_host_statemachine.go
```diff
@@ -27,6 +27,7 @@ func NewPoolHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler)
 			stateswitch.State(models.HostStatusKnownUnbound),
 			stateswitch.State(models.HostStatusUnbinding),
 			stateswitch.State(models.HostStatusUnbindingPendingUserAction),
+			stateswitch.State(models.HostStatusReclaimingPendingReboot),
```
avishayt, line 30: PendingReboot sounds like the user needs to do it
```suggestion
			stateswitch.State(models.HostStatusReclaimingRebooting),
```

avishayt:
> I'm not sure I understand. If we have the BMH I'd prefer we use that and not try to start an agent on the node.
> We could probably improve the way it's done currently, but I think it's out of scope for this feature.

1. Originally when we were thinking about this feature we discussed all kinds of shortcuts like kexec, but now that we need to reboot and this isn't necessarily more reliable than BMH, we can skip it for the BMH flow.
2. We need to think about how to detect the BMH flow.  Currently, the only controller that deals with BMH is BMAC and I wouldn't want that logic to creep into the rest of Assisted.

carbonin:
> now that we need to reboot and this isn't necessarily more reliable than BMH, we can skip it for the BMH flow.

Yeah, I think BMH will handle more situations than this will so it's better to default to using that if it's present.

> We need to think about how to detect the BMH flow. Currently, the only controller that deals with BMH is BMAC and I wouldn't want that logic to creep into the rest of Assisted.

This is one of the reasons that we can assume this feature will only be available in the kube API.
Given that it's only available in the kube API I don't see why the agent controller wouldn't be able to search for a BMH on unbind by sharing the matching logic with BMAC (or by just using a label selector).
Hopefully that doesn't qualify as creeping into the rest of assisted.

carbonin:
Yeah, right now the plan is just to indicate implicitly (linked to a bmh) that we shouldn't reclaim the agent. We could add something explicit as well, but I'm not sure it's relevant for this PR, right?

avishayt:
> Yeah, right now the plan is just to indicate implicitly (linked to a bmh) that we shouldn't reclaim the agent. We could add something explicit as well, but I'm not sure it's relevant for this PR, right?

Not relevant for this PR, just noting it.

PR #4044, opened 2022-06-30
NO-ISSUE: Use `go:embed` for ignition templates

jhernand:
Currently the ignition files are generated from Go templates that are
defined in constants inside the Go code. Those templates also contain
the unit files for the agent and for other systemd units. Changing those
unit files is uncomfortable, because they need to be manually encoded as
a long JSON string. To make those ignition and systemd files easier to
change this patch moves them to external files, using the `embed`
mechanism introduced in Go 1.16.

This patch also introduces to new template functions named
`executeTemplate` and `toJson` that simplify including templates from
other templates. For example, the `agent.service` template is included
from the `discovery.ign` file with this code:

```
"contents": {{ executeTemplate "agent.service" . | toJson }}
```

## List all the issues related to this PR

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

Manually verified that the ignition files included in the generated .iso files are exactly the same.

## Assignees

/cc @
/cc @

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #4052, opened 2022-06-30
NO-ISSUE: Use template function go generate data URL

jhernand:
Currently some of the files included in the ignition configurations are
generated with opaque data URLs like this:

```
{
  "overwrite": true,
  "path": "/etc/multipath.conf",
  "mode": 420,
  "user": {
    "name": "root"
  },
  "contents": { "source": "data:text/plain;charset=utf-8;base64,ZGVmYXVsdHMgewogICAgdXNlcl9mcmllbmRseV9uYW1lcyB5ZXMKICAgIGZpbmRfbXVsdGlwYXRocyB5ZXMKICAgIGVuYWJsZV9mb3JlaWduICJeJCIKfQpibGFja2xpc3RfZXhjZXB0aW9ucyB7CiAgICBwcm9wZXJ0eSAiKFNDU0lfSURFTlRffElEX1dXTikiCn0KYmxhY2tsaXN0IHsKfQo=" }
}
```

That is difficult to change and to review, because the contents of the
file aren't immediately visible. To improve that this patch adds an new
`toBase64` template function that combined with the `executeTemplate`
template function introduced in a previous patch make it possible to
write those files in plain text, and then include it in the ignition
file like this:

```
{
  "overwrite": true,
  "path": "/etc/multipath.conf",
  "mode": 420,
  "user": {
    "name": "root"
  },
  "contents": { "source": "data:text/plain;charset=utf-8;base64,{{ executeTemplate "multipath.conf" . | toBase64 }}" }
}
```

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/ignition/templates.go
+++ b/internal/ignition/templates.go
@@ -3,6 +3,7 @@ package ignition
 import (
 	"bytes"
 	"embed"
+	"encoding/base64"
 	"encoding/json"
 	"text/template"
 )
@@ -12,12 +13,13 @@ var templatesFS embed.FS
 
 // loadTemplates loads the templates from the embedded 'templates' directory.
 //
-// In addition to the default functions the templates will also have available the 'executeTemplate'
-// and 'toJson' functions.
+// In addition to the default functions the templates will also have available the
+// 'executeTemplate', 'toJson' and 'toBase64' functions.
 func loadTemplates() (result *template.Template, err error) {
 	initial := template.New("")
 	initial.Funcs(template.FuncMap{
 		"executeTemplate": makeExecuteTemplateFunc(initial),
+		"toBase64":        toBase64Func,
 		"toJson":          toJsonFunc,
 	})
 	_, err = initial.ParseFS(templatesFS, "templates/*")
@@ -81,4 +83,9 @@ func toJsonFunc(data interface{}) (result string, err error) {
 	}
 	result = string(text)
 	return
+}
+
+// toBase64 is a template function that encodes the given data using Base64.
+func toBase64Func(data string) string {
+	return base64.StdEncoding.EncodeToString([]byte(data))
 }
--- a/internal/ignition/templates/discovery.ign
+++ b/internal/ignition/templates/discovery.ign
@@ -66,7 +66,7 @@
       "user": {
           "name": "root"
       },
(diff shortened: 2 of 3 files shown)


jhernand:
Note that this is an additional commit on top of #4044.

eranco74:
Looks good, just a nit
data url doesn't have to be base64 encoded, consider renaming `toDataUrl` to reflect that it's encoded

jhernand:
@eranco74 That makes sense. However note that the content type and the character encoding (and maybe other parameters) are also configurable in a data URL. If we wanted to reflect all that in the name of the template function it would need to have a very long name, or many parameters. Considering that I believe it is better to write that initial part of the data URL directly in the template, and then use a `toBase64` template function to generate the rest. Something like this:

```
"source": "data:text/plain;charset=utf-8;base64,{{ executeTemplate "multipath.conf" . | toBase64 }}"
```

I will do that in the next version of the patch.

jhernand:
Rebased and replaced `toDataUrl` with `toBase64`.

jhernand:
@eranco74 @omertuc any other suggestions before merging this?

internal/ignition/templates.go
```diff
@@ -82,3 +84,8 @@ func toJsonFunc(data interface{}) (result string, err error) {
 	result = string(text)
 	return
 }
+
+// toBase64 is a template function that encodes the given data using Base64.
+func toBase64Func(data string) string {
+	return base64.StdEncoding.EncodeToString([]byte(data))
```
omertuc, line 90: I'm not a fan of `makeExecuteTemplateFunc` calling `buffer.String()` because it assumes the file we wish to embed is a utf-8 encoded text file, which limits us to only store textual files. Ideally makeExecuteTemplateFunc would return a byte array, and toBase64Func would receive a byte array, and so the []byte(data) conversion on this line would not be needed

omertuc: No concrete need for binary files at the moment, just striving to do what's right without unnecessary assuming encoding where it's not needed.

jhernand:
@eranco74 do you have any objection to unhold and therefore merge this?

jhernand:
@omertuc the change to make the `executeTemplate` return an array of bytes instead of a string is here: https://github.com/openshift/assisted-service/pull/4126.

@omertuc, @eranco74 can we merge this and continue the discussion there?

jhernand:
@eranco74 a kind reminder that this is waiting for you to unhold.

PR #4057, opened 2022-07-01
NO-ISSUE: Improve error reporting in ephemeral installer client

zaneb:
The errors returned from the API are not especially human-readable, they
print something like:

    [POST /v2/clusters][400] v2RegisterClusterBadRequest  &{Code:0xc00052e790 Href:0xc00052e7a0 ID:0xc00053ed84 Kind:0xc00052e7b0 Reason:0xc00052e7c0}

Passing through GetAssistedError() returns an error that is readable in
the logs, including describing the reason for the failure.

### DIFF
--- a/cmd/agentbasedinstaller/register.go
+++ b/cmd/agentbasedinstaller/register.go
@@ -13,6 +13,7 @@ import (
 	"github.com/openshift/assisted-service/internal/controller/controllers"
 	"github.com/openshift/assisted-service/internal/oc"
 	"github.com/openshift/assisted-service/models"
+	errorutil "github.com/openshift/assisted-service/pkg/error"
 	"github.com/openshift/assisted-service/pkg/executer"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


cmd/agentbasedinstaller/register.go
```diff
@@ -49,7 +50,7 @@ func RegisterCluster(ctx context.Context, log *log.Logger, bmInventory *client.A
 	}
 	clusterResult, registerClusterErr := bmInventory.Installer.V2RegisterCluster(ctx, clientClusterParams)
 	if registerClusterErr != nil {
-		return nil, registerClusterErr
+		return nil, errorutil.GetAssistedError(registerClusterErr)
```
rwsu, line 53: I did some testing and I am not seeing the improved error reporting. I think errorutil.GetAssistedError is not returning the detailed error message because the errors type check is flowing into the default case here: https://github.com/openshift/assisted-service/blob/master/pkg/error/error_utils.go#L44

zaneb: It looks like you ran into a local error and thus didn't get the response from the server:
https://github.com/openshift/assisted-service/blob/master/client/installer/v2_register_cluster_responses.go#L63

rwsu: I was able to trigger a non-local error and now seeing the extra information in the error message. The trigger was to deploy a SNO cluster with API_VIP and INGRESS_VIP defined in the AgentClusterInstall.

`Jul 07 05:30:13 master-0.ostest.test.metalkube.org create-cluster-and-infraenv[2668]: time="2022-07-07T05:30:13Z" level=fatal msg="Failed to register cluster with assisted-service: AssistedServiceError Code: 400 Href:  ID: 400 Kind: Error Reason: Ingress VIP cannot be set with User Managed Networking"`

rwsu (approved):
/lgtm

PR #4061, opened 2022-07-03
MGMT-11000: Failed to update the cluster when select Integrate with vsphere

michaellevy101:
Getting "no condition passed to run transition RefreshHost from state pending-for-input"
Updating it to be part of sufficient for install checks will resolve this issue

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/statemachine.go
+++ b/internal/host/statemachine.go
@@ -536,7 +536,7 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 	})
 
 	var hasMinRequiredHardware = stateswitch.And(If(HasMinValidDisks), If(HasMinCPUCores), If(HasMinMemory),
-		If(CompatibleWithClusterPlatform), If(DiskEncryptionRequirementsSatisfied), If(VSphereHostUUIDEnabled))
+		If(CompatibleWithClusterPlatform), If(DiskEncryptionRequirementsSatisfied))
 
 	var requiredInputFieldsExist = stateswitch.And(If(IsMachineCidrDefined))
 
@@ -572,7 +572,7 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 			stateswitch.State(models.HostStatusKnown),
 			stateswitch.State(models.HostStatusPendingForInput),
 		},
-		Condition: stateswitch.And(If(IsConnected), If(IsMediaConnected), If(HasInventory),
+		Condition: stateswitch.And(If(IsConnected), If(IsMediaConnected), If(HasInventory), If(VSphereHostUUIDEnabled),
 			hasMinRequiredHardware,
 			stateswitch.Not(requiredInputFieldsExist)),
 		DestinationState: stateswitch.State(models.HostStatusPendingForInput),
@@ -593,8 +593,9 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 		},
 		Condition: stateswitch.And(If(IsConnected), If(IsMediaConnected), If(HasInventory),
 			hasMinRequiredHardware,
-			requiredInputFieldsExist,
-			stateswitch.Not(isSufficientForInstall)),
+			stateswitch.Or(stateswitch.Not(If(VSphereHostUUIDEnabled)),
+				stateswitch.And(requiredInputFieldsExist, stateswitch.Not(isSufficientForInstall)),
+			)),
 		DestinationState: stateswitch.State(models.HostStatusInsufficient),
 		PostTransition:   th.PostRefreshHost(statusInfoNotReadyForInstall),
 	})
@@ -611,7 +612,9 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 		Condition: stateswitch.And(If(IsConnected), If(IsMediaConnected), If(HasInventory),
 			hasMinRequiredHardware,
 			requiredInputFieldsExist,
-			isSufficientForInstall),
+			isSufficientForInstall,
+			If(VSphereHostUUIDEnabled),
+		),
 		DestinationState: stateswitch.State(models.HostStatusKnown),
 		PostTransition:   th.PostRefreshHost(statusInfoKnown),
 	})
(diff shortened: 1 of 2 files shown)


gamli75:
Please retitle to: https://issues.redhat.com/browse/MGMT-11000 an add tests

gamli75:
there are lot of testing in the transition test and I'm not sure where we should add a new test but maybe add something like https://github.com/openshift/assisted-service/blob/master/internal/host/transition_test.go#L2851

michaellevy101:
> there are lot of testing in the transition test and I'm not sure where we should add a new test but maybe add something like https://github.com/openshift/assisted-service/blob/master/internal/host/transition_test.go#L2851

have the same test

filanov (commented):
Is `VSphereHostUUIDEnabled` something that can be fixed by the user?

internal/host/statemachine.go
```diff
@@ -593,8 +593,9 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 		},
 		Condition: stateswitch.And(If(IsConnected), If(IsMediaConnected), If(HasInventory),
 			hasMinRequiredHardware,
-			requiredInputFieldsExist,
-			stateswitch.Not(isSufficientForInstall)),
+			stateswitch.Or(stateswitch.Not(If(VSphereHostUUIDEnabled)),
```
filanov, line 596: To see if i understand correctly, this is the only transition that actually change? 
Why the or include `requiredInputFieldsExist`? shouldn't it be only `stateswitch.Not(If(VSphereHostUUIDEnabled)` and ` stateswitch.Not(isSufficientForInstall)`?

michaellevy101: if requiredInputFieldsExist has failed (missing machine CIDR) the status should be pending for input unless VSphereHostUUIDEnabled check failed (the status should be insufficient)

michaellevy101:
> Is `VSphereHostUUIDEnabled` something that can be fixed by the user?

yes, update the VM and reboot the machine or turn vsphere integration off

filanov:
> > Is `VSphereHostUUIDEnabled` something that can be fixed by the user?
> 
> yes, update the VM and reboot the machine or turn vsphere integration off

filanov:
> > Is `VSphereHostUUIDEnabled` something that can be fixed by the user?
> 
> yes, update the VM and reboot the machine or turn vsphere integration off

Then maybe it should be pending input instead of insufficient? 
@avishayt what do you think?

avishayt:
> Then maybe it should be pending input instead of insufficient?

I don't think so, I would consider it a (virtual) hardware property, like the disk's type or the VM's memory or CPU which can also be fixed by modifying the VM and rebooting.  Pending input should be something that can be fixed via our API.

internal/host/statemachine.go
ori-amizur: The second `requiredInputFieldsExist`  (In the last `And` condition), should be removed because it is always true

ori-amizur: Why not add `VSphereHostUUIDEnabled` to `isSufficientForInstall` and leave the rules as before?

filanov: How about adding `HostStatusPendingForInput` to  https://github.com/openshift/assisted-service/blob/f40267245a6bc7e170b4e6acb368b7c41d8f0397/internal/host/statemachine.go#L550-L562 and returning `VSphereHostUUIDEnabled` to be part of `hasMinRequiredHardware` ?

filanov: or moving it to `isSufficientForInstall` as Ori suggested

michaellevy101:
@ori-amizur not sure that i understand your concerns. 
If we want to change the status to insufficient even when the machine CIDR is empty the vsphere uuid validation cannot be part of to the isSufficientForInstall.

filanov (commented):
Is there a subsystem test for vsphere flows? 
It looks like there are flows that can be related to the order that things are defined in

internal/host/statemachine.go
```diff
@@ -572,7 +572,7 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 			stateswitch.State(models.HostStatusKnown),
 			stateswitch.State(models.HostStatusPendingForInput),
 		},
-		Condition: stateswitch.And(If(IsConnected), If(IsMediaConnected), If(HasInventory),
+		Condition: stateswitch.And(If(IsConnected), If(IsMediaConnected), If(HasInventory), If(VSphereHostUUIDEnabled),
```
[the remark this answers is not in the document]
michaellevy101: if VSphereHostUUIDEnabled is false the status should be Insufficient instead of PendingForInput

internal/host/statemachine.go
michaellevy101: if Vsphere validation will be part of isSufficientForInstall and the cluster doesn't have machine cidr the status will be "Pending for input" instead of "insufficient".
Adding it to hasMinRequiredHardware (as i did before this PR) prevent us from integration with vsphere (and updating the host in general) since there is no transition for  "Pending for input" and hasMinRequiredHardware validation failed.
The ultimate solution is to add the vsphere condition to hasMinRequiredHardware  and add "pending for input" to the status list of the following transition.

```
sm.AddTransition(stateswitch.TransitionRule{
		TransitionType: TransitionTypeRefresh,
		SourceStates: []stateswitch.State{
			stateswitch.State(models.HostStatusDisconnected),
			stateswitch.State(models.HostStatusDiscovering),
			stateswitch.State(models.HostStatusInsufficient),
			stateswitch.State(models.HostStatusKnown),
		},
		Condition: stateswitch.And(If(IsConnected), If(IsMediaConnected), If(HasInventory),
			stateswitch.Not(hasMinRequiredHardware)),
		DestinationState: stateswitch.State(models.HostStatusInsufficient),
		PostTransition:   th.PostRefreshHost(statusInfoInsufficientHardware),
	})
```

michaellevy101: Implementing the ultimate solution is risky for production.
As a short term solution i covered the vsphere case only.
Afterward we should implement the ultimate solution that will resolve any modification in Pending for input status.

filanov: Did you try to test https://github.com/openshift/assisted-service/pull/4061/files#r913489952? 
Is any of the existing tests breaks? does a regression pass?

gamli75 (approved):

PR #4062, opened 2022-07-03
Change default ISO image type in cloud envs

osherdp:
PSI installations are still using ``full-iso`` as the default image type. Seems like we only changed the default in the template, which doesn't apply to all installation types.

This changes the default when no ``ISO_IMAGE_TYPE`` has been given in configmap. Operator should remain installing the operand with ``minimal-iso`` as required.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @lalon4 
/cc @gamli75 

## Checklist

## Reviewers Checklist


osherdp:
/hold
inspecting ZTP installations, to make sure we're not making any change there

danielerez:
@filanov Are we ready to change the default value? or we still want to see the effect of the change in SaaS?

gamli75:
> @filanov Are we ready to change the default value? or we still want to see the effect of the change in SaaS?

If this change will only affect the cloud environment - no need to hold it

danielerez:
> > @filanov Are we ready to change the default value? or we still want to see the effect of the change in SaaS?
> 
> If this change will only affect the cloud environment - no need to hold it

This PR would would affect also onprem as it changes the default value of the config directly. I.e. it changes the default value for all envs.

osherdp:
> > > @filanov Are we ready to change the default value? or we still want to see the effect of the change in SaaS?
> > 
> > 
> > If this change will only affect the cloud environment - no need to hold it
> 
> This PR would would affect also onprem as it changes the default value of the config directly. I.e. it changes the default value for all envs.

That's right
I think it deserves adding support of it in configmap, and changing it only for PSI

filanov:
> > > > @filanov Are we ready to change the default value? or we still want to see the effect of the change in SaaS?
> > > 
> > > 
> > > If this change will only affect the cloud environment - no need to hold it
> > 
> > 
> > This PR would would affect also onprem as it changes the default value of the config directly. I.e. it changes the default value for all envs.
> 
> That's right I think it deserves adding support of it in configmap, and changing it only for PSI

I changed it only for cloud so that we can get the statistics and then if it actually resolve media disconnections issues we can change it to be minimal by default as you did. 
Suggesting to hold until we have the data from the cloud.

osherdp:
> I changed it only for cloud so that we can get the statistics and then if it actually resolve media disconnections issues we can change it to be minimal by default as you did.
> Suggesting to hold until we have the data from the cloud.

I changed the scope to just enable us customizing it wherever we want. Please take another look

osherdp:
> ISO_IMAGE_TYPE=minimal-iso

yes, you're right
thank you!

osherdp:
/unhold
late-binding job is hightly flaky, and either way it doesn't seem to affect it

PR #4068, opened 2022-07-04
MGMT-10921: Ensure that API DNS validation checks for resolution to a master.

paul-maidment:
This commit adds a check to ensure that when validating the DNS resolution of the API address that none of these addresses are matched to a current master as this is the incorrect API address.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -1276,7 +1276,7 @@ func (v *validator) isAPIInternalDomainNameResolvedCorrectly(c *validationContex
 
 func (v *validator) printIsAPIInternalDomainNameResolvedCorrectly(c *validationContext, status ValidationStatus) string {
 	apiInternalDomainName := domainNameToResolve(c, constants.APIInternalName)
-	return printIsDomainNameResolvedCorrectly(c, status, apiInternalDomainName, "API load balancer")
+	return printIsDomainNameResolvedCorrectly(c, status, apiInternalDomainName, "internal API load balancer")
 }
 
 func (v *validator) isAppsDomainNameResolvedCorrectly(c *validationContext) ValidationStatus {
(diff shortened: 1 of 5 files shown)


internal/host/validator.go
```diff
@@ -1305,6 +1305,22 @@ func checkDomainNameResolution(c *validationContext, domainName string) Validati
 	for _, domain := range response.Resolutions {
 		if domain.DomainName != nil && *domain.DomainName == domainName {
 			if len(domain.IPV4Addresses) != 0 || len(domain.IPV6Addresses) != 0 {
+				for _,host := range c.cluster.Hosts{
+					ipV4Addresses, ipV6Addresses, err := network.GetHostIPAddresses(host)
+					if (err != nil) {
```
ori-amizur, line 1310: Please remove parenthesis

internal/host/validator.go
```diff
@@ -1305,6 +1305,22 @@ func checkDomainNameResolution(c *validationContext, domainName string) Validati
 	for _, domain := range response.Resolutions {
 		if domain.DomainName != nil && *domain.DomainName == domainName {
 			if len(domain.IPV4Addresses) != 0 || len(domain.IPV6Addresses) != 0 {
+				for _,host := range c.cluster.Hosts{
```
ori-amizur, line 1308: Why in host validation, we are iterating over all the hosts ? I think we need to check only the current host.
Also please use the already parsed inventory

paul-maidment: We are iterating over the cluster hosts as the intention is to check whether or not the currently resolved address is actually one of the cluster hosts. If this is the case then the aim of the ticket is to reject that host.

internal/host/validator.go
```diff
@@ -1320,6 +1346,12 @@ func printIsDomainNameResolvedCorrectly(c *validationContext, status ValidationS
 		}
 		return fmt.Sprintf("Domain name resolution was successful for domain %s", domainName)
 	case ValidationFailure:
+		domain, err := getFirstMatchingResolution(c, domainName)
+		if err == nil && domain != nil {
```
ori-amizur, line 1350: Is the case err != nil handled ?

internal/host/hosts_suite_test.go
```diff
@@ -11,6 +12,7 @@ import (
 func TestHost(t *testing.T) {
 	RegisterFailHandler(Fail)
 	common.InitializeDBTest()
+	time.Sleep(10 * time.Second)
```
tsorya, line 15: why we need sleep? we should not have it

paul-maidment: Oops! Left some debug stuff in...deleted!

PR #3404, opened 2022-02-25
MGMT-9365: assisted kube-api SNO clusters should be converted to…

paul-maidment:
… day-2 automatically

This PR ensures that an SNO cluster is converted to a day2 cluster to
allow the addition of workers to the cluster.

## List all the issues related to this PR

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #4072, opened 2022-07-04
MGMT-10973: Enable DNS validations if coredns or keepalived static pod manifests in day-2 connectivity-check ignition

omertuc:
For context, this is a service-side follow-up to:

https://github.com/openshift/assisted-installer-agent/pull/388

and also this small agent fix https://github.com/openshift/assisted-installer-agent/pull/403

# Goal

During day-2 installations, we want the service to optionally perform
DNS validations when the worker attempts to join none-platform clusters.

# Issue

When the cluster is imported via our `.../v2/clusters/import` endpoint,
we have no way to know whether that cluster is a none-platform cluster
or cluster with managed networking (e.g. baremetal), so we have no way
to know whether we should perform the DNS validation or not. We wouldn't
like to have that validation on all the time, because it's unnecessarily
annoying for clusters with managed networking.

# Background

As part of our existing API connectivity-check, the service asks the
agent to download the worker.ign file from the to-be-joined-cluster's
machine-config-server, then send that back to the service.

# Solution

The contents of that file include information that will allow the
service to make an educated guess about whether the ignition originated
from a cluster with managed networking or not.

Also, a new "imported" column has been added to clusters, to indicate
whether they were created through a conversion or through being
imported. This is important because the solution should only be
applied for imported clusters, and this will provide a good way
to tell apart imported from non-imported clusters.

Also, when a user imports a cluster, we will inspect the `params.NewImportClusterParams.APIVipDnsname`
parameter and extract:

- The cluster name
- The cluster base domain

The cluster name will override the name given in `params.NewImportClusterParams.Name`,
see diff comment for more information on why.

The cluster base domain will be used to set the `BaseDNSDomain` of the
cluster, because up until now we didn't set it for imported cluster.

The reason `BaseDNSDomain` and `Name` have to be correctly set is
because the DNS validations use them to construct the domain names
that are being validated.

Also updated some validation messages for the API connectivity check and
DNS validations.

Also, the clusterdeployments_controller can now import SNO clusters,
it was an oversight that should have been done as part of https://github.com/openshift/assisted-service/commit/4cda26533d377f453f68783e8b2eae438734555d (https://github.com/openshift/assisted-service/pull/3404)

# Ignition Files

The ignition files we currently look at are:

```
/etc/kubernetes/manifests/coredns.yaml
/etc/kubernetes/manifests/keepalived.yaml
```

This is a hack - since we have no official way to know whether a worker
ignition file originated from a cluster with managed networking or not,
we instead rely on the presence of coredns and keepalived pod manifests
to indicate that. We only expect those to be present in clusters with
managed networking. To be a bit more robust, the service will consider
the presence of any one of them to mean that the cluster has managed
networking. This gives us better forwards compatibility if one of them
gets renamed / replaced with other technologies in future OCP versions.

Another way in which this is hacky is that users could manually create
static pods with the same name as part of their machine-configs, in
which case we would have a false-positive detection. But that is
admittedly very unlikely.

Hopefully we can negotiate with the relevant OCP teams to have a more
official, stable way to have this detection - like a magic empty file
placed somewhere in the ignition that we can check for the presence of.
Once we have such file, we can slowly deprecate this detection mechanism
and fully move to the new one by inspecting that file instead.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- Tested by manually doing day2 against a SNO using the service with imported cluster 
- [ ] No tests needed

## Assignees

/cc @tsorya
/cc @avishayt
/cc @eranco74

## Checklist

## Reviewers Checklist


internal/host/validator.go
```diff
@@ -1229,18 +1229,179 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
...
+
 		return false
 	}
 
-	// If its an SNO cluster with DNSMasq manifests enabled the check should be skipped
+	// We've exhausted all day-2 hosts, none of them seem to indicate that the
+	// cluster doesn't (or does) have managed networking, so we're forced to
+	// assume that it does in order to preserve our existing behavior of
+	// assuming the cluster has managed networking
+	return true
```
eranco74, line 1310: :+1: 
good call

internal/host/validator.go
```diff
@@ -1229,18 +1229,179 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
...
+		if host.APIVipConnectivity != "" {
+			var response models.APIVipConnectivityResponse
+			if err := json.Unmarshal([]byte(host.APIVipConnectivity), &response); err != nil {
+				v.log.WithError(err).Warnf("Invalid API connectivity response")
+				continue
+			}
+
+			if !response.IsSuccess {
+				continue
+			}
```
eranco74, line 1329: This method seem to share a lot of code with `importedClusterHasManagedNetworking` consider calling one from the other.

internal/host/validator.go
```diff
@@ -1229,18 +1229,179 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
...
+	for _, host := range cluster.Hosts {
+		if host.APIVipConnectivity != "" {
+			var response models.APIVipConnectivityResponse
+			if err := json.Unmarshal([]byte(host.APIVipConnectivity), &response); err != nil {
+				v.log.WithError(err).Warnf("Invalid API connectivity response")
+				continue
+			}
+
+			if !response.IsSuccess {
+				continue
```
eranco74, line 1328: Why continue? this should return true.
Otherwise, if the connectivity check fails, we might hang with validation pending forever.

internal/host/validator.go
```diff
@@ -1229,18 +1229,179 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
...
+
 	networkCfg, err := network.NewConfig()
 	if err != nil {
-		return false
+		// This should never happen, so the values chosen here are arbitrary
+		// and have no effect on anything
+		return false, ValidationSuccess
+	}
+
+	if common.IsSingleNodeCluster(cluster) && networkCfg.EnableSingleNodeDnsmasq {
```
eranco74, line 1397: We take care of the DNS on the SNO, what about the day2 worker? do we take care of it's DNS configuration as well (as if it is being added to a clsuter with managedNetworking)?

tsorya:
I think we should apply this solution not only for imported clusters to say the truth but i don't fully know the logic in ocm. If i remember right we are creating day2 when we want to add nodes in saas , will it be "imported" one?

internal/host/validator.go
```diff
@@ -1229,18 +1229,179 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
...
+		if file.Path == path {
+			return true
+		}
+	}
+
+	return false
+}
+
+func ignitionContainsManagedNetworkingFiles(config *ignition_types.Config) bool {
+	if ignitionHasFile(config, "/etc/kubernetes/manifests/coredns.yaml") ||
```
tsorya, line 1243: if we check both files then we need && .
I still think one of them is enough but if we decide to have both lets validate that both exists

internal/host/validator.go
```diff
@@ -1229,18 +1229,179 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
...
+		if err := json.Unmarshal([]byte(day2Host.APIVipConnectivity), &apiConnectivityResponse); err != nil {
+			v.log.WithError(err).Warnf("Invalid API connectivity response")
+			continue
+		}
+
+		if !apiConnectivityResponse.IsSuccess {
+			continue
+		}
+
+		config, _, err := v3_2.Parse([]byte(apiConnectivityResponse.Ignition))
```
tsorya, line 1293: will it be always 3.2 ?

omertuc: Probably not, but I think v3_2.Parse will handle any 3.x versions as they're supposed to be backwards compatible. 

@carbonin what would be the correct way to behave here? Parsing worker.ign ignition files from arbitrary clusters the user want us to help them join

internal/host/validator.go
omertuc: If connectivity check fails, the host is blocked on the connectivity check forever anyway. Basically we're pending here until at-least one of the hosts passes the connectivity check, which has to happen to all of them some time anyway

internal/host/validator.go
omertuc: See comment of why I check either and not both:

> To be a bit morerobust, we consider the presence of any one of them to mean that the cluster has managed networking. This gives us better forwards compatibility if either one of them gets renamed / replaced with other technologies in future OCP versions.

internal/host/validator.go
omertuc: >We take care of the DNS on the SNO,

correct

>what about the day2 worker?

Irrelevant at this point in the function. See "	// This is a day-1, user-managed-networking cluster" comment above that indicates anything below that comment deals with day-1 UMN clusters

>do we take care of it's DNS configuration as well (as if it is being added to a clsuter with managedNetworking)?

We definitely could hack DNS for them like we do in day-1 SNO, but we currently don't do that, so instead we annoy the user with validations to make the user do it manually

omertuc: Once we decide to hack DNS for day-2 UMN, we could drop those validations. But it's a bit more involved as we can no longer use simple MC as that would apply to all workers, which might not be desired, so we need a new pool... and it gets complicated

omertuc:
>I think we should apply this solution not only for imported clusters to say the truth but i don't fully know the logic in ocm. If i remember right we are creating day2 when we want to add nodes in saas , will it be "imported" one?

Yes, in REST when the user enters the OCM "add-hosts" tab it automatically creates a cluster by running import, so it will be an imported one.

internal/host/validator.go
```diff
@@ -1229,18 +1229,179 @@ func (v *validator) printDefaultRoute(c *validationContext, status ValidationSta
...
+		// create static pods with the same name as part of their
+		// machine-configs, in which case we would have a false-positive
+		// detection. But that is admittedly very unlikely.
+		//
+		// Hopefully we can negotiate with the relevant OCP teams to have a
+		// more official, stable way to have this detection - like a magic
+		// empty file placed somewhere in the ignition that we can check for
+		// the presence of. Once we have such file, we can slowly deprecate
+		// this detection mechanism and fully move to the new one by inspecting
+		// that file instead.
```
carbonin, line 1265: I think this longer explanation is probably better in the commit. I think something smaller would be good for a comment if you can come up with a useful smaller description.

internal/host/validator.go
carbonin: I think you might want something more like [`ParseCompatibleVersion`](https://github.com/coreos/ignition/blob/d508eb2e63df37d0069d93ef85cfafd548b6c63e/config/v3_2/config.go#L64). I'm not sure if the current implementation would error for 3.1, but it's probably worth a unit test.

omertuc: And what about forwards compatibility with future ignition versions?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1247,9 +1239,14 @@ func (r *ClusterDeploymentsReconciler) createNewDay2Cluster(
 		clusterParams.OpenshiftClusterID = &cid
 	}
 
+	// TODO: Remove this once CAPI tests are fixed to create the right DNS entries.
+	// For now we set it to false because CAPI tests don't have the right DNS entries
+	// and it will cause DNS validations to be disabled. This is just a temporary hack.
+	imported := false
```
eranco74, line 1245: Thanks

Makefile
```diff
@@ -377,7 +377,7 @@ deploy-test: _verify_cluster generate-keys update-local-image
 	$(MAKE) deploy-wiremock deploy-all
 
 # An alias for the deploy-test target
-deploy-service-for-subsystem-test: deploy-test enable-kube-api-for-subsystem
+deploy-service-for-subsystem-test: deploy-test
```
eranco74, line 380: Why?
I think this will break the operator-ztp-converged test

omertuc: Users who want to deploy the service for kube-api subsystem can follow [these](https://github.com/openshift/assisted-service/blob/a314b89e470e47ccc91651ae6775f3727bf95870/docs/dev/running-test.md?plain=1#L60-L65) instructions

You can't run it all the time because otherwise you break [the instructions](https://github.com/openshift/assisted-service/blob/a314b89e470e47ccc91651ae6775f3727bf95870/docs/dev/running-test.md?plain=1#L54-L58) for regular subsystem tests

PR #4073, opened 2022-07-05
NO-ISSUE: WIP

eranco74:
No need to review
just want to run a test...


eranco74:
The issue presist 
should get fixed by this https://github.com/openshift/assisted-service/pull/4075

PR #4075, opened 2022-07-05
Revert "MGMT-10850: Message for Day2 reboot is misleading (#3970)"

eranco74:
This PR seems to break the capi flow
it seems that as a result of this PR the agent CR of a day 2 worker will hang in rebooting and never get to done. 
This reverts commit e34274ae119fa9a3d5cd6786dd6de4716897729a.


omertuc:
[Failed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_assisted-service/4075/pull-ci-openshift-assisted-service-master-edge-e2e-metal-assisted-operator-capi/1544280109492473856), does this mean the revert was ineffective? Or is it something else?

PR #4079, opened 2022-07-05
MGMT-10844: Add virtual interface metrics for SaaS

CrystalChun:
[MGMT-10844](https://issues.redhat.com/browse/MGMT-10844)
As part of epic [MGMT-10087](https://issues.redhat.com/browse/MGMT-10087)
Virtual interface data per host and per cluster are available
in the SaaS environment, even if the feature is not enabled.

This PR records the virtual interfaces each host/cluster has and
exposes it as a metric. The data will help determine if this is a
viable feature we can enable in SaaS by allowing us to predict
performance/scalability and allowing us to see any patterns/common
interfaces.

**Note**: These metrics only cover the first two data points listed in this
[document](https://docs.google.com/document/d/1eFNcEHGv14iR50Z7cSN8lQuyQx4JaPVc4uaIbGvPdgI/edit?usp=sharing) since the rest require enabling the feature. (See Caveats under 
each data point type)

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

1. Deployed this image with assisted-test-infra, made sure `ENABLE_VIRTUAL_INTERFACES` is set to `false` in the configmap, and ran `make deploy_nodes` to create a cluster
2. Get the metrics for this cluster to make sure metrics were set
    ```
    $ curl "${ASSISTED_SERVICE_URL}/api/assisted-install/v2/events?categories=metrics&cluster_id=${CLUSTER_ID}" | jq '.[] | select(.message | contains("interfaces"))'
      {
        "cluster_id": "c28283b0-3ddf-45c3-a1f5-b31a8dcf7c98",
        "event_time": "2022-07-05T17:28:18.218Z",
        "host_id": "2c5557d6-447e-4611-bc55-be6b0b267c8f",
        "message": "nic.virtual_interfaces",
        "props": "{\"count\":2,\"types\":[\"device\",\"bridge\"]}",
        "severity": "info"
      },
      {
        "cluster_id": "c28283b0-3ddf-45c3-a1f5-b31a8dcf7c98",
        "event_time": "2022-07-05T17:28:18.219Z",
        "host_id": "2c5557d6-447e-4611-bc55-be6b0b267c8f",
        "message": "nic.physical_interfaces",
        "props": "{\"count\":2}",
        "severity": "info"
      }
    ```

3. Checked interface of a host to make sure no virtual interfaces were stored
    ```
    $ curl "${ASSISTED_SERVICE_URL}/api/assisted-install/v2/clusters/${CLUSTER_ID} | jq -r '.hosts[0].inventory' | jq '.interfaces[].type'
    "physical"
    "physical"
    ```
## Assignees

/cc @carbonin 
/cc @avishayt 
/cc @gamli75 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -352,15 +352,32 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster, 
 		return common.NewApiError(http.StatusInternalServerError, err)
 	}
 
-	// Remove virtual interfaces if the feature is not enabled
-	if !m.Config.EnableVirtualInterfaces {
-		var physicalInterfaces []*models.Interface = make([]*models.Interface, 0)
-		for _, intf := range inventory.Interfaces {
-			if intf.Type == "" || intf.Type == "physical" || intf.Type == "vlan" || intf.Type == "bond" {
-				// Empty interface type indicates an older agent, which only passes physical interfaces
-				physicalInterfaces = append(physicalInterfaces, intf)
-			}
+	existingHostInventory, err := common.UnmarshalInventory(h.Inventory)
+	if err != nil {
+		log.WithError(err).Debugf("failed to unmarshal host existing inventory %s", h.Inventory)
+	}
+
+	var physicalInterfaces []*models.Interface = make([]*models.Interface, 0)
+	var virtualInterfaces []string = make([]string, 0)
+	for _, intf := range inventory.Interfaces {
+		if intf.Type == "" || intf.Type == "physical" || intf.Type == "vlan" || intf.Type == "bond" {
+			// Empty interface type indicates an older agent, which only passes physical interfaces
+			physicalInterfaces = append(physicalInterfaces, intf)
+		} else {
+			virtualInterfaces = append(virtualInterfaces, intf.Type)
 		}
+	}
+
+	// Only record metrics if it's new/changed inventory and the host is associated with a cluster
+	if existingHostInventory == nil || (existingHostInventory != nil && len(physicalInterfaces) != len(existingHostInventory.Interfaces)) {
+		if h.ClusterID != nil {
+			m.eventsHandler.AddMetricsEvent(ctx, *h.ClusterID, h.ID, models.EventSeverityInfo, "nic.virtual_interfaces", time.Now(), "count", len(virtualInterfaces), "types", virtualInterfaces)
+			m.eventsHandler.AddMetricsEvent(ctx, *h.ClusterID, h.ID, models.EventSeverityInfo, "nic.physical_interfaces", time.Now(), "count", len(physicalInterfaces))
+		}
+	}
+
+	// Don't store virtual interfaces if the feature is not enabled
+	if !m.Config.EnableVirtualInterfaces {
 		inventory.Interfaces = physicalInterfaces
 	}
 
--- a/internal/common/test_configuration.go
+++ b/internal/common/test_configuration.go
@@ -252,30 +252,29 @@ func GenerateTestDefaultInventory() string {
 	return string(b)
 }
 
-func GenerateTestInventoryWithVirtualInterface() string {
+func generateInterfaces(amount int, intfType string) []*models.Interface {
+	interfaces := make([]*models.Interface, amount)
+	for i := 0; i < amount; i++ {
+		intf := models.Interface{
+			Name: fmt.Sprintf("eth%d", i),
+			IPV4Addresses: []string{
+				fmt.Sprintf("192.%d.2.0/24", i),
+			},
+			IPV6Addresses: []string{
+				fmt.Sprintf("2001:db%d::/32", i),
+			},
+			Type: intfType,
+		}
+		interfaces[i] = &intf
+	}
+	return interfaces
+}
+
+func GenerateTestInventoryWithVirtualInterface(physicalInterfaces, virtualInterfaces int) string {
+	interfaces := generateInterfaces(physicalInterfaces, "physical")
+	interfaces = append(interfaces, generateInterfaces(virtualInterfaces, "device")...)
 	inventory := &models.Inventory{
-		Interfaces: []*models.Interface{
-			{
-				Name: "eth0",
-				IPV4Addresses: []string{
-					"192.0.2.0/24",
-				},
-				IPV6Addresses: []string{
-					"2001:db8::/32",
-				},
-				Type: "physical",
-			},
-			{
-				Name: "cni1",
-				IPV4Addresses: []string{
-					"198.51.100.0/24",
-				},
-				IPV6Addresses: []string{
-					"2001:db8::/32",
-				},
-				Type: "device",
-			},
-		},
+		Interfaces: interfaces,
 		Disks: []*models.Disk{
 			TestDefaultConfig.Disks,
 		},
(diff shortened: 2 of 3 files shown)


internal/host/host.go
```diff
@@ -351,15 +351,30 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
...
+			// Empty interface type indicates an older agent, which only passes physical interfaces
+			physicalInterfaces = append(physicalInterfaces, intf)
+		} else {
+			virtualInterfaces = append(virtualInterfaces, intf.Type)
 		}
+	}
+
+	// Only record metrics if it's new/changed inventory
+	if existingHostInventory == nil || (existingHostInventory != nil && len(physicalInterfaces) != len(existingHostInventory.Interfaces)) {
+		m.eventsHandler.AddMetricsEvent(ctx, *h.ClusterID, h.ID, models.EventSeverityInfo, "nic.virtual_interfaces", time.Now(), "count", len(virtualInterfaces), "types", virtualInterfaces)
```
romfreiman, line 372: u should not add cluster/host id to the metrics. It will make it not really useful or queriable. 
And why do we actually need it per host/cluster. We just want to see how many of such are out there, not to use metrics to track down the cluster - for that we can query the inventory, dont we?

romfreiman:
@omertuc @rccrdpccl

internal/host/host.go
gamli75: @romfreiman this metric event API will add an event to DB and we will see it in the elastic - not in Prometheus. I don't think it's a problem to have cluster-id & host id if we are calling it only **once** for each host. 
@CrystalChun I think that we are calling updateInventory method many times so I don't think it's a good place to call this code.
Do we want to see this data only on clusters that started installation or on all clusters?

romfreiman: @gamli75 thanks for clarifying. then please ignore my comment :) I saw metric in the name and in the pr title

CrystalChun: @gamli75 this [condition](https://github.com/openshift/assisted-service/pull/4079/files#diff-549fd6f951612ce99043912371fffacd86d8249a40ef3905825be3d45f81dba4R371) ensures that the metrics is only called if the interfaces are new or have changed. 

> Do we want to see this data only on clusters that started installation or on all clusters?

I think it would be useful to see it on all clusters, regardless if they started install or not. What do you all think? @gamli75 @avishayt @filanov

internal/host/host.go
```diff
@@ -351,15 +351,30 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
...
+		m.eventsHandler.AddMetricsEvent(ctx, *h.ClusterID, h.ID, models.EventSeverityInfo, "nic.physical_interfaces", time.Now(), "count", len(physicalInterfaces))
```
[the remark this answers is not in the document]
carbonin: Yup, looks like subsystem caught it. From the assited service logs:
```
2022/07/05 18:39:18 http: panic serving 10.130.0.1:33518: runtime error: invalid memory address or nil pointer dereference
goroutine 1134 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1802 +0xb9
panic({0x2bddde0, 0x51c42b0})
	/usr/local/go/src/runtime/panic.go:1047 +0x266
github.com/openshift/assisted-service/internal/host.(*Manager).updateInventory(0xc0001ce2a0, {0x36fc5d0, 0xc00104bb60}, 0x0, 0xc0010d5080, {0xc001b1c000, 0x378}, 0x2f94901)
	/go/src/github.com/openshift/origin/internal/host/host.go:372 +0xbd8
github.com/openshift/assisted-service/internal/host.(*Manager).UpdateInventory(0x2f949a0, {0x36fc5d0, 0xc00104bb60}, 0x25a6085, {0xc001b1c000, 0xc0017ec1a0})
	/go/src/github.com/openshift/origin/internal/host/host.go:295 +0x37
```

internal/host/host.go
```diff
@@ -351,15 +351,30 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
...
+	}
+
+	// Don't store virtual interfaces if the feature is not enabled
+	if !m.Config.EnableVirtualInterfaces {
```
avishayt, line 377: It looks like this PR doesn't handle enabling virtual interfaces, right?

internal/host/host.go
avishayt: I don't know if we will care about interfaces for clusters that didn't install, but does having the information cost us anything?

gamli75: it will cost us more events in the DB

gamli75: @CrystalChun can we run cluster installation with this code and verify the number of events?

avishayt: Assuming the number of interfaces doesn't change it should be 1 event per cluster

avishayt: @CrystalChun Please make sure this is covered by UT

internal/host/host.go
CrystalChun: That's right. Do you think enabling should be part of this iteration too?

internal/host/host.go
CrystalChun: That's correct, it should be 1 event per cluster

@avishayt will do!

filanov: If we want to skip clusters that didn't start the installation it will need to hold the information until the installation is starting, meaning that we will need to store it in the DB, the flow that we have right now we don't store extra data. 
It will make the solution much more complicated, having another event is a small price to pay.

internal/host/host.go
avishayt: No, just wanted to make sure it was intentional

internal/host/host.go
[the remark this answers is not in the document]
filanov: +1 need to consider unbound hosts.

internal/host/host.go
```diff
@@ -351,15 +351,30 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
...
 
-	// Remove virtual interfaces if the feature is not enabled
-	if !m.Config.EnableVirtualInterfaces {
-		var physicalInterfaces []*models.Interface = make([]*models.Interface, 0)
-		for _, intf := range inventory.Interfaces {
-			if intf.Type == "" || intf.Type == "physical" || intf.Type == "vlan" || intf.Type == "bond" {
-				// Empty interface type indicates an older agent, which only passes physical interfaces
-				physicalInterfaces = append(physicalInterfaces, intf)
-			}
+	existingHostInventory, err := common.UnmarshalInventory(h.Inventory)
```
filanov, line 354: How about moving all this part into a function and place it under https://github.com/openshift/assisted-service/blob/9b5bb30c4e1c7ba42566f11eabb09fe16eaf3942/internal/bminventory/inventory.go#L2948 switch? 
There is a log of logic here that is not related to updateInventroy
@carbonin @CrystalChun what do you think?

internal/host/host.go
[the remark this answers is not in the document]
CrystalChun: Going to modify this to only publish metrics if the host is bound to a cluster - wdyt?

internal/host/host.go
CrystalChun: Agreed with Nick that the logic fits in this section since we're modifying/updating the interfaces (one aspect of the inventory which is similar to disk eligibility, etc.)

filanov: OK

carbonin:
/hold

Looks good, but we're waiting for the stage promotion to merge.

gamli75:
/unhold

I reviewed this PR and it's OK to merge it to 2.6 version

PR #4097, opened 2022-07-08
MGMT-11092: Merge the format and the condition methods in host validation

paul-maidment:
The present implementation of host validation has a problem, it splits the validation into two phases. The validation and the printing of the validation message. 

This leads to a situation in some cases where it is not possible to use the output of a validation to determine the message that should be sent back to the user, there are some workarounds but these involve poor practices such as code repetition.

This commit changes the validation to be a single method. Due to the complexity of the refactor, no effort has been made to make each of the validation methods their "most efficient" it is assumed that these fixes will take place in subsequent commits.

Because the validation methods made heavy use of return statements, the style has had to be changed to an if/else style and some code has had to be reorganised a little.

In theory, the existing tests should be sufficient to act as a "smoke" test for this functionality provided that there is enough coverage.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @tsorya 
/cc @ori-amizur 

## Checklist

## Reviewers Checklist


mkowalski:
/lgtm

It's a big change, but looks sane. Thanks!

internal/host/validator.go
```diff
@@ -425,366 +435,363 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
 	return boolValue(response.IsSuccess)
 }
 
-func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) ValidationStatus {
+func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) (ValidationStatus, string) {
```
ori-amizur, line 438: The main reason to do the change is not to perform checks twice, but we are doing double checks here.
Do you intend to refactor this function (and others), to eliminate double checks ?

internal/host/hosts_suite_test.go
```diff
@@ -11,6 +12,7 @@ import (
 func TestHost(t *testing.T) {
 	RegisterFailHandler(Fail)
 	common.InitializeDBTest()
+	time.Sleep(10 * time.Second)
```
avishayt, line 15: Debug leftovers (the import too)?

avishayt:
The new code looks much simpler.  Thank you!

internal/host/validator.go
paul-maidment: I have made changes that should address this.

gamli75 (approved):

carbonin (commented):
I think there are a lot of missed opportunities for simplification if that's what you're striving for.
There are quite a few places where you could return early and flatten some of the larger if-else blocks.

As someone who isn't particularly familiar with a lot of these validations it's hard to review this from a functional level. The diff doesn't really make it clear where code is being reused versus where you've had to add new logic.

If the goal was to change the validation struct to remove the formatter function I think a good first step/commit/PR would have been simply calling the formatter from the condition function and leave everything else alone for now. That would have been very easy for me to review. I'm fine if someone else wants to lgtm this as is, but it will take me a while to become comfortable with it all.

internal/host/validator.go
```diff
@@ -271,114 +269,87 @@ type validator struct {
...
-		return fmt.Sprintf("Unexpected status %s", status)
+		return status, fmt.Sprintf("Unexpected status %s", status)
 	}
 }
 
-func (v *validator) hasMinCpuCores(c *validationContext) ValidationStatus {
-	if c.inventory == nil {
-		return ValidationPending
+func (v *validator) hasMinCpuCores(c *validationContext) (ValidationStatus, string) {
+	if c.inventory != nil {
```
carbonin, line 309: I think this would be better reversed. Generally people reading this expect the success case to be the least nested.

internal/host/validator.go
```diff
@@ -271,114 +269,87 @@ type validator struct {
...
-		return fmt.Sprintf("The host is not eligible to participate in Openshift Cluster because the minimum required CPU cores for any role is %d, found only %d", c.minCPUCoresRequirement, c.inventory.CPU.Count)
-	case ValidationPending:
-		return "Missing inventory"
-	default:
-		return fmt.Sprintf("Unexpected status %s", status)
-	}
-}
+func (v *validator) hasMinMemory(c *validationContext) (ValidationStatus, string) {
+	status := ValidationPending
+	if c.inventory != nil {
```
carbonin, line 320: Same here, reversing this condition will make this more readable.

internal/host/validator.go
```diff
@@ -417,271 +388,214 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
-	switch status {
-	case ValidationSuccess:
-		if swag.BoolValue(context.cluster.UserManagedNetworking) {
-			return "No Machine Network CIDR needed: User Managed Networking"
-		}
-		if swag.StringValue(context.cluster.Kind) == models.ClusterKindAddHostsCluster {
-			return "No Machine Network CIDR needed: Day2 cluster"
+func (v *validator) isMachineCidrDefined(c *validationContext) (ValidationStatus, string) {
+	status := ValidationSuccessSuppressOutput
+	if c.infraEnv == nil {
```
carbonin, line 492: I think this would be better reversed.
Then the rest of the function could be one level less indented.

internal/host/validator.go
```diff
@@ -417,271 +388,214 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
-	case ValidationSuccess:
-		return fmt.Sprintf("Sufficient CPU cores for role %s", common.GetEffectiveRole(c.host))
-	case ValidationFailure:
-		return fmt.Sprintf("Require at least %d CPU cores for %s role, found only %d", c.clusterHostRequirements.Total.CPUCores, common.GetEffectiveRole(c.host), c.inventory.CPU.Count)
-	case ValidationPending:
-		return "Missing inventory or role"
-	default:
-		return fmt.Sprintf("Unexpected status %s", status)
+func (v *validator) hasMemoryForRole(c *validationContext) (ValidationStatus, string) {
+	if c.inventory != nil {
```
carbonin, line 522: Same, reverse this condition.

internal/host/validator.go
```diff
@@ -417,271 +388,214 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
-	}
-}
-
-func (v *validator) belongsToMachineCidr(c *validationContext) ValidationStatus {
+func (v *validator) belongsToMachineCidr(c *validationContext) (ValidationStatus, string) {
 	if c.infraEnv != nil {
-		return ValidationSuccessSuppressOutput
+		return ValidationSuccessSuppressOutput, ""
 	}
 	if swag.StringValue(c.cluster.Kind) == models.ClusterKindAddHostsCluster || (swag.BoolValue(c.cluster.UserManagedNetworking) && !common.IsSingleNodeCluster(c.cluster)) {
```
carbonin, line 583: Is there any reason for this condition now?
At the very least the `AddHostsCluster` case doesn't need to be inside this conditional.

The other one ends up being this I think. But it might be worth getting someone who actually knows this validation to sort out if this logic is correct.
```
if swag.BoolValue(c.cluster.UserManagedNetworking) && (swag.StringValue(c.cluster.Kind) == models.ClusterKindAddHostsCluster || !common.IsSingleNodeCluster(c.cluster)) {
```

internal/host/validator.go
```diff
@@ -691,103 +605,41 @@ func getRealHostname(host *models.Host, inventory *models.Inventory) string {
...
 
-func (v *validator) isIgnitionDownloadable(c *validationContext) ValidationStatus {
+func (v *validator) isIgnitionDownloadable(c *validationContext) (ValidationStatus, string) {
 	if c.infraEnv != nil {
-		return ValidationSuccessSuppressOutput
+		return ValidationSuccessSuppressOutput, ""
 	}
-	if !hostutil.IsDay2Host(c.host) {
-		return ValidationSuccessSuppressOutput
+	if !hostutil.IsDay2Host(c.host) || swag.BoolValue(c.cluster.UserManagedNetworking) {
```
carbonin, line 626: This user managed network case used to be `ValidationSuccess`, not `ValidationSuccessSuppressOutput`

Looks like you actually have it handled below so maybe just remove that condition from here?

internal/host/validator.go
```diff
@@ -691,103 +605,41 @@ func getRealHostname(host *models.Host, inventory *models.Inventory) string {
...
-		return "Ignition is downloadable"
-	case ValidationFailure:
-		return "Ignition is not downloadable. Please ensure host connectivity to the cluster's API"
-	case ValidationPending:
-		return "Ignition is not yet available, pending API connectivity"
-	default:
-		return fmt.Sprintf("Unexpected status %s", status)
+		return ValidationSuccess, "Ignition is downloadable"
 	}
+	return ValidationFailure, "Ignition is not downloadable. Please ensure host connectivity to the cluster's API"
```
carbonin, line 642: Since this message is in two places maybe it's worth defining it as a variable in the function?

internal/host/validator.go
```diff
@@ -849,56 +701,49 @@ func (v *validator) belongsToL3MajorityGroup(c *validationContext, majorityGroup
...
-		return ValidationSuccessSuppressOutput
+		return ValidationSuccessSuppressOutput, ""
 	}
-	if hostutil.IsDay2Host(c.host) || common.IsSingleNodeCluster(c.cluster) {
-		return ValidationSuccess
+	if hostutil.IsDay2Host(c.host) {
+		return ValidationSuccess, "Day2 host is not required to be connected to other hosts in the cluster"
+	}
+	if common.IsSingleNodeCluster(c.cluster) {
+		return ValidationSuccess, "Host has connectivity to the majority of hosts in the cluster"
```
carbonin, line 713: Maybe define this message in one place to reduce string duplication.

internal/host/validator.go
```diff
@@ -921,67 +766,57 @@ func (v *validator) missingNTPSyncResult(db *gorm.DB, host *models.Host) Validat
...
-
-	return boolValue(allImagesValid(imageStatuses))
-}
-
-func (v *validator) printSucessfullOrUnknownContainerImagesAvailability(c *validationContext, status ValidationStatus) string {
-	switch status {
-	case ValidationSuccess:
-		return "All required container images were either pulled successfully or no attempt was made to pull them"
-	case ValidationFailure:
+	if boolValue(allImagesValid(imageStatuses)) == ValidationFailure {
```
carbonin, line 810: `allImagesValid` returns a bool, just use that in the condition. Why convert and compare to a different type?

internal/host/validator.go
```diff
@@ -1022,49 +857,44 @@ func allImagesValid(imageStatuses common.ImageStatuses) bool {
...
+		return ValidationError, "Validation error"
 	}
-
-	return boolValue(info == nil || info.DiskSpeed == nil || !info.DiskSpeed.Tested || info.DiskSpeed.ExitCode == 0)
-}
-
-func (v *validator) printSufficientOrUnknownInstallationDiskSpeed(c *validationContext, status ValidationStatus) string {
-	switch status {
-	case ValidationSuccess:
+	if boolValue(info == nil || info.DiskSpeed == nil || !info.DiskSpeed.Tested || info.DiskSpeed.ExitCode == 0) == ValidationSuccess {
```
carbonin, line 865: Same, just use the condition directly. Why convert and compare to some other type?

internal/host/validator.go
```diff
@@ -1110,41 +940,42 @@ func comparisonBuilder(value float64) string {
...
+	if c.infraEnv != nil {
+		status = ValidationSuccessSuppressOutput
+	} else if len(c.cluster.Hosts) == 1 || c.clusterHostRequirements.Total.PacketLossPercentage == nil || common.GetEffectiveRole(c.host) == models.HostRoleAutoAssign || hostutil.IsDay2Host(c.host) {
+		// Single Node use case || no requirements defined || role is auto assign
+		status = ValidationSuccess
+	} else if len(c.host.Connectivity) == 0 {
+		status = ValidationPending
+	} else {
+		s, _, _ := v.validatePacketLossForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts)
+		status = s
```
carbonin, line 957: Assign directly to `status`, no need for the intermediate variable.

internal/host/validator.go
```diff
@@ -1498,51 +1270,81 @@ func (v *validator) shouldValidateDNSResolution(cluster *common.Cluster) (bool,
...
-		return ret
+		if c.cluster != nil {
+			apiDomainName := domainNameToResolve(c, constants.InternalAPIClusterSubdomain)
+			return ret, printIsDomainNameResolvedCorrectly(c, ret, apiDomainName, target)
+		}
 	}
 	apiInternalDomainName := domainNameToResolve(c, constants.InternalAPIClusterSubdomain)
-	return checkDomainNameResolution(c, apiInternalDomainName)
+	status := checkDomainNameResolution(c, apiInternalDomainName)
+	return status, printIsDomainNameResolvedCorrectly(c, status, apiInternalDomainName, target)
```
carbonin, line 1333: I think you could simplify this as:
```go
	shouldValidate, status := v.shouldValidateDNSResolution(c.cluster)

	apiInternalDomainName := domainNameToResolve(c, constants.InternalAPIClusterSubdomain)
        if shouldValidate {
                status = checkDomainNameResolution(c, apiInternalDomainName)
        }
	return status, printIsDomainNameResolvedCorrectly(c, status, apiInternalDomainName, target)
```

Where did that cluster conditional come from? It doesn't seem to really be needed.

internal/host/validator.go
```diff
@@ -1498,51 +1270,81 @@ func (v *validator) shouldValidateDNSResolution(cluster *common.Cluster) (bool,
...
 	if shouldValidate, ret := v.shouldValidateDNSResolution(c.cluster); !shouldValidate {
-		return ret
+		if c.cluster != nil {
+			return ret, v.printIsAppsDomainNameResolvedCorrectly(c, ret)
+		}
 	}
 	appsDomainName := fmt.Sprintf("%s.apps.%s.%s", constants.AppsSubDomainNameHostDNSValidation, c.cluster.Name, c.cluster.BaseDNSDomain)
-	return checkDomainNameResolution(c, appsDomainName)
+	status := checkDomainNameResolution(c, appsDomainName)
+	return status, v.printIsAppsDomainNameResolvedCorrectly(c, status)
```
carbonin, line 1347: Same simplification as above could be made here.

carbonin (commented):
Mostly just style and small optimizations.

Looking a lot better :+1:

internal/host/validator.go
paul-maidment: I discovered during testing that if c.cluster is nil then you will get crashes when you call printIsDomainNameResolvedCorrectly, arguably I could put the defensive code in the print function. But I had problems during tests until I introduced this.

Not really sure why it did not show up in tests before this.

paul-maidment: Not sure that removing this condition is a good idea

paul-maidment: The nil cluster name stuff, I moved this to here

`func domainNameToResolve(c *validationContext, name string) string {
	if c.cluster == nil {
		return ""
	}
	return fmt.Sprintf("%s.%s.%s", name, c.cluster.Name, c.cluster.BaseDNSDomain)
}`

the original function was

`func domainNameToResolve(c *validationContext, name string) string {
	return fmt.Sprintf("%s.%s.%s", name, c.cluster.Name, c.cluster.BaseDNSDomain)
}`

so, if c.cluster were to be nil, there would be a panic

I'm not entirely sure why the solution is now sensitive to this nil value (maybe the old formatter code handled errors differently) but I had to add this to get tests to pass.

internal/host/validator.go
paul-maidment: See above comment

internal/host/validator.go
```diff
@@ -417,271 +389,211 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
-	switch status {
-	case ValidationSuccess:
-		return "Sufficient disk capacity"
-	case ValidationFailure:
-		return "No eligible disks were found, please check specific disks to see why they are not eligible"
-	case ValidationPending:
-		return "Missing inventory"
-	default:
-		return fmt.Sprintf("Unexpected status %s", status)
+	if boolValue(len(disks) > 0) == ValidationSuccess {
```
carbonin, line 485: `if len(disks) > 0 {`

internal/host/validator.go
```diff
@@ -417,271 +389,211 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
 			c.host.ID.String(), models.StepTypeTangConnectivityCheck, c.host.TangConnectivity, c.host.DiscoveryAgentVersion,
 		)
-		return ValidationSuccessSuppressOutput
+		return ValidationSuccessSuppressOutput, ""
 	}
-
 	var response models.TangConnectivityResponse
-	if err := json.Unmarshal([]byte(c.host.TangConnectivity), &response); err != nil {
-		return ValidationFailure
+	if err := json.Unmarshal([]byte(c.host.TangConnectivity), &response); err == nil && boolValue(response.IsSuccess) == ValidationSuccess {
```
carbonin, line 395: Just use `response.IsSuccess` directly

internal/host/validator.go
```diff
@@ -271,114 +269,88 @@ type validator struct {
...
-	switch status {
-	case ValidationSuccess:
-		return "Sufficient CPU cores"
-	case ValidationFailure:
-		return fmt.Sprintf("The host is not eligible to participate in Openshift Cluster because the minimum required CPU cores for any role is %d, found only %d", c.minCPUCoresRequirement, c.inventory.CPU.Count)
-	case ValidationPending:
-		return "Missing inventory"
-	default:
-		return fmt.Sprintf("Unexpected status %s", status)
+	if boolValue(c.inventory.CPU.Count >= c.minCPUCoresRequirement) == ValidationSuccess {
```
carbonin, line 312: `if c.inventory.CPU.Count >= c.minCPUCoresRequirement {`

internal/host/validator.go
```diff
@@ -417,271 +389,211 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
+			return ValidationSuccess, "No Machine Network CIDR needed: Day2 cluster"
 		}
-	default:
-		return fmt.Sprintf("Unexpected status %s", status)
+		return ValidationSuccess, "Machine Network CIDR is defined"
 	}
+	if swag.BoolValue(c.cluster.VipDhcpAllocation) {
+		return ValidationFailure, "Machine Network CIDR is undefined"
+	}
+	return ValidationFailure, "Machine Network CIDR is undefined; the Machine Network CIDR can be defined by setting either the API or Ingress virtual IPs"
```
carbonin, line 509: This whole middle bit of the function could be better I think. A lot of the conditions are duplicated. I'm pretty sure this is identical.

```go
	if swag.BoolValue(c.cluster.UserManagedNetworking) {
		return ValidationSuccess, "No Machine Network CIDR needed: User Managed Networking"
	}
	if swag.StringValue(c.cluster.Kind) == models.ClusterKindAddHostsCluster {
		return ValidationSuccess, "No Machine Network CIDR needed: Day2 cluster"
	}
	if network.IsMachineCidrAvailable(c.cluster) {
		return ValidationSuccess, "Machine Network CIDR is defined"
	}
	if swag.BoolValue(c.cluster.VipDhcpAllocation) {
		return ValidationFailure, "Machine Network CIDR is undefined"
	}
	return ValidationFailure, "Machine Network CIDR is undefined; the Machine Network CIDR can be defined by setting either the API or Ingress virtual IPs"
```

internal/host/validator.go
```diff
@@ -417,271 +389,211 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
 }
 
-func (v *validator) hasCPUCoresForRole(c *validationContext) ValidationStatus {
-	if c.inventory == nil {
-		return ValidationPending
+func (v *validator) hasCPUCoresForRole(c *validationContext) (ValidationStatus, string) {
+	if c.inventory != nil {
+		if boolValue(c.inventory.CPU.Count >= c.clusterHostRequirements.Total.CPUCores) == ValidationSuccess {
```
carbonin, line 514: Don't need boolValue here.

internal/host/validator.go
```diff
@@ -417,271 +389,211 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
-	case ValidationPending:
-		return "Missing inventory or role"
-	default:
-		return fmt.Sprintf("Unexpected status %s", status)
+func (v *validator) hasMemoryForRole(c *validationContext) (ValidationStatus, string) {
+	if c.inventory == nil {
+		return ValidationPending, "Missing inventory or role"
+	}
+	requiredBytes := conversions.MibToBytes(c.clusterHostRequirements.Total.RAMMib)
+	if boolValue(c.inventory.Memory.PhysicalBytes >= requiredBytes) == ValidationSuccess {
```
carbonin, line 527: Don't need `boolValue` here.

internal/host/validator.go
```diff
@@ -417,271 +389,211 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
 	}
 	if funk.ContainsString(invalidPlatforms, c.inventory.SystemVendor.ProductName) {
 		// In case there is no cluster validation is pending
 		if c.infraEnv != nil {
-			return ValidationSuccessSuppressOutput
+			return ValidationSuccessSuppressOutput, ""
 		} else {
 			//In case userManagedNetworking is true, we don't care about the platform
-			return boolValue(swag.BoolValue(c.cluster.UserManagedNetworking))
+			if boolValue(swag.BoolValue(c.cluster.UserManagedNetworking)) == ValidationSuccess {
```
carbonin, line 571: Don't need `boolValue` here.

internal/host/validator.go
```diff
@@ -417,271 +389,211 @@ func (v *validator) areTangServersReachable(c *validationContext) ValidationStat
...
-			return "No machine network CIDR validation needed: Day2 cluster"
-		}
-		return "Host belongs to all machine network CIDRs"
-	case ValidationFailure:
-		return "Host does not belong to machine network CIDRs. Verify that the host belongs to every CIDR listed under machine networks"
-	case ValidationPending:
-		return "Missing inventory or machine network CIDR"
-	default:
-		return fmt.Sprintf("Unexpected status %s", status)
+	if boolValue(network.IsHostInPrimaryMachineNetCidr(v.log, c.cluster, c.host)) == ValidationFailure {
```
carbonin, line 593: Don't need `boolValue` here.

internal/host/validator.go
carbonin: Any ideas about this @paul-maidment ?

internal/host/validator.go
```diff
@@ -691,103 +603,41 @@ func getRealHostname(host *models.Host, inventory *models.Inventory) string {
...
+		return ValidationFailure, "Ignition is not downloadable. Please ensure host connectivity to the cluster's API"
 	}
-
-	return boolValue(response.IsSuccess)
-}
-
-func (v *validator) printIgnitionDownloadable(c *validationContext, status ValidationStatus) string {
-	switch status {
-	case ValidationSuccess:
+	if boolValue(response.IsSuccess) == ValidationSuccess {
```
carbonin, line 634: Don't need `boolValue`

internal/host/validator.go
```diff
@@ -1022,49 +855,44 @@ func allImagesValid(imageStatuses common.ImageStatuses) bool {
...
 	}
-
-	return boolValue(info == nil || info.DiskSpeed == nil || !info.DiskSpeed.Tested || info.DiskSpeed.ExitCode == 0)
-}
-
-func (v *validator) printSufficientOrUnknownInstallationDiskSpeed(c *validationContext, status ValidationStatus) string {
-	switch status {
-	case ValidationSuccess:
+	if info == nil || info.DiskSpeed == nil || !info.DiskSpeed.Tested || info.DiskSpeed.ExitCode == 0 {
 		info, _ := v.getBootDeviceInfo(c.host)
```
carbonin, line 864: This is already being called as the first line of this function. You should be able to simplify this.

internal/host/validator.go
```diff
@@ -1110,41 +938,41 @@ func comparisonBuilder(value float64) string {
...
 }
 
-func (v *validator) printSufficientNetworkLatencyRequirementForRole(c *validationContext, status ValidationStatus) string {
+func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationContext) (ValidationStatus, string) {
+
+	var status ValidationStatus
+	var message string
+
+	if c.infraEnv != nil {
+		status = ValidationSuccessSuppressOutput
```
carbonin, line 947: I think this function can be simplified. For example, you can return early here as this is the only place you get this status.

internal/host/validator.go
```diff
@@ -1110,41 +938,41 @@ func comparisonBuilder(value float64) string {
...
+	} else if len(c.cluster.Hosts) == 1 || c.clusterHostRequirements.Total.PacketLossPercentage == nil || common.GetEffectiveRole(c.host) == models.HostRoleAutoAssign || hostutil.IsDay2Host(c.host) {
+		// Single Node use case || no requirements defined || role is auto assign
+		status = ValidationSuccess
+	} else if len(c.host.Connectivity) == 0 {
+		status = ValidationPending
+	} else {
+		status, _, _ = v.validatePacketLossForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts)
```
carbonin, line 954: I think you can avoid calling this twice (it's also called in the `ValidationFailure` case below) if you save off the packet loss and err values into variables here.

internal/host/validator.go
```diff
@@ -1110,41 +938,41 @@ func comparisonBuilder(value float64) string {
...
+	var status ValidationStatus
+	var message string
+
+	if c.infraEnv != nil {
+		status = ValidationSuccessSuppressOutput
+	} else if len(c.cluster.Hosts) == 1 || c.clusterHostRequirements.Total.PacketLossPercentage == nil || common.GetEffectiveRole(c.host) == models.HostRoleAutoAssign || hostutil.IsDay2Host(c.host) {
+		// Single Node use case || no requirements defined || role is auto assign
+		status = ValidationSuccess
+	} else if len(c.host.Connectivity) == 0 {
+		status = ValidationPending
```
carbonin, line 952: This can return early.

internal/host/validator.go
```diff
@@ -1498,51 +1267,73 @@ func (v *validator) shouldValidateDNSResolution(cluster *common.Cluster) (bool,
...
-func (v *validator) isAPIInternalDomainNameResolvedCorrectly(c *validationContext) ValidationStatus {
+func (v *validator) isAPIInternalDomainNameResolvedCorrectly(c *validationContext) (ValidationStatus, string) {
+	target := "internal API load balancer"
 	if c.infraEnv != nil {
-		return ValidationSuccessSuppressOutput
+		return ValidationSuccessSuppressOutput, ""
 	}
 	if shouldValidate, ret := v.shouldValidateDNSResolution(c.cluster); !shouldValidate {
-		return ret
+		apiDomainName := domainNameToResolve(c, constants.InternalAPIClusterSubdomain)
```
carbonin, line 1319: I thought I had a comment here about how this could be improved.

There's no reason to call this function twice.

carbonin: ref https://github.com/openshift/assisted-service/pull/4097#discussion_r929195361

internal/host/validator.go
carbonin: It's still relevant. You should be able to deduplicate the `v.printIsAppsDomainNameResolvedCorrectly` call.

carbonin (commented):
Mostly pointing out places you can use bools as bools rather than unnecessarily converting them to status strings to just later compare them.

carbonin (approved):

PR #3700, opened 2022-04-24
Bug 2077162: Can not provision SNOs in a disconnected environment

eranco74:
Recent updates to the oc client used by the assisted-service seems to break ICSP in disconnected environment.
This PR revert these changes:
1. Revert "NO-ISSUE: Add a retry when fetching oc with curl (#3687)"
This reverts commit 0da8d3899d8e421a424a78816bbcc75b22683384.

2. Revert "[MGMT-5442](https://issues.redhat.com//browse/MGMT-5442): assisted-service should use official oc client (#3657)"
This reverts commit cdbb65b98076673d156f5a5b787bad9378528531.

3. Revert "Bug 2069976: Error with oc-mirror using an old outdated unsupported oc client in AI (#3598)"
This reverts commit 53616082111984bbb43f4c848f7ce11c915f579f.

- Should this PR be tested by the reviewer?
- Is this PR relying on CI for an e2e test run? Yes
- Should this PR be tested in a specific environment? Disconnected env
- Any logs, screenshots, etc that can help with the review process? 
Running:
./oc adm -v=5 release extract --command=openshift-install --to=./ --insecure=false <mirrored release image> --registry-config=/tmp/registry.json 2>&1 >/dev/null | grep "Will extract"
and checking where it's extracting from (should be from an image in the mirror registry and not quay.io...)

-->

## List all the issues related to this PR

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [x] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[...]

PR #4115, opened 2022-07-11
MGMT-10209: Pass in icsp file to be used for 'oc adm release extract'

bfournie:
Pass in an ICSP file to the 'oc' command to extract the installer.  The ICSP file is created using the ImageContentSources in the install-config.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


pkg/generator/generator.go
```diff
@@ -32,6 +32,7 @@ type Config struct {
 	ReleaseImageMirror string
 	DummyIgnition      bool   `envconfig:"DUMMY_IGNITION"`
 	InstallInvoker     string `envconfig:"INSTALL_INVOKER" default:"assisted-installer"`
+	IcspFile           string `envconfig:"ICSP_FILE" default:""`
```
eranco74, line 35: Who should set this env?
What should be the content of `IcspFile` is it the imageContentSourcePolicy content or file path?

eranco74 (commented):
This ICSP configuration is used solely for extracting the installer binary from the release image in the assisted-service pod (running on the HUB cluster).
I think that the HUB cluster should have the ICSP (I think it's even mounted inside the pod)
so we probably would get it from there.
@avishayt perhaps we should define an API for installing from mirror registry and allow users to provide ICSP (currently they need to use install-config override)

pkg/generator/generator.go
avishayt: This will be set when running BILLI

avishayt:
If linking to github in the commit message, please use a permalink

avishayt:
> This ICSP configuration is used solely for extracting the installer binary from the release image in the assisted-service pod (running on the HUB cluster). I think that the HUB cluster should have the ICSP (I think it's even mounted inside the pod) so we probably would get it from there. @avishayt perhaps we should define an API for installing from mirror registry and allow users to provide ICSP (currently they need to use install-config override)

1. We can consider adding it as an API, though it would probably be more useful in the k8s API which isn't being used by BILLI.
2. Why is this PR needed if BILLI can just use install-config override?

pkg/generator/generator.go
bfournie: Yes this will be set when running BLLI, it will be the file path of a mounted ICSP API file.

bfournie: We are currently passing into to the assisted-service container the registries.conf file, mounted in /etc/containers, and this is used [here](https://github.com/openshift/assisted-service/blob/dd79548c59fb0756b5a6536a8868ab29123dfcd7/internal/installcfg/builder/builder.go#L137)
to set up the ImageContentSources in the install-config. We could also use that to build up an ImageContenSourcePolicy [API](https://docs.openshift.com/container-platform/4.10/rest_api/operator_apis/imagecontentsourcepolicy-operator-openshift-io-v1alpha1.html) to pass into the 'oc' command if that is the preferred way to go.

zaneb: This should also be set in MCE. Doing this is a pre-requisite to switching to the official `oc` command instead of using a fork.
BILLI is the first user only because it has to use official builds.

bfournie:
> 2. Why is this PR needed if BILLI can just use install-config override?

I'm not sure I understand the question. We weren't planning on doing a install-config override from BLLI, but even so I don't think that would help in passing the ICSP to 'oc'.

zaneb (commented):
My main question is whether it'd be possible to generate the ICSP from the `/etc/containers/registries.conf` file and write it to a tempfile. That would save having to configure it in two places, and possibly eliminate a lot of the boilerplate here.
Talk of adding an API for this makes it unclear whether that would be flexible enough for all use cases though.

internal/ignition/ignition.go
```diff
@@ -1513,3 +1524,67 @@ func proxySettingsForIgnition(httpProxy, httpsProxy, noProxy string) (string, er
...
+
+	icspFile, err := ioutil.TempFile("", "icsp-file")
+	if err != nil {
+		return "", err
+	}
+	if _, err := icspFile.Write(contents); err != nil {
+		os.Remove(icspFile.Name())
+		return "", err
+	}
+	icspFile.Close()
```
zaneb, line 1545: Need to defer this prior to line 1541, otherwise we will leak a file descriptor if the Write() fails.

internal/ignition/ignition.go
```diff
@@ -288,15 +290,24 @@ func (g *installerGenerator) UploadToS3(ctx context.Context) error {
...
+	var icspFile string
 	log := logutil.FromContext(ctx, g.log)
 
 	// In case we don't want to override image for extracting installer use release one
 	if g.installerReleaseImageOverride == "" {
 		g.installerReleaseImageOverride = g.releaseImage
 	}
 
+	// If ImageContentSources are defined, store in a file for the 'oc' command
+	defer removeIcspFile(icspFile)
```
zaneb, line 302: Arguments are evaluated at the time of the defer, so this will always pass `""`. Need to move below line 304.

internal/ignition/ignition.go
```diff
@@ -1513,3 +1524,67 @@ func proxySettingsForIgnition(httpProxy, httpsProxy, noProxy string) (string, er
...
+	if contents == nil {
+		return "", nil
+	}
+
+	icspFile, err := ioutil.TempFile("", "icsp-file")
+	if err != nil {
+		return "", err
+	}
+	if _, err := icspFile.Write(contents); err != nil {
+		os.Remove(icspFile.Name())
```
zaneb, line 1542: I would defer this as well, just so it executes after the file is already closed.

internal/ignition/ignition.go
bfournie: Good catch, fixed.

internal/ignition/ignition.go
bfournie: Changed to add the Close if the Write fails, seems more readable than using defer

internal/ignition/ignition.go
bfournie: Changed to add close if Write() fails

internal/ignition/ignition.go
```diff
@@ -1513,3 +1523,68 @@ func proxySettingsForIgnition(httpProxy, httpsProxy, noProxy string) (string, er
...
+	}
+	if contents == nil {
+		return "", nil
+	}
+
+	icspFile, err := ioutil.TempFile("", "icsp-file")
+	if err != nil {
+		return "", err
+	}
+	if _, err := icspFile.Write(contents); err != nil {
```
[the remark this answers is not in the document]
bfournie: Good idea, added.

internal/ignition/ignition.go
```diff
@@ -1513,3 +1524,76 @@ func proxySettingsForIgnition(httpProxy, httpsProxy, noProxy string) (string, er
...
 	return buf.String(), nil
 }
+
+func getIcspFileFromInstallConfig(cfg []byte) (string, error) {
+	contents, err := getIcsp(cfg)
+	if err != nil {
+		return "", err
+	}
+	if contents == nil {
+		log.Infof("No ImageContentsSources in install-config to build ICSP file")
```
zaneb, line 1534: Need to pass log in to this function.

zaneb: also typo in ImageContentSources

zaneb:
/lgtm cancel
Actually, I think we also need to revert #3700 in this PR because the oc fork from https://github.com/eranco74/oc/commits/ai-bug-1823143-multi-arch-ai-bug-2069976 doesn't have the icsp-file option

zaneb:
Hopefully https://github.com/openshift/release/pull/30508 should fix IPv6.
/test edge-e2e-ai-operator-ztp-disconnected

pawanpinjarkar (approved):
/approve

PR #4117, opened 2022-07-11
MGMT-10739: Reclaim agents on unbind

carbonin:
Run a pod on the spoke node when the agent associated with that node is unbound.
This agent will boot the host back into the discovery image, allowing it to be reused from the same infraEnv.

This is only done when the agent does not have an associated BareMetalHost.

This PR also adds a feature flag for all of this that will prevent us from moving to `HostStatusReclaiming` even if the BMH isn't present. This is added so that the feature can be worked on incrementally. Without this we would need all the agent commands to be implemented before this could be merged.

https://issues.redhat.com/browse/MGMT-10739

Part of the implementation for the [Automatically Return Agents to InfraEnv enhancement](https://github.com/openshift/assisted-service/blob/6b8401d3b0f0ceea5a837f33c17b9b1cd5ec52a3/docs/enhancements/auto-return-agent-to-infra-env.md)

~~Built on https://github.com/openshift/assisted-service/pull/4139~~ Merged

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

Ran this version of assisted-service in the hub cluster of a dev-scripts setup.
This allowed the hub cluster and the spokes to exist on the same libvirt network.

Created an SNO spoke and added 2 workers as a day-2 install step.
Unbound a worker and observed the pod get started on the spoke node.

## Assignees

/cc @danielerez 
/cc @ori-amizur 
/cc @CrystalChun 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -339,10 +339,11 @@ func (r *AgentReconciler) tryApproveDay2CSRs(ctx context.Context, agent *aiv1bet
 }
 
 func (r *AgentReconciler) unbindHost(ctx context.Context, log logrus.FieldLogger, agent, origAgent *aiv1beta1.Agent, h *common.Host) (ctrl.Result, error) {
-	host, err2 := r.Installer.UnbindHostInternal(ctx, installer.UnbindHostParams{
+	params := installer.UnbindHostParams{
 		HostID:     *h.ID,
 		InfraEnvID: h.InfraEnvID,
-	})
+	}
+	host, err2 := r.Installer.UnbindHostInternal(ctx, params, false)
 	if err2 != nil {
 		return r.updateStatus(ctx, log, agent, origAgent, &h.Host, nil, err2, !IsUserError(err2))
 	}
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
+	bmh_v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	. "github.com/openshift/assisted-service/api/common"
 	aiv1beta1 "github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/internal/bminventory"
@@ -84,6 +85,9 @@ type AgentReconciler struct {
 	AuthType                   auth.AuthType
 	SpokeK8sClientFactory      spoke_k8s_client.SpokeK8sClientFactory
 	ApproveCsrsRequeueDuration time.Duration
+	EnableHostReclaim          bool
+	AgentContainerImage        string
+	reclaimer                  *agentReclaimer
 }
 
 // +kubebuilder:rbac:groups=agent-install.openshift.io,resources=agents,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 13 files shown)


internal/controller/controllers/spoke_k8s_client.go
```diff
@@ -109,9 +111,60 @@ func (c *spokeK8sClient) ApproveCsr(csr *certificatesv1.CertificateSigningReques
...
+		return errors.Wrapf(err, "failed to find node %s", hostname)
+	}
+
+	namespaceName := "assisted-installer"
+	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespaceName}}
+	if err := c.Get(ctx, types.NamespacedName{Name: ns.Name}, ns); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return errors.Wrapf(err, "failed to get namespace %s", namespaceName)
+		}
+		if err := c.Create(ctx, ns); err != nil {
```
ori-amizur, line 134: Are we leaving the namespace in the spoke after reclaim ?

internal/controller/controllers/spoke_k8s_client.go
```diff
@@ -109,9 +111,60 @@ func (c *spokeK8sClient) ApproveCsr(csr *certificatesv1.CertificateSigningReques
...
-	node, err := c.nodesClient.Get(context.TODO(), name, metav1.GetOptions{})
+	node := &corev1.Node{}
+	if err := c.Get(context.TODO(), types.NamespacedName{Name: name}, node); err != nil {
+		return nil, err
+	}
+	return node, nil
+}
+
+func (c *spokeK8sClient) StartAgentForReclaim(ctx context.Context, agent *aiv1beta1.Agent, image string) error {
+	hostname := getAgentHostname(agent)
```
ori-amizur, line 122: Maybe we should consider leaving this client to be more generic, and just add a function CreatePod, or CreatePosOnHost ?

internal/controller/controllers/spoke_k8s_client.go
carbonin: In clusters we install this namespace is actually present.
It's where the controller job runs.

But also, yes. We would be leaving anything we create there.
I think this could probably be handled as a part of https://issues.redhat.com/browse/MGMT-10006

internal/controller/controllers/spoke_k8s_client.go
carbonin: :man_shrugging:

The generic client is `client.Client`.
Honestly `GetNode`, and `ListCsrs` feel like they shouldn't exist. Those functions don't add anything meaningful over `client.Client`.
I'd rather see this client object wrap some significant logic that maybe we don't want to include in other files.

`ApproveCsr` is a pretty good example. We're adding a message that it was approved by assisted service and we need to build a condition in a specific way. I actually think it could include even more logic if it were up to me.

Is there anything in particular you object to in here? As far as I can tell the hostname is all I really need from the agent (this is still WIP after all), so maybe if I just passed the hostname as a parameter?

carbonin: Actually in the case of `ListCsrs` I think it _definitely_ shouldn't exist. It's just an alias that prevents us from using contexts properly.

carbonin:
A note for reviewers @ori-amizur @danielerez @CrystalChun 

I'm sorry this PR is so large, but I tried to make the individual commits reviewable.
Going through each commit should make this a lot easier to look at.

Also cc @avishayt @mhrivnak in case you want to take a look

internal/controller/controllers/agent_controller.go
```diff
@@ -337,14 +345,97 @@ func (r *AgentReconciler) tryApproveDay2CSRs(ctx context.Context, agent *aiv1bet
...
+	return true, nil
+}
+
+func (r *AgentReconciler) shouldReclaimOnUnbind(ctx context.Context, agent *aiv1beta1.Agent, clusterRef *aiv1beta1.ClusterReference) bool {
+	if !r.EnableHostReclaim {
+		return false
+	}
+
+	// default to not attempting to reclaim as that's the safer route
+	if foundBMH, err := r.bmhExists(ctx, agent); err != nil || foundBMH {
```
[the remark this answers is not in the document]
carbonin: Nope, it's correct as is.

We don't reclaim (return false) if we fail to find a bmh or if we do find one.

internal/controller/controllers/agent_controller.go
```diff
@@ -337,14 +345,97 @@ func (r *AgentReconciler) tryApproveDay2CSRs(ctx context.Context, agent *aiv1bet
...
+		if err != nil || cluster.KubeKeyName == "" || cluster.KubeKeyNamespace == "" {
+			if err != nil {
+				r.Log.WithError(err).Warnf("failed to get cluster %s, not attempting reclaim", h.ClusterID)
+			} else {
+				r.Log.Warnf("cluster %s missing kube key (%s/%s), not attempting reclaim", h.ClusterID, cluster.KubeKeyNamespace, cluster.KubeKeyName)
+			}
+		} else {
+			clusterRef := &aiv1beta1.ClusterReference{Namespace: cluster.KubeKeyNamespace, Name: cluster.KubeKeyName}
+			reclaim = r.shouldReclaimOnUnbind(ctx, origAgent, clusterRef)
+			fmt.Printf("should reclaim returned %v\n", reclaim)
```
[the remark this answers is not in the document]
carbonin: oops, yup that was for debugging, removed

PR #4118, opened 2022-07-11
MGMT-11129: Only set bootstrap host when it's attached to a cluster

CrystalChun:
[MGMT-11129](https://issues.redhat.com/browse/MGMT-11129)
Initially there wasn't a check to see if the host was attached to a
cluster before setting the bootstrap host. In a scenario like
late-binding, there could be hosts without a cluster, which would
cause a panic.

This checks to make sure the cluster is not nil before executing
the logic to set a bootstrap host.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] Manual (Elaborate on how it was tested): created a late-binding cluster and set the BOOTSTRAP_MAC_HOST in the assisted-service configmap to a host that was not the bootstrap host. Ensured that the panic didn't occur
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -327,7 +327,7 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster, 
 		return common.NewApiError(http.StatusNotFound, err)
 	}
 
-	if m.Config.BootstrapHostMAC != "" && !h.Bootstrap {
+	if cluster != nil && m.Config.BootstrapHostMAC != "" && !h.Bootstrap {
 		for _, iface := range inventory.Interfaces {
 			if iface.MacAddress == m.Config.BootstrapHostMAC {
 				log.Infof("selected local bootstrap host %s for cluster %s", h.ID, cluster.ID)


internal/host/host.go
```diff
@@ -327,7 +327,7 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
 		return common.NewApiError(http.StatusNotFound, err)
 	}
 
-	if m.Config.BootstrapHostMAC != "" && !h.Bootstrap {
+	if cluster != nil && m.Config.BootstrapHostMAC != "" && !h.Bootstrap {
```
carbonin, line 330: Maybe it would be cleaner to refactor this a bit.

We could move the infraEnv query above the `if h.ClusterID != nil` block.
Then put this entire block inside that one.

That way it's very clear that this shouldn't run if we can't find a clusterID.

Alternatively it feels like an error case to have `BootstrapHostMAC` set and not have a clusterID. Maybe we should check for that explicitly and just fail? @zaneb what do you think?

CrystalChun: I did unset the clusterid while running [this unit test](https://github.com/openshift/assisted-service/blob/e27247fe07ace5f6212ea5896871faa84e329fe7/internal/host/host_test.go#L1176), sorry I guess I didn't fully understand late-binding and the usage of this Bootstrap MAC. 

If this scenario can't happen then I'll go ahead and close this issue and PR.

CrystalChun:
Closing for now since it's lower priority, but will reopen if there are cycles to refactor this bit of code.

PR #4130, opened 2022-07-13
MGMT-11156: Cluster validation, merge the print and validate methods

paul-maidment:
Same as https://github.com/openshift/assisted-service/commit/4e04230522992aad4d93551b8d935aa0d8b2ed0b but for cluster validations instead of host validations

The present implementation of cluster validation has a problem, it splits the validation into two phases. The validation and the printing of the validation message.

This leads to a situation in some cases where it is not possible to use the output of a validation to determine the message that should be sent back to the user, there are some workarounds but these involve poor practices such as code repetition.

This commit changes the validation to be a single method. Due to the complexity of the refactor, no effort has been made to make each of the validation methods their "most efficient" it is assumed that these fixes will take place in subsequent commits.

Because the validation methods made heavy use of return statements, the style has had to be changed to an if/else style and some code has had to be reorganised a little.

In theory, the existing tests should be sufficient to act as a "smoke" test for this functionality provided that there is enough coverage.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @tsorya 
/cc @ori-amizur 

## Checklist

## Reviewers Checklist


gamli75 (approved):

internal/cluster/cluster_suite_test.go
```diff
@@ -11,6 +12,7 @@ import (
 func TestCluster(t *testing.T) {
 	RegisterFailHandler(Fail)
 	common.InitializeDBTest()
+	time.Sleep(10 * time.Second)
```
nmagnezi, line 15: Is this intentional? or just for debugging purposes?

nmagnezi (commented):
The `hold` label is intentional?

internal/cluster/cluster_suite_test.go
paul-maidment: Just debugging, removed it now

paul-maidment: and unheld it as at the time of writing, the hold was intentional but no longer necessary

Dockerfile.assisted-service-build
```diff
@@ -28,4 +28,4 @@ COPY ./hack/setup_env.sh ./dev-requirements.txt ./
 RUN ./setup_env.sh podman_remote && ./setup_env.sh assisted_service
 RUN chmod g+xw -R $GOPATH && chmod g+xw -R $(go env GOROOT)
 COPY ./data /tmp/data
-COPY . .
\ No newline at end of file
+COPY . .
```
omertuc, line 31: ?

omertuc:
The commit message would be much more helpful if it were just "Same as 4e04230522992aad4d93551b8d935aa0d8b2ed0b but for cluster validations instead of host validations"

This way I don't have to think "Does this change make sense" but instead I can think "This was already done & approved for host validations, of course it makes sense we should also do it for cluster validations"

internal/cluster/validator.go
```diff
@@ -473,38 +355,22 @@ func isReadyToInstall(status string) bool {
 	return funk.ContainsString(allowedStatuses, status)
 }
 
-func (v *clusterValidator) allHostsAreReadyToInstall(c *clusterPreprocessContext) ValidationStatus {
+func (v *clusterValidator) allHostsAreReadyToInstall(c *clusterPreprocessContext) (ValidationStatus, string) {
 	readyToInstall := true
 	for _, host := range c.cluster.Hosts {
 		readyToInstall = readyToInstall && isReadyToInstall(swag.StringValue(host.Status))
 	}
```
carbonin, line 362:
```suggestion
	readyToInstall := true
	for _, host := range c.cluster.Hosts {
		if !isReadyToInstall(swag.StringValue(host.Status)) {
		        readyToInstall = false
		        break
		}
	}
```

I know this is unchanged, but it seems strange to loop through all of the hosts if you already know one isn't ready.

internal/cluster/validator.go
```diff
@@ -528,177 +379,127 @@ func checkCidrsOverlapping(cluster *common.Cluster) error {
...
+	serviceCidrDefined, _ = v.isServiceCidrDefined(c)
+
+	//If one of the required Cidr fields is empty return Pending status
+	if !validationStatusToBool(clusterCidrDefined) || !validationStatusToBool(serviceCidrDefined) {
+		status = ValidationPending
+	} else {
+		machineCidrDefined, _ := v.isMachineCidrDefined(c)
+		machineNetworkRequired := network.IsMachineNetworkRequired(c.cluster)
+		if machineNetworkRequired && !validationStatusToBool(machineCidrDefined) {
+			status = ValidationPending
```
carbonin, line 420: I think you could probably combine this with the message handling and return early, no?

`network.IsMachineNetworkRequired` is checking user managed networking so it'd imagine it's more accurate than the check below.

internal/cluster/validator.go
```diff
@@ -528,177 +379,127 @@ func checkCidrsOverlapping(cluster *common.Cluster) error {
...
-		return "No ntp problems found"
-	case ValidationFailure:
-		return fmt.Sprintf("Hosts' clocks are not synchronized (there's more than a %d minutes gap between clocks), "+
-			"please configure an NTP server via DHCP or set clocks manually.", common.MaximumAllowedTimeDiffMinutes)
-	default:
-		return fmt.Sprintf("Unexpected status %s", status)
+	if synced {
+		return ValidationSuccess, "No ntp problems found"
 	}
+	return ValidationFailure, fmt.Sprintf("Hosts' clocks are not synchronized (there's more than a %d minutes gap between clocks), "+
```
carbonin, line 503: Small nit, but whenever possible I think the success case should be the least indented. So I'd reverse this.
```suggestion
	if !synced {
		return ValidationFailure, fmt.Sprintf("Hosts' clocks are not synchronized (there's more than a %d minutes gap between clocks), "+
	}
	return ValidationSuccess, "No ntp problems found"
```

internal/cluster/cluster_suite_test.go
carbonin, line 15: Please remove this

internal/cluster/validator.go
```diff
@@ -402,58 +402,50 @@ func (v *clusterValidator) noCidrsOverlapping(c *clusterPreprocessContext) (Vali
...
+		v.log.WithError(err).Errorf("Getting service address families for cluster %s", c.cluster.ID.String())
+		return ValidationError, "Bad CIDR(s) appears in one of the networks"
+	}
+	clusterNetworkFamilies, err := network.CidrsToAddressFamilies(network.GetClusterNetworkCidrs(c.cluster))
+	if err != nil {
+		v.log.WithError(err).Errorf("Getting cluster address families for cluster %s", c.cluster.ID.String())
+		return ValidationError, "Bad CIDR(s) appears in one of the networks"
+	}
+	// serviceNetworkFamilies should only ever have a maximum of two distinct families.
+	// clusterNetworkFamilies may have multiple indistinct families so need to be canonized prior to comparison.
```
carbonin, line 432: Something I was confused about here was what "canonize" meant.
Maybe that's worth explaining too?

carbonin (changes requested):
New test cases look good, but the names and structure could be better

carbonin (approved):

PR #4135, opened 2022-07-14
MGMT-10789: Make sure that service can connect to KubeAPI during Day2 installation

paul-maidment:
Failures have been witnessed during Day2 installation that have been
caused by the absence of a valid KubeAPI connection from the service to
the spoke cluster.

This validation will execute whenever we add a host to a Day2 cluster to
ensure that a check is carried out as early as possible so that the user
can rectify the problem with appropriate DNS entries etc.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @filanov 
/cc @omertuc 

## Checklist

## Reviewers Checklist


internal/host/validator.go
```diff
@@ -1714,3 +1716,54 @@ func (v *validator) printVSphereUUIDEnabled(_ *validationContext, status Validat
...
+		_, err := net.DialTimeout("tcp", apiHostPort, 10*time.Second)
+		if err != nil {
+			status = ValidationFailure
+		}
+	}
+
+	return status
+}
+
+func (v *validator) printServiceCanConnectToKubeAPI(c *validationContext, status ValidationStatus) string {
```
omertuc, line 1754: Perhaps you want to merge your validations refactor PR before working on this?

internal/host/validator.go
```diff
@@ -1714,3 +1716,54 @@ func (v *validator) printVSphereUUIDEnabled(_ *validationContext, status Validat
...
+	var status ValidationStatus
+
+	// Check for the "Day2" status "adding-hosts"
+	addingHosts := models.ClusterStatusAddingHosts
+
+	// If we are not in day2 then this validation does not apply
+	// TODO: Only applies if Kube-API is enabled - we need to add this to the check...
+	if !c.kubeApiEnabled || (c.cluster != nil && c.cluster.Status != &addingHosts) {
+		status = ValidationSuccessSuppressOutput
+	} else {
```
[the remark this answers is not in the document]
paul-maidment: This was a deliberate style choice as when the "Validation merge" https://github.com/openshift/assisted-service/pull/4097 PR is on master, this will merge the validation and print methods. At which point returning in this way will make less sense.

internal/host/validator.go
paul-maidment: Agreed

paul-maidment: Just wanted to get a good feel of how this could be solved, at this stage it should really just be small changes to make it work.

internal/host/validator.go
```diff
@@ -1818,3 +1820,50 @@ func (v *validator) printCompatibleAgent(c *validationContext, status Validation
...
+func (v *validator) serviceCanConnectToSpokeKubeAPI(c *validationContext) ValidationStatus {
+	// If we do not have a cluster then we definitely are not in Day2 status!
+	if c.cluster == nil {
+		return ValidationSuccessSuppressOutput
+	}
+
+	var status ValidationStatus
+
+	// If we are not in day2 then this validation does not apply
+	if !c.kubeApiEnabled || (c.cluster != nil && *c.cluster.Status != models.ClusterStatusAddingHosts) {
```
omertuc, line 1833: redundant cluster nil check

internal/host/validator.go
```diff
@@ -1626,3 +1628,21 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
 	}
 	return ValidationSuccess, successMessage
 }
+
+func (v *validator) serviceCanConnectToSpokeKubeAPI(c *validationContext) (ValidationStatus, string) {
+	// This only applies to scenarios where we have a cluster, kube-api is enabled and the cluster is in "Day2" mode.
```
omertuc, line 1633: >	// This only applies to scenarios where we have a cluster, kube-api is enabled and the cluster is in "Day2" mode.

That much is apparent from the condition, but this comment doesn't explain why it only applies in that situation

internal/host/validator.go
```diff
@@ -1626,3 +1628,21 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+	if c.cluster == nil || !c.kubeApiEnabled || *c.cluster.Status != models.ClusterStatusAddingHosts {
+		return ValidationSuccessSuppressOutput, "Host is not part of a Day2 cluster, no need to check KubeAPI connectivity"
+	}
+
+	// Validate that the service can connect to kube-api - this will be needed for some Day2 operations.
+	// This will only verify host/port connectivity.
+	apiAddress := common.GetAPIHostname(c.cluster)
+	apiHostPort := net.JoinHostPort(apiAddress, "6443")
+	_, err := net.DialTimeout("tcp", apiHostPort, 10*time.Second)
```
omertuc, line 1642: What is the reason we do a simple TCP connection instead of actually checking we have full, privileged API access?

paul-maidment: I guess we should, I tried to keep things quick and simple here but for accuracy, we should make sure we can make an API call. I will look into this.

internal/host/validator.go
```diff
@@ -1626,3 +1635,73 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
 	}
 	return ValidationSuccess, successMessage
 }
+
+func (v *validator) serviceCanConnectToSpokeKubeAPI(c *validationContext) (ValidationStatus, string) {
```
filanov, line 1639: What is the performance of this validation? 
Imagine managing 2K day2 clusters

filanov: Actually it's a validation that  will run only before we start day2 installation so maybe it's fine. 
I confused it with a validation on the cluster and not on the host.

filanov (commented):
A small comment about comments in the code. 
I prefer the code to be clear and there should be a good reason to have comments, for example explaining why specific things is done and not how. 
In addition the parameters should be clear enough so that you will not need to describe them 
for example 
```
	// Define a buffer to hold the kubeconfig
	buffer := make([]byte, contentLength)
```
can be 
```
	kubeconfigbuff := make([]byte, contentLength)
```

internal/host/validator.go
```diff
@@ -1626,3 +1635,73 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+	// Calling ReadAtLeast method with its parameters
+	bytesRead, err := io.ReadAtLeast(respBody, buffer, int(contentLength))
+
+	// If error is not nil then panics
+	if err != nil {
+		return ValidationError, fmt.Sprintf("Could not read spoke cluster kubeconfig from internal storage: %s", err.Error())
+	}
+	if bytesRead > int(contentLength) {
+		return ValidationError, "Too many bytes read when reading spoke cluster kubeconfig from internal storage"
+	}
```
omertuc, line 1664: This function is getting crowded, this should be extracted into a method

internal/host/validator.go
```diff
@@ -1626,3 +1635,73 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+            },
+        },
+	}
+	sarResponse, err = client.CreateSubjectAccessReview(csrApproveSAR)
+	if  err != nil {
+		return ValidationFailure, "Could not create subject access review to detemine csr approve permissions"
+	}
+	if !sarResponse.Status.Allowed {
+		return ValidationFailure, "The spoke does not permit the approval of CSRs"
+	}
```
omertuc, line 1705: This too, should be extracted into a method

internal/host/validator.go
```diff
@@ -1626,3 +1635,73 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+            ResourceAttributes: &authorizationv1.ResourceAttributes{
+                Namespace: namespace,
+                Verb:      "list",
+                Resource:  "CertificateSigningRequest",
+            },
+        },
+	}
+	sarResponse, err := client.CreateSubjectAccessReview(csrListSAR)
+	if  err != nil {
+		return ValidationFailure, "Could not create subject access review to detemine csr list permissions"
```
omertuc, line 1685: This is a ValidationError, not ValidationFailure

omertuc: Also could be nice to include the error itself in a ValidationError

internal/host/validator.go
```diff
@@ -1626,3 +1635,73 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+	}
+	if !sarResponse.Status.Allowed {
+		return ValidationFailure, "The spoke does not permit the listing of CSRs"
```
omertuc, line 1688: Remember this is a user-facing message. I wouldn't say the "the spoke doesn't permit the listing of CSRs" as that is not true in the general sense. I think the message should be more specific - "The cluster credentials available to the Assisted Installer do not have permissions to <verb> <resource>s on the spoke cluster"

swagger.yaml
```diff
@@ -5768,6 +5768,8 @@ definitions:
       - 'compatible-agent'
       - 'no-skip-installation-disk'
       - 'no-skip-missing-disk'
+      - 'service-can-connect-to-spoke-kube-api'
```
omertuc, line 5771: Perhaps now that the validation is more specific it should be renamed - something like `service-has-sufficient-spoke-kube-api-access`

internal/host/validator.go
```diff
@@ -1626,3 +1636,34 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+		{"list", "CertificateSigningRequest"},
+		{"approve", "CertificateSigningRequest"},
+		{"list", "Node"},
+	} {
+		state, err := client.IsActionPermitted(check[0], check[1])
+		if err != nil {
+			return ValidationError, err.Error()
+		}
+		if !state {
+			return ValidationFailure, fmt.Sprintf("%s of %s is not permitted", check[0], check[1])
```
omertuc, line 1663: not permitted to who? give the user some context about what's happening - this is the only error they're going to see

internal/host/validator.go
```diff
@@ -1626,3 +1634,38 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+		{"list", "CertificateSigningRequest"},
+		{"approve", "CertificateSigningRequest"},
+		{"list", "Node"},
+	} {
+		state, err := client.IsActionPermitted(check[0], check[1])
+		if err != nil {
+			return ValidationError, err.Error()
+		}
+		if !state {
+			return ValidationFailure, fmt.Sprintf("the spoke cluster credentials provided to assisted-service do not have permissions to %s %s on the spoke cluster", check[0], check[1])
```
omertuc, line 1665: The credentials are (as of today) not "provided" to the assisted-service and the user has no way to change them. Perhaps "held by" is a better fit than "provided to"

omertuc: `to %s %s` sounds odd when the latter %s is singular, so perhaps change the loop iteratee to the plural (I don't think SAR would mind, but I'm not entirely sure)

internal/host/validator.go
```diff
@@ -1626,3 +1634,38 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
 	}
 	return ValidationSuccess, successMessage
 }
+
+
+
+func (v *validator) serviceCanConnectToSpokeKubeAPI(c *validationContext) (ValidationStatus, string) {
+	// When assisted-service operates with kubeApiEnabled for a day2 cluster, it will automatically sign CSR's by making a connection from the assisted-service to the spoke kubeApi.
```
omertuc, line 1641:
```suggestion
	// When assisted-service operates in kube-api mode, when adding a day2 host, when CSRs are not otherwise automatically approved, the service will automatically attempt to approve CSRs by using its stored spoke kubeconfig 
```

omertuc: Which also means we only need this validation when the spoke is UMN

omertuc: However even in non-UMN we (shoulud) track the installation progress using List nodes, so at-least those permissions are always needed regardless of UMN

omertuc: Perhaps this entire text should be simplified to something a bit less specific like "When adding a day2 host to a spoke cluster and the service operates in kube-API mode, the service needs to access the spoke cluster's kube API in order to perform some operations, such as listing nodes and approving CSRs (depending on platform). This validation makes sure in advance that the service is authorized to do that"

internal/host/validator.go
paul-maidment: You can just chuck an extra s into the message rather than do that

`return ValidationFailure, fmt.Sprintf("the spoke cluster credentials provided to assisted-service do not have permissions to %s %ss on the spoke cluster", check[0], check[1]) `

Because all of the entities in the list are perfectly valid to express as plural in this messaging. No need to change what we send to kubernetes.

internal/spoke_k8s_client/spoke_k8s_client.go
```diff
@@ -0,0 +1,219 @@
...
+                Resource:  resource,
+            },
+        },
+	}
+	sarResponse, err := c.CreateSubjectAccessReview(&sar)
+	if err != nil {
+		return false, fmt.Errorf("could not create subject access review to detemine %s %s permissions %w", verb, resource, err)
+	}
+
+	if (sarResponse.Status.Allowed) {
```
omertuc, line 174: no parenthesis around `if` conditions

internal/spoke_k8s_client/spoke_k8s_client.go
```diff
@@ -0,0 +1,219 @@
...
+func (cf *spokeK8sClientFactory) CreateFromKubeconfig(kubeconfig []byte) (SpokeK8sClient, error) {
+	clientConfig, err := cf.getRestConfigFromKubeConfig(kubeconfig)
+	if err != nil {
+		cf.log.WithError(err).Warnf("Getting client from kubeconfig cluster")
+		return nil, err
+	}
+	return cf.createFromClientConfig(clientConfig)
+}
+
+func (cf *spokeK8sClientFactory) CreateFromSpokeKubeconfig(ctx context.Context, clusterId *strfmt.UUID, objectHandler s3wrapper.API) (SpokeK8sClient, error) {
```
omertuc, line 92:
```suggestion
func (cf *spokeK8sClientFactory) CreateFromStorageKubeconfig(ctx context.Context, clusterId *strfmt.UUID, objectHandler s3wrapper.API) (SpokeK8sClient, error) {
```

internal/spoke_k8s_client/spoke_k8s_client.go
```diff
@@ -0,0 +1,219 @@
...
+	kubeConfigReader, contentLength, err := objectHandler.Download(ctx, fmt.Sprintf("%s/%s", clusterId, constants.Kubeconfig))
+	if err != nil {
+		return nil, fmt.Errorf("could not load kubeconfig from internal storage with cluster id %s and filename %s", clusterId, constants.Kubeconfig)
```
omertuc, line 95: unused `err`

internal/spoke_k8s_client/spoke_k8s_client.go
```diff
@@ -0,0 +1,219 @@
...
+	log logrus.FieldLogger
+}
+
+func NewSpokeK8sClientFactory(log logrus.FieldLogger) SpokeK8sClientFactory {
+	return &spokeK8sClientFactory{
+		log: log,
+	}
+}
+
+func (cf *spokeK8sClientFactory) Create(secret *corev1.Secret) (SpokeK8sClient, error) {
```
omertuc, line 74:
```suggestion
func (cf *spokeK8sClientFactory) CreateFromSecret(secret *corev1.Secret) (SpokeK8sClient, error) {
```

internal/spoke_k8s_client/spoke_k8s_client.go
```diff
@@ -0,0 +1,219 @@
...
+	clientConfig, err := cf.getRestConfigFromSecret(secret)
+	if err != nil {
+		cf.log.WithError(err).Warnf("Getting client from kubeconfig cluster")
+		return nil, err
+	}
+	return cf.createFromClientConfig(clientConfig)
+}
+
+func (cf *spokeK8sClientFactory) CreateFromKubeconfig(kubeconfig []byte) (SpokeK8sClient, error) {
```
omertuc, line 83:
```suggestion
func (cf *spokeK8sClientFactory) CreateFromRawKubeconfig(kubeconfig []byte) (SpokeK8sClient, error) {
```

cmd/main.go
```diff
@@ -702,14 +703,8 @@ func createStorageClient(deployTarget string, storage string, s3cfg *s3wrapper.C
 		switch storage {
 		case storage_s3:
 			storageClient = s3wrapper.NewS3Client(s3cfg, log)
-			if storageClient == nil {
```
omertuc, line 705: why were these removed

paul-maidment: the compiler indicated that the value of storageclient could never be nil and that these checks were not required.

paul-maidment: I can double check this and put them back if needed

internal/host/validator.go
```diff
@@ -1626,3 +1634,35 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+		{"list", "CertificateSigningRequests"},
+		{"approve", "CertificateSigningRequests"},
+		{"list", "Nodes"},
+	} {
+		state, err := client.IsActionPermitted(check[0], check[1])
+		if err != nil {
+			return ValidationError, err.Error()
+		}
+		if !state {
+			return ValidationFailure, fmt.Sprintf("the spoke cluster credentials held by assisted-service do not have permissions to %s %s on the spoke cluster", check[0], check[1])
```
omertuc, line 1663:
```suggestion
			return ValidationFailure, fmt.Sprintf("the spoke cluster credentials held by the Assisted Installer do not have permissions to %s %s on the spoke cluster", check[0], check[1])
```

The product's name is Assisted Installer, assisted-service is a component and an implementation detail the user should not be exposed to, ideally

internal/host/validator.go
```diff
@@ -1626,3 +1634,35 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+	if err != nil {
+		return ValidationFailure, "Could not create the spoke k8s client connection using kubeconfig"
+	}
+
+	for _, check := range [][]string{
+		{"list", "CertificateSigningRequests"},
+		{"approve", "CertificateSigningRequests"},
+		{"list", "Nodes"},
+	} {
+		state, err := client.IsActionPermitted(check[0], check[1])
```
omertuc, line 1658: Some aliases could make the rest of the code nicer to read and emphasize the intention / meaning of those strings

```suggestion
		verb, resource := check[0], check[1]
		state, err := client.IsActionPermitted(verb, resource)
```

internal/host/validator.go
```diff
@@ -1626,3 +1634,35 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+	// such as listing nodes and approving CSRs (depending on platform).
+	// This validation makes sure in advance that the service is authorized to do that.
+	if c.cluster == nil || !c.kubeApiEnabled || *c.cluster.Status != models.ClusterStatusAddingHosts {
+		return ValidationSuccessSuppressOutput, "Host is not part of a Day2 cluster, no need to check KubeAPI connectivity"
+	}
+
+	spokeK8sClientFactory := spoke_k8s_client.NewSpokeK8sClientFactory(v.log)
+	client, err := spokeK8sClientFactory.CreateFromStorageKubeconfig(context.TODO(), c.host.ClusterID, c.objectHandler)
+	if err != nil {
+		return ValidationFailure, "Could not create the spoke k8s client connection using kubeconfig"
```
omertuc, line 1650: This is a ValidationError, not failure. Also should probably wrap the error to aid in troubleshooting what went wrong

PR #4139, opened 2022-07-14
MGMT-10734: Add more host reclaim states and transitions

carbonin:
These were missed in 9fab7bd07fbaa9de1bdcecc4563f17a815cc1c9b

* Ensures that the reclaim transition works the same as
unbind in the states that are not particularly interesting for the new
work. This means that any time that we don't need to reboot the host it should
act exactly the same as unbind already does.

* Adds a transition type to be used when we fail the reclaim
process. Currently this will bring a host from either `Reclaiming` or
`ReclaimingRebooting` to `UnbindingPendingUserAction`

* Adds a transition to timeout hosts in the reclaim states from the
monitor

https://issues.redhat.com/browse/MGMT-10734

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

This just adds new states and transitions, none of these are currently accessible.

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @CrystalChun 
/cc @filanov 
/cc @tsorya 

## Checklist

## Reviewers Checklist


CrystalChun (approved):
/lgtm

ori-amizur:
Unit tests ?

carbonin:
> Unit tests ?

There are not currently any ways for a host to enter any of these states. And no code exists to run the related transitions.
I will add the tests when I add that code.

Also there are no existing unit tests for `NewHostStateMachine`.
Do you think we should add some for the state machine itself separate from the code that we expect to actually run the transitions?

internal/host/statemachine.go
```diff
@@ -265,6 +266,40 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 		PostTransition:   th.PostUnbindHost,
 	})
 
+	sm.AddTransition(stateswitch.TransitionRule{
```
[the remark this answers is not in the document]
carbonin: Added.

internal/host/statemachine.go
```diff
@@ -255,16 +257,69 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
...
+		TransitionType: TransitionTypeRebootingForReclaim,
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusReclaiming),
+		},
+		DestinationState: stateswitch.State(models.HostStatusReclaimingRebooting),
+	})
+
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeReclaimFailed,
```
filanov, line 280: who is going to use this transition?

skipper.env
```diff
@@ -40,3 +40,4 @@ MINIKUBE_HOME
 OPENSHIFT_VERSION
 ENABLE_ORG_TENANCY
 ENABLE_ORG_BASED_FEATURE_GATES
+TEST
```
filanov, line 43: what is that for?

internal/host/statemachine.go
carbonin: This would be used when the agent returns an error from one of the steps in this process.

skipper.env
carbonin: >This will allow specifying a package to test

It's in the commit message :wink: 

It was easier for me to specify the package and then focus the tests I wanted in the code rather than using the focus regex on the command line when running tests.

internal/host/statemachine.go
filanov: handling get step reply?

skipper.env
filanov: i'm reviewing everything together so i looked at the description not the commit message, thanks

internal/host/statemachine.go
carbonin: Yes that's the case, but that handling isn't being implemented in this PR. I just opened https://issues.redhat.com/browse/MGMT-11335 yesterday for that work.

carbonin: I intended this PR to just be the states and transitions.

I added the monitor stuff because it was fairly minor (two lines and some tests), but I can separate that out if you'd prefer?

carbonin: I also updated the diagram in the enhancement to reflect this state's usage and the monitor change, hopefully that helps.

carbonin:
@tsorya @ori-amizur could one of you take a look at this since @filanov is going to be out for a few weeks.
Thanks.

ori-amizur:
The transitions tests are in transition_test.go

carbonin:
> The transitions tests are in transition_test.go

Those are not testing the transitions directly. They're testing functions like `RegisterHost`, `HandleInstallationFailure`, or `RefreshStatus`.

As I said in my previous comment, if we want to add new tests that actually test the state machine rather than code that calls out to the state machine then this would be testable, but we don't currently have tests like that.

tsorya:
lgtm as for me

tsorya:
maybe squash commits?

internal/host/statemachine.go
```diff
@@ -255,16 +257,69 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
...
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusReclaiming),
+			stateswitch.State(models.HostStatusReclaimingRebooting),
```
ori-amizur, line 283: Once we are in HostStatusReclaimingRebooting, how can a failure occur ?  Isn't it just timeout ?

internal/host/statemachine.go
```diff
@@ -255,16 +257,69 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
...
+		},
+		DestinationState: stateswitch.State(models.HostStatusUnbindingPendingUserAction),
+		PostTransition:   th.PostUnbindHost,
+	})
+
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeRefresh,
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusReclaiming),
```
ori-amizur, line 292: Do they have the same timeout (create pod, boot for discovery ISO)?

internal/host/statemachine.go
```diff
@@ -255,16 +257,69 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
...
+		},
+		Condition:        th.HasStatusTimedOut(ReclaimTimeout),
+		DestinationState: stateswitch.State(models.HostStatusUnbindingPendingUserAction),
+		PostTransition:   th.PostRefreshReclaimTimeout,
+	})
+
+	// ReclaimHost in other states acts like Unbind
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeReclaimHost,
+		SourceStates: []stateswitch.State{
```
ori-amizur, line 303: Also for installing host before rebooting

internal/host/host.go
```diff
@@ -58,6 +58,8 @@ var WrongBootOrderIgnoreTimeoutStages = []models.HostStage{
 
 var InstallationTimeout = 20 * time.Minute
 
+var ReclaimTimeout = 60 * time.Minute
```
ori-amizur, line 61: My opinion is that reclaim timeout should be shorter.
Pod creation should be quite short.
Booting from discovery ISO is not dependent on getting ignition from MCS, so it should be also quite short.
Also, maybe consider making these timeouts separate.

carbonin: > Booting from discovery ISO is not dependent on getting ignition from MCS

No, but it does pull the rootfs, which is quite a bit larger.

Also the step to do the reclaim will pull the initrd and kernel from the service which are smaller, but still something we need to take into account.

Do you have any suggestions for times?

internal/host/statemachine.go
```diff
@@ -255,16 +257,69 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 		PostTransition:   th.PostUnbindHost,
 	})
 
-	// Reclaim host
+	// ReclaimHost when installed moves to Reclaiming
 	sm.AddTransition(stateswitch.TransitionRule{
 		TransitionType: TransitionTypeReclaimHost,
 		SourceStates: []stateswitch.State{
 			stateswitch.State(models.HostStatusInstalled),
+			stateswitch.State(models.HostStatusAddedToExistingCluster),
```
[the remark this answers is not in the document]
carbonin: What's the difference between joined and these states?
Should we be able to see the node using the spoke kubeconfig and schedule a pod on it?

internal/host/statemachine.go
carbonin: The agent could fail to `chroot` or not have required permissions to run reboot.

internal/host/statemachine.go
carbonin: Currently, yes. The reclaim timeout is being passed to the condition a few lines below this.

internal/host/statemachine.go
carbonin: :man_shrugging: I took this from [the transition for `UnbindHost`](https://github.com/openshift/assisted-service/blob/1f7410a2d43220908c4ef1b136d2075ac284ec96/internal/host/statemachine.go#L232-L256) so if we don't support that state for unbind, I don't think we should here.

carbonin:
> maybe squash commits?

I'd prefer not to. I think separating out the addition to the monitor from the state definition, for example, makes it easier to understand the patch.

It's a shame the bot will squash these when it merges anyway.

internal/host/statemachine.go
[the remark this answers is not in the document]
ori-amizur: It can be seen (node exists) - but the node is not ready yet

internal/host/statemachine.go
ori-amizur: OK

internal/host/statemachine.go
```diff
@@ -255,16 +257,69 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
...
 		DestinationState: stateswitch.State(models.HostStatusReclaiming),
 		PostTransition:   th.PostUnbindHost,
 	})
 
+	sm.AddTransition(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeRebootingForReclaim,
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusReclaiming),
+		},
+		DestinationState: stateswitch.State(models.HostStatusReclaimingRebooting),
```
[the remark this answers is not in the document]
carbonin: :man_facepalming: That's irritating. One would think a state transition would update the state.
Thanks for catching this.

internal/host/host.go
ori-amizur: It seemed to me that it needs to be shorter, but maybe you know better.
Anyway, they should be separate.  Pod creation should not take a long time (I think).

internal/host/statemachine.go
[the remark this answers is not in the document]
carbonin: If by "ready" you mean it's not schedulable then no, I don't think we should add joined.

internal/host/host.go
carbonin: I definitely don't know better.

I left this as long as it is because I have no idea how long these operations might take.
Given that we'll never succeed after we time out I felt that more time was better at least on the first pass.

If it turns out that this times out fairly frequently in a way that we can't do anything about, then maybe it will make sense to split or decrease the time, but for now I'd rather give it as much time as I can. What difference will timing out 10 or 20 minutes earlier really make?

internal/host/statemachine.go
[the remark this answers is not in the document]
ori-amizur: ok

PR #4128, opened 2022-07-13
MGMT-11131: Track hosts' registration times

mkowalski:
This PR introduces a new host property RegisteredAt in order to track
all the registration attemps. This is because an already existing host
can e.g. reboot what would trigger a new registration attempt that
currently is not tracked as a property of the host even though it
triggers an event.

This is going to be helpful as there are scenarios when we want to know
if the host rebooted after the initial registration. With this change it
will be as simple as comparing timestamps of CreatedAt and RegisteredAt.

Contributes-to: [MGMT-11131](https://issues.redhat.com//browse/MGMT-11131)

## List all the issues related to this PR

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [x] Automation (CI, tools, etc)
- [x] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] Reviewers have been listed
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #4140, opened 2022-07-15
[WIP] MGMT-11131: Use host's registration time as key for cache

mkowalski:
It has been discovered that hosts can get different IP addresses when
they reboot post-discovery. Currently our connectivity validations use
only Host ID and InfraEnv ID as a key and never invalidate the entry.
This causes problems if the host changes its IP address, as the
validation can still rely on the old inventory.

Given that the cache exists inside the validator and is not connected
with the inventory and also the fact that unmarshalling the inventory is
an expensive operation, we decided to additionally use registration
timestamp as a key for the cache.

This limits the exposure to the described issue as with this change
whenever the host reboots it will use a new entry in the cache. Doing so
allows us to avoid redesigning the cache mechanism (the performance hit
for invalidating on every inventory update is too big) while solving the
observed issue.

Please note that if for any reason host changes its IP address without
rebooting or restarting the agent, the wrong cache may still be used.
This has been discussed internally and is at the moment accepted to be a
known issue.

Closes: [MGMT-11131](https://issues.redhat.com//browse/MGMT-11131)

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -370,7 +370,7 @@ func CanonizeStrings(slice []string) (ret []string) {
 }
 
 func GetHostKey(host *models.Host) string {
-	return host.ID.String() + "@" + host.InfraEnvID.String()
+	return host.ID.String() + "@" + host.InfraEnvID.String() + "@" + host.RegisteredAt.String()
 }
 
 // GetTagFromImageRef returns the tag of the given container image reference. For example, if the


mkowalski:
We need https://github.com/openshift/assisted-service/pull/4128 to land in order to rebase on top of it. Master branch if frozen at the moment, therefore `[WIP]`

carbonin:
cc @ori-amizur 

Also want to make you aware.

ori-amizur:
This is not a correct fix to the problem.  I think we should fix the invalidation and not change the key.  The host key should not depend on the registration time.

mkowalski:
> This is not a correct fix to the problem. I think we should fix the invalidation and not change the key. The host key should not depend on the registration time.

The way cache has been implemented in https://github.com/openshift/assisted-service/pull/3804 makes it impossible to easily invalidate it from the Host's context, that's the first issue that has been spotted. We cannot implement the following use cases

* "I want to invalidate the cache when Host reregisters"
* "I want to invalidate the cache when Host changes its IP address'

The main issue though, the most complex - NICs and IP addresses are stored inside inventory. The cache has been introduced because unmarshalling inventory is expensive. If we want to invalidate cache when IP address changes, then we need to unmarshall inventory to detect it. QED, we have just made a cycle.

avishayt:
Now that we removed the timestamp and full smartctl output from the inventory, it should not change often.  In that case, the agent should notify the service only for real changes.  Is that not the case?
If it is the case, the service can cache the unmarshalled inventory and invalidate it upon every update.

mkowalski:
/close

After offline discussions, superseded by https://github.com/openshift/assisted-service/pull/4170

PR #4150, opened 2022-07-18
MGMT-8818: clusterdeployments: set HTTPS Ignition URL when finalizing the cluster

vrutkovs:
When possible, fetch MCS certificate from spoke cluster and use it to provision day2 workers via HTTPS Ignition endpoint.

- [x] Enhancement 

## How was this code tested?

TODO:
* [x] atm the code assumes that if IgnitionOverrides has CA its the CA set by BMH controller (it fetches MCS cert after cluster is transitioned to day 2) and its an operator-backed instance. So, there is at least one usecase where this might have a false-positive case:
* * hub cannot contact spoke cluster and won't add MCS cert to ignition override
* * user has set Ignition override with CA manually

In that case day2 worker would attempt to use HTTPS and fail as Ignition server cannot be contacted using user-provided CA.
Do we consider this usecase likely to happen and serious enough to fix?

Discussed with mko and omer - hub->spoke connection is a requirement on day2, so we can rely on having MCS cert in ignition override. Users are rarely setting host-specific overrides for CA (this would be replaced by a new API later), so it looks like a low-risk assumption to me

* [x] add unit tests

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -862,6 +862,10 @@ func (b *bareMetalInventory) createAndUploadDay2NodeIgnition(ctx context.Context
 	if err != nil {
 		return errors.Wrapf(err, "Failed to build ignition endpoint for host %s in cluster %s", host.ID, cluster.ID)
 	}
+	ignitionEndpointUrl, err = b.UpdateIgnitionEndpointIfHasMCSCert(log, host, cluster, ignitionEndpointUrl)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to build ignition endpoint for host %s in cluster %s", host.ID, cluster.ID)
+	}
 
 	var caCert *string = nil
 	if cluster.IgnitionEndpoint != nil {
@@ -880,6 +884,31 @@ func (b *bareMetalInventory) createAndUploadDay2NodeIgnition(ctx context.Context
 		return errors.Errorf("Failed to upload worker ignition for cluster %s", cluster.ID)
 	}
 	return nil
+}
+
+func (b *bareMetalInventory) UpdateIgnitionEndpointIfHasMCSCert(log logrus.FieldLogger, host *models.Host, cluster *common.Cluster, ignitionEndpointUrl string) (string, error) {
+	// No ignition override set by bmh agent controller
+	if host.IgnitionConfigOverrides == "" {
+		return ignitionEndpointUrl, nil
+	}
+	// No CA certs set in ignition override
+	if !ignition.HasCACertInIgnition(host.IgnitionConfigOverrides) {
+		return ignitionEndpointUrl, nil
+	}
+	// Not an operator-backed deployment
+	if b.k8sClient == nil {
+		return ignitionEndpointUrl, nil
+	}
+	log.Infof("Found CA certificate in ignition override for cluster %s, host %s", cluster.ID, host.ID)
+	url, err := url.Parse(ignitionEndpointUrl)
+	if err != nil {
+		return ignitionEndpointUrl, err
+	}
+	url.Scheme = "https"
+	url.Host = fmt.Sprintf("%s.%s.%s:%d", constants.InternalAPIClusterSubdomain, cluster.Name, cluster.BaseDNSDomain, constants.SecureMCSPort)
+	ignitionEndpointUrl = url.String()
+	log.Infof("Resetting Ignition endpoint URL to %s for %s, host %s", ignitionEndpointUrl, cluster.ID, host.ID)
+	return ignitionEndpointUrl, nil
 }
 
 func (b *bareMetalInventory) integrateWithAMSClusterDeregistration(ctx context.Context, cluster *common.Cluster) error {
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -15253,3 +15253,88 @@ var _ = Describe("GetHostByKubeKey", func() {
 		Expect(err).To(HaveOccurred())
 	})
 })
+
+var _ = Describe("UpdateIgnitionEndpointIfHasMCSCert", func() {
+	var (
+		bm      *bareMetalInventory
+		cfg     Config
+		db      *gorm.DB
+		dbName  string
+		cluster *common.Cluster
+		log     logrus.FieldLogger
+	)
+	const (
+		httpIgnitionEndpointUrl  = "http://api.foo.bar.com:22624/config/custom-pool"
+		httpsIgnitionEndpointUrl = "https://api-int.foo.bar.com:22623/config/custom-pool"
+		masterIgn                = `{
+		  "ignition": {
+		    "config": {
+		      "merge": [
+			{
+			  "source": "https://192.168.126.199:22623/config/master"
+			}
+		      ]
+		    },
+		    "security": {
+		      "tls": {
+			"certificateAuthorities": [
+			  {
+			    "source": "data:text/plain;charset=utf-8;base64,LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURFRENDQWZpZ0F3SUJBZ0lJUk90aUgvOC82ckF3RFFZSktvWklodmNOQVFFTEJRQXdKakVTTUJBR0ExVUUKQ3hNSmIzQmxibk5vYVdaME1SQXdEZ1lEVlFRREV3ZHliMjkwTFdOaE1CNFhEVEl3TURreE9ERTVORFV3TVZvWApEVE13TURreE5qRTVORFV3TVZvd0pqRVNNQkFHQTFVRUN4TUpiM0JsYm5Ob2FXWjBNUkF3RGdZRFZRUURFd2R5CmIyOTBMV05oTUlJQklqQU5CZ2txaGtpRzl3MEJBUUVGQUFPQ0FROEFNSUlCQ2dLQ0FRRUE1c1orVWtaaGsxUWQKeFU3cWI3YXArNFczaS9ZWTFzZktURC8ybDVJTjFJeVhPajlSL1N2VG5SOGYvajNJa1JHMWN5ZXR4bnNlNm1aZwpaOW1IRDJMV0srSEFlTTJSYXpuRkEwVmFwOWxVbVRrd3Vza2Z3QzhnMWJUZUVHUlEyQmFId09KekpvdjF4a0ZICmU2TUZCMlcxek1rTWxLTkwycnlzMzRTeVYwczJpNTFmTTJvTEM2SXRvWU91RVVVa2o0dnVUbThPYm5rV0t4ZnAKR1VGMThmNzVYeHJId0tVUEd0U0lYMGxpVGJNM0tiTDY2V2lzWkFIeStoN1g1dnVaaFYzYXhwTVFMdlczQ2xvcQpTaG9zSXY4SWNZbUJxc210d2t1QkN3cWxibEo2T2gzblFrelorVHhQdGhkdWsrZytzaVBUNi9va0JKU2M2cURjClBaNUNyN3FrR3dJREFRQUJvMEl3UURBT0JnTlZIUThCQWY4RUJBTUNBcVF3RHdZRFZSMFRBUUgvQkFVd0F3RUIKL3pBZEJnTlZIUTRFRmdRVWNSbHFHT1g3MWZUUnNmQ0tXSGFuV3NwMFdmRXdEUVlKS29aSWh2Y05BUUVMQlFBRApnZ0VCQU5Xc0pZMDY2RnNYdzFOdXluMEkwNUtuVVdOMFY4NVJVV2drQk9Wd0J5bHluTVRneGYyM3RaY1FsS0U4CjVHMlp4Vzl5NmpBNkwzMHdSNWhOcnBzM2ZFcUhobjg3UEM3L2tWQWlBOWx6NjBwV2ovTE5GU1hobDkyejBGMEIKcGNUQllFc1JNYU0zTFZOK0tZb3Q2cnJiamlXdmxFMU9hS0Q4dnNBdkk5YXVJREtOdTM0R2pTaUJGWXMrelRjSwphUUlTK3UzRHVYMGpVY001aUgrMmwzNGxNR0hlY2tjS1hnUWNXMGJiT28xNXY1Q2ExenJtQ2hIUHUwQ2NhMU1MCjJaM2MxMHVXZnR2OVZnbC9LcEpzSjM3b0phbTN1Mmp6MXN0K3hHby9iTmVSdHpOMjdXQSttaDZ6bXFwRldYKzUKdWFjZUY1SFRWc0FkbmtJWHpwWXBuek5qb0lFPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg=="
+			  }
+			]
+		      }
+		    },
+		    "version": "3.2.0"
+		  },
+		  "storage": {}
+		}`
+	)
+
+	BeforeEach(func() {
+		db, dbName = common.PrepareTestDB()
+		bm = createInventory(db, cfg)
+		cluster = createClusterWithAvailability(db, models.ClusterStatusReady, models.ClusterCreateParamsHighAvailabilityModeNone)
+		cluster.Name = "foo"
+		cluster.BaseDNSDomain = "bar.com"
+		// this doesn't need to be a VM, but any host works for this test
+		addVMToCluster(cluster, db)
+		log = logrus.New()
+	})
+
+	AfterEach(func() {
+		common.DeleteTestDB(db, dbName)
+		ctrl.Finish()
+	})
+
+	It("no ignition overrides for host set", func() {
+		h := cluster.Hosts[0]
+		ignitionEndpointUrl, err := bm.UpdateIgnitionEndpointIfHasMCSCert(log, h, cluster, httpIgnitionEndpointUrl)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(ignitionEndpointUrl).To(Equal(httpIgnitionEndpointUrl))
+	})
+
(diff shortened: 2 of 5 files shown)


internal/common/common.go
```diff
@@ -392,3 +393,11 @@ func GetTagFromImageRef(ref string) string {
 		return ""
 	}
 }
+
+func GetMCSUrlBase(cluster *Cluster) string {
+	// URL needs to use api-int DNS name if APIVIP is not set (i.e. platform:none)
+	if cluster.APIVip == "" {
+		return fmt.Sprintf("https://%s.%s:22623", constants.InternalAPIClusterSubdomain, cluster.BaseDNSDomain)
```
eranco74, line 400: I think you are missing cluster name here, no?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -225,6 +227,19 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 		}
 	}
 
+	// Set HTTPS Ignition URL
+	if *cluster.Status == models.ClusterStatusFinalizing {
+		mcsSecretKey, err1 := r.createMCSCertSecret(ctx, log, clusterDeployment, cluster, clusterInstall)
```
eranco74, line 232: err1 -> err

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -225,6 +227,19 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 		}
 	}
 
+	// Set HTTPS Ignition URL
+	if *cluster.Status == models.ClusterStatusFinalizing {
```
eranco74, line 231: We need to do it once, and only if ignitionEndpoint isn't already set.
There is no reason to keep getting the MCS cert from the spoke if the ignitionEndpoint is set.

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -241,6 +256,91 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
...
+		Namespace: clusterDeployment.Namespace,
+		Name:      getMCSSecretName(clusterDeployment),
+	}
+	log.Debugf("Expected key: %s", key)
+
+	secret, getErr := getSecret(ctx, r.Client, r.APIReader, key)
+
+	log.Debugf("Fetched secret %v err %v", secret, getErr)
+
+	if !k8serrors.IsNotFound(getErr) {
```
eranco74, line 296: In case getErr is nil , `k8serrors.IsNotFound(getErr)` will return false...
I'm not sure what you are trying to achieve here

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -241,6 +256,91 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
...
+		log.WithError(getErr).Errorf("failed to get secret %s", key)
+		return key, getErr
+	}
+
+	// Exit early if the secret already exists
+	if k8serrors.IsNotFound(getErr) {
+		return key, nil
+	}
+
+	spokeClient, err := r.SpokeK8sClientFactory.Create(secret)
```
eranco74, line 306: Why do you need the MCS secret to create the spoke cluster client?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -241,6 +256,91 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
...
+
+	log.Debugf("Fetching MCS secret from spoke cluster")
+
+	key := types.NamespacedName{
+		Namespace: clusterDeployment.Namespace,
+		Name:      getMCSSecretName(clusterDeployment),
+	}
+	log.Debugf("Expected key: %s", key)
+
+	secret, getErr := getSecret(ctx, r.Client, r.APIReader, key)
```
eranco74, line 292: Perhaps you meant to get the kubeconfig secret?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -2013,3 +2113,10 @@ func getClusterDeploymentAdminKubeConfigSecretName(cd *hivev1.ClusterDeployment)
 	}
 	return fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name)
 }
+
+func getMCSSecretName(cd *hivev1.ClusterDeployment) string {
```
eranco74, line 2117: OK, now it's clear, this method name is misleading:smiley_cat:

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -241,6 +256,91 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
...
+		if err := r.Update(ctx, clusterInstall); err != nil {
+			return err
+		}
+		log.Debugf("Endpoint updated")
+	}
+	return nil
+}
+
+func (r *ClusterDeploymentsReconciler) createMCSCertSecret(ctx context.Context, log logrus.FieldLogger, clusterDeployment *hivev1.ClusterDeployment, cluster *common.Cluster, clusterInstall *hiveext.AgentClusterInstall) (types.NamespacedName, error) {
+
```
eranco74, line 283: Consider refactoring it to:
GetSpokeClusterClient (pretty sure it's already implemented)
GetMCSCertInfo

internal/controller/controllers/clusterdeployments_controller.go
vrutkovs: First we want to check that MCS cert doesn't exist already. If its not we proceed to fetching kubeconfig

eranco74:
```suggestion
	MCSsecret, getErr := getSecret(ctx, r.Client, r.APIReader, key)
```

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -234,13 +236,133 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
...
 	return r.updateStatus(ctx, log, clusterInstall, cluster, nil)
 }
 
+func (r *ClusterDeploymentsReconciler) setHTTPSIgnitionURL(ctx context.Context, log logrus.FieldLogger, cluster *common.Cluster, clusterInstall *hiveext.AgentClusterInstall, secretRef types.NamespacedName) error {
+
+	if clusterInstall.Spec.IgnitionEndpoint == nil {
+		log.Debugf("Empty endpoint found")
+		// Set custom Ignition endpoint so that day2 workers would join using HTTPS endpoint
+		// ensureMCSCert would add ignition override with MCS secret
+		ignitionEndpoint := fmt.Sprintf("%s/config/%s", common.GetMCSUrlBase(cluster), models.HostRoleWorker)
```
tsorya, line 265: should not pool part be per host? I think that mcs pool can be specified per host so we should actually set endpoint without pool and add pool part afterwards? no?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -234,13 +236,133 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
...
+	certData, ok := configMap.Data[MCS_CERT_NAME]
+	log.Debugf("Fetched certData ok %v", ok)
+	if !ok || len(certData) == 0 {
+		return key, errors.Errorf("Configmap data for %s/%s does not contain %s", configMap.Namespace, configMap.Name, MCS_CERT_NAME)
+	}
+	mcsCrt := configMap.Data[MCS_CERT_NAME]
+	data := map[string][]byte{
+		corev1.TLSCertKey: []byte(mcsCrt),
+	}
+	log.Debugf("Creating MCS secret")
```
tsorya, line 355: info?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -234,13 +236,133 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
...
+	kubeConfigSecret, getErr := getSecret(ctx, r.Client, r.APIReader, kubeConfigSecretKey)
+	log.Debugf("Fetched kubeconfig secret err %v", getErr)
+
+	if getErr != nil {
+		log.WithError(getErr).Errorf("failed to get secret %s", key)
+		return key, getErr
+	}
+
+	spokeClient, err := r.SpokeK8sClientFactory.Create(kubeConfigSecret)
+	log.Debugf("Created spokeClient %v err %v", spokeClient, err)
```
vrutkovs, line 329: not sure why some tests runs get stuck here with
```
time="2022-07-20T14:47:08Z" level=warning msg="failed to get spoke kube client" func="github.com/openshift/assisted-service/internal/controller/controllers.(*spokeK8sClientFactory).Create" file="/assisted-service/internal/controller/controllers/spoke_k8s_client.go:83" error="Get \"https://api.assisted-test-cluster.redhat.com:6443/api?timeout=32s\": dial tcp: lookup api.assisted-test-cluster.redhat.com on 172.30.0.10:53: no such host"
```
error, and some runs wipe away APIVIP/IngressVIP after `setHTTPSIgnitionURL`

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -234,13 +236,133 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
...
+func (r *ClusterDeploymentsReconciler) setHTTPSIgnitionURL(ctx context.Context, log logrus.FieldLogger, cluster *common.Cluster, clusterInstall *hiveext.AgentClusterInstall, secretRef types.NamespacedName) error {
+
+	if clusterInstall.Spec.IgnitionEndpoint == nil {
+		log.Debugf("Empty endpoint found")
+		// Set custom Ignition endpoint so that day2 workers would join using HTTPS endpoint
+		// ensureMCSCert would add ignition override with MCS secret
+		ignitionEndpoint := fmt.Sprintf("%s/config/%s", common.GetMCSUrlBase(cluster), models.HostRoleWorker)
+		log.Infof("setting worker ignition URL to %s", ignitionEndpoint)
+
+		freshClusterInstall := &hiveext.AgentClusterInstall{}
```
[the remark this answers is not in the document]
vrutkovs: We're reconciling `ClusterDeployment`, so we know it won't change, but nothing is preventing `AgentClusterInstal` from changing meanwhile. This is why we have to refetch it so make sure we don't overwrite the status.

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -235,13 +240,181 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 		}
 		return r.installDay1(ctx, log, clusterDeployment, clusterInstall, cluster, pullSecret)
 	} else if swag.StringValue(cluster.Kind) == models.ClusterKindAddHostsCluster {
+		// Set HTTPS Ignition URL
+		mcsSecretKey, err := r.createMCSCertSecret(ctx, log, clusterDeployment, cluster, clusterInstall)
```
eranco74, line 244: Since this controller is the one creating this secret I think it should also add a finalizer that deletes the secret when the cluster deployment get deleted.

vrutkovs:
>time="2022-09-13T23:17:08Z" level=debug msg="Created spokeClient <nil> err Get \"https://192.168.111.26:6443/api?timeout=32s\": x509: certificate is valid for 172.30.0.1, not 192.168.111.26" func="github.com/openshift/assisted-service/internal/controller/controllers.(*ClusterDeploymentsReconciler).createMCSCertSecret" file="/assisted-service/internal/controller/controllers/clusterdeployments_controller.go:390" agent_cluster_install=assisted-agent-cluster-install agent_cluster_install_namespace=assisted-spoke-cluster cluster_deployment=assisted-test-cluster cluster_deployment_namespace=assisted-spoke-cluster go-id=619 request_id=84bedd64-57cb-46c7-bbf7-c4ec39174dbd

We can't use API DNS as SNO clusters are exposing it via internal DNSMasq. We can't use APIVIP either - certificate is issued for DNS name only

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -242,6 +245,107 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 	return r.updateStatus(ctx, log, clusterInstall, cluster, nil)
 }
 
+func (r *ClusterDeploymentsReconciler) setHTTPSIgnitionURL(ctx context.Context, log logrus.FieldLogger, cluster *common.Cluster, clusterInstall *hiveext.AgentClusterInstall, secretRef types.NamespacedName) error {
+
+	if clusterInstall.Spec.IgnitionEndpoint == nil {
+		// Set custom Ignition endpoint so that day2 workers would join using HTTPS endpoint
+		// ensureMCSCert would add ignition override with MCS secret
+		ignitionEndpoint := fmt.Sprintf("%s/config/%s", common.GetMCSUrlBase(cluster), hiveext.WorkerAgentMachinePool)
```
mkowalski, line 253: Is this function used **only** for day2 worker or for any day2 ? If for any, then why does it have hardcoded `hiveext.WorkerAgentMachinePool`? If only for day2 worker, then maybe we can make it clear in the name of the function?

internal/controller/controllers/clusterdeployments_controller_test.go
```diff
@@ -1380,12 +1386,98 @@ var _ = Describe("cluster reconcile", func() {
...
+			Expect(err).To(BeNil())
+			// First reconciliation will update Ignition URL and requeue updates
+			Expect(result).To(Equal(ctrl.Result{Requeue: true}))
+			result, err = cr.Reconcile(ctx, request)
+			Expect(err).To(BeNil())
+			Expect(result).To(Equal(ctrl.Result{Requeue: false}))
+			aci = getTestClusterInstall()
 			Expect(aci.Status.DebugInfo.State).To(Equal(models.ClusterStatusAddingHosts))
+			Expect(aci.Spec.IgnitionEndpoint).NotTo(BeNil())
+			ignitionEndpoint := fmt.Sprintf("https://1.2.3.5:22623/config/%s", hiveext.WorkerAgentMachinePool)
```
mkowalski, line 1479: As I see that, AgentClusterInstall as a whole will get ignition endpoint point to the **worker** ignition. Aren't we breaking day2 master with this approach?

mkowalski (changes requested):

internal/controller/controllers/clusterdeployments_controller.go
tsorya: you are right. I think we talked about it once but missed in review. We should MachineConfigPoolName and not hard code any value

mkowalski: We cannot blindly use MachineConfigPoolName here for two reasons

- `MachineConfigPoolName` is very often empty because it's a field used for override and has no default value; if anything, `GetEffectiveRole()` is a good fit
- this code sets ignition endpoint for the whole ClusterInstall object, whereas MachineConfigPoolName or EffectiveRole are properties of a single Agent; there is no any contract saying that all the Agents for a specific ClusterInstall will contain hosts with only a single role

vrutkovs:
Flakes, both tests failed to setup ztp cluster
/retest

vrutkovs:
/hold cancel

All tests passing, ready for review

internal/bminventory/inventory.go
```diff
@@ -882,6 +886,32 @@ func (b *bareMetalInventory) createAndUploadDay2NodeIgnition(ctx context.Context
 	return nil
 }
 
+func (b *bareMetalInventory) UpdateIgnitionEndpointIfHasMCSCert(log logrus.FieldLogger, host *models.Host, cluster *common.Cluster, ignitionEndpointUrl string) (string, error) {
+	// No ignition override set by bmh agent controller
+	if host.IgnitionConfigOverrides == "" {
+		return ignitionEndpointUrl, nil
+	}
+	// No CA certs set in ignition override
+	if !ignition.HasCACertInIgnition(host.IgnitionConfigOverrides) {
```
carbonin, line 895: Is there any chance that we're providing a CACert for some purpose other than the spoke cluster?
Worried that this could cause failures if someone needs a cert for some other reason.

vrutkovs: It has to be a host-specific CA override (cluster-wide Ignition with proxy CA won't trigger this). We know of only one usecase for this - a custom Ignition service and this PR won't break it

internal/bminventory/inventory.go
```diff
@@ -882,6 +886,32 @@ func (b *bareMetalInventory) createAndUploadDay2NodeIgnition(ctx context.Context
...
+		return ignitionEndpointUrl, nil
+	}
+	log.Infof("Found CA certificate in ignition override for cluster %s, host %s", cluster.ID, host.ID)
+	url, err := url.Parse(ignitionEndpointUrl)
+	if err != nil {
+		return ignitionEndpointUrl, err
+	}
+	url.Scheme = "https"
+	apiIntHostname := fmt.Sprintf("%s.%s.%s", constants.InternalAPIClusterSubdomain, cluster.Name, cluster.BaseDNSDomain)
+	url.Host = fmt.Sprintf("%s:22623", apiIntHostname)
```
carbonin, line 909: Do we have this format and port somewhere else? Seems like it would be a good thing to share.

carbonin: Also I don't see why this needs to be two `Sprintf` calls.

vrutkovs: Moved ports to `constants`, removed intermediate variable

carbonin:
Looks okay to me.

Would want @mkowalski or @tsorya to take a look though since they were involved in earlier reviews.

PR #4151, opened 2022-07-18
MGMT-11174: Add IPXE boot control flag to infra env in kube API

ori-amizur:
This is needed to support boot control in Kube API.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @carbonin 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/v1beta1/infraenv_types.go
+++ b/api/v1beta1/infraenv_types.go
@@ -88,6 +88,16 @@ type InfraEnvSpec struct {
 	// +kubebuilder:default=x86_64
 	// +optional
 	CpuArchitecture string `json:"cpuArchitecture,omitempty"`
+
+	// IPXEScriptType the script type that should be served (HostRedirect/Boot)
+	// When the value is HostRedirect, the served script consists from URL redirection (i.e chain <url>).  The
+	// <url> is the same URL that contains query parameter mac=${net0/mac}.  The value is IPXE macro is replaced
+	//  with the mac address.  The mac address is used to identify the invoking script.
+	// When the value is Boot (or default), the served script contains a script that causes the invoker to boot
+	// the discovery ISO from the network.
+	// +kubebuilder:default=Boot
+	// +optional
+	IPXEScriptType IPXEScriptType `json:"ipxeScriptType"`
 }
 
 // Proxy defines the proxy settings for agents and clusters that use the InfraEnv.
@@ -168,6 +178,18 @@ type InfraEnvList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []InfraEnv `json:"items"`
 }
+
+// IPXEScriptType is the script type that should be served (HostRedirect/Boot)
+// +kubebuilder:validation:Enum="";Boot;HostRedirect
+type IPXEScriptType string
+
+const (
+	// HostRedirect - Boot with mac identification redirect script
+	HostRedirect IPXEScriptType = "HostRedirect"
+
+	// Boot - Boot from network
+	Boot IPXEScriptType = "Boot"
+)
 
 func init() {
 	SchemeBuilder.Register(&InfraEnv{}, &InfraEnvList{})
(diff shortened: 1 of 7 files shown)


carbonin:
Generally I think we should [avoid](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) bools in k8s APIs.

@avishayt @filanov @mhrivnak any ideas for how to phrase this concept as non-boolean?

Maybe `IPXEScriptType` with values `HostRedirect` or `Boot`. Not sure how something like this could be expanded though.

avishayt:
> Generally I think we should [avoid](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) bools in k8s APIs.
> 
> @avishayt @filanov @mhrivnak any ideas for how to phrase this concept as non-boolean?
> 
> Maybe `IPXEScriptType` with values `HostRedirect` or `Boot`. Not sure how something like this could be expanded though.

Definitely agree.  I was thinking the same.  Maybe "IPXEMode" with values "HostRedirect" or "AlwaysBoot"?

ori-amizur:
> > Generally I think we should [avoid](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) bools in k8s APIs.
> > @avishayt @filanov @mhrivnak any ideas for how to phrase this concept as non-boolean?
> > Maybe `IPXEScriptType` with values `HostRedirect` or `Boot`. Not sure how something like this could be expanded though.
> 
> Definitely agree. I was thinking the same. Maybe "IPXEScriptMode" with values "AutoSelect" or "DiskOnly"? In the future, it can be expanded to force booting from ISO, maybe something else.

If this is the case, it has to be reflected in the REST API as well.

api/v1beta1/infraenv_types.go
```diff
@@ -88,6 +88,10 @@ type InfraEnvSpec struct {
 	// +kubebuilder:default=x86_64
 	// +optional
 	CpuArchitecture string `json:"cpuArchitecture,omitempty"`
+
+	// IPXEScriptType the script type that should be served (HostRedirect/Boot)
+	// +optional
```
mhrivnak, line 93: It seems like effectively the default behavior is to act as if "Boot" is the selection. WDYT about making that the explicit default here?

mhrivnak: And perhaps also limit the values to these two using a validation marker. https://book.kubebuilder.io/reference/markers/crd-validation.html

mhrivnak (changes requested):

api/v1beta1/infraenv_types.go
```diff
@@ -88,6 +88,12 @@ type InfraEnvSpec struct {
 	// +kubebuilder:default=x86_64
 	// +optional
 	CpuArchitecture string `json:"cpuArchitecture,omitempty"`
+
+	// IPXEScriptType the script type that should be served (HostRedirect/Boot)
+	// +kubebuilder:default=Boot
+	// +kubebuilder:validation:Pattern=Boot|HostRedirect
```
mhrivnak, line 94: I think the Enum validation is a better option since it enables UIs to present the acceptable values and let the user pick one.

mhrivnak (changes requested):

api/v1beta1/infraenv_types.go
ori-amizur: Done

api/v1beta1/infraenv_types.go
```diff
@@ -88,6 +88,11 @@ type InfraEnvSpec struct {
 	// +kubebuilder:default=x86_64
 	// +optional
 	CpuArchitecture string `json:"cpuArchitecture,omitempty"`
+
+	// IPXEScriptType the script type that should be served (HostRedirect/Boot)
```
mhrivnak, line 92: Sorry, this is my last comment I think. :) Here we should have clear documentation about what each option does. And it should not assume that the user knows about the two different scripts already and what they do.

mhrivnak (changes requested):

avishayt:
I don't love the enum options names. Specifically "boot" is ambiguous because all options boot. Can we go with something clearer?

PR #4158, opened 2022-07-19
MGMT-10816: Returning the actual resulting cluster platform

eliorerz:
- Make `none platform` as a valid platform
- When setting `UserManagedNetworking` (on cluster creation and on cluster update), if `UserManagedNetworking=true && Platform.Type == "baremetal"` than `cluster.Platform.Type` is set to `"none"`
- Allow creating and updating cluster with `Platform.Type = none`  
- `GetSupportedProvidersByHosts` returns also `none` platform alongside `baremetal`

- [x] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @eranco74 
/cc @avishayt 
/cc @gamli75

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -291,7 +291,8 @@ func (b *bareMetalInventory) updatePullSecret(pullSecret string, log logrus.Fiel
 	return pullSecret, nil
 }
 
-func (b *bareMetalInventory) setDefaultRegisterClusterParams(_ context.Context, params installer.V2RegisterClusterParams) installer.V2RegisterClusterParams {
+func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context, params installer.V2RegisterClusterParams) (installer.V2RegisterClusterParams, error) {
+	log := logutil.FromContext(ctx, b.log)
 	if params.NewClusterParams.ClusterNetworks == nil {
 		params.NewClusterParams.ClusterNetworks = []*models.ClusterNetwork{
 			{Cidr: models.Subnet(b.Config.DefaultClusterNetworkCidr), HostPrefix: b.Config.DefaultClusterNetworkHostPrefix},
@@ -309,20 +310,27 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(_ context.Context, 
 	if params.NewClusterParams.VipDhcpAllocation == nil {
 		params.NewClusterParams.VipDhcpAllocation = swag.Bool(false)
 	}
-	if params.NewClusterParams.UserManagedNetworking == nil {
-		params.NewClusterParams.UserManagedNetworking = swag.Bool(false)
-	}
 	if params.NewClusterParams.Hyperthreading == nil {
 		params.NewClusterParams.Hyperthreading = swag.String(models.ClusterHyperthreadingAll)
 	}
 	if params.NewClusterParams.SchedulableMasters == nil {
 		params.NewClusterParams.SchedulableMasters = swag.Bool(false)
 	}
-	if params.NewClusterParams.Platform == nil {
-		params.NewClusterParams.Platform = &models.Platform{
-			Type: common.PlatformTypePtr(models.PlatformTypeBaremetal),
-		}
+	if params.NewClusterParams.HighAvailabilityMode == nil {
+		params.NewClusterParams.HighAvailabilityMode = swag.String(models.ClusterHighAvailabilityModeFull)
 	}
+
+	log.Infof("Verifying cluster platform and user-managed-networking, got platform=%v and userManagedNetworking=%v", params.NewClusterParams.Platform, params.NewClusterParams.UserManagedNetworking)
+	platform, userManagedNetworking, err := getActualCreateClusterPlatformParams(params.NewClusterParams.Platform, params.NewClusterParams.UserManagedNetworking, params.NewClusterParams.HighAvailabilityMode)
+	if err != nil {
+		log.Error(err)
+		return params, err
+	}
+
+	params.NewClusterParams.Platform = platform
+	params.NewClusterParams.UserManagedNetworking = userManagedNetworking
+	log.Infof("Cluster high-availability-mode is set to %v, setting platform type to %v and user-managed-networking to %v", params.NewClusterParams.HighAvailabilityMode, platform, userManagedNetworking)
+
 	if params.NewClusterParams.AdditionalNtpSource == nil {
 		params.NewClusterParams.AdditionalNtpSource = &b.Config.DefaultNTPSource
 	}
@@ -333,7 +341,7 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(_ context.Context, 
 		}
 	}
 
-	return params
+	return params, nil
 }
 
 func (b *bareMetalInventory) validateRegisterClusterInternalParams(params *installer.V2RegisterClusterParams, log logrus.FieldLogger) error {
@@ -421,11 +429,15 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 	if err = validations.ValidateDualStackNetworks(params.NewClusterParams, false); err != nil {
 		return nil, common.NewApiError(http.StatusBadRequest, err)
 	}
-	if err = b.validateRegisterClusterInternalParams(&params, log); err != nil {
+
+	params, err = b.setDefaultRegisterClusterParams(ctx, params)
+	if err != nil {
 		return nil, err
 	}
 
-	params = b.setDefaultRegisterClusterParams(ctx, params)
+	if err = b.validateRegisterClusterInternalParams(&params, log); err != nil {
+		return nil, err
+	}
 
 	cpuArchitecture, err := b.getNewClusterCPUArchitecture(params.NewClusterParams)
 	if err != nil {
@@ -1668,6 +1680,93 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
 	return b.v2UpdateClusterInternal(ctx, params, NonInteractive)
 }
 
+func checkPlatformWrongParamsInput(platform *models.Platform, userManagedNetworking *bool) error {
+	if platform != nil && userManagedNetworking != nil {
+		userManagedNetworkingStatus := "enabled"
+		if !*userManagedNetworking {
+			userManagedNetworkingStatus = "disabled"
+		}
+
+		if (!*userManagedNetworking && *platform.Type == models.PlatformTypeNone) || (*userManagedNetworking && models.PlatformTypeBaremetal == *platform.Type) {
+			return common.NewApiError(http.StatusBadRequest, errors.Errorf("Can't set %s platform with user-managed-networking %s", *platform.Type, userManagedNetworkingStatus))
+		}
+	}
+
+	return nil
+}
+
+func getActualUpdateClusterPlatformParams(platform *models.Platform, userManagedNetworking *bool, cluster *common.Cluster) (*models.Platform, *bool, error) {
+	if platform == nil && userManagedNetworking == nil {
+		return nil, nil, nil
+	}
+
+	if err := checkPlatformWrongParamsInput(platform, userManagedNetworking); err != nil {
+		return nil, nil, err
+	}
+
+	if *cluster.Platform.Type == models.PlatformTypeBaremetal {
+		if !swag.BoolValue(userManagedNetworking) && (platform == nil || *platform.Type == models.PlatformTypeBaremetal) {
+			// Platform is already baremetal, nothing to do
+			return nil, nil, nil
+		}
+
+		if (platform != nil && *platform.Type == models.PlatformTypeNone) || (swag.BoolValue(userManagedNetworking) && platform == nil) {
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
+		}
+	} else if *cluster.Platform.Type == models.PlatformTypeNone {
+		if (userManagedNetworking == nil || *userManagedNetworking) && (platform == nil || *platform.Type == models.PlatformTypeNone) {
+			// Platform is already none, nothing to do
+			return nil, nil, nil
+		}
+
+		if *cluster.HighAvailabilityMode == models.ClusterHighAvailabilityModeNone {
+			if !swag.BoolValue(userManagedNetworking) || (platform != nil && *platform.Type != models.PlatformTypeNone) {
+				return nil, nil, common.NewApiError(http.StatusBadRequest, errors.New("disabling User Managed Networking or setting platform different than none platform is not allowed in single node Openshift"))
+			}
+		}
+
+		if !swag.BoolValue(userManagedNetworking) {
+			if platform == nil || *platform.Type == models.PlatformTypeBaremetal {
+				if cluster.CPUArchitecture != common.X86CPUArchitecture && !featuresupport.IsFeatureSupported(cluster.OpenshiftVersion, models.FeatureSupportLevelFeaturesItems0FeatureIDARM64ARCHITECTUREWITHCLUSTERMANAGEDNETWORKING) {
+					return nil, nil, common.NewApiError(http.StatusBadRequest, errors.New("disabling User Managed Networking or setting Bare-Metal platform is not allowed for clusters with non-x86_64 CPU architecture"))
+				}
+
+				return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeBaremetal)}, swag.Bool(false), nil
+			}
+		}
+	}
+
+	return platform, userManagedNetworking, nil
+}
+
+func getActualCreateClusterPlatformParams(platform *models.Platform, userManagedNetworking *bool, highAvailabilityMode *string) (*models.Platform, *bool, error) {
+	if err := checkPlatformWrongParamsInput(platform, userManagedNetworking); err != nil {
+		return nil, nil, err
+	}
+
+	if *highAvailabilityMode == models.ClusterHighAvailabilityModeFull {
+		if (platform == nil || *platform.Type == models.PlatformTypeBaremetal) && !swag.BoolValue(userManagedNetworking) {
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeBaremetal)}, swag.Bool(false), nil
+		}
+
+		if swag.BoolValue(userManagedNetworking) || (platform != nil && *platform.Type == models.PlatformTypeNone) {
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
+		}
+	} else { // *highAvailabilityMode == models.ClusterHighAvailabilityModeNone
+		if platform != nil && *platform.Type == models.PlatformTypeBaremetal {
+			return nil, nil, common.NewApiError(http.StatusBadRequest, errors.Errorf("Can't set %s platform on single node OpenShift", *platform.Type))
+		}
+
+		if userManagedNetworking != nil && !*userManagedNetworking {
+			return nil, nil, common.NewApiError(http.StatusBadRequest, errors.New("Can't disable user-managed-networking on single node OpenShift"))
+		}
+
+		return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
+	}
+
+	return platform, userManagedNetworking, nil
+}
+
 func (b *bareMetalInventory) v2UpdateClusterInternal(ctx context.Context, params installer.V2UpdateClusterParams, interactivity Interactivity) (*common.Cluster, error) {
 	log := logutil.FromContext(ctx, b.log)
 	var cluster *common.Cluster
@@ -1719,6 +1818,18 @@ func (b *bareMetalInventory) v2UpdateClusterInternal(ctx context.Context, params
 		log.WithError(err).Errorf("cluster %s can't be updated in current state", params.ClusterID)
 		return nil, common.NewApiError(http.StatusConflict, err)
 	}
+
+	log.Infof("Current cluster platform is set to %v and user-managed-networking is set to %v", cluster.Platform, cluster.UserManagedNetworking)
+	log.Infof("Verifying cluster platform and user-managed-networking, got platform=%v and userManagedNetworking=%v", params.ClusterUpdateParams.Platform, params.ClusterUpdateParams.UserManagedNetworking)
+	platform, userManagedNetworking, err := getActualUpdateClusterPlatformParams(params.ClusterUpdateParams.Platform, params.ClusterUpdateParams.UserManagedNetworking, cluster)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
+	params.ClusterUpdateParams.Platform = platform
+	params.ClusterUpdateParams.UserManagedNetworking = userManagedNetworking
+	log.Infof("Platform verification completed, setting platform type to %v and user-managed-networking to %v", platform, userManagedNetworking)
 
 	if err = b.v2NoneHaModeClusterUpdateValidations(cluster, params); err != nil {
 		log.WithError(err).Warnf("Unsupported update params in none ha mode")
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -210,6 +210,7 @@ func getDefaultClusterCreateParams() *models.ClusterCreateParams {
 		Platform: &models.Platform{
 			Type: common.PlatformTypePtr(models.PlatformTypeBaremetal),
 		},
+		HighAvailabilityMode: swag.String(models.ClusterHighAvailabilityModeFull),
 	}
 }
 
@@ -1858,8 +1859,12 @@ var _ = Describe("cluster", func() {
 		})
 	})
 	Context("Create non HA cluster", func() {
+		var clusterParams *models.ClusterCreateParams
+
 		BeforeEach(func() {
 			mockDurationsSuccess()
+			clusterParams = getDefaultClusterCreateParams()
+			clusterParams.Platform = &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}
 		})
 		It("happy flow", func() {
 			bm.clusterApi = cluster.NewManager(cluster.Config{}, common.GetTestLog().WithField("pkg", "cluster-monitor"),
@@ -1868,7 +1873,6 @@ var _ = Describe("cluster", func() {
 			mockClusterRegisterSuccess(true)
 			noneHaMode := models.ClusterHighAvailabilityModeNone
 			MinimalOpenShiftVersionForNoneHA := "4.8.0-fc.0"
-			clusterParams := getDefaultClusterCreateParams()
 			clusterParams.OpenshiftVersion = swag.String(MinimalOpenShiftVersionForNoneHA)
 			clusterParams.HighAvailabilityMode = &noneHaMode
 			reply := bm.V2RegisterCluster(ctx, installer.V2RegisterClusterParams{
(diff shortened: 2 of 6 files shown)


internal/bminventory/inventory.go
```diff
@@ -1653,6 +1660,23 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
 	return b.v2UpdateClusterInternal(ctx, params, NonInteractive)
 }
 
+func updateClusterNonePlatformParams(params installer.V2UpdateClusterParams, cluster *common.Cluster) {
+	if params.ClusterUpdateParams.UserManagedNetworking == nil || cluster.Platform == nil || cluster.UserManagedNetworking == nil {
```
eranco74, line 1664: This should be && no?

internal/bminventory/inventory.go
```diff
@@ -1653,6 +1660,23 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+		return
+	}
+
+	if *params.ClusterUpdateParams.UserManagedNetworking && !*cluster.UserManagedNetworking {
+		if *cluster.Platform.Type == models.PlatformTypeBaremetal {
+			params.ClusterUpdateParams.Platform = &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}
+
+		} else if params.ClusterUpdateParams.Platform != nil && models.PlatformTypeBaremetal == *params.ClusterUpdateParams.Platform.Type {
+			params.ClusterUpdateParams.Platform = &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}
```
eranco74, line 1673: You are basically overriding a param provided by the users because it's conflicting with another param,
I think this deserves returning bad request, @avishayt?

internal/bminventory/inventory.go
```diff
@@ -1653,6 +1660,23 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+		}
+	} else if params.ClusterUpdateParams.Platform != nil && *params.ClusterUpdateParams.Platform.Type == models.PlatformTypeNone {
+		params.ClusterUpdateParams.UserManagedNetworking = swag.Bool(true)
+	}
```
eranco74, line 1677: What about the case where user sends `UserManagedNetworking = false`, and `cluster.UserManagedNetworking = true`?
it seems that that case is ignored.
Also, if we are not to returning badRequest we should make it clear what param takes presedence Platform or UMN flag.

internal/bminventory/inventory.go
```diff
@@ -1689,6 +1713,8 @@ func (b *bareMetalInventory) v2UpdateClusterInternal(ctx context.Context, params
 		return nil, common.NewApiError(http.StatusNotFound, err)
 	}
 
+	updateClusterNonePlatformParams(params, cluster)
```
eranco74, line 1716: Consider renaming e.g. `alignClusterPlatformParams`

internal/bminventory/inventory.go
avishayt: I agree.  If they selected user-managed networking and baremetal that's 400.

internal/bminventory/inventory.go
eliorerz: No, If one of those properties is not set it will not be able to update the platform/UMN

internal/bminventory/inventory.go
eliorerz: > What about the case where user sends UserManagedNetworking = false, and cluster.UserManagedNetworking = true?

You right, but I think we can handle this case instead `*params.ClusterUpdateParams.UserManagedNetworking == false && *cluster.Platform.Type == models.PlatformTypeNone`
WDYT?

eranco74: Unsure how we can handle it though.

internal/bminventory/inventory.go
```diff
@@ -1653,6 +1660,23 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
 	return b.v2UpdateClusterInternal(ctx, params, NonInteractive)
 }
 
+func updateClusterNonePlatformParams(params installer.V2UpdateClusterParams, cluster *common.Cluster) {
+	if params.ClusterUpdateParams.UserManagedNetworking == nil || cluster.Platform == nil || cluster.UserManagedNetworking == nil {
+		return
+	}
+
+	if *params.ClusterUpdateParams.UserManagedNetworking && !*cluster.UserManagedNetworking {
```
eranco74, line 1668: if cluster.UserManagedNetworking is nil you can hit nil pointer dereference

internal/bminventory/inventory.go
eranco74: Hmmm, unsure I understand the logic here.
Don't we have a default for `cluster.Platform`? under what circumstances it will be nil?
Same question for `cluster.UserManagedNetworking`.

If `params.ClusterUpdateParams.UserManagedNetworking` is nil but `params.ClusterUpdateParams.Platform.Type == models.PlatformTypeNone` you would want to update `params.ClusterUpdateParams.UserManagedNetworking` to `swag.Bool(true)` no?

internal/bminventory/inventory.go
eliorerz: I think that in that case we should return to baremetal. The "duplication" of UMN and platform=none casing some edge cases that not that trivial to handle (e.g. selected vsphere platfrom with UMN=true and change back to baremetal, is it disabling the UMN?)

eranco74: These 2 parameters are optional, if the user chooses to provide both with conflicting values we shouldn't overthink it and simply return bad request.
In case one was provided we should align the other param as well (within the boundary of None / Baremetal platform and not overriding any other platforms provided by the user).

internal/bminventory/inventory.go
eliorerz: i updated it and now returning bed-request when `UserManagedNetworking=false and Platform=none or UserManagedNetworking=true and Platform=baremetal`

eliorerz:
> Hmmm, unsure I understand the logic here.
Don't we have a default for cluster.Platform? under what circumstances it will be nil?
Same question for cluster.UserManagedNetworking.

@eranco74 You are right, I removed all the unnecessary `nil` "validations"

internal/bminventory/inventory.go
eliorerz: @eranco74 You are right, I removed all the unnecessary nil "validations"

internal/bminventory/inventory.go
```diff
@@ -458,6 +458,13 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 		monitoredOperators = append(monitoredOperators, newOLMOperators...)
 	}
 
+	platform := params.NewClusterParams.Platform
+	if models.PlatformTypeBaremetal == *platform.Type && *params.NewClusterParams.UserManagedNetworking {
```
eranco74, line 462: params.NewClusterParams.UserManagedNetworking might be nil!

internal/bminventory/inventory.go
```diff
@@ -458,6 +458,13 @@ func (b *bareMetalInventory) RegisterClusterInternal(
...
+		*platform.Type = models.PlatformTypeNone
```
eranco74, line 463: Didn't we agree that this case should return bad request?

internal/bminventory/inventory.go
```diff
@@ -458,6 +458,13 @@ func (b *bareMetalInventory) RegisterClusterInternal(
...
+	} else if models.PlatformTypeNone == *platform.Type {
+		params.NewClusterParams.UserManagedNetworking = swag.Bool(true)
```
eranco74, line 465: I think you want to check that that params.NewClusterParams.UserManagedNetworking is nil before overriding it

internal/bminventory/inventory.go
```diff
@@ -458,6 +458,13 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 		monitoredOperators = append(monitoredOperators, newOLMOperators...)
 	}
 
+	platform := params.NewClusterParams.Platform
+	if models.PlatformTypeBaremetal == *platform.Type && *params.NewClusterParams.UserManagedNetworking {
+		*platform.Type = models.PlatformTypeNone
+	} else if models.PlatformTypeNone == *platform.Type {
```
eranco74, line 464: Consider moving all this to a function than handles the platform and UMN params.
e.g. `alignClusterPlatformParams`

internal/bminventory/inventory.go
eranco74, line 462: Note that in line 421 `setDefaultRegisterClusterParams` will set the platform to Baremetal if the user didn't specify it.
Meaning whatever checks you are doing for the params input it should happen before we set defaults.

internal/bminventory/inventory.go
```diff
@@ -1660,6 +1667,28 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+		if !*paramsUserManagedNetworking && platformType == models.PlatformTypeNone || *paramsUserManagedNetworking && models.PlatformTypeBaremetal == platformType {
+			return common.NewApiError(http.StatusBadRequest, errors.Errorf("Can't set %s platform with user-managed-networking %v", platformType, *paramsUserManagedNetworking))
+		}
+	}
+
+	if swag.BoolValue(paramsUserManagedNetworking) && *cluster.Platform.Type == models.PlatformTypeBaremetal {
+		params.ClusterUpdateParams.Platform = &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}
+	} else if params.ClusterUpdateParams.Platform != nil && *params.ClusterUpdateParams.Platform.Type == models.PlatformTypeNone {
+		params.ClusterUpdateParams.UserManagedNetworking = swag.Bool(true)
+	} else if paramsUserManagedNetworking != nil && !*paramsUserManagedNetworking && *cluster.Platform.Type == models.PlatformTypeNone {
```
eranco74, line 1686: This case (UMN=false, platform=None) should return bad request

internal/bminventory/inventory.go
```diff
@@ -1660,6 +1667,28 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+	if paramsUserManagedNetworking != nil && params.ClusterUpdateParams.Platform != nil {
+		platformType := *params.ClusterUpdateParams.Platform.Type
+		if !*paramsUserManagedNetworking && platformType == models.PlatformTypeNone || *paramsUserManagedNetworking && models.PlatformTypeBaremetal == platformType {
+			return common.NewApiError(http.StatusBadRequest, errors.Errorf("Can't set %s platform with user-managed-networking %v", platformType, *paramsUserManagedNetworking))
+		}
+	}
+
+	if swag.BoolValue(paramsUserManagedNetworking) && *cluster.Platform.Type == models.PlatformTypeBaremetal {
+		params.ClusterUpdateParams.Platform = &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}
+	} else if params.ClusterUpdateParams.Platform != nil && *params.ClusterUpdateParams.Platform.Type == models.PlatformTypeNone {
```
eranco74, line 1684: So if platform == none it will just override params.ClusterUpdateParams.UserManagedNetworking to true?
why?

eliorerz: It will set the platform_type to none and will update UserManagedNetworking=true.
The update will be later when UserManagedNetworking and platform is updated

internal/bminventory/inventory.go
eliorerz: Didn't we said we want to keep the current behavior?
If the user set the UMN=true that resulting platform=none, now updating UMN=false will return bad request.
In this case we will need to also change the UI behavior?

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,107 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+		if swag.BoolValue(userManagedNetworking) || (platform != nil && *platform.Type == models.PlatformTypeNone) {
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
+		}
+	}
+
+	if *highAvailabilityMode == models.ClusterHighAvailabilityModeNone {
+		if platform != nil && *platform.Type == models.PlatformTypeBaremetal && userManagedNetworking != nil && !*userManagedNetworking {
+			return nil, nil, common.NewApiError(http.StatusBadRequest,
+				errors.Errorf("Can't set %s platform and disable user-managed-networking on single node OpenShift", models.PlatformTypeBaremetal))
+		}
```
eranco74, line 1762: No need for this statement (and since it's complicated as it is I think we can remove it), the user will get  badRequest from the next 2 if statements until if he try to set platform=Baremetal or userManagedNetworking=True.

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,107 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+			return nil, nil, common.NewApiError(http.StatusBadRequest, errors.Errorf("Can't set %s platform on single node OpenShift", *platform.Type))
+		}
+
+		if userManagedNetworking != nil && !*userManagedNetworking {
+			return nil, nil, common.NewApiError(http.StatusBadRequest, errors.New("Can't disable user-managed-networking on single node OpenShift"))
+		}
+
+		if (userManagedNetworking == nil || *userManagedNetworking) && (platform == nil || *platform.Type == models.PlatformTypeNone) {
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
+		}
```
eranco74, line 1774: Don't even need to check it, just return:
```
return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
```

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,107 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+		// case 4
+		if (userManagedNetworking == nil || *userManagedNetworking) && (platform == nil || *platform.Type == models.PlatformTypeNone) {
+			return nil, nil, nil
+		}
+
+		// case 6
+		if !swag.BoolValue(userManagedNetworking) && (platform == nil || *platform.Type == models.PlatformTypeBaremetal) {
+			if cluster.CPUArchitecture != common.X86CPUArchitecture && !featuresupport.IsFeatureSupported(cluster.OpenshiftVersion, models.FeatureSupportLevelFeaturesItems0FeatureIDARM64ARCHITECTUREWITHCLUSTERMANAGEDNETWORKING) {
+				return nil, nil, common.NewApiError(http.StatusBadRequest, errors.New("disabling User Managed Networking or setting Bare-Metal platform is not allowed for clusters with non-x86_64 CPU architecture"))
+			}
```
eranco74, line 1729: Consider moving this to another PR just in case it creates unexpected noise and might get reverted

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,107 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+		if (!*userManagedNetworking && *platform.Type == models.PlatformTypeNone) || (*userManagedNetworking && models.PlatformTypeBaremetal == *platform.Type) {
+			return common.NewApiError(http.StatusBadRequest, errors.Errorf("Can't set %s platform with user-managed-networking %s", *platform.Type, userManagedNetworkingStatus))
+		}
+	}
+
+	return nil
+}
+
+func getActualUpdateClusterPlatformParams(platform *models.Platform, userManagedNetworking *bool, cluster *common.Cluster) (*models.Platform, *bool, error) {
+	// case 5
```
eranco74, line 1695: Before every return log what we end up with

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,107 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+		}
+
+		// case 6
+		if !swag.BoolValue(userManagedNetworking) && (platform == nil || *platform.Type == models.PlatformTypeBaremetal) {
+			if cluster.CPUArchitecture != common.X86CPUArchitecture && !featuresupport.IsFeatureSupported(cluster.OpenshiftVersion, models.FeatureSupportLevelFeaturesItems0FeatureIDARM64ARCHITECTUREWITHCLUSTERMANAGEDNETWORKING) {
+				return nil, nil, common.NewApiError(http.StatusBadRequest, errors.New("disabling User Managed Networking or setting Bare-Metal platform is not allowed for clusters with non-x86_64 CPU architecture"))
+			}
+
+			if *cluster.HighAvailabilityMode == models.ClusterHighAvailabilityModeNone {
+				return nil, nil, common.NewApiError(http.StatusBadRequest, errors.New("disabling User Managed Networking or setting Bare-Metal platform is not allowed in single node mode"))
```
eranco74, line 1732:
```suggestion
				return nil, nil, common.NewApiError(http.StatusBadRequest, errors.New("disabling User Managed Networking or setting Bare-Metal platform is not allowed in single node Openshift"))
```

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,107 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+			}
+
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeBaremetal)}, swag.Bool(false), nil
+		}
+	}
+
+	return platform, userManagedNetworking, nil
+}
+
+func getActualCreateClusterPlatformParams(platform *models.Platform, userManagedNetworking *bool, highAvailabilityMode *string) (*models.Platform, *bool, error) {
```
eranco74, line 1742: Before every return log what we end up with

eranco74 (commented):
looks good

internal/bminventory/inventory.go
avishayt: The message here is clearer, but not sure it's worth the extra code.  I agree, can remove.

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,107 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+		if (platform == nil || *platform.Type == models.PlatformTypeBaremetal) && !swag.BoolValue(userManagedNetworking) {
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeBaremetal)}, swag.Bool(false), nil
+		}
+
+		if swag.BoolValue(userManagedNetworking) || (platform != nil && *platform.Type == models.PlatformTypeNone) {
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
+		}
+	}
+
+	if *highAvailabilityMode == models.ClusterHighAvailabilityModeNone {
```
avishayt, line 1758: Why two separate "ifs" instead of an else?

internal/bminventory/inventory.go
avishayt: And please remove all the "// case N" comments

eliorerz: > And please remove all the "// case N" comments

Sure, that was here just for the first review. I'll remove it.

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,107 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+	return platform, userManagedNetworking, nil
+}
+
+func getActualCreateClusterPlatformParams(platform *models.Platform, userManagedNetworking *bool, highAvailabilityMode *string) (*models.Platform, *bool, error) {
+
+	if err := checkPlatformWrongParamsInput(platform, userManagedNetworking); err != nil {
+		return nil, nil, err
+	}
+
+	if *highAvailabilityMode == models.ClusterHighAvailabilityModeFull {
```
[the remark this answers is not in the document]
eliorerz: It is covered [here](https://github.com/openshift/assisted-service/blob/572b40607fc29728b43a9b418aa2d68d16988760/internal/bminventory/inventory.go#L1749) as part of `!swag.BoolValue(userManagedNetworking)` (it's check if user managed networking is nil or false)

```go
func BoolValue(v *bool) bool {
	if v != nil {
		return *v
	}
	return false
}
```

internal/bminventory/inventory.go
eliorerz: I think it's more clearer to read it this way.
We already have a lot of conditions there and to my opinion it's more clearer to "see" the condition and not need to scroll up to see the if statement.

internal/bminventory/inventory.go
eliorerz: I think that on platforms different than `baremetal` and `none` with `HighAvailabilityMode=None` we still want to return `params.NewClusterParams.Platform` and `params.NewClusterParams.UserManagedNetworking` as they are

eliorerz: I added a test to check it for vsphere

internal/bminventory/inventory.go
eliorerz: Without it `[It] Fail with non-x86_64 CPU architecture with 4.10` will fail

internal/bminventory/inventory.go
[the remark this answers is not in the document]
avishayt: Right, thanks

internal/bminventory/inventory.go
avishayt: I disagree because this way it's disconnected from the previous "if" even though they are mutually exclusive.  Now I need to scroll up to understand that the code can't actually reach both "ifs".

internal/bminventory/inventory.go
avishayt: SNO is always none platform

internal/bminventory/inventory.go
eliorerz: @avishayt I think that `else if` will solve that issue and still will be more clear, WDYT?

internal/bminventory/inventory.go
eliorerz: Currently, but what about other platforms? For example I don't think that there is any limitation for SNO + Nutanix

avishayt: @eranco74 is that possible?

internal/bminventory/inventory.go
avishayt: OK

internal/bminventory/inventory.go
eranco74: You can install SNO on Nutanix but the platform will be None!

eranco74: In summary, in case the cluster HighAvailabilityMode is None, any changes to the UMN or platform that result in a platform other than None should return a bad request.

eliorerz: Thanks. I'll change it

eliorerz: @avishayt @eranco74 I moved the if block that checks if the cluster `HighAvailabilityMode==Full` and `Platform!=None` outside the `platform == nil || *platform.Type == models.PlatformTypeBaremetal` block.
Now the service will always return `BedRequest` when trying to set platform != none on SNO

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,113 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+		if (platform == nil || *platform.Type == models.PlatformTypeBaremetal) && !swag.BoolValue(userManagedNetworking) {
+			log.Info("Cluster high-availability-mode is set to Full, setting platform type to baremetal and disable user-managed-networking")
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeBaremetal)}, swag.Bool(false), nil
+		}
+
+		if swag.BoolValue(userManagedNetworking) || (platform != nil && *platform.Type == models.PlatformTypeNone) {
+			log.Info("Cluster high-availability-mode is set to Full, setting platform type to none and enable user-managed-networking")
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
+		}
+	} else if *highAvailabilityMode == models.ClusterHighAvailabilityModeNone {
```
avishayt, line 1765:
```suggestion
	} else {  // if *highAvailabilityMode == models.ClusterHighAvailabilityModeNone
```

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,113 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+			}
+		}
+	}
+
+	log.Infof("No platform manipulation is needed. Setting platform=%v and userManagedNetworking=%v", platform, userManagedNetworking)
+	return platform, userManagedNetworking, nil
+}
+
+func getActualCreateClusterPlatformParams(log logrus.FieldLogger, platform *models.Platform, userManagedNetworking *bool, highAvailabilityMode *string) (*models.Platform, *bool, error) {
+
```
avishayt, line 1749:
```suggestion
```

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,113 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+			err := errors.New("Can't disable user-managed-networking on single node OpenShift")
+			log.Error(err)
+			return nil, nil, common.NewApiError(http.StatusBadRequest, err)
+		}
+
+		log.Info("Cluster high-availability-mode is set to None, setting platform type to none and enable user-managed-networking")
+		return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
+	}
+
+	log.Infof("No platform manipulation is needed. Setting platform=%v and userManagedNetworking=%v", platform, userManagedNetworking)
```
avishayt, line 1782: This isn't really accurate because there are several cases above where the function returned the same thing that was passed.  I recommend logging the input to the function and the output.

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,113 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+					return nil, nil, common.NewApiError(http.StatusBadRequest, err)
+				}
+
+				log.Infof("Updating %s platform to baremetal platform", *cluster.Platform.Type)
+				return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeBaremetal)}, swag.Bool(false), nil
+			}
+		}
+	}
+
+	log.Infof("No platform manipulation is needed. Setting platform=%v and userManagedNetworking=%v", platform, userManagedNetworking)
```
avishayt, line 1744: Same comment above.  Just log the inputs and outputs for all cases.

internal/bminventory/inventory.go
```diff
@@ -1668,6 +1676,113 @@ func (b *bareMetalInventory) UpdateClusterNonInteractive(ctx context.Context, pa
...
+			return nil, nil, nil
+		}
+
+		if (platform != nil && *platform.Type == models.PlatformTypeNone) || (swag.BoolValue(userManagedNetworking) && platform == nil) {
+			log.Infof("Updating %s platform to none platform", *cluster.Platform.Type)
+			return &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeNone)}, swag.Bool(true), nil
+		}
+	}
+
+	if *cluster.Platform.Type == models.PlatformTypeNone {
```
avishayt, line 1716:
```suggestion
	else if *cluster.Platform.Type == models.PlatformTypeNone {
```

avishayt, line 1716:
```suggestion
	} else if *cluster.Platform.Type == models.PlatformTypeNone {
```

avishayt (approved):

PR #4160, opened 2022-07-20
MGMT-11174: Add Boot Control Kube API documentation.

ori-amizur:
1.   Add Boot Control Kube API documentation
2.   Refactor boot control parameter to be ipxe-script-type both in the REST API and Kube API.
      This parameter is enum with 2 possible values:
    - DiscoveryImageAlways: Boot unconditionaly from the network discovery image
    - BootOrderControl: Boot from discovery ISO depending on the state of
      the host.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @carbonin 
/cc @mhrivnak 

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/v1beta1/infraenv_types.go
+++ b/api/v1beta1/infraenv_types.go
@@ -89,13 +89,13 @@ type InfraEnvSpec struct {
 	// +optional
 	CpuArchitecture string `json:"cpuArchitecture,omitempty"`
 
-	// IPXEScriptType the script type that should be served (HostRedirect/Boot)
-	// When the value is HostRedirect, the served script consists from URL redirection (i.e chain <url>).  The
-	// <url> is the same URL that contains query parameter mac=${net0/mac}.  The value is IPXE macro is replaced
-	//  with the mac address.  The mac address is used to identify the invoking script.
-	// When the value is Boot (or default), the served script contains a script that causes the invoker to boot
-	// the discovery ISO from the network.
-	// +kubebuilder:default=Boot
+	// IPXEScriptType the script type that should be served (DiscoveryImageAlways/BootOrderControl)
+	// DiscoveryImageAlways: Boot unconditionaly from the network discovery image
+	// BootOrderControl: Boot from discovery ISO depending on the host's state.
+	// When the value is BootOrderControl, the service will look for an Agent record that matches the host's MAC address;
+	// if found, and if that Agent is in a state where it is provisioned and attached to a cluster, then host will boot the host disk.
+	// Otherwise it will boot the discovery ISO using the same script as the DiscoveryImageAlways option.
+	// +kubebuilder:default=DiscoveryImageAlways
 	// +optional
 	IPXEScriptType IPXEScriptType `json:"ipxeScriptType"`
 }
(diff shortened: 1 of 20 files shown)


api/v1beta1/infraenv_types.go
```diff
@@ -89,13 +89,13 @@ type InfraEnvSpec struct {
 	// +optional
 	CpuArchitecture string `json:"cpuArchitecture,omitempty"`
 
-	// IPXEScriptType the script type that should be served (HostRedirect/Boot)
-	// When the value is HostRedirect, the served script consists from URL redirection (i.e chain <url>).  The
+	// IPXEScriptType the script type that should be served (BootOrderControl/DiscoveryImageAlways)
+	// When the value is BootOrderControl, the served script consists from URL redirection (i.e chain <url>).  The
```
mhrivnak, line 93: I like the description from the git commit message better. "Boot from discovery ISO depending on the state of
the host." The stuff about the mac address and redirection could be confusing I think and is just an implementation detail. I suggest listing `DiscoveryImageAlways` first, since it's the simple and default case. Then something like this:

> When the value is BootOrderControl, the served ipxe script will look for an Agent record that matches the host's MAC address; if found, and if that Agent is in a state where it is provisioned and attached to a cluster, then the IPXE script will boot the host from disk. Otherwise it will boot the discovery ISO using the same script as the `DiscoveryImageAlways` option.

ori-amizur: @mhrivnak So do you want to take the text from the commit message, or what you wrote above ?

api/v1beta1/infraenv_types.go
```diff
@@ -89,13 +89,13 @@ type InfraEnvSpec struct {
...
-	// <url> is the same URL that contains query parameter mac=${net0/mac}.  The value is IPXE macro is replaced
-	//  with the mac address.  The mac address is used to identify the invoking script.
-	// When the value is Boot (or default), the served script contains a script that causes the invoker to boot
-	// the discovery ISO from the network.
-	// +kubebuilder:default=Boot
+	// IPXEScriptType the script type that should be served (DiscoveryImageAlways/BootOrderControl)
+	// DiscoveryImageAlways: Boot unconditionaly from the network discovery image
+	// BootOrderControl: Boot from discovery ISO depending on the host's state.
+	// When the value is BootOrderControl, the service will look for an Agent record that matches the host's MAC address;
+	// if found, and if that Host is in a state where disk image was already written to disk, then host will boot the host disk.
```
mhrivnak, line 96: > where disk image was already written to disk,
... and the Agent is still bound to a cluster,
?

ori-amizur: The condition is according to the installing stage the host is in - not if it is provisioned.  Only after disk image has been written to disk, the boot from discovery ISO is skipped.

mhrivnak: Ok. Maybe the part about being bound to a cluster will be added after we fully support this: https://github.com/openshift/assisted-service/pull/3849

mhrivnak (approved):

PR #4170, opened 2022-07-21
MGMT-11131: Cache results of inventory unmarshalling in connectivity test

mkowalski:
This PR changes the key for the cache used in the
`GetHostValidInterfaces` function. Instead of caching based on host's ID
and infraenv's ID, we will be using marshalled inventory string as a key
and unmarshalled as a value. This allows us to prevent the cache being
stale when the host changes its network configuration, as the current
design of the cache does not allow for an easy invalidation of entry for
such an event.

The new key schema leverages the fact that the main purpose of the cache
is to remove the performance hit from JSON unmarshalling. By having a
cache that uses marshalled JSON as a key, we are using it to optimize
exactly this operation.

The logic of extracting interfaces from the full inventory based on the
string slices instead of regular expressions has been chosen after
benchmarking, as a method giving the smallest performance hit, i.e.

* 100000 string operations
  * evaluateWithoutRegex took 398.699662ms
  * evaluateWithRegex took 7.286950236s
* 1000 string operations
  * evaluateWithoutRegex took 3.344477ms
  * evaluateWithRegex took 152.738194ms

Closes: [MGMT-11131](https://issues.redhat.com//browse/MGMT-11131)

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @ori-amizur 
/cc @carbonin 

## Checklist

## Reviewers Checklist


internal/common/common.go
```diff
@@ -377,6 +377,16 @@ func GetHostKey(host *models.Host) string {
 	return host.ID.String() + "@" + host.InfraEnvID.String()
 }
 
+func GetInventoryInterfaces(inventory string) string {
+	if (strings.Index(inventory, "\"interfaces\":")) == -1 {
```
ori-amizur, line 381: No reason to search for interfaces twice

internal/common/common.go
```diff
@@ -377,6 +377,16 @@ func GetHostKey(host *models.Host) string {
...
+		return ""
+	}
+
+	interfacesLocation := strings.Index(inventory, "\"interfaces\":") + len(`"interfaces:"`)
+	endLocation := strings.Index(inventory[interfacesLocation:], "}],") + len("}]")
```
ori-amizur, line 386: What if the endLocation is not found?

PR #4188, opened 2022-07-26
MGMT-11334: Add API models for service-to-agent host reclaim

CrystalChun:
[MGMT-11334](https://issues.redhat.com/browse/MGMT-11334)
As part of epic https://issues.redhat.com/browse/MGMT-7503
Adds the following:
1. Two new step types: download boot artifacts and reboot
2. Download boot artifacts request
3. Download boot artifacts response

Host reclaim requires two new steps for the agent to run,
and these models contain the necessary information for the
communication between the agent and the service.

[MGMT-11334](https://issues.redhat.com/browse/MGMT-11334)
As part of epic https://issues.redhat.com/browse/MGMT-7503

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed - only adds models/swagger

## Assignees

/cc @carbonin 
/cc @filanov 

## Checklist

## Reviewers Checklist


swagger.yaml
```diff
@@ -5428,6 +5430,34 @@ definitions:
...
+        type: string
+        description: URL address to download the initrd.
+
+  download_boot_artifacts_response:
+    type: object
+    description: The result of downloading artifacts to a host.
+    required:
+      - error
+    properties:
+      error:
```
carbonin, line 5458: Do we think it's worth creating this struct if it just has an error string?
The reply already has an `error` field.

I guess this is better if we ever wanted to expand it? @avishayt what do you think?

swagger.yaml
```diff
@@ -5428,6 +5430,34 @@ definitions:
...
+        type: string
+        description: URL address to download the kernel.
+      rootfs_url:
+        type: string
+        description: URL address to download the rootfs.
+      initrd_url:
+        type: string
+        description: URL address to download the initrd.
+
+  download_boot_artifacts_response:
```
carbonin, line 5452: Do we want/need a similar request or response for the reboot step?

@avishayt ? Is that a convention we want to continue even if the step is incredibly simple?

swagger.yaml
```diff
@@ -4100,6 +4100,8 @@ definitions:
       - logs-gather
       - next-step-runner
       - upgrade-agent
+      - download-boot-artifacts
+      - reboot
```
avishayt, line 4104: Should we make this one more specific?  If I were to see this outside of the context of this PR, I would assume that it had to do with the reboot during installation.
If the instructions to the agent are generic in this step and will always remain generic, so that the step can be used for any scenario where we want to tell the agent to reboot, then leave it as "reboot".  If, on the other hand, the agent will perform some action specific to this step, I would call it something more specific.

swagger.yaml
avishayt: step-reply already defines exit_code, output, and error.  Not all steps replies define their own response and it's fine not to.  In the worst case, we now have the ability to upgrade agents and can always expand it later if necessary.

swagger.yaml
avishayt: As I wrote above, I think it's not necessary.  Better to not redefine "error" here, and no need to define an empty struct.  We can always create it in the future if it becomes necessary and use agent-upgrade.

swagger.yaml
CrystalChun: Good points! I think this instruction can be generic since it simply reboots the host and it can be reused in other scenarios too.

@carbonin what do you think?

CrystalChun: I guess the only thing that could be more specific is also running `chroot /host` before `reboot` on the agent. I'm not sure if `chroot /host` would be necessary to reboot the host in other scenarios

swagger.yaml
CrystalChun: Makes sense we can use the pre-existing step-reply arguments. Thank you for your input!

@carbonin if you agree, I'll go ahead and remove this model

swagger.yaml
carbonin: Yeah, that bit is rather specific to our use case. I think it's probably safer to rename it to `reboot-for-reclaim` or something.

avishayt (approved):

PR #4191, opened 2022-07-27
MGMT-11362: Set api vip and not dnsname for day2 cluster (transform) for bm platform

tsorya:
[MGMT-11362](https://issues.redhat.com//browse/MGMT-11362): Set api vip and not dnsname for day2 cluster (transform) for bm platform
We see cases where user are not able to manage dns inside their org and
they are stuck with trying to add hosts. We know that in BM case there
is no need to set dns name if we can set api vip.
Changing tranform to day2 logic to set vip in case it can

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/cluster_test.go
+++ b/internal/cluster/cluster_test.go
@@ -2905,82 +2905,124 @@ var _ = Describe("Transform day1 cluster to a day2 cluster", func() {
 	})
 
 	tests := []struct {
-		name          string
-		clusterStatus string
-		clusterKind   string
-		errorExpected bool
+		name               string
+		clusterStatus      string
+		clusterKind        string
+		errorExpected      bool
+		userManagedNetwork bool
+		apiVip             string
 	}{
 		{
-			name:          "successfully transform day1 cluster to a day2 cluster - status installed",
-			clusterStatus: models.ClusterStatusInstalled,
-			clusterKind:   models.ClusterKindCluster,
-			errorExpected: false,
+			name:               "successfully transform day1 cluster to a day2 cluster - status installed - user managed network true",
+			clusterStatus:      models.ClusterStatusInstalled,
+			clusterKind:        models.ClusterKindCluster,
+			errorExpected:      false,
+			userManagedNetwork: true,
+			apiVip:             common.TestIPv4Networking.APIVip,
 		},
 		{
-			name:          "fail to transform day1 cluster to a day2 cluster - kind AddHostsCluster",
-			clusterStatus: models.ClusterStatusInstalled,
-			clusterKind:   models.ClusterKindAddHostsCluster,
-			errorExpected: true,
+			name:               "successfully transform day1 cluster to a day2 cluster - status installed - api vip is not set",
+			clusterStatus:      models.ClusterStatusInstalled,
+			clusterKind:        models.ClusterKindCluster,
+			errorExpected:      false,
+			userManagedNetwork: false,
+			apiVip:             "",
 		},
 		{
-			name:          "fail to transform day1 cluster to a day2 cluster - status insufficient",
-			clusterStatus: models.ClusterStatusInsufficient,
-			clusterKind:   models.ClusterKindCluster,
-			errorExpected: true,
+			name:               "successfully transform day1 cluster to a day2 cluster - status installed",
+			clusterStatus:      models.ClusterStatusInstalled,
+			clusterKind:        models.ClusterKindCluster,
+			errorExpected:      false,
+			userManagedNetwork: false,
+			apiVip:             common.TestIPv4Networking.APIVip,
 		},
 		{
-			name:          "fail to transform day1 cluster to a day2 cluster - status ready",
-			clusterStatus: models.ClusterStatusReady,
-			clusterKind:   models.ClusterKindCluster,
-			errorExpected: true,
+			name:               "fail to transform day1 cluster to a day2 cluster - kind AddHostsCluster",
+			clusterStatus:      models.ClusterStatusInstalled,
+			clusterKind:        models.ClusterKindAddHostsCluster,
+			errorExpected:      true,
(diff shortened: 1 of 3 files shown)


internal/common/common.go
```diff
@@ -393,6 +393,16 @@ func GetTagFromImageRef(ref string) string {
 	}
 }
 
+func GetAPIVipDNSName(c *Cluster) string {
```
eranco74, line 396: Consider renaming to GetIgnitionEndpoint?

internal/common/common.go
```diff
@@ -393,6 +393,16 @@ func GetTagFromImageRef(ref string) string {
...
+	//In case cluster that is not None platform and api vip is set we should set api vip as APIVipDNSName
+	// I know that APIVipDNSName name is confusing but in order not to add new variable we will use it
+	// as we allow already setting ip to it
+	if !swag.BoolValue(c.Cluster.UserManagedNetworking) && c.Cluster.APIVip != "" {
```
eranco74, line 400: Unsure we can trust UMN, but I guess if we have the API VIP we can just use it

internal/cluster/cluster_test.go
```diff
@@ -2905,82 +2905,124 @@ var _ = Describe("Transform day1 cluster to a day2 cluster", func() {
...
-		name          string
-		clusterStatus string
-		clusterKind   string
-		errorExpected bool
+		name               string
+		clusterStatus      string
+		clusterKind        string
+		errorExpected      bool
+		userManagedNetwork bool
+		apiVip             string
```
omertuc, line 2913:
```suggestion
		expectedAPIVipDNSName             string
```

internal/common/common.go
omertuc: I disagree, the literal APIVipDNSName cluster field gets set with the value being the string returned by this function

omertuc, line 396: Perhaps this function should be called "GetConvertedClusterAPIVipDNSName", because it assumes UMN is correctly set, which is only true during conversion

internal/common/common.go
tsorya: i prefer to verify both, in SNO case we will set api_vip but we still need to use dns name

internal/cluster/cluster_test.go
tsorya: this should be ip as i need  away to verify case where apiVip is not set

omertuc: The test is literally just checking whether the APIVipDNSName field on the cluster is equal to this parameter, so the name of this parameter should be modified to emphasize that it's an expectation (i.e. "test output") and not a test "input" like the rest of the parameters in this struct

omertuc: Ah sorry I misread the test

internal/cluster/cluster_test.go
```diff
@@ -3007,8 +3050,13 @@ var _ = Describe("Transform day1 cluster to a day2 cluster", func() {
 				Expect(db.Take(&c, "id = ?", cluster.ID.String()).Error).ToNot(HaveOccurred())
 				Expect(c.Kind).To(Equal(swag.String(models.ClusterKindAddHostsCluster)))
 				Expect(c.Status).To(Equal(swag.String(models.ClusterStatusAddingHosts)))
-				apiVipDnsname := fmt.Sprintf("api.%s.%s", c.Name, c.BaseDNSDomain)
-				Expect(c.APIVipDNSName).To(Equal(swag.String(apiVipDnsname)))
+				if t.apiVip == "" || t.userManagedNetwork {
+					apiVipDnsname := fmt.Sprintf("api.%s.%s", c.Name, c.BaseDNSDomain)
+					Expect(c.APIVipDNSName).To(Equal(swag.String(apiVipDnsname)))
+				} else {
+					Expect(c.APIVipDNSName).To(Equal(swag.String(t.apiVip)))
```
omertuc, line 3057: I think this demonstrates the intention better:

```suggestion
					Expect(c.APIVipDNSName).To(Equal(swag.String(cluster.APIVip)))
```

internal/cluster/cluster_test.go
omertuc: See https://github.com/openshift/assisted-service/pull/4191/files#r931196094 instead, it's the reason I got confused

internal/cluster/cluster_test.go
tsorya: sounds good

internal/common/common.go
```diff
@@ -393,6 +393,16 @@ func GetTagFromImageRef(ref string) string {
 	}
 }
 
+func GetConvertedClusterAPIVipDNSName(c *Cluster) string {
+	//In case cluster that is not None platform and api vip is set we should set api vip as APIVipDNSName
+	// I know that APIVipDNSName name is confusing but in order not to add new variable we will use it
```
eranco74, line 398: I don't see `APIVipDNSName` here, this comment seems unrelated

internal/common/common.go
```diff
@@ -393,6 +393,16 @@ func GetTagFromImageRef(ref string) string {
 	}
 }
 
+func GetConvertedClusterAPIVipDNSName(c *Cluster) string {
+	//In case cluster that is not None platform and api vip is set we should set api vip as APIVipDNSName
```
eranco74, line 397:
```suggestion
	//In case cluster that isn't configred with user-managed-networking and api vip is set we should set api vip as APIVipDNSName
```

eranco74 (commented):
/lgtm

PR #4196, opened 2022-07-28
bug-2109637: Agent validations out of sync with the host validation

eranco74:
Added NotifyInternalEvent to events Sender intervace

Now when the host/cluster monitor calls reportValidationStatusChanged it will send notification in case the validation status changed.
This notification will trigger reconciliation of the relevant agent
Prior to this commit status changes from Pending to Success/Failure were ignored by reportValidationStatusChanged

- Is this PR relying on CI for an e2e test run? yes
- Should this PR be tested in a specific environment? kube-api
- Any logs, screenshots, etc that can help with the review process? no

-->

https://bugzilla.redhat.com/show_bug.cgi?id=2109637

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @omertuc 
/cc @tsorya 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -266,9 +266,12 @@ func (m *Manager) reportValidationStatusChanged(ctx context.Context, c *common.C
 				if v.Status == ValidationFailure && currentStatus == ValidationSuccess {
 					m.metricAPI.ClusterValidationChanged(models.ClusterValidationID(v.ID))
 					eventgen.SendClusterValidationFailedEvent(ctx, m.eventsHandler, *c.ID, v.ID.String(), v.Message)
-				}
-				if v.Status == ValidationSuccess && currentStatus == ValidationFailure {
+				} else if v.Status == ValidationSuccess && currentStatus == ValidationFailure {
 					eventgen.SendClusterValidationFixedEvent(ctx, m.eventsHandler, *c.ID, v.ID.String(), v.Message)
+				} else if v.Status != currentStatus {
+					msg := fmt.Sprintf("Cluster %s: validation '%s' status changed from %s to %s",
+						*c.ID, v.ID.String(), currentStatus, v.Status)
+					m.eventsHandler.NotifyInternalEvent(ctx, c.ID, nil, nil, msg)
 				}
 			}
 		}
--- a/internal/events/api/event.go
+++ b/internal/events/api/event.go
@@ -17,6 +17,12 @@ type Sender interface {
 	// Use the prop field to add list of arbitrary key value pairs when additional information is needed (for example: "vendor": "RedHat")
 	AddEvent(ctx context.Context, clusterID strfmt.UUID, hostID *strfmt.UUID, severity string, msg string, eventTime time.Time, props ...interface{})
 	V2AddEvent(ctx context.Context, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID, name string, severity string, msg string, eventTime time.Time, props ...interface{})
+	// Used for events that are not interesting for metrics or for users, but we still
+	// want to raise them for internal notification between systems. e.g. notify kube-api
+	// that a cluster / host validation status changed so the CR conditions messages could
+	// be updated, but the status that changed is trivial (e.g. from pending to success) so it's
+	// not interesting to display to user or to be included in metrics
+	NotifyInternalEvent(ctx context.Context, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID, msg string)
 
 	//Add metric-related event. These events are hidden from the user and has 'metrics' Category field
 	AddMetricsEvent(ctx context.Context, clusterID strfmt.UUID, hostID *strfmt.UUID, severity string, msg string, eventTime time.Time, props ...interface{})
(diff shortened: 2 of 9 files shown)


internal/host/host.go
```diff
@@ -1050,11 +1050,14 @@ func (m *Manager) reportValidationStatusChanged(ctx context.Context, vc *validat
 					m.metricApi.HostValidationChanged(models.HostValidationID(v.ID))
 					eventgen.SendHostValidationFailedEvent(ctx, m.eventsHandler, *h.ID, h.InfraEnvID, h.ClusterID,
 						hostutil.GetHostnameForMsg(h), v.ID.String())
-				}
-				if v.Status == ValidationSuccess && currentStatus == ValidationFailure {
+				} else if v.Status == ValidationSuccess && currentStatus == ValidationFailure {
 					log.Infof("Host %s: validation '%s' is now fixed", hostutil.GetHostnameForMsg(h), v.ID)
 					eventgen.SendHostValidationFixedEvent(ctx, m.eventsHandler, *h.ID, h.InfraEnvID, h.ClusterID,
 						hostutil.GetHostnameForMsg(h), v.ID.String())
+				} else if v.Status != currentStatus {
```
omertuc, line 1057: is this code just for host validations, or does it also work for cluster validations?

internal/events/api/event.go
```diff
@@ -17,6 +17,7 @@ type Sender interface {
 	// Use the prop field to add list of arbitrary key value pairs when additional information is needed (for example: "vendor": "RedHat")
 	AddEvent(ctx context.Context, clusterID strfmt.UUID, hostID *strfmt.UUID, severity string, msg string, eventTime time.Time, props ...interface{})
 	V2AddEvent(ctx context.Context, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID, name string, severity string, msg string, eventTime time.Time, props ...interface{})
+	NotifyInternalEvent(ctx context.Context, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID, msg string)
```
omertuc, line 20:
```suggestion
	// Used for events that are not interesting for metrics or for users, but we still
	// want to raise them for internal notification between systems. e.g. notify kube-api
	// that a cluster / host validation status changed so the CR conditions messages could
	// be updated, but the status that changed is trivial (e.g. from pending to success) so it's
	// not interesting to display to user or to be included in metrics
	NotifyInternalEvent(ctx context.Context, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID, msg string)
```

internal/host/host.go
eranco74: Just host validatoins, need to think/check if we have the same issue there.

omertuc: Couldn't hurt to have it for cluster as well... It's hard to come up with / notice cases where these silent events are needed, to save future headache I think we should take care of cluster as well

eranco74: OK, from the code it seems that we might have the same problem there.
I'll add it

internal/host/host.go
```diff
@@ -1050,11 +1050,14 @@ func (m *Manager) reportValidationStatusChanged(ctx context.Context, vc *validat
 					m.metricApi.HostValidationChanged(models.HostValidationID(v.ID))
 					eventgen.SendHostValidationFailedEvent(ctx, m.eventsHandler, *h.ID, h.InfraEnvID, h.ClusterID,
 						hostutil.GetHostnameForMsg(h), v.ID.String())
-				}
-				if v.Status == ValidationSuccess && currentStatus == ValidationFailure {
+				} else if v.Status == ValidationSuccess && currentStatus == ValidationFailure {
```
omertuc, line 1053:
```suggestion
				} else if v.Status == ValidationSuccess || v.Status == ValidationPending && currentStatus == ValidationFailure {
```

omertuc: there are nicer ways to do this but you get the idea

internal/cluster/cluster.go
```diff
@@ -266,9 +266,12 @@ func (m *Manager) reportValidationStatusChanged(ctx context.Context, c *common.C
 				if v.Status == ValidationFailure && currentStatus == ValidationSuccess {
 					m.metricAPI.ClusterValidationChanged(models.ClusterValidationID(v.ID))
 					eventgen.SendClusterValidationFailedEvent(ctx, m.eventsHandler, *c.ID, v.ID.String(), v.Message)
-				}
-				if v.Status == ValidationSuccess && currentStatus == ValidationFailure {
+				} else if v.Status == ValidationSuccess && currentStatus == ValidationFailure {
```
omertuc, line 269: Same as [this](https://github.com/openshift/assisted-service/pull/4196/files#r932467611) but for clusters you don't have supress and you also don't have disabled

internal/host/host.go
eranco74: Unsure I understand what you suggested here.
This `else if` handles the case of a validation changing from success to failure.

eranco74: OK, this comment was ment for the validation failed section.
You are right we should send an event for every validation that failed (and wasn't failing before) e.g.
```
if v.Status == ValidationFailure && currentStatus != ValidationFailure {
```

Although it seems that we don't send events for failing validation (unless they move from success to failed) I'd like to look into it and fix it in another PR.

eranco74:
/hold
 the ztp test seems broken

eranco74:
@omertuc capital is broken due to changes in assisted test infra.
Thus can merge
/unhold

internal/host/host.go
eranco74: https://github.com/openshift/assisted-service/pull/4226

PR #4198, opened 2022-07-28
MGMT-10969: Ensure infraEnv name label is set on agents

carbonin:
Previously the infraEnv name label was only set on agent resources when
the agent was originally created. If the label was changed or deleted
this would break processes that rely on it being present.

https://issues.redhat.com/browse/MGMT-10969

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Unit test and subsystem tests should be sufficient.

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @ori-amizur 
/cc @danielerez 
/cc @CrystalChun 

## Checklist

## Reviewers Checklist


internal/controller/controllers/agent_controller.go
```diff
@@ -1022,23 +1026,29 @@ func (r *AgentReconciler) updateLabels(log logrus.FieldLogger, ctx context.Conte
...
+	err := r.Get(ctx, agentRef, agent)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get agent %s", agentRef)
+	}
+	return nil
+}
+
 func setAgentAnnotation(log logrus.FieldLogger, agent *aiv1beta1.Agent, key string, value string) bool {
 	annotations := agent.GetAnnotations()
+	if annotations == nil {
```
danielerez, line 1049: Is there a specific scenario it can be nil? Worth adding a test for it?

internal/controller/controllers/agent_controller.go
```diff
@@ -1226,6 +1236,20 @@ func (r *AgentReconciler) getIgnitionToken(ctx context.Context, ignitionEndpoint
...
+func (r *AgentReconciler) setInfraEnvNameLabel(ctx context.Context, log logrus.FieldLogger, h *common.Host, agent *aiv1beta1.Agent) error {
+	infraEnv, err := r.Installer.GetInfraEnvInternal(ctx, installer.GetInfraEnvParams{InfraEnvID: h.InfraEnvID})
+	if err != nil {
+		return err
+	}
+	if infraEnv.Name == nil {
+		return errors.Errorf("infraEnv %s name is nil", h.InfraEnvID)
+	}
+	if setAgentLabel(log, agent, aiv1beta1.InfraEnvNameLabel, *infraEnv.Name) {
+		return r.updateAndReplaceAgent(ctx, agent)
```
danielerez, line 1248: Do we need some debug/info log here to note that the agent label was set?

internal/controller/controllers/agent_controller.go
carbonin: I ran into this when running tests. I don't think it should really ever be nil in a real scenario.

I don't remember which of these tests I changed ran into it though. I could add an explicit one if you think it's needed.

internal/controller/controllers/agent_controller.go
carbonin: :man_shrugging: sure?

Not sure it's such an important thing to log, but I definitely can.

internal/controller/controllers/agent_controller.go
danielerez: Oh ok, I was just curious when it happens:) No need for test if you don't think it's necessary.

internal/controller/controllers/agent_controller.go
danielerez: Yeah not mandatory, just thought it could help on debug and such.

danielerez:
/lgtm
/hold

Unhold if ready.

PR #4199, opened 2022-07-28
MGMT-10542: Allow users to skip writing over specific disks - SaaS

omertuc:
tl;dr This is an implementation of a feature that gives users the
ability to tell Assisted to not format disks that it otherwise would.

# Background

After the assisted-installer writes the OCP operating system to the
host's installation disk and then reboots the host, we can only hope
that the disk that will boot after that reboot is the installation disk
and not some other device.

After the reboot, one of the following three things could happen:

1) The discovery ISO media gets booted, again. This is bad, but we can
detect this and let the user know.

2) Some disk other than the installation disk gets booted. This is bad,
and there's nothing we can do about it. We can't even tell for sure if
that happened.

3) The installation disk boots and everything is fine and installation
/ cluster initialization proceeds as expected.

On UEFI based systems we can usually make (3) happen using `efibootmgr`,
on BIOS based systems we cannot influence the boot process in the same
way. However, there is one thing yet we try to do to avoid (2) - during
installation the assisted-service asks the assisted-installer to attempt
to format the boot partition of some of the disks that the
assisted-agent recognized as "Bootable" (these are disks that have an
MBR bootloader as their first sector).

By attempting to clear the boot sector of these disks, we hope that BIOS
will no longer recognize these disks as bootable and so will not try to
automatically boot from them, thus preventing (2).

# The issue

ALL disks that we recognize as bootable (save for some special
exceptions) get formatted, all for the sake of ensuring the user won't
run into confusing boot order issues. This is an opinionated decision we
made to improve the ease of use of the Assisted Installer, but it comes
at a cost of destroying user disks, some of which could potentially be
important and might be used by the cluster / server hardware. Some users
may be unhappy with this decision and may want to have more control.

# The solution

Give users an abiltiy to opt out of formatting particular disk(s). When
users choose to opt out of formatting, the UI (not as part of this PR)
will show them a warning explaining that it is on them to ensure they
won't run into any boot order issues, and it's up to them to take care
of it.

# Implementation

Two new API changes have been introduced -

* New read-only `disks_to_be_formatted` column added to every host. This
  column is managed by the service. It's a comma separated list of disks
  that are automatically planned to undergo formatting once installation
  begins, unless otherwise set to be skipped by `skip_formatting_disks`.
  The reason this new API was added is because we would like the new UI
  for this feature to also show the users which disks are planned to
  undergo formatting.

* New user-updatable `skip_formatting_disks` column. Can be modified
  through a newly added host UPDATE API. The structure of this new
  UPDATE API is a list of disks, and for each disk, whether the user
  wants it removed-from or added-to the `skip_formatting_disks` list. In
  order to make sure that direct API users didn't make a typo (which
  could lead to potentially harmful formatting), this API would also
  validate that the disk ID that the user asked to skip actually appears
  in the inventory. This also ensures the user doesn't choose to skip a
  disk before the inventory arrives - as that could lead to the
  installation disk being chosen to be one of the skipped disks, which
  would be awkward.

Finally, when determining which disks the assisted-service should ask
the assisted-installer to format, the assisted-service will consult both
`disks_to_be_formatted` and `skip_formatting_disks` to decide on the
final list (the final list is a simple `set(disks_to_be_formatted)`
minus `set(skip_formatting_disks)`)

# Installation disk

I've also added a blocking validation to hosts which checks that the
installation disk of the host does not appear in the host's
`skip_formatting_disks` list. This validation would be the simplest way
to block the user from asking to "skip the formatting of the
installation disk" (which would obviously be non-sensical, as the
installation disk has to be formatted). Another way to solve this would
be to block the user during the host UPDATE API call, but that would be
mixing / tightly coupling the installation disk choice API with the new
skip formatting disks API. An "offline" validation happening in the
background is simpler.

# Missing skip-formatting disks validation

If for some reason one of the disks that is set to be skipped no longer
appears in the inventory, a new validation will block installation. This
is to ensure that if the user reboots the system and somehow the disk
they asked to skip the formatting of changed ID, then they would have to
manually and consciously remove it from the `skip_formatting_disks` list.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
Manual API interaction, unit-tests
- [] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist


internal/host/validator.go
```diff
@@ -1552,3 +1552,58 @@ func (v *validator) compatibleAgent(c *validationContext) (ValidationStatus, str
 	return ValidationFailure, "This host's agent is in the process of being upgraded to a compatible " +
 		"version. This might take a few minutes"
 }
+
+func (v *validator) noSkipInstallationDisk(c *validationContext) (ValidationStatus, string) {
+	const (
+		successMessage string = "No request to skip formatting of the installation disk"
+		failureMessage string = "Requesting to skip the formatting of the installation disk is not allowed. The installation disk has to be formatted. Please either change this host's installation disk or don't skip the formatting of the installation disk"
```
avishayt, line 1559:
```suggestion
		failureMessage string = "Requesting to skip the formatting of the installation disk is not allowed. The installation disk must be formatted. Please either change this host's installation disk or do not skip the formatting of the installation disk"
```

internal/host/validator.go
```diff
@@ -1552,3 +1552,58 @@ func (v *validator) compatibleAgent(c *validationContext) (ValidationStatus, str
...
+	if c.host.InstallationDiskID == "" || !funk.ContainsString(
+		common.GetSkippedFormattingDiskIdentifiers(c.host), c.host.InstallationDiskID) {
+
+		return ValidationSuccess, successMessage
+	}
+
+	return ValidationFailure, failureMessage
+}
+
+func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, string) {
```
avishayt, line 1573: This should never happen because the identifiers that we now use are hardware-based, but I appreciate your paranoia :)

avishayt (commented):
A few minor comments, but otherwise really nice job

swagger.yaml
```diff
@@ -5655,6 +5675,20 @@ definitions:
       role:
         $ref: "#/definitions/disk-role"
 
+  disk-skip-formatting-params:
+    type: object
+    description: Allows an addition or removal of a host disk from the host's skip_formatting_disks list
+    required:
+      - 'disk_id'
+      - 'skip_formatting'
```
jkilzi, line 5683: @omertuc why do you need this boolean flag? Isn't the presence of the disk-id in the disks-skip-formatting list enough?

avishayt: The API isn't to provide all the disks to skip formatting each time, and that replaces the previous list.  If a disk was previously selected to be skipped, you undo it by passing skip_formatting=false next time.
If you pass skip_formatting=true for a disk and it was already set to be skipped, or pass false and it was already false, then the API will accept it.

jkilzi: OK now I get it, 10x

PR #4201, opened 2022-07-29
Bug 2112321: Stop reconcile loop after changing CR inside BMAC

mkowalski:
This PR fixes the issue where returning `reconcileComplete{dirty: true}`
does not stop the reconciliation loop what leads to unexpected races and
objects being modified multiple times.

Closes: Bug-2112321

/cc @flaper87 
/cc @omertuc


flaper87 (approved):
I'm amazed how we got this far with things working considering the races we could've hit because of this bug.

Glad we found it :muscle:

flaper87:
Also, I think we should backport this.... all the way to 2.4 if needed

PR #4205, opened 2022-07-29
NO-ISSUE: Fix conversion of disk size for device hints

mkowalski:
This PR fixes the wrong formula for converting bytes to gigabytes.

/cc @omertuc


omertuc:
Let's use our functions from `assisted-service/pkg/conversions/conversions.go`

mkowalski:
> Let's use our functions from `assisted-service/pkg/conversions/conversions.go`

Of course, didn't realize we have that widely used

PR #4212, opened 2022-08-01
Bug 2112817: Stop reconcile loop after changing CR inside BMAC

mkowalski:
This PR fixes the issue where returning `reconcileComplete{dirty: true}`
does not stop the reconciliation loop what leads to unexpected races and
objects being modified multiple times.

Cherry-picks: #4201
Closes: Bug-2112817

/cc @flaper87


flaper87 (approved):
LGTM! Thanks for the backports

PR #4219, opened 2022-08-01
MGMT-11416: support platform in kube-api

slaviered:
Add support for defining PlatformType in Kube API.

The PlatformType names are aligned with OpenShift PlatformTypes.
Since the BE can set the platform type and the UserManagedNetworking flag,they are both reflected in the Status section.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/hiveextension/v1beta1/agentclusterinstall_types.go
+++ b/api/hiveextension/v1beta1/agentclusterinstall_types.go
@@ -171,6 +171,10 @@ type AgentClusterInstallSpec struct {
 	// Proxy defines the proxy settings used for the install config
 	// +optional
 	Proxy *Proxy `json:"proxy,omitempty"`
+
+	// PlatformType is the name for the specific platform upon which to perform the installation.
+	// +optional
+	PlatformType PlatformType `json:"platformType,omitempty"`
 }
 
 // IgnitionEndpoint stores the data to of the custom ignition endpoint.
@@ -227,6 +231,14 @@ type AgentClusterInstallStatus struct {
 	// IngressVIP is the virtual IP used for cluster ingress traffic.
 	// +optional
 	IngressVIP string `json:"ingressVIP,omitempty"`
+
+	// UserManagedNetworking indicates if the networking is managed by the user.
+	// +optional
+	UserManagedNetworking *bool `json:"userManagedNetworking,omitempty"`
+
+	// PlatformType is the name for the specific platform upon which to perform the installation.
+	// +optional
+	PlatformType PlatformType `json:"platformType,omitempty"`
 
 	// ValidationsInfo is a JSON-formatted string containing the validation results for each validation id grouped by category (network, hosts-data, etc.)
 	// +optional
@@ -327,6 +339,21 @@ const (
 const (
 	MasterAgentMachinePool string = "master"
 	WorkerAgentMachinePool string = "worker"
+)
+
+// PlatformType is a specific supported infrastructure provider.
+// +kubebuilder:validation:Enum="";BareMetal;None;VSphere
+type PlatformType string
+
+const (
+	// BareMetalPlatformType represents managed bare metal infrastructure.
+	BareMetalPlatformType PlatformType = "BareMetal"
+
+	// NonePlatformType means there is no infrastructure provider.
+	NonePlatformType PlatformType = "None"
+
+	// VSpherePlatformType represents VMWare vSphere infrastructure.
+	VSpherePlatformType PlatformType = "VSphere"
 )
 
 // AgentMachinePool is a pool of machines to be installed.
--- a/config/crd/bases/extensions.hive.openshift.io_agentclusterinstalls.yaml
+++ b/config/crd/bases/extensions.hive.openshift.io_agentclusterinstalls.yaml
@@ -310,6 +310,15 @@ spec:
                       true or leave empty.
                     type: boolean
                 type: object
+              platformType:
+                description: PlatformType is the name for the specific platform upon
+                  which to perform the installation.
+                enum:
+                - ""
+                - BareMetal
+                - None
+                - VSphere
+                type: string
               provisionRequirements:
                 description: ProvisionRequirements defines configuration for when
                   the installation is ready to be launched automatically.
@@ -443,6 +452,15 @@ spec:
                   - cidr
                   type: object
                 type: array
+              platformType:
+                description: PlatformType is the name for the specific platform upon
+                  which to perform the installation.
+                enum:
+                - ""
+                - BareMetal
+                - None
+                - VSphere
+                type: string
               progress:
                 description: Progress shows the installation progress of the cluster
                 properties:
@@ -453,6 +471,10 @@ spec:
                 required:
                 - totalPercentage
                 type: object
+              userManagedNetworking:
+                description: UserManagedNetworking indicates if the networking is
+                  managed by the user.
+                type: boolean
               validationsInfo:
                 additionalProperties:
                   items:
--- a/deploy/olm-catalog/manifests/extensions.hive.openshift.io_agentclusterinstalls.yaml
+++ b/deploy/olm-catalog/manifests/extensions.hive.openshift.io_agentclusterinstalls.yaml
@@ -308,6 +308,15 @@ spec:
                       true or leave empty.
                     type: boolean
                 type: object
+              platformType:
+                description: PlatformType is the name for the specific platform upon
+                  which to perform the installation.
+                enum:
+                - ""
+                - BareMetal
+                - None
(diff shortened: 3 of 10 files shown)


api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -171,6 +171,11 @@ type AgentClusterInstallSpec struct {
 	// Proxy defines the proxy settings used for the install config
 	// +optional
 	Proxy *Proxy `json:"proxy,omitempty"`
+
+	// PlatformType is the name for the specific platform upon which to perform the installation.
+	// +kubebuilder:validation:Enum=baremetal;vsphere;ovirt;none
```
mhrivnak, line 176: Can we use the same values as the openshift API? https://github.com/openshift/api/blob/9368b92/config/v1/types_infrastructure.go#L127

slaviered: Sounds like a good idea. It will require adding a conversion to assisted-service values.

slaviered: @avishayt Please also review and comment on the topic of API conversion between OpenShift's and AI's platform type names. I uploaded a code version that exposes the Openshift type for our ztp API.

avishayt:
Thanks.  Some high-level comments:
1. We recently removed support for oVirt, so there is no need to expose that.
2. We should check that the specified platform is supported by us (baremetal, none, vsphere) - I guess in a validating webhook?
3. We should check that platform, UserManagedNetworking, and SNO (ProvisionRequirements.ControlPlaneAgents==1) can all work together - also in a validating webhook.

slaviered:
@avishayt  
(2): This suppose to do by the validation logic in kubeapi server following the kubebuilder:validation tag. I am running a manual test to confirm this indeed work as expected.
(3): Those checks are already done by the BE.

avishayt:
(2) OK, validate that it actually has an effect
(3) Right, so we'll get a sync error, I guess that's fine.  Just please be careful - in the REST API we can have all kinds of strange effects, for example setting SNO will change the platform to "none" and things like that.  We need to make sure there are no leftover effects from previous updates.

config/crd/bases/extensions.hive.openshift.io_agentclusterinstalls.yaml
```diff
@@ -310,6 +310,31 @@ spec:
                       true or leave empty.
                     type: boolean
                 type: object
+              platformType:
+                allOf:
+                - enum:
```
slaviered, line 315: @mhrivnak it seems that we can not redefine the validation list but only add on top of it. I suggest to leave it then as a custom type of PlatformType and only align the names. Is that OK? The option of a webhook validation is misleading because the CRD explicitly declares the full list as something we support.

deploy/olm-catalog/manifests/extensions.hive.openshift.io_agentclusterinstalls.yaml
```diff
@@ -308,6 +308,31 @@ spec:
...
+                - enum:
+                  - ""
+                  - BareMetal
+                  - None
+                  - VSphere
+                  - oVirt
+                description: PlatformType is the name for the specific platform upon
+                  which to perform the installation.
+                type: string
               provisionRequirements:
```
slaviered, line 336: After discussions with @avishayt: we should add the actual PlatfromType and user managed networking to the status because the BE may change / set the values of these fields according to the combination of requested Platform, UMN and HA mode

config/crd/bases/extensions.hive.openshift.io_agentclusterinstalls.yaml
mhrivnak: Yes, I think that's fine. The important part is that the string values are the same as the openshift API, and the validation works correctly.

avishayt (approved):

filanov:
@slaviered the commit is pointing to the epic, please use the tasks and not the epics for the commits

filanov:
@slaviered @avishayt i'm not sure if you already talked about it but will it be confusing to the user that he have two playform types? on in [Cluster Deployment](https://github.com/openshift/hive/blob/56adaaacf5f8075e3ad0896dac35243a863ec07b/apis/hive/v1/clusterdeployment_types.go#L124)  and one in the agent install?

avishayt:
@filanov ClusterDeployment has a Platform struct, but not a PlatformType string (unless I missed something).  There are other properties in ClusterDeployment that we ignore, so I think it's fine.  Do you have a better suggestion?

slaviered:
The PlatformType in ACI is extension to the standard Platform in ClusterDeployment which is a standard one and does not fit our purposes.

filanov:
I don't have a better suggestion. 
Maybe the problem is with the agent platform that is talking about bare metal `AgentBareMetal`. 
and there is already a [VSphere](https://github.com/openshift/hive/blob/56adaaacf5f8075e3ad0896dac35243a863ec07b/apis/hive/v1/clusterdeployment_types.go#L624)
Just want to make sure that this is clear to the users.

avishayt:
`AgentBareMetal` is an unfortunate name.  But I think it should be clear to users reading the docs.

filanov:
How was this tested?

slaviered:
This was tested both via subsystem and manually and there are couple of PRs in assisted-test-infra by @danielerez that insert the new functionality to the CI and to test-infra and they were tested E2E with actual VSphere platforms to prove that an atual cluster was created.

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -171,6 +171,10 @@ type AgentClusterInstallSpec struct {
 	// Proxy defines the proxy settings used for the install config
 	// +optional
 	Proxy *Proxy `json:"proxy,omitempty"`
+
+	// PlatformType is the name for the specific platform upon which to perform the installation.
+	// +optional
+	PlatformType PlatformType `json:"platformType,omitempty"`
```
danielerez, line 177: Can you change the object to something like 'PlatformTypeEnum'? Seems ACI isn't populating PlatformType correctly when using the same name for both.

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -329,6 +341,21 @@ const (
 	WorkerAgentMachinePool string = "worker"
 )
 
+// PlatformType is a specific supported infrastructure provider.
+// +kubebuilder:validation:Enum="";BareMetal;None;VSphere
```
filanov, line 345: @avishayt if i remember correctly we wanted to avoid enums no?

filanov: i don't remember the reason, just remember that we talked about it and somehow changes here are considered breaking changes or something like that

api/hiveextension/v1beta1/agentclusterinstall_types.go
slaviered: It is working on the CI and on my test machine. I wish to leave it like this because this is the naming convention of Openshift.

api/hiveextension/v1beta1/agentclusterinstall_types.go
avishayt: The problem with enums is when reporting status because adding a new value breaks clients (e.g., a client checking progress in a state machine and all of the sudden a new state is added).  However, in the spec it should be OK because if we add a new value, it wouldn't affect the others.

filanov: but the same type is used in the spec and in the status here

avishayt: That's a good point.  The cleanest API would create several conditions: PlatformBareMetal, PlatformNone, PlatformVsphere, and UserManagedNetworking.  Maybe overkill though?

danielerez: Won't it break clients as well if we drop support for a platform? Or we'll have to keep maintaining separate conditions also for deprecated platforms :/ Any way, not sure a scenario of a state machine is relevant here, as the platform in status is simply derived from the spec.

avishayt: I think it's fine the way it is

PR #4228, opened 2022-08-02
MGMT-3156: Enhancement - easy discovery password

omertuc:

- [x] Documentation

- [x] None

## How was this code tested?

- [x] No tests needed

## Checklist

## Reviewers Checklist


mkowalski (commented):
I like it very much, thanks! I wonder what will be the way for us to implement it in the most elegant way (I'm worried that it will be yet another ignition override that we will be internally merging) but I don't feel the need to cover it deeply in the enhancement.

/approve

docs/enhancements/easy-discovery-password.md
```diff
@@ -0,0 +1,174 @@
...
+needed.
+
+## Proposal
+
+### New service APIs
+
+#### REST API
+
+- Add a new optional `discovery_core_user_password_hash` parameter to the
+`infra-env-create-params` API structure. Ensure this parameter has
```
avishayt, line 68: "discovery" is redundant because it's already part of the infra-env.  "core_user_password_hash" is sufficient.

docs/enhancements/easy-discovery-password.md
```diff
@@ -0,0 +1,174 @@
...
+`description:` documentation explaining how the hash can be calculated and what
+its format should look like, preferably with an example. The service will use
+this new parameter to set the `core` user discovery ISO password hash. Since
+this is just a salted hash, the service has no practical way to figure out what
+the password actually is.
+
+#### Kubernetes API
+
+- Extend the InfraEnv CR to support a similar (see REST API section above)
+`discovery_core_user_password_hash` parameter.
```
avishayt, line 78: Same comment as the REST API - "core_user_password_hash" is sufficient.

docs/enhancements/easy-discovery-password.md
avishayt: It should also be in infra-env-update-params and infra-env, otherwise it sounds like it can be set on create and never updated or viewed.

docs/enhancements/easy-discovery-password.md
```diff
@@ -0,0 +1,174 @@
...
+
+### UI changes
+
+- Extend the discovery ISO generation UI modal to let users input a password.
+If and when the user provides a password in that field, UI will calculate the
+hash for the user-provided password and will set the `discovery_core_user_password_hash`
+parameter accordingly.
```
avishayt, line 85: If the user exits that modal and returns, the UI can't show the password because the backend doesn't return it (not a bad thing, just a fact).  Suggest showing ********* if core_user_password_hash is not empty string.

avishayt: @nirfarkas should review this

omertuc: >the UI can't show the password because the backend doesn't return it

The UI can technically store the password in browser local storage, but I agree just forgetting the password is probably best

>Suggest showing ********* if core_user_password_hash is not empty string.

A bit problematic because we don't even know the length of the password

avishayt: It doesn't need to have the right amount of *, actually having it always show the same amount is more secure.  Anyway, @nirfarkas can figure out the best way to indicate that a password was set and to allow updating it.

omertuc:
We put this on hold for now due to potential security compliance concerns

omertuc:
Closed. Might pick it up again in the future

PR #4189, opened 2022-07-26
MGMT-8890: Integrate support for multiarch release images

mkowalski:
This PR adds support of multiarch OCP release images to the Assisted
Service.

The following happens now when a multiarch release image is provided via
the ClusterImageSet

1) cluster.CPUArchitecture is set as "multiarch"
2) validation of existence of OS image matching version and architecture
   is offloaded to the InfraEnv when the ISO is created
3) validation of matching cluster's and InfraEnv's architecture is
   removed

The rationale for (1) is that if using multiarch release image, the
cluster is no longer bound to one particular architecture but can rather
be deployed with any architecture as long as release image contains an
image for it and OS image that matches the version and architecture
exists.

The rationale for (2) is that the set of architectures supported by the
release image can be bigger than set of architectures supported by OS
images available for the Assisted Service. Given that we cannot enforce
users to provide OS image for every arch from the OCP release image,
this validation happens when InfraEnv tries to create an ISO. If it
happens that InfraEnv is created for an architecture supported by the
release image but without OS image available, a clear error message will
be shown.

The rationale for (3) is that this work is a preliminary work to support
heterogenous OCP clusters. Therefore we don't want to strictly limit
InfraEnvs bound to the specific ACI to contain only single architecture.

Given that (2) and (3) are technically allowing for some installation
schemas that we may not want to immediately support, we may want to add
additional validations in a subsequent PR. This is however left outside
of the scope of this change here as it clearly depends on the timelines
for releasing specific features like multiarch OCP images and
heterogenous OCP clusters.

Supersedes: #4185
Contributes-to: [MGMT-8890](https://issues.redhat.com//browse/MGMT-8890)
Closes: [MGMT-11353](https://issues.redhat.com//browse/MGMT-11353)
Closes: [MGMT-11354](https://issues.redhat.com//browse/MGMT-11354)

## List all the issues related to this PR

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [x] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [x] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Assignees

/cc @ori-amizur 
[...]

PR #4230, opened 2022-08-02
MGMT-8890: Handle registries.conf without any mirrors correctly

mkowalski:
Currently the code assumes that existence of registries.conf file means
there are always registry mirrors configured. It has been observed that
the default file has no `[[registry]]` section but is still a correct
file. For such a scenario we want to proceed as the file did not exist
instead of failing.

/cc @carbonin 
/cc @danielerez

### DIFF
--- a/pkg/mirrorregistries/generator.go
+++ b/pkg/mirrorregistries/generator.go
@@ -2,7 +2,7 @@ package mirrorregistries
 
 import (
 	"fmt"
-	"io/ioutil"
+	"os"
 
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/pelletier/go-toml"
@@ -17,10 +17,15 @@ type MirrorRegistriesConfigBuilder interface {
 }
 
 type mirrorRegistriesConfigBuilder struct {
+	MirrorRegistriesConfigPath      string
+	MirrorRegistriesCertificatePath string
 }
 
 func New() MirrorRegistriesConfigBuilder {
-	return &mirrorRegistriesConfigBuilder{}
+	return &mirrorRegistriesConfigBuilder{
+		MirrorRegistriesConfigPath:      common.MirrorRegistriesConfigPath,
+		MirrorRegistriesCertificatePath: common.MirrorRegistriesCertificatePath,
+	}
 }
 
 type RegistriesConf struct {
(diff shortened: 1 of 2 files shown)


mkowalski:
I opened this PR because of https://coreos.slack.com/archives/CUPJTHQ5P/p1659446457033599 and the fact that CI in https://github.com/openshift/assisted-service/pull/4189 fails. What I observe there is that we run with a default `/etc/containers/containers.conf` which has no `[[registry]]` section but is still a valid file.

carbonin:
How did we get into this situation?

Did we start providing the containers.conf for some other reason?

pkg/mirrorregistries/generator.go
```diff
@@ -66,21 +66,23 @@ func extractLocationMirrorDataFromRegistries(registriesConfToml string) ([]Regis
 
 	registriesTree, ok := tomlTree.Get("registry").([]*toml.Tree)
 	if !ok {
-		return nil, fmt.Errorf("Failed to cast registry key to toml Tree")
+		// If /etc/containers/containers.conf exists but has no "registry" key we don't want to fail
+		// but we proceed as if the file did not exist at all.
+		return nil, nil
```
carbonin, line 71: Would this be better in `IsMirrorRegistriesConfigured`?

carbonin:
Also, can you add a unit test for this?

pkg/mirrorregistries/generator.go
romfreiman: Log if not?

mkowalski:
>How did we get into this situation?
>Did we start providing the containers.conf for some other reason?

Yes, in fact yes. We are adding `skopeo` as a dependency now and one of its dependencies is creating this file with the default content

mkowalski:
/retitle MGMT-8890: Handle registries.conf without any mirrors correctly

pkg/mirrorregistries/generator.go
mkowalski: Given how those functions are called from around the code, I think it's the safest to return `false` in `IsMirrorRegistriesConfigured` when file exists but has no registries, but also to return nil without error in `extractLocationMirrorDataFromRegistries`. That's because there are calls around to both of those functions so I think it is safer to handle it twice

carbonin: This should never happen now, right?

pkg/mirrorregistries/generator_test.go
```diff
@@ -60,4 +72,51 @@ var _ = Describe("MirrorRegistriesConfig", func() {
...
+		m := mirrorRegistriesConfigBuilder{
+			MirrorRegistriesConfigPath:      tempRegistriesFile.Name(),
+			MirrorRegistriesCertificatePath: tempCAFile.Name(),
+		}
+		res := m.IsMirrorRegistriesConfigured()
+		Expect(res).Should(Equal(false))
+	})
+
+	It("check if mirrors configured - registry.conf has registries, CA exists", func() {
+		tempRegistriesFile, err := ioutil.TempFile(os.TempDir(), "registries.*.conf")
```
carbonin, line 105: This function is deprecated. ref: https://pkg.go.dev/io/ioutil@go1.19#TempFile

Can you use `os.CreateTemp` instead?

pkg/mirrorregistries/generator_test.go
```diff
@@ -45,19 +48,90 @@ var _ = Describe("MirrorRegistriesConfig", func() {
...
+	It("check if mirrors configured - registry.conf exists with garbage, CA exists", func() {
+		tempRegistriesFile, err := ioutil.TempFile(os.TempDir(), "registries.*.conf")
+		defer os.Remove(tempRegistriesFile.Name())
+		Expect(err).NotTo(HaveOccurred())
+		_, _ = tempRegistriesFile.WriteString(configWithGarbage)
+
+		tempCAFile, err := ioutil.TempFile(os.TempDir(), "ca.*.crt")
+		defer os.Remove(tempCAFile.Name())
+		Expect(err).NotTo(HaveOccurred())
+		_, _ = tempCAFile.WriteString("some random CA certificate")
```
carbonin, line 109: You could extract most of this into a `BeforeEach` and `AfterEach` in a context like `Context("with temp files"...` or something.

Then each test only needs to write the content to the file (which is the important part).

mkowalski: Yes that makes sense

carbonin (approved):
Great work, thanks for your patience with all my opinions about test code :sweat_smile:

PR #4234, opened 2022-08-03
Bug 2113954: Fix order when merging ignitions

mkowalski:
This PR changes the order of calling ignition merge function in BMAC.
This is because the function is not performing a merge towards the
latest provided ignition but merges towards the ignition version
provided as a second parameter.

Because of this behaviour we have observed cases where user providing
ignition override v3.2.0 was receiving a rendered ignition v3.1.0 with
some data lost (i.e. features introduced in 3.2 and not available in
3.1).

Fixes: Bug-2113954
Fixes: Bug-2099929
Closes: [MGMTBUGSM-447](https://issues.redhat.com//browse/MGMTBUGSM-447)
Closes: [MGMTBUGSM-499](https://issues.redhat.com//browse/MGMTBUGSM-499)

/cc @carbonin

### DIFF
--- a/internal/ignition/ignition_test.go
+++ b/internal/ignition/ignition_test.go
@@ -411,6 +411,7 @@ SV4bRR9i0uf+xQ/oYRvugQ25Q7EahO5hJIWRf4aULbk36Zpw3++v2KFnF26zqwB6
 	})
 })
 
+// TODO(deprecate-ignition-3.1.0)
 var _ = Describe("createHostIgnitions", func() {
 	const masterIgn = `{
 		  "ignition": {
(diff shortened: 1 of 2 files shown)


internal/ignition/ignition_test.go
```diff
@@ -813,69 +816,154 @@ var _ = Describe("downloadManifest", func() {
...
-		Expect(err).ToNot(HaveOccurred())
-		v32Config, _, err := config_32.Parse(bytes)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(v32Config.Ignition.Version).To(Equal("3.2.0"))
 	})
 
-	It("does not parse v99 config", func() {
-		_, err := ParseToLatest([]byte(v99ignition))
-		Expect(err.Error()).To(ContainSubstring("unsupported config version"))
+	Context("using MergeStructTranscribe", func() {
```
carbonin, line 905: I don't understand these tests. They look like they're just testing the ignition library, shouldn't these be using `MergeIgnitionConfig`?

internal/ignition/ignition_test.go
```diff
@@ -813,69 +816,154 @@ var _ = Describe("downloadManifest", func() {
...
+			res, err := json.Marshal(mergeResult)
+			Expect(err).ToNot(HaveOccurred())
+
+			version, err := jsonparser.GetString(res, "ignition", "version")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(version).To(Equal("3.2.0"))
+		})
+
+		// Be aware, this combination is counterintuitive and comes from the fact that MergeStructTranscribe()
+		// is not order-agnostic and prefers the field coming from the override rather from the base.
```
carbonin, line 952: >prefers the field coming from the override rather from the base

This is actually something I'm a bit concerned about. Specifically, will switching to order to achieve the higher version end up causing issues when we want to actually override an existing value in the base?

Say we provide some file at `/tmp/example` in the base ignition at version 3.2 and someone provides an override for `/tmp/example` with version 3.1. What happens and what do we expect to happen?

I think the main thing we want to ensure is that the override file content is what ends up in the resulting ignition. Can you add a test for this case?

internal/ignition/ignition_test.go
mkowalski: We have tests that use library (because my impression was that we seem to use library without fully understand it) and also tests that use `MergeIgnitionConfig` (just below)

carbonin:
We also have a table for the pointer (host) ignition versions in the same doc:

https://github.com/openshift/assisted-service/blob/master/docs/user-guide/install-customization.md#pointer-ignition

mkowalski:
> We have had a bunch of issues like this and documented that the discovery ignition override needs to be 3.1.
> 
> ref: https://github.com/openshift/assisted-service/blob/master/docs/user-guide/install-customization.md#discovery-ignition
> 
> > The discovery ignition must use version 3.1.0 regardless of the version of the cluster that will eventually be created.

There is no any explanation for this therefore I treat it as stale and not valid (of course unless proven otherwise).


> We also have a table for the pointer (host) ignition versions in the same doc:
> 
> https://github.com/openshift/assisted-service/blob/master/docs/user-guide/install-customization.md#pointer-ignition

This is exactly the problem this PR is trying to address. We have installation of 4.10 and we rendered pointer ignition 3.1 - so we already now are not complying to the table linked

internal/ignition/ignition_test.go
mkowalski: I see the scenario... It's just by looking at our base template I'm very close to saying that if someone tries to override a value from the base template, they are doing something that's wrong...

I agree having `/tmp/example` in the base and in the override is problematic, but here we have more components

- our base hardcoded in the service
- base coming from the MCS (that we link as part of `merge` stanza)
- override provided by the user

If any 2 out of 3 above are trying to touch the same field, it's already a shady area

mkowalski:
As discussed offline, all changes were now squashed to the changing order of the call in BMAC. Unit tests added so that behaviour of merger function is clear.

internal/ignition/ignition_test.go
```diff
@@ -813,69 +814,92 @@ var _ = Describe("downloadManifest", func() {
 	})
 })
 
-var _ = Describe("ParseToLatest", func() {
+var _ = Describe("parse and merge ignitions", func() {
```
carbonin, line 817:
```suggestion
var _ = Context("with test ignitions", func() {
```

internal/ignition/ignition_test.go
```diff
@@ -813,69 +814,92 @@ var _ = Describe("downloadManifest", func() {
...
-		Expect(err).ToNot(HaveOccurred())
-		Expect(config.Ignition.Version).To(Equal("3.2.0"))
 
-		bytes, err := json.Marshal(config)
-		Expect(err).ToNot(HaveOccurred())
-		v32Config, _, err := config_32.Parse(bytes)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(v32Config.Ignition.Version).To(Equal("3.2.0"))
-	})
+	Context("using ParseToLatest", func() {
```
carbonin, line 824:
```suggestion
	Describe("ParseToLatest", func() {
```

internal/ignition/ignition_test.go
```diff
@@ -813,69 +814,92 @@ var _ = Describe("downloadManifest", func() {
...
+			v31Config, _, err := config_31.Parse(bytes)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(v31Config.Ignition.Version).To(Equal("3.1.0"))
+		})
 	})
 
-	It("parses a v31 config with v31 override as 3.1.0", func() {
-		merge, err := MergeIgnitionConfig([]byte(v31ignition), []byte(v31override))
-		Expect(err).ToNot(HaveOccurred())
+	Context("using MergeIgnitionConfig", func() {
```
carbonin, line 850:
```suggestion
	Describe("MergeIgnitionConfig", func() {
```

internal/ignition/ignition_test.go
```diff
@@ -813,69 +814,92 @@ var _ = Describe("downloadManifest", func() {
...
+			Expect(config.Ignition.Version).To(Equal("3.1.0"))
+
+			bytes, err := json.Marshal(config)
+			Expect(err).ToNot(HaveOccurred())
+			v31Config, _, err := config_31.Parse(bytes)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(v31Config.Ignition.Version).To(Equal("3.1.0"))
+		})
+
+		It("does not parse v99 config", func() {
```
carbonin, line 898: This should go into the `ParseToLatest` describe block

carbonin (commented):
Looks good, just some comments on the test block naming.

carbonin (approved):

mkowalski:
/test edge-unit-test

Worked locally, not sure what's the glitch

mkowalski:

We hit a race

```
    Operation cannot be fulfilled on agents.agent-install.openshift.io "dbf22936-eb97-44a8-9730-f6b1311605b5": the object has been modified; please apply your changes to the latest version and try again
/assisted-service/subsystem/kubeapi_test.go:3109}
```

mkowalski:
multiarch job detects everything as expected
```
"For release image quay.io/openshift-release-dev/ocp-release:4.11.0-rc.6-multi detected version: 4.11.0-multi-rc.6"
"For release image quay.io/openshift-release-dev/ocp-release:4.11.0-rc.6-multi detected architecture: multiarch"
"selected cluster release image: arch=multiarch, openshiftVersion=4.11.0-multi-rc.6, url=quay.io/openshift-release-dev/ocp-release:4.11.0-rc.6-multi""Successfully registered cluster assisted-test-cluster with id ce2e08eb-c27c-4589-a028-1724177263d0"
```

PR #4237, opened 2022-08-03
MGMT-11447: Improve ignition downloadable (day2 API connectivity check) validation message

omertuc:
The current message (constructed by the UI) is "API IP connectivity
failure - To continue installation, configure your DNS or alternatively
<a>Update cluster with API IP</a>"

There are multiple issues with the current message:

1) The title says "API IP connectivity failure". This is not technically
   correct. The issue is reaching the API in general. It doesn't
   necessarily have anything to do with an "IP address".

2) The suggestion to "configure your DNS or Update cluster with API IP"
   is misleading - it seems to imply that if the user chooses to not
   configure their DNS, they can simply "update the cluster with API IP"
   and everything will work fine. This is not necessarily correct -
   none-platform clusters require the user to configure DNS, regardless
   of whether they provide us with the IP address of their cluster or
   not. The worker will simply not be able to join without it.

3) "update the cluster with API IP" needs rewording. "Updating a
   cluster" doesn't mean much to someone without internal AI developer
   perspective. It's extra confusing in day-2 because they already have
   a cluster. The concept of a "cluster" that the message is referring
   to is unknown to the user - us reusing the day-1 internal "cluster"
   entity for adding day-2 workers is merely an implementation detail
   and not something they user should be aware of, yet this message
   expects them to be aware of it. When read superficially by a user,
   this message seems to imply they have to make some change to their
   actual cluster, and it's confusing, because they don't.

To solve this we're gonna make the backend build most of the message.

1) The title should instead say "API connectivity failure"
- This will be done as part of the UI PR

2) The message should be more verbose about what actually happened:
   "This host has failed to contact the cluster's API at <API URL> with
   the following error: <error>. Please ensure the host can reach the
   API", and to that the UI will add to that something like
   "Alternatively, set a different API hostname <link>" which will allow
   lead the users to change the hostname setting dialog

It's important to include the entire URL in the message because we're
trying to contact the API at a very specific port number, and this is a
port number that's likely to be blocked by a firewall, so they user
should be aware of it. It's also important to include the entire URL
because the API domain name is something the UI crafts without user
interaction, when importing the cluster from OCM, so it's possible that
it's not even correct - the user needs to see it.

The agent will be separately improved to provide the service with both
the URL it tried to contact (which it itself will get from the service,
but we include it in the response anyway in case it changed and the
agent didn't get a chance to try the new URL yet) and the agent will
also include the error message it encountered when attempting to
download the ignition from the API - DNS error, connection refused,
whatever - the user should not be forced to SSH and watch agent logs to
figure out what went wrong.

Also changed the validation to be a `ValidationError` whenever the
service fails to deserialize the agent response, instead of a
"ValidationFailure".

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
    Unit tests
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -620,20 +620,35 @@ func (v *validator) isIgnitionDownloadable(c *validationContext) (ValidationStat
 	if c.infraEnv != nil {
 		return ValidationSuccessSuppressOutput, ""
 	}
+
 	if !hostutil.IsDay2Host(c.host) {
 		return ValidationSuccessSuppressOutput, ""
 	}
+
 	if c.host.APIVipConnectivity == "" {
 		return ValidationPending, "Ignition is not yet available, pending API connectivity"
 	}
-	var response models.APIVipConnectivityResponse
-	if err := json.Unmarshal([]byte(c.host.APIVipConnectivity), &response); err != nil || !response.IsSuccess {
-		return ValidationFailure, "Ignition is not downloadable. Please ensure host connectivity to the cluster's API"
+
+	var apiConnectivityResponse models.APIVipConnectivityResponse
+	if err := json.Unmarshal([]byte(c.host.APIVipConnectivity), &apiConnectivityResponse); err != nil {
+		return ValidationError, "Internal error - failed to parse agent API connectivity response"
 	}
-	if swag.BoolValue(c.cluster.UserManagedNetworking) {
-		return ValidationSuccess, "No API connectivity needed: User Managed Networking"
+
+	if apiConnectivityResponse.IsSuccess {
+		return ValidationSuccess, "Ignition is downloadable"
 	}
-	return ValidationSuccess, "Ignition is downloadable"
+
+	if apiConnectivityResponse.URL == "" {
+		// Missing URL means this is a response from an older agent version,
+		// without much information about what went wrong with the download -
+		// so return an undetailed error message
+		return ValidationFailure, "This host has failed to download the ignition file from the cluster, please ensure the host can reach the cluster"
+	}
+
+	return ValidationFailure, fmt.Sprintf(
+		"This host has failed to download the ignition file from %s with the following error: %s. "+
+			"Please ensure the host can reach this URL",
+		apiConnectivityResponse.URL, apiConnectivityResponse.DownloadError)
 }
 
 func (v *validator) belongsToL2MajorityGroup(c *validationContext, majorityGroups map[string][]strfmt.UUID) ValidationStatus {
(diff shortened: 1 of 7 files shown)


avishayt:
1. Please gofmt to fix lint
2. The main problem, IMO, is that we're checking "ignition downloadable" and reporting on "API connectivity."  The two aren't necessarily related.  They usually happen to be on the same IP but different port, but don't have to be (it's possible to override the ignition endpoint, which is done in the hypershift use case).  I was sure we renamed this, but I guess not enough.

internal/host/validator.go
```diff
@@ -845,13 +860,14 @@ func allImagesValid(imageStatuses common.ImageStatuses) bool {
...
 }
 
-/*
-   This is a pre-install validation that checks that the boot device was either not tested for sufficient disk speed
-   or the disk speed check has been successful.  Since disk speed test is performed after installation has started,
-   in order to have result for such test, the result has to be from a previous installation attempt.
-   Since all pre-install validations have to pass before starting installation, it is mandatory that in case installation
-   on the current boot device has not been attempted yet, this validation must pass.
-*/
+// This is a pre-install validation that checks that the boot device was either
```
omertuc, line 863: Recent gopls versions are opinionated about how this should be indented due to changes to the go doc format, I changed it to regular single-line comments to avoid this mismatch

omertuc:
> 2. The main problem, IMO, is that we're checking "ignition downloadable" and reporting on "API connectivity."  The two aren't necessarily related.  They usually happen to be on the same IP but different port, but don't have to be (it's possible to override the ignition endpoint, which is done in the hypershift use case).  I was sure we renamed this, but I guess not enough.

true, I'll try to re-work the phrasing

swagger.yaml
```diff
@@ -5482,16 +5482,22 @@ definitions:
...
     type: object
-    description: |
+    description: |-
       The response from the day-2 agent's attempt to download the worker ignition file from the API machine config server of the target cluster.
       Note - the name "API VIP connectivity" is old and misleading and is preserved for backwards compatibility.
     properties:
       is_success:
         type: boolean
-        description: Whether the agent was able to contact the API of the target cluster or not
+        description: Whether the agent was able to contact the API of the target cluster and download the ignition or not
```
avishayt, line 5491:
```suggestion
        description: Whether the agent was able to download the ignition or not
```

internal/host/validator.go
```diff
@@ -620,20 +620,35 @@ func (v *validator) isIgnitionDownloadable(c *validationContext) (ValidationStat
...
+
+	if apiConnectivityResponse.IsSuccess {
+		return ValidationSuccess, "Ignition is downloadable"
+	}
+
+	if apiConnectivityResponse.URL == "" {
+		// Missing URL means this is a response from an older agent version,
+		// without much information about what went wrong with the download -
+		// so return an undetailed error message
+		return ValidationFailure, "This host has failed to contact the cluster's API, please ensure the host can reach the API"
```
avishayt, line 645: We can remove this, now that we have agent upgrade, right?

internal/host/validator.go
```diff
@@ -620,20 +620,35 @@ func (v *validator) isIgnitionDownloadable(c *validationContext) (ValidationStat
...
 	}
-	return ValidationSuccess, "Ignition is downloadable"
+
+	return ValidationFailure, fmt.Sprintf(
+		"This host has failed to download the ignition file from the cluster's API at %s with the following error: %s. "+
+			"Please ensure the host can reach this URL",
```
avishayt, line 650:
```suggestion
		"This host has failed to download the ignition file from %s with the following error: %s. "+
			"Please ensure the host can reach this URL",
```

internal/host/validator.go
omertuc: I didn't write the agent part yet, so no

avishayt: I missed that this was new code, my bad

avishayt (approved):

avishayt (approved):

PR #4245, opened 2022-08-07
MGMT-9915: Enhancement for dual-stack VIPs

mkowalski:
This is an enhancement for dual-stack VIPs. It describes the proposed
design of the integration with the feature that has already a stable API
in the upstream OCP.

Closes: [MGMT-11525](https://issues.redhat.com//browse/MGMT-11525)

/cc @mhrivnak 
/cc @avishayt 
/cc @filanov 
/cc @ori-amizur 
/cc @carbonin

### DIFF
--- a/docs/enhancements/dual-stack-vips.md
+++ b/docs/enhancements/dual-stack-vips.md
@@ -1,0 +1,147 @@
+---
+title: dual-stack-vips
+authors:
+  - "@mkowalski"
+creation-date: 2022-08-07
+last-updated: 2022-10-31
+tracking-link:
+  - https://issues.redhat.com/browse/MGMT-9915
+see-also:
+  - https://github.com/openshift/enhancements/blob/master/enhancements/network/on-prem-dual-stack-vips.md
+---
+
+# Dual-stack VIPs
+
+## Summary
+
+Originally Assisted Installer supported only single-stack IPv4 or IPv6 deployments. When dual-stack support was added, we carried over the VIP configuration that only allowed for a single VIP to be specified.
+
+The initial rationale was that in dual-stack clusters every required communication can happen over IPv4 and only the workload running in the cluster may need IPv6 connectivity.
+
+Given this is no longer the case and that OCP for the baremetal platform supports dual-stack VIPs, this design is proposing the addition of a Cluster property allowing configuration of the aforementioned feature.
+
+## Motivation
+
+The current implementation of dual-stack support assumes that everything in the cluster should be reachable via IPv4 and that IPv6 is an addition for the workload running in the cluster. This assumption makes it impossible to interact with such a cluster using IPv6-only client, i.e. makes it impossible for IPv6-only Hub to manage a dual-stack Spoke.
+
+### Goals
+
+* Allow configuring both IPv4 and IPv6 VIP for dual-stack clusters.
+* Allow configuring only IPv4 VIPs for dual-stack clusters (it is supported already today and will not be removed with implementation of this enhancement).
+
+### Non-goals
+
+* Allow configuring more than single IPv4 and single IPv6 VIP. This is still not supported by the OCP. MetalLB is a better solution to creating arbitrary loadbalancers.
+* Allow configuring IPv6 as the primary networking stack. This is still not supported by the OCP. For dual-stack deployments the IPv4 stack is always to be the primary.
+* Allow DHCPv6 allocation of IPv6 VIPs. This is because DHCPv6 is not the model way of deploying IPv6 networks as well as that we intentionally decided not to offer IPv4 DHCP allocation for dual-stack clusters.
+* Allow VIP(s) to be placed outside of subnet(s) defined as machine network(s).
+* Change behaviour of VIPs in a single-stack scenario.
+
+## Proposal
+
+### User Stories
+
+* As an operator of the IPv6-only Hub cluster I want to deploy dual-stack Spoke cluster.
+
+* As an user of a dual stack OCP cluster I want to run IPv6-only applications that need access to the API and/or Ingress VIPs.
+
+### API Extensions
+
+New fields will have to be added to the Cluster object to persist the specified VIPs.
+
+### Implementation Details
+
+Our change can be simply described as "making a singular field plural". For this, we will follow the [Kubernetes API change guide](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#making-a-singular-field-plural).
+
+#### Phase 1
+
+`cluster-create-params`, `v2-cluster-update-params` and `cluster` will get new fields
+
+* `api_vips` of type `array`
+* `ingress_vips` of type `array`
+
+Respective AgentClusterInstall CR will also get those new fields.
+
+The database schema will be updated by creating new tables `api_vips` and `ingress_vips` with the following columns
+
+* `cluster_id`
+* `ip`
+
+Creation of the new tables in the database happens transparently thanks to the golang libraries in use. Copying current data to the new tables will be done using DB migration (as defined by the GORM library).
+
+Given that all the newly introduced fields are optional, there is no need to release a new version of the API.
+
+#### Phase 2
+
+A set of functions for cluster creation will provide a compatiblity between already existing fields (`api_vip` and `ingress_vip`) and new ones (`api_vips` and `ingress_vips`). The behaviour will be as follows
+
+* If only `api_vip` is provided, the `api_vips` will be automatically populated by a simple conversion from `string` to `[]string`. This guarantees no change in behaviour for users not aware of introduction of the new field.
+
+* If both `api_vip` and `api_vips` are provided, the value of `api_vips[0]` must match the value of `api_vip`.
+
+* If only `api_vips` are provided, the error will be raised. This is the case when plural field is specified and the singular is not. The rationale is that in an update it's impossible to tell the difference between an old client clearing the singular field via patch and a new client setting the plural field. The update semantics must not differ from create, therefore behaviour as described.
+
+For cluster update the following logic will apply
+
+* If `api_vip` is cleared and `api_vips` are not provided, we will clear both fields. This is a scenario when an old client (not aware of plural field) is used.
+
+* If `api_vips` is cleared and `api_vip` is not provided, the change will not apply. This is a scenario when an old client is used and it can't send fields it doesn't know about.
+
+* If `api_vip` is changed (but not cleared), `api_vips` will be forcefully set to the value of `api_vip` as a one-element list. This is a scenario when an old client is used to change the field it knows about.
+
+All the points described above will behave in an equivalent way for `ingress_vip` and `ingress_vips`. The logic above covers all CRUD operations that are reasonable for a specific combination in order to provide compatibility between various versions of clients.
+
+#### Phase 3
+
+After providing a required period of backwards-compatibility the `api_vip` and `ingress_vip` fields will be deprecated and removed. The value in the database will be nulled. Removal of the column in the database will follow.
+
+The new version of API will be released as we are removing an existing field.
+
+### UI Impact
+
+UI changes will be required in order to
+
+* display details of dual-stack clusters with dual-stack VIPs correctly
+* allow configuring dual-stack VIPs for dual-stack clusters
+
+For displaying the changes, the UI will need to display the value of `api_vips` and `ingress_vips` instead of the current `api_vip` and `ingress_vip`.
+
+For setting the values, the UI will need to include `api_vips` and `ingress_vips` in the cluster payload. The previous fields i.e. `api_vip` and `ingress_vip` must not be provided.
+
+Given that the implementation will provide a backwards compatibility, the rollout of the UI changes does not have to be synchronized with the rollout of the backend.
+
+As UI is a web client, it's going to transition to the new API faster than other clients. Users will only need to refresh the browser in order to get the latest client version.
+
+### Tooling impact
+
+The process outlined here ultimately leads to the deprecation of some API fields and introduction of the new ones. This has impact on all the tooling that contacts the Assisted Service API. With the relation to the mentioned before phases, the following applies
+
+* Phase 1
+    * no impact for tooling; no changes needed
+
+* Phase 2
+    * no changes needed for tools that only use single-stack VIPs
+    * changes needed for tools that want to offer dual-stack VIPs
+
+* Phase 3
+    * changes needed for all the tools
+
+### Risks and Mitigations
+
+The similar manouever of introducing `[]string` in place of `string` has already happened in the past when implementing the initial dual-stack support. An example of the field that got transformed is `machine_network_cidr` that became `machine_networks`.
+
+There is already an experience and established path, therefore the risk is considered minimal.
+
+Given that the core of the feature is implemented in the baremetal networking component of the OCP, the implementation in Assisted Installer is only responsible for providing the correct syntax to the install-config.
+
+Before moving from phase 2 to phase 3 we need to make sure that developers of the supported tools around Assisted Service are aware of this change. This includes but is not limited to:
+
+* aicli
+* Crucible
+* siteconfig generator
+
+### Test Plan
+
+The new fields will be covered by unit tests. The backwards compatiblity described above will be covered by unit tests.
+
+The whole of the installation will be covered by respective E2E tests using both test-infra as well as dev-scripts. This will provide a sufficient coverage for both REST API and Kube API flows.


docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,104 @@
...
+
+#### Phase 3
+
+After providing a required period of backwards-compatibility the `api_vip` and `ingress_vip` fields will be deprecated and removed. The value in the database will be nulled. Removal of the column in the database will follow.
+
+### Risks and Mitigations
+
+The similar manouever of introducing `[]string` in place of `string` has already happened in the past when implementing the initial dual-stack support. An example of the field that got transformed is `machine_network_cidr` that became `machine_networks`.
+
+There is already an experience and established path, therefore the risk is considered minimal.
```
avishayt, line 83: That's true, but we've still never broken a k8s API (the machine networks change happened before).  We'll have to create a new API version for phase 3, but that's fine.

mkowalski: You are right. To be honest what I meant here was the fact that in the 1st dual-stack iteration we *broke* API for real by changing the type of an existing field. That's why my focus here was to explain that we will not repeat this mistake.

For the K8s API you are right, we already had `[]string` fields back then. Here we need to introduce a new field, release a new version (keeping both fields aside) and then in the future release yet newer version with the old field removed

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,104 @@
...
+Respective AgentClusterInstall CR will also get those new fields.
+
+The database schema will be updated by creating new tables `api_vips` and `ingress_vips` with the following columns
+
+* `cluster_id`
+* `vip`
+
+#### Phase 2
+
+A set of functions for cluster creation and cluster update will provide a compatiblity between already existing fields (`api_vip` and `ingress_vip`) and new ones (`api_vips` and `ingress_vips`). The behaviour will be as follows
```
mhrivnak, line 65: What about just following the [k8s API change guide](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#making-a-singular-field-plural)?

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,104 @@
...
+
+* If only `api_vips` are provided, the `api_vip` will be populated by taking the first value from `[]string`. This guarantees that older versions of the UI do not break when interacting with the new field.
+
+* If both `api_vip` and `api_vips` are provided, the value of `api_vips` will take precedence. This is an undesired behaviour indicating use of the wrong syntax or client but for the completeness the behaviour has to be defined.
+
+All the points described above will behave in an equivalent way for `ingress_vip` and `ingress_vips`.
+
+#### Phase 3
+
+After providing a required period of backwards-compatibility the `api_vip` and `ingress_vip` fields will be deprecated and removed. The value in the database will be nulled. Removal of the column in the database will follow.
```
mhrivnak, line 77: ... and a new version of the API will be released?

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,104 @@
...
+### UI Impact
+
+UI changes will be required in order to
+
+* display details of dual-stack clusters with dual-stack VIPs correctly
+* allow configuring dual-stack VIPs for dual-stack clusters
+
+For displaying the changes, the UI will need to display the value of `api_vips` and `ingress_vips` instead of the current `api_vip` and `ingress_vip`.
+
+For setting the values, the UI will need to include `api_vips` and `ingress_vips` in the cluster payload. The previous fields i.e. `api_vip` and `ingress_vip` must not be provided.
```
mhrivnak, line 96: Note that this differs from the [k8s API change guide](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#making-a-singular-field-plural), which says a new client must supply both values.

mhrivnak (changes requested):

docs/enhancements/dual-stack-vips.md
mkowalski: Agree, that’s a good catch. Seems like what we did in the past was not in line with k8s guidelines. I will of course fix it here so we are following them now

docs/enhancements/dual-stack-vips.md
filanov: if i remember correctly for cloud we can have the same version, for kube-api we will probably need to have a new version 
@avishayt what do you think?

filanov (commented):
overall LGTM from me, lets wait for another one

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,104 @@
...
+
+* Allow configuring more than single IPv4 and single IPv6 VIP. This is still not supported by the OCP.
+
+## Proposal
+
+### User Stories
+
+* As an operator of the IPv6-only Hub cluster I want to deploy dual-stack Spoke cluster.
+
+* As a deployer of a dual stack OCP cluster I want to run IPv6-only applications that need access to the API and Ingress VIPs.
```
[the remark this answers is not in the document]
mkowalski: Fixed

docs/enhancements/dual-stack-vips.md
mkowalski: Done

docs/enhancements/dual-stack-vips.md
mkowalski: Fixed, I have rewritten the above to follow the API change guide

docs/enhancements/dual-stack-vips.md
mkowalski: At least according to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md, K8s API should get a new version

docs/enhancements/dual-stack-vips.md
lranjbar: That sounds like the path to have the least amount of API breakage. Before when the API was changed it was broken but we had the excuse of not being a released product. But since AI has gone GA now, you have to take more steps to change an API like this. Sounds like a good plan.

lranjbar:
@mkowalski Have you thought about how this will change the functionality in the Rest API and k8s API?  Will there be a divergence in the functionality? Will this only be available in the k8s API at first, or only in the Rest API at first? When I read this it feels like this is a primarily a change for the k8s API but perhaps I'm misunderstanding the goal of the enchancement.

mkowalski:
@lranjbar the general aim of this enhancement is to cover both REST and K8s API; this is in fact outlined as one of the goals of https://issues.redhat.com//browse/MGMT-9915 which this enhancement is the first part of

I agree, overall when reading the enhancement it may feel like it focuses only on K8s API, but that's only not to lose the readers' attention. We aim for feature parity between REST and K8s APIs in this regards. Which API is going to get it first depends kind of on the person implementing that. The epic https://issues.redhat.com//browse/MGMT-9915 will not be closed until both APIs have this feature. Some people are comfortable with writing K8s API first and then just filling the gaps for REST, whereas other people prefer to focus on REST and then fill the gaps for K8s.

lranjbar:
Thanks for the background Mat :)

rwsu:
Looks good.

ori-amizur:
Couple of questions:
1. What will be the effect of this change concerning vip-dhcp-allocation? 
2. Is there a plan for none platform (dual LB addresses) ?
3. Currently we have auto-assign functionality for machine network.  Is this functionality going to be updated?

mkowalski:
> 1. What will be the effect of this change concerning vip-dhcp-allocation?

There is no auto-allocation allowed for dual-stack today and this enhancement is not changing anything in this respect

> 2. Is there a plan for none platform (dual LB addresses) ?

AFAIK it makes no difference whether we use platform None or platform Baremetal for dual-stack. Saying in AI language - dual-stack for SNO and multi-node has feature-parity and this enhancement is not changing anything in this respect. Dual-stack VIPs are expected to work for SNO and multi-node as long as upstream OCP allows it

> 3. Currently we have auto-assign functionality for machine networks.  Is this functionality going to be updated?

We do not have auto-assign of machine networks for dual-stack and this enhancement is not changing anything in this respect

ori-amizur:
> There is no auto-allocation allowed for dual-stack today and this enhancement is not changing anything in this respect
The data structure is going to change and vip-dhcp-allocation relies on it.
Also, is this going to change in OCP? If yes, it might effect lease renewal.

> AFAIK it makes no difference whether we use platform None or platform Baremetal for dual-stack. Saying in AI language - >dual-stack for SNO and multi-node has feature-parity and this enhancement is not changing anything in this respect. Dual->stack VIPs are expected to work for SNO and multi-node as long as upstream OCP allows it
We probably need to have the DNS populated with both stacks. 

> We do not have auto-assign of machine networks for dual-stack and this enhancement is not changing anything in this respect
Do we always expect them to be provided?

mkowalski:
> > There is no auto-allocation allowed for dual-stack today and this enhancement is not changing anything in this respect
> The data structure is going to change and vip-dhcp-allocation relies on it.

VIP auto-allocation for single-stack should rely on the fact that there can be only one item in the list of VIPs. For this reason the behaviour is not changing, it will only use [0] element instead of vanilla variable

> Also, is this going to change in OCP? If yes, it might effect lease renewal.

VIP auto-allocation is a feature of Assisted Installer. It does not exist in vanilla OCP

> > AFAIK it makes no difference whether we use platform None or platform Baremetal for dual-stack. Saying in AI language - dual-stack for SNO and multi-node has feature-parity and this enhancement is not changing anything in this respect. Dual-stack VIPs are expected to work for SNO and multi-node as long as upstream OCP allows it
> 
> We probably need to have the DNS populated with both stacks.

ACK

> 
> > We do not have auto-assign of machine networks for dual-stack and this enhancement is not changing anything in this respect
> 
> Do we always expect them to be provided?

Not only “expect” but we require them to be provided. It is part of validation already today and will not disappear

ori-amizur:
Please note that the VIPs allocated by DHCP are passed by ignition  [here](https://github.com/openshift/assisted-service/blob/b22b165f9b273194f8c15b70b155b48713a50f32/internal/ignition/ignition.go#L327
)
It is OCP responsibility to continue renewal.

ori-amizur:
> > Please note that the VIPs allocated by DHCP are passed by ignition [here](https://github.com/openshift/assisted-service/blob/b22b165f9b273194f8c15b70b155b48713a50f32/internal/ignition/ignition.go#L327) It is OCP responsibility to continue renewal.
> 
> Yes. From core OCP perspective there is no difference between VIP that has been selected by the user or auto-allocated by some service (in this case, by Assisted). This is not changing, OCP will receive two VIPs instead of a single one.

OCP does not renew VIPs that were not allocated by DHCP

mkowalski:
> > > Please note that the VIPs allocated by DHCP are passed by ignition [here](https://github.com/openshift/assisted-service/blob/b22b165f9b273194f8c15b70b155b48713a50f32/internal/ignition/ignition.go#L327) It is OCP responsibility to continue renewal.
> > 
> > 
> > Yes. From core OCP perspective there is no difference between VIP that has been selected by the user or auto-allocated by some service (in this case, by Assisted). This is not changing, OCP will receive two VIPs instead of a single one.
> 
> OCP does not renew VIPs that were not allocated by DHCP. So there is a clear difference.

Please provide upstream doc for this? I don't see any field in install-config.yaml that would be differentiating between user-selected or auto-allocated VIP?

If this is something that we do inside e.g. ignition override, then it is our feature. Upstream OCP features in this context are features that are available in the core OCP irrespective from the installation method (so available in IPI, UPI and so on)

ori-amizur:
> > > > Please note that the VIPs allocated by DHCP are passed by ignition [here](https://github.com/openshift/assisted-service/blob/b22b165f9b273194f8c15b70b155b48713a50f32/internal/ignition/ignition.go#L327) It is OCP responsibility to continue renewal.
> > > 
> > > 
> > > Yes. From core OCP perspective there is no difference between VIP that has been selected by the user or auto-allocated by some service (in this case, by Assisted). This is not changing, OCP will receive two VIPs instead of a single one.
> > 
> > 
> > OCP does not renew VIPs that were not allocated by DHCP. So there is a clear difference.
> 
> Please provide upstream doc for this? I don't see any field in install-config.yaml that would be differentiating between user-selected or auto-allocated VIP?
> 
> If this is something that we do inside e.g. ignition override, then it is our feature. Upstream OCP features in this context are features that are available in the core OCP irrespective from the installation method (so available in IPI, UPI and so on)

[This](https://github.com/openshift/assisted-service/blob/b22b165f9b273194f8c15b70b155b48713a50f32/internal/network/dhcp_param_file.go#L46) is the place where we decide if to add the file in ignition or not.
[Here](https://github.com/openshift/assisted-service/blob/b22b165f9b273194f8c15b70b155b48713a50f32/internal/ignition/ignition.go#L945
) is the location where we decide decide if to update the ignition for DHCP.  It will be done only for DHCP VIP allocation.
Adding these files affects OCP .  It causes it to perform DHCP renewal on the VIPs.

mkowalski:
In the linked file we have 

```
setFileInIgnition(config, "/etc/keepalived/unsupported-monitor.conf", g.encodedDhcpFileContents, false, 0o644, false)
```

so we inject a custom file to the configuration of keepalived, is that right? If yes, then this is a feature of Assisted Installer and not a feature of OpenShift

ori-amizur:
> In the linked file we have
> 
> ```
> setFileInIgnition(config, "/etc/keepalived/unsupported-monitor.conf", g.encodedDhcpFileContents, false, 0o644, false)
> ```
> 
> so we inject a custom file to the configuration of keepalived, is that right? If yes, then this is a feature of Assisted Installer and not a feature of OpenShift

OK.  I can accept it.
Just keep in mind that assisted is the only one that uses this custom file (and this functionality).

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Risks and Mitigations
+
+The similar manouever of introducing `[]string` in place of `string` has already happened in the past when implementing the initial dual-stack support. An example of the field that got transformed is `machine_network_cidr` that became `machine_networks`.
+
+There is already an experience and established path, therefore the risk is considered minimal.
+
+Given that the core of the feature is implemented in the baremetal networking component of the OCP, the implementation in Assisted Installer is only responsible for providing the correct syntax to the install-config.
+
+### UI Impact
```
mkowalski, line 97: @nirfarkas @jkilzi: please review and comment before any real-time kick-off meeting happens

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,114 @@
...
+* display details of dual-stack clusters with dual-stack VIPs correctly
+* allow configuring dual-stack VIPs for dual-stack clusters
+
+For displaying the changes, the UI will need to display the value of `api_vips` and `ingress_vips` instead of the current `api_vip` and `ingress_vip`.
+
+For setting the values, the UI will need to include `api_vips` and `ingress_vips` in the cluster payload. The previous fields i.e. `api_vip` and `ingress_vip` must not be provided.
+
+Given that the implementation will provide a backwards compatibility, the rollout of the UI changes does not have to be synchronized with the rollout of the backend.
+
+### Test Plan
```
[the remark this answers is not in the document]
mkowalski: @lalon4 this is a ping, I'm planning to proceed with the agreed implementation and I will be taking no comments as an agreement

docs/enhancements/dual-stack-vips.md
mkowalski: @nirfarkas @jkilzi this is a ping, I'm planning to proceed with the agreed implementation and I will be taking no comments as an agreement

docs/enhancements/dual-stack-vips.md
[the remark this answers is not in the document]
mkowalski: @lalon4 this is a ping, I'm planning to proceed with the agreed implementation and I will be taking no comments as an agreement

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,114 @@
...
+
+* If `api_vips` is cleared and `api_vip` is not provided, the change will not apply. This is a scenario when an old client is used and it can't send fields it doesn't know about.
+
+* If `api_vip` is changed (but not cleared), `api_vips` will be forcefully set to the value of `api_vip` as a one-element list. This is a scenario when an old client is used to change the field it knows about.
+
+All the points described above will behave in an equivalent way for `ingress_vip` and `ingress_vips`.
+
+#### Phase 3
+
+After providing a required period of backwards-compatibility the `api_vip` and `ingress_vip` fields will be deprecated and removed. The value in the database will be nulled. Removal of the column in the database will follow.
```
gamli75, line 87: It will impact not only UI & test-infra but also external tools such as: aicli

mkowalski: That's correct, I will write it explicitly. Also Crucible and siteconfig generator will be affected (that's all interfaces that I know about)

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,138 @@
...
+
+Given that all the newly introduced fields are optional, there is no need to release a new version of the API.
+
+#### Phase 2
+
+A set of functions for cluster creation will provide a compatiblity between already existing fields (`api_vip` and `ingress_vip`) and new ones (`api_vips` and `ingress_vips`). The behaviour will be as follows
+
+* If only `api_vip` is provided, the `api_vips` will be automatically populated by a simple conversion from `string` to `[]string`. This guarantees no change in behaviour for users not aware of introduction of the new field.
+
+* If both `api_vip` and `api_vips` are provided, the value of `api_vip[0]` must match the value of `api_vip`.
```
jkilzi, line 73:
```suggestion
* If both `api_vip` and `api_vips` are provided, the value of `api_vips[0]` must match the value of `api_vip`.
```

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,138 @@
...
+
+* `cluster_id`
+* `ip`
+
+Given that all the newly introduced fields are optional, there is no need to release a new version of the API.
+
+#### Phase 2
+
+A set of functions for cluster creation will provide a compatiblity between already existing fields (`api_vip` and `ingress_vip`) and new ones (`api_vips` and `ingress_vips`). The behaviour will be as follows
+
```
jkilzi, line 70: I think you forgot to mention this part as per [making-a-singular-field-plural](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#making-a-singular-field-plural)
> Upon any read operation:  
> * If plural is not populated, API logic must populate plural as a one-element list, with plural[0] set to the singular value.

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,138 @@
...
+The database schema will be updated by creating new tables `api_vips` and `ingress_vips` with the following columns
+
+* `cluster_id`
+* `ip`
+
+Given that all the newly introduced fields are optional, there is no need to release a new version of the API.
+
+#### Phase 2
+
```
jkilzi, line 68: I think you forgot to mention this part as per [making-a-singular-field-plural](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#making-a-singular-field-plural)
> Upon any read operation:  
> * If plural is not populated, API logic must populate plural as a one-element list, with plural[0] set to the singular value.

docs/enhancements/dual-stack-vips.md
jkilzi: Questions:
1. How many entries per field (api_vips/ingress_vips) are users expected to enter?
2. Is it valid to pass IPv4 addresses intertwined with IPv6?
3. Is the order of the entries in the lists important?
4. WLG. Can users specify only IPv4 addresses in one list but mixed IP versions in the other?
5. Is there any relationship between the api_vips and ingress_vips, meaning, are the addresses required to be part of the same subnet when entered manually (no DHCP)?
6. Should we expect any special case for when VIP DHCP allocation is enabled/disabled?

docs/enhancements/dual-stack-vips.md
mkowalski: The first bullet mentions it - „If only `api_vip` is provided, the `api_vips` will be automatically populated”

docs/enhancements/dual-stack-vips.md
mkowalski: Answered 1-6 in goals and non-goals sections. Let me know if any of questions still needs some further clarification

docs/enhancements/dual-stack-vips.md
mkowalski: Of course, thanks for finding this out!

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,142 @@
...
+#### Phase 1
+
+`cluster-create-params`, `v2-cluster-update-params` and `cluster` will get new fields
+
+* `api_vips` of type `array`
+* `ingress_vips` of type `array`
+
+Respective AgentClusterInstall CR will also get those new fields.
+
+The database schema will be updated by creating new tables `api_vips` and `ingress_vips` with the following columns
```
nmagnezi, line 64: Perhaps the document should mention that this will be done via DB migration.

nmagnezi (commented):
I added just a few minor comments and questions.
Great job with this enhancement doc @mkowalski!

docs/enhancements/dual-stack-vips.md
jkilzi: Oh I see, since you were talking there about *creation* and *updates* I thought the section about *read* operations wasn't explicitly discussed.

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,142 @@
...
+
+After providing a required period of backwards-compatibility the `api_vip` and `ingress_vip` fields will be deprecated and removed. The value in the database will be nulled. Removal of the column in the database will follow.
+
+The new version of API will be released as we are removing an existing field.
+
+### UI Impact
+
+UI changes will be required in order to
+
+* display details of dual-stack clusters with dual-stack VIPs correctly
```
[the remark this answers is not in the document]
jkilzi: On that regard, the proposed changes are going to be backward-compatible. Old clients keep working as usual with the singular fields until we announce they'll be deprecated and new clients will use both fields single and plural whenever they need to issue update or create calls, that's the TL;DR of this proposal.  
BTW, keep in mind that the web UI, given it's a web client, is going to transition to the new API faster, users will only need to refresh the browser in order to get the latest client version, once it is implemented)

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,142 @@
...
+* allow configuring dual-stack VIPs for dual-stack clusters
+
+For displaying the changes, the UI will need to display the value of `api_vips` and `ingress_vips` instead of the current `api_vip` and `ingress_vip`.
```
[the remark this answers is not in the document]
jkilzi: No it isn't necessary, IMO, the UI should now allow users to enter multiple values instead of just one. The API fields transitioning from singular to plural is just an implementation detail.

docs/enhancements/dual-stack-vips.md
jkilzi: We still do need the following clarifications:
1. Is this enhancement document applicable if users opt to not use dual-stack?
1. Do users still need to enter multiple IPv4 addresses if they don't use dual-stack?
1. Goal: "Allow configuring only IPv4 VIPs for dual-stack clusters.". Isn't this something we already allow?
1. What kind of constrains should be applied to these fields?
1. Could you please provide some examples of valid or invalid user input?

jkilzi: > How many entries per field (api_vips/ingress_vips) are users expected to enter?

This question is not addressed in the Goals section. Right now my understanding is that the UI should allow users to enter one IPv4 and one IPv6 per VIP when they choose to use dual-stack, is my assumption correct?

mkowalski: Questions 1-2: This enhancement is only about dual-stack users. For people doing single-stack nothing changes. Hence, we don't mention them because we focus here only on people who do dual-stack.

Question 3: Yes you are right, we allow it now. It's in the goals explicitly so that it's clear that we will still allow it

Question 4: Explained in the doc. It's only about the belonging to the machine networks and about the order

>This question is not addressed in the Goals section. Right now my understanding is that the UI should allow users to enter one IPv4 and one IPv6 per VIP when they choose to use dual-stack, is my assumption correct?

It is addressed in the doc. We explain that enabling more than 1 VIP per IP stack is in the realm of MetalLB hence is not the goal of this enhancement here. TLDR is that you can provide 1 IPv4 and 1 IPv6, not more.

docs/enhancements/dual-stack-vips.md
```diff
@@ -0,0 +1,142 @@
...
+
+#### Phase 2
+
+A set of functions for cluster creation will provide a compatiblity between already existing fields (`api_vip` and `ingress_vip`) and new ones (`api_vips` and `ingress_vips`). The behaviour will be as follows
+
+* If only `api_vip` is provided, the `api_vips` will be automatically populated by a simple conversion from `string` to `[]string`. This guarantees no change in behaviour for users not aware of introduction of the new field.
+
+* If both `api_vip` and `api_vips` are provided, the value of `api_vips[0]` must match the value of `api_vip`.
+
+* If only `api_vips` are provided, the error will be raised. This is the case when plural field is specified and the singular is not. The rationale is that in an update it's impossible to tell the difference between an old client clearing the singular field via patch and a new client setting the plural field. The update semantics must not differ from create, therefore behaviour as described.
```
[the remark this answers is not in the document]
mkowalski: >This is very confusing and worth a discussion.

This is how k8s upstream defines it. I'm not sure if we are in authority to try changing it. But I do agree that the error needs to be very clear to the end user

docs/enhancements/dual-stack-vips.md
mkowalski: This is a good comment, I added a sentence explaining explicitly that this logic is meant to cover all of CRUD

docs/enhancements/dual-stack-vips.md
mkowalski: Of course, just added it

docs/enhancements/dual-stack-vips.md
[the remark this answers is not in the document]
mkowalski: Because of all the magic dance we are doing in Phase 2, every tool we have will be backwards-compatible. If we had the possibility to skip that, we would move from Phase 1 directly to Phase 3

mkowalski:
/approve
/lgtm

The implementation is finished, there is nothing to discuss here anymore

nmagnezi (approved):
/lgtm

mkowalski:
/override ci/prow/e2e-agent-compact

No seriously this is just and enhancement, just a piece of markdown

PR #4247, opened 2022-08-07
NO-ISSUE: CI to be more graceful with ZTP log

mkowalski:
This PR changes the way we collect ZTP logs to be more graceful. This is
so that we try to lazily collect as much as possible instead of failing
in the first unsuccessful "oc get".

/cc @osherdp

### DIFF
--- a/deploy/operator/gather.sh
+++ b/deploy/operator/gather.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -xeo pipefail
+set -x
 
 __dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source ${__dir}/common.sh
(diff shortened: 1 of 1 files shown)


deploy/operator/gather.sh
```diff
@@ -17,9 +17,9 @@ function gather_hive_data() {
 }
 
 function gather_olm_data() {
-  oc get catalogsource "${ASSISTED_SERVICE_OPERATOR_CATALOG}" -n openshift-marketplace -o yaml > ${LOGS_DEST}/${ASSISTED_SERVICE_OPERATOR_CATALOG}.log
-  oc get subscription assisted-service-operator -n "${ASSISTED_NAMESPACE}" -o yaml > ${LOGS_DEST}/assisted-service-operator-subscription.log
-  oc get installplan -n "${ASSISTED_NAMESPACE}" -o yaml  > ${LOGS_DEST}/oc_install_plan.log
+  oc get catalogsource "${ASSISTED_SERVICE_OPERATOR_CATALOG}" -n openshift-marketplace -o yaml > ${LOGS_DEST}/${ASSISTED_SERVICE_OPERATOR_CATALOG}.log || true
```
osherdp, line 20: maybe we should apply it to the 3rd line (``set -xeo pipefail``) and just make the whole script optional in its run?
because I think there's no justification to make some lines optional and the others not

osherdp (changes requested):

deploy/operator/gather.sh
mkowalski: I agree very much. I find it generally confusing when I see Prow telling me "e2e-ai-operator-ztp-disconnected-assisted-baremetal-operator-gather failed" because I'd expect -gather to be just an opportunistic step

mkowalski: How about having just `set -x` for this particular script? Anything against solving the problem like that?

osherdp: yes sounds good

osherdp: on the job level it is best-effort btw: https://github.com/openshift/release/blob/eec4d78a6279dcc32cd3b622783297b59920eb59/ci-operator/step-registry/assisted/baremetal/operator/ztp/assisted-baremetal-operator-ztp-workflow.yaml#L4

osherdp (approved):

osherdp:
and as always, hold it until jobs are passing. rehearse jobs are waiting for no one :)
/hold

osherdp:
Be aware of: https://github.com/openshift/assisted-service/pull/4266

mkowalski:
Ok, rebased, let's re-review @osherdp ?

mkowalski:
/unhold

We are not in o/release, there are no rehearses

osherdp (approved):

osherdp:
/override ci/prow/edge-e2e-ai-operator-ztp-converged
seems unrelated (and it's a required job that will just keep retesting)

PR #4248, opened 2022-08-08
MGMT-8890: Extend release image struct to store all archs

mkowalski:
This PR extends the ReleaseImage with a new `[]string` field that stores
all the architectures referenced by the release image. This is useful
for working with multiarch images when `releaseImage.CPUArchitecture`
field has only "multiarch" as a value but it carries no knowledge about
specific architectures of such an image.

With this change a multiarch release image will be stored as e.g.

```
CPUArchitecture: "multiarch"
CPUArchitectures: ["x86_64", "ppc64le", "s390x", "arm64"]
```

In order to allow for further simplifications in the future and
potentially a drop of singular `CPUArchitecture` field, the value of
plural field for a single-arch image will be a one-element list, e.g.

```
CPUArchitecture: "x86_64"
CPUArchitectures: ["x86_64"]
```

Closes: [MGMT-11494](https://issues.redhat.com//browse/MGMT-11494)

/cc @carbonin 
/cc @ori-amizur 
/cc @danielerez


cmd/agentbasedinstaller/register.go
```diff
@@ -192,7 +192,7 @@ func getReleaseVersionAndCpuArch(log *log.Logger, releaseImage string, releaseMi
 		return "", "", archError
 	}
 
-	return version, cpuArch, nil
+	return version, cpuArch[0], nil
```
paul-maidment, line 195: is it intentional that we always return the 1st cpuArch here?
Is there a concept of a "default" index or some way for the user to pick a different index?

mkowalski: @zaneb that's your part of the code that does not have a coverage anywhere in `cmd/agentbasedinstaller/*test.go`; can you have a look please? The semantics of `GetReleaseArchitecture()` changes to return `[]string` instead of `string`. I tried to accommodate your function as much as I understood it, but given there are no other handlers apart for error returned by `GetReleaseArchitecture()`, we end up with such a simple change

swagger.yaml
```diff
@@ -6221,6 +6221,11 @@ definitions:
         default: "x86_64"
         description: The CPU architecture of the image (x86_64/arm64/etc).
         x-go-custom-tag: gorm:"default:'x86_64'"
+      cpu_architectures:
```
danielerez, line 6224: As this is an API change, do we want to mark as deprecated the old cpu_architecture? Or we'll keep use both for now?

internal/versions/versions.go
```diff
@@ -307,8 +307,10 @@ func (h *handler) GetLatestOsImage(cpuArchitecture string) (*models.OsImage, err
 	return latest, nil
 }
 
-func (h *handler) AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion, cpuArchitecture string) (*models.ReleaseImage, error) {
+func (h *handler) AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion, cpuArchitecture string, cpuArchitectures []string) (*models.ReleaseImage, error) {
```
danielerez, line 310: Are we still using cpuArchitecture?

models/release_image.go
```diff
@@ -24,6 +24,9 @@ type ReleaseImage struct {
 	// Required: true
 	CPUArchitecture *string `json:"cpu_architecture" gorm:"default:'x86_64'"`
 
+	// List of CPU architectures provided by the image.
+	CPUArchitectures []string `json:"cpu_architectures"`
```
danielerez, line 28: Should we update default_release_images.json file with cpu_architectures as well? Is it the next step to have cpu_architectures required and remove/deprecate cpu_architecture?

cmd/agentbasedinstaller/register.go
danielerez, line 195: Should ensure cpuArch is not empty just to be safe?

mkowalski: Yeah, just look above at the discussion with Paul. It's not really our piece of code and it has no tests, so I agree there should be _some_ handler but I could as well write
```
if cpuArch == nil {
  panic()
}
```

swagger.yaml
mkowalski: We keep using both for now because we can't drop a field anyway... How do we mark field as deprecated? I'm happy to do so but the last time I checked, the swagger version we use did not support it

internal/versions/versions.go
mkowalski: Yes we still are, because there may be users of `default_release_images.json` not aware of the new field - they will keep use `cpu_architecture` and not provide `cpu_architectures` at all

models/release_image.go
mkowalski: That's a very good suggestion, just added it.

By the way, given that the code detects it automatically, I wonder whether we even need CPU arch(s) and release version in this file at all... Any thoughts?

mkowalski: Maybe @osherdp can suggest?

The current entry looks like
```
    {
        "openshift_version": "4.11",
        "cpu_architecture": "arm64",
        "cpu_architectures": ["arm64"],
        "url": "quay.io/openshift-release-dev/ocp-release:4.11.0-rc.7-aarch64",
        "version": "4.11.0-rc.7",
        "support_level": "beta"
    }
```

but given the implementation that handles ClusterImageSet containing only the URL (so we have capability to detect CPU arch and version), it looks to me like some of those here are redundant

swagger.yaml
danielerez: Yeah, I usually just add a comment somewhere to indicate it's deprecated. Maybe simply in the old property description.

internal/versions/versions.go
danielerez: Oh, AddReleaseImage is an internal API so should be fine to change it. it'll probably look clearer if we have only cpu_architectures in the signature, and the caller (clusterdeployments_controller) would just provide an array with a single arch.
Anyway, we should check in line 315 if cpu_architectures is provided. I mean, no need to use oc for fetching arch in case cpu_architectures is not empty.

models/release_image.go
mkowalski: Forget, after analyzing stuff we cannot remove anything and we must keep both `arch` and `archs`. Sorry for the noise

internal/versions/versions.go
```diff
@@ -307,32 +307,41 @@ func (h *handler) GetLatestOsImage(cpuArchitecture string) (*models.OsImage, err
...
-func (h *handler) AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion, cpuArchitecture string) (*models.ReleaseImage, error) {
+func (h *handler) AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion string, cpuArchitectures []string) (*models.ReleaseImage, error) {
 	var err error
-	// If release version or cpu architecture are not specified, use oc to fetch values
-	if ocpReleaseVersion == "" || cpuArchitecture == "" {
+	var cpuArchitecture string
+	var osImage *models.OsImage
+
+	// If release version or cpu architectures are not specified, use oc to fetch values
+	if ocpReleaseVersion == "" || cpuArchitectures == nil {
```
danielerez, line 316: cpuArchitectures shouldn't be an empty array, but maybe worth checking as well just in case.

mkowalski: That's a good catch, thanks!

cmd/agentbasedinstaller/register.go
```diff
@@ -187,12 +187,17 @@ func getReleaseVersionAndCpuArch(log *log.Logger, releaseImage string, releaseMi
...
-	cpuArch, archError := releaseHandler.GetReleaseArchitecture(log, releaseImage, releaseMirror, pullSecret)
+	cpuArchs, archError := releaseHandler.GetReleaseArchitecture(log, releaseImage, releaseMirror, pullSecret)
 	if archError != nil {
 		return "", "", archError
 	}
 
-	return version, cpuArch, nil
+	// This is a safety compatibility handler. GetReleaseArchitecture() should never return nil without an error
+	// but given the caller of this function here does not check that, we want to explicitly handle this scenario.
+	if cpuArchs == nil {
```
danielerez, line 197: Nice. Good enough:) Also an empty cpuArchs just in case..

swagger.yaml
```diff
@@ -6221,6 +6221,11 @@ definitions:
         default: "x86_64"
         description: The CPU architecture of the image (x86_64/arm64/etc).
```
danielerez, line 6222: Maybe add here 'deprecated' mark, just as a note to remove it later on.

mkowalski: Right, added `(DEPRECATED)` as a prefix

cmd/agentbasedinstaller/register.go
mkowalski: Yes, fixed

data/default_okd_release_images.json
```diff
@@ -2,6 +2,7 @@
     {
         "openshift_version": "4.11",
         "cpu_architecture": "x86_64",
+        "cpu_architectures": ["x86_64"],
```
mkowalski, line 5: @vrutkovs this change would benefit from your approval

data/default_release_images.json
```diff
@@ -2,46 +2,53 @@
     {
         "openshift_version": "4.7",
         "cpu_architecture": "x86_64",
+        "cpu_architectures": ["x86_64"],
```
mkowalski, line 5: @osherdp this change would benefit from your approval

swagger.yaml
```diff
@@ -6210,6 +6210,7 @@ definitions:
     required:
       - openshift_version
       - cpu_architecture
+      - cpu_architectures
```
mkowalski, line 6213: @danielerez I wonder about this part, what do you think?

internal/oc/release.go
```diff
@@ -171,18 +168,36 @@ func (r *release) GetReleaseArchitecture(log logrus.FieldLogger, releaseImage st
...
 		}
-		_, _, _, err2 = jsonparser.Get([]byte(skopeoImageRaw), "manifests")
+
+		var multiarchContent []string
+		_, err2 = jsonparser.ArrayEach([]byte(skopeoImageRaw), func(value []byte, dataType jsonparser.ValueType, offset int, err error) {
+			res, _ := jsonparser.GetString(value, "platform", "architecture")
+
+			// Convert architecture naming to supported values
+			if res == "amd64" {
+				res = common.DefaultCPUArchitecture
```
carbonin, line 180: I hope we never change the default to arm :sweat_smile:

internal/oc/release_test.go
```diff
@@ -259,29 +263,88 @@ var _ = Describe("oc", func() {
...
+				args := splitStringToInterfacesArray(command)
+				args2 := splitStringToInterfacesArray(command2)
+				mockExecuter.EXPECT().Execute(args[0], args[1:]...).Return("", "the image is a manifest list", 1).Times(1)
+				mockExecuter.EXPECT().Execute(args2[0], args2[1:]...).Return(test_skopeo_multiarch_image_output, "", 0).Times(1)
 
-			arch, err := oc.GetReleaseArchitecture(log, releaseImage, "", pullSecret)
-			Expect(arch).Should(Equal(common.MultiCPUArchitecture))
-			Expect(err).ShouldNot(HaveOccurred())
+				arch, err := oc.GetReleaseArchitecture(log, releaseImage, "", pullSecret)
+				Expect(arch).Should(Equal([]string{"x86_64", "ppc64le", "s390x", "arm64"}))
```
carbonin, line 300: Does this check validate the order? If so, is there a matcher that doesn't?

internal/versions/versions.go
```diff
@@ -307,32 +307,41 @@ func (h *handler) GetLatestOsImage(cpuArchitecture string) (*models.OsImage, err
...
 
-func (h *handler) AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion, cpuArchitecture string) (*models.ReleaseImage, error) {
+func (h *handler) AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion string, cpuArchitectures []string) (*models.ReleaseImage, error) {
 	var err error
-	// If release version or cpu architecture are not specified, use oc to fetch values
-	if ocpReleaseVersion == "" || cpuArchitecture == "" {
+	var cpuArchitecture string
+	var osImage *models.OsImage
+
+	// If release version or cpu architectures are not specified, use oc to fetch values
```
carbonin, line 315: If the arch is specified, what happens if the arch list is wrong?

There are a few cases, but the most obvious is someone provides an arch that actually isn't present in the multiarch image.
I also see there's some handling around `amd64` vs `x86_64`, do we care about handling that at a later stage if we aren't detecting it? For example, can a user specify either in the release image arch?

I guess the same conversation could be had around the single arch image though ... :man_shrugging:

swagger.yaml
carbonin: I think adding a new required field is a breaking API change.

mkowalski: ACK, removing the requiredness right now

internal/oc/release.go
mkowalski: 😅

internal/versions/versions.go
mkowalski: Right, so in the current logic with single-arch if you lied and provided a wrong valued we just let it be. I copy the same behaviour. But I think it's important to note here that `AddReleaseImage()` is a function used only internally and never interfaces the user. User talks to us via `ClusterImageSet` in which my expectation (correct me if I'm wrong) is that you only give us URL of the release image and we do all the magic with detection for you.

internal/oc/release_test.go
mkowalski: Yes it does, but it runs against a manually crafted `test_skopeo_multiarch_image_output` which has the same order. Because of that, as long as we don't change our implementation of unmarshalling that would stop preserving the order, this is a "safe" check.

Does it answer or did I get the question wrong?

carbonin: Ah I looked it up, looks like there's a matcher for what I'm talking about.

`ConsistOf` https://onsi.github.io/gomega/#consistofelement-interface

Really my suggestion is that no matter what happens with the implementation as long as the elements are the same the test should still pass. Using a matcher like this just makes the test a bit less fragile.

swagger.yaml
danielerez: Right, should be optional for now. Should be probably handled when removing cpu_architecture later on, hope it won't be too much mess:(

internal/oc/release_test.go
mkowalski: Done

swagger.yaml
mkowalski: Reverted. Yeah, in the future we should just swap which one is required

internal/versions/versions.go
mkowalski: The same applies to `amd64` vs `x86_64` - whatever the rule was now, the same rule will apply tomorrow. Again, I would personally prefer that we just drop this logic "if function is called with values pre-filled just trust them" but I don't want to make too many changes in a single PR.

It was meant to allow passing `[]string` instead `string` so as much as the question "what if the value passed is incorrect" is a valid one, I see it out of scope of the change

carbonin: :+1: yeah, I agree this isn't something you are changing here.
Just something I noticed while reading through.

Maybe the choice in the future is just to pull the image and detect the arch ourselves every time :man_shrugging:

mkowalski: Yes, I'll be very much in favour of just doing `oc` and/or `skopeo` ourselves instead of relying on what's passed. I'll definitely do it in a subsequent PR in this story, just want to get this one properly reviewed so it does not grow every day :)

mkowalski:
/test edge-e2e-ai-operator-ztp-multiarch-sno-ocp-411

Some flake with Agent rebooting, but multiarch extraction succeeded

```
time="2022-08-09T22:36:16Z" level=info msg="Stored release version 4.11.0-multi-rc.6 for architecture multiarch" func="github.com/openshift/assisted-service/internal/versions.(*handler).AddReleaseImage" file="/assisted-service/internal/versions/versions.go:369" pkg=versions
time="2022-08-09T22:36:16Z" level=info msg="Full list or architectures: [x86_64 ppc64le s390x arm64]" func="github.com/openshift/assisted-service/internal/versions.(*handler).AddReleaseImage" file="/assisted-service/internal/versions/versions.go:371" pkg=versions
```

mkowalski:

```
waiting.exceptions.TimeoutExpired: Timeout of 3600 seconds expired waiting for Nodes to be in of the statuses ['known']
```

mkowalski:
@carbonin @danielerez @vrutkovs @osherdp anyone with any further changes please?

data/default_okd_release_images.json
vrutkovs: LGTM

carbonin:
Test failure looks unrelated:
```
2022-08-10 11:51:26,200 - root - INFO - 139750863722304 - Asked hosts to be in one of the statuses from ['known'] and currently hosts statuses are [(1, 'd6c54288-26b3-4bba-9b9a-754254aa6564', 'test-infra-cluster-4a1e0fdd-worker-0', 'worker', 'insufficient', "Host cannot be installed due to following failing validation(s): Couldn't resolve domain name *.apps.test-infra-cluster-4a1e0fdd.redhat.com on the host. To continue installation, create the necessary DNS entries to resolve this domain name to your cluster's application ingress IP address")]
```

/lgtm

PR #4252, opened 2022-08-09
MGMT-11404: add rest-api doc for cluster tags

danielerez:
Added documentation for creating/updating cluster tags using the rest-api (in docs/user-guide).

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @filanov 
/cc @avishayt 
/cc @gamli75 
/cc @iovadia 

## Checklist

## Reviewers Checklist


avishayt:
Please add something about their use.  For example, that they are returned via cluster GET and can be used by the user for tracking installations.

docs/user-guide/rest-api-cluster-tags.md
```diff
@@ -0,0 +1,38 @@
+# REST-API - Cluster Tags
+
+The `tags` property in cluster object is a comma-separated list of tags that are associated to the cluster. Each tag is a free-text string that can consist of the following characters:
+Alphanumeric (aA-zZ, 0-9), underscore (_) and white-spaces.
```
gamli75, line 4: how do we remove tags? is an empty string a valid value?

danielerez:
> Please add something about their use. For example, that they are returned via cluster GET and can be used by the user for tracking installations (and that they're ignored by the system).

Sure, added some more details and an example for getting tags.

docs/user-guide/rest-api-cluster-tags.md
danielerez: right, it can be cleared by specifying an empty string, added a note for that.

docs/user-guide/rest-api-cluster-tags.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Create tags (using v2RegisterCluster)
+
+```bash
+cat register_cluster.json
+{
+    "name":"test",
+    "pull_secret":"<pull_secret>",
+    "openshift_version":"4.10",
+    "tags":"tag1,tag_2,tag 3"
```
romfreiman, line 25: do we allow spaces? what will happen in elk?

danielerez: @rccrdpccl Should we drop support for spaces inside tags to make it simpler for elasticsearch? IIUC, there could be an issue only with trailing spaces in elk? Or we would need a special tokenizer any way for spaces?

rccrdpccl: I think spaces within the word is fine, but we should trim leading and trailing spaces.
Changing analyzer and tokenizer right now it's not straightforward as we would need to reindex current data.

The current behaviour with above example it's the following (using `match` query on `cluster.tags` field):
* search term `tag`: no results found
* search term `tag1`: finds the document
* search term `tag2` or `tag3`: finds nothing
* search term `tag 3`: finds the correct document
* search term `tag `: finds nothing

However, the following example can be ambiguous:
`  "tags":"tag1 ,tag_2 ,tag 3 "`
* search term `tag1`: no result found
* search term `tag1 `: finds the document
* same with the other terms: `tag_2` `tag 3` will find nothing, `tag_2 ` and `tag 3 ` will find it

danielerez: So the current implementation already prevent having leading and trailing spaces. I.e. the validation ensures 'tag1,tag2,tag3' format, so the only concern was regarding spaces inside the tag - which seems indeed fine. Thanks!

gamli75 (approved):

PR #4253, opened 2022-08-09
MGMT-11542: Fix connectivity check running forever on large cluster

ori-amizur:
When having a large cluster that has many varying connectivity problems,
there are many possible groups.  Finding the optimal one takes a very
long time. This causes cluster monitor to never return which causes the
assisted installer pod to be restarted due to health problems.
The fix to this problem, is to provide a semi optimal problem. In case
there is full mesh connectivity between all nodes, the change will
provide the correct answer which means full connectivity.
In case that there are many possible full mesh connectivity groups, it
might not find the optimal one  (although in the tests that were run the
previous algorithm and the current one produce same results).

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @filanov 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/network/connectivity_groups.go
+++ b/internal/network/connectivity_groups.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
-	"sort"
 
 	"github.com/go-openapi/strfmt"
 	"github.com/golang-collections/go-datastructures/bitarray"
@@ -94,18 +93,15 @@ func (c *connectivitySet) intersect(other *connectivitySet) *connectivitySet {
 	}
 }
 
+func (c *connectivitySet) union(other *connectivitySet) *connectivitySet {
+	return &connectivitySet{
+		array: c.array.Or(other.array),
+	}
+}
+
 func (c *connectivitySet) containsElement(id int) bool {
 	b, err := c.array.GetBit(uint64(id))
 	return err == nil && b
-}
-
-func (c *connectivitySet) equals(other *connectivitySet) bool {
-	return c.array.Equals(other.array)
-}
-
-func (c *connectivitySet) isSupersetOf(other *connectivitySet) bool {
-	intersection := c.intersect(other)
-	return intersection.equals(other)
 }
 
 func (c *connectivitySet) id() groupId {
@@ -113,6 +109,10 @@ func (c *connectivitySet) id() groupId {
 		c.groupId = c.array.ToNums()
 	}
 	return c.groupId
+}
+
+func (c *connectivitySet) equals(other *connectivitySet) bool {
+	return c.array.Equals(other.array)
 }
 
 func (c *connectivitySet) toList(hosts []*models.Host) []strfmt.UUID {
@@ -123,7 +123,7 @@ func (c *connectivitySet) toList(hosts []*models.Host) []strfmt.UUID {
 	return ret
 }
 
-func (c *connectivitySet) len() int {
+func (c *connectivitySet) size() int {
 	return len(c.id())
 }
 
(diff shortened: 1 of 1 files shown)


internal/network/connectivity_groups.go
```diff
@@ -142,116 +143,235 @@ func (g groupId) isLess(other groupId) bool {
...
+	for _, cg := range groupsBySize {
+
+		// Take only the participants with the highest number of participating hosts
+		switch {
+		case cg.numParticipatingHosts() > numParticipatingHosts:
+			numParticipatingHosts = cg.numParticipatingHosts()
+			ret = nil
+			fallthrough
+		case cg.numParticipatingHosts() == numParticipatingHosts:
+			ret = append(ret, cg)
```
eranco74, line 265: Instead of appending and later sorting by id, consider updating ret if
`if cg.id.isLess(currentId)...`

internal/network/connectivity_groups.go
```diff
@@ -142,116 +143,235 @@ func (g groupId) isLess(other groupId) bool {
...
+	var ret []*connectivityGroup
+	var numParticipatingHosts int
+	for _, cg := range groupsBySize {
+
+		// Take only the participants with the highest number of participating hosts
+		switch {
+		case cg.numParticipatingHosts() > numParticipatingHosts:
+			numParticipatingHosts = cg.numParticipatingHosts()
+			ret = nil
+			fallthrough
```
eranco74, line 263: why fallthrough and not update ret?

internal/network/connectivity_groups.go
```diff
@@ -142,116 +143,235 @@ func (g groupId) isLess(other groupId) bool {
...
+	// First iteration - gather the original groups
+	for _, candidate := range groupCandidates {
+
+		// Add the set of the current candidate to the pending list
+		groupList.addOrMergeGroup(m.toConnectivityGroup(candidate))
+	}
+
+	// The primary groups are used for intersection with group candidate which is not full mesh
+	primaryGroups := groupList.setAndMergePrimaryGroups()
+	for {
```
eranco74, line 320: If `groupCandidate.isFullMesh()` is false we can hang here forever

internal/network/connectivity_groups.go
```diff
@@ -142,116 +143,235 @@ func (g groupId) isLess(other groupId) bool {
...
+		groupCandidate := groupList.largestGroup()
+		if groupCandidate != nil {
+			currentSize := groupCandidate.len()
+
+			if groupCandidate.isFullMesh() {
+				return groupCandidate
+			}
+
+			// Remove all groups with the current size, since we are now looking for smaller size groups
+			groupList.reduce(currentSize)
```
eranco74, line 330:
```suggestion
			if groupCandidate.isFullMesh() {
				return groupCandidate
			}

			currentSize := groupCandidate.len()
			// Remove all groups with the current size, since we are now looking for smaller size groups
			groupList.reduce(currentSize)
```

internal/network/connectivity_groups.go
```diff
@@ -142,116 +143,235 @@ func (g groupId) isLess(other groupId) bool {
...
+
+			// We now want to intersect with groups not smaller than the biggest group(s) still in the group list
+			biggestGroupSize := groupList.biggestGroupSize()
+			for i := range primaryGroups {
+
```
eranco74, line 335: Unsure what's going on here.
why do we need to check all these interactions and add them to the `groupList`?
why are these groups not part of the original `groupCandidates`?

internal/network/connectivity_groups.go
ori-amizur: OK

internal/network/connectivity_groups.go
ori-amizur: OK. Will keep single candidate

internal/network/connectivity_groups.go
ori-amizur: OK

internal/network/connectivity_groups.go
ori-amizur: Since the current group size removed (reduce), and new intersected group must be less than the current size, the size will be decreased in the next iteration

internal/network/connectivity_groups.go
ori-amizur: Intersections may be not the ones that were part of the original groups.
Assuming we have the following original groups
(1, 2, 3, 4,5)  (1,2,3,4) (1,2,3,5) then only the intersection (1,2,3) is common to all original groups.

internal/network/connectivity_groups.go
ori-amizur: Changed the loop so it will not seem as infinite loop.

eranco74 (commented):
/lgtm

PR #4259, opened 2022-08-10
MGMT-11543: When there are many connectivity check packet losses, the host monitoring thread gets stuck

ori-amizur:
When there are many connectivity check packet losses, the packet loss validation
status message contains list of the failed hosts, and the failure
percentage for each.
In order to create the list of failed hosts, each failed host is
identified by its IP address.  Since the IP address is in the inventory,
to identify the host all inventories in the cluster need to be
unmarshalled.
The change is to identify the host by host-id and not by the IP address.
In this case, inventory unmarshalling is not needed for host
identification, because the search is by host id.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @filanov 
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/common.go
+++ b/internal/host/common.go
@@ -2,7 +2,6 @@ package host
 
 import (
 	"fmt"
-	"net"
 	"net/http"
 	"time"
 
@@ -155,29 +154,25 @@ func updateRole(log logrus.FieldLogger, h *models.Host, role models.HostRole, su
 		*h.ID, h.InfraEnvID, srcRole).Updates(fields).Error
 }
 
-func GetHostnameAndEffectiveRoleByIP(ip string, hosts []*models.Host) (string, models.HostRole, error) {
+func FindHostByID(hostID strfmt.UUID, hosts []*models.Host) *models.Host {
 	for _, h := range hosts {
-		if h.Inventory == "" {
-			continue
-		}
-		inv, err := common.UnmarshalInventory(h.Inventory)
-		if err != nil {
-			return "", "", fmt.Errorf("unable to unmarshall cluster inventory for host %s: %s", h.RequestedHostname, err)
-		}
-		for _, i := range inv.Interfaces {
-			ips := append(i.IPV4Addresses, i.IPV6Addresses...)
-			for _, cidr := range ips {
-				parsedIP, _, err := net.ParseCIDR(cidr)
-				if err != nil {
-					return "", "", err
-				}
-				if ip == parsedIP.String() {
-					return getRealHostname(h, inv), common.GetEffectiveRole(h), nil
-				}
-			}
+		if hostID == *h.ID {
+			return h
 		}
 	}
-	return "", "", fmt.Errorf("host with IP %s not found in inventory", ip)
+	return nil
+}
+
+func GetHostnameAndEffectiveRoleByHostID(hostId strfmt.UUID, hosts []*models.Host, inventoryCache InventoryCache) (string, models.HostRole, error) {
+	host := FindHostByID(hostId, hosts)
+	if host == nil {
+		return "", "", fmt.Errorf("host with ID %s was not found", hostId.String())
+	}
+	inventory, err := inventoryCache.GetOrUnmarshal(host)
+	if err != nil {
+		return "", "", err
+	}
+	return getRealHostname(host, inventory), common.GetEffectiveRole(host), nil
 }
 
 func FindMatchingStages(role models.HostRole, bootstrap, isSNO bool) []models.HostStage {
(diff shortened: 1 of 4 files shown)


internal/host/validator.go
```diff
@@ -927,7 +926,7 @@ func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRole
 		for _, l3 := range r.L3Connectivity {
 			if l3.AverageRTTMs > *clusterRoleReqs.Total.NetworkLatencyThresholdMs {
 				if _, ok := failedHostIPs[l3.RemoteIPAddress]; !ok {
-					hostname, role, err := GetHostnameAndEffectiveRoleByIP(l3.RemoteIPAddress, hosts)
```
danielerez, line 930: Can we remove the old GetHostnameAndEffectiveRoleByIP func? Seems it would be unused after this change?

internal/host/common.go
```diff
@@ -180,6 +180,27 @@ func GetHostnameAndEffectiveRoleByIP(ip string, hosts []*models.Host) (string, m
...
+func FindHostByID(hostID strfmt.UUID, hosts []*models.Host) *models.Host {
+	for _, h := range hosts {
+		if hostID == *h.ID {
+			return h
+		}
+	}
+	return nil
+}
+
+func GetHostnameAndEffectiveRoleByHostID(hostId strfmt.UUID, hosts []*models.Host, inventoryCache InventoryCache) (string, models.HostRole, error) {
```
danielerez, line 192: Do we need a test for this func in common_test? Could be instead of the current 'GetHostnameAndEffectiveRoleByIP'

internal/host/common.go
```diff
@@ -180,6 +180,27 @@ func GetHostnameAndEffectiveRoleByIP(ip string, hosts []*models.Host) (string, m
 	return "", "", fmt.Errorf("host with IP %s not found in inventory", ip)
 }
 
+func FindHostByID(hostID strfmt.UUID, hosts []*models.Host) *models.Host {
+	for _, h := range hosts {
```
danielerez, line 184: Not crucial, but just for clarity can be simplified to something like:
```
return funk.Find(hosts, func(host *models.Host) bool {
    return host.ID == hostID
})
```

internal/host/validator.go
ori-amizur: It is still in use

danielerez: > It is still in use

Oh, I assumed only the validator is using it. Where is it in use elsewhere?

internal/host/common.go
ori-amizur: OK

internal/host/common.go
ori-amizur: There is cast on the return value only if it is not nil.  So doesn't seem simpler.

internal/host/common_test.go
```diff
@@ -15,81 +15,85 @@ import (
 type hostNetProfile struct {
 	role     models.HostRole
 	hostname string
+	id       strfmt.UUID
 	ip       string
 }
 
 var _ = Describe("GetHostnameAndEffectiveRoleByIP", func() {
```
danielerez, line 22: can be renamed now to GetHostnameAndEffectiveRoleByHostID

internal/host/common_test.go
```diff
@@ -15,81 +15,85 @@ import (
...
-	hostrolesIpv6 := []hostNetProfile{{role: models.HostRoleMaster, hostname: "master-1", ip: "1001:db8::11"}, {role: models.HostRoleWorker, hostname: "worker-1", ip: "1001:db8::12"}}
+	hostRolesIpv4 := []hostNetProfile{{role: models.HostRoleMaster, hostname: "master-0", ip: "1.2.3.1", id: strfmt.UUID(uuid.New().String())},
+		{role: models.HostRoleWorker, hostname: "worker-0", ip: "1.2.3.2", id: strfmt.UUID(uuid.New().String())}}
+	hostrolesIpv6 := []hostNetProfile{{role: models.HostRoleMaster, hostname: "master-1", ip: "1001:db8::11", id: strfmt.UUID(uuid.New().String())},
+		{role: models.HostRoleWorker, hostname: "worker-1", ip: "1001:db8::12", id: strfmt.UUID(uuid.New().String())}}
 	clusterID := strfmt.UUID(uuid.New().String())
 	infraEnvID := strfmt.UUID(uuid.New().String())
+	nonExistantHostID := strfmt.UUID(uuid.New().String())
 
 	Context("resolves hostname and role based on IP", func() {
```
danielerez, line 32: '...based on host ID'

danielerez (commented):
looks good, just a couple of minor comments in test

PR #4263, opened 2022-08-10
MGMT-11334: Add additional API models for service-to-agent host reclaim

CrystalChun:
[MGMT-11334](https://issues.redhat.com/browse/MGMT-11334)
Adds the host directory to the download artifacts request model
and to a new model (reboot for reclaim request) from discussion in
https://github.com/openshift/assisted-installer-agent/pull/415#discussion_r942442609

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Assignees

/cc @carbonin 
/cc @avishayt 

## Checklist

## Reviewers Checklist


swagger.yaml
```diff
@@ -5469,6 +5469,7 @@ definitions:
       - kernel_url
       - rootfs_url
       - initrd_url
+      - host_fs_mount_dir
```
carbonin, line 5472: Maybe `host_filesystem_dir` or `host_filesystem_mount` if we want to fully avoid abbreviations?

@avishayt thoughts? Naming is hard :confounded:

swagger.yaml
```diff
@@ -5479,6 +5480,19 @@ definitions:
       initrd_url:
         type: string
         description: URL address to download the initrd.
+      host_fs_mount_dir:
+        type: string
+        description: The base directory on the host that contains the /boot folder.
```
omertuc, line 5485: what is this directory for? if the agent is going to download the artifacts into that directory when this request is sent, I think the description should mention that

swagger.yaml
```diff
@@ -5479,6 +5480,19 @@ definitions:
...
+
+  reboot_for_reclaim_request:
+    type: object
+    description: Information sent to the agent for rebooting a host into discovery.
+    required:
+      - host_fs_mount_dir
+    properties:
+      host_fs_mount_dir:
+        type: string
+        description: The base directory on the host that contains the /boot folder.
```
omertuc, line 5495: >reboot_for_reclaim_request

How is the directory used in a `reboot_for_reclaim_request`? I would assume the agent would simply need to reboot at this stage, wouldn't it? Is it doing anything with that directory? If yes it could be nice if it were mentioned in the description

swagger.yaml
CrystalChun: Thank you for the feedback! I've added that to the description

swagger.yaml
CrystalChun: Thanks for the feedback! I've modified the description to mention what it's used for

swagger.yaml
```diff
@@ -5479,6 +5480,23 @@ definitions:
       initrd_url:
         type: string
         description: URL address to download the initrd.
+      host_fs_mount_dir:
```
avishayt, line 5483: How does the service know this?  Can the agent detect it so that it's more future-proof?

CrystalChun: Currently it's hard-coded [here](https://github.com/openshift/assisted-service/blob/347339cb7e089c34d51e7085d9fe20c16826b319/internal/controller/controllers/agent_reclaimer.go#L159), but I've made it an assisted-service environment variable in my latest [PR](https://github.com/openshift/assisted-service/pull/4271/files#diff-5fb1b62b486cd67165e89476a57536baa3fc9822734d1e2922b8a9c9484e9f18R32). May I get feedback on that? Should it be provided in a different way?

Good question about agent detection, since the agent is running in a pod, it could also be provided as an environment variable of the pod itself when the agent is spun up. @carbonin what do you think?

carbonin: Either way we should have one place for the source of this data.

I suppose we could provide it as an env var rather than through the request, but I recall thinking that adding more env vars or args to the agent that only applied to a single operation didn't seem like a good pattern.

> Can the agent detect it so that it's more future-proof?

I think @avishayt is asking if there's something we can look at in the `mount` output that might tell us where the host rootfs is mounted.

Maybe, but I'm not sure it would be particular straight forward.

carbonin:
I'm good with this as is.

Will let @avishayt have another look though

avishayt (approved):

PR #4274, opened 2022-08-17
Bug-2118970: Create an utility GetIgnitionEndpoint function

mkowalski:
Contributes-to: [MGMTBUGSM-525](https://issues.redhat.com//browse/MGMTBUGSM-525)

### DIFF
--- a/internal/host/hostutil/host_utils.go
+++ b/internal/host/hostutil/host_utils.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
+	"path"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


internal/host/hostutil/host_utils.go
```diff
@@ -298,3 +300,20 @@ func IsDiskEncryptionEnabledForRole(encryption models.DiskEncryption, role model
 		return false
 	}
 }
+
+func GetIgnitionEndpoint(cluster *common.Cluster, host *models.Host) (string, error) {
+	poolName := string(common.GetEffectiveRole(host))
+	if host.MachineConfigPoolName != "" {
+		poolName = host.MachineConfigPoolName
+	}
```
eranco74, line 308: Can we have a default "worker" here or in GetEffectiveRole?

slaviered: Not in GetEffectiveRole since the default of role is auto-assign

mkowalski: Sure, just added it here. I hope it's a dead code, but we never know

PR #4275, opened 2022-08-17
MGMT-8890: Provide a multiarch RELEASE_IMAGE

mkowalski:
This commit provides a sample of multiarch image that should be passed
to the RELEASE_IMAGES.


javipolo (approved):

mkowalski:
/hold

As discussed with @osherdp, we will wait with this change till we agree on using multiarch image in SaaS. This change here will populate to some places (thought not our prod env) so we don't want to introduce it transparently without people being aware.

mkowalski:
`edge-e2e-metal-assisted-day2-single-node` is failing already for a month ¯\_(ツ)_/¯

mkowalski:
https://github.com/openshift/assisted-service/pull/4293 should merge first so that we use `cpu_architecture: multi`

tools/handle_ocp_versions.py
```diff
@@ -31,7 +31,7 @@ def verify_images(release_images: List[str]):
 
 def verify_release_version(ocp_version: str, release_image: str, release_version: str):
     oc_version = get_oc_version(release_image)
-    assert oc_version == release_version, f"{release_image} full version is {oc_version} not {release_version}"
+    assert oc_version == release_version or oc_version == release_version+"-multi", f"{release_image} full version is {oc_version} not {release_version}"
```
osherdp, line 34: pep8 dictates:
```suggestion
    assert oc_version == release_version or oc_version == release_version + "-multi", f"{release_image} full version is {oc_version} not {release_version}"
```
also let's split this line to make it more readable

data/default_release_images.json
```diff
@@ -37,15 +37,15 @@
     },
     {
         "openshift_version": "4.11",
-        "cpu_architecture": "x86_64",
-        "url": "quay.io/openshift-release-dev/ocp-release:4.11.0-x86_64",
+        "cpu_architecture": "multi",
+        "cpu_architectures": [
```
osherdp, line 41: so what is the intended difference between ``cpu_architecture`` and ``cpu_architectures``?
maybe it's better to rename ``cpu_architecture`` to ``image_architecture`` to annotate it as the one relevant to the docker image arch type?

osherdp (changes requested):

tools/handle_ocp_versions.py
mkowalski: Sure, just fixed

data/default_release_images.json
mkowalski: Both fields have the same meaning. One is `string`, the other is `[]string` but the underlying meaning is the same for both of them. The naming (i.e. `cpu_architectures`) has been agreed on https://github.com/openshift/assisted-service/pull/4248. In the future we will deprecate `string` in favour of `[]string`, but now we need to keep backwards compatibility

mkowalski:
/unhold

Dependencies have been merged, ready to test

osherdp (approved):

PR #4279, opened 2022-08-18
MGMT-11694: Add `NoSkipInstallationDisk` and `NoSkipMissingDisk` to `isSufficientForInstall`

omertuc:
They were supposed to be added during b0d095d7bf1e2c6fbbc935b7107f13c98ee86f8a but they weren't.

Without adding them to this list, installation is allowed to proceed
even with those validations failing

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
    - Testing is still in progress
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/host/transition.go
```diff
@@ -30,6 +30,39 @@ type transitionHandler struct {
 	eventsHandler eventsapi.Handler
 }
 
+//go:generate mockgen -source=transition.go -package=host -destination=mock_transition.go
+type TransitionHandler interface {
```
omertuc, line 34: Made transitionHandler struct conform to an interface to it can be mocked. This mocking allows us to easily test the state machine in isolation

internal/host/statemachine.go
```diff
@@ -637,6 +637,8 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th *transitionHandler) sta
 		If(SufficientOrUnknownInstallationDiskSpeed),
 		If(NonOverlappingSubnets),
 		If(CompatibleAgent),
+		If(NoSkipInstallationDisk),
```
omertuc, line 640: These 2 lines are the only "real" change, the rest of the changes are just tests / changes that allow me to test them

internal/host/transition_test.go
```diff
@@ -5988,3 +5989,153 @@ func validateEqualProgress(p1, p2 *models.HostProgressInfo) {
...
+func (ts *testState) State() stateswitch.State {
+	return stateswitch.State(ts.state)
+}
+
+func (ts *testState) SetState(state stateswitch.State) error {
+	ts.state = string(state)
+	return nil
+}
+
+var _ = Describe("State machine test", func() {
```
filanov, line 6012: Suggesting to have several different tests cases here. 
For example this describe can scope only refresh functions. 
Refresh is not the only operations of the state machine.

internal/host/transition_test.go
```diff
@@ -5988,3 +5989,153 @@ func validateEqualProgress(p1, p2 *models.HostProgressInfo) {
...
+		refreshHostArgs       TransitionArgsRefreshHost
+		testState             *testState
+	)
+
+	BeforeEach(func() {
+		mockController = gomock.NewController(GinkgoT())
+
+		mockTransitionHandler = NewMockTransitionHandler(mockController)
+
+		mockTransitionHandler.EXPECT().HasStatusTimedOut(gomock.Any()).Return(func(_ stateswitch.StateSwitch, _ stateswitch.TransitionArgs) (bool, error) {
```
filanov, line 6027: isn't that specific for your tests? maybe it should be under the same context or something like that.

internal/host/transition_test.go
```diff
@@ -5988,3 +5989,153 @@ func validateEqualProgress(p1, p2 *models.HostProgressInfo) {
...
+			return false, nil
+		}).AnyTimes()
+
+		mockTransitionHandler.EXPECT().PostRefreshHost(gomock.Any()).Return(
```
filanov, line 6031: IMO using any times will make the tests more shallow

internal/host/transition_test.go
```diff
@@ -5988,3 +5989,153 @@ func validateEqualProgress(p1, p2 *models.HostProgressInfo) {
...
+			func(_ stateswitch.StateSwitch, _ stateswitch.TransitionArgs) error { return nil },
+		).AnyTimes()
+
+		mockTransitionHandler.EXPECT().PostPreparingForInstallationHost(gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
+
+		stateMachine = NewHostStateMachine(stateswitch.NewStateMachine(), mockTransitionHandler)
+
+		testState = newTestState(models.HostStatusKnown)
+
+		allValidationIDs := []validationID{
```
filanov, line 6045: Suggesting to have a helper function that will create this list so it can be reused in different tests and we will have it once

internal/host/transition_test.go
```diff
@@ -5988,3 +5989,153 @@ func validateEqualProgress(p1, p2 *models.HostProgressInfo) {
...
+			refreshHostArgs.validationResults[validationCategory] = append(
+				refreshHostArgs.validationResults[validationCategory], ValidationResult{
+					ID:      validationID,
+					Status:  ValidationSuccess,
+					Message: "Test validation",
+				})
+			refreshHostArgs.conditions[string(validationID)] = true
+		}
+
+		Expect(stateMachine.Run(TransitionTypeRefresh, testState, &refreshHostArgs)).To(Succeed())
```
filanov, line 6114: why do you need it in the before each? isn't that part of the test? 
if it's to bring the host into a known state then again it's very specific for your case and not part of the generic refresh test
you can have a context that will define that you want to start from a known state

filanov (commented):
I really like this change, have few suggestions to make it more helpful for later usage in different cases of the state machine

internal/host/transition_test.go
omertuc: My attitude towards mocks call counts are -

- If I use a mock to test that a particular external function is actually being called as expected, I probably want to specify the exact number of times I expect it to be called. 

- If I use mock just so my code can run without issues, I use `AnyTimes()`, because I don't want to be too opinionated regarding how many times it gets called - it might change in the future, and my test couldn't care less. Since my test is not dealing with testing that this particular call happens, my test also doesn't care how many times it gets called.

mkowalski: I agree with Omer on that one. From my experience, when writing a big test I tend to spend more time on figuring out _what_ and _how many times_ I need to mock to get it passing instead of focusing on whether my feature is being tested properly. I think AnyTimes, Maybe & co. are perfectly acceptable as long as they are not core of the test

internal/host/transition_test.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/filanov/stateswitch"
```
mkowalski, line 13: May be opinionated, but shouldn't we own that?

internal/host/transition_test.go
```diff
@@ -5988,3 +5989,176 @@ func validateEqualProgress(p1, p2 *models.HostProgressInfo) {
...
+var allConditions []conditionId = []conditionId{
+	InstallationDiskSpeedCheckSuccessful,
+	ClusterPreparingForInstallation,
+	ClusterPendingUserAction,
+	ClusterInstalling,
+	ValidRoleForInstallation,
+	StageInWrongBootStages,
+	ClusterInError,
+	SuccessfulContainerImageAvailability,
+}
```
mkowalski, line 6061: I have somehow hard time believing that the next person adding new validation and/or condition will remember to add it here. Can we invent any mechanism to somehow ensure/enforce that? If no that's okay, but I just want to raise awareness here

omertuc: No clean way of doing that that I can think of

internal/host/transition_test.go
omertuc: That's up to the stateswitch BDFL @filanov

internal/host/transition_test.go
omertuc: No, this function gets immediately called during `NewHostStateMachine`

omertuc: If this will cause problems to anyone in the future they can change it, this is not set in stone

internal/host/transition_test.go
omertuc: To ensure we've actually set up the validations correctly by running a refresh and checking that we're still in Known

internal/host/transition_test.go
filanov: i don't mind to move it

internal/host/transition_test.go
filanov: then maybe it show that the code need to be more simple and break it more. 
not a deal breaker for me.

filanov (approved):

PR #4283, opened 2022-08-19
MGMT-8890: Unit test to ensure compatibility of release images

mkowalski:
This PR adds a set of explicit unit tests to ensure that extending
release image with `cpu_architectures` didn't break compatibility with
images using `cpu_architecture` and ignoring the old fields.

This tests adds explicit coverage for the following two scenarios:

1) image containing `cpu_architecture` and missing `cpu_architectures`
2) image containing both `cpu_architecture` and `cpu_architectures`

Contributes-to: [MGMT-11494](https://issues.redhat.com//browse/MGMT-11494)

### DIFF
--- a/internal/versions/versions_test.go
+++ b/internal/versions/versions_test.go
@@ -71,11 +71,20 @@ var defaultOsImages = models.OsImages{
 
 var defaultReleaseImages = models.ReleaseImages{
 	&models.ReleaseImage{
-		CPUArchitecture:  swag.String(common.MultiCPUArchitecture),
-		CPUArchitectures: []string{common.X86CPUArchitecture, common.ARM64CPUArchitecture},
+		// This image uses a syntax with missing "cpu_architectures". It is crafted
+		// in order to make sure the change in MGMT-11494 is backwards-compatible.
+		CPUArchitecture:  swag.String("fake-architecture-chocobomb"),
+		CPUArchitectures: []string{},
 		OpenshiftVersion: swag.String("4.11.1"),
 		URL:              swag.String("release_4.11.1"),
-		Version:          swag.String("4.11.1-candidate"),
+		Version:          swag.String("4.11.1-fake-chocobomb"),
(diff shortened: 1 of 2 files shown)


internal/versions/versions_test.go
```diff
@@ -70,12 +70,20 @@ var defaultOsImages = models.OsImages{
 }
 
 var defaultReleaseImages = models.ReleaseImages{
+	&models.ReleaseImage{
+		// This image uses a syntax with missing "cpu_architectures". It is crafted
+		// in order to make sure the change in MGMT-11494 is backwards-compatible.
+		CPUArchitecture:  swag.String("fake-stripped-architecture"),
+		OpenshiftVersion: swag.String("4.11.1"),
```
omertuc, line 77: To be more explicit

```suggestion
		CPUArchitectures: []string{},
		OpenshiftVersion: swag.String("4.11.1"),
```

internal/versions/versions_test.go
```diff
@@ -70,12 +70,20 @@ var defaultOsImages = models.OsImages{
 }
 
 var defaultReleaseImages = models.ReleaseImages{
+	&models.ReleaseImage{
+		// This image uses a syntax with missing "cpu_architectures". It is crafted
+		// in order to make sure the change in MGMT-11494 is backwards-compatible.
+		CPUArchitecture:  swag.String("fake-stripped-architecture"),
```
omertuc, line 76: What does stripped mean in this context? The meaning isn't obvious to me, so if you can think of a different english word(s) to describe it it could be more clear

internal/versions/versions_test.go
mkowalski: Done

internal/versions/versions_test.go
mkowalski: Just removed it. With adding explicit empty array it should be clear enough

PR #4285, opened 2022-08-19
MGMT-8890: InfraEnv to validate content of release image

mkowalski:
This PR adds a logic that ensures we create InfraEnvs that refer to
existing release images. This is in order to prevent the following
scenario

* release image contains architecture [x86_64 ppc64le s390x]
* OS images available for [x86_64 arm64]
* infraenv created for arm64

The current logic for creating the ISO already detects the case when
InfraEnv uses architecture for which no RHCOS image is available.
However a scenario where RHCOS exists but OpenShift release image does
not was not covered till now. Without the change introduced by this PR
the missing release image would be detected only during the
installation. With this change it will happen very early in the
process so that we can prevent the installation from even starting.

Closes: [MGMT-11500](https://issues.redhat.com//browse/MGMT-11500)
Closes: [MGMT-11569](https://issues.redhat.com//browse/MGMT-11569)

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -3804,6 +3804,35 @@ func validateProxySettings(httpProxy, httpsProxy, noProxy, ocpVersion *string) e
 	return nil
 }
 
+// validateArchitectureAndVersion validates if architecture specified inside Infraenv matches one
+// specified for the cluster. For single-arch clusters the validation needs to only compare values
+// of the params. For multiarch cluster we want to see if the multiarch release image contains the
+// the architecture specifically requested by the InfraEnv. We don't need to explicitly validate if
+// the OS image exists because if not, this will be detected by the function generating the ISO.
+func validateArchitectureAndVersion(v versions.Handler, c *common.Cluster, cpuArch, ocpVersion string) error {
+	var err error
+
+	// For late-binding we don't know the cluster yet
+	if c == nil {
+		return nil
+	}
+	if ocpVersion == "" {
+		ocpVersion = c.OpenshiftVersion
+	}
+	if c.CPUArchitecture != common.MultiCPUArchitecture {
+		if c.CPUArchitecture != "" && c.CPUArchitecture != cpuArch {
+			return errors.Errorf("Specified CPU architecture (%s) doesn't match the cluster (%s)", cpuArch, c.CPUArchitecture)
+		}
+	} else {
+		_, err = v.GetReleaseImage(ocpVersion, cpuArch)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 func secretValidationToUserError(err error) error {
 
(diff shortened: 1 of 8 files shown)


mkowalski:
/cc @danielerez
/cc @ori-amizur 
/cc @carbonin 

Merging https://github.com/openshift/assisted-service/pull/4275 is a hard-requirement for this one, but we already have passing subsystem tests here so it is ready for review

mkowalski:
/retest-required

We suspect a flake with RHCOS 4.11; more details in https://coreos.slack.com/archives/C014N2VLTQE/p1661283378019209?thread_ts=1661236700.443779&cid=C014N2VLTQE but here let's just retry to get a sample

internal/versions/versions.go
```diff
@@ -168,16 +168,6 @@ func (h *handler) GetDefaultReleaseImage(cpuArchitecture string) (*models.Releas
...
 // Returns the OsImage entity
 func (h *handler) GetOsImage(openshiftVersion, cpuArchitecture string) (*models.OsImage, error) {
-	// For multiarch clusters we are offloading the validation checking for existing OS image matching the version
-	// and the architecture directly to the InfraEnv. This is because Cluster object on its own cannot know which
-	// architecture is going to be used and we cannot enforce that OS images exist for every architecture from the
-	// multiarch release image.
-	// In the multiarch scenario it's up to the InfraEnv's ISO generation part to detect whether the required
-	// OS image exists or not. This is because we still have InfraEnvs per-architecture and don't allow for multiarch
-	// InfraEnvs. Based on that, it's reasonable to skip the validation for Cluster and let InfraEnv handle it.
-	if cpuArchitecture == common.MultiCPUArchitecture {
```
danielerez, line 178: Is it safe to remove this now? No scenario calls GetOsImage with common.MultiCPUArchitecture?

mkowalski: Yeah, no one should ever call this (because OS Image is function of InfraEnv). If by any reason it happens though, we will not fail, we will just return an error saying `The requested CPU architecture (multiarch) isn't specified in OS images list`

internal/bminventory/inventory.go
```diff
@@ -3804,6 +3804,36 @@ func validateProxySettings(httpProxy, httpsProxy, noProxy, ocpVersion *string) e
...
+// of the params. For multiarch cluster we want to see if the multiarch release image contains the
+// the architecture specifically requested by the InfraEnv. We don't need to explicitly validate if
+// the OS image exists because if not, this will be detected by the function generating the ISO.
+func validateArchitectureAndVersion(v versions.Handler, c *common.Cluster, cpuArch, ocpVersion string) error {
+	var err error
+
+	// For late-binding we don't know the cluster yet
+	if c == nil {
+		return nil
+	} else {
```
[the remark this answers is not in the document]
mkowalski: Sure thing, just fixed

PR #4299, opened 2022-08-23
MGMT-11182: Allow base level domain

slaviered:
Enable top level domain DNS name to be used for support of local
dns servers.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/subsystem/cluster_test.go
+++ b/subsystem/cluster_test.go
@@ -909,9 +909,9 @@ var _ = Describe("Validate BaseDNSDomain when creating a cluster", func() {
 	}
 	tests := []DNSTest{
 		{
-			It:            "V2RegisterCluster should throw an error. BaseDNSDomain='example', not a valid DNS structure string",
+			It:            "V2RegisterCluster should not throw an error. BaseDNSDomain='example', valid DNS",
 			BaseDNSDomain: "example",
-			ShouldThrow:   true,
+			ShouldThrow:   false,
 		},
 		{
 			It:            "V2RegisterCluster should throw an error. BaseDNSDomain='example.c', Invalid top-level domain name ",
--- a/pkg/validations/validations.go
+++ b/pkg/validations/validations.go
@@ -14,6 +14,7 @@ import (
 )
 
 const (
+	baseDomainRegex          = "^([a-z0-9]+(-[a-z0-9]+)*)+$"
 	dnsNameRegex             = "^([a-z0-9]+(-[a-z0-9]+)*[.])+[a-z]{2,}$"
 	hostnameRegex            = `^[a-z0-9][a-z0-9\-\.]{0,61}[a-z0-9]$`
 	installerArgsValuesRegex = `^[A-Za-z0-9@!#$%*()_+-=//.,";':{}\[\]]+$`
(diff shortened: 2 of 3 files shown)


subsystem/cluster_test.go
```diff
@@ -909,9 +909,9 @@ var _ = Describe("Validate BaseDNSDomain when creating a cluster", func() {
 	}
 	tests := []DNSTest{
 		{
-			It:            "V2RegisterCluster should throw an error. BaseDNSDomain='example', not a valid DNS structure string",
+			It:            "V2RegisterCluster should not throw an error. BaseDNSDomain='example', valid DNS",
 			BaseDNSDomain: "example",
-			ShouldThrow:   true,
+			ShouldThrow:   false,
```
mkowalski, line 914: Can you please add a subsystem test with a negative scenario? I.e. it's good to have `domain=example` that will succeed, but I'd also like to see a subsystem test with some domain that will fail, e.g. `basednsdomain=a-.com` and `shouldthrow: true`.

Apart from that looks good, this one test and LGTM

subsystem/cluster_test.go
```diff
@@ -909,9 +909,9 @@ var _ = Describe("Validate BaseDNSDomain when creating a cluster", func() {
...
 		},
 		{
 			It:            "V2RegisterCluster should throw an error. BaseDNSDomain='example.c', Invalid top-level domain name ",
```
mkowalski, line 917: I am confused here, that's what we discussed offline. Why is `example.c` invalid? Top Level Domain of 1 character is technically a valid one. It is defined in https://www.rfc-editor.org/rfc/rfc952 as `<name>  ::= <let>[*[<let-or-digit-or-hyphen>]<let-or-digit>]`.

Apart from that there is https://datatracker.ietf.org/doc/html/rfc3696#section-2 which states
```
additional rule
that essentially requires that top-level domain names not be all-
numeric
```

So `example.c` is a valid name whereas `example.1` is invalid

subsystem/cluster_test.go
slaviered: This is a classic example for a code that does not need subsystem at all. The algorithm is tested thoroughly with unit test (as it should be) and there is no need for extra subsystem test that just add a thin layer of calling the algorithm through REST.

slaviered: Plus I missed the part of the subsystem but of course fixed it already. "example" should be a legal domain now

subsystem/cluster_test.go
slaviered: @mkowalski  For some reason the existing validation (which I did not change as it was not part of the PR requirement strictly speaking) requires it to be a 2 or more letter suffix. I Believe it is because we are aiming for a real world DNS address where the last part will be .com or .au). I am not sure we need to relax this limitation. What do you think?

subsystem/cluster_test.go
mkowalski: Ok, fair enough 👍🏻

subsystem/cluster_test.go
mkowalski: As I said, real world does allow for a suffix that contains only 1 letter. Sadly I cannot answer why the original regex had the requirement of "minimum 2 characters" but this requirement does not come from the standard (i.e. RFC, anything what is not in RFC is someone's implementation, not "standard"). It may be one of the vendor's implementation (e.g. Microsoft has more strict requirements for DNS names), it may be someone's invented rule, I just can't say why we had it in the first place (just like we don't know why we had a limitation that we need a "dot" so we landed with the issue you are fixing)

Anyway, you are fixing what got initially reported so I'm okay keeping it like this till we get a request to soften it again in the future. The change is too small and too irrelevant to keep it hanging forever :)

PR #4302, opened 2022-08-23
WIP: MGMT-11736: Support ASC in multi namespaces

slaviered:
To support deploying an assisted infrastructure operator on zero workers hyper shift we need first to support deploying the operator on multiple namespaces.

To support backward compatibility, if the operator is deployed to the assisted-installer namespace we assume that the setup is a cluster scoped deployment. Otherwise, we implicitly assume that the hyper shift mode is on.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @
/cc @

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -106,9 +106,6 @@ type AgentServiceConfigReconciler struct {
 	Scheme   *runtime.Scheme
 	Recorder record.EventRecorder
 
-	// Namespace the operator is running in
-	Namespace string
-
 	// selector and tolerations the Operator runs in and propagates to its deployments
 	NodeSelector map[string]string
 	Tolerations  []corev1.Toleration
@@ -161,10 +158,18 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
 	log.Info("AgentServiceConfig Reconcile started")
 
 	instance := &aiv1beta1.AgentServiceConfig{}
+	ascKey := types.NamespacedName{Name: req.NamespacedName.Name}
 
-	// NOTE: ignoring the Namespace that seems to get set on request when syncing on namespaced objects
-	// when our AgentServiceConfig is ClusterScoped.
-	if err := r.Get(ctx, types.NamespacedName{Name: req.NamespacedName.Name}, instance); err != nil {
+	//for backwards compatability, if the namespace is assisted-installer we are looking
+	//for a cluster scoped CRD. Otherwise, we will look for a namespace-scoped resource
+	//and implicitly assume that we are in hypershift mode (L0-L1)
+	if req.NamespacedName.Namespace != "" {
+		ascKey.Namespace = req.NamespacedName.Namespace
+	} else {
+		log.Warning("DEBUG ==> ASC is not namespace scopped")
+	}
+
+	if err := r.Get(ctx, ascKey, instance); err != nil {
 		if errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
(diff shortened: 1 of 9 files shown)


omertuc:
>To support backward compatibility, if the operator is deployed to the assisted-installer namespace we assume that the setup is a cluster scoped deployment. Otherwise, we implicitly assume that the hyper shift mode is on.

Which part of this PR diff is this paragraph referring to?

carbonin:
> To support backward compatibility, if the operator is deployed to the assisted-installer namespace we assume that the setup is a cluster scoped deployment. Otherwise, we implicitly assume that the hyper shift mode is on.

I'm also not convinced this is the best way to achieve this.
I'd prefer some explicit switch somewhere for the user to dictate that they want the behavior to switch.

carbonin:
I think it's also worth explaining fully in the commit message and PR description why hypershift zero-worker mode requires this change.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -161,10 +158,18 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
...
 	instance := &aiv1beta1.AgentServiceConfig{}
+	ascKey := types.NamespacedName{Name: req.NamespacedName.Name}
 
-	// NOTE: ignoring the Namespace that seems to get set on request when syncing on namespaced objects
-	// when our AgentServiceConfig is ClusterScoped.
-	if err := r.Get(ctx, types.NamespacedName{Name: req.NamespacedName.Name}, instance); err != nil {
+	//for backwards compatability, if the namespace is assisted-installer we are looking
+	//for a cluster scoped CRD. Otherwise, we will look for a namespace-scoped resource
+	//and implicitly assume that we are in hypershift mode (L0-L1)
+	if req.NamespacedName.Namespace != "" {
```
carbonin, line 166: When would this be empty?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -176,15 +181,7 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
...
-	// We only support one AgentServiceConfig per cluster, and it must be called "agent". This prevents installing
-	// AgentService more than once in the cluster.
-	if instance.Name != agentServiceConfigName {
-		reason := fmt.Sprintf("Invalid name (%s)", instance.Name)
-		msg := fmt.Sprintf("Only one AgentServiceConfig supported per cluster and must be named '%s'", agentServiceConfigName)
-		log.Info(fmt.Sprintf("%s: %s", reason, msg), req.NamespacedName)
-		r.Recorder.Event(instance, "Warning", reason, msg)
-		return reconcile.Result{}, nil
-	}
+	log.Info("DEBUG ==> ASC instance is", instance.Name, instance.Namespace)
```
carbonin, line 184: I think this log can be removed, but if you do actually want a debug message here, you should use `log.Debug`

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -918,9 +915,9 @@ func (r *AgentServiceConfigReconciler) newImageServiceConfigMap(ctx context.Cont
 	return cm, mutateFn, nil
 }
 
-func (r *AgentServiceConfigReconciler) urlForRoute(ctx context.Context, routeName string) (string, error) {
+func (r *AgentServiceConfigReconciler) urlForRoute(ctx context.Context, routeName string, instance *aiv1beta1.AgentServiceConfig) (string, error) {
```
carbonin, line 918: If namespace is the only thing you need for this function, why not just pass that instead of the whole instance?

carbonin (commented):
What happens on upgrade?

How will an existing `AgentServiceConfig` instance receive a namespace?

internal/controller/controllers/agentserviceconfig_controller.go
slaviered: It is empty now (with the cluster scoped version). All other resources however do have the namespace "assisted-installer"

internal/controller/controllers/agentserviceconfig_controller.go
slaviered: I am passing an instance reference because it is more clear that way. On the called function then it is obvious what namespace is being used.

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: Right, but `req.NamespacedName` will always be the key for an `AgentServiceConfig` instance.
And after this PR that should always have a namespace, right?

carbonin: I think this ties into what happens on upgrade.

carbonin:
@slaviered can we close this one since we've decided to use a different CRD?

slaviered:
This design option is not relevant any more. There is a new CRD for the zero-worker-hypershift case

PR #4307, opened 2022-08-24
NO-ISSUE: For ZTP CI add more Hive condition checks

mkowalski:


osherdp (approved):

mkowalski:
/hold

Let me check the status of the optional ZTP jobs before shipping it

mkowalski:
/close

From discussion with Hive team, we should not use `ProvisionStopped` as a marker for installation being successful. Using `.spec.installed` is the good way

PR #4309, opened 2022-08-24
MGMT-11442: Add override ISO type parameter for v2DownloadInfraEnvFiles

paul-maidment:
This adds a non-mandatory field that may contain the values 'full-iso' or 'minimal-iso' when it is set in the request.
This will override any defaults that we read from the infra-env
and will inject or omit static network config from the discovery ignition as applicable.

If the parameter is left undefined then the existing behaviour of generating static network config based on the infra env type will be used.

This corresponds to the truth table shown below.

-----------------------------------------------------------------------------------------------
infraenv type | overrideImageType          | discovery ignition contains static network config?
-----------------------------------------------------------------------------------------------
full iso      | nil                        | yes
full iso      | full iso                   | yes
full iso      | minimal iso                | no
minimal iso   | nil                        | no
minimal iso   | minimal iso                | no
minimal iso   | full iso                   | yes

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?

This has not been tested yet, this is a draft PR, I will fill this out later.

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @nmagnezi 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -1014,7 +1014,7 @@ func (b *bareMetalInventory) createAndUploadNewImage(ctx context.Context, log lo
 }
 
 func (b *bareMetalInventory) getIgnitionConfigForLogging(ctx context.Context, infraEnv *common.InfraEnv, log logrus.FieldLogger, imageType models.ImageType) string {
-	ignitionConfigForLogging, _ := b.IgnitionBuilder.FormatDiscoveryIgnitionFile(ctx, infraEnv, b.IgnitionConfig, true, b.authHandler.AuthType())
+	ignitionConfigForLogging, _ := b.IgnitionBuilder.FormatDiscoveryIgnitionFile(ctx, infraEnv, b.IgnitionConfig, true, b.authHandler.AuthType(), string(imageType))
 	log.Infof("Generated infra env <%s> image with ignition config", infraEnv.ID)
 	log.Debugf("Ignition for infra env <%s>: %s", infraEnv.ID, ignitionConfigForLogging)
 	var msgDetails []string
@@ -3360,12 +3360,6 @@ func (b *bareMetalInventory) DownloadMinimalInitrd(ctx context.Context, params i
 		return common.GenerateErrorResponder(err)
 	}
 
-	if common.ImageTypeValue(infraEnv.Type) != models.ImageTypeMinimalIso {
-		err = fmt.Errorf("Only %v image type supported but %v specified.", models.ImageTypeMinimalIso, infraEnv.Type)
-		log.WithError(err)
-		return common.NewApiError(http.StatusConflict, err)
-	}
-
 	var netFiles []staticnetworkconfig.StaticNetworkConfigData
 	if infraEnv.StaticNetworkConfig != "" {
 		netFiles, err = b.staticNetworkConfig.GenerateStaticNetworkConfigData(ctx, infraEnv.StaticNetworkConfig)
@@ -5299,7 +5293,8 @@ func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params
 	var content, filename string
 	switch params.FileName {
 	case "discovery.ign":
-		content, err = b.IgnitionBuilder.FormatDiscoveryIgnitionFile(ctx, infraEnv, b.IgnitionConfig, false, b.authHandler.AuthType())
+		discoveryIsoType := swag.StringValue(params.DiscoveryIsoType)
+		content, err = b.IgnitionBuilder.FormatDiscoveryIgnitionFile(ctx, infraEnv, b.IgnitionConfig, false, b.authHandler.AuthType(), discoveryIsoType)
 		if err != nil {
 			b.log.WithError(err).Error("Failed to format ignition config")
 			return common.GenerateErrorResponder(err)
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -48,7 +48,6 @@ import (
 	"github.com/openshift/assisted-service/internal/ignition"
 	"github.com/openshift/assisted-service/internal/infraenv"
 	installcfg "github.com/openshift/assisted-service/internal/installcfg/builder"
-	"github.com/openshift/assisted-service/internal/isoeditor"
 	"github.com/openshift/assisted-service/internal/metrics"
 	"github.com/openshift/assisted-service/internal/operators"
 	"github.com/openshift/assisted-service/internal/provider/registry"
@@ -157,7 +156,7 @@ func mockInfraEnvRegisterSuccess() {
 	mockVersions.EXPECT().GetOsImage(gomock.Any(), gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).AnyTimes()
 	mockStaticNetworkConfig.EXPECT().FormatStaticNetworkConfigForDB(gomock.Any()).Return("", nil).Times(1)
 	mockSecretValidator.EXPECT().ValidatePullSecret(gomock.Any(), gomock.Any()).Return(nil).Times(1)
-	mockIgnitionBuilder.EXPECT().FormatDiscoveryIgnitionFile(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(discovery_ignition_3_1, nil).Times(1)
+	mockIgnitionBuilder.EXPECT().FormatDiscoveryIgnitionFile(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(discovery_ignition_3_1, nil).Times(1)
 	mockEvents.EXPECT().SendInfraEnvEvent(gomock.Any(), eventstest.NewEventMatcher(
 		eventstest.WithNameMatcher(eventgen.ImageInfoUpdatedEventName))).AnyTimes()
 }
@@ -165,7 +164,7 @@ func mockInfraEnvRegisterSuccess() {
 func mockInfraEnvUpdateSuccess() {
 	mockVersions.EXPECT().GetOsImageOrLatest(gomock.Any(), gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).AnyTimes()
 	mockVersions.EXPECT().GetOsImage(gomock.Any(), gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).AnyTimes()
-	mockIgnitionBuilder.EXPECT().FormatDiscoveryIgnitionFile(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(discovery_ignition_3_1, nil).Times(1)
+	mockIgnitionBuilder.EXPECT().FormatDiscoveryIgnitionFile(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(discovery_ignition_3_1, nil).Times(1)
 	mockEvents.EXPECT().SendInfraEnvEvent(gomock.Any(), eventstest.NewEventMatcher(
 		eventstest.WithNameMatcher(eventgen.ImageInfoUpdatedEventName))).AnyTimes()
 }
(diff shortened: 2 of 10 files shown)


swagger.yaml
```diff
@@ -1654,7 +1654,8 @@ paths:
           description: Overrides the ISO type for the disovery ignition, either 'full-iso' or 'minimal-iso'.
           required: false
           type: string
-          enum: ['full-iso', 'minimal-iso']
+          enum: ['use-infraenv-type', 'full-iso', 'minimal-iso']
```
carbonin, line 1657: `use-infraenv-type` shouldn't be necessary here. If it isn't specified in the url we should use the type from the infra-env.
That also ensures backwards compatibility.

internal/ignition/ignition.go
```diff
@@ -1359,7 +1359,8 @@ func (ib *ignitionBuilder) FormatDiscoveryIgnitionFile(ctx context.Context, infr
 		ignitionParams["ServiceIPs"] = dataurl.EncodeBytes([]byte(GetServiceIPHostnames(cfg.ServiceIPs)))
 	}
 
-	if infraEnv.StaticNetworkConfig != "" && common.ImageTypeValue(infraEnv.Type) == models.ImageTypeFullIso {
+	shouldAddStaticNetworkConfig := string(overrideDiscoveryISOType) == string(DiscoveryISOTypeFullIso) || (string(overrideDiscoveryISOType) == string(DiscoveryISOTypeUseInfraEnv) && infraEnv.StaticNetworkConfig != "" && common.ImageTypeValue(infraEnv.Type) == models.ImageTypeFullIso)
+	if shouldAddStaticNetworkConfig {
```
carbonin, line 1363: I think this would be more clear as multiple steps.

```
isoType := overrideDiscoveryISOType
if isoType == "" {
  isoType = infraEnv.Type
}
if infraEnv.StaticNetworkConfig != "" && common.ImageTypeValue(isoType) == models.ImageTypeFullIso {
...
```

This more clearly shows that we're prioritizing the function parameter over the infra env and separates out the check for actually having static networking.

It's less efficient, but I think the readability is worth it in this case.

internal/bminventory/inventory.go
```diff
@@ -5202,7 +5202,7 @@ func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params
 	var content, filename string
 	switch params.FileName {
 	case "discovery.ign":
-		content, err = b.IgnitionBuilder.FormatDiscoveryIgnitionFile(ctx, infraEnv, b.IgnitionConfig, false, b.authHandler.AuthType(), ignition.DiscoveryISOTypeUseInfraEnv)
+		content, err = b.IgnitionBuilder.FormatDiscoveryIgnitionFile(ctx, infraEnv, b.IgnitionConfig, false, b.authHandler.AuthType(), ignition.DiscoveryISOType(*params.DiscoveryIsoType))
```
carbonin, line 5205: If this can be nil we should be checking for that.
I don't think the boilerplate generated code actually fills in defaults for us.

carbonin (changes requested):

internal/ignition/ignition.go
```diff
@@ -56,6 +56,14 @@ import (
 	k8syaml "sigs.k8s.io/yaml"
 )
 
+type DiscoveryISOType string
+
+const (
+	DiscoveryISOTypeUseInfraEnv DiscoveryISOType = "use-infraenv-type"
```
carbonin, line 62: If we get rid of this then we can just use the existing models.

swagger.yaml
carbonin: Actually I think it would be best if you could do something like this here:

```yaml
- in: query
  name: discovery_iso_type
  description: ...
  required: false
  schema:
    $ref: '#/definitions/image_type'
```

internal/bminventory/inventory_test.go
```diff
@@ -7296,7 +7278,7 @@ var _ = Describe("DownloadMinimalInitrd", func() {
 			InfraEnv: models.InfraEnv{
 				ID:            &clusterID,
 				PullSecretSet: true,
-				Type:          common.ImageTypePtr(models.ImageTypeMinimalIso),
+				Type:          common.ImageTypePtr(models.ImageTypeFullIso),
```
carbonin, line 7281: I think it would be good to have a sanity test case for both types.
I think it would be fine to put the validation code here into a helper function then have an example for each iso type.

internal/ignition/ignition_test.go
```diff
@@ -1270,7 +1269,26 @@ var _ = Describe("IgnitionBuilder", func() {
...
 			mockMirrorRegistriesConfigBuilder.EXPECT().IsMirrorRegistriesConfigured().Return(false).Times(1)
-			text, err := builder.FormatDiscoveryIgnitionFile(context.Background(), &infraEnv, IgnitionConfig{}, false, auth.TypeRHSSO)
+			text, err := builder.FormatDiscoveryIgnitionFile(context.Background(), &infraEnv, IgnitionConfig{}, false, auth.TypeRHSSO, "")
+			Expect(err).NotTo(HaveOccurred())
+			config, report, err := config_31.Parse([]byte(text))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(report.IsFatal()).To(BeFalse())
+			count := 0
+			for _, f := range config.Storage.Files {
+				if strings.HasSuffix(f.Path, "nmconnection") || strings.HasSuffix(f.Path, "mac_interface.ini") || strings.HasSuffix(f.Path, "02-hostname-mode.conf") {
```
carbonin, line 1279: Why do two of these tests have this bit about a hostname file, but the last doesn't?

If they can all be the same maybe this could be a helper to return the count of network files in an ignition config and we could deduplicate this logic.

carbonin:
Looking good, just a few comments on the tests then this should be good to run through CI (undraft? not sure how to say that :sweat_smile: )

internal/bminventory/inventory.go
```diff
@@ -5299,7 +5293,11 @@ func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params
 	var content, filename string
 	switch params.FileName {
 	case "discovery.ign":
-		content, err = b.IgnitionBuilder.FormatDiscoveryIgnitionFile(ctx, infraEnv, b.IgnitionConfig, false, b.authHandler.AuthType())
+		discoveryIsoType := ""
+		if params.DiscoveryIsoType != nil {
+			discoveryIsoType = *params.DiscoveryIsoType
+		}
```
nmagnezi, line 5299:
```suggestion
		discoveryIsoType := swag.StringValue(params.DiscoveryIsoType)
```

internal/ignition/ignition_test.go
```diff
@@ -1017,7 +1016,7 @@ var _ = Describe("IgnitionBuilder", func() {
 
 	It("auth_disabled_no_pull_secret_token", func() {
 		mockMirrorRegistriesConfigBuilder.EXPECT().IsMirrorRegistriesConfigured().Return(false).Times(1)
-		text, err := builder.FormatDiscoveryIgnitionFile(context.Background(), &infraEnv, IgnitionConfig{}, false, auth.TypeNone)
+		text, err := builder.FormatDiscoveryIgnitionFile(context.Background(), &infraEnv, IgnitionConfig{}, false, auth.TypeNone, "")
```
nmagnezi, line 1019: @paul-maidment maybe I'm missing something here; Why are these test changes?

nmagnezi (changes requested):

internal/ignition/ignition_test.go
paul-maidment: FormatDiscoveryIgnitionFile now expects the parameter discovery_iso_type which has been added to the signature, the test needed to be updated to match the signature change.

nmagnezi: Allow me to better phrase what I don't understand here: Why do all tests pass `""` as a value for `overrideDiscoveryISOType` instead of some actual values?

paul-maidment: discovery_iso_type is not mandatory in the API and when it is not present (I think) it will be interpreted as an empty string, I did originally accept a pointer but I was discouraged from this in other feedback.

nmagnezi:
@carbonin I had a discussion with Paul about ignition tests. Noticed that some of the added tests were based on existing tests such as:
https://github.com/openshift/assisted-service/blob/dedbd0275cc62ae9d850df9f0cb0f2c3f76ebf5f/internal/ignition/ignition_test.go#L1268-L1285

Something does not seem right here. Why would a minimal ISO not include static network config? Isn't that the basis of what we do in KubeAPI (which defaults to minimal-iso).

carbonin:
>Something does not seem right here. Why would a minimal ISO not include static network config? Isn't that the basis of what we do in KubeAPI (which defaults to minimal-iso).

It doesn't include static networking in the ignition because it's included in the initrd and is carried forward through boot to install.

internal/ignition/ignition_test.go
carbonin: I just looked. It seems like references to this file were removed in 5ccdf83ab301eb668593296d53c02c647deac3b0 we can just get rid of these conditions that are looking for it.

internal/ignition/ignition_test.go
```diff
@@ -1281,6 +1299,25 @@ var _ = Describe("IgnitionBuilder", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			config, report, err := config_31.Parse([]byte(text))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(report.IsFatal()).To(BeFalse())
+			count := 0
+			for _, f := range config.Storage.Files {
+				if strings.HasSuffix(f.Path, "nmconnection") || strings.HasSuffix(f.Path, "mac_interface.ini") {
+					count += 1
+				}
+			}
```
carbonin, line 1320: You could put this in a helper function so that we don't have it in multiple places.

carbonin (commented):
Some comments on unit tests.

Can we also add some of this to subsystem?
It looks like we have a good place to add this stuff in `subsystem/infra_env_tests.go`

internal/bminventory/inventory_test.go
```diff
@@ -7623,6 +7568,36 @@ var _ = Describe("DownloadMinimalInitrd", func() {
...
+
+	It("returns no content without network customizations", func() {
+		createInfraEnv(models.ImageTypeFullIso)
+		params := installer.DownloadMinimalInitrdParams{InfraEnvID: id}
+		response := bm.DownloadMinimalInitrd(ctx, params)
+		Expect(response).Should(BeAssignableToTypeOf(&installer.DownloadMinimalInitrdNoContent{}))
+	})
+
+	FIt("returns legit archive when full ISO is requested", func() {
+		validateArchive(createInfraEnv(models.ImageTypeFullIso), proxy())
```
carbonin, line 7596: If we're always using the same proxy definition, why pass it separately?

internal/bminventory/inventory_test.go
```diff
@@ -7506,96 +7506,56 @@ var _ = Describe("KubeConfig download", func() {
...
-		dbName    string
+		bm     *bareMetalInventory
+		cfg    Config
+		db     *gorm.DB
+		ctx    = context.Background()
+		id     strfmt.UUID
+		dbName string
+		httpProxy = "http://10.10.1.1:3128"
+		httpsProxy = "https://10.10.1.1:3128"
+		noProxy = "quay.io"
```
carbonin, line 7517: Lack of alignment here might fail the linter. I'd be sure to `skipper make validate` before making this a non-draft.

internal/bminventory/inventory_test.go
```diff
@@ -7506,96 +7506,56 @@ var _ = Describe("KubeConfig download", func() {
...
 			},
 			PullSecret: "{\"auths\":{\"cloud.openshift.com\":{\"auth\":\"dG9rZW46dGVzdAo=\",\"email\":\"coyote@acme.com\"}}}",
 		}
+		return result
+	}
+
+	storeInfraEnv := func(infraEnv common.InfraEnv) {
 		err := db.Create(&infraEnv).Error
 		Expect(err).ShouldNot(HaveOccurred())
+	}
```
carbonin, line 7555: This function really is unnecessary. This could be one line:

```go
Expect(db.Create(&infraEnv).Error).NotTo(HaveOccurred())
```

internal/bminventory/inventory_test.go
```diff
@@ -7620,9 +7580,39 @@ var _ = Describe("DownloadMinimalInitrd", func() {
...
+
+	It("returns no content without network customizations", func() {
+		infraEnv := createInfraEnv(models.ImageTypeFullIso)
+		storeInfraEnv(infraEnv)
+		params := installer.DownloadMinimalInitrdParams{InfraEnvID: id}
+		response := bm.DownloadMinimalInitrd(ctx, params)
+		Expect(response).Should(BeAssignableToTypeOf(&installer.DownloadMinimalInitrdNoContent{}))
+	})
+
+	It("returns legit archive when full ISO is requested", func() {
```
carbonin, line 7604: I think I asked about this before, but while we're here can we clarify what "legit" is?

internal/bminventory/inventory_test.go
```diff
@@ -6271,7 +6271,7 @@ var _ = Describe("infraEnvs", func() {
 					common.FormatStaticConfigHostYAML("0200003ef73c", "02000048ba38", "192.168.126.40", "192.168.141.40", "192.168.126.1", map2),
 					common.FormatStaticConfigHostYAML("0200003ef75c", "02000048ba58", "192.168.126.42", "192.168.141.42", "192.168.126.1", map3),
 				}
-				mockStaticNetworkConfig.EXPECT().ValidateStaticConfigParams(staticNetworkConfig).Return(nil).Times(1)
+				mockStaticNetworkConfig.EXPECT().ValidateStaticConfigParams(gomock.Any()).Return(nil).Times(1)
```
carbonin, line 6274: Why is this changing?

carbonin (changes requested):
Looks like you have a sleep in there still.

Also one of the commits looks like a revert?
Maybe best to squash at this point. I don't think the separate commits are helpful right now.

carbonin (approved):

Issue #4310, opened 2022-08-25
Incorrect modes in the discovery.ign template


gmarcy:
Ignition storage file modes are represented in decimal while their natural representation in Linux is typically octal, i.e. 755, 644, etc.

These two modes are incorrect in the discovery.ign template:

      "path": "/usr/local/bin/agent-fix-bz1964591",
      "mode": 755,

and

      "path": "/usr/local/bin/okd-binaries.sh",
      "mode": 755,

both should be changed to:

        "mode": 493,

PR #4311, opened 2022-08-25
Update discovery.ign to fix incorrect file modes

gmarcy:
Fix the modes of scripts to their proper decimal representation.

This addresses issue #4310

### DIFF
--- a/internal/ignition/templates/discovery.ign
+++ b/internal/ignition/templates/discovery.ign
@@ -44,7 +44,7 @@
     "files": [{
       "overwrite": true,
       "path": "/usr/local/bin/agent-fix-bz1964591",
-      "mode": 755,
+      "mode": 493,
       "user": {
           "name": "root"
       },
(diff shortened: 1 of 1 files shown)


internal/ignition/templates/discovery.ign
```diff
@@ -44,7 +44,7 @@
     "files": [{
       "overwrite": true,
       "path": "/usr/local/bin/agent-fix-bz1964591",
-      "mode": 755,
+      "mode": 493,
```
jhernand, line 47: As this is a Go template you may want to use `{{ 0755 }}`, that is easier to read and change, and translates into `493`.

carbonin (approved):
/ok-to-test

We use the decimal representation in other places in this file so I'm fine approving this without using the template features.

internal/ignition/templates/discovery.ign
gmarcy: I was going for the smallest change that addressed the specific problem but agree there may be an issue with a broader scope that would address more consistent use of accepted best practices.

PR #4312, opened 2022-08-25
MGMT-8890: Add OS image validator to openshift-versions API call

mkowalski:
With this commit we are changing behaviour of the "openshift-versions"
API call so that it returns only versions and architectures if there is
a matching release image as well as RHCOS image.

This is in order to cover a scenario when a multi-arch release image is
used by the service, but available OS images do not contain every
architecture that is provided by the release. For such a scenario we
only want to mark availability for those combinations that have a
coverage in both release as well as RHCOS image.

Closes: [MGMT-11780](https://issues.redhat.com//browse/MGMT-11780)

/cc @danielerez

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -109,6 +109,13 @@ func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params o
 				arch = common.DefaultCPUArchitecture
 			}
 
+			// In order to mark a specific version and architecture as supported we do not
+			// only need to have an available release image, but we need RHCOS image as well.
+			if _, err := h.GetOsImage(key, arch); err != nil {
+				h.log.Debugf("Marking architecture %s for version %s as not available because no matching OS image found", arch, key)
+				continue
+			}
+
 			openshiftVersion, exists := openshiftVersions[key]
 			if !exists {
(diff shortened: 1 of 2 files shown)


internal/versions/versions.go
```diff
@@ -109,6 +109,12 @@ func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params o
 				arch = common.DefaultCPUArchitecture
 			}
 
+			// In order to mark a specific version and architecture as supported we do not
+			// only need to have an available release image, but we need RHCOS image as well.
+			if _, err := h.GetOsImage(key, arch); err != nil {
+				continue
```
danielerez, line 115: Probably should log the err here as a warning. It would help users identify the issue in case of a missing OS image.

mkowalski: Done. As debug because we definitely don't want to see those in SaaS. For on-prem users we anyway troubleshoot with debug

mkowalski:
I have already had subsystem tests succeeding for this commit so I take the authority to override given that the failure is related to the out-of-project dependency

/override ci/prow/edge-subsystem-aws
/override ci/prow/edge-subsystem-kubeapi-aws

PR #4324, opened 2022-08-29
MGMT-8890: Add 4.11 single-arch release image to defaults

mkowalski:


data/default_release_images.json
```diff
@@ -35,6 +35,25 @@
...
+      "version": "4.11.0",
+      "default": true
+    },
+    {
+      "openshift_version": "4.11",
+      "cpu_architecture": "arm64",
+      "cpu_architectures": [
+          "arm64"
+      ],
+      "url": "quay.io/openshift-release-dev/ocp-release:4.11.0-arm64",
```
osherdp, line 54:
```suggestion
      "url": "quay.io/openshift-release-dev/ocp-release:4.11.0-aarch64",
```

osherdp (changes requested):

data/default_release_images.json
mkowalski: Fixed, thanks!

osherdp (approved):

osherdp:
/override ci/prow/edge-e2e-metal-assisted-day2-single-node
ignore, @eliorerz is fixing it

osherdp:
/hold
Failure seems somewhat related to the changes of this PR

mkowalski:
/unhold

Indeed that was the case; we didn't foresee shipping single-arch next to multi-arch so the tests were opinionated towards some logic implemented for "only multi-arch" scenario. As the files used by subsystem are directly coming from this repo and not from the separate test files, the test need to be adjusted. Didn't find an easy way to provide a separate configuration to the test so as long as our tests and prod consume the same file, one has to be opinionated ¯\_(ツ)_/¯

osherdp (approved):

mkowalski:

Seems to almost always fail, let's just retry

PR #4336, opened 2022-08-31
[WIP] MGMT-11739: added KubeconfigSecretRef property to ASC

danielerez:
In order to communicate with an external control-plane (e.g. hypershift HostedCluster), the service should be supplied with the relevant kubeconfig. I.e. When applying the service deployment in AgentServiceConfigReconciler, a path for a custom kubeconfig should be supplied in an env var.

Hence, introduced the following changes, added KubeconfigSecretRef property to AgentServiceConfigSpec.
This property is a reference for a secret that contains the kubeconfig for an external cluster.

The logic for adding relevant Volume, VolumeMount and EnvVar for the service deployment will be added as part of: https://github.com/openshift/assisted-service/pull/4322

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @filanov 
/cc @carbonin 
/cc @slaviered 

## Checklist

## Reviewers Checklist


api/v1beta1/agentserviceconfig_types.go
```diff
@@ -102,6 +102,12 @@ type AgentServiceConfigSpec struct {
 	// +kubebuilder:validation:default:=disabled
 	// +optional
 	IPXEHTTPRoute string `json:"iPXEHTTPRoute,omitempty"`
+
+	// KubeconfigSecretRef is the reference to the Secret containing a kubeconfig of an external
+	// control plane (e.g. hypershift HostCluster).
```
carbonin, line 107: I think we should add some more words about what happens when you add this.

It's not clear that this changes where the controllers are looking to do reconcile.

carbonin (commented):
cc @avishayt @mhrivnak for API sanity check

api/v1beta1/agentserviceconfig_types.go
danielerez: right, added some explanation.

danielerez:
/cc @avishayt 
/cc @mhrivnak 
Can you please take a look?

mhrivnak (commented):
We should still have a follow-up about whether this is better as an addon to ASC or makes more sense as a separate CRD. It'll help to think through any other behavioral and/or API changes that could be related. For example, will status be any different? Might upgrades work differently at any point? etc.

I also still struggle to see a reasonable path toward making ASC namespaced.

PR #4341, opened 2022-08-31
NO-ISSUE: Specify Hive channel explicitly for RHACM 2.4

mkowalski:
/cc @carbonin

### DIFF
--- a/deploy/operator/setup_hive.sh
+++ b/deploy/operator/setup_hive.sh
@@ -50,6 +50,7 @@ spec:
   name: hive-operator
   source: community-operators
   sourceNamespace: openshift-marketplace
+  channel: ocm-2.4
 EOCR
 
     wait_for_operator "hive-operator" "${HIVE_NAMESPACE}"


mkowalski:
/retest-required

Failures in non-operator tests so not relevant

carbonin:
Looks like both runs of those tests failed.
They haven't been run in some time.
Is there any chance the failure is an actual problem even if it isn't an issue with this patch?

mkowalski:
Yes it's perfectly possible. Is ACM 2.4 even supported?

carbonin:
I'm not sure, maybe @oourfali or @filanov can comment on if this is worth doing for this branch.

oourfali:
what are we trying to solve here?

mkowalski:
> what are we trying to solve here?

During discussions about some bug in Hive we have discovered that for our RHACM 2.x CI we don't use Hive version that is shipped with RHACM 2.x, but always the latest. Therefore, our 2.x CI is not testing Hive from this release. Therefore, if there is a bug in Hive that affects Assisted installations (and e.g. has been fixed in latest but never backported), we are just letting it go through.

Based on all of this, we have decided that our CIs for 2.x should use Hive shipped with 2.x; this PR implements such a change for RHACM 2.4 branch, but the question is whether we still support 2.4

filanov:
2.4 still supported and get new releases

mkowalski:
/retest-required

Failures in non-operator tests so not relevant

filanov (approved):

PR #4331, opened 2022-08-30
NO-ISSUE: Fix condition to get kubeconfig secret name from CR

javipolo:
This condition was reverted, it should run getClusterDeploymentAdminKubeConfigSecretName if clusterDeployment CR was retrieved successfully

Identified by @carbonin -> https://github.com/openshift/assisted-service/pull/3411/files#r955343608

## List all the issues related to this PR

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [x] No tests needed

## Assignees

/cc @carbonin

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] Reviewers have been listed
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #4348, opened 2022-09-02
MGMT-10968: Use a daemon set for the spoke cluster agent during node reclaim

carbonin:
Testing reclaim with CAPI revealed an issue with the ordering in which CAPI cordons and drains the node and when we attempt to run our pod and reboot into discovery.

Originally the plan was to use a "pre-drain" hook where we would do our work then CAPI would drain the node. During testing I realized that this wouldn't work because "our work" involved rebooting the node and that wouldn't actually give CAPI time to drain. In practice CAPI actually stalled after we rebooted and never removed the Node, Machine, or AgentMachine because it couldn't drain.

This meant that we needed to run our agent _after_ the node was drained using a "pre-terminate" hook. By extension, this meant finding a way to run a pod on a node that had been cordoned and drained. Based on the [drain documentation](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#drain) the only real option was to use a DaemonSet (the other pod type that doesn't get removed is a static pod which we couldn't use for obvious reasons).

Moving to a DaemonSet also required explicitly assigning a service account bound to the privileged scc.

Another change to note here is that the DaemonSet is marked as owed by the node we're trying to remove. This is done to ensure that the DaemonSet doesn't persist past when we finish the reclaim work. This would be a problem if the node (or another one with the same hostname) were ever re-added to the cluster. In that case the pod would start again and run for the entire time the node existed. This owner reference prevents this and also saves us from having to do some more manual cleanup from the service.

https://issues.redhat.com/browse/MGMT-10968
Related CAPI provider PR: https://github.com/openshift/cluster-api-provider-agent/pull/54
Requires https://github.com/openshift/assisted-service/pull/4331 to work properly

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

This PR was tested in conjunction with https://github.com/openshift/cluster-api-provider-agent/pull/54 to ensure nodes were removed from a hypershift cluster using CAPI and added back to the infraEnv.

## Assignees

/cc @CrystalChun 
/cc @mhrivnak 
/cc @ori-amizur 

## Checklist

## Reviewers Checklist


CrystalChun:
Looks good to me 👍 will let Michael and Ori take a look

carbonin:
I have another branch I'm working on to handle assigning the privileged scc in a better way.
Since this hasn't had much review yet I'm going to mark it as WIP and include those changes here.

internal/controller/controllers/agent_reclaimer.go
```diff
@@ -48,16 +51,73 @@ func newAgentReclaimer(hostFSMountDir string) (*agentReclaimer, error) {
...
-		if !k8serrors.IsNotFound(err) {
-			return errors.Wrapf(err, "failed to get namespace %s", spokeReclaimNamespaceName)
-		}
-		if err := c.Create(ctx, ns); err != nil {
-			return errors.Wrapf(err, "failed to create namespace %s", spokeReclaimNamespaceName)
+	mutate := func() error {
+		if ns.Labels == nil {
+			ns.Labels = make(map[string]string)
 		}
+		ns.Labels["pod-security.kubernetes.io/enforce"] = "privileged"
```
danielerez, line 58: Just wonder if it worth adding some comment here about why we need to change the security mode. I mean, as it affects the entire namespace it could be confusing if missing the label set here.

internal/controller/controllers/agent_reclaimer.go
```diff
@@ -74,68 +134,45 @@ func (r *agentReclaimer) ensureSpokeAgentSecret(ctx context.Context, c client.Cl
...
+		secret.Type = corev1.SecretTypeOpaque
 		secret.Data = map[string][]byte{"auth-token": []byte(authToken)}
-		if err := c.Create(ctx, secret); err != nil {
-			return errors.Wrapf(err, "failed to create secret %s", spokeReclaimNamespaceName)
-		}
+		return nil
 	}
-
-	return nil
+	_, err := controllerutil.CreateOrUpdate(ctx, c, secret, mutate)
```
danielerez, line 148: We don't want to have a log if the result is not none (unchanged)?

internal/spoke_k8s_client/spoke_k8s_client.go
```diff
@@ -160,11 +161,11 @@ func (c *spokeK8sClient) CreateSubjectAccessReview(subjectAccessReview *authoriz
 func (c *spokeK8sClient) IsActionPermitted(verb string, resource string) (bool, error) {
 	sar := authorizationv1.SelfSubjectAccessReview{
 		Spec: authorizationv1.SelfSubjectAccessReviewSpec{
-            ResourceAttributes: &authorizationv1.ResourceAttributes{
```
danielerez, line 163: yikes, how was it missed :/

internal/controller/controllers/agent_reclaimer.go
carbonin: :shrug: I added this in the commit message:

```
These will ensure that we're able to run privileged pods in this
namespace when Pod Security Standards are adopted in OCP 4.12
```

Also this is not strictly necessary with OpenShift. There should be some label syncer that labels the namespace when we have some SA assigned with privileged permissions, but I think this is the recommended way to go about it.

internal/controller/controllers/agent_reclaimer.go
carbonin: :shrug: We could but I'm not sure it matters.
In that case we're doing what we're supposed to do.

But if you would prefer a log I can add one.

internal/controller/controllers/agent_reclaimer.go
danielerez: ok, should be good enough I guess to rely on the commit message for reference, or just add this comment about OCP 4.12.

internal/controller/controllers/agent_reclaimer.go
danielerez: Not very crucial, but this is the behavior on other controllers, so thought it worth adding it for consistency.

internal/controller/controllers/agent_reclaimer.go
```diff
@@ -178,35 +215,58 @@ func (r *agentReclaimer) createNextStepRunnerPod(ctx context.Context, c client.C
...
-					ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{
-						Key: "auth-token",
-						LocalObjectReference: corev1.LocalObjectReference{
-							Name: spokeReclaimSecretName(infraEnvID),
-						},
-					}},
-				}},
-			}},
+		Spec: appsv1.DaemonSetSpec{
+			Selector: &metav1.LabelSelector{MatchLabels: labels},
```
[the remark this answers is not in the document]
danielerez: Ok makes sense. Same for the Template in next line? If no need to update them then scratch that...

internal/spoke_k8s_client/spoke_k8s_client.go
danielerez: Probably. golangci-lint should have caught it though :(

internal/controller/controllers/agent_reclaimer.go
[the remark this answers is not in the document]
carbonin: Yup, these really shouldn't ever be updated.

PR #4350, opened 2022-09-02
MGMT-11888: Pass usable NIC's MAC address to install-config

mkowalski:
Currently in order to fill the BootMACAddress in install-config.yaml we
are picking any NIC that has non-zero MAC address and use this value.
It's been proven that this logic is prone to fail for servers that have
NICs with duplicated MACs even though those NICs are disabled and not
used.

This PR changes the logic so that we will only pick MAC address of a NIC
if it has an IP address assigned. This should remove changes of hitting
the scenario described above, as in a regular network topology there
should be no duplicated MACs in the network.

Closes: MGMT-11888

/cc @CrystalChun 
/cc @carbonin


carbonin:
I saw the original PR changed some subsystem tests.
Do you think we could reproduce this issue there and add a test for it?

mkowalski:
> I saw the original PR changed some subsystem tests. Do you think we could reproduce this issue there and add a test for it?

As discussed, subsystem tests should never hit this code in the negative scenario. This is because we should never ever try to install a host in a cluster, if the host does not belong to the machine network. If that was happening, it would mean that for some reasons we are proceeding with installation of the host that cannot communicate with rest of the cluster.

Instead of that I enhanced unit tests to cover what happens here.

internal/provider/registry/registry_test.go
```diff
@@ -302,6 +302,36 @@ var _ = Describe("Test AddPlatformToInstallConfig", func() {
 			Expect(cfg.Platform.Baremetal.Hosts[1].Name).Should(Equal("hostname1"))
 			Expect(cfg.Platform.Baremetal.Hosts[2].Name).Should(Equal("hostname2"))
 		})
+		It("fails for host without interface from machine networks", func() {
```
tsorya, line 305: one success case were we validate that we took right mac?

mkowalski: Sure, just added a 2nd interface to all the hosts and added a test to confirm that we pick MAC of the interface where address is assigned

internal/provider/baremetal/installConfig.go
```diff
@@ -42,14 +43,34 @@ func (p baremetalProvider) AddPlatformToInstallConfig(
...
+			// We are looking for the NIC that matches the first Machine Network configured
+			// for the cluster. This is to ensure that BootMACAddress belongs to the NIC that
+			// is really used and not to any fake interface even if this interface has IPs.
+			match := false
+			ipAddressses := append(iface.IPV4Addresses, iface.IPV6Addresses...)
+			for _, ip := range ipAddressses {
+				overlap, _ := network.NetworksOverlap(ip, string(cluster.MachineNetworks[0].Cidr))
+				if overlap {
+					match = true
+				}
```
CrystalChun, line 62: Tiny suggestion to reuse the `match` variable and to break out of this loop early if a match is found
```suggestion
				if match, _ = network.NetworksOverlap(ip, string(cluster.MachineNetworks[0].Cidr)); match {
					// maybe even set this here?
					// hosts[yamlHostIdx].BootMACAddress = iface.MacAddress
					break
				}
```

CrystalChun (commented):
Small suggestion, otherwise looks good! Will let others re-review

internal/provider/baremetal/installConfig.go
mkowalski: That's a very good suggestion, thanks! Looks like I was trying to over-engineer and completely missed something as trivial as this

internal/installcfg/builder/builder_test.go
```diff
@@ -154,7 +154,6 @@ var _ = Describe("installcfg", func() {
 		cluster.HTTPProxy = proxyURL
 		cluster.HTTPSProxy = proxyURL
 		cluster.NoProxy = "no-proxy.com"
-		cluster.MachineNetworks = []*models.MachineNetwork{}
```
mkowalski, line 157: Cluster with empty machine networks is illegal, not sure why this ever worked

internal/installcfg/builder/builder_test.go
```diff
@@ -56,37 +56,37 @@ var _ = Describe("installcfg", func() {
 			ClusterNetworks:        []*models.ClusterNetwork{{Cidr: "1.1.1.0/24"}},
 			ServiceNetworks:        []*models.ServiceNetwork{{Cidr: "2.2.2.0/24"}},
 			MachineNetworks:        []*models.MachineNetwork{{Cidr: "1.2.3.0/24"}},
-			APIVip:                 "102.345.34.34",
-			IngressVip:             "376.5.56.6",
```
mkowalski, line 60: Those are not even IP addresses, changed to a proper syntax matching the machine network

internal/installcfg/builder/builder_test.go
```diff
@@ -56,37 +56,37 @@ var _ = Describe("installcfg", func() {
...
+			APIVip:                 "1.2.3.11",
+			IngressVip:             "1.2.3.12",
 			InstallConfigOverrides: `{"fips":true}`,
 			ImageInfo:              &models.ImageInfo{},
 			Platform:               &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeBaremetal)},
 			NetworkType:            swag.String("OpenShiftSDN"),
 		}}
 		id := strfmt.UUID(uuid.New().String())
-		// By default all the hosts in the cluster are IPv4-only
+		// By default all the hosts in the cluster are dual-stack
```
mkowalski, line 67: The tests below are relying on the fact that hosts have interface in IPv6 subnet, so host with IPv4-only address cannot work in such a cluster

internal/installcfg/builder/builder_test.go
```diff
@@ -609,10 +609,12 @@ func getInventoryStr(hostname, bootMode string, ipv4 bool, ipv6 bool) string {
 		},
 	}
 	if ipv4 {
-		inventory.Interfaces[0].IPV4Addresses = []string{"10.35.20.10/24"}
```
mkowalski, line 612: All the tests are using `1.2.3.0/24` as machine network, so we need to have a hosts' interface in this subnet

mkowalski:
@tsorya need re-LGTM after fixing tests

mkowalski:
@tsorya need re-LGTM after fixing tests

PR #4353, opened 2022-09-04
[WIP] MGMT-11751: create agent-install CRDs on external cluster

danielerez:
When KubeconfigSecretRef is specific in ASC, we should create agent-install CRDs on the cluster associated with the kubeconfig. This is required as the assisted-service is deployed on that external cluster, and kube-api is using these CRDs.

Since the relevant CRDs are already deployed as part of the infrastructure operator installation, we can use them when deploying CRDs on the external cluster. I.e. the flow is as follows:
- List CRDs using the in-cluster config and filter by 'assisted-service-operator.namespace' filter; this should retrieve the agent-install CRDs.
- Create the retrieved CRDs using the external cluster config.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @filanov 
/cc @slaviered 

## Checklist

## Reviewers Checklist


config/rbac/role.yaml
```diff
@@ -181,6 +181,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
```
carbonin, line 187: This rbac will get applied to the management cluster when we install the operator, but we don't really need these permissions there. Or at least we don't need `create` there, right?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -2288,3 +2301,43 @@ func (r *AgentServiceConfigReconciler) getImageService(ctx context.Context, log
 	}
 	return imageServiceURL
 }
+
+func (r *AgentServiceConfigReconciler) deployAgentInstallCRDs(secret *corev1.Secret, namespace string, log logrus.FieldLogger) error {
+	// Get in-cluster client
+	kubeClient, err := r.K8sApiExtensionsClientFactory.CreateFromInClusterConfig()
```
carbonin, line 2307: Why use a different client type here? Is there any reason to not use the same kind we're already using everywhere else in this file?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -2288,3 +2301,43 @@ func (r *AgentServiceConfigReconciler) getImageService(ctx context.Context, log
...
+	if err != nil {
+		return err
+	}
+
+	// Fetch all agent-install CRDs in cluster
+	crdClient := kubeClient.ApiextensionsV1().CustomResourceDefinitions()
+	crds, err := crdClient.List(context.TODO(), metav1.ListOptions{
```
carbonin, line 2314: Why not pass the context from the caller in and use that?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -2288,3 +2301,43 @@ func (r *AgentServiceConfigReconciler) getImageService(ctx context.Context, log
...
+		return err
+	}
+
+	// Create the CRDs on the external cluster
+	crdClient = kubeClient.ApiextensionsV1().CustomResourceDefinitions()
+	for _, crd := range crds.Items {
+		// ResourceVersion should not be set on objects to be created
+		crd.ResourceVersion = ""
+		c := crd
+		_, err1 := crdClient.Create(context.TODO(), &c, metav1.CreateOptions{})
```
carbonin, line 2334: Same here.

Use contexts whenever you can.

pkg/k8sclient/k8s_apiextensions_client.go
```diff
@@ -0,0 +1,94 @@
+package k8sclient
```
carbonin, line 1: How is this significantly different from [the other client factory implementation](https://github.com/openshift/assisted-service/blob/3541ee5294a89fe9a10a03f1b561790a30a10cc2/internal/spoke_k8s_client/spoke_k8s_client.go) we already have? Why create a new one for this?

carbonin (commented):
While this looks like it will work. I wonder how the other teams involved in this effort (installing ACM in hypershift zero-worker mode) are handling this problem for their CRDs.

I would think it would be better to handle this in whatever is installing the operator (as OLM/ACM installer does for us currently).

internal/controller/controllers/agentserviceconfig_controller.go
danielerez: We need specifically the 'ApiextensionsV1' API (for listing/creatind CRDs with ApiextensionsV1().CustomResourceDefinitions()). Which seems separate from  [kubernetes.clientset](https://github.com/openshift/assisted-service/blob/84b83af19531313df8186ea4f075894fb518f9a8/vendor/k8s.io/client-go/kubernetes/clientset.go#L19)

internal/controller/controllers/agentserviceconfig_controller.go
danielerez: Had an issue with the context for ApiextensionsV1 with creating CRDs, I'll try to reproduce to see the specific error.

pkg/k8sclient/k8s_apiextensions_client.go
danielerez: This is using specifically the apiextensions [clientset](k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset). Can try to make the SpokeK8sClient more generic and get a specific clientset. Just not sure it really worth the effort, as it serves different use-cases:/

danielerez:
> While this looks like it will work. I wonder how the other teams involved in this effort (installing ACM in hypershift zero-worker mode) are handling this problem for their CRDs.
> 
> I would think it would be better to handle this in whatever is installing the operator (as OLM/ACM installer does for us currently).

Yeah, couldn't find a better approach yet. The similar work in ACM is still in planning: https://issues.redhat.com/browse/ACM-1566

config/rbac/role.yaml
danielerez: Right, basically we need list/get for the in-cluster and create from the external cluster. I'll try to separate these.

mhrivnak:
> > While this looks like it will work. I wonder how the other teams involved in this effort (installing ACM in hypershift zero-worker mode) are handling this problem for their CRDs.
> > I would think it would be better to handle this in whatever is installing the operator (as OLM/ACM installer does for us currently).
> 
> Yeah, couldn't find a better approach yet. The similar work in ACM is still in planning: https://issues.redhat.com/browse/ACM-1566

I also agree. We should coordinate with ACM and hive to figure out a consistent way to install what's needed with minimal re-invention.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -102,6 +102,12 @@ type AgentServiceConfigSpec struct {
 	// +kubebuilder:validation:default:=disabled
 	// +optional
 	IPXEHTTPRoute string `json:"iPXEHTTPRoute,omitempty"`
+
+	// KubeconfigSecretRef is the reference to the Secret containing a kubeconfig of an external
+	// control plane (e.g. hypershift HostCluster).
+	// +optional
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Control plane kubeconfig reference"
+	KubeconfigSecretRef *corev1.LocalObjectReference `json:"kubeconfigSecretRef"`
```
mhrivnak, line 110: Per discussion earlier today, we need to further discuss what the right API is for this and to signal the other behavioral changes that are needed for the hypershift case.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -172,7 +178,7 @@ type AgentServiceConfigStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:resource:scope=Namespaced
```
mhrivnak, line 181: We can't do this without a plan to prove that it can be backward-compatible.

cmd/operator/main.go
```diff
@@ -136,14 +137,15 @@ func main() {
...
 	if err = (&controllers.AgentServiceConfigReconciler{
-		Client:       mgr.GetClient(),
-		Log:          logrus.New(),
-		Scheme:       mgr.GetScheme(),
-		Recorder:     mgr.GetEventRecorderFor("agentserviceconfig-controller"),
-		Namespace:    ns,
-		NodeSelector: nodeSelector,
-		Tolerations:  tolerations,
+		Client:                        mgr.GetClient(),
+		Log:                           log,
```
mhrivnak, line 143: Why the change here?

internal/controller/controllers/agentserviceconfig_controller.go
mhrivnak: I'm not following. controller-runtime gives us a great dynamic client with automatic caching, and it's almost always the client we want to use. CRD is [just another type](https://github.com/kubernetes/apiextensions-apiserver/blob/520f4b4/pkg/apis/apiextensions/v1beta1/types.go#L398) that you can [Get](https://sdk.operatorframework.io/docs/building-operators/golang/references/client/#get) or [List](https://sdk.operatorframework.io/docs/building-operators/golang/references/client/#list) like any other resource. Is there some reason not to use it?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -2237,11 +2293,51 @@ func getStorageRequests(pvcSpec *corev1.PersistentVolumeClaimSpec) map[corev1.Re
...
+	crds, err := crdClient.List(context.TODO(), metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("operators.coreos.com/assisted-service-operator.%s=", namespace),
+	})
+	if err != nil || len(crds.Items) == 0 {
+		log.Error(err)
+		return pkgerror.New("agent-install CRDs are not available")
+	}
+
+	// Get external cluster client (using kubeconfig specified in ASC)
+	kubeClient, err = r.K8sApiExtensionsClientFactory.CreateFromSecret(secret)
```
mhrivnak, line 2323: Setting up a client every time could get expensive. We should consider options.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -2237,11 +2293,51 @@ func getStorageRequests(pvcSpec *corev1.PersistentVolumeClaimSpec) map[corev1.Re
...
+	if err != nil {
+		return err
+	}
+
+	// Create the CRDs on the external cluster
```
mhrivnak, line 2328: This whole function is looking very imperative, even from the name "deploy...". When reconciling, we want a declarative pattern that:
1. figures out the desired state
1. inspects the real world to identify the current state
1. changes the real world to bring it closer to the desired state

Per comments elsewhere, we should figure out a coordinated strategy with ACM and hive for how to install and manage the CRDs in the hypershift control planes. If we do it through code like this, we'll probably want to setup a long-lived dynamic client for each control plane that can be reused.

internal/controller/controllers/agentserviceconfig_controller.go
danielerez: Seems ACM is using exactly the same apiextensions client for that: https://github.com/open-cluster-management-io/registration-operator/blob/79d280ad8a82055848a25bc9bcb83717f8db27e9/pkg/helpers/helpers.go#L448
I could indeed try to do the same with the dynamic client.
@mhrivnak WDYT, worth the effort to try that or we should keep follow ACM approach with apiextensionsclient?

mhrivnak: Is there some advantage to using that separate client? Generally it seems simpler to just one client for everything unless there's a compelling reason to bring in a new one.

PR #4358, opened 2022-09-05
MGMT-11506: Present a more useful validation message for packet loss

paul-maidment:
The current message for host packet loss is not helpful to end users and does little to help them get to the root of their issues.
This update attempts to remedy that by providing more concise information on how to troubleshoot the issue.

- [ ] New Feature 
- [X] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [x] None

## How was this code tested?
This code has unit test coverage.

- [ ] dev-scripts environment
- [] No tests needed

## Assignees
/cc @nmagnezi 
/cc @omertuc 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -44,6 +44,8 @@ const (
 	ValidationDisabled              ValidationStatus = "disabled"
 	maxServiceAheadOfHostTimeDiff                    = 20 * time.Minute
 	maxHostAheadOfServiceTimeDiff                    = 1 * time.Hour
+	maxHostTimingMetrics                             = 4
+	maxPingCommandExamples                           = 4
 )
 
 const FailedToFindAction = "failed to find action for step"
@@ -916,41 +918,54 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
 	return ValidationFailure, "While preparing the previous installation the installation disk speed measurement failed or was found to be insufficient"
 }
 
-func (v *validator) hasSufficientNetworkLatencyRequirementForRole(c *validationContext) (ValidationStatus, string) {
-	if c.infraEnv != nil {
-		return ValidationSuccessSuppressOutput, ""
-	}
-	if len(c.cluster.Hosts) == 1 || c.clusterHostRequirements.Total.NetworkLatencyThresholdMs == nil || common.GetEffectiveRole(c.host) == models.HostRoleAutoAssign || hostutil.IsDay2Host(c.host) {
-		// Single Node use case || no requirements defined || role is auto assign
-		return ValidationSuccess, "Network latency requirement has been satisfied."
-	}
-	if len(c.host.Connectivity) == 0 {
-		return ValidationPending, "Missing network latency information."
-	}
-	s, hostLatencies, err := v.validateNetworkLatencyForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts, c.inventoryCache)
-	if s == ValidationFailure {
-		if err != nil {
-			return ValidationFailure, fmt.Sprintf("Error while attempting to validate network latency: %s", err)
-		}
-		return ValidationFailure, fmt.Sprintf("Network latency requirements of %s %.2f ms not met for connectivity between %s and%s.", comparisonBuilder(*c.clusterHostRequirements.Total.NetworkLatencyThresholdMs), *c.clusterHostRequirements.Total.NetworkLatencyThresholdMs, c.host.ID, strings.Join(hostLatencies, ","))
-	}
-	if s == ValidationError {
-		return ValidationError, "Parse error while attempting to process the connectivity report"
-	}
-	return ValidationSuccess, "Network latency requirement has been satisfied."
+type hostTimingMetric struct {
+	otherHostName string
+	timingMetric  float64
+	timingSuffix  string
 }
 
-func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host, inventoryCache InventoryCache) (ValidationStatus, []string, error) {
+func (v *validator) summarizeHostTimingMetrics(packetLossInfo []hostTimingMetric, truncateMetrics bool) string {
+	result := []string{}
+	for i, p := range packetLossInfo {
+		//If there a lot of hosts in the cluster, this list could be rather large, so we shorten it
+		if truncateMetrics && i > maxHostTimingMetrics {
+			result = append(result, fmt.Sprintf("%s (%.2f%s) and others...", p.otherHostName, p.timingMetric, p.timingSuffix))
+			break
+		}
+		result = append(result, fmt.Sprintf("%s (%.2f%s)", p.otherHostName, p.timingMetric, p.timingSuffix))
+	}
+	return strings.Join(result, ", ")
+}
+
+type thresholdTestType int
+
+const (
+	thresholdTestL3AverageRTTMs thresholdTestType = 0
+	thresholdTestL3PacketLoss   thresholdTestType = 1
+)
+
+func (v *validator) thresholdExceededTest(testType thresholdTestType, host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host, inventoryCache InventoryCache) (ValidationStatus, []hostTimingMetric, error) {
+
 	connectivityReport, err := hostutil.UnmarshalConnectivityReport(host.Connectivity)
 	if err != nil {
 		v.log.Errorf("Unable to unmarshall host connectivity for %s:%s", host.ID, err)
 		return ValidationError, nil, nil
 	}
 	failedHostIPs := map[string]struct{}{}
-	failedHostLatencies := []string{}
+	failedHostMetrics := []hostTimingMetric{}
 	for _, r := range connectivityReport.RemoteHosts {
 		for _, l3 := range r.L3Connectivity {
-			if l3.AverageRTTMs > *clusterRoleReqs.Total.NetworkLatencyThresholdMs {
+
+			var hostHasExceededThreshold bool
+			switch testType {
+			case thresholdTestL3AverageRTTMs:
+				hostHasExceededThreshold = l3.AverageRTTMs > *clusterRoleReqs.Total.NetworkLatencyThresholdMs
+			case thresholdTestL3PacketLoss:
+				hostHasExceededThreshold = l3.PacketLossPercentage > *clusterRoleReqs.Total.PacketLossPercentage
+			default:
+				return ValidationError, nil, fmt.Errorf("unexpected testType")
+			}
+			if hostHasExceededThreshold {
 				if _, ok := failedHostIPs[l3.RemoteIPAddress]; !ok {
 					hostname, role, err := GetHostnameAndEffectiveRoleByHostID(r.HostID, hosts, inventoryCache)
 					if err != nil {
@@ -959,28 +974,22 @@ func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRole
 					}
 					if role == common.GetEffectiveRole(host) {
 						failedHostIPs[l3.RemoteIPAddress] = struct{}{}
-						failedHostLatencies = append(failedHostLatencies, fmt.Sprintf(" %s (%.2f ms)", hostname, l3.AverageRTTMs))
+						switch testType {
+						case thresholdTestL3AverageRTTMs:
+							failedHostMetrics = append(failedHostMetrics, hostTimingMetric{otherHostName: hostname, timingMetric: l3.AverageRTTMs, timingSuffix: " ms"})
+						case thresholdTestL3PacketLoss:
+							failedHostMetrics = append(failedHostMetrics, hostTimingMetric{otherHostName: hostname, timingMetric: l3.PacketLossPercentage, timingSuffix: "%"})
+						}
 					}
 				}
 			}
 		}
 	}
-	if len(failedHostLatencies) > 0 {
-		return ValidationFailure, failedHostLatencies, nil
+	if len(failedHostMetrics) > 0 {
+		return ValidationFailure, failedHostMetrics, nil
 	}
 	return ValidationSuccess, nil, nil
-}
 
-const (
-	lessThanOr = "less than or"
-	equals     = "equals"
-)
-
-func comparisonBuilder(value float64) string {
-	if value > 0 {
-		return fmt.Sprintf("%s %s", lessThanOr, equals)
-	}
-	return equals
 }
 
 func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationContext) (ValidationStatus, string) {
@@ -994,7 +1003,7 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
 	if len(c.host.Connectivity) == 0 {
 		return ValidationPending, "Missing packet loss information."
 	}
-	status, hostPacketLoss, err := v.validatePacketLossForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts, c.inventoryCache)
+	status, hostMetrics, err := v.thresholdExceededTest(thresholdTestL3PacketLoss, c.host, c.clusterHostRequirements, c.cluster.Hosts, c.inventoryCache)
 	if err != nil {
 		return status, fmt.Sprintf("Error while attempting to validate packet loss validation: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


internal/host/validator.go
```diff
@@ -977,8 +1004,9 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
 	case ValidationSuccess:
 		return status, "Packet loss requirement has been satisfied."
 	case ValidationFailure:
```
omertuc, line 1006: and%s

internal/host/validator.go
```diff
@@ -977,8 +1004,9 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
...
-
-		return status, fmt.Sprintf("Packet loss percentage requirement of %s %.2f%% not met for connectivity between %s and%s.", comparisonBuilder(*c.clusterHostRequirements.Total.PacketLossPercentage), *c.clusterHostRequirements.Total.PacketLossPercentage, c.host.ID, strings.Join(hostPacketLoss, ","))
+		packetLossAdvisoryMessage := fmt.Sprintf("A total packet loss above the tolerated threshold of %.2f%% was encountered when performing connectivity validation between host %s and%s", *c.clusterHostRequirements.Total.PacketLossPercentage, c.host.ID, strings.Join(hostPacketLoss, ","))
```
omertuc, line 1007: and%s

internal/host/validator.go
```diff
@@ -957,6 +957,33 @@ func comparisonBuilder(value float64) string {
 	return equals
 }
 
+func (v *validator) generatePacketLossAdvisoryMessageForHost(c *validationContext, hostPacketLoss [] string) string {
+	message := "Actions:\n"
+	message += "1: Check if there are multiple devices on the same L2 network, if so then create a bond or disable all but one of the interfaces.\n"
+
+	var inventory models.Inventory
+	err := json.Unmarshal([]byte(c.host.Inventory), &inventory)
```
omertuc, line 965: There's a deserialized inventory cache Somewhere:tm: in this file, could it be used here?

internal/host/validator.go
```diff
@@ -957,6 +957,33 @@ func comparisonBuilder(value float64) string {
 	return equals
 }
 
+func (v *validator) generatePacketLossAdvisoryMessageForHost(c *validationContext, hostPacketLoss [] string) string {
+	message := "Actions:\n"
+	message += "1: Check if there are multiple devices on the same L2 network, if so then create a bond or disable all but one of the interfaces.\n"
```
omertuc, line 962: As a customer unfamiliar with our product, they might simply reach out to log in to the host and set up all this networking manually on the live environment - so perhaps we should tell them how they can go about setting up bonds or disabling interfaces etc (point them at the relevant nmstate networking config feature of AI)

internal/host/validator.go
```diff
@@ -957,6 +957,33 @@ func comparisonBuilder(value float64) string {
...
+	}
+	hostName := getRealHostname(c.host, &inventory)
+	if len(inventory.Interfaces) > 0 {
+		message = fmt.Sprintf("2: Please try the following commands on the host %s to investigate the packet loss further\n\n", hostName)
+		for _, intf := range inventory.Interfaces {
+			for _, hostPacketLossEntry := range hostPacketLoss {
+				entryParts := strings.Split(hostPacketLossEntry, " ")
+				if len(entryParts) > 0 {
+					targetHostName := entryParts[0]
+					message += fmt.Sprintf("ping -c 10 -W 3 -q -I %s %s\n", intf.Name, targetHostName)
```
omertuc, line 977:
```suggestion
					// This command must be kept in sync with the agent repo <path to go file in agent repo with this command>
					message += fmt.Sprintf("ping -c 10 -W 3 -q -I %s %s\n", intf.Name, targetHostName)
```

internal/host/validator.go
omertuc: The exact phrasing of these directions should take into account both UI users and API users, as both of them might see this message

internal/host/validator.go
omertuc: oh actually this seems to be on purpose, nevermind

internal/host/validator.go
```diff
@@ -957,6 +957,33 @@ func comparisonBuilder(value float64) string {
...
+	err := json.Unmarshal([]byte(c.host.Inventory), &inventory)
+	if err != nil {
+		v.log.WithError(err).Warnf("Could not parse inventory of host %s\n", *c.host.ID)
+	}
+	hostName := getRealHostname(c.host, &inventory)
+	if len(inventory.Interfaces) > 0 {
+		message = fmt.Sprintf("2: Please try the following commands on the host %s to investigate the packet loss further\n\n", hostName)
+		for _, intf := range inventory.Interfaces {
+			for _, hostPacketLossEntry := range hostPacketLoss {
+				entryParts := strings.Split(hostPacketLossEntry, " ")
```
omertuc, line 974: All this formatting and splitting is rather confusing, can we change `validatePacketLossForRole` to return structured data rather than a formatted string, so we can use it here rather than extracting it from an already formatted string?

internal/host/validator.go
mkowalski: NIT, missing space in `between host %s and%s` ?

internal/host/validator.go
nmagnezi: IMO the wording here serves the purpose of this PR.
There's only so much we can cram in one error, and different users want to act upon this case of two interfaces on the same L2 network (which could be validation on its own) differently, as you mentioned.

Say we do point users to some URL (example: https://nmstate.io/examples.html#interfaces-bond), they might end up copying the configs as-is, and it would not work for them.

@paul-maidment, is this case failing the validation or just producing a warning?

mkowalski: It is important to note that we have disabled the packet loss validation in prod (/cc @romfreiman @gamli75). The reasons for doing so are probably not valid any more, but we should be aware that whatever the PR produces here, won't be immediately exposed to users.

In case we re-enable the validation, it should be a blocking one.

omertuc: When a user sees "create a bond or disable all but one of the interfaces" - the only way a new user might know how to perform the action mentioned in that quote is by logging into the host and actively disabling the interface. This will be fruitless as it will only solve the issue during discovery and the issue will return once the host reboots.

We need to tell them they have to perform network configuration using our nmstate API / UI, regenerate the ISO, etc. All of that stuff is not at all obvious. I don't think we need to cram an entire tutorial inside the error message, just point the user at the relevant documentation for our networking config feature

mkowalski: I couldn't agree more with Omer about the above. Also, is it worth to redact it all inside this error message given that we are going to spend the whole next week in rewriting our docs? Sounds like potentially a duplicated work

paul-maidment: So to clarify, is the suggestion that we include a link to the markdown files here?

https://github.com/openshift/assisted-service/blob/master/docs/user-guide/network-configuration/static-configuration.md

mkowalski: In a few days/weeks we will have this markdown file rewritten and published directly in docs.openshift.com; my personal suggestion is to wait till it's done and then link to the official documentation

filanov:
the commit message is not very helpful, it doesn't explain what you did to improve the message.

internal/host/validator.go
```diff
@@ -977,43 +980,82 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
 	case ValidationSuccess:
 		return status, "Packet loss requirement has been satisfied."
 	case ValidationFailure:
-
-		return status, fmt.Sprintf("Packet loss percentage requirement of %s %.2f%% not met for connectivity between %s and%s.", comparisonBuilder(*c.clusterHostRequirements.Total.PacketLossPercentage), *c.clusterHostRequirements.Total.PacketLossPercentage, c.host.ID, strings.Join(hostPacketLoss, ","))
+		packetLossAdvisoryMessage := fmt.Sprintf("A total packet loss above the tolerated threshold of %.2f%% was encountered when performing connectivity validation between host %s and %s\n", *c.clusterHostRequirements.Total.PacketLossPercentage, c.host.ID, v.summarizeHostTimingMetrics(hostMetrics))
```
filanov, line 983: please break long lines

internal/host/validator.go
```diff
@@ -890,41 +890,50 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
...
 
-func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host, inventoryCache InventoryCache) (ValidationStatus, []string, error) {
+type ThresholdTestType int
+
+const (
+	ThresholdTestL3AverageRTTMs ThresholdTestType = 0
+	ThresholdTestL3PacketLoss   ThresholdTestType = 1
+)
+
+func (v *validator) validatePacketLossForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host, inventoryCache InventoryCache) (ValidationStatus, []HostTimingMetric, error) {
```
filanov, line 913: why do you need to create two functions that call the same function with a different parameter? 
Is there a reason not to expose the parameter?

internal/host/validator.go
```diff
@@ -977,43 +980,82 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
...
+	}
+	if len(c.host.Connectivity) == 0 {
+		return ValidationPending, "Missing network latency information."
+	}
+	s, hostMetrics, err := v.validateNetworkLatencyForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts, c.inventoryCache)
+	if s == ValidationFailure {
+		if err != nil {
+			return ValidationFailure, fmt.Sprintf("Error while attempting to validate network latency: %s", err)
+		}
+		networkLatencyAdvisoryMessage := fmt.Sprintf("A total network latency above the tolerated threshold of %.2f%% was encountered when performing network latency tests between host %s and %s\n", *c.clusterHostRequirements.Total.PacketLossPercentage, c.host.ID, v.summarizeHostTimingMetrics(hostMetrics))
```
filanov, line 1030: please break long lines

internal/host/validator.go
filanov: @avishayt what do you think?

avishayt: I agree that some will configure it manually via SSH and that more information would be helpful.
However, linking to a doc isn't a great solution IMO because the URL can break and because we would need to point to different docs depending on SaaS, ACM, MCE, agent-based, etc.
How about something more generic like "..., if so then use the built-in nmstate-based advanced networking configuration to create a bond or disable all but one of the interfaces"

filanov: i like that idea and agree with the point about the links

ori-amizur:
When there are many hosts in cluster, with many packet losses, this message becomes very long and unusable.
A better approach would be to provide a link to a location where there is an explanation about the type of problem that was encountered.
We should probably have a database of validations and explanation of each validation. Here we should keep the message as short as possible.

internal/host/validator.go
```diff
@@ -977,43 +980,82 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
...
-				}
+		v.log.WithError(err).Warnf("Could not parse inventory of host %s\n", *c.host.ID)
+	}
+	hostName := getRealHostname(c.host, inventory)
+	if len(inventory.Interfaces) > 0 {
+		message += fmt.Sprintf("2: Please try the following commands on the host %s to investigate the packet loss further\n\n", hostName)
+		for _, intf := range inventory.Interfaces {
+			for _, p := range hostPacketLoss {
+				// This command must be kept in sync with the agent repo https://github.com/openshift/assisted-installer-agent/blob/a35f7c36951313f6a6948a190ca6b56d1472516b/src/connectivity_check/connectivity_check.go#L87
+				message += fmt.Sprintf("ping -c 10 -W 3 -q -I %s %s\n", intf.Name, p.HostName)
```
ori-amizur, line 1006: Connectivity check checks ips.  DNS setting is not guaranteed.

internal/host/validator.go
ori-amizur: I am for the links too

internal/host/validator.go
```diff
@@ -890,41 +890,42 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
...
-	s, hostLatencies, err := v.validateNetworkLatencyForRole(c.host, c.clusterHostRequirements, c.cluster.Hosts, c.inventoryCache)
-	if s == ValidationFailure {
-		if err != nil {
-			return ValidationFailure, fmt.Sprintf("Error while attempting to validate network latency: %s", err)
-		}
-		return ValidationFailure, fmt.Sprintf("Network latency requirements of %s %.2f ms not met for connectivity between %s and%s.", comparisonBuilder(*c.clusterHostRequirements.Total.NetworkLatencyThresholdMs), *c.clusterHostRequirements.Total.NetworkLatencyThresholdMs, c.host.ID, strings.Join(hostLatencies, ","))
-	}
-	if s == ValidationError {
-		return ValidationError, "Parse error while attempting to process the connectivity report"
+type HostTimingMetric struct {
```
omertuc, line 893: should be private

internal/host/validator.go
```diff
@@ -890,41 +890,42 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
...
+	result := []string{}
+	for _, p := range packetLossInfo {
+		result = append(result, fmt.Sprintf("%s (%.2f%s)", p.HostName, p.TimingMetric, suffix))
 	}
-	return ValidationSuccess, "Network latency requirement has been satisfied."
+	return strings.Join(result, ", ")
 }
 
-func (v *validator) validateNetworkLatencyForRole(host *models.Host, clusterRoleReqs *models.ClusterHostRequirements, hosts []*models.Host, inventoryCache InventoryCache) (ValidationStatus, []string, error) {
+type ThresholdTestType int
```
omertuc, line 906: should be private

internal/host/validator.go
```diff
@@ -890,41 +890,42 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
...
+
+const (
+	ThresholdTestL3AverageRTTMs ThresholdTestType = 0
```
omertuc, line 909: should be private

internal/host/validator.go
```diff
@@ -890,41 +890,42 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
...
+	ThresholdTestL3PacketLoss   ThresholdTestType = 1
```
omertuc, line 910: should be private

internal/host/validator.go
```diff
@@ -890,41 +890,42 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
...
-	if s == ValidationFailure {
-		if err != nil {
-			return ValidationFailure, fmt.Sprintf("Error while attempting to validate network latency: %s", err)
-		}
-		return ValidationFailure, fmt.Sprintf("Network latency requirements of %s %.2f ms not met for connectivity between %s and%s.", comparisonBuilder(*c.clusterHostRequirements.Total.NetworkLatencyThresholdMs), *c.clusterHostRequirements.Total.NetworkLatencyThresholdMs, c.host.ID, strings.Join(hostLatencies, ","))
-	}
-	if s == ValidationError {
-		return ValidationError, "Parse error while attempting to process the connectivity report"
+type HostTimingMetric struct {
+	HostName     string
```
omertuc, line 894: should be private

internal/host/validator.go
```diff
@@ -890,41 +890,42 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
...
+	TimingMetric float64
```
omertuc, line 895: should be private

internal/host/validator.go
omertuc, line 894: Is my understanding correct that `otherHostHostname` might be a better name for this member?

internal/host/validator.go
```diff
@@ -977,43 +977,101 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
...
-		return ValidationFailure, failedHostPacketLoss, nil
+	if s == ValidationError {
+		return ValidationError, "Parse error while attempting to process the connectivity report"
 	}
-	return ValidationSuccess, nil, nil
+	return ValidationSuccess, "Network latency requirement has been satisfied."
+}
+
+func (v *validator) generateExcessiveLatencyAdvisoryForHost(c *validationContext) string {
+	message := "Actions:\n"
```
nmagnezi, line 1063: Please make sure to return an empty message if len(inventory.Interfaces) == 0

nmagnezi (changes requested):

nmagnezi:
please update the commit-msg to include some info about the refactor done with this PR.

internal/host/validator.go
```diff
@@ -916,41 +916,57 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
...
-	if s == ValidationFailure {
-		if err != nil {
-			return ValidationFailure, fmt.Sprintf("Error while attempting to validate network latency: %s", err)
+type hostTimingMetric struct {
+	otherHostName string
+	timingMetric  float64
+	timingSuffix  string
+}
+
+const MAX_HOST_TIMING_METRICS = 4
```
nmagnezi, line 925: Please put the const at the top of the file.
Additionally, the naming convention should be `camelCase`

internal/host/validator.go
```diff
@@ -916,41 +916,57 @@ func (v *validator) sufficientOrUnknownInstallationDiskSpeed(c *validationContex
...
+const MAX_PING_COMMAND_EXAMPLES = 4
```
nmagnezi, line 926: ditto

internal/host/validator.go
```diff
@@ -1003,43 +1013,120 @@ func (v *validator) hasSufficientPacketLossRequirementForRole(c *validationConte
 	case ValidationSuccess:
 		return status, "Packet loss requirement has been satisfied."
 	case ValidationFailure:
-
-		return status, fmt.Sprintf("Packet loss percentage requirement of %s %.2f%% not met for connectivity between %s and%s.", comparisonBuilder(*c.clusterHostRequirements.Total.PacketLossPercentage), *c.clusterHostRequirements.Total.PacketLossPercentage, c.host.ID, strings.Join(hostPacketLoss, ","))
+		// When logging, make sure the full timing metrics are logged.
+		fullHostTimingMetrics := v.summarizeHostTimingMetrics(hostMetrics, false)
+		v.log.Info(fmt.Sprintf(`A total packet loss above the tolerated threshold of %.2f%% was encountered when performing connectivity validation between host %s and %s\n`,
```
nmagnezi, line 1018: If we log this under `ValidationFailure`, shouldn't this be a `log.Error`?
@filanov what do you think?

nmagnezi (changes requested):

internal/host/validator.go
paul-maidment: I agree and have made this change

internal/host/validator.go
paul-maidment: Done

internal/host/validator.go
paul-maidment: Done and done

nmagnezi (approved):
/lgtm

PR #3814, opened 2022-05-15
[WIP] MGMT-10195: arm feature enablement for v2 list supported openshift versions

eliorerz:
DO NOT REVIEW

## List all the issues related to this PR

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [x] No tests needed

## Assignees

/cc @nmagnezi 

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #4368, opened 2022-09-08
MGMT-11563: Allow org-based filtering for multiarch images

mkowalski:
Contributes-to: [MGMT-11563](https://issues.redhat.com//browse/MGMT-11563)
Contributes-to: [MGMT-11859](https://issues.redhat.com//browse/MGMT-11859)
Implements: https://github.com/openshift/assisted-service/pull/3744
Implements-enhancement: saas-per-organization-feature-access

/cc @nmagnezi 
/cc @danielerez

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -463,6 +463,11 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 	if len(releaseImage.CPUArchitectures) > 1 {
 		log.Infof("Setting cluster as multi-arch because of the release image (requested was %s)", cpuArchitecture)
 		cpuArchitecture = common.MultiCPUArchitecture
+		// (MGMT-11859) Additional check here ensures that the customer cannot just guess a version
+		//              with multiarch in order to get access to that release payload.
+		if err = b.versionsHandler.ValidateAccessToMultiarch(ctx, b.authzHandler); err != nil {
+			return nil, err
+		}
 	}
 
 	if kubeKey == nil {
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -11153,6 +11153,7 @@ var _ = Describe("TestRegisterCluster", func() {
 		mockAMSSubscription(ctx)
 		mockVersions.EXPECT().GetCPUArchitectures(gomock.Any()).Return(
 			[]string{common.TestDefaultConfig.OpenShiftVersion, common.ARM64CPUArchitecture}).Times(1)
+		mockVersions.EXPECT().ValidateAccessToMultiarch(gomock.Any(), gomock.Any()).Times(1)
 
 		params := getDefaultClusterCreateParams()
 		params.CPUArchitecture = common.ARM64CPUArchitecture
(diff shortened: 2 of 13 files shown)


mkowalski:
/test edge-subsystem-aws

Subsystem does not have 4.6 so the failure was valid
```
--release-images '[{\"openshift_version\":\"4.8\",\"cpu_architecture\":\"x86_64\",\"cpu_architectures\":[\"x86_64\"],\"url\":\"quay.io/openshift-release-dev/ocp-release:4.8.48-x86_64\",\"version\":\"4.8.48\"},{\"openshift_version\":\"4.9\",\"cpu_architecture\":\"x86_64\",\"cpu_architectures\":[\"x86_64\"],\"url\":\"quay.io/openshift-release-dev/ocp-release:4.9.46-x86_64\",\"version\":\"4.9.46\"},{\"openshift_version\":\"4.10\",\"cpu_architecture\":\"x86_64\",\"cpu_architectures\":[\"x86_64\"],\"url\":\"quay.io/openshift-release-dev/ocp-release:4.10.28-x86_64\",\"version\":\"4.10.28\"},{\"openshift_version\":\"4.10\",\"cpu_architecture\":\"arm64\",\"cpu_architectures\":[\"arm64\"],\"url\":\"quay.io/openshift-release-dev/ocp-release:4.10.28-aarch64\",\"version\":\"4.10.28\"},{\"openshift_version\":\"4.11\",\"cpu_architecture\":\"x86_64\",\"cpu_architectures\":[\"x86_64\"],\"url\":\"quay.io/openshift-release-dev/ocp-release:4.11.0-x86_64\",\"version\":\"4.11.0\",\"default\":true},{\"openshift_version\":\"4.11\",\"cpu_architecture\":\"arm64\",\"cpu_architectures\":[\"arm64\"],\"url\":\"quay.io/openshift-release-dev/ocp-release:4.11.0-aarch64\",\"version\":\"4.11.0\"},{\"openshift_version\":\"4.11\",\"cpu_architecture\":\"multi\",\"cpu_architectures\":[\"x86_64\",\"arm64\",\"ppc64le\",\"s390x\"],\"url\":\"quay.io/openshift-release-dev/ocp-release:4.11.0-multi\",\"version\":\"4.11.0-multi\"}]'
```

internal/versions/versions.go
```diff
@@ -103,6 +106,29 @@ func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params o
 			supportedArchs = []string{*releaseImage.CPUArchitecture}
 		}
 
+		// (MGMT-11859) We are filtering out multiarch release images so that they are available
+		//              only for customers allowed to use them. This is in order to be able to
+		//              expose them in OCP pre-4.13 without making them generally available.
+		if len(supportedArchs) > 1 {
+			hasMultiarchAuthorization := false
```
danielerez, line 113: I think we should use go-cache for this instead of using a couple of flags. It would allow expiration in case we want to grant/revoke perms, and should look nicer. For example see: https://github.com/openshift/assisted-service/blob/096a0f99b15dbb9a5afb71c86c146617511f14d6/pkg/auth/rhsso_authz_handler.go#L303

internal/versions/versions.go
```diff
@@ -103,6 +106,29 @@ func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params o
...
+			checkedForMultiarchAuthorization := false
+
+			if !checkedForMultiarchAuthorization {
+				checkedForMultiarchAuthorization = true
+				if err := auth.ValidateAccessToMultiarch(ctx, h.authzHandler); err != nil {
```
danielerez, line 118: Maybe we can incorporate the caching in this func.

pkg/auth/authz_utils.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/openshift/assisted-service/internal/common"
+	"github.com/openshift/assisted-service/pkg/ocm"
+	"github.com/pkg/errors"
+)
+
+const (
+	NoAccessToMultiarch = "multiarch clusters are not available"
+)
+
+func ValidateAccessToMultiarch(ctx context.Context, authzHandler Authorizer) error {
```
danielerez, line 17: Not sure we need this new utils file. If we pass authzHandler to versions any way, we can add this func to the handler. That way we could reuse the same cache. Also, please add unit-tests to this func.

internal/versions/versions.go
nmagnezi: We decided not to do caching for the ARM capability at the time since it was a flow that is not triggered too often. Isn't it the case here as well?

pkg/auth/authz_utils.go
nmagnezi: +1
This is how we did if for `V2ListSupportedOpenshiftVersions` in https://github.com/openshift/assisted-service/pull/3814/

internal/versions/versions.go
```diff
@@ -103,6 +106,29 @@ func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params o
...
 		}
 
+		// (MGMT-11859) We are filtering out multiarch release images so that they are available
+		//              only for customers allowed to use them. This is in order to be able to
+		//              expose them in OCP pre-4.13 without making them generally available.
+		if len(supportedArchs) > 1 {
+			hasMultiarchAuthorization := false
+			checkedForMultiarchAuthorization := false
+
+			if !checkedForMultiarchAuthorization {
```
nmagnezi, line 116: Please explain.
This if statement is always `true`, after which you change `checkedForMultiarchAuthorization` to `true`. Why?

pkg/auth/authz_utils.go
nmagnezi: `ValidateAccessToMultiarch` should be a part of `versionsHandler`.

internal/versions/versions.go
```diff
@@ -103,6 +106,29 @@ func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params o
...
+				if err := auth.ValidateAccessToMultiarch(ctx, h.authzHandler); err != nil {
+					if strings.Contains(err.Error(), auth.NoAccessToMultiarch) {
+						continue
+					} else {
+						return common.GenerateErrorResponder(err)
+					}
+				}
+				hasMultiarchAuthorization = true
+			}
+			if !hasMultiarchAuthorization {
```
nmagnezi, line 127: Given the if statement in line 118 + the assignment in 125, this is always `false`

nmagnezi (changes requested):

internal/versions/versions.go
filanov: i think that it's the same case, if the api is called only on update and create then cache is not needed.

internal/versions/versions.go
mkowalski: Yeah you are right, the initial value for the bools is now moved to the very top of this function to ensure that we do "check for authorization" only once in the whole call

internal/versions/versions.go
mkowalski: We have now merged the prerequisite to master, so subsystem test should tell us if this part works. I think after evacuating the initial values to the top of the function it's okay now

internal/versions/versions.go
danielerez: iirc, this API is called by the UI on every draft cluster creation, so caching is not crucial but would still be nice to have I think. Anyway, it'll make the func clearer if this logic can be extracted.

pkg/auth/authz_utils.go
mkowalski: >Also, please add unit-tests to this func.

I may be missing something, but the only thing this function does is calling `authzHandler.HasOrgBasedCapability` and blindly parsing the result (a touple of boolean and error). So unit test would be here something like

```
result, error := true, nil
if result == true {
    do something
} else {
    do something else
}
```

>Not sure we need this new utils file. If we pass authzHandler to versions any way, we can add this func to the handler
>ValidateAccessToMultiarch should be a part of versionsHandler.

ACK, will do and move. Maybe we should have a generic utility library for handling stuff around `authzHandler.HasOrgBasedCapability`? I'm the second person doing org-based capability and the first implementation was different from the one I'm going to do. This sounds like a something that should be potentially "frameworkable"

internal/versions/versions.go
mkowalski: That's true, the API is called every time you open UI for cluster creation. I just want to highlight the fact that this code is only temporary because in 4.13 everything will be multiarch. So everything I do in this PR will need to be reverted. And probably everyone here agrees that it will be not as simple as a commit-revert. Maybe we should classify it as "not ideal but good enough to keep us running for one release"?

danielerez: Oh ok, if that's the case then it's indeed not crucial:)

pkg/auth/authz_utils.go
danielerez: Right, some refactoring for that would be nice. Need to think how to structure it, maybe some lib just for capability checks:/ About test, it'd indeed should be a very simple test, don't you think it worth it? I mean, just to ensure we handle the error correctly and coverage..

mkowalski: I think it's worth in the function of increasing our coverage. I think it's not worth in the function of usefulness :) I'll add it as it's indeed not difficult so no reason not to do it

mkowalski: @danielerez unit test in place, I think we are complete now

subsystem/versions_test.go
```diff
@@ -22,4 +24,37 @@ var _ = Describe("[minimal-set]test versions", func() {
...
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(hasMultiarch(reply.GetPayload())).Should(BeTrue())
+		})
+	})
 })
+
+func hasMultiarch(versions models.OpenshiftVersions) bool {
+	hasMultiarch := false
+	for _, version := range versions {
+		if strings.HasSuffix(*version.DisplayName, "-multi") {
```
nmagnezi, line 54: Is this an agreed-upon name convention?
Maybe we should have `Contains` instead of `HasSuffix`? WDYT?

nmagnezi (changes requested):

subsystem/versions_test.go
mkowalski: It's an internal contract that `-multi` is a suffix, so I feel that it's fair to keep it as it is

internal/bminventory/inventory_test.go
```diff
@@ -11309,6 +11310,129 @@ var _ = Describe("TestRegisterCluster", func() {
...
+					Version:          swag.String("doesnt-really-matter"),
+				}, nil).Times(1)
+				mockOperatorManager.EXPECT().GetSupportedOperatorsByType(models.OperatorTypeBuiltin).Return([]*models.MonitoredOperator{&common.TestDefaultConfig.MonitoredOperator}).Times(1)
+				mockProviderRegistry.EXPECT().SetPlatformUsages(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
+				mockMetric.EXPECT().ClusterRegistered().Times(1)
+				mockEvents.EXPECT().SendClusterEvent(gomock.Any(), eventstest.NewEventMatcher(
+					eventstest.WithNameMatcher(eventgen.ClusterRegistrationSucceededEventName))).Times(1)
+				mockAMSSubscription(authCtx)
+				mockUsageReports()
+				// mockOcmAuthz.EXPECT().CapabilityReview(context.Background(), userName1, ocm.MultiarchCapabilityName, ocm.OrganizationCapabilityType).Return(true, nil)
```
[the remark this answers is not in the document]
mkowalski: Missed that, thanks!

internal/bminventory/inventory.go
```diff
@@ -463,6 +463,11 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 	if len(releaseImage.CPUArchitectures) > 1 {
 		log.Infof("Setting cluster as multi-arch because of the release image (requested was %s)", cpuArchitecture)
 		cpuArchitecture = common.MultiCPUArchitecture
+		// (MGMT-11859) Additional check here ensures that the customer cannot just guess a version
```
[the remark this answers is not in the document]
mkowalski: `cluster.CPUArchitecture: multi` and `len(releaseImage.CPUArchitectures) > 1` is not a symmetric relationship; technically you can use multi-arch release image with a cluster that sets CPU architecture as e.g. arm64

I think the statement to summarize what happens here should be - we are going to filter availability of multi-arch release images, therefore the validation check happens directly when we try to use multi-arch release image

It is *not* a validator check whether cluster's CPU architecture can be multi-arch.

Does it answer the question?

mkowalski:
/test edge-e2e-metal-assisted

Random failure

subsystem/versions_test.go
nmagnezi: ack 👍

internal/bminventory/inventory.go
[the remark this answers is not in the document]
nmagnezi: Yes. Thanks!

nmagnezi (commented):
LGTM.
@danielerez please vote if your comments/questions were addressed.

PR #4374, opened 2022-09-08
MGMT-8890: Add subsystem test for overlapping multiarch releases

mkowalski:
This commit extends subsystem test to ensure that if for a specific OCP version there is single- and multi-arch release payload, we are never accidentally picking the multi-arch unless requested.


vrutkovs (approved):
/lgtm

PR #4375, opened 2022-09-08
MGMT-11973: Fix vSphere platform with enabled user-managed-networking results none platform

eliorerz:
Fix changes made on https://github.com/openshift/assisted-service/pull/4158.
One case was missed, if platform (e.g. vsphere) is different than none or baremetal with user-managed-networking enabled it will result none platform instead of the desired platform.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @gamli75 
/cc @vrutkovs

### DIFF
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -5106,6 +5106,23 @@ var _ = Describe("[V2ClusterUpdate] cluster", func() {
 
 				})
 
+				It("Update platform=vsphere while and UMN=true - success", func() {
+					mockSuccess()
+					mockProviderRegistry.EXPECT().SetPlatformUsages(models.PlatformTypeVsphere, gomock.Any(), mockUsage)
+
+					reply := bm.V2UpdateCluster(ctx, installer.V2UpdateClusterParams{
+						ClusterID: clusterID,
+						ClusterUpdateParams: &models.V2ClusterUpdateParams{
+							Platform:              &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeVsphere)},
+							UserManagedNetworking: swag.Bool(true),
+						},
+					})
+					Expect(reply).Should(BeAssignableToTypeOf(installer.NewV2UpdateClusterCreated()))
+					actual := reply.(*installer.V2UpdateClusterCreated).Payload
+					Expect(swag.BoolValue(actual.UserManagedNetworking)).To(Equal(true))
+					Expect(*actual.Platform.Type).To(Equal(models.PlatformTypeVsphere))
+				})
+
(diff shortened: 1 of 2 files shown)


romfreiman:
do we have a UT that sets all the pltaform things and verify that the install config generated as expected?

internal/bminventory/inventory_test.go
```diff
@@ -9874,6 +9874,22 @@ var _ = Describe("TestRegisterCluster", func() {
...
+				params := getClusterCreateParams(nil)
+				params.Platform = &models.Platform{Type: common.PlatformTypePtr(models.PlatformTypeVsphere)}
+				params.UserManagedNetworking = swag.Bool(true)
+				reply := bm.V2RegisterCluster(ctx, installer.V2RegisterClusterParams{
+					NewClusterParams: params,
+				})
+				Expect(reply).Should(BeAssignableToTypeOf(installer.NewV2RegisterClusterCreated()))
+				actual := reply.(*installer.V2RegisterClusterCreated).Payload
+				Expect(swag.BoolValue(actual.UserManagedNetworking)).To(Equal(true))
+				Expect(*actual.Platform.Type).To(Equal(models.PlatformTypeVsphere))
```
romfreiman, line 9890: dont we have a way to verify final install config?

eliorerz: > do we have a UT that sets all the pltaform things and verify that the install config generated as expected?

We had for the most cases, but we didn't had a case where platform is vSphere and UMN enabled

romfreiman: and now? is this UT reads the generated install config? seems to me that it's only 'reading' from the DB

eliorerz: No, this UT is only checking the result cluster created on the database. 
I think that if we want to test the generated install-config it should be a separate effort. I'm not familiar with UT that register cluster and verify it install-config

romfreiman: cool. so lets verify it as well - even if it takes extra time.

i dont think there was an issue with db in the issue that we recently encountered tbh

eliorerz: OK, I'll add a task for it
FYI @gamli75

vrutkovs (approved):
/lgtm
/hold

for last test to complete

vrutkovs:
All tests completed
/hold cancel

PR #4380, opened 2022-09-11
MGMT-11056: Additional stages when adding workers with k8s API

eranco74:
When adding a worker to an existing cluster we move straight from Rebooting to Done.  However,  with the k8s API we have the kubeconfig and can determine the stages more accurately.

### This PR adds to the following changes:
1. If kubeAPIEnabled, do not move the Agent to Done, leave it in Rebooting.
2. If the agent-controller sees:
    a. a ready node for that host, move the Agent to Done
    b. a non-ready node for that host, move the Agent to Joined

- Is this PR relying on CI for an e2e test run? for regression
- Should this PR be tested in a specific environment? yes, with the infrastructure-operator
- Any logs, screenshots, etc that can help with the review process? nope

-->

https://issues.redhat.com/browse/MGMT-11047

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @carbonin 
/cc @nmagnezi 
/cc @ori-amizur 

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -311,26 +311,17 @@ func (r *AgentReconciler) spokeKubeClient(ctx context.Context, clusterRef *aiv1b
 
 // Attempt to approve CSRs for agent. If already approved then the node will be marked as done
 // requeue means that approval will be attempted again
-func (r *AgentReconciler) tryApproveDay2CSRs(ctx context.Context, agent *aiv1beta1.Agent) bool {
+func (r *AgentReconciler) tryApproveDay2CSRs(ctx context.Context, agent *aiv1beta1.Agent, client spoke_k8s_client.SpokeK8sClient) bool {
 	r.Log.Infof("Approving CSRs for agent %s/%s", agent.Namespace, agent.Name)
-
-	hostname := getAgentHostname(agent)
 	var (
 		validateNodeCsr       nodeCsrValidator
 		shouldApproveMoreCSRs bool
 	)
 
-	clients, err := r.spokeKubeClient(ctx, agent.Spec.ClusterDeploymentName)
-	if err != nil {
-		r.Log.WithError(err).Errorf("Agent %s/%s: Failed to create spoke client", agent.Namespace, agent.Name)
-		return false
-	}
-
-	// TODO: Node name might be FQDN and not just host name if cluster is IPv6
-	node, err := clients.GetNode(hostname)
+	node, err := r.getNode(agent, client)
 	if err != nil {
 		if !k8serrors.IsNotFound(err) {
-			r.Log.WithError(err).Errorf("Agent %s/%s: Failed to get node %s", agent.Namespace, agent.Name, hostname)
+			r.Log.WithError(err).Errorf("agent %s/%s: failed to get node", agent.Namespace, agent.Name)
 			return false
 		}
 		validateNodeCsr = validateNodeClientCSR
@@ -341,9 +332,16 @@ func (r *AgentReconciler) tryApproveDay2CSRs(ctx context.Context, agent *aiv1bet
 	}
 
 	// Even if node is already ready, we try approving last time
-	r.approveAIHostsCSRs(clients, agent, validateNodeCsr)
+	r.approveAIHostsCSRs(client, agent, validateNodeCsr)
 
 	return !shouldApproveMoreCSRs
+}
+
+func (r *AgentReconciler) getNode(agent *aiv1beta1.Agent, client spoke_k8s_client.SpokeK8sClient) (*corev1.Node, error) {
+	hostname := getAgentHostname(agent)
+
+	// TODO: Node name might be FQDN and not just host name if cluster is IPv6
+	return client.GetNode(hostname)
 }
 
 func (r *AgentReconciler) bmhExists(ctx context.Context, agent *aiv1beta1.Agent) (bool, error) {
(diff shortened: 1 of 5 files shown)


eranco74:
packet setup failed
/test ci/prow/edge-e2e-ai-operator-ztp

internal/controller/controllers/agent_controller.go
```diff
@@ -588,11 +594,31 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 	} else {
 		setConditionsUnknown(agent)
 	}
+	// get node
+	var node *corev1.Node
+	if funk.Contains([]models.HostStage{models.HostStageRebooting, models.HostStageJoined, models.HostStageConfiguring}, agent.Status.Progress.CurrentStage) {
```
ori-amizur, line 599: 1. Is it also relevant to day1? According to the code it seems it does. Day1 flow shouldn't be handled here.
2. For day2 flow, seems that only none platform is handled.  Is this the intention?

internal/controller/controllers/agent_controller.go
```diff
@@ -588,11 +594,31 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
...
+		node, err = r.getNode(ctx, agent)
+		if err != nil && !k8serrors.IsNotFound(err) {
+			r.Log.WithError(err).Errorf("agent %s/%s: failed to get node", agent.Namespace, agent.Name)
+			return ctrl.Result{}, err
+		}
+		if node != nil {
+			if isNodeReady(node) {
+				err = r.updateHostInstallProgress(ctx, h, models.HostStageDone)
```
ori-amizur, line 607: Day1 flow should not be handled here.

internal/controller/controllers/agent_controller.go
```diff
@@ -588,11 +594,31 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
...
+				agent.Status.Progress.CurrentStage = models.HostStageDone
+			} else {
+				err = r.updateHostInstallProgress(ctx, h, models.HostStageJoined)
+				agent.Status.Progress.CurrentStage = models.HostStageJoined
```
ori-amizur, line 611: Since the status/stage change logic is now separate from the CSR approval logic, maybe the status/stage calculation can be moved to a separate function ? 
Also consider to include the code [between lines 558-563](https://github.com/openshift/assisted-service/blob/3cc5fe4735485ef23b6e7dc6469157749e27379a/internal/controller/controllers/agent_controller.go#L558-L561)

internal/controller/controllers/agent_controller.go
```diff
@@ -588,11 +594,31 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
...
+				return ctrl.Result{}, err
+			}
+		}
+	}
 	if shouldAutoApproveCSRs {
-		alreadyApproved := r.tryApproveDay2CSRs(ctx, agent)
-		if alreadyApproved {
-			agent.Status.Progress.CurrentStage = models.HostStageDone
-		} else {
+		alreadyApproved := r.tryApproveDay2CSRs(ctx, agent, node)
```
ori-amizur, line 620: Since only none platform day2 CSRs are auto approved, the logic for updating status/stage will not work correctly for (day2 , BM) since no requeue will be done for BM

eranco74: Why would we need to requeue?
The reconcile should get called the same way it get's called for approving the day2 CSRs

eranco74: Hmmm, I think you are right (though it does seems to work without the requeue), we should requeue the all agents once they reach status `Rebooting` until they reach `Done`  to make sure we keep reconciling them

internal/controller/controllers/agent_controller.go
```diff
@@ -312,41 +312,46 @@ func (r *AgentReconciler) spokeKubeClient(ctx context.Context, clusterRef *aiv1b
...
 	// Even if node is already ready, we try approving last time
 	r.approveAIHostsCSRs(clients, agent, validateNodeCsr)
 
 	return !shouldApproveMoreCSRs
 }
 
+func (r *AgentReconciler) getNode(ctx context.Context, agent *aiv1beta1.Agent) (*corev1.Node, error) {
+	hostname := getAgentHostname(agent)
+
+	clients, err := r.spokeKubeClient(ctx, agent.Spec.ClusterDeploymentName)
```
carbonin, line 345: Can you try to reduce the number of times you create a client?

I think it would be better to create it earlier and pass the reference around.

filanov (commented):
Unless there is a backend event the controller will reconcile every 10h is that case handled here?

internal/controller/controllers/agent_controller.go
```diff
@@ -312,27 +312,17 @@ func (r *AgentReconciler) spokeKubeClient(ctx context.Context, clusterRef *aiv1b
...
-
-	// TODO: Node name might be FQDN and not just host name if cluster is IPv6
-	node, err := clients.GetNode(hostname)
+	node, err := r.getNode(agent, client)
 	if err != nil {
 		if !k8serrors.IsNotFound(err) {
-			r.Log.WithError(err).Errorf("Agent %s/%s: Failed to get node %s", agent.Namespace, agent.Name, hostname)
-			return false
+			r.Log.WithError(err).Errorf("agent %s/%s: failed to get node", agent.Namespace, agent.Name)
+			return true
```
ori-amizur, line 325: Why did this change ?

internal/controller/controllers/agent_controller.go
```diff
@@ -610,6 +609,40 @@ func (r *AgentReconciler) updateStatus(ctx context.Context, log logrus.FieldLogg
 	return ret, nil
 }
 
+func (r *AgentReconciler) UpdateDay2InstallPogress(ctx context.Context, h *models.Host, agent *aiv1beta1.Agent, spokeClient spoke_k8s_client.SpokeK8sClient) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
eranco74: > the update of the stage can be in the same place for day1 and day2

The day1 host update is determined by the installer running on the node and later by the assisted-installer-controller running inside the installed cluster.
Day2 host installation progress is different because we don't have the assisted-installer-controller to report the host progress, hence we need this code...

internal/controller/controllers/agent_controller.go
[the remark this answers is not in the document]
eranco74: bug. thanks

internal/controller/controllers/agent_controller.go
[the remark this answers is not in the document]
ori-amizur: This is correct, but this is already set by the code that changes the stage in the database.  Therefore, the stage is Kube API is just a reflection of what exists in the database.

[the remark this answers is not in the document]
eranco74: I agree that it's better, but I'm concerned that this will lead to a UT changes nightmare.
I can give it a try in another PR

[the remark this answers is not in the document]
eranco74: Hmm, actually in order for the stage in Kube API to reflect the stage we have in the database.
I'll need to fetch the agent from the DB again (or set it manually on the host object here).
I think there isn't much value in it, and perhaps the current implementation is clearer - in the case of day2 host we set the installation stage here (with some logic behind it) in the case of day 1 we just copy it from the host (under the `else` condition).

eranco74:
`Failed to create equinix device`
/test edge-e2e-metal-assisted-day2

ori-amizur (approved):

PR #4382, opened 2022-09-12
MGMT-11739: create spoke client using specified kubeconfig

danielerez:
HypershiftServiceConfig should contain a reference for a kubeconfig secret in the KubeconfigSecretRef property (the kubeconfig is for an hypershift control-plane).

This change introduces logic for:
* Fetching the referenced kubeconfig secret in Reconcile.
* Creating a SpokeK8sClient using that secret.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @slaviered 
/cc @mhrivnak 
/cc @carbonin 
/cc @filanov 

## Checklist

## Reviewers Checklist


internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -41,6 +48,9 @@ type HypershiftAgentServiceConfigReconciler struct {
 	// on the management cluster
 	NodeSelector map[string]string
 	Tolerations  []corev1.Toleration
+
+	SpokeK8sClientFactory spoke_k8s_client.SpokeK8sClientFactory
+	SpokeClient           spoke_k8s_client.SpokeK8sClient
```
carbonin, line 53: This should not be a field on the reconciler as the same reconciler can handle requests for multiple spokes at the same time.

This would cause some _very_ strange races and bugs.

carbonin (changes requested):

internal/controller/controllers/hypershiftagentserviceconfig_controller_test.go
```diff
@@ -51,31 +57,102 @@ var _ = Describe("HypershiftAgentServiceConfig reconcile", func() {
...
 
 	It("runs without error", func() {
-		hsc := newHSCDefault()
+		mockSpokeClientFactory.EXPECT().CreateFromSecret(gomock.Any()).Return(mockSpokeClient, nil)
 		res, err := hr.Reconcile(ctx, newHypershiftAgentServiceConfigRequest(hsc))
 		Expect(err).To(BeNil())
 		Expect(res).To(Equal(ctrl.Result{}))
 	})
+
+	Context("Create spoke cluster client", func() {
```
carbonin, line 111: You don't need a context here.

internal/spoke_k8s_client/spoke_k8s_client_manager.go
```diff
@@ -0,0 +1,44 @@
...
+type SpokeK8sClientManager interface {
+	AddFromSecret(secret *corev1.Secret) (SpokeK8sClient, error)
+}
+
+type spokeK8sClientManager struct {
+	clientFactory SpokeK8sClientFactory
+	clientMap     map[string]SpokeK8sClient
+}
+
+func NewSpokeK8sClientManager(log logrus.FieldLogger) SpokeK8sClientManager {
```
mhrivnak, line 21: A [general golang naming convention](https://go.dev/doc/effective_go#package-names) is to avoid stuttering in names. Here this will get used as `spoke_k8s_client.NewSpokeK8sClientManager()`. That would be better as `spoke_k8s_client.NewManager()`

FWIW there's also a convention to avoid snake-cases in package names and otherwise, but that seems to have appeared in a prior recent PR.

internal/spoke_k8s_client/spoke_k8s_client_manager.go
```diff
@@ -0,0 +1,44 @@
...
+	return &spokeK8sClientManager{
+		clientFactory: NewSpokeK8sClientFactory(log),
+		clientMap:     make(map[string]SpokeK8sClient),
+	}
+}
+
+// Add a SpokeK8sClient to the client map (create the client using specified secret).
+// If already created, return the existing client.
```
mhrivnak, line 29: This describes to different behaviors. It adds a client, but also returns a client regardless of whether it was added just now or not. WDYT about renaming this `FromSecret`, so it does not make implications about when or how the client was created?

internal/spoke_k8s_client/spoke_k8s_client_manager.go
```diff
@@ -0,0 +1,44 @@
...
+func (m *spokeK8sClientManager) AddFromSecret(secret *corev1.Secret) (SpokeK8sClient, error) {
+	client := m.clientMap[secret.Name]
```
mhrivnak, line 31: We need a lock around reading and writing this map since access could happen from multiple goroutines. Just acquiring a lock at the top of this function with a `defer` unlock is probably sufficient.

internal/spoke_k8s_client/spoke_k8s_client_manager.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// SpokeK8sClientManager maintains a set of spoke clients
+
+//go:generate mockgen --build_flags=--mod=mod -package=spoke_k8s_client -destination=mock_spoke_k8s_client_manager.go . SpokeK8sClientManager
+type SpokeK8sClientManager interface {
+	AddFromSecret(secret *corev1.Secret) (SpokeK8sClient, error)
+}
+
+type spokeK8sClientManager struct {
```
mhrivnak, line 16: What happens when a kubeconfig changes? We need a way to handle that. Storing a hash of the kubeconfig contents is one way that could be identified, so the old client could be evicted from the cache.

mhrivnak (changes requested):

internal/spoke_k8s_client/spoke_k8s_client_manager.go
carbonin: I think this cache key needs to be name+namespace.

internal/spoke_k8s_client/spoke_k8s_client_manager.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// SpokeK8sClientManager maintains a set of spoke clients
+
+//go:generate mockgen --build_flags=--mod=mod -package=spoke_k8s_client -destination=mock_spoke_k8s_client_manager.go . SpokeK8sClientManager
+type SpokeK8sClientManager interface {
```
carbonin, line 12: I would prefer we call this a "Cache" rather than a "Manager".

internal/spoke_k8s_client/spoke_k8s_client_manager.go
```diff
@@ -0,0 +1,44 @@
...
+	AddFromSecret(secret *corev1.Secret) (SpokeK8sClient, error)
```
carbonin, line 13: Is it worth considering making this a generic client cache rather than one that requires the caller create the client from a kubeconfig stored in a secret?

I could see us taking just the kubeconfig as input and using the hostname as a cache key, for example.

carbonin: If it's not worth the effort, maybe this belongs in the controller package rather than here, which is meant to be general purpose.

danielerez: yeah, just moved it the controllers package

internal/spoke_k8s_client/spoke_k8s_client_manager.go
danielerez: ok, makes sense.

internal/spoke_k8s_client/spoke_k8s_client_manager.go
danielerez: Yeah good point, added kubeconfig hash to the cache map.

internal/spoke_k8s_client/spoke_k8s_client_manager.go
danielerez: right, added.

internal/controller/controllers/spoke_client_cache.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+type spokeClientCache struct {
+	sync.Mutex
+	clientFactory spoke_k8s_client.SpokeK8sClientFactory
+	clientMap     map[types.NamespacedName]*spokeClient
+}
+
+type spokeClient struct {
+	sync.Mutex
```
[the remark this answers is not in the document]
danielerez: oops. yeah, no need for this.

internal/spoke_k8s_client/spoke_k8s_client_manager.go
danielerez: Sure, makes sense. Moved to controllers package, so it'd be controllers.NewSpokeClientCache

internal/controller/controllers/hypershiftagentserviceconfig_controller_test.go
```diff
@@ -51,31 +57,85 @@ var _ = Describe("HypershiftAgentServiceConfig reconcile", func() {
...
+			Scheme:       c.Scheme(),
+			Log:          common.GetTestLog(),
+			SpokeClients: mockSpokeClientCache,
+		}
+	}
+
 	BeforeEach(func() {
-		c = fakeclient.NewClientBuilder().WithScheme(GetKubeClientSchemes()).Build()
 		mockCtrl = gomock.NewController(GinkgoT())
+		mockSpokeClient = spoke_k8s_client.NewMockSpokeK8sClient(mockCtrl)
```
carbonin, line 92: I know you're not using it for anything currently, but you may eventually want to make this a fakeclient for better tests later.

internal/controller/controllers/spoke_client_cache.go
```diff
@@ -0,0 +1,79 @@
...
+
+//go:generate mockgen --build_flags=--mod=mod -package=controllers -destination=mock_spoke_client_cache.go . SpokeClientCache
+type SpokeClientCache interface {
+	Get(secret *corev1.Secret) (spoke_k8s_client.SpokeK8sClient, error)
+}
+
+type spokeClientCache struct {
+	sync.Mutex
+	clientFactory spoke_k8s_client.SpokeK8sClientFactory
+	clientMap     map[types.NamespacedName]*spokeClient
```
carbonin, line 22: Not a huge fan of using a struct as a map key, but I don't have enough expertise in exactly how go handles map keys to know how much of an issue this is.

I'd rather use a simple type. `types.NamespacedName` has a `.String()` method. Maybe use the output of that instead of the whole struct?

internal/controller/controllers/spoke_client_cache_test.go
```diff
@@ -0,0 +1,114 @@
...
+			mockSpokeFactory.EXPECT().CreateFromRawKubeconfig(gomock.Any()).Return(mockSpokeClient, nil)
+			client, err := clientCache.Get(kubeconfigSecret)
+
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(client).To(Equal(mockSpokeClient))
+		})
+
+		It("successfully returns an existing client", func() {
+			// create a client
+			mockSpokeFactory.EXPECT().CreateFromRawKubeconfig(gomock.Any()).Return(mockSpokeClient, nil)
```
carbonin, line 64: Is there any reason not to assert that this parameter is `BASIC_KUBECONFIG`?

internal/controller/controllers/spoke_client_cache_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+			// get created client
+			client, err = clientCache.Get(kubeconfigSecret)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(client).To(Equal(mockSpokeClient))
+		})
+
+		It("successfully creates a new client on hash mismatch", func() {
+			// create a client
+			mockSpokeFactory.EXPECT().CreateFromRawKubeconfig(gomock.Any()).Return(mockSpokeClient, nil)
```
carbonin, line 77: Same here, we can check this parameter.

internal/controller/controllers/spoke_client_cache_test.go
```diff
@@ -0,0 +1,114 @@
...
+			client, err := clientCache.Get(kubeconfigSecret)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(client).To(Equal(mockSpokeClient))
+
+			// create a client from a new kubeconfig
+			kubeconfigSecret.Data["kubeconfig"] = []byte("new")
+			mockSpokeFactory.EXPECT().CreateFromRawKubeconfig(gomock.Any()).Return(mockSpokeClient, nil)
```
carbonin, line 84: Same, this should be `[]byte("new")`, right?

internal/controller/controllers/spoke_client_cache_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/assisted-service/internal/spoke_k8s_client"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	testKubeconfigSecretName      = "test-secret"
```
carbonin, line 15: Any reason these can't be scoped in the `Context`?

Specifically this same constant is defined in the `"HypershiftAgentServiceConfig reconcile"` describe block.
If they were different values this might be confusing.

carbonin (commented):
A few mostly minor comments.

Looks good otherwise.

@mhrivnak did you have anything to add?

internal/controller/controllers/spoke_client_cache_test.go
danielerez: done

carbonin (approved):

PR #4399, opened 2022-09-14
MGMT-12063: Fix update cluster UMN when platform is different from baremetal/none

eliorerz:
We missed a case when first we have (creating or creating + updating) cluster with platform different than baremetal/none and next updating the cluster UMN to with sone value (true or false)

Issue: https://issues.redhat.com/browse/MGMT-12063

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

/cc @eranco74 
/cc @gamli75

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -1694,6 +1694,18 @@ func checkPlatformWrongParamsInput(platform *models.Platform, userManagedNetwork
 	return nil
 }
 
+func doesPlatformDifferentThanBaremetalOrNone(platform *models.Platform, cluster *common.Cluster) bool {
+	if platform != nil && *platform.Type != models.PlatformTypeBaremetal && *platform.Type != models.PlatformTypeNone {
+		return true
+	}
+
+	if platform == nil && *cluster.Platform.Type != models.PlatformTypeBaremetal && *cluster.Platform.Type != models.PlatformTypeNone {
+		return true
+	}
+
+	return false
+}
+
 func getActualUpdateClusterPlatformParams(platform *models.Platform, userManagedNetworking *bool, cluster *common.Cluster) (*models.Platform, *bool, error) {
 	if platform == nil && userManagedNetworking == nil {
 		return nil, nil, nil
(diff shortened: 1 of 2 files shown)


internal/bminventory/inventory.go
```diff
@@ -1694,6 +1694,18 @@ func checkPlatformWrongParamsInput(platform *models.Platform, userManagedNetwork
 	return nil
 }
 
+func doesPlatformDifferentThanBaremetalOrNone(platform *models.Platform, cluster *common.Cluster) bool {
+	if platform != nil && *platform.Type != models.PlatformTypeBaremetal && *platform.Type != models.PlatformTypeNone {
+		return true
+	}
+
+	if platform == nil && *cluster.Platform.Type != models.PlatformTypeBaremetal && *cluster.Platform.Type != models.PlatformTypeNone {
```
gamli75, line 1702: why do we need to check both cluster.Platform.Type & platform.Type?

eranco74: Because if the user didn't provide the platform in the request, and the platform is already set to let say vsphere, we should honer the request.
We only care about the UMN & platform alignment when the platform is Baremetal or None.

internal/bminventory/inventory_test.go
```diff
@@ -5133,6 +5133,37 @@ var _ = Describe("[V2ClusterUpdate] cluster", func() {
 					Expect(*actual.Platform.Type).To(Equal(models.PlatformTypeVsphere))
 				})
 
+				It("Update UMN=true while cluster platform already set to vsphere - success", func() {
```
eranco74, line 5136: Please add the following tests:
1. cluster platform none and an update request that set the cluster platform to vsphere + UMN=true
2. cluster platform none and an update request that set the cluster platform to vsphere + UMN=false
3. cluster platform baremetal and an update request that set the cluster platform to vsphere + UMN=false
4. cluster platform baremetal and an update request that set the cluster platform to vsphere + UMN=true

internal/bminventory/inventory.go
eliorerz: Because if the cluster platform type is currently set to vsphere and we just want to update UMN, it's once case .
Another case is where we set them both together.
On both of those cases we do not want to go and check all the if cases for none/bm platform

internal/bminventory/inventory_test.go
eliorerz: Done

eranco74 (commented):
/lgtm

PR #4401, opened 2022-09-15
MGMT-11751: ensure agent-install CRDs exist spoke cluster

danielerez:
In order to deploy assisted-service on a spoke hypershift cluster, the agent-install CRDs must first exist on it.
Hence, introduced the following flow to ensure each CRD exists (and not stale) on the spoke cluster:
* List the relevant CRDs on the management cluster.
* For each one, create/update on the spoke cluster.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Assignees

/cc @slaviered 
/cc @mhrivnak 
/cc @carbonin 
/cc @filanov 

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/v1beta1/hypershiftagentserviceconfig_types.go
+++ b/api/v1beta1/hypershiftagentserviceconfig_types.go
@@ -53,7 +53,7 @@ type HypershiftAgentServiceConfig struct {
 	Status HypershiftAgentServiceConfigStatus `json:"status,omitempty"`
 }
 
-//+kubebuilder:object:root=true
+// +kubebuilder:object:root=true
 // HypershiftAgentServiceConfigList contains a list of HypershiftAgentServiceConfigs
 type HypershiftAgentServiceConfigList struct {
 	metav1.TypeMeta `json:",inline"`
--- a/internal/controller/controllers/hypershiftagentserviceconfig_controller.go
+++ b/internal/controller/controllers/hypershiftagentserviceconfig_controller.go
@@ -19,18 +19,22 @@ package controllers
 import (
 	"context"
 	_ "embed"
+	"fmt"
 
 	aiv1beta1 "github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/internal/spoke_k8s_client"
 	logutil "github.com/openshift/assisted-service/pkg/log"
 	pkgerror "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/thoas/go-funk"
 	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
@@ -45,11 +49,15 @@ type HypershiftAgentServiceConfigReconciler struct {
 	NodeSelector map[string]string
 	Tolerations  []corev1.Toleration
 
+	// Namespace the operator is running in
+	Namespace string
+
 	SpokeClients SpokeClientCache
 }
 
 // +kubebuilder:rbac:groups=agent-install.openshift.io,resources=hypershiftagentserviceconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=agent-install.openshift.io,resources=hypershiftagentserviceconfigs/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
 
 func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	ctx := addRequestIdIfNeeded(origCtx)
@@ -73,10 +81,15 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
 	}
 
 	// Creating spoke client using specified kubeconfig secret reference
-	// TODO: use spoke client to create CRDs/CRs
-	_, err := hr.createSpokeClient(ctx, instance.Spec.KubeconfigSecretRef.Name, instance.Namespace)
+	spokeClient, err := hr.createSpokeClient(ctx, instance.Spec.KubeconfigSecretRef.Name, instance.Namespace)
 	if err != nil {
-		log.WithError(err).Error("Failed to create client using remote kubeconfig", req.NamespacedName)
+		log.WithError(err).Error("Failed to create client using specified kubeconfig", req.NamespacedName)
+		return ctrl.Result{}, err
+	}
+
+	// Ensure agent-install CRDs exist on spoke cluster and clean stale ones
+	if err := hr.syncSpokeAgentInstallCRDs(ctx, spokeClient); err != nil {
+		log.WithError(err).Error(fmt.Sprintf("Failed to sync agent-install CRDs on spoke cluster in namespace '%s'", instance.Namespace))
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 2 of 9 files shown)


internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -118,6 +130,34 @@ func (hr *HypershiftAgentServiceConfigReconciler) getKubeconfigSecret(ctx contex
 	return secret, nil
 }
 
+func (hr *HypershiftAgentServiceConfigReconciler) ensureSpokeAgentInstallCRDs(ctx context.Context) error {
+	// Fetch agent-install CRDs using in-cluster client
+	localCRDs := &apiextensionsv1.CustomResourceDefinitionList{}
+	listOpts := []client.ListOption{
+		client.MatchingLabels{fmt.Sprintf("operators.coreos.com/assisted-service-operator.%s", hr.Namespace): ""},
```
carbonin, line 137: CRDs aren't namespaced, right? Why would this need the be the namespace the operator is running in?

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -118,6 +130,34 @@ func (hr *HypershiftAgentServiceConfigReconciler) getKubeconfigSecret(ctx contex
...
+	}
+	if len(localCRDs.Items) == 0 {
+		return pkgerror.New("agent-install CRDs are not available")
+	}
+
+	// Ensure CRDs exist on the spoke cluster
+	for _, crd := range localCRDs.Items {
+		c := crd
+		c.ResourceVersion = "" // ResourceVersion should not be set on objects to be created
+		result, err := controllerutil.CreateOrUpdate(ctx, hr.SpokeClient, &c, func() error { return nil })
```
carbonin, line 150: This won't update anything I think.
We'd need to set the spec to what we want it to be in the mutate function.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -118,6 +130,34 @@ func (hr *HypershiftAgentServiceConfigReconciler) getKubeconfigSecret(ctx contex
...
+	if err := hr.Client.List(ctx, localCRDs, listOpts...); err != nil {
+		return pkgerror.Wrap(err, "Failed to list in-cluster CRDs")
+	}
+	if len(localCRDs.Items) == 0 {
+		return pkgerror.New("agent-install CRDs are not available")
+	}
+
+	// Ensure CRDs exist on the spoke cluster
+	for _, crd := range localCRDs.Items {
+		c := crd
```
carbonin, line 148: Pretty sure you don't need this here.

internal/controller/controllers/hypershiftagentserviceconfig_controller_test.go
```diff
@@ -155,4 +183,30 @@ var _ = Describe("HypershiftAgentServiceConfig reconcile", func() {
 			Expect(err.Error()).To(ContainSubstring("Failed to create client"))
 		})
 	})
+
+	Context("Ensure agent-install CRDs exist on spoke clister", func() {
```
carbonin, line 187: You don't need a context here as you're not adding any shared context.

Also you should add a test for update.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -41,10 +51,17 @@ type HypershiftAgentServiceConfigReconciler struct {
...
+	// Namespace the operator is running in
+	Namespace string
+
+	SpokeK8sClientFactory spoke_k8s_client.SpokeK8sClientFactory
+	SpokeClient           spoke_k8s_client.SpokeK8sClient
 }
 
 // +kubebuilder:rbac:groups=agent-install.openshift.io,resources=hypershiftagentserviceconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=agent-install.openshift.io,resources=hypershiftagentserviceconfigs/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;create;watch
```
mhrivnak, line 64: I don't think we need or want the ability to create CRDs on the host cluster, right?

mhrivnak (changes requested):

internal/controller/controllers/hypershiftagentserviceconfig_controller_test.go
danielerez: Maybe 'Describe' instead? Just for grouping these separately.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
danielerez: Not sure what we want to update actually, maybe just the spec.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
danielerez: Right, but the label includes the namespace (i.e. operators.coreos.com/assisted-service-operator.assisted-installer: ""). Using it as an indication for finding the relevant CRDs, didn't find a cleaner way:/

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
mhrivnak: We should figure that out so we can drop this permission on the hub.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
carbonin: But is that actually the namespace the operator is installed in?
I'm not sure why the CRD would have a label related to where the operator is running.

That said, is there some way to query all crds for a particular api group/version?
For example, anything that has `spec: group: "agent-install.openshift.io"` in the CRD?

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -118,6 +130,34 @@ func (hr *HypershiftAgentServiceConfigReconciler) getKubeconfigSecret(ctx contex
 	return secret, nil
 }
 
+func (hr *HypershiftAgentServiceConfigReconciler) ensureSpokeAgentInstallCRDs(ctx context.Context) error {
```
carbonin, line 133: Does this need to ensure the lists are exactly the same?
For example, if we add v1beta2, do need to delete the v1beta1 CRDs from the spoke?

We would need a "true" desired state calculation in that case where we list the CRDs on both the hub and spoke to ensure they're the same.

danielerez: Right. Probably a rare (uninteresting?) scenario, but added logic for deleting stale CRDs on spoke clusters.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
danielerez: sure, removed. WIll test again without it.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
danielerez: Added mutate func, so need it now :) I.e. to avoid: "Using the variable on range scope `crd` in function literal"

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -113,6 +122,69 @@ func (hr *HypershiftAgentServiceConfigReconciler) getKubeconfigSecret(ctx contex
...
+	// Fetch agent-install CRDs using spoke client
+	spokeCRDs, err := hr.getAgentInstallCRDs(ctx, spokeClient)
+	if err != nil {
+		return err
+	}
+
+	// Remove spoke CRDs that don't exist locally in-cluster
+	for _, item := range spokeCRDs.Items {
+		crd := item
+		existsInCluster := funk.Contains(localCRDs.Items, func(localCRD apiextensionsv1.CustomResourceDefinition) bool {
```
carbonin, line 164: Is this comparing the entire CRD object?

Feels like that might be a bit fragile. Is there some simpler comparison we can do that won't fail if something trivial (like a status) happens to be different?

danielerez: /The comparison is actually by CRD name (in the compare func). Any other property worth comparing

carbonin: :facepalm: Missed that entirely.

Nope I think that covers it.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
mhrivnak: Please run this by the OLM team. They of course have experience managing the lifecycle of CRDs and may have some useful insights, especially in case this is more complex to handle than is obvious.

danielerez: Started a thread with OLM team.
btw, for reading the CRDs we can also go with the approach used by [registration-operator ](https://github.com/open-cluster-management-io/registration-operator/blob/79c6b55faef94d1a456cbfaeaa67f2d9f21dc89a/pkg/operators/clustermanager/controllers/clustermanagercontroller/clustermanager_controller.go#L416) of using "[embed](https://pkg.go.dev/embed#FS.ReadDir)" for the relevant [yamls](https://github.com/open-cluster-management-io/registration-operator/blob/dcf93fa3b9d6a55dd7f800b2af6a9e42d3efa01a/manifests/fs.go).

filanov (commented):
do you think that we need to add a condition that will inform the user about the sync?

danielerez:
> do you think that we need to add a condition that will inform the user about the sync?

yeah, probably worth it. added a condition for sync failure.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -73,14 +82,36 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
...
 	if err != nil {
-		log.WithError(err).Error("Failed to create client using remote kubeconfig", req.NamespacedName)
+		log.WithError(err).Error("Failed to create client using specified kubeconfig", req.NamespacedName)
+		return ctrl.Result{}, err
+	}
+
+	// Ensure agent-install CRDs exist on spoke cluster and clean stale ones
+	if err := hr.syncSpokeAgentInstallCRDs(ctx, spokeClient); err != nil {
+		log.WithError(err).Error(fmt.Sprintf("Failed to sync agent-install CRDs on spoke cluster in namespace '%s'", instance.Namespace))
+		conditionsv1.SetStatusConditionNoHeartbeat(&instance.Status.Conditions, conditionsv1.Condition{
```
filanov, line 94: it's problematic to set single condition, usually when reporting a status you will set all the condition in each reconcile, maybe a different pattern is required here, something similar to what we have in the other controllers that you have a function that update the status with all the conditions.
I'm not sure how it will feet, we can split it into another PR if needed.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -73,14 +82,36 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
...
+			Type:    aiv1beta1.ConditionReconcileCompleted,
+			Status:  corev1.ConditionFalse,
+			Reason:  aiv1beta1.ReasonSpokeClusterCRDsSyncFailure,
+			Message: "Failed to sync agent-install CRDs on spoke cluster.",
```
filanov, line 98: consider adding the the `err` here as well, otherwise users will not know if it's something that they can fix or not. 
for example if it's internal issue then no, but if there is no connection to the other cluster then maybe it's something that they can check at least.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -73,14 +82,36 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
...
+		})
+		if statusErr := hr.Status().Update(ctx, instance); statusErr != nil {
+			log.WithError(err).Error("Failed to update status")
+			return ctrl.Result{}, statusErr
+		}
 		return ctrl.Result{}, err
 	}
 
-	return ctrl.Result{}, nil
+	conditionsv1.SetStatusConditionNoHeartbeat(&instance.Status.Conditions, conditionsv1.Condition{
```
filanov, line 107: same comment as above about the conditions.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -113,6 +144,88 @@ func (hr *HypershiftAgentServiceConfigReconciler) getKubeconfigSecret(ctx contex
...
+		crd := item
+		c := crd.DeepCopy()
+		c.ResourceVersion = "" // ResourceVersion should not be set on objects to be created
+		mutate := func() error {
+			c.Spec = crd.Spec
+			c.Annotations = crd.Annotations
+			c.Labels = crd.Labels
+			return nil
+		}
+		result, err1 := controllerutil.CreateOrUpdate(ctx, spokeClient, c, mutate)
```
filanov, line 183: no reason to call it `err1` right? it can be just `err`

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
danielerez: ok then I'll remove the condition for this PR for now, and add it separately. It would require some work to do it as in cd_ctrl: https://github.com/openshift/assisted-service/blob/ef1a92878d313976b142abf6dbd0909639cfe7f5/internal/controller/controllers/clusterdeployments_controller.go#L1943

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
danielerez: yep, no need now.

api/v1beta1/hypershiftagentserviceconfig_types.go
```diff
@@ -22,6 +22,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// ConditionType related to our reconcile loop in addition to all the reasons
+// why ConditionStatus could be true or false.
+const (
+	// ReasonSpokeClusterCRDsSyncFailure when agent-install CRDs sync fails on spoke cluster.
+	ReasonSpokeClusterCRDsSyncFailure string = "SpokeClusterCRDsSyncFailure"
```
filanov, line 29: not used, you can probably delete

danielerez: right, removed.

filanov (approved):

PR #4402, opened 2022-09-15
MGMT-12052: Swagger changes for kernel parameters support

ori-amizur:

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @carbonin 
/cc @mkowalski


swagger.yaml
```diff
@@ -6491,6 +6496,12 @@ definitions:
         x-nullable: false
         default: 'x86_64'
         description: The CPU architecture of the image (x86_64/arm64/etc).
+      discovery_kernel_parameters:
+        description: List of string values to be passed to discovery image as kernel parameters.
```
mkowalski, line 6500: Sounds good, but not clear to me why we have two different descriptions for the same field?

1) `Json formatted string array representing the discovery image kernel parameters.`
2) `List of string values to be passed to discovery image as kernel parameters.`

swagger.yaml
```diff
@@ -6419,6 +6419,11 @@ definitions:
         x-nullable: false
         default: 'x86_64'
         description: The CPU architecture of the image (x86_64/arm64/etc).
+      discovery_kernel_parameters:
+        type: string
+        x-nullable: true
+        x-go-custom-tag: gorm:"type:text"
+        description: Json formatted string array representing the discovery image kernel parameters.
```
mkowalski, line 6426: nit: `JSON` and not `Json` because we already use the former around our API description, so we should not introduce a new way

swagger.yaml
ori-amizur: In the parameters to the API call we get them as array, while in the database it is stored as JSON formatted string

swagger.yaml
```diff
@@ -6491,6 +6496,12 @@ definitions:
         x-nullable: false
         default: 'x86_64'
         description: The CPU architecture of the image (x86_64/arm64/etc).
+      discovery_kernel_parameters:
+        description: List of string values to be passed to discovery image as kernel parameters.
+        type: array
+        items:
+          type: string
+          pattern: '^[^ \n\t=(){}\[\]]+(?:=[^ \n\t=(){}\[\]]+)?$'
```
carbonin, line 6504: What is this validating?
Is this some well-known pattern for a kernel parameter syntax?

carbonin: FWIW this is the link I found when trying to implement kernel parameters at installation time the first time around: https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html

ori-amizur: Actually, I didn't find any pattern.  Mainly it eliminates the characters that I thought cannot appear in a kernel parameter.  Maybe we can make it more restrictive, but since I didn't find such pattern, I set this one.
I am open for suggestions on it.

carbonin: So it's a bit hard for me to read without a description, but it looks like excluding `=` would cause an issue at least. Based on the link I sent it seems whitespace would also be okay if it's enclosed in quotes.

We'd expect people to pass things like `rd.net.timeout.carrier=60`, right? This was what prompted the epic originally IIRC.
I've personally used `rd.break=pre-mount` to debug issues with the minimal iso not being able to pull rootfs.

What are you thinking would be the elements in this array? Maybe give some examples in the swagger and we can at least ensure they conform to this pattern.

Alternatively I'd say to lean less restrictive or no regex at all unless we can find actual guidelines for what is and isn't allowed. Worst case scenario the machine doesn't boot and it should be pretty clear if the kernel cmd line failed to parse.

Maybe also investigate if coreos is doing any validation when they embed these parameters into the iso since they already have the feature.

carbonin: Also in that doc `:` seems to be okay:
```
For example one can add to the command line following parameter:

isolcpus=1,2,10-20,100-2000:2/25
```

ori-amizur: All the examples that were mentioned above were accepted by the regex.
Concerning `=`, parameters can have a form `<name>` or `<name>=<value>` but only single `=` can appear in a parameter and it is not at the beginning or the end.  Right? If yes then it is accepted by the regex.
BTW, the regex can be tested [here](https://regex101.com/r/3Qp7IS/1).
Will try to add explanation for the regex.

carbonin: The regex as is doesn't seem to allow for spaces in quoted values such as:

`param="spaces in here"`

ori-amizur: Should it ?

ori-amizur: OK.  Saw [here](https://codebrowser.dev/linux/linux/lib/cmdline.c.html#next_arg) it should be supported.  Actually is provides information what should be the pattern  of the parameters.  Will update.

avishayt (approved):

PR #4414, opened 2022-09-19
OCPBUGS-1482: Don't override schedulableMasters unnecessarily

zaneb:
When generating the Ignition files, the installer already sets schdulableMasters to true when there are no worker nodes (i.e. in the SNO and compact cluster topologies. (See
https://github.com/openshift/installer/pull/2004) Therefore it is unnecessary to override it here (though it may be preferred to avoid a warning log from the installer).

Since https://github.com/openshift/installer/pull/6247, attempting to override the schedulableMasters setting causes installation to fail, because there are two manifests of the same type and name that conflict.

Since we don't need to set this override when the installer would already do it, avoid doing so and triggering the error when the value is determined by the number of hosts rather than explicitly set by the user.

The conflict still needs to be resolved so that the user can enable schedulableMasters, but this at least allows the SNO and compact topologies to install OpenShift 4.12 again.

This partially reverts commit c45f3699d83665a5e27a00939a9ff40907f39111.

https://issues.redhat.com/browse/OCPBUGS-1482

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -120,7 +120,14 @@ func IsImportedCluster(cluster *Cluster) bool {
 }
 
 func AreMastersSchedulable(cluster *Cluster) bool {
-	return swag.BoolValue(cluster.SchedulableMastersForcedTrue) || swag.BoolValue(cluster.SchedulableMasters)
+	// If the topology forces schedulable masters (i.e. there are no workers),
+	// SchedulableMastersForcedTrue will be true, but also it will be enabled
+	// by default in the installer. Since overriding it currently causes a
+	// failure due to a conflict starting with 4.12, we prefer to avoid it. We
+	// only need to override the installer when the user has set
+	// SchedulableMasters explicitly and that is not already the default in the
+	// installer.
+	return !swag.BoolValue(cluster.SchedulableMastersForcedTrue) && swag.BoolValue(cluster.SchedulableMasters)
 }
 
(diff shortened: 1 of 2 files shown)


carbonin:
> When generating the Ignition files, the installer already sets schdulableMasters to true when there are no worker nodes (i.e. in the SNO and compact cluster topologies

What versions is this true for?
Are we going to end up not setting schedulable masters for compact clusters for older versions that we still support like 4.8?

carbonin:
>Are we going to end up not setting schedulable masters for compact clusters for older versions that we still support like 4.8?

Looks like the change to the installer is in 4.8 as openshift/installer@64f96df25fe24adc53a391183f6e50a6ed7d86fc

carbonin (approved):

carbonin:
Let's make sure we handle this properly in a follow up.
Whatever "properly" ends up being.

That said it doesn't seem like this will cause a problem and it will unblock the agent team so I'm good with merging.

carbonin:
This is indeed referencing a bug, and it's an issue regarding the agent installer targeted to 4.12.

IMO this shouldn't be an invalid bug.
I'll manually remove the label and hope it doesn't get added back.

PR #4420, opened 2022-09-21
MGMT-12083: Create enhancement document for discovery kernel parameters

ori-amizur:

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @filanov 
/cc @carbonin

### DIFF
--- a/docs/enhancements/add-kernel-parameters.md
+++ b/docs/enhancements/add-kernel-parameters.md
@@ -1,0 +1,128 @@
+---
+title: add-discovery-kernel-arguments
+authors:
+- "@oamizur"
+  creation-date: 2022-09-20
+  last-updated: 2022-10-23
+---
+
+# Add discovery kernel arguments
+
+## Summary
+
+Some environments require additional kernel arguments to boot host machines.  
+We would like to support additional kernel arguments while booting the discovery image.
+
+## Motivation
+
+Previously, adding kernel arguments was required by several customers in order to adapt to various environments.
+For example **rd.net.timeout.carrier** may be required in order for either static networking to be applied or 
+to successfully get a dhcp address before pulling the rootfs.
+### Goals
+Add capability to manipulate kernel arguments in the following ways:
+- Append: Append new argument to the existing argument set
+- Replace: Replace an existing kernel argument with a new value
+- Delete: Delete a kernel argument from the existing argument set 
+
+Image types: Full ISO, minimal ISO, iPXE
+
+
+ Deployment types: Cloud, operator
+
+### Non-Goals
+
+## Proposal
+
+The discovery kernel arguments will be stored in the infra-env database record in a designated JSON encoded field.
+During infra-env registration and update the full set of the kernel arguments can be set.
+Since the kernel arguments are defined at infra-env level, if different set of arguments is needed for
+different hosts, then different infra-env is needed (one per kernel arguments set). 
+
+Setting the arguments can either be done by either REST API or Kube API (Part of InfraEnv CR)
+### User Stories
+
+#### Story 1
+
+When having slow NIC, increase the timeout to have the IP address configured.
+
+### Implementation Details/Notes/Constraints
+
+#### iPXE implementation
+
+iPXE implementation should be quite straight forward.  The arguments will be concatenated at the end of the 
+**kernel** command which is part of the iPXE script.
+
+#### Minimal and full ISO
+
+When streaming the discovery ISO file, the kernel arguments are included
+as part of some files: grub.cfg and isolinux.cfg
+
+Each of these files contains an area where the kernel arguments can be
+embedded. This can be identified by the regex **"(\n#+) COREOS_KARG_EMBED_AREA"**.
+The arguments should overwrite this are from the beginning and should
+not pass the last # (hash sign). The added kernel will be injected in the beginning.
+These arguments will be seperated by space **(" ")**.  When embedding the kernel arguments,
+the size of the above files will not change.
+
+Starting from version 4.10, there is a file that contains the file names,
+while in older versions the file does not exist. If the file exists, the
+files to embed the kernel arguments are taken from this config file.
+Otherwise, the default files **[/EFI/redhat/grub.cfg, /isolinux/isolinux.cfg]**
+are used.
+
+
+We will need the image service to query the assisted service to get the kernel arguments.
+This will require authentication.  The authentication methodology may have 2 variants:  
+
+1. When running in cloud
+2. When running in disconnected environment
+
+The image service already queries the assisted service.  We will use the same authentication 
+methods.  We will need to verify that the infra-env endpoint may be accessed using the same authentication
+mechanism.
+
+#### UI
+
+Eventually, UI may support kernel arguments modification in all places that infra-env is manipulated.
+#### Kube API
+
+The discovery kernel arguments should be defined in the spec section in InfraEnv CR.  
+They will have the following format:
+
+In InfraEnv Spec section:
+
+```sh
+DiscoveryKernelArguments []KernelArgument `json:"discoveryKernelArguments,omitempty"`
+```
+
+In addition there is the definition of **KernelArgument**:
+
+```sh
+type KernelArgument struct {
+
+	Operation string `json:"operation,omitempty"`
+
+	Value string `json:"value,omitempty"`
+}
+```
+
+
+
+### Risks and Mitigations
+
+## Design Details [optional]
+
+### Open Questions
+Coreos Installer already supports the operations mentioned in this document but does not 
+have full-support for streaming.  Once it will have such support, the image service will move to 
+use the Coreos Installer.  It will be done not as part of the effort to add discovery kernel parameters.
+
+Here is a link to the Coreos Installer kernel arguments modification reference:
+https://coreos.github.io/coreos-installer/cmd/iso/#coreos-installer-iso-kargs-modify
+### UI Impact
+
+### Test Plan
+
+## Drawbacks
+
+## Alternatives


docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,81 @@
...
+
+iPXE implementation should be quite straight forward.  The arguments will be concatenated at the end of the 
+**kernel** command which is part of the iPXE script.
+
+#### Minimal and full ISO
+
+The embedding of the kernel arguments into the ISO is an issue that needs to be investigated.
+
+We will also might need the image service to query the assisted service to get the kernel arguments.
+This will require authentication.  The authentication methodology may have 2 variants:
```
avishayt, line 57: The image service already queries the assisted service to get the ignition info

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,81 @@
...
+
+We will also might need the image service to query the assisted service to get the kernel arguments.
+This will require authentication.  The authentication methodology may have 2 variants:  
+
+1. When running in cloud
+2. When running in disconnected environment
+
+#### UI
+
+UI will need to add support for kernel arguments in all places that infra-env is manipulated.
```
avishayt, line 64: Maybe eventually, but until now we didn't expose these advanced settings in the UI

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,81 @@
...
+### Implementation Details/Notes/Constraints
+
+#### iPXE implementation
+
+iPXE implementation should be quite straight forward.  The arguments will be concatenated at the end of the 
+**kernel** command which is part of the iPXE script.
+
+#### Minimal and full ISO
+
+The embedding of the kernel arguments into the ISO is an issue that needs to be investigated.
```
avishayt, line 54: Where does it go in the ISO?  Currently, we know how to embed ignition.

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,81 @@
...
+For example **rd.net.timeout.carrier** may be required in order for either static networking to be applied or 
+to successfully get a dhcp address before pulling the rootfs.
+### Goals
+Add capability to set additional kernel arguments to discovery image.
+
+Support the following scenarios:
+- Full ISO image
+- Minimal ISO image
+- iPXE
+- Cloud/operator (when considering authentication from image service to assisted-service)
```
avishayt, line 28:
```suggestion
- Image types: Full ISO, minimal ISO, iPXE
- Deployment types: Cloud, operator
```

docs/enhancements/add-kernel-parameters.md
ori-amizur: Not sure yet

docs/enhancements/add-kernel-parameters.md
ori-amizur: OK

docs/enhancements/add-kernel-parameters.md
ori-amizur: OK

docs/enhancements/add-kernel-parameters.md
ori-amizur: OK

docs/enhancements/add-kernel-parameters.md
carbonin: There is an embed area reserved for kernel arguments, but it's within a file.
For ignition we currently find the file and use that as the offset to start embedding.

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,82 @@
...
+#### iPXE implementation
+
+iPXE implementation should be quite straight forward.  The arguments will be concatenated at the end of the 
+**kernel** command which is part of the iPXE script.
+
+#### Minimal and full ISO
+
+The embedding of the kernel arguments into the ISO is an issue that needs to be investigated.
+
+We will also might need the image service to query the assisted service to get the kernel arguments.
```
carbonin, line 53:
```suggestion
We will need the image service to query the assisted service to get the kernel arguments.
```

carbonin (commented):
I think it would be good to link to and describe the behavior of `coreos-installer iso kargs`.

We should compare how our approach will behave with what is available through that CLI.

docs/enhancements/add-kernel-parameters.md
avishayt: When you say "it's within a file", do you mean that it's part of a file as opposed to an entire file?  If so, is there an easy way to determine the offset within that file?

carbonin: > When you say "it's within a file", do you mean that it's part of a file as opposed to an entire file?

Yes

> If so, is there an easy way to determine the offset within that file?

I'm not sure. I can think of some things we might do, but it would be good to investigate what coreos is doing to sort out the "right" way.

ori-amizur: Added description how it is done

docs/enhancements/add-kernel-parameters.md
ori-amizur: OK

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,98 @@
...
+### Implementation Details/Notes/Constraints
+
+#### iPXE implementation
+
+iPXE implementation should be quite straight forward.  The arguments will be concatenated at the end of the 
+**kernel** command which is part of the iPXE script.
+
+#### Minimal and full ISO
+
+hen streaming the discovery ISO file, the kernel arguments are included
```
oourfali, line 52: s/hen/When

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,98 @@
...
+- Image types: Full ISO, minimal ISO, iPXE
+- Deployment types: Cloud, operator
+
+### Non-Goals
+Manipulation of existing kernel parameters (update/delete).  Update and delete of exiting
+kernel arguments in ISO file is supported by Coreos Installer, but this implementation will not support it. 
+
+## Proposal
+
+The discovery kernel arguments will be stored in the infra-env database record in a designated JSON encoded field.
```
oourfali, line 33: So if I need different args for different hosts, I'll need to create two infra envs?
That's a limitation. reasonable one I guess, but still a limitation.

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,98 @@
...
+to successfully get a dhcp address before pulling the rootfs.
+### Goals
+Add capability to set additional kernel arguments to discovery image.
+
+- Image types: Full ISO, minimal ISO, iPXE
+- Deployment types: Cloud, operator
+
+### Non-Goals
+Manipulation of existing kernel parameters (update/delete).  Update and delete of exiting
+kernel arguments in ISO file is supported by Coreos Installer, but this implementation will not support it.
```
avishayt, line 29: Why not?

ori-amizur: We decided this during the design meeting.

avishayt: IIRC we said that we *could* defer it as long as adding it later won't break the current API.  Will it?

ori-amizur: I don't recall that we said it will not break the current API. As I see it now, the API will have to change in order to support it.

avishayt: We can't plan to have API breakages.  It's fine if we don't implement update/delete right now, but we must be able to add them without breaking users.  This is a blocker as far as I'm concerned.

carbonin: FWIW this API was already merged in https://github.com/openshift/assisted-service/pull/4402 and implemented in https://github.com/openshift/assisted-service/pull/4425

Is there something we want to do now to prevent this from getting to stage/prod or are we okay just as long as it doesn't get in the UI?

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,101 @@
...
+Previously, adding kernel arguments was required by several customers in order to adapt to various environments.
+For example **rd.net.timeout.carrier** may be required in order for either static networking to be applied or 
+to successfully get a dhcp address before pulling the rootfs.
+### Goals
+Add capability to set additional kernel arguments to discovery image for append.
+
+- Image types: Full ISO, minimal ISO, iPXE
+- Deployment types: Cloud, operator
+
+API which supports full set of operations as defined by Coreos Installer: append, replace, delete.
```
carbonin, line 27: Should this replace the first line in this section?

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Non-Goals
+Manipulation of existing kernel parameters (replace/delete).  Replace and delete of exiting
```
carbonin, line 30: Given that this is now a goal it shouldn't be a non-goal.

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,101 @@
...
+as part of some files: grub.cfg and isolinux.cfg
+
+Each of these files contains an area where the kernel arguments can be
+embedded. This can be identified by the regex **"(\n#+) COREOS_KARG_EMBED_AREA"**.
+The arguments should overwrite this are from the beginning and should
+not pass the last # (hash sign). The added kernel will be injected in the beginning.
+These arguments will be seperated by space **(" ")**.  When embedding the kernel arguments,
+the size of the above files will not change.
+
+In new Openshift versions, there is a file that contains the file names,
```
carbonin, line 65: Maybe specify the version this was added just for reference? Not sure how much it matters, but "new" is relative and this document will be around for a while I assume.

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,101 @@
...
+
+We will need the image service to query the assisted service to get the kernel arguments.
+This will require authentication.  The authentication methodology may have 2 variants:  
+
+1. When running in cloud
+2. When running in disconnected environment
+
+The image service already queries the assisted service.  We will probably use the same authentication 
+methods.  We will need to verify that the infra-env endpoint may be accessed using the same authentication
+mechanism.
```
carbonin, line 80: I think we're sure of all of this now, right?

You can drop "probably" and state directly what we're doing.
If there's more to investigate I think this should be moved to "Open Questions"

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### UI
+
+Eventually, UI may support kernel arguments modification in all places that infra-env is manipulated.
+#### Kube API
+
+The kernel arguments should be defined in the spec section in InfraEnv CR.
```
carbonin, line 87: @avishayt @mhrivnak do you think it's worth trying to define this API here or is this enough and we can debate what this will look like in the PR that adds it?

docs/enhancements/add-kernel-parameters.md
avishayt: Agree, they can be implemented as separate tasks if we choose but can be covered by the same enhancement.

docs/enhancements/add-kernel-parameters.md
avishayt: Worth stating explicitly.

docs/enhancements/add-kernel-parameters.md
avishayt: Better to have it here, as it may impact other text in this document.

docs/enhancements/add-kernel-parameters.md
ori-amizur: Done

docs/enhancements/add-kernel-parameters.md
ori-amizur: Changed

docs/enhancements/add-kernel-parameters.md
ori-amizur: The API will only support all operations, but only the append capability will be implemented.

docs/enhancements/add-kernel-parameters.md
ori-amizur: @avishayt you indicated that the API needs to be stable but the backend implementation may change later.  Therefore, this is still non-goal.

docs/enhancements/add-kernel-parameters.md
ori-amizur: OK

docs/enhancements/add-kernel-parameters.md
ori-amizur: OK

docs/enhancements/add-kernel-parameters.md
ori-amizur: OK

docs/enhancements/add-kernel-parameters.md
avishayt: It's a non-goal for the current task that you're implementing.  But it is a goal of this enhancement that should cover delete and replace as well.

ori-amizur: Changed

avishayt (approved):

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,128 @@
...
+1. When running in cloud
+2. When running in disconnected environment
+
+The image service already queries the assisted service.  We will use the same authentication 
+methods.  We will need to verify that the infra-env endpoint may be accessed using the same authentication
+mechanism.
+
+#### UI
+
+Eventually, UI may support kernel arguments modification in all places that infra-env is manipulated.
```
jkilzi, line 86: Since a "different set of arguments is needed for different hosts, then different infra-env is needed (one per kernel arguments set)" (and given the infra-env is not exposed through the UI) then in the SaaS UI probably we should prompt the user to enter the kargs at the new cluster step

docs/enhancements/add-kernel-parameters.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Open Questions
+Coreos Installer already supports the operations mentioned in this document but does not 
+have full-support for streaming.  Once it will have such support, the image service will move to 
+use the Coreos Installer.  It will be done not as part of the effort to add discovery kernel parameters.
+
+Here is a link to the Coreos Installer kernel arguments modification reference:
+https://coreos.github.io/coreos-installer/cmd/iso/#coreos-installer-iso-kargs-modify
+### UI Impact
+
```
jkilzi, line 123: SaaS UI:
- What validations should be performed before sending the kernel args?
- If users modify the kernel args, will they be required to discover the hosts again?

PR #4424, opened 2022-09-21
MGMT-11293: Added default network type setting during registration

danmanor:

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -125,6 +125,7 @@ type Config struct {
 }
 
 const minimalOpenShiftVersionForSingleNode = "4.8.0-0.0"
+const minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes = "4.12.0-0.0"
 
 type Interactivity bool
 
@@ -340,7 +341,33 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
 		}
 	}
 
+	params.NewClusterParams.NetworkType, err = getDefaultNetworkType(params)
+	if err != nil {
+		return params, err
+	}
+
 	return params, nil
+}
+
+// If the cluster is SNO or the OpenShift version >= 4.12.0-0.0,
+// the network type will be set to "OVNKubernetes", otherwise it will be set to "OpenShiftSDN"
+func getDefaultNetworkType(params installer.V2RegisterClusterParams) (*string, error) {
+	if params.NewClusterParams.NetworkType != nil {
+		return params.NewClusterParams.NetworkType, nil
+	}
+
+	isOpenShiftVersionRecentEnough, err := common.VersionGreaterOrEqual(swag.StringValue(params.NewClusterParams.OpenshiftVersion), minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes)
+	if err != nil {
+		return nil, err
+	}
+
+	isSingleNodeCluster := swag.StringValue(params.NewClusterParams.HighAvailabilityMode) == models.ClusterCreateParamsHighAvailabilityModeNone
+
+	if isOpenShiftVersionRecentEnough || isSingleNodeCluster {
+		return swag.String(models.ClusterCreateParamsNetworkTypeOVNKubernetes), nil
+	} else {
+		return swag.String(models.ClusterCreateParamsNetworkTypeOpenShiftSDN), nil
+	}
 }
 
 func (b *bareMetalInventory) validateRegisterClusterInternalParams(params *installer.V2RegisterClusterParams, log logrus.FieldLogger) error {
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -10018,6 +10018,110 @@ var _ = Describe("TestRegisterCluster", func() {
 		ctrl.Finish()
 	})
 
+	// Testing the default network type setting during register flow, with various combinations of platforms,
+	// OpenShift version and high availability modes
+	// Note: SNO is supported starting OpenShift v4.8, and currently only supports Platform "none"
+	It("Default Network Type Setting", func() {
+		type clusterParamsForTest struct {
+			OpenShiftVersion     *string
+			HighAvailabilityMode *string
+			DesiredNetworkType   string
+			Platform             models.Platform
+		}
+
+		tests := []clusterParamsForTest{
+			// OpenShift version >= 4.12.0-0.0, not SNO => OVNKubernetes
+			{
+				OpenShiftVersion:     swag.String("4.13"),
+				HighAvailabilityMode: swag.String(models.ClusterCreateParamsHighAvailabilityModeFull),
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOVNKubernetes,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeBaremetal)},
+			},
+			// OpenShift version >= 4.12.0-0.0, SNO => OVNKubernetes
+			{
+				OpenShiftVersion:     swag.String("4.14.2"),
+				HighAvailabilityMode: swag.String(models.ClusterCreateParamsHighAvailabilityModeNone),
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOVNKubernetes,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeNone)},
+			},
+			// OpenShift version < 4.12.0-0.0, not SNO => OpenShiftSDN
+			{
+				OpenShiftVersion:     swag.String("4.10.3"),
+				HighAvailabilityMode: swag.String(models.ClusterCreateParamsHighAvailabilityModeFull),
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOpenShiftSDN,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeBaremetal)},
+			},
+			// OpenShift version < 4.12.0-0.0, SNO => OVNKubernetes
+			{
+				OpenShiftVersion:     swag.String("4.11.17"),
+				HighAvailabilityMode: swag.String(models.ClusterCreateParamsHighAvailabilityModeNone),
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOVNKubernetes,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeNone)},
+			},
+			// OpenShift version >= 4.12.0-0.0, HighAvailability = nil => OVNKubernetes
+			{
+				OpenShiftVersion:     swag.String("4.13.2"),
+				HighAvailabilityMode: nil,
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOVNKubernetes,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeBaremetal)},
+			},
+			// OpenShift version < 4.12.0-0.0, HighAvailability = nil => OpenShiftSDN
+			{
+				OpenShiftVersion:     swag.String("4.8.5"),
+				HighAvailabilityMode: nil,
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOpenShiftSDN,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeBaremetal)},
+			},
+			// OpenShift version >= 4.12.0-0.0, SNO, IPv4 => OVNKubernetes
+			{
+				OpenShiftVersion:     swag.String("4.12.0-ec.3"),
+				HighAvailabilityMode: swag.String(models.ClusterCreateParamsHighAvailabilityModeNone),
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOVNKubernetes,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeNone)},
+			},
+			// OpenShift version < 4.12.0-0.0, SNO => OVNKubernetes
+			{
+				OpenShiftVersion:     swag.String("4.11.0-rc.3"),
+				HighAvailabilityMode: swag.String(models.ClusterCreateParamsHighAvailabilityModeNone),
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOVNKubernetes,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeNone)},
+			},
+			// OpenShift version >= 4.12.0-0.0, SNO => OVNKubernetes
+			{
+				OpenShiftVersion:     swag.String("4.12-fc.2"),
+				HighAvailabilityMode: swag.String(models.ClusterCreateParamsHighAvailabilityModeNone),
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOVNKubernetes,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeNone)},
+			},
+			// OpenShift version >= 4.12.0-0.0, SNO => OVNKubernetes
+			{
+				OpenShiftVersion:     swag.String("4.12-rc.1"),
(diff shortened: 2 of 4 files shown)


eliorerz:
Please remove default `network_type` definition from [swagger.yaml](https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L4293-#L4297) (don't forget to run `skipper make generate` after you delete it)

internal/bminventory/inventory.go
```diff
@@ -340,9 +340,82 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+		// If an OpenShift version is not specified, we do not assume it is >= 4.12
+		if params.NewClusterParams.OpenshiftVersion == nil {
+			isOpenShiftVersionRecentEnough = false
+		} else {
+			openShiftVersion, err := version.NewVersion(swag.StringValue(params.NewClusterParams.OpenshiftVersion))
+			openShiftVersionParam = openShiftVersion
+			if err != nil {
+				return params.NewClusterParams.NetworkType, err
+			}
+			thresholdVersion, err := version.NewVersion("4.12")
```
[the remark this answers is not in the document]
danmanor: Currently version library compares as follow: x.y.z > x.y.z-rc > x.y.z-fc > x.y.z-ec, so it will not catch pre-release versions of 4.12.0. If we wish to do so I can change the threshold version to 4.12.0-ec or use a different function that ignores pre-release data

internal/bminventory/inventory.go
```diff
@@ -340,9 +340,82 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+	}
+}
+
+// returns whether the IP version is IPv6
+func isIPv6(cidr *string) (bool, error) {
+	// if no ClusterNetworkCidr specified, we do not assume nothing of it
+	if cidr == nil {
+		return false, nil
+	}
+	_, _, err := net.ParseCIDR(*cidr)
```
gamli75, line 408: this code is redundant - https://github.com/openshift/assisted-service/blob/a7e4f7485b3b7939e93e7c319865df3a6b18ee68/internal/network/family_utils.go#L16 is doing the same.
you just need to call to network.IsIPV6CIDR

internal/bminventory/inventory.go
```diff
@@ -340,9 +340,82 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+	if err != nil {
+		return false, errors.Wrapf(err, "%s is not a valid cidr", *cidr)
+	}
+	if network.IsIPV4CIDR(*cidr) {
```
gamli75, line 412: the method name is IsIPv6 but you are asking about IPv4(network.IsIPV4CIDR) - it is better to call network.IsIPV6CIDR but as I mentioned above you don't need it

internal/bminventory/inventory_test.go
```diff
@@ -9812,6 +9812,108 @@ var _ = Describe("TestRegisterCluster", func() {
...
+			It("NetworkType default value", func() {
+				type clusterParamsForTest struct {
+					OpenShiftVersion     *string
+					HighAvailabilityMode *string
+					ClusterNetworkCidr   *string
+					DesiredNetworkType   string
+					Platform             models.Platform
+				}
+
+				tests := []clusterParamsForTest{
```
gamli75, line 9828: please add another test for a case that CIDR is null or empty

internal/bminventory/inventory.go
```diff
@@ -340,9 +340,82 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+	}
+
 	return params, nil
 }
 
+// If the cluster is SNO, or the IP version is IPv6 (based on ClusterNetworkCidr param), or the OpenShift version >= 4.12,
+// the network type will be set to "OVNKubernetes", otherwise it will be set to "OpenShiftSDN"
+func getDefaultNetworkType(params installer.V2RegisterClusterParams) (*string, error) {
+	// If there is not a network specified already
+	if params.NewClusterParams.NetworkType == nil {
```
avishayt, line 356: Avoid nesting by checking "!= nil" and returning early

internal/bminventory/inventory.go
```diff
@@ -340,9 +340,82 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+
+		var (
+			isOpenShiftVersionRecentEnough bool
+			openShiftVersionParam          *version.Version
+		)
+
+		// If an OpenShift version is not specified, we do not assume it is >= 4.12
+		if params.NewClusterParams.OpenshiftVersion == nil {
```
avishayt, line 364: OpenshiftVersion should be required.  Anyway, you can also check this up front and return early to avoid more nesting.

internal/bminventory/inventory.go
```diff
@@ -340,9 +341,62 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+	// If one of the conditions match, we will return OVNKubernetes, otherwise OpenShiftSDN
+	if isOpenShiftVersionRecentEnough || isSingleNodeCluster || isIPv6 {
+		return swag.String(models.ClusterCreateParamsNetworkTypeOVNKubernetes), nil
+	} else {
+		return swag.String(models.ClusterCreateParamsNetworkTypeOpenShiftSDN), nil
+	}
+}
+
+func isSingleNodeCluster(params installer.V2RegisterClusterParams) bool {
+	if swag.StringValue(params.NewClusterParams.HighAvailabilityMode) == "None" {
```
osherdp, line 378: replace "none" with ``models.ClusterHighAvailabilityModeNone``

internal/bminventory/inventory.go
```diff
@@ -340,9 +341,62 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+		return true
+	} else {
+		return false
+	}
```
osherdp, line 382:
```suggestion
	return swag.StringValue(params.NewClusterParams.HighAvailabilityMode) == "None"
```

internal/bminventory/inventory.go
```diff
@@ -340,9 +341,62 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+
+	isOpenShiftVersionRecentEnough, err := verifyMinimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes(swag.StringValue(params.NewClusterParams.OpenshiftVersion))
+	if err != nil {
+		return params.NewClusterParams.NetworkType, err
+	}
+
+	isSingleNodeCluster := isSingleNodeCluster(params)
+	isIPv6 := network.IsIPv6CIDR(swag.StringValue(params.NewClusterParams.ClusterNetworkCidr))
+
+	// If one of the conditions match, we will return OVNKubernetes, otherwise OpenShiftSDN
```
osherdp, line 369: delete

internal/bminventory/inventory.go
osherdp: same goes to other ``return <bool-value>`` cases

osherdp (changes requested):
partial review

internal/bminventory/inventory.go
```diff
@@ -340,9 +341,51 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+		return swag.String(models.ClusterCreateParamsNetworkTypeOpenShiftSDN), nil
+	}
+}
+
+func verifyMinimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes(requestedOpenshiftVersion string) (bool, error) {
+	ocpVersion, err := version.NewVersion(requestedOpenshiftVersion)
+	if err != nil {
+		return false, errors.Errorf("Failed to parse OCP version %s", requestedOpenshiftVersion)
+	}
+	minimalVersionForSno, err := version.NewVersion(minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes)
```
gamli75, line 379: why did the member call minimalVersionForSno? it's not relevant to SNO
should be minimalVersionForNetworkTypeOVN or something similar

internal/bminventory/inventory.go
```diff
@@ -340,9 +341,51 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
 
+	params.NewClusterParams.NetworkType, err = getDefaultNetworkType(params)
+	if err != nil {
+		return params, err
+	}
+
 	return params, nil
 }
 
+// If the cluster is SNO, or the IP version is IPv6 (based on ClusterNetworkCidr param), or the OpenShift version >= 4.12,
```
gamli75, line 352: in the comment, you wrote that you are checking if "the IP version is IPv6" but I don't see you handle IPv6 in the method. 
I'm not sure what should be the fix: Add check for IPv6 or remove it from the comment

danmanor: Mistake, was corrected now

internal/bminventory/inventory.go
danmanor: Mistake, was corrected now

internal/bminventory/inventory_test.go
```diff
@@ -10029,6 +10029,160 @@ var _ = Describe("TestRegisterCluster", func() {
...
+					NewClusterParams.Platform = &t.Platform
+
+					reply := bm.V2RegisterCluster(ctx, installer.V2RegisterClusterParams{
+						NewClusterParams: NewClusterParams,
+					})
+
+					Expect(reply).Should(BeAssignableToTypeOf(installer.NewV2RegisterClusterCreated()))
+					actual := reply.(*installer.V2RegisterClusterCreated).Payload
+					Expect(swag.StringValue(actual.NetworkType)).To(Equal(t.DesiredNetworkType))
+				}
```
osherdp, line 10130: it seems to be the same code as above cases
why not merge ``availability-mode=nil`` to the above dictionary cases?

osherdp: same question for the pre-release cases

danmanor: I wanted to create a semantic separation between the tests, thought it would be too long in one test

internal/bminventory/inventory.go
```diff
@@ -340,9 +341,51 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+	isSingleNodeCluster := swag.StringValue(params.NewClusterParams.HighAvailabilityMode) == models.ClusterCreateParamsHighAvailabilityModeNone
+
+	if isOpenShiftVersionRecentEnough || isSingleNodeCluster {
+		return swag.String(models.ClusterCreateParamsNetworkTypeOVNKubernetes), nil
+	} else {
+		return swag.String(models.ClusterCreateParamsNetworkTypeOpenShiftSDN), nil
+	}
+}
+
+func verifyMinimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes(requestedOpenshiftVersion string) (bool, error) {
```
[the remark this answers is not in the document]
danmanor: True, but it doesn't make use of Core() ("Core returns a new version constructed from only the MAJOR.MINOR.PATCH segments of the version, without prerelease or metadata.") . we need Core for example in order to determine that 4.12.0-ec.3 >= 4.12. Should I add a new function in internal/common/version.go that uses Core() ?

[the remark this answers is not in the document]
avishayt: The unit tests on that method look convincing:
230:		is410Version, _ := VersionGreaterOrEqual("4.10.0", "4.10.0-0.alpha")
234:		is410Version, _ := VersionGreaterOrEqual("4.10.0-fc.1", "4.10.0-0.alpha")
238:		is410Version, _ := VersionGreaterOrEqual("4.10.1-fc.1", "4.10.0-0.alpha")
242:		is410Version, _ := VersionGreaterOrEqual("4.10.0-0.nightly-2022-01-23-013716", "4.10.0-0.alpha")

If it's broken, then we can fix it, but no need to replicate functionality.

swagger.yaml
```diff
@@ -4294,7 +4294,7 @@ definitions:
         type: string
         description: "The desired network type used."
         enum: ['OpenShiftSDN', 'OVNKubernetes']
-        default: 'OpenShiftSDN'
+        x-nullable: true
```
avishayt, line 4297: Does this break clients/API?

osherdp: I think it shouldn't break
At the schema-level, we don't disallow anything that wasn't possible before
At the data & values level, UI team is preparing to remove their extra-logic of setting ovn on "ocp>=4.12 || SNO", and provide no value for the network-type (letting us decide on cluster creation what should be the smart-default)

internal/bminventory/inventory.go
```diff
@@ -340,9 +341,36 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+// the network type will be set to "OVNKubernetes", otherwise it will be set to "OpenShiftSDN"
+func getDefaultNetworkType(params installer.V2RegisterClusterParams) (*string, error) {
+	// If the network type is specified already, we return it
+	if params.NewClusterParams.NetworkType != nil {
+		return params.NewClusterParams.NetworkType, nil
+	}
+
+	isOpenShiftVersionRecentEnough, err := common.VersionGreaterOrEqual(swag.StringValue(params.NewClusterParams.OpenshiftVersion), minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes)
+	if err != nil {
+		return params.NewClusterParams.NetworkType, err
```
mkowalski, line 362: I may be too picky, but I find it slightly confusing. We enter the error-path here and return effectively a tuple (nil,err) but instead of nil we write it as params.somethingsomething. The first argument is effectively nil because if it wasn’t, we would have caught it in the conditional two lines above. So if this is true, could we just write here “return nil, err”? To anyone consuming this function in the future it will be clear that this path is not a good one

internal/bminventory/inventory_test.go
```diff
@@ -10029,6 +10029,111 @@ var _ = Describe("TestRegisterCluster", func() {
 			}
 		}
 
+		// Testing the default network type setting during register flow, with various combinations of platforms,
+		// OpenShift version and high availability modes
+		// Note: clusters with high availability mode "None" must not have have Platform "baremetal", and OpenShift version >= 4.8
```
mkowalski, line 10034: Where is that sentence coming from? I agree that SNO is only OCP 4.8+, but nothing stops OCP 4.13 to support SNO with platform=baremetal

swagger.yaml
mkowalski: I ran through the code and I also think this is a safe change. If user provided the value, we will just carry it. If user does not provide a value, then immediately when we go to cluster creation we will fill that value with something real.

So in the current form it looks like there is no path to end up with empty network type in cluster creation.

How about cluster update though? I may not read it carefully enough but I have a feeling that it’s now possible to unset network type

internal/bminventory/inventory_test.go
avishayt: SNO must be platform "none"

swagger.yaml
danmanor: I don't think it is possible to unset the network type in cluster update, because: 
https://github.com/openshift/assisted-service/blob/2ee5cb8ca37b7cf6d96ec27f4d7980aaeea111e6/internal/bminventory/inventory.go#L2350-L2351

internal/bminventory/inventory_test.go
```diff
@@ -10018,6 +10018,110 @@ var _ = Describe("TestRegisterCluster", func() {
 		ctrl.Finish()
 	})
 
+	// Testing the default network type setting during register flow, with various combinations of platforms,
+	// OpenShift version and high availability modes
+	// Note: SNO currently must have have Platform "none", and OpenShift version >= 4.8
```
osherdp, line 10023:
```suggestion
	// Note: SNO is supported starting OpenShift v4.8, and currently only supports Platform "none"
```

osherdp (changes requested):

osherdp (approved):

PR #4425, opened 2022-09-21
MGMT-12067: Support discovery kernel arguments during infra-env registration and updates

ori-amizur:

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @filanov 
/cc @carbonin

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4261,23 +4261,34 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
 		return nil, err
 	}
 
+	var discoveryKernelArguments *string
+	if len(params.InfraenvCreateParams.DiscoveryKernelArguments) > 0 {
+		var b []byte
+		b, err = json.Marshal(&params.InfraenvCreateParams.DiscoveryKernelArguments)
+		if err != nil {
+			return nil, common.NewApiError(http.StatusBadRequest, errors.Wrap(err, "failed to format kernel arguments as json"))
+		}
+		discoveryKernelArguments = swag.String(string(b))
+	}
+
 	infraEnv := common.InfraEnv{
 		Generated: false,
 		InfraEnv: models.InfraEnv{
-			ID:                     &id,
-			Href:                   swag.String(url.String()),
-			Kind:                   swag.String(models.InfraEnvKindInfraEnv),
-			Name:                   params.InfraenvCreateParams.Name,
-			UserName:               ocm.UserNameFromContext(ctx),
-			OrgID:                  ocm.OrgIDFromContext(ctx),
-			EmailDomain:            ocm.EmailDomainFromContext(ctx),
-			OpenshiftVersion:       *osImage.OpenshiftVersion,
-			IgnitionConfigOverride: params.InfraenvCreateParams.IgnitionConfigOverride,
-			StaticNetworkConfig:    staticNetworkConfig,
-			Type:                   common.ImageTypePtr(params.InfraenvCreateParams.ImageType),
-			AdditionalNtpSources:   swag.StringValue(params.InfraenvCreateParams.AdditionalNtpSources),
-			SSHAuthorizedKey:       swag.StringValue(params.InfraenvCreateParams.SSHAuthorizedKey),
-			CPUArchitecture:        params.InfraenvCreateParams.CPUArchitecture,
+			ID:                       &id,
+			Href:                     swag.String(url.String()),
+			Kind:                     swag.String(models.InfraEnvKindInfraEnv),
+			Name:                     params.InfraenvCreateParams.Name,
+			UserName:                 ocm.UserNameFromContext(ctx),
+			OrgID:                    ocm.OrgIDFromContext(ctx),
+			EmailDomain:              ocm.EmailDomainFromContext(ctx),
+			OpenshiftVersion:         *osImage.OpenshiftVersion,
+			IgnitionConfigOverride:   params.InfraenvCreateParams.IgnitionConfigOverride,
+			StaticNetworkConfig:      staticNetworkConfig,
+			Type:                     common.ImageTypePtr(params.InfraenvCreateParams.ImageType),
+			AdditionalNtpSources:     swag.StringValue(params.InfraenvCreateParams.AdditionalNtpSources),
+			SSHAuthorizedKey:         swag.StringValue(params.InfraenvCreateParams.SSHAuthorizedKey),
+			CPUArchitecture:          params.InfraenvCreateParams.CPUArchitecture,
+			DiscoveryKernelArguments: discoveryKernelArguments,
 		},
 		KubeKeyNamespace: kubeKey.Namespace,
 		ImageTokenKey:    imageTokenKey,
@@ -4597,6 +4608,17 @@ func (b *bareMetalInventory) updateInfraEnvData(ctx context.Context, infraEnv *c
 			optionalParam(params.InfraEnvUpdateParams.Proxy.HTTPSProxy, "proxy_https_proxy", updates)
 			optionalParam(params.InfraEnvUpdateParams.Proxy.NoProxy, "proxy_no_proxy", updates)
 			updates["proxy_hash"] = proxyHash
+		}
+	}
+	if params.InfraEnvUpdateParams.DiscoveryKernelArguments != nil {
+		if len(params.InfraEnvUpdateParams.DiscoveryKernelArguments) > 0 {
+			b, err := json.Marshal(&params.InfraEnvUpdateParams.DiscoveryKernelArguments)
+			if err != nil {
+				return common.NewApiError(http.StatusBadRequest, errors.Wrap(err, "failed to format kernel arguments as json"))
+			}
+			updates["discovery_kernel_arguments"] = string(b)
+		} else {
+			updates["discovery_kernel_arguments"] = gorm.Expr("NULL")
 		}
 	}
 
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/kelseyhightower/envconfig"
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	gomega_format "github.com/onsi/gomega/format"
 	amgmtv1 "github.com/openshift-online/ocm-sdk-go/accountsmgmt/v1"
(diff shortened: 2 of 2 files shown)


internal/bminventory/inventory_test.go
```diff
@@ -6338,12 +6368,94 @@ var _ = Describe("infraEnvs", func() {
...
+						InfraEnvUpdateParams: &models.InfraEnvUpdateParams{
+							DiscoveryKernelArguments: kp1,
+						},
+					})
+					Expect(reply).To(BeAssignableToTypeOf(installer.NewUpdateInfraEnvCreated()))
+					i, err = bm.GetInfraEnvInternal(ctx, installer.GetInfraEnvParams{InfraEnvID: *i.ID})
+					Expect(err).ToNot(HaveOccurred())
+					Expect(swag.StringValue(i.DiscoveryKernelArguments)).To(Equal(jsonEncodeStringArray(kp1)))
+				})
+				By("Do not update DiscoveryKernelArguments -> DiscoveryKernelArguments with previous value", func() {
```
carbonin, line 6414: `It("does not change existing kernel arguments when none are sent in an update"`

internal/bminventory/inventory_test.go
```diff
@@ -6338,12 +6368,94 @@ var _ = Describe("infraEnvs", func() {
...
+						InfraEnvUpdateParams: &models.InfraEnvUpdateParams{
+							AdditionalNtpSources: swag.String("2.2.2.2"),
+						},
+					})
+					Expect(reply).To(BeAssignableToTypeOf(installer.NewUpdateInfraEnvCreated()))
+					i, err = bm.GetInfraEnvInternal(ctx, installer.GetInfraEnvParams{InfraEnvID: *i.ID})
+					Expect(err).ToNot(HaveOccurred())
+					Expect(swag.StringValue(i.DiscoveryKernelArguments)).To(Equal(jsonEncodeStringArray(kp1)))
+				})
+				By("Update DiscoveryKernelArguments with new values. Expected to be set in db", func() {
```
carbonin, line 6427: `It("overwrites existing kernel arguments with new ones"`

internal/bminventory/inventory_test.go
```diff
@@ -6338,12 +6368,94 @@ var _ = Describe("infraEnvs", func() {
...
+					mockInfraEnvUpdateSuccess()
+					kp2 := models.KernelArguments{
+						"p1",
+						"p2",
+						"p3",
+					}
```
carbonin, line 6433: I think it would be good to make these completely different from the previously set ones to explicitly show that we're not doing some kind of set union and only adding the new value (and keeping the old).

internal/bminventory/inventory_test.go
```diff
@@ -6338,12 +6368,94 @@ var _ = Describe("infraEnvs", func() {
...
+						InfraEnvUpdateParams: &models.InfraEnvUpdateParams{
+							DiscoveryKernelArguments: kp2,
+						},
+					})
+					Expect(reply).To(BeAssignableToTypeOf(installer.NewUpdateInfraEnvCreated()))
+					i, err = bm.GetInfraEnvInternal(ctx, installer.GetInfraEnvParams{InfraEnvID: *i.ID})
+					Expect(err).ToNot(HaveOccurred())
+					Expect(swag.StringValue(i.DiscoveryKernelArguments)).To(Equal(jsonEncodeStringArray(kp2)))
+				})
+				By("Update DiscoveryKernelArguments with empty array. Expected to be set in db", func() {
```
carbonin, line 6445: `It("clears the discovery kernel args when updated to an empty array"`

carbonin (commented):
Minor comments about tests and error messages. Everything else looks good.

internal/bminventory/inventory.go
```diff
@@ -4224,23 +4224,34 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
 		return nil, err
 	}
 
+	var discoveryKernelArguments *string
+	if len(params.InfraenvCreateParams.DiscoveryKernelArguments) > 0 {
+		var b []byte
+		b, err = json.Marshal(&params.InfraenvCreateParams.DiscoveryKernelArguments)
+		if err != nil {
+			return nil, common.NewApiError(http.StatusBadRequest, errors.Wrap(err, "bad discovery kernel parameters"))
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/bminventory/inventory.go
```diff
@@ -4562,6 +4573,17 @@ func (b *bareMetalInventory) updateInfraEnvData(ctx context.Context, infraEnv *c
 			updates["proxy_hash"] = proxyHash
 		}
 	}
+	if params.InfraEnvUpdateParams.DiscoveryKernelArguments != nil {
+		if len(params.InfraEnvUpdateParams.DiscoveryKernelArguments) > 0 {
+			b, err := json.Marshal(&params.InfraEnvUpdateParams.DiscoveryKernelArguments)
+			if err != nil {
+				return common.NewApiError(http.StatusBadRequest, errors.Wrap(err, "bad discovery kernel parameters"))
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/bminventory/inventory_test.go
```diff
@@ -6338,12 +6368,94 @@ var _ = Describe("infraEnvs", func() {
...
 				Expect(err).ToNot(HaveOccurred())
 				err = db.Model(&common.InfraEnv{}).Where("id = ?", i.ID).Update("generated_at", strfmt.DateTime(time.Now().AddDate(0, 0, -1))).Error
 				Expect(err).ToNot(HaveOccurred())
 			})
+			jsonEncodeStringArray := func(array []string) string {
+				b, e := json.Marshal(&array)
+				Expect(e).ToNot(HaveOccurred())
+				return string(b)
+			}
+			It("Update DiscoveryKernelArguments", func() {
```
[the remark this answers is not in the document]
ori-amizur: This tests update after a previous value already present - so this way is better for the test

internal/bminventory/inventory_test.go
```diff
@@ -6338,12 +6368,94 @@ var _ = Describe("infraEnvs", func() {
...
+				By("Do not update DiscoveryKernelArguments -> DiscoveryKernelArguments empty", func() {
```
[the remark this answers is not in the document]
ori-amizur: Not sure it is better.  Is it clearer ?

internal/bminventory/inventory_test.go
ori-amizur: OK

internal/bminventory/inventory_test.go
ori-amizur: Changed order added new param and removed existing one.

internal/bminventory/inventory_test.go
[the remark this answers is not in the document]
carbonin: I don't think it's better. If we have to change the behavior in a way where only a subset of these cases fail then we'll only see the first failure when we make changes.

Updating with a new value has nothing to do with changing the existing value. Don't tie tests together like this just to save setup code. Use a context and a before block to share setup.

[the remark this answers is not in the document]
ori-amizur: We need to test change after update was already applied.  This is the way to test it.

internal/bminventory/inventory_test.go
```diff
@@ -6546,12 +6576,100 @@ var _ = Describe("infraEnvs", func() {
...
+					},
+					{
+						name:                   "clears the discovery kernel args when updated to an empty array",
+						initialKernelArguments: ka2,
+						updateKernelArguments:  models.KernelArguments{},
+					},
+				}
+				for index := range tests {
+					t := tests[index]
+					It(t.name, func() {
```
carbonin, line 6633: I very much don't like this format of tests. It makes the test body very difficult to understand.

If you truly think a table test is the best option (which I don't given the amount of conditionals in the body) can you use the [ginkgo table test API](https://onsi.github.io/ginkgo/#table-specs)?

internal/bminventory/inventory_test.go
```diff
@@ -6546,12 +6576,100 @@ var _ = Describe("infraEnvs", func() {
...
+						Expect(i.DiscoveryKernelArguments).To(BeNil())
+						var reply middleware.Responder
+						if t.initialKernelArguments != nil {
+							mockInfraEnvUpdateSuccess()
+							reply = bm.UpdateInfraEnv(ctx, installer.UpdateInfraEnvParams{
```
carbonin, line 6638: I would prefer setting this using the database for the precondition.

internal/bminventory/inventory_test.go
```diff
@@ -6546,12 +6576,100 @@ var _ = Describe("infraEnvs", func() {
...
+						initialKernelArguments: ka2,
+						updateKernelArguments:  models.KernelArguments{},
+					},
+				}
+				for index := range tests {
+					t := tests[index]
+					It(t.name, func() {
+						Expect(i.DiscoveryKernelArguments).To(BeNil())
+						var reply middleware.Responder
+						if t.initialKernelArguments != nil {
```
carbonin, line 6636: This isn't a pointer, right? Wouldn't it be empty rather than nil if it wasn't explicitly set?

carbonin: Same for all the other conditional cases also.

ori-amizur: nil and empty slices are different in go

internal/bminventory/inventory_test.go
ori-amizur: I don't agree.  I prefer to set it with API.

internal/bminventory/inventory_test.go
carbonin: Ah I didn't realize it was a slice type. Thought it was a struct.

internal/bminventory/inventory_test.go
carbonin: Okay, pretty much every other infraenv test uses the DB to initialize the test state, but if you feel strongly I guess this one can do setup over the API :man_shrugging:

internal/bminventory/inventory_test.go
ori-amizur: Moved to https://onsi.github.io/ginkgo/#table-specs

carbonin (approved):

PR #4432, opened 2022-09-21
MGMT-11443: Add API to retrieve Network Manager Config Files

CrystalChun:
[MGMT-11443](https://issues.redhat.com/browse/MGMT-11443)
Adds an API to get the network manager config files at the endpoint `/v2/infra-envs/<infra-env-id>/downloads/files?file_name=static-network-config`

Returns a tar archive of the network config files for this infra-env.

WIP: Still need to add unit tests

[MGMT-11443](https://issues.redhat.com/browse/MGMT-11443)

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @carbonin

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -5342,6 +5342,21 @@ func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params
 			return common.GenerateErrorResponder(err)
 		}
 		filename = fmt.Sprintf("%s-%s", params.InfraEnvID, params.FileName)
+	case "static-network-config":
+		if infraEnv.StaticNetworkConfig != "" {
+			netFiles, err := b.staticNetworkConfig.GenerateStaticNetworkConfigData(ctx, infraEnv.StaticNetworkConfig)
+			if err != nil {
+				b.log.WithError(err).Errorf("Failed to create static network config data")
+				return common.GenerateErrorResponder(err)
+			}
+			buffer, err := staticnetworkconfig.GenerateStaticNetworkConfigArchive(netFiles)
+			if err != nil {
+				b.log.WithError(err).Errorf("Failed to create static network config archive")
+				return common.GenerateErrorResponder(err)
+			}
+			content = buffer.String()
+			filename = fmt.Sprintf("%s-%s.tar", params.InfraEnvID, params.FileName)
+		}
 	default:
 		return common.NewApiError(http.StatusBadRequest, fmt.Errorf("unknown file type for download: %s", params.FileName))
 	}
--- a/pkg/staticnetworkconfig/generator.go
+++ b/pkg/staticnetworkconfig/generator.go
@@ -1,6 +1,7 @@
 package staticnetworkconfig
 
 import (
+	"archive/tar"
 	"bytes"
 	"context"
 	"fmt"
@@ -294,4 +295,30 @@ func (s *StaticNetworkConfigGenerator) formatMacInterfaceMap(macInterfaceMap mod
 	}
 	sort.Strings(lines)
 	return strings.Join(lines, "\n")
+}
+
+func GenerateStaticNetworkConfigArchive(files []StaticNetworkConfigData) (*bytes.Buffer, error) {
+	buffer := new(bytes.Buffer)
+	w := tar.NewWriter(buffer)
+	for _, file := range files {
+		path := filepath.Join("/etc/assisted/network", file.FilePath)
+		content := file.FileContents
+
+		// add the file content
+		hdr := &tar.Header{
+			Name: path,
+			Mode: 0600,
+			Size: int64(len(content)),
+		}
+		if err := w.WriteHeader(hdr); err != nil {
+			return nil, err
+		}
+		if _, err := w.Write([]byte(content)); err != nil {
+			return nil, err
+		}
+	}
+	if err := w.Close(); err != nil {
+		return nil, err
+	}
+	return buffer, nil
 }
--- a/subsystem/infra_env_test.go
+++ b/subsystem/infra_env_test.go
@@ -163,6 +163,42 @@ var _ = Describe("Infra_Env", func() {
 		Expect(s.Size()).ShouldNot(Equal(0))
 	})
 
+	It("download infra-env files static network config file", func() {
+		By("Patching the infra env with a static network config")
+		netYaml := `interfaces:
+- ipv4:
+    address:
+    - ip: 192.0.2.1
+      prefix-length: 24
+    dhcp: false
+    enabled: true
+  name: eth0
+  state: up
+  type: ethernet`
+		staticNetworkConfig := models.HostStaticNetworkConfig{
+			NetworkYaml: netYaml,
+		}
+		staticNetworkConfigs := []*models.HostStaticNetworkConfig{&staticNetworkConfig}
+		updateParams := &installer.UpdateInfraEnvParams{
+			InfraEnvID: infraEnvID,
+			InfraEnvUpdateParams: &models.InfraEnvUpdateParams{
+				StaticNetworkConfig: staticNetworkConfigs,
+			},
+		}
+		_, err := userBMClient.Installer.UpdateInfraEnv(ctx, updateParams)
+		Expect(err).NotTo(HaveOccurred())
+		By("Downloading the static network config archive")
+		buf := &bytes.Buffer{}
+		_, err = userBMClient.Installer.V2DownloadInfraEnvFiles(ctx, &installer.V2DownloadInfraEnvFilesParams{InfraEnvID: infraEnvID, FileName: "static-network-config"}, buf)
+		Expect(err).NotTo(HaveOccurred())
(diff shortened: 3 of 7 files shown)


swagger.yaml
```diff
@@ -1635,7 +1635,7 @@ paths:
           name: file_name
           description: The file to be downloaded.
           type: string
-          enum: [discovery.ign, ipxe-script]
+          enum: [discovery.ign, ipxe-script, static-network-config]
```
carbonin, line 1638: This endpoint is becoming a bit crowded. At what point should we consider something not a "file" and give it a different endpoint under `/infra-envs/{infra_env_id}/downloads/`

Already these three cases are implemented completely differently. Curious what @avishayt thinks.

pkg/staticnetworkconfig/generator.go
```diff
@@ -295,3 +296,29 @@ func (s *StaticNetworkConfigGenerator) formatMacInterfaceMap(macInterfaceMap mod
 	sort.Strings(lines)
 	return strings.Join(lines, "\n")
 }
+
+func GenerateStaticNetworkConfigArchive(files []StaticNetworkConfigData, infraEnvID string) (*bytes.Buffer, error) {
+	buffer := new(bytes.Buffer)
+	w := tar.NewWriter(buffer)
+	for _, file := range files {
+		path := filepath.Join("/network", infraEnvID, file.FilePath)
```
carbonin, line 304: Why are we using this particular path?
I feel like it would be better to just keep the path the files would have on the host.

CrystalChun: Gotcha, I'll revert it back!

My initial thought process was to contain the files in a folder such as `network-files` so that ways the user can easily find them on their local filesystem when they untar the archive. 

Although, I can see how having it be the same path as it exists on the host(s) makes sense

swagger.yaml
CrystalChun: What about having one common endpoint such as `/v2/infra-envs/{infra_env_id}/downloads` with a query for the item you'd like to download? E.g. `/v2/infra-envs/{infra_env_id}/downloads?artifact=minimal-initrd` or `/v2/infra-envs/{infra_env_id}/downloads?artifact=ipxe-script`

carbonin: >What about having one common endpoint

This is effectively what we have now, right? Are there any other sub-paths under `/downloads`?

> Should we allow users to download static config files by host?

We apply the same static network config to all hosts since this is the discovery image.

CrystalChun: > > What about having one common endpoint
> 
> This is effectively what we have now, right? Are there any other sub-paths under `/downloads`?

`/downloads/minimal-initrd` is a different subpath. I meant combining that and the `/downloads/files` path 
> > Should we allow users to download static config files by host?
> 
> We apply the same static network config to all hosts since this is the discovery image.
Oh gotcha, then a follow up question: is it useful to generate and send all the network config files as opposed to just the one that's applied to all hosts?

carbonin: > I meant combining that and the /downloads/files path

This is what I'm considering a problem. We're downloading too many unrelated things each with their own separate options from this one endpoint. I'm thinking each should have a separate endpoint, but we can't do anything about the ones that are already here.

I'm not sure if starting with a new endpoint for this will make thinks better or worse :worried:

carbonin: >is it useful to generate and send all the network config files

What other ones? The scripts and such? I think the use for this endpoint is mostly just so admins can verify that the network manager config files look as they would expect.

CrystalChun: > This is what I'm considering a problem. We're downloading too many unrelated things each with their own separate options from this one endpoint. I'm thinking each should have a separate endpoint, but we can't do anything about the ones that are already here.
> 
> I'm not sure if starting with a new endpoint for this will make thinks better or worse 😟

Oh I see, is it usually better to have different endpoints for each type of download? So it would be split into
`/downloads/ipxe-script`
`/downloads/minimal-initrd`
`/downloads/discovery.ign`

Would this API definition benefit the end-user, be easier to maintain, etc?

carbonin: I think the problem right now is that some url params on this point only apply when you're downloading certain files and that's just not a great experience. @avishayt brought this up when we added the boot control stuff.

That said, I don't think we can change things like the ipxe script and discovery ignition without breaking the current API so I'm not sure how much we gain by splitting just this one out :man_shrugging:

CrystalChun: Would it be a good idea to start separating the APIs now instead of adding more to this `/downloads/files` endpoint?

---
What would be the process if we were to create new separate API endpoints for discovery ignition and ipxe script?

Would it be to introduce these APIs and keep (but start deprecating) the original downloads/files API for them?

avishayt: In the cluster download API it made sense to have a single endpoint.  We have 10 files there available for download so I wanted to avoid 10 download APIs.  The implementation shouldn't affect the API, but all those files are stored in S3 so the implementation for all 10 is identical - just pass the right file name.  We split out 'credentials' because that API has different authorization than 'files'.
Here I understand that the implementation is different but that shouldn't affect the API.  It would be better to keep things under 'files' to be consistent with cluster downloads.  In that case, we should move minimal-initrd to 'files' and deprecate the current API.  I'm also OK with making separate endpoints if you feel strongly about it, but then we should deprecate 'files'.

carbonin: :+1: lets just keep it the same as clusters then.

Easier to deprecate one endpoint and move it here than create a bunch of others.

I remember having a conversation with someone, I thought it was @avishayt, about how the parameters only apply if other parameters are set in a certain way, but I agree that's it's not worth the trouble.

carbonin (commented):
Looks good.

I think you could probably add a subsystem test.

subsystem/infra_env_test.go
```diff
@@ -163,6 +163,42 @@ var _ = Describe("Infra_Env", func() {
...
+		_, err := userBMClient.Installer.UpdateInfraEnv(ctx, updateParams)
+		Expect(err).NotTo(HaveOccurred())
+		By("Downloading the static network config archive")
+		buf := &bytes.Buffer{}
+		_, err = userBMClient.Installer.V2DownloadInfraEnvFiles(ctx, &installer.V2DownloadInfraEnvFilesParams{InfraEnvID: infraEnvID, FileName: "static-network-config"}, buf)
+		Expect(err).NotTo(HaveOccurred())
+		By("Verifying the contents of the archive")
+		contents := buf.String()
+		Expect(len(contents)).ShouldNot(Equal(0))
+		Expect(contents).To(ContainSubstring("/etc/assisted/network/host0"))
```
carbonin, line 197: Aren't these contents an archive?
Do we need to unpack this before doing these kinds of comparisons or will these strings always exist even in the archive directly?

carbonin: I suppose this also applies to the unit test

CrystalChun: It is an archive. 
When I print out the archive string (from the unit test) it looks like this and contains these strings:
```
/etc/assisted/network/host10000600000000000000000000000005000000000000014313 0ustar0000000000000000static network config data of first host
```
Is it better to always unpack it? 

I thought it only needed to be unpacked if it was compressed since the bytes need to be transformed?

internal/bminventory/inventory.go
```diff
@@ -5342,6 +5342,21 @@ func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params
...
+				b.log.WithError(err).Errorf("Failed to create static network config data")
+				return common.GenerateErrorResponder(err)
+			}
+			buffer, err := staticnetworkconfig.GenerateStaticNetworkConfigArchive(netFiles)
+			if err != nil {
+				b.log.WithError(err).Errorf("Failed to create static network config archive")
+				return common.GenerateErrorResponder(err)
+			}
+			content = buffer.String()
+			filename = fmt.Sprintf("%s-%s.tar", *infraEnv.Name, params.FileName)
```
avishayt, line 5358: The ipxe-script above uses the infra-env ID rather than the name in the filename.  Any reason to change that here?

CrystalChun: Thank you for catching! Sorry, originally changed it to the infra-env name for easy debugging. Changed it to the infra-env ID

subsystem/infra_env_test.go
carbonin: :man_shrugging: I suppose it works and the tar format isn't likely to change so I'm fine with it.

carbonin (approved):

subsystem/infra_env_test.go
CrystalChun: Thanks for bringing it up! I'll keep an eye out for unit test failures regarding it if it does change in the future

PR #4438, opened 2022-09-23
NO-ISSUE: Abstract how to get spoke kubeconfig from ClusterReference and add tests

javipolo:
Abstract the way that we're getting the spoke Admin KubeConfig name
This way it can be reused if needed and we can also run some tests against it

/cc @carbonin

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] Unit tests
- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller_test.go
+++ b/internal/controller/controllers/agent_controller_test.go
@@ -3135,3 +3135,105 @@ var _ = Describe("TestConditions", func() {
 		})
 	}
 })
+
+var _ = Describe("spokeKubeClient", func() {
+	var (
+		clusterName       = "test-cluster"
+		adminSecretName   = "admin-secret"
+		ctx               = context.Background()
+		c                 client.Client
+		cdSpec            hivev1.ClusterDeploymentSpec
+		hr                *AgentReconciler
+		mockCtrl          *gomock.Controller
+		mockClientFactory *spoke_k8s_client.MockSpokeK8sClientFactory
+		ref               *v1beta1.ClusterReference
+	)
+
+	BeforeEach(func() {
+		c = fakeclient.NewClientBuilder().WithScheme(scheme.Scheme).Build()
+		mockCtrl = gomock.NewController(GinkgoT())
+		mockClientFactory = spoke_k8s_client.NewMockSpokeK8sClientFactory(mockCtrl)
+		hr = &AgentReconciler{
+			Client:                c,
+			Log:                   common.GetTestLog(),
+			APIReader:             c,
+			SpokeK8sClientFactory: mockClientFactory,
+		}
+		cdSpec = hivev1.ClusterDeploymentSpec{
+			ClusterName:     clusterName,
+			ClusterMetadata: &hivev1.ClusterMetadata{},
+		}
+		ref = &v1beta1.ClusterReference{
+			Name:      clusterName,
+			Namespace: testNamespace,
+		}
+	})
+
+	AfterEach(func() {
+		mockCtrl.Finish()
+	})
+
+	It("uses the cluster deployment secret ref when the cluster deployment exists", func() {
+		cdSpec.ClusterMetadata = &hivev1.ClusterMetadata{
+			AdminKubeconfigSecretRef: corev1.LocalObjectReference{
+				Name: adminSecretName,
+			},
+		}
+		cd := newClusterDeployment(clusterName, testNamespace, cdSpec)
+		Expect(c.Create(ctx, cd)).To(Succeed())
+
+		secret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      adminSecretName,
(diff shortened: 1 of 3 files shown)


internal/controller/controllers/clusterdeployments_controller_test.go
```diff
@@ -3519,7 +3519,42 @@ var _ = Describe("selectClusterNetworkType", func() {
 	}
 })
 
-var _ = Describe("Getting ClusterDeployment admin kubeconfig secret name", func() {
+var _ = Describe("Getting spoke admin kubeconfig secret name from ClusterReference", func() {
```
carbonin, line 3522: Usually this should be the name of the function you're testing.

internal/controller/controllers/clusterdeployments_controller_test.go
```diff
@@ -3519,7 +3519,42 @@ var _ = Describe("selectClusterNetworkType", func() {
...
+		c = fakeclient.NewClientBuilder().WithScheme(scheme.Scheme).Build()
+		cd = newClusterDeployment(clusterName, testNamespace, defaultClusterSpec)
+		cd.Spec.ClusterMetadata = &hivev1.ClusterMetadata{
+			AdminKubeconfigSecretRef: corev1.LocalObjectReference{
+				Name: adminSecretName,
+			},
+		}
+	})
+	Context("ClusterDeployment does not exist", func() {
+		It("can make secret name from template", func() {
```
carbonin, line 3544: There's no need for a context when it contains a single test and doesn't have any setup code.

internal/controller/controllers/clusterdeployments_controller_test.go
```diff
@@ -3519,7 +3519,42 @@ var _ = Describe("selectClusterNetworkType", func() {
...
+			Expect(getClusterReferenceAdminKubeConfigSecretName(ctx, c, &clusterReference)).To(Equal(fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name)))
+		})
+	})
+	Context("ClusterDeployment exists", func() {
+		It("can extract secret name from ClusterDeployment", func() {
```
carbonin, line 3549: Same here, this should not have a context and be named something like:
`It("gets the secret name from the cluster deployment when it exists", func() {`

carbonin (commented):
Just some comments on the test names, but I'm curious what the blocker is to testing the function as it existed before.

We should be able to use the test client object for the queries and a mock client factory, right?
I don't see a great reason to break this function apart outside of testing so if we can do it without breaking it up that would probably be better.

javipolo:
> Just some comments on the test names, but I'm curious what the blocker is to testing the function as it existed before.
> 
> We should be able to use the test client object for the queries and a mock client factory, right? I don't see a great reason to break this function apart outside of testing so if we can do it without breaking it up that would probably be better.

My main reason to start the splitting was to not have logics on creating the default secretname in two different files of the code:

https://github.com/openshift/assisted-service/blob/268172d230603597fe28a6de4c329a072833b2bc/internal/controller/controllers/agent_controller.go#L284
https://github.com/openshift/assisted-service/blob/268172d230603597fe28a6de4c329a072833b2bc/internal/controller/controllers/clusterdeployments_controller.go#L2059

Once there, I thought that maybe it would be better to split the way I did, putting all logic at least in the same file, and it would also accomodate for nicer testing and possibly reusing that code if needed

Let me know what do you think, but if you feel that not splitting the functions is better I'll rewrite the tests to accomodate the situation :)

carbonin:
>My main reason to start the splitting was to not have logics on creating the default secretname in two different files of the code

If this is the reason, then I'd expect the new function to be called from multiple places, but it seems it's only called from one. It looks like you moved the duplicated code so it's co-located, but didn't deduplicate.

internal/controller/controllers/agent_controller_test.go
```diff
@@ -3135,3 +3136,112 @@ var _ = Describe("TestConditions", func() {
...
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("uses the kubeconfig template format when the cluster deployment is missing", func() {
+		cdSpec := hivev1.ClusterDeploymentSpec{
+			ClusterMetadata: &hivev1.ClusterMetadata{},
+		}
+
+		cd := newClusterDeployment(clusterName, testNamespace, cdSpec)
+		Expect(c.Create(ctx, cd)).To(Succeed())
```
carbonin, line 3225: Shouldn't this not be created here (just based on the test name)?

javipolo: You're totally right :sweat_smile: fixed!

carbonin (approved):

PR #4439, opened 2022-09-23
MGMT-9984: Support iPXE in converged ZTP flow

carbonin:
Set the PXE artifacts on the preprovisioningimage CR in the
preprovisioningimage controller will set in case the
preprovisioningimage accepts Initrd image format (and doesn't accept ISO)

https://issues.redhat.com/browse/MGMT-9984

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- Deploy a dev-scripts cluster with extra hosts.
- Create BMH for extra hosts using `redfish` protocol (not `redfish-virtualmedia`)
- Observe artifact urls in preprovisioning image

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


carbonin:
Still need to fully test this.

I was only able to see the pxe artifacts set by reversing the logic in the service because I don't know how to properly set the accepted formats. I have a thread open with the metal team for this, but if anyone knows here I'd welcome the info.

cc @eranco74

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -154,17 +155,29 @@ func (r *PreprovisioningImageReconciler) getIronicIgnitionConfig(log logrus.Fiel
...
+		image.Status.ImageUrl = infraEnv.Status.ISODownloadURL
+		image.Status.KernelUrl = ""
+		image.Status.ExtraKernelParams = ""
+	} else if funk.Contains(image.Spec.AcceptFormats, metal3_v1alpha1.ImageFormatInitRD) {
+		r.Log.Infof("Updating PreprovisioningImage ImageUrl with InitRD artifacts")
+		image.Status.Format = metal3_v1alpha1.ImageFormatInitRD
+		image.Status.ImageUrl = infraEnv.Status.BootArtifacts.InitrdURL
+		image.Status.KernelUrl = infraEnv.Status.BootArtifacts.KernelURL
+		// should this respect existing settings? do some kind of search and append?
+		image.Status.ExtraKernelParams = fmt.Sprintf("coreos.live.rootfs_url=%s", infraEnv.Status.BootArtifacts.RootfsURL)
```
carbonin, line 173: The question in the comment is still something I'm curious about, but maybe it's just as easy as leaving that for a follow up.

The other question is if we need to include the rest of the kernel parameters here?
[All these, for example.](https://github.com/openshift/assisted-service/blob/6d3452d546b8a3c67068015e4e94ce64278792b5/internal/bminventory/inventory_v2_handlers.go#L619)

vrutkovs: I don't think we should include those, so that user would have an option to override some of these

carbonin: I suppose that's part of the question I asked originally in the comment.
Is this a value we expect users to be editing? If so, this logic needs to be significantly more sophisticated. If not, then we should probably be setting all those values, right?

carbonin: @eranco74 I think we talked about this and said "no we don't expect users to be touching this", but how else will they set kernel params on discovery?

cc @ori-amizur as you're working on that epic

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -154,17 +166,28 @@ func (r *PreprovisioningImageReconciler) getIronicIgnitionConfig(log logrus.Fiel
...
 }
 
-func (r *PreprovisioningImageReconciler) setImage(generation int64, status *metal3_v1alpha1.PreprovisioningImageStatus, infraEnv aiv1beta1.InfraEnv) error {
+func (r *PreprovisioningImageReconciler) setImage(image *metal3_v1alpha1.PreprovisioningImage, infraEnv aiv1beta1.InfraEnv) error {
 	r.Log.Infof("Updating PreprovisioningImage ImageUrl to: %s", infraEnv.Status.ISODownloadURL)
-	status.ImageUrl = infraEnv.Status.ISODownloadURL
-	status.Architecture = infraEnv.Spec.CpuArchitecture
-	status.Format = metal3_v1alpha1.ImageFormatISO
+	image.Status.Architecture = infraEnv.Spec.CpuArchitecture
+	if funk.Contains(image.Spec.AcceptFormats, metal3_v1alpha1.ImageFormatISO) {
```
eranco74, line 172: Hmm, pretty sure image.Spec.AcceptFormats can contain both ISO and InitRD.
Since the ImageUrl is required for both formats (different artifacts) We should decide which one we set in case it accepts both.

eranco74 (commented):
Just need to figure out which format take precedence in case the image accepts both, other than than looks good

internal/controller/controllers/preprovisioningimage_controller.go
carbonin: I consciously did it this way to prefer ISO. Is there a significant difference between the two if both are accepted?

eranco74: Good call

internal/controller/controllers/preprovisioningimage_controller.go
eranco74: I guess there are few options for supporting this (annotations on the BMH, adding another attribute, getting it from infraEnv etc).
I think we can start with the current implementation and figure out how to support user provided kernel params later.

eranco74 (commented):
/lgtm

eranco74:
ai-operator-ztp failure seems unrelated:
```
[2022-10-26 21:00:30] Failed with exit code 1, retrying "oc annotate project openshift-local-storage openshift.io/node-selector= --overwrite=true"...
```
/retest-required

carbonin:
/retest-required

This _should_ work now that https://github.com/openshift/assisted-service/pull/4545 is merged

PR #4441, opened 2022-09-26
ARMOCP-417: enable arm64 for agent installer

jeffdyoung:
Translate arm64-> aarch64 in instances where we reference rhcos (should mirror x86_64 -> amd64 mapping)
In parallel with openshift-installer change for agent installer: https://github.com/openshift/installer/pull/6401

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment (dev-scripts needs additional work to support arm)
Manually created agent.iso and installed OCP with arm64 BM HW in our RDU2 lab
- [] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"runtime"
 	"runtime/debug"
 	"strings"
 	"time"
@@ -4291,6 +4292,8 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
 	}
 	url := installer.GetInfraEnvURL{InfraEnvID: id}
 
+	params.InfraenvCreateParams.CPUArchitecture = common.NormalizeCPUArchitecture(params.InfraenvCreateParams.CPUArchitecture)
+
 	log = log.WithField(ctxparams.ClusterId, id)
 	log.Infof("Register infraenv: %s with id %s", swag.StringValue(params.InfraenvCreateParams.Name), id)
 
@@ -6129,7 +6132,7 @@ func (b *bareMetalInventory) GetKnownApprovedHosts(clusterId strfmt.UUID) ([]*co
 // will automatically use the image matching the Hub's architecture.
 func isBaremetalBinaryFromAnotherReleaseImageRequired(cpuArchitecture, version string, platform *models.PlatformType) bool {
 	return cpuArchitecture != common.MultiCPUArchitecture &&
-		cpuArchitecture != common.DefaultCPUArchitecture &&
+		cpuArchitecture != common.NormalizeCPUArchitecture(runtime.GOARCH) &&
 		common.PlatformTypeValue(platform) == models.PlatformTypeBaremetal &&
 		featuresupport.IsFeatureSupported(version,
 			models.FeatureSupportLevelFeaturesItems0FeatureIDARM64ARCHITECTUREWITHCLUSTERMANAGEDNETWORKING)
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -305,6 +305,15 @@ func (h *handler) validateVersions() error {
 		if swag.StringValue(release.Version) == "" {
 			return errors.Errorf(fmt.Sprintf(missingValueTemplate, "version"))
 		}
+		// Normalize release.CPUArchitecture and release.CPUArchitectures
+		// TODO: remove this block when AI starts using aarch64 instead of arm64
+		if swag.StringValue(release.CPUArchitecture) == common.MultiCPUArchitecture || swag.StringValue(release.CPUArchitecture) == common.AARCH64CPUArchitecture {
+			*release.CPUArchitecture = common.NormalizeCPUArchitecture(*release.CPUArchitecture)
+			for i := 0; i < len(release.CPUArchitectures); i++ {
+				release.CPUArchitectures[i] = common.NormalizeCPUArchitecture(release.CPUArchitectures[i])
+			}
+
+		}
 	}
 
 	return nil
--- a/subsystem/cluster_v2_test.go
+++ b/subsystem/cluster_v2_test.go
@@ -474,7 +474,7 @@ var _ = Describe("[V2ClusterTests] multiarch", func() {
 			IgnoreStateInfo)
 	})
 
-	It("Fail to register infraenv with missing release image and OS ", func() {
+	It("Fail to register an infraenv with a non-supported CPUArchitecture ", func() {
 		_, err := userBMClient.Installer.RegisterInfraEnv(context.Background(), &installer.RegisterInfraEnvParams{
 			InfraenvCreateParams: &models.InfraEnvCreateParams{
 				Name:             swag.String("test-infra-env"),
@@ -488,7 +488,5 @@ var _ = Describe("[V2ClusterTests] multiarch", func() {
 		})
 
 		Expect(err).To(HaveOccurred())
-		actual := err.(*installer.RegisterInfraEnvBadRequest)
-		Expect(*actual.Payload.Reason).To(ContainSubstring("does not have a matching OpenShift release image"))
 	})
 })
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -4390,6 +4390,8 @@ definitions:
         type: string
         x-nullable: false
         default: 'x86_64'
+        # TODO: remove arm64 when AI moves to using aarch64
+        enum: ['x86_64', 'aarch64', 'arm64','ppc64le','s390x','multi']
         description: The CPU architecture of the image (x86_64/arm64/etc).
       disk_encryption:
         $ref: '#/definitions/disk-encryption'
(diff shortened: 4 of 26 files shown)


internal/bminventory/inventory.go
```diff
@@ -3955,7 +3957,9 @@ func validateArchitectureAndVersion(v versions.Handler, c *common.Cluster, cpuAr
 		ocpVersion = c.OpenshiftVersion
 	}
 	if c.CPUArchitecture != common.MultiCPUArchitecture {
-		if c.CPUArchitecture != "" && c.CPUArchitecture != cpuArch {
+		if c.CPUArchitecture == "aarch64" && cpuArch == "arm64" {
```
cs-zhang, line 3960: Should not fix the arch to x86 and arm, the original check should work for multi-arch support.

jeffdyoung: @cs-zhang I moved this logic out of the MultiCPUArchitecture block. It needs to be there otherwise I get a `release = aarch64 and cpu arch = arm64 does` not match error

internal/versions/versions.go
```diff
@@ -138,6 +138,9 @@ func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params o
 				arch = common.DefaultCPUArchitecture
 			}
 
+			if arch == "arm64" {
+				arch = "aarch64"
+			}
```
mkowalski, line 143: If we take this path, can we use both of those as `common.*`? E.g. `common.ARM64CPUArchitecture` and `common.AARCH64CPUArchitecture` or similar

internal/versions/versions.go
```diff
@@ -228,6 +231,10 @@ func (h *handler) GetOsImage(openshiftVersion, cpuArchitecture string) (*models.
 		// Empty implies default CPU architecture
 		cpuArchitecture = common.DefaultCPUArchitecture
 	}
+	// RHCOS uses aarch64 for arm64
+	if cpuArchitecture == "arm64" {
+		cpuArchitecture = "aarch64"
+	}
```
mkowalski, line 237: This function has quite an extensive set of unit tests in around https://github.com/openshift/assisted-service/blob/master/internal/versions/versions_test.go#L929-L934 and https://github.com/openshift/assisted-service/blob/master/internal/versions/versions_test.go#L403-L421

It would be very appreciated if you created a set of cases there that cover `GetOsImage` via `arm64` as well as via `aarch64` so that by reading the test I can understand the rationale behind the change.

Look e.g. at the test https://github.com/openshift/assisted-service/blob/master/internal/versions/versions_test.go#L579-L592 where we explicitly have a statement "gets X using X query" and "gets X using Y query"

internal/bminventory/inventory.go
```diff
@@ -1535,7 +1537,7 @@ func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, c
 
 	installerReleaseImageOverride := ""
 	if isBaremetalBinaryFromAnotherReleaseImageRequired(cluster.CPUArchitecture, cluster.OpenshiftVersion, cluster.Platform.Type) {
-		defaultArchImage, err := b.versionsHandler.GetReleaseImage(cluster.OpenshiftVersion, common.DefaultCPUArchitecture)
+		defaultArchImage, err := b.versionsHandler.GetReleaseImage(cluster.OpenshiftVersion, cluster.CPUArchitecture)
```
mkowalski, line 1540: Instead of modifying the content of this conditional, could we maybe modify the content of the function `isBaremetalBinaryFromAnotherReleaseImageRequired` so that it does not enter this code at all? AFAIK this block is meant to support cross-installation where Hub is x86 but spoke is ARM (but not the reverse). I.e. this block handles a scenario when `openshift-install` is executed on platform x86 but manifests that it generate may be used by ARM. By writing here `cluster.CPUArchitecture` you are saying "use openshift-install binary that matches the installed cluster architecture" which is different from "use openshift-install binary that matches the Hub cluster architecture"

mkowalski: TLDR: Have you tested that with this change you can install single-arch ARM cluster from X86 hub with no regression?

internal/controller/controllers/infraenv_controller.go
```diff
@@ -345,14 +345,15 @@ func (r *InfraEnvReconciler) ensureISO(ctx context.Context, log logrus.FieldLogg
 }
 
 func CreateInfraEnvParams(infraEnv *aiv1beta1.InfraEnv, imageType models.ImageType, pullSecret string, clusterID *strfmt.UUID, openshiftVersion string) installer.RegisterInfraEnvParams {
+	normalizedRHCOSARCH := infraEnv.Spec.CpuArchitecture
```
mkowalski, line 348: Maybe too quick and missed some other logic, but don't you want to write here `... := NormalizeRHCOSArch(infraEnv.Spec.CpuArchitecture)`?

jeffdyoung: I meant to remove this all together, but will add `NormalizeRHCOSArch(infraEnv.Spec.CpuArchitecture)` back based on our slack conversation.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -36,7 +36,7 @@ type OSImage struct {
 	// +optional
 	// Deprecated: this field is ignored (will be removed in a future release).
 	RootFSUrl string `json:"rootFSUrl"`
-	// The CPU architecture of the image (x86_64/arm64/etc).
+	// The CPU architecture of the image (x86_64/aarch64/etc).
```
jkilzi, line 39: Can we make this an enum instead of any string?

jeffdyoung:
Hey @mkowalski @jkilzi Do you guys have time to re-review? Its been a few months, so to refresh memories:

These changes are needed b/c the AI expects `arm64` for its cpu_architecture values. The agent installer will produce `aarch64` artifacts, because it generates os_images and release_image straight from the release image (like `x86_64`)  This PR's purpose is to allow agent to use `aarch64` without breaking the other AI components that require `arm64`.

After this merges, revisit migrating the entire assisted stack to use `aarch64`: https://issues.redhat.com/browse/ARMOCP-487

Arm Agent installer PR: https://github.com/openshift/installer/pull/6401

swagger.yaml
```diff
@@ -4390,6 +4390,8 @@ definitions:
         type: string
         x-nullable: false
         default: 'x86_64'
+        # TODO: remove arm64 when AI moves to using aarch64
+        enum: ['','x86_64', 'aarch64', 'arm64','ppc64le','s390x','multi']
```
carbonin, line 4394: What purpose does the empty value in this enum serve if there's also a default set?
(applies for all instances of this in the swagger file).

jeffdyoung: Ack, missed that... Thanks.

carbonin:
Do we think we need to normalize [these](https://github.com/openshift/assisted-service/blob/204d6fc0d4601690f41cb5c38ef3ced4fb00b17b/internal/controller/controllers/preprovisioningimage_controller.go#L131)?

This is comparing the architecture in the infraEnv CR with the one in the preprovisioningimage (detected by ironic in some way).

jeffdyoung:
> Do we think we need to normalize [these](https://github.com/openshift/assisted-service/blob/204d6fc0d4601690f41cb5c38ef3ced4fb00b17b/internal/controller/controllers/preprovisioningimage_controller.go#L131)?
> 
I haven't hit this in my local testing w/agent installer. FWIW, my first attempts was normalizing all comparisons, didn't work and was kinda messy. Normalizing the inputs seems to be the best route. 

> This is comparing the architecture in the infraEnv CR with the one in the preprovisioningimage (detected by ironic in some way).
> 
> Although I don't think this PR will change either of those as one is provided by the user and the other is provided by ironic so maybe it's fine (or at least as fine as it was before).

carbonin:
>I haven't hit this in my local testing w/agent installer.

You wouldn't, this is a ZTP-specific section of the code, but I think we can leave it for now.

Looks like you'll need to address the subsystem test failure though.

subsystem/cluster_v2_test.go
```diff
@@ -488,7 +489,7 @@ var _ = Describe("[V2ClusterTests] multiarch", func() {
 		})
 
 		Expect(err).To(HaveOccurred())
-		actual := err.(*installer.RegisterInfraEnvBadRequest)
-		Expect(*actual.Payload.Reason).To(ContainSubstring("does not have a matching OpenShift release image"))
+		actual := err.(*runtime.APIError)
```
carbonin, line 492: Ah I guess this is failing because you added the enum, right?

Can you also just change the name of the test?
Maybe something like `Fails to register an infraenv with a non-supported OS architecture` or something to distinguish it from the release/OS images provided in the config since that's actually not relevant to this test anymore.

swagger.yaml
```diff
@@ -455,6 +455,10 @@ paths:
           description: Error.
           schema:
             $ref: '#/definitions/error'
+        "422":
```
carbonin, line 458: Not sure I understand this change. When would we use this?

carbonin: Yeah, subsystem tests are kinda annoying to run locally, but you should be able to just get the system up and running in minikube with `make deploy-test` then just run the query the subsystem test is running manually.

carbonin: Also, however you want to handle testing is fine, I just wasn't sure if I should review this or if you were just testing, maybe mark the whole PR as WIP until you want reviews on the changes?

carbonin (approved):

carbonin:
FWIW this also resolves https://issues.redhat.com/browse/MGMT-13386

PR #4442, opened 2022-09-26
MGMT-8578: Describe process for day2 master for OCP 4.11

mkowalski:
Contributes-to: [MGMT-8578](https://issues.redhat.com//browse/MGMT-8578)
Closes: [MGMT-11945](https://issues.redhat.com//browse/MGMT-11945)

### DIFF
--- a/docs/user-guide/day2-master/411-healthy.md
+++ b/docs/user-guide/day2-master/411-healthy.md
@@ -1,0 +1,302 @@
+# Adding 4th master node on OpenShift 4.11 (and newer) with 3 healthy nodes
+
+## Add day-2 master
+
+Please follow the pre-existing guide describing the process of adding a day 2 node. Please note that you need to annotate the agent with `role: master`.
+
+A sample guide for adding a day-2 node is available [here](../rest-api-day2.md).
+
+## Accept CSRs
+
+During the installation there may be some CSRs to approve:
+
+```
+[root@test1001 ~]# oc get csr
+NAME        AGE     SIGNERNAME                                    REQUESTOR                                                                   REQUESTEDDURATION   CONDITION
+csr-sc2rk   73s     kubernetes.io/kubelet-serving                 system:node:extraworker-5                                                   <none>              Approved,Issued
+csr-tv9rs   7m59s   kubernetes.io/kube-apiserver-client-kubelet   system:serviceaccount:openshift-machine-config-operator:node-bootstrapper   <none>              Approved,Issued
+```
+
+## Confirm 4th master node is ready
+
+```
+[root@test1001 ~]# oc get nodes
+NAME            STATUS   ROLES    AGE     VERSION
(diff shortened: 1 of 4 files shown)


filanov:
consider adding it to a sub directory

docs/user-guide/day2-master/411-healthy.md
```diff
@@ -0,0 +1,298 @@
...
+NAME            STATUS   ROLES    AGE     VERSION
+extraworker-0   Ready    master   4h42m   v1.24.0+3882f8f
+extraworker-1   Ready    worker   4h29m   v1.24.0+3882f8f
+extraworker-2   Ready    master   4h43m   v1.24.0+3882f8f
+extraworker-3   Ready    master   4h27m   v1.24.0+3882f8f
+extraworker-4   Ready    worker   4h30m   v1.24.0+3882f8f
+extraworker-5   Ready    master   105s    v1.24.0+3882f8f
+```
+
+## Create Machine object linked with BMH and Node
```
avishayt, line 33: What if the cluster isn't baremetal platform?

docs/user-guide/day2-master/README.md
```diff
@@ -0,0 +1,20 @@
+# Installing control plane (master node) as day 2 operation
```
avishayt, line 1: The majority of the document focuses on the cluster itself, rather than on Assisted.  Is there any difference between Assisted SaaS and MCE/ACM in this regard?  Is it just how to download the ISO for the master?

mkowalski: This is right, that's because the whole process is really an OCP process and not Assisted's feature. On the side of Assisted Service the only difference is how you get an ISO for day2. I intentionally don't describe this here as we have enough documentation about day2 process itself.

> Is there any difference between Assisted SaaS and MCE/ACM in this regard

The difference is more about whether you do ZTP or BIY, not SaaS vs ACM

docs/user-guide/day2-master/411-healthy.md
mkowalski: I have raised this with etcd team (https://coreos.slack.com/archives/C027U68LP/p1664799799748179). But the current behaviour that I observe is

```
clustermembercontroller.go:203] Ignoring node (extraworker-5) for scale-up: no Machine found referencing this node's internal IP (192.168.111.28)
```

that implies that you need a Machine CR. Just note that it does not mean that you need `platform: baremetal` as you can create Machine CR even if the entry is "fake".

mkowalski: The outcome of the discussion is that Machine object is needed only if there is a functional Machine API in the cluster. I have adjusted the docs accordingly to explain that the step is not always needed and to link a doc that explains in details what is the functional MAPI

avishayt (approved):

PR #4444, opened 2022-09-28
MGMT-11090: Enhancement Doc: Assisted boot-reporter service

nmagnezi:

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/boot-status-reporter.md
+++ b/docs/enhancements/boot-status-reporter.md
@@ -1,0 +1,143 @@
+---
+title: boot-status-reporter
+authors:
+  - "@nmagnezi"
+creation-date: 2022-09-22
+last-updated: 2022-09-22
+---
+
+# Assisted Boot Reporter
+
+## Summary
+
+Create a new reporting service that runs as early as possible when hosts are booted from disk to collect additional 
+information and send it to `assisted-service`.
+
+## Motivation
+
+There are multiple occurrences of clusters that fail to install at the stage where
+hosts manage to pull ignition but fail to get to a point where the `assisted-controller`
+pod runs and calls home to `assisted-service` with logs.
+
+With that missing information, there's not a whole lot we can do to understand what
+happened with those installations.
+
+A new service that starts right after the boot process on all cluster hosts with no dependency on `assisted-controller`
+or `crio` will report logs we don't currently have and shed some light on why the host failed to progress and
+what exactly happened.
+
+
+### Goals
+
+1. Gather information for a host boot stage to improve our ability to debug failures and better inform 
+   users about the host progress stages (first boot, second boot, etc).
+
+2. The new reporting service should run independently on all cluster hosts, regardless of their role.
+
+3. The new service should be thoroughly cleaned up from the hosts and should never fail installations by itself.
+   If the boot reporter service fails to run or report logs, the cluster installation should continue, 
+   and the new host progress stages should be skipped in such a scenario.
+
+## Proposal
+
+
+### Changes To Assisted Service
+For cases as mentioned in `Motivation`, we need a new service (started via `systemd`) as soon as possible after hosts
+are booting from the disk and contact `assisted-service` to:
+
+1. Change the installation stage to a new `Booted with local ignition` between `Rebooting` and `Configuring`.
+   This stage refers to a boot with pointer ignition.
+
+2. Change the installation stage to a new `Booted with control plane ignition` Between
+   `Waiting for control plane` and  `Joined`.
+
+
+3. For each of the above-mentioned, Read the current host stage from `assisted-service`, to determine if:
+   - the service runs on the first boot :arrow_right: collect logs and report.
+   - or the second boot :arrow_right: collect logs, report, and self-cleanup after a configurable threshold.
+
+4. Add a new `LogsType` so the collected logs won't override the existing logs.
+   - current names: `host`, `controller`, `all`
+   - proposing to add a new name: `host_boot`
+
+### Information Needed And Changes To The Discovery Agent
+
+In order to communicate with `assisted-service`, The assisted boot reported service needs to know:
+1. The `assisted-service` URL.
+2. The pull-secret to use for agent authentication.
+3. The InfraEnv ID and host ID
+
+It will need this information starting from the first boot, which is a pointer ignition-based boot.
+
+For that, we need to modify the discovery agent to store this information on disk in a pre-defined location for the assisted boot reporter service to use.
+
+The assisted boot reporter service should start via `systemd` and determine its operating stage. As mentioned above:
+
+### First Boot (Pointer Ignition)
+Expected stage to be read from assisted-service backend: `Rebboting`.
+Update the host stage to: `Booted with local ignition`.
+Start a timer to compress and send logs every 10 minutes to `v2UpdateHostInstallProgress` API endpoint.
+
+
+### Second Boot (Control Plane Ignition)
+Expected stage to be read from assisted-service backend: `Waiting for control plane`.
+Update the host stage to: `Booted with control plane ignition`.
+Start a timer to compress send logs every 10 minutes to `v2UpdateHostInstallProgress` API endpoint.
+Start another timer to halt operation and self-cleanup as follows:
+
+1. If 1 hour has passed.
+2. If cluster install progress reached a point where it is no longer needed (now counting in `assisted-controller` pod). An indication could be that `kubelet` is up and running.
+   Cleanup should include:
+
+   * The stored files for pull-secret, host id, and infraEnv id.
+   * Service itself - should not run on additional host boot.
+
+### Collected Information
+
+Kept under a new logs type: `host_boot`:
+
+1. journalctl
+2. A list of mounted file systems
+3. crio logs
+4. network settings
+
+
+### Open Questions
+
+1. What should we use for implementation? Our options are:
+   - Bash:
+      - Advantages:
+          1. Has very little to no dependencies.
+          2. Expected to be present in future versions of RHCOS.
+      - Disadvantages:
+         1. Harder to maintain and test.
+         2. No code discovery agent code reuse (written in Go).
+         3. Client work will involve a lot of parsing.
+   - Go:
+      - Advantages:
+         1. Discovery agent code reuse (also written in Go).
+         2. Simpler to test and maintain, as we do for other assisted code bases. 
+         3. Can be shipped as a binary, which helps with dependencies.
+      - Disadvantages:
+         1. Not clear if we can include a Go binary in the ignition.
+         2. Multi-Platform support: we need to have a statically-linked binary not to introduce
+            more dependencies to the platform. Note that it needs to compile for any platform that
+            we want to support (x86, ARM, etc.)
+
+   - Python:
+      - Advantages:
+         1. Code reuse to some extent. Swagger is able to generate a Python-based client. 
+         2. Simpler to test and maintain, comparing to bash.
+      - Disadvantages:
+         1. Python is not included with RHCOS. We will need to install it and some third-party python libs it will probably need (e.g., python-requests).
+
+
+2. How to handle pending user action state?
+
+3. Are there any SNO specific things to take into account?
+
+4. Logs should be compressed before we send them. Any limitation with what we have installed on RHCOS?
+### UI Impact
+
+UI is expected to present the newly added host stages. Those will get reflected by the backend in the same way as the 
+current host stages.


docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+# Assisted Boot Reporter
+
+## Summary
+
+Create a new reporting service that runs as early as possible when hosts are booted from disk to collect additional 
+information and send it to `assisted-service`.
+
+## Motivation
+
+There are multiple occurrences of clusters that fail to install at the stage where
```
eranco74, line 18: So the case here is when the host did boot successfully from the right disk, configured network, and pulled the ignition.
In this case we have the installer running on the bootstrap node that should collect the relevant logs from the rest of the nodes.
In case the problem happened in the bootstrap node, the assisted-installer-controller should collect the relevant logs.

omertuc: >In case the problem happened in the bootstrap node, the assisted-installer-controller should collect the relevant logs.

The controller can't collect anything about the bootstrap node's post reboot early boot processr

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+hosts manage to pull ignition but fail to get to a point where the `assisted-controller`
```
omertuc, line 19: This line seems to imply that all hosts run the controller. That is not true. Only one arbitrary host runs the controller

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+
+2. The new service should be thoroughly cleaned up from the hosts and should never fail installations by itself.
+   If the boot reporter service fails to run or report logs, the cluster installation should continue, 
+   and the new host progress stages should be skipped in such a scenario.
+
+## Proposal
+
+
+### Changes To Assisted Service
+For cases as mentioned in `Motivation`, we need a new service (start via `systemd`) soon as possible after hosts
```
omertuc, line 40:
```suggestion
For cases as mentioned in `Motivation`, we need a new service (started via `systemd`) as soon as possible after hosts
```

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+are booting from the disk and contact `assisted-service` to:
+
+1. Change the installation stage to a new `Booted with local ignition` Between `Configuring`
```
omertuc, line 43: I don't like the new stage name too much because it kinda sounds like what Configuring already is. How about `Reporting`?

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+   and  `Waiting for control plane`. This stage refers to a boot with pointer ignition.
```
omertuc, line 44: Waiting for control plane comes before Configuring, but this sentence seems to imply otherwise 

Also waiting for control plane is not a universal stage, non bootstrap masters don't go through that stage so it's not a good example.

Maybe the sentence should read "between Configuring and Joined"

Also Configuring is not a guaranteed stage, hosts can go from rebooting to joined/done immediately when there's user managed networking and an API LoadBalancer

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+It will need this information starting from the first boot, which is a pointer ignition-based boot.
+
+For that, we need to modify the discovery agent to store this information on disk in a pre-defined location for the assisted boot reporter service to use.
+
+The assisted boot reporter service should start via `systemd` and determine its operating stage. As mentioned above:
+
+### First Boot (Pointer Ignition)
+Expected stage to be read from assisted-service backend: `Configuring`.
+Update the host stage to: `Booted with local ignition`.
+Start a timer to compress and send logs every 10 minutes to `v2UpdateHostInstallProgress` API endpoint.
```
eranco74, line 71: Unsure this is the endpoint for sending logs.

eranco74, line 71: I think this should use the `send_logs` functionality we have in the assisted-installer-agent

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+Update the host stage to: `Booted with control plane ignition`.
+Start a timer to compress send logs every 10 minutes to `v2UpdateHostInstallProgress` API endpoint.
+Start another timer to halt operation and self-cleanup as follows:
+
+1. If 1 hour has passed.
+2. If cluster install progress reached a point where it is no longer needed (now counting in `assisted-controller` pod). An indication could be that `kubelet` is up and running.
+   Cleanup should include:
+
+   * The stored files for pull-secret, host id, and infraEnv id.
+   * Service itself - should not run on additional host boot.
```
eranco74, line 85: @avishayt @tsorya @nmagnezi,
Can someone please explain why we need the boot-reporter service needs to run at this stage (for multi-node cluster installation)?
Perhaps the name (boot-reporter service) is misleading since it seems that it's expected to do more than just report that the host booted...
And it's unclear in which cases it provides the information we don't get without it.

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+### Collected Information
+1. journalctl
+2. A list of mounted file systems
+3. crio logs
+4. network settings
+
+
+### Open Questions
+
+1. What should we use for implementation? Our options are:
```
eranco74, line 96: I think we can use both bash and go (container): 
For the boot reporting - use bash, a simple curl command should do the trick, and we don't really on anything other than network access to the service.
For sending logs - I'd prefer if we reuse the send logs functionality already implemented in the assisted-installer-agent. but since this depends on other things (container run time, access to the registry) I guess we can try to run the agent and in case this method fails, fall back to a bash script as well.

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+
+   - Python:
+      - Advantages:
+         1. Code reuse to some extent. Swagger is able to generate a Python-based client. 
+         2. Simpler to test and maintain, comparing to bash.
+      - Disadvantages:
+         1. Python is not included with RHCOS. We will need to install it and some third-party python libs it will probably need (e.g., python-requests).
+
+
+2. How to handle pending user action state?
```
eranco74, line 121: `pending user action` might be the node current stage instead of `Rebooting` (the node moves to `pending user action` in case the node failed to boot from the installation disk.
So I think we should handle it the same way as we handle `Rebooting` if the `boot-reporter` service started it means that the node booted with the right disk and we can move to `Booted with control plane ignition`

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+
+1. Change the installation stage to a new `Booted with local ignition` Between `Configuring`
+   and  `Waiting for control plane`. This stage refers to a boot with pointer ignition.
+
+
+2. Change the installation stage to a new `Booted with control plane ignition` Between
+   `Waiting for control plane` and  `Joined`.
+
+
+3. For each of the above-mentioned, Read the current host stage from `assisted-service`, to determine if:
```
eranco74, line 51: Unsure what is the added value of this service after the first boot.

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+Create a new reporting service that runs as early as possible when hosts are booted from disk to collect additional 
+information and send it to `assisted-service`.
+
+## Motivation
+
+There are multiple occurrences of clusters that fail to install at the stage where
+hosts manage to pull ignition but fail to get to a point where the `assisted-controller`
+pod runs and calls home to `assisted-service` with logs.
+
+With that missing information, there's not a whole lot we can do to understand what
```
eranco74, line 22: I think this enhancement should discuss the boot-reporter functionality in multi nodes cluster installation and SNO installation. 
The node stages updates and the logs gathering we have in place are different (lacking in case of SNO).

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+      - Advantages:
+         1. Code reuse to some extent. Swagger is able to generate a Python-based client. 
+         2. Simpler to test and maintain, comparing to bash.
+      - Disadvantages:
+         1. Python is not included with RHCOS. We will need to install it and some third-party python libs it will probably need (e.g., python-requests).
+
+
+2. How to handle pending user action state?
+
+3. Are there any SNO specific things to take into account?
```
eranco74, line 123: Yes, in the case of SNO ( there is no configuring stage, the node installation is more cumbersome and there is no installer/controller to report the progress or send logs)  it makes sense for this service to keep reporting and send logs until the assisted-installer-controller is running.
In multi-node cluster installation it does not and this service might override the logs gathered by the current logs collection implementation.

docs/enhancements/boot-status-reporter.md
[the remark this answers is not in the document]
eranco74: I think one motivation for this service (in case we intend to use it for more than just reporting successful boot) is to align the SNO stages and logs gathering functionality with the multi-node stages.

docs/enhancements/boot-status-reporter.md
tsorya: This case is mostly relevant for sno, in case controller is not started due to some issues like kube-api/kubelet and etc
we will get logs

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+### Information Needed And Changes To The Discovery Agent
+
+In order to communicate with `assisted-service`, The assisted boot reported service needs to know:
+1. The `assisted-service` IP address and port.
+2. The pull-secret to use for agent authentication.
+3. The InfraEnv ID and host ID
+
+It will need this information starting from the first boot, which is a pointer ignition-based boot.
+
+For that, we need to modify the discovery agent to store this information on disk in a pre-defined location for the assisted boot reporter service to use.
```
omertuc, line 64: Why not embed the files in the installed ignition?

docs/enhancements/boot-status-reporter.md
omertuc: For the bootstrap host post reboot issue diagnosting, must gather / controller don't give you that info

omertuc: We should run masters-gather and upload the result

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+---
+
+# Assisted Boot Reporter
+
+## Summary
+
+Create a new reporting service that runs as early as possible when hosts are booted from disk to collect additional 
+information and send it to `assisted-service`.
+
+## Motivation
```
tsorya, line 16: I think our motivation should be 
"No matter what, host should send logs after reboot" from my point status is even less interesting

docs/enhancements/boot-status-reporter.md
tsorya, line 64: Not sure i get it right. Why this is needed in discovery agent? 
I think assisted service should patch pointer ignition with all relevant values, no need for anything in agent

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+Start another timer to halt operation and self-cleanup as follows:
+
+1. If 1 hour has passed.
+2. If cluster install progress reached a point where it is no longer needed (now counting in `assisted-controller` pod). An indication could be that `kubelet` is up and running.
+   Cleanup should include:
+
+   * The stored files for pull-secret, host id, and infraEnv id.
+   * Service itself - should not run on additional host boot.
+
+### Collected Information
```
tsorya, line 87: Please point that this data will require adding new log type in order not to override host logs before reboot

docs/enhancements/boot-status-reporter.md
tsorya: I don't think we need to rely on pulling image. Idea is to send logs no matter what if we will need to pull image it will break this idea. 
I assume we need to use bash , it should be easy code to write and maintain

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+         3. Client work will involve a lot of parsing.
+   - Go:
+      - Advantages:
+         1. Discovery agent code reuse (also written in Go).
+         2. Simpler to test and maintain, as we do for other assisted code bases. 
+         3. Can be shipped as a binary, which helps with dependencies.
+      - Disadvantages:
+         1. Not clear if we can include a Go binary in the ignition.
+
+   - Python:
```
tsorya, line 113: There is no python, how it will run?

docs/enhancements/boot-status-reporter.md
eranco74: I agree that this service is mostly relevant for SNO and I think the enhancement motivation, background and goals should reflect that.

>  The controller can't collect anything about the bootstrap node's post reboot early boot process 
This is something we can solve with the current implementation (copy the private key to the controller).

On the other hand, if the motivation is to have a robust mechanism to get the information about the host boot status and the logs from the host regardless of the scenario (SNO, worker stuck in configuring, host changed IP) than this service makes sense.

docs/enhancements/boot-status-reporter.md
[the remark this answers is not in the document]
eranco74: What scenario is this?
Why would it apply the new boot-reporter service but not the rest of the ignition payload?

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+3. The InfraEnv ID and host ID
+
+It will need this information starting from the first boot, which is a pointer ignition-based boot.
+
+For that, we need to modify the discovery agent to store this information on disk in a pre-defined location for the assisted boot reporter service to use.
+
+The assisted boot reporter service should start via `systemd` and determine its operating stage. As mentioned above:
+
+### First Boot (Pointer Ignition)
+Expected stage to be read from assisted-service backend: `Configuring`.
```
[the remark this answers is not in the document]
eranco74: Hmm, I think you are right,
If the host did pull the ignition from the MCS it's already configuring (so what's the point in adding the new stage??). 
If the host didn't pull the ignition from the MCS (failed to reach the API or some other issue with the bootstrap control plane) it will fail during ignition and I doubt if the new boot reporter service will start in that case.

docs/enhancements/boot-status-reporter.md
eranco74: I tend to agree, the main value of this service is the ability to get the logs from the node with minimum dependencies.
in that regard bash script is the way to go.

docs/enhancements/boot-status-reporter.md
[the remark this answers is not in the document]
omertuc: >If the host did pull the ignition from the MCS it's already configuring (so what's the point in adding the new stage??).

It's still nice to see a new stage because you know not only it tried to pull ignition but it also began applying it and our service was able to run and our service was able to communicate with the assisted-service, tells you the networking is working great

 >If the host didn't pull the ignition from the MCS (failed to reach the API or some other issue with the bootstrap control plane) it will fail during ignition and I doubt if the new boot reporter service will start in that case.
 
 This is indeed a problem. I wish we could start the new service even when the ignition "remote merge" fails

docs/enhancements/boot-status-reporter.md
eranco74: I think that was the intention here, probably requires rephrasing.

docs/enhancements/boot-status-reporter.md
eranco74: Agree, and we should make sure that these logs don't override/conflict with the current logs gathering we have in place

docs/enhancements/boot-status-reporter.md
omertuc: Why bash? it's not compiled (compilation tells you about mistakes long before you run the program), it's not typed (much easier to write bugs), it's a language riddled with horrible pitfalls and I have no doubt we will be chasing bugs if we use it. 

You can compile a static go binary and place it in the ignition just as easily, then you can use the proper assisted go api library and make structured API calls without relying on curl CLI mess with a real language

docs/enhancements/boot-status-reporter.md
romfreiman: when will it be even relevantl:
1. wrong boot order - wont help
2. no networking post reboot (static networking was not applied correctly) - low chances that despite that, the host will still have networking to reach the saas / hub cluster
3. wrong dns configuration - lets add better validations before triggering the installation
4. what else?

omertuc: Literally any error that prevents crio/kubelet from functioning properly (e.g. expired certs? the crio ipv4/ipv6 bug we had recently?)

omertuc: i.e. crio and kubelet and super complex and prone to failure, we want to see why and how they fail even on SNO/bootstrap nodes

omertuc: Also we can't really tell why some installations fail when it's the bootstrap node post-reboot or a SNO.

This service is a best-effort attempt at maybe finding an answer (or maybe not, depending on the failure)

docs/enhancements/boot-status-reporter.md
omertuc: >The problem with go library is to provide it through ignition.

Why?

>If we decide to go with go library we will need to bring it through pulling image

no images, just a static binary embedded directly inside the ignition

omertuc: For example a statically built controller executable with upx + removing debugging information (`-ldflags="-s -w"`) from the executable it only weighs 20M base64 encoded, and a controller does much more things than what we want to build, and I bet we can make it even smaller if we try

docs/enhancements/boot-status-reporter.md
nmagnezi: Still testing this. However, I had chat with @carbonin about this, and he thinks we should avoid it.
Instead, he proposed to have a script file downloaded via assisted-service API (probably a bash script).
I'll complete the test anyhow but mentioned this in case Nick would want to expand on this.

carbonin: Yes, but my suggestion was just to avoid an incredibly large payload of base64 data in ignition.
Ignition itself also allows files to be created using regular http urls so I'm suggesting to download whatever script/executable thing we use with that format rather than embedding directly into the pointer ignition.

I also suggested bash because it's the most portable.
I don't think we want to get into the business of compiling go programs for every architecture we might soon support.

I think there's a lot of room for discussion on how to go about distributing whatever this service is, and a lot of it depends on how complicated it ends up being.

docs/enhancements/boot-status-reporter.md
mkowalski: I discussed with Nir quite a bit about the motivation and I would like to write it explicitly here - any solution that involves running assisted-installer-controller is not a solution here.

Reasoning - in order to run assisted-installer-controller we need to have a working k8s cluster. In order to have a working k8s cluster we need to have a working kubelet. In order to have a working kubelet we need to have a working crio. Therefore, before booting a machine and getting to run assisted-installer-controller there is a bunch of components out of which none reports its status mid-way.

To have a robust solution we need to have something that is plugged at the beginning of this dependency chain. Thus, we need something with no dependency on crio. Thus, our solution cannot be running a container. This takes assisted-installer-controler out of the game here.

We keep discussing improving assisted-installer-controller for ages now (and it's good, it needs improvements) but none of those discussions focuses on the problem we try to solve here - installation can fail because a component of the RHCOS is dodged. We need ability to debug this, otherwise we will keep receive tickets that are unsolvable without overseeing the installation in real-time.

We don't need an ideal solution. We need a solution good enough that will tell us e.g. "machine booted with network connectivity but XYZ happened so kubelet never started". Omer summarizes it perfectly - we need best-effort, not an ideal.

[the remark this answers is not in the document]
mkowalski: Service should run independently on all the hosts. In the initial form we always discussed (i.e. architecture call) it was a standalone entity and not a clustered service. Clustered service introduces yet another path to fail, so we should rather avoid it

docs/enhancements/boot-status-reporter.md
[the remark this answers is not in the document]
mkowalski: Agree with Igal here. We had cases where we managed to pull the ignition but RHCOS failed to apply the ignition. E.g. using `storage` stanza that refers to a disk via non-existing name. That's syntax-wise a totally correct ignition override but RHCOS will refuse to apply it and will not continue the boot process.

Whether this service here would help is another question as we would need to design its dependency chain quite carefully.

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+
+There are multiple occurrences of clusters that fail to install at the stage where
+hosts manage to pull ignition but fail to get to a point where the `assisted-controller`
+pod runs and calls home to `assisted-service` with logs.
+
+With that missing information, there's not a whole lot we can do to understand what
+happened with those installations.
+
+
+### Goals
```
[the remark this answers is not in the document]
mkowalski: Maybe we should write explicitly in the "Non-Goals" section that it is not a self-healer service but only an observer

docs/enhancements/boot-status-reporter.md
mkowalski: Could we piggy-back here and maybe rethink renaming `Configuring`? I'm not sure we have a common agreement that everyone understands that Configuring means "ignition pulled from MCS but not necessarily applied correctly". Maybe something like following states: `Configuring (ignition pulled)` followed by `Configuring (ignition applied)`?

docs/enhancements/boot-status-reporter.md
mkowalski: I think it was about the self-cleanup, am I right? But maybe the service should be a one-shot, so that the very first instance of the service is responsible for collecting the logs, reporting them back to SaaS and immediately killing itself afterwards? That way there will never be sign of our service in any subsequent reboot.

docs/enhancements/boot-status-reporter.md
mkowalski: I have to disagree here with anything that says "container". We need to ship as a standalone binary at most because we want to catch cases where Container Runtime on the host is broken. No matter whether we would need to pull a new image or just run something that we shipped, if someone thinks about running this service via e.g. `podman run` my very first question will be - what are you going to do if podman binary doesn't execute as you expected?

We had cases where e.g. c/storage was broken. You weren't able to run any container. Not only to pull a container, but not even run what you already have.

If we want to be robust, we should use the simplest stack possible. I agree with Omer on superiority of statically-built go over bash. If we combine it with Nick's idea of just pointing towards the payload instead of embedding it all in the ignition, we will not even increase the ignition's size.

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+         3. Can be shipped as a binary, which helps with dependencies.
+      - Disadvantages:
+         1. Not clear if we can include a Go binary in the ignition.
+
+   - Python:
+      - Advantages:
+         1. Code reuse to some extent. Swagger is able to generate a Python-based client. 
+         2. Simpler to test and maintain, comparing to bash.
+      - Disadvantages:
+         1. Python is not included with RHCOS. We will need to install it and some third-party python libs it will probably need (e.g., python-requests).
```
mkowalski, line 118: This requirement may be a deal-breaker. I'm not sure if pulling `python-requests` as a core dependency in OCP will pass. Whatever we would need to install here will become a standard package included everywhere, unless we have a mechanism that cleans it up later (but then we need to write more of a framework rather than a simple log collector)

docs/enhancements/boot-status-reporter.md
mkowalski, line 96: CoreOS moves to Rust in some of its components. Should we at least add a point about it? Not saying that it should be our choice, but to at least have it under consideration

docs/enhancements/boot-status-reporter.md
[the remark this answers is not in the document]
filanov: maybe need to make it clear in the proposal part

docs/enhancements/boot-status-reporter.md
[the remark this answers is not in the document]
filanov: yes at least for now :)

docs/enhancements/boot-status-reporter.md
nmagnezi: @tsorya, I think this is pretty much what the `Summary` section states. Let me know if you think a change of wording is needed here.

nmagnezi: @eranco74 By "the current logs gathering" do you mean the logs from the discovery phase? 
Based on [send_logs.go](https://github.com/openshift/assisted-installer-agent/blob/584421642dfb3cbd32ef9d81fcf3be91debe164c/src/logs_sender/send_logs.go#L113), it uses the `V2UploadLogs`. Should we use something else for this?

docs/enhancements/boot-status-reporter.md
nmagnezi: Thank you all for the feedback. 
I think a "best effort" is a fair description here.
I agree with the reasoning that we should not depend on `crio` or other things that are prone to fail the installation, leaving us as blind as we are today.

Therefore, this service needs to be basic and with little to no dependencies. Start via systemd as soon as boot happens and immediately try to report logs (and status).

I *think* this should cover both SNO and non-SNO cases; however, I am aware of some SNO-specific concerns @eranco74 has, and we will probably keep discussing that.

@romfreiman  as for:
>     2. no networking post reboot (static networking was not applied correctly) - low chances that despite that, the host will still have networking to reach the saas / hub cluster

as we have seen [today](https://coreos.slack.com/archives/CUPJTHQ5P/p1664781693190949?thread_ts=1664774218.591499&cid=CUPJTHQ5P) numbers do show that some clusters are using dhcp alongside nmstate, so we might also get logs in those situations.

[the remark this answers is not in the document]
nmagnezi: fair point. I will add that.

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Changes To Assisted Service
+For cases as mentioned in `Motivation`, we need a new service (start via `systemd`) soon as possible after hosts
+are booting from the disk and contact `assisted-service` to:
+
+1. Change the installation stage to a new `Booted with local ignition` Between `Configuring`
+   and  `Waiting for control plane`. This stage refers to a boot with pointer ignition.
+
+
+2. Change the installation stage to a new `Booted with control plane ignition` Between
```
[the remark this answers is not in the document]
nmagnezi: I was aiming for injecting this new stage right after the boot from MCS ignition was completed. From your comment, I understand it should be changed to:

> 2. Change the installation stage to a new `Booted with control plane ignition` Between `Configuring` and `Joined`.
 

What do you think?

docs/enhancements/boot-status-reporter.md
nmagnezi: I had a discussion with @avishayt and @tsorya about this, in which we determined that we want to reflect each boot in its own stage.
This will:
1. Allow more visibility about the host's progress.
2. Allow to service to determine the first or second boot (and by that, activate the cleanup) by reading the host stage from `assisted-service`.

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+
+
+3. For each of the above-mentioned, Read the current host stage from `assisted-service`, to determine if:
+   - the service runs on the first boot :arrow_right: collect logs and report.
+   - or the second boot :arrow_right: collect logs, report, and self-cleanup after a configurable threshold.
+
+### Information Needed And Changes To The Discovery Agent
+
+In order to communicate with `assisted-service`, The assisted boot reported service needs to know:
+1. The `assisted-service` IP address and port.
```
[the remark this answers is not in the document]
nmagnezi: Done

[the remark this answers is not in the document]
nmagnezi: Done

docs/enhancements/boot-status-reporter.md
nmagnezi: Is it okay to place the pull secret in the pointer ignition?

docs/enhancements/boot-status-reporter.md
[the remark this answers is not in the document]
nmagnezi: I'm changing it from `Configuring` to `Rebooting` so it will be on-par with the section above.
Also, indeed it does not catch all cases but as Omer said it does let you know that networking is up and running.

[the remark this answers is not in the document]
nmagnezi: Done.

docs/enhancements/boot-status-reporter.md
nmagnezi: Should `V2UploadLogs` be used instead?

nmagnezi: Yes, if we end up implementing this in Go. Some opinions lean towards bash.

docs/enhancements/boot-status-reporter.md
nmagnezi: Ack.
Added to `Changes To Assisted Service`:

> 4. Add a new `LogsType` so the collected logs won't override the existing logs.
>       - current names: host, controller, all
>       - proposing to add a new name: host_boot

let me know what you think

docs/enhancements/boot-status-reporter.md
nmagnezi: +1 @mkowalski.
To download that script, won't that mean we will need to have an endpoint with no auth? (or, "empty auth")?
Won't that be a problem with the SaaS deployment?

Unless.. we can place the pull secret inside the pointer ignition, which allows us to use agent auth.

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,129 @@
...
+         1. Harder to maintain and test.
+         2. No code discovery agent code reuse (written in Go).
+         3. Client work will involve a lot of parsing.
+   - Go:
+      - Advantages:
+         1. Discovery agent code reuse (also written in Go).
+         2. Simpler to test and maintain, as we do for other assisted code bases. 
+         3. Can be shipped as a binary, which helps with dependencies.
+      - Disadvantages:
+         1. Not clear if we can include a Go binary in the ignition.
```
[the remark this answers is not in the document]
nmagnezi: Ack. added this:
> 2. Multi-Platform support: we need to have a statically-linked binary not to introduce more dependencies to the platform. Note that it needs to compile for any platform that we want to support (x86, ARM, etc.)

docs/enhancements/boot-status-reporter.md
nmagnezi: Is there a need to write anything to disk? we aim to upload logs that can be pulled in real-time (journal etc)

docs/enhancements/boot-status-reporter.md
carbonin: I think we're going to likely have a problem with either release engineering or our build chain if we try to use golang here.

I see a few options and issues with all of them:
1. We compile these binaries from go source and commit them to our code base:
  - Generally poor form. Source repo should have source, not compiled code.
  - I think this will run afoul of some security guidelines that I can't find a reference to at the moment
  - We will need a build chain with all the architectures we support. (with power and Z on the way this could get complicated very quickly)
2. We ship source code to the host and compile there
  - Needs go tooling and dependencies on the host
3. Build an RPM with our changes
  - RPMs are kind of annoying to deal with
  - Need to deal with releng
  - Hard to install on coreos

I see the benefits that @omertuc is referencing, but practically I'm not sure how we would ship something like this to a coreos host without a container image.

docs/enhancements/boot-status-reporter.md
```diff
@@ -0,0 +1,143 @@
...
+   and the new host progress stages should be skipped in such a scenario.
+
+## Proposal
+
+
+### Changes To Assisted Service
+For cases as mentioned in `Motivation`, we need a new service (started via `systemd`) as soon as possible after hosts
+are booting from the disk and contact `assisted-service` to:
+
+1. Change the installation stage to a new `Booted with local ignition` between `Rebooting` and `Configuring`.
```
filanov, line 48: how can we identify that the host already was installed? 
for example if the host just reboot for some reason the service will start again no? what will it do?

filanov (approved):

PR #4451, opened 2022-09-29
MGMT-8460: Automatically rename discovered hosts

jhernand:
Currently host that use a forbidden name like `hostname` need to be manually renamed by users before proceeding to the installation. That isn't convenient for users. To avoid it this patch changes the code so that when the inventory is updated those hosts will be automatically renamed with a name calculated from the MAC address of a network interface card of the host that has a non-local IP address.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Tested manually adding a host with `localhost.localdomain` as the name and
checking that it was automatically renamed.

## Checklist

## Reviewers Checklist

### DIFF
--- a/config/crd/bases/agent-install.openshift.io_agents.yaml
+++ b/config/crd/bases/agent-install.openshift.io_agents.yaml
@@ -34,7 +34,7 @@ spec:
       name: Stage
       type: string
     - description: The hostname of the Agent.
-      jsonPath: .status.inventory.hostname
+      jsonPath: .status.hostname
       name: Hostname
       priority: 1
       type: string
@@ -168,6 +168,18 @@ spec:
                       the Agent
                     type: string
                 type: object
+              hostname:
+                description: "Hostname is the name that will be assigned to the host
+                  when the cluster is eventually installed. This will be different
+                  to the host name included in the inventory if that is a forbidden
+                  host name (like localhost, localhost4 or localhost6) or if the user
+                  has explicitly requested a different name using the .spec.hostname
+                  field. \n When the host name included in the inventory is forbidden
+                  and the user hasn't explicitly requested a different one, this will
+                  be calculated from the MAC address of the host. For example, if
+                  the MAC address of the host is A8:CD:16:AE:79:01 then this will
+                  be a8-cd-16-ae-79-01."
+                type: string
               inventory:
                 properties:
                   bmcAddress:
--- a/internal/host/host.go
+++ b/internal/host/host.go
@@ -21,6 +21,7 @@ import (
 	"github.com/openshift/assisted-service/internal/host/hostcommands"
 	"github.com/openshift/assisted-service/internal/host/hostutil"
 	"github.com/openshift/assisted-service/internal/metrics"
+	"github.com/openshift/assisted-service/internal/network"
 	"github.com/openshift/assisted-service/internal/operators"
 	"github.com/openshift/assisted-service/internal/provider/registry"
 	"github.com/openshift/assisted-service/models"
@@ -286,6 +287,8 @@ func (m *Manager) UpdateInventory(ctx context.Context, h *models.Host, inventory
 func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster, h *models.Host, inventoryStr string, db *gorm.DB) error {
 	log := logutil.FromContext(ctx, m.log)
 
+	updates := map[string]interface{}{}
+
 	hostStatus := swag.StringValue(h.Status)
 	allowedStatuses := append(hostStatusesBeforeInstallation[:], models.HostStatusInstallingInProgress)
 	allowedStatuses = append(allowedStatuses, hostStatusesInInfraEnv[:]...)
@@ -304,6 +307,10 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster, 
 	if err != nil {
 		log.WithError(err).Errorf("not updating inventory - failed to find infra env %s", h.InfraEnvID.String())
 		return common.NewApiError(http.StatusNotFound, err)
+	}
+
+	if m.autorenameHost(h, inventory) {
+		updates["requested_hostname"] = h.RequestedHostname
 	}
 
 	if h.ClusterID != nil && h.ClusterID.String() != "" {
@@ -394,12 +401,11 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster, 
 	// If there is substantial change in the inventory that might cause the state machine to move to a new status
 	// or one of the validations to change, then the updated_at field has to be modified.  Otherwise, we just
 	// perform update with touching the updated_at field
-	return db.Model(h).Updates(map[string]interface{}{
-		"inventory":              inventoryStr,
-		"installation_disk_path": installationDiskPath,
-		"installation_disk_id":   installationDiskID,
-		"disks_to_be_formatted":  disksToBeFormatted,
-	}).Error
+	updates["inventory"] = inventoryStr
+	updates["installation_disk_path"] = installationDiskPath
+	updates["installation_disk_id"] = installationDiskID
+	updates["disks_to_be_formatted"] = disksToBeFormatted
+	return db.Model(h).Updates(updates).Error
 }
 
 func (m *Manager) UpdateMediaConnected(_ context.Context, h *models.Host) error {
@@ -1385,4 +1391,90 @@ func (m *Manager) HandleReclaimBootArtifactDownload(ctx context.Context, h *mode
 
 func (m *Manager) HandleReclaimFailure(ctx context.Context, h *models.Host) error {
 	return m.sm.Run(TransitionTypeReclaimFailed, newStateHost(h), &TransitionArgsUnbindHost{ctx: ctx, db: m.db})
+}
+
+// autorenameHost renames a host if it is necessary and possible. For example, if the name is
+// localhost.localdomain and the host has a network interface card with MAC address
+// A8:CD:16:AE:79:01 it will be renamed to a8-cd-16-ae-79-01. Returns a boolean indicating if the
+// host was actually renamed.
+func (m *Manager) autorenameHost(host *models.Host, inventory *models.Inventory) bool {
+	if host.RequestedHostname != "" {
+		return false
+	}
+	if inventory == nil || inventory.Hostname == "" {
+		return false
+	}
+	if !funk.Contains(hostutil.ForbiddenHostnames, inventory.Hostname) {
+		return false
+	}
+	if len(inventory.Interfaces) < 1 {
+		return false
+	}
+	selectedNIC, err := m.findUsableNIC(inventory)
+	if err != nil {
+		m.log.WithError(err).WithFields(logrus.Fields{
+			"host": host.ID,
+			"name": inventory.Hostname,
+		}).Error(
+			"Host hasn't been automatically renamed because an unexpected error " +
+				"occurred while trying to find a NIC with a non-local IP address " +
+				"to calculate the new name",
+		)
+		return false
+	}
+	if selectedNIC == nil {
+		m.log.WithFields(logrus.Fields{
+			"host": host.ID,
+			"name": inventory.Hostname,
+		}).Info(
+			"Host hasn't been automatically renamed because it doesn't have a NIC " +
+				"with a non-local IP address that can be used to calculate the " +
+				"new name",
+		)
+		return false
+	}
+	host.RequestedHostname = strings.ToLower(strings.ReplaceAll(selectedNIC.MacAddress, ":", "-"))
+	m.log.WithFields(logrus.Fields{
+		"host": host.ID,
+		"old":  inventory.Hostname,
+		"new":  host.RequestedHostname,
+		"nic":  selectedNIC.Name,
+		"mac":  selectedNIC.MacAddress,
+	}).Info("Host has been automatically renamed")
+	return true
+}
+
+// findUsableNIC returns a network interface card of the given host inventory that has a MAC address
+// and a non-local IP address. Returns nil if the host doesn't have such network interface card, and
+// an error if the process fails, for example if some of the IP addresses of the host can't be
+// parsed.
+func (m *Manager) findUsableNIC(inventory *models.Inventory) (result *models.Interface, err error) {
+	for _, candidate := range inventory.Interfaces {
+		hasMAC := candidate.MacAddress != ""
+		hasV4 := false
+		for _, ip := range candidate.IPV4Addresses {
+			hasV4, err = network.IsGlobalCIDR(ip)
+			if err != nil {
+				return
+			}
+			if hasV4 {
+				break
+			}
+		}
+		hasV6 := false
+		for _, ip := range candidate.IPV6Addresses {
+			hasV6, err = network.IsGlobalCIDR(ip)
+			if err != nil {
+				return
+			}
+			if hasV6 {
+				break
+			}
+		}
+		if hasMAC && (hasV4 || hasV6) {
+			result = candidate
+			return
+		}
+	}
+	return
 }
--- a/internal/host/host_test.go
+++ b/internal/host/host_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/kelseyhightower/envconfig"
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	"github.com/openshift/assisted-service/internal/common"
 	eventgen "github.com/openshift/assisted-service/internal/common/events"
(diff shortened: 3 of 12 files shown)


jhernand:
@ori-amizur @CrystalChun I see that you recently made changes around the code that this is changing. Would you be so kind to review it?

internal/host/host.go
```diff
@@ -1412,3 +1417,49 @@ func (m *Manager) HandleReclaimBootArtifactDownload(ctx context.Context, h *mode
...
+// will be renamed to a8-cd-16-ae-79-01. Returns a boolean indicating if the host was actually
+// renamed.
+func (m *Manager) autorenameHost(host *models.Host, inventory *models.Inventory) bool {
+	if host.RequestedHostname != "" {
+		return false
+	}
+	if inventory == nil || inventory.Hostname == "" {
+		return false
+	}
+	if !funk.Contains(hostutil.ForbiddenHostnames, inventory.Hostname) {
```
CrystalChun, line 1432: Suggestion to use the `ValidateHostname` function here https://github.com/openshift/assisted-service/blob/3f535bd36578f5b1a5576610f19df4daee541138/internal/host/hostutil/host_utils.go#L80

and return false if err != nil 

Since the function has all the validations for a valid hostname.

subsystem/metrics_test.go
```diff
@@ -688,12 +688,12 @@ var _ = Describe("Metrics tests", func() {
 			oldFailedMetricCounter := getValidationMetricCounter(string(models.HostValidationIDHostnameValid), hostValidationFailedMetric)
 
 			// create a validation failure
-			// 'localhost' is a forbidden host name
-			generateHWPostStepReply(ctx, h, validHwInfo, "localhost")
+			// 'my host' isn't a valid host name
```
CrystalChun, line 691: Sorry, I need some clarification, how come the hostnames in these tests were changed?

internal/host/host.go
jhernand: @CrystalChun the intent here is to automatically rename the host only if it is using a forbidden host name, not if it using an incorrect host name. For example, `my host` is an incorrect host name but we don't want to automatically replace it with the name calculated from the MAC address (or so I understood). If we use that `ValidateHostname` function here then we will be doing the rename for both forbidden and incorrect host names.

subsystem/metrics_test.go
jhernand: @CrystalChun my understanding is that these tests are posting the inventory for a host with a forbidden host name, and then checking that the metrics that count the number of failed validations are updated correctly. But now a host with a name like `localhost` is automatically renamed to the MAC address immediately after receiving the inventory, so the number of failed validations will not increase. That is what I thought that this could be replaced by an incorrect host name that isn't automatically replaced.

internal/host/host.go
CrystalChun: Oh gotcha! How come we're only renaming forbidden and not incorrect host names?

subsystem/metrics_test.go
CrystalChun: Makes sense! Thanks! I misunderstood when the renaming would occur and thought it would be for both forbidden and incorrect hostnames

internal/host/host.go
jhernand: That is my interpretation of the ticket: https://issues.redhat.com/browse/MGMT-8460. I assume that we only want to do this automatic renaming when the user didn't explicitly set a name for the host. I may be wrong. Can you take a look at that ticket and share your thoughts?

CrystalChun: I see, I do agree with the assumption to set a name when the user didn't explicitly set a name. 

However, I'm wondering if it should automatically change the hostname when the inventory hostname is:
1. just forbidden (like in this check) 
2. forbidden AND invalid (e.g. `my host` or a hostname > 64 characters)

I'm not too sure which experience would be better for the user. 

E.g. If they have a host that doesn't have a forbidden name but the hostname is invalid (>64 characters) would it be better to warn the user, change it automatically, or something else?

jhernand: @CrystalChun I'd say that would want to automatically rename the host only when it is `localhost` (or any of its variants). My reasoning is that when the host name is `localhost` it most probably means that the user didn't make any effort to assign a name to the host. In cases like `my host` or longer than 64 characters it means that the user made the effort, but incorrectly, so we would want to tell her, and her to fix it, like we do today.

CrystalChun: That's a good point to only assign the host name when the user hasn't attempted to change it. I agree with you! Thanks for explaining!

internal/host/host.go
```diff
@@ -1412,3 +1417,49 @@ func (m *Manager) HandleReclaimBootArtifactDownload(ctx context.Context, h *mode
...
+		return false
+	}
+	if len(inventory.Interfaces) < 1 {
+		return false
+	}
+	var selectedIf *models.Interface
+	for _, candidateIf := range inventory.Interfaces {
+		if candidateIf.MacAddress != "" {
```
CrystalChun, line 1440: It might be possible for two hosts to have the same mac address https://issues.redhat.com//browse/MGMT-11888

jhernand: Thanks for pointing that out @CrystalChun! In the new version of the patch I added logic to select as main interface the first one that has both a MAC address and an non-local (not 127.0.0.1 or ::1) IP address. I think this should address this issue. I believe it is not necessary to add additional information to the name to make it unique, because MAC addresses of interfaces that are actually connected to the network should be unique. Please take another look and let me know what you think.

internal/host/host.go
```diff
@@ -1412,3 +1418,82 @@ func (m *Manager) HandleReclaimBootArtifactDownload(ctx context.Context, h *mode
...
+}
+
+// isLocalIP determines if the given IP address is a local one, like 127.0.0.1 or ::1.
+func (m *Manager) isLocalIP(ip string) bool {
+	parsed := net.ParseIP(ip)
+	if parsed == nil {
+		m.log.WithFields(logrus.Fields{
+			"ip": ip,
+		}).Info("Failed to parse IP address, will assume that isn't local")
+		return false
```
CrystalChun, line 1496: Sorry this logic is a little confusing to me, if it's an IP that can't be parsed, shouldn't it fail the condition? Since the condition used looks for a `false` in order to set the IP https://github.com/openshift/assisted-service/pull/4451/files#diff-549fd6f951612ce99043912371fffacd86d8249a40ef3905825be3d45f81dba4R1477

internal/host/host.go
CrystalChun: Sounds good! Reviewed and had one comment. Would also ask @mkowalski to take a look too!

internal/host/host.go
jhernand: This methods returns `true` if the IP is local (like `127.0.0.1` or `::1`) and `false` otherwise. For unparseable IP addresses, for example `junk`, it returns `false`. The logic being that `junk` is not `127.0.0.1` or `::1`. It is indeed weak logic because `junk` it isn't a valid IP address at all. But it doesn't affect much the rest of the system because I assume that we will never have an unparseable IP address in the inventory. I will rework this to return an error in that case.

jhernand: @CrystalChun the new version of the patch handles the non parseable addresses returning an error. Hope that makes it clearer.

internal/host/host.go
```diff
@@ -1412,3 +1418,99 @@ func (m *Manager) HandleReclaimBootArtifactDownload(ctx context.Context, h *mode
...
+			result = candidate
+			return
+		}
+	}
+	return
+}
+
+// isGlobal determines if the given IP address is a global one and not a local one like 127.0.0.1
+// or ::1. Returns an error if the IP address can't be parsed.
+func (m *Manager) isGlobal(ip string) (result bool, err error) {
```
mkowalski, line 1508: Just a nit regarding naming: I would rather have the function name a more verbose, because currently we are in the `host` package and look at `isGlobal` function. Anyone seeing it outside of where it's implemented will be confused about it's meaning, i.e. seeing a function `host.isGlobal("some string")` it's hard to guess it's about IP address.

Apart from the above, I would think whether we could move this function to `internal/network/utils.go` or `internal/network/family_utils.go` as those files already have bunch of helper functions related to networking. Apart from this, I can see like such a function could be reused in other places so having it as a public utility function is useful (we are doing "is address non-local" validation in other places too)

internal/host/host.go
```diff
@@ -1412,3 +1418,99 @@ func (m *Manager) HandleReclaimBootArtifactDownload(ctx context.Context, h *mode
...
+		"nic":  mainNIC.Name,
+		"mac":  mainNIC.MacAddress,
+	}).Info("Host has been automatically renamed")
+	return true
+}
+
+// mainNIC returns the main network interface of the given host inventory, or nil if there is no
+// such network interface card. The main network interface card is the first that has assigned both
+// a MAC address and a global IP address.
+func (m *Manager) mainNIC(inventory *models.Inventory) (result *models.Interface, err error) {
```
mkowalski, line 1475: Also a naming nit: I have a hard time with digesting the concept of using the name "main network interface" for an interface that just have a MAC and IP address assigned. I definitely like the logic you do here so that you will pick an usable NIC and build a hostname out of it. It's just that in a global perspective "main network interface" may have another meaning.

I wonder if something as simple as `findUsableNIC` may satisfy what you are trying to do here? Given that the result is not guaranteed to be deterministic as a function of the host (because the content of `inventory.Interfaces` has no guaranteed order) but also given that we effectively don't care which NIC is used for this autorenaming, this could be some tradeoff.

If you really really wanted it to be about detecting the main network interface, then I'd suggest not picking "the first" interface but rather the interface that matches the Machine Networks of the cluster. But this path requires you to parse the whole Cluster object here and cross-check it with host's inventory. So looks like a big overkill for what you try to achieve

internal/host/host.go
```diff
@@ -1412,3 +1418,99 @@ func (m *Manager) HandleReclaimBootArtifactDownload(ctx context.Context, h *mode
...
+		return false
+	}
+	mainNIC, err := m.mainNIC(inventory)
+	if err != nil {
+		m.log.WithError(err).WithFields(logrus.Fields{
+			"host": host.ID,
+			"name": inventory.Hostname,
+		}).Error(
+			"Host hasn't been automatically renamed because an unexpected error " +
+				"ocurred while trying to find the main network interface card",
```
mkowalski, line 1446: As already wrote, as a user I would be quite confused if I've seen `trying to find the main network interface card` error. Maybe something as simple as `error [...] while trying to find a NIC with non-local IP address`? This implicitly carry on the assumption that NIC must have a MAC address should be true for all the types of interfaces we support in our stack.

internal/host/host_test.go
```diff
@@ -4519,3 +4520,270 @@ var _ = Describe("Rebooting day2", func() {
...
+						Name:          "eth0",
+						MacAddress:    "71:A0:A4:6F:BE:C8",
+						IPV4Addresses: []string{"192.168.0.1"},
+					},
+				},
+				Expected: "71-a0-a4-6f-be-c8",
+			},
+		),
+		Entry(
+			"Ignores NIC with MAC but no IP address",
```
mkowalski, line 4709: Typo alert: `ingress`

internal/host/host_test.go
```diff
@@ -4519,3 +4520,270 @@ var _ = Describe("Rebooting day2", func() {
...
+						Name:          "eth1",
+						MacAddress:    "42:5a:90:c8:24:dc",
+						IPV4Addresses: []string{"192.168.0.1"},
+					},
+				},
+				Expected: "42-5a-90-c8-24-dc",
+			},
+		),
+		Entry(
+			"Ignores NIC with local IPv4 address",
```
mkowalski, line 4729: Typo(?): `loopback` instead of `local`. Local IP address is something different

internal/host/host_test.go
```diff
@@ -4519,3 +4520,270 @@ var _ = Describe("Rebooting day2", func() {
...
+						Name:          "eth1",
+						MacAddress:    "42:5a:90:c8:24:dc",
+						IPV4Addresses: []string{"192.168.0.1"},
+					},
+				},
+				Expected: "42-5a-90-c8-24-dc",
+			},
+		),
+		Entry(
+			"Ignores NIC with local IPv6 address",
```
mkowalski, line 4749: Typo(?): `loopback` instead of `local`. Local IP address is something different

mkowalski:
Thanks a lot! I added a bunch of comments but they are related mostly to naming and organization. The overall design and functionality looks good to me.

internal/host/host.go
mkowalski: Thanks, looks safe after the change. The bug in subject was exposed for NICs that were using link-local addresses and those will return `false` when fed into `net.IsGlobalUnicast`

internal/host/host.go
jhernand: Renamed to `IsGlobalIPAddress` and moved to `internal/network/utils.go`.

internal/host/host.go
jhernand: Renamed to `findUsableNIC`. I agree that cross checking with the cluster networks isn't required here.

internal/host/host.go
jhernand: Done, removed the "main NIC" words everywhere.

internal/host/host_test.go
jhernand: Done.

jhernand:
Thanks for the review @mkowalski! I think that the new version of the patch addresses all your concerns. Please have another look.

mkowalski:
/lgtm

Thank you!

internal/host/host.go
CrystalChun: Looks good! Thanks Juan!

internal/host/host.go
CrystalChun: Thanks for taking a look @mkowalski!

CrystalChun:
/lgtm
Good work @jhernand and thanks for hopping in to review @mkowalski!

jhernand:
@CrystalChun @mkowalski the subsystem tests were failing because I didn't realize that we store the IP addresses in the inventory with the network mas, something like `127.0.0.1/32`. The code was expecting it without the network mask, and therefore failing. I have fixed that in the new version of the patch. I need your review again.

jhernand:
/hold

I need to check why the `edge-e2e-ai-operator-ztp` job is failing, and also want to test this manually with the Kubernetes operator.

mkowalski:

Last 2 failures of this job were not related to the change

internal/host/host.go
```diff
@@ -1412,3 +1418,89 @@ func (m *Manager) HandleReclaimBootArtifactDownload(ctx context.Context, h *mode
...
+	if host.RequestedHostname != "" {
+		return false
+	}
+	if inventory == nil || inventory.Hostname == "" {
+		return false
+	}
+	if !funk.Contains(hostutil.ForbiddenHostnames, inventory.Hostname) {
+		return false
+	}
+	if len(inventory.Interfaces) < 1 {
```
filanov, line 1436: how can that happen? doesn't it mean that host was not able to register?

mkowalski: I can imagine a scenario when the host has outside connectivity, but because we have a strictly-defined list of interface types that we collect, the list is empty.

If for example my server uses interface type `chocobomb` because I have a custom driver for it and at the same time it has no ethernet interfaces, we will end up in a scenario when the host has been discovered (because IP connectivity is available and kernel can route traffic) but Assisted does not see any interface because our collector ignores interfaces of type `chocobomb`

filanov: good to know thanks

mkowalski:

It failed for a reasonably valid reason, but I'm not convinced that it's you who introduced this, therefore blindly retest

jhernand:
@CrystalChun @mkowalski I was asked to test this with the Kubernetes API and I found that the calculated host name is not reflected at all in the `Agent` object. I discussed that with @filanov and @omertuc and we agreed to add a new `assigned_hostname` field to the API, to the database and to the status of the `Agent` object. I think that is conceptually correct, but unfortunately it requires a lot of changes to the server and also to the UI. So instead of doing that I just added a new `.status.hostname` field that contains a copy of the `requested_hostname` field of the API. That should have almost the same effect, and doesn't require so many changes. Please take a look and let me know what you think. Note that there is no change in the logic that renames the host; the only change is adding the new `.status.localhost` field to the `Agent` type.

mkowalski:
The trick with `Agent.status.hostname` looks good to me, just for the sake of completeness cc @avishayt

avishayt:
Overall looks OK but let me do a full code review on Wednesday.  We should also review with UX. @nirfarkas

omertuc:
>added a new .status.hostname field that contains a copy of the requested_hostname field of the API

But an automatically generated hostname should not be in requested_hostname, so how would that automatically generated hostname be reflected in the status?

config/crd/bases/agent-install.openshift.io_agents.yaml
```diff
@@ -168,6 +168,18 @@ spec:
...
+              hostname:
+                description: "Hostname is the name that will be assigned to the host
+                  when the cluster is eventually installed. This will be different
+                  to the host name included in the inventory if that is a forbidden
+                  host name (like localhost, localhost4 or localhost6) or if the user
+                  has explicitly requested a different name using the .spec.hostname
+                  field. \n When the host name included in the inventory is forbidden
+                  and the user hasn't explicitly requested a different one, this will
+                  be calculated from the MAC address of the host. For example, if
+                  the MAC address of the host is A8:CD:16:AE:79:01 then this will
```
omertuc, line 180: There's no "The MAC address" for a host, what if it has multiple interfaces? Phrasing should be adjusted

internal/host/host.go
```diff
@@ -306,6 +309,10 @@ func (m *Manager) updateInventory(ctx context.Context, cluster *common.Cluster,
 		return common.NewApiError(http.StatusNotFound, err)
 	}
 
+	if m.autorenameHost(h, inventory) {
```
omertuc, line 312: So are we officially giving up on delayed rDNS? Can you please state this in the commit message and a code comment?

omertuc (changes requested):

internal/host/host.go
jhernand: Yes @omertuc, that is one of the side effects of not adding an additional field: there is no way to know if the name in "requested_hostname" came from the user or was automatically calculated from a MAC address. So either we allow host name discovered via reverse DNS after the initial discovery to override the hostname explicitly requested by the user or else we ignore it. I will explain that.

config/crd/bases/agent-install.openshift.io_agents.yaml
jhernand: @omertuc I wanted to be vague there on purpose, because the algorithm is not deterministic, and I think explaining it to the user will not be very useful. It would be something like this:

>  The host name will be calculated from the MAC address of one of the network interfaces
>  of the host that has both a MAC address and a non-local IP address.

Note the "one of the network interfaces" part. That is what the code does; it iterates the list of network interfaces
of the host and picks the first that has a MAC and a non-local IP address. If I understand correctly there is no
guarantee of any kind of ordering in that list, so the selected network interface may change if the agent or
the service decide to sort this list in a different way. @mkowalski mentioned that we could make this selection
deterministic by looking at the cluster networks and selecting a network interface that is connected to the
machine network of the cluster. I thought that the benefit of doing that is not worth the effort: all we need here
is a unique host name. Do we want make this deterministic?

internal/host/host.go
omertuc: Also it would be nice to investigate whether my hypothetical scenario is something that happens often in prod or not, before we go through with this

config/crd/bases/agent-install.openshift.io_agents.yaml
omertuc: I have no opinion about the interface selection itself, just about the description being literally incorrect. 

Vagueness can be achieved by just saying a instead of the.

Ideally we should pick the machine network interface, I think we already have functions that do that if you feel like reusing them, but again I'm unopinionated about it

jhernand:
> But an automatically generated hostname should not be in requested_hostname, so how would that automatically generated hostname be reflected in the status?

That is the key point @omertuc. If we don't use the "requested_hostname" field to store the automatically calculated host name then we need to add a new one and change the API and the UI to use it. That is the conceptually robust thing to do, but it requires additional effort, something like this:

1. Change the UI so that it uses the new "assigned_hostname" field, and if not present then it uses the old "requested_hostname" field. This would need to be deployed before the new version of the service.
2. Deploy the new version of the service that populates and returns the new "assigned_hostname" field.
3. Change the UI so that it no longer uses the "requested_hostname" field. This can be deployed independently of additional changes to the service.

In the service side I think that the right thing to do would be to have that new "assigned_hostname" field and an additional "assigned_hostname_source" field. This "..._source" would be an enumerated value indicating what was the source for the assigned hostname: the user, the inventory or the service. That would more explicitly represent the reality of how these host names are assigned, and would make the code clearer.

My only concern about that is that it will take longer, as it requires additional effort in both the service and the UI, as well as more coordination for the phased deployment of the service and the UI.

config/crd/bases/agent-install.openshift.io_agents.yaml
avishayt: The algorithm should definitely be deterministic

avishayt:
> > But an automatically generated hostname should not be in requested_hostname, so how would that automatically generated hostname be reflected in the status?
> 
> That is the key point @omertuc. If we don't use the "requested_hostname" field to store the automatically calculated host name then we need to add a new one and change the API and the UI to use it. That is the conceptually robust thing to do, but it requires additional effort, something like this:
> 
> 1. Change the UI so that it uses the new "assigned_hostname" field, and if not present then it uses the old "requested_hostname" field. This would need to be deployed before the new version of the service.
> 2. Deploy the new version of the service that populates and returns the new "assigned_hostname" field.
> 3. Change the UI so that it no longer uses the "requested_hostname" field. This can be deployed independently of additional changes to the service.
> 
> In the service side I think that the right thing to do would be to have that new "assigned_hostname" field and an additional "assigned_hostname_source" field. This "..._source" would be an enumerated value indicating what was the source for the assigned hostname: the user, the inventory or the service. That would more explicitly represent the reality of how these host names are assigned, and would make the code clearer.
> 
> My only concern about that is that it will take longer, as it requires additional effort in both the service and the UI, as well as more coordination for the phased deployment of the service and the UI.

Would it be better to move all of this into the agent, to override the discovered hostname only in case no hostname was discovered?  Then if a hostname is discovered later it will override this generated one, and there is no conflict with requested_hostname, which only the user should modify?

jhernand:
> Would it be better to move all of this into the agent, to override the discovered hostname only in case no hostname was discovered? Then if a hostname is discovered later it will override this generated one, and there is no conflict with requested_hostname, which only the user should modify?

@avishayt here is the equivalent change in the agent: https://github.com/openshift/assisted-installer-agent/pull/449 . Let me know if I should close this.

omertuc:
I like @avishayt 's idea, let's close #4451

PR #4455, opened 2022-09-29
MGMT-12188: Fix panic when infraEnv image has not been created

carbonin:
Previously the preprovisioning controller was unconditionally looking
for the created time in the infraEnv status. This would cause the entire
assisted-service container to crash if the created time was not set.

This is only set after the infraEnv controller sets the image URLs and
may not be set for some time if there are, for example, format errors in
nmstate config associated with the infraEnv. Once the container starts
crashing it is unlikely to recover even if the underlying issues are
resolved as the preprovisioning controller seemed to consistently reconcile
before the infraenv controller got a chance to set the created time.

This patch checks that the image created time is set and sets a
condition if not. The preprovisionsing image controller is watching for
infraenv events so it will reconcile when the image gets created.

https://issues.redhat.com/browse/MGMT-12188

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/preprovisioningimage_controller.go
+++ b/internal/controller/controllers/preprovisioningimage_controller.go
@@ -114,6 +114,18 @@ func (r *PreprovisioningImageReconciler) Reconcile(origCtx context.Context, req 
 		return r.AddIronicAgentToInfraEnv(ctx, log, infraEnv)
 	}
 
+	if infraEnv.Status.CreatedTime == nil {
+		log.Info("InfraEnv image has not been created yet")
+		setNotCreatedCondition(image)
+		err = r.Status().Update(ctx, image)
+		if err != nil {
+			log.WithError(err).Error("failed to update status")
+			return ctrl.Result{}, err
+		}
+		// no need to requeue, the change in the infraenv should trigger a reconcile
+		return ctrl.Result{}, nil
+	}
+
 	// The image has been created sooner than the specified cooldown period
 	imageTimePlusCooldown := infraEnv.Status.CreatedTime.Time.Add(InfraEnvImageCooldownPeriod)
 	if imageTimePlusCooldown.After(time.Now()) {
@@ -178,6 +190,17 @@ func (r *PreprovisioningImageReconciler) setImage(generation int64, status *meta
 		metal3_v1alpha1.ConditionImageError, imageErrorStatus,
 		reason, message)
 	return nil
+}
+
+func setNotCreatedCondition(image *metal3_v1alpha1.PreprovisioningImage) {
+	message := "Waiting for InfraEnv image to be created"
+	reason := imageConditionReason(strcase.ToCamel(message))
+	setImageCondition(image.GetGeneration(), &image.Status,
+		metal3_v1alpha1.ConditionImageReady, metav1.ConditionFalse,
+		reason, message)
+	setImageCondition(image.GetGeneration(), &image.Status,
+		metal3_v1alpha1.ConditionImageError, metav1.ConditionFalse,
+		reason, message)
 }
 
 func setCoolDownCondition(image *metal3_v1alpha1.PreprovisioningImage) {
--- a/internal/controller/controllers/preprovisioningimage_controller_test.go
+++ b/internal/controller/controllers/preprovisioningimage_controller_test.go
@@ -288,6 +288,15 @@ var _ = Describe("PreprovisioningImage reconcile", func() {
 			Expect(infraEnv.ObjectMeta.Annotations[EnableIronicAgentAnnotation]).To(Equal("true"))
 		})
 
+		It("doesn't fail when the infraEnv image has not been created yet", func() {
+			infraEnv.Status = aiv1beta1.InfraEnvStatus{}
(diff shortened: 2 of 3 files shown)


internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -114,6 +114,17 @@ func (r *PreprovisioningImageReconciler) Reconcile(origCtx context.Context, req
...
 
+	if infraEnv.Status.CreatedTime == nil {
+		log.Info("InfraEnv image has not been created yet. Requeuing and retrying again soon")
+		setNotCreatedCondition(image)
+		err = r.Status().Update(ctx, image)
+		if err != nil {
+			log.WithError(err).Error("failed to update status")
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true, RequeueAfter: InfraEnvImageCooldownPeriod}, nil
```
nmagnezi, line 125: Have we considered adding a `Watches` on InfraEnv resources instead of this requeue?
An example of how we do it on other places (see NMState): 
https://github.com/openshift/assisted-service/blob/80c73bc2ce02fd4d9ac84d47776f8c49b7d0b47a/internal/controller/controllers/infraenv_controller.go#L725-L731

carbonin: We have a watch on an event channel already. Let me check to see if we'll get an event for this.
I did this mostly because this is how it is already being done for the "cooldown" check below, but I suppose that's a different use case.

carbonin: I suppose using that would still give us a race between the infraenv controller saving a change back to the kubeapi and the preprovisioning image controller reading that infraenv from the kube api after getting an event from the service code, right?

Feels like a direct watch would be better assuming we only don't need to be notified in situations where nothing has actually changed in the CR.

Curious about why we went this direction initially ...

Based on the [original commit](https://github.com/openshift/assisted-service/commit/18fcad2904816ba9023f2fec1550089989d5f531) that added this in place of a direct watch it sounds like we have an underlying issue that coincidentally I've trying to track down separately. It feels like we should be updating the urls in the infraenv CR more frequently than we currently are.

I was doing a test with static networking config and couldn't seem to get the new config to propagate through ironic to the hosts and I think this is why. The URL must not be changing which would not trigger ironic to re-download the initrd/iso.

I'll put a hold on this until I can sort out this issue.

carbonin:
/hold

Trying to sort out an underlying issue with infraEnvs not updating urls when they should be.
See https://github.com/openshift/assisted-service/pull/4455#discussion_r984014142 for more details.
Solving this might allow the preprovisioning image controller to watch infraenvs in a more normal way and remove the need for the requeue in this patch.

That said, this is actually crashing the service so if tracking down the issue with infraenvs takes a while or the fix is invasive it might be worth merging this and following up to improve it later.

internal/controller/controllers/preprovisioningimage_controller.go
carbonin: I'm working on just adding a subsystem test to ensure this all works as we assume it will.

I'm going to try to remove this requeue, but not change the current watches and make sure the new test reproduces the issue I saw originally and also passes with these changes (adds the new condition and then sets the image when the infraenv is fixed).

eranco74: Unsure I understand the issue here, the preprovisioning image controller is watching infraEnv:
https://github.com/openshift/assisted-service/blob/3b5982f3d050dae906fef918d5ece9039de58aa8/internal/controller/controllers/preprovisioningimage_controller.go#L256
I think we can remove the requeue

eranco74 (commented):
/lgtm

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -180,6 +192,17 @@ func (r *PreprovisioningImageReconciler) setImage(generation int64, status *meta
 	return nil
 }
 
+func setNotCreatedCondition(image *metal3_v1alpha1.PreprovisioningImage) {
+	message := "Waiting for InfraEnv image to be created"
+	reason := imageConditionReason(strcase.ToCamel(message))
+	setImageCondition(image.GetGeneration(), &image.Status,
+		metal3_v1alpha1.ConditionImageReady, metav1.ConditionFalse,
+		reason, message)
+	setImageCondition(image.GetGeneration(), &image.Status,
```
nmagnezi, line 201: why do we need both `metal3_v1alpha1.ConditionImageReady` and `metal3_v1alpha1.ConditionImageError`?

eranco74: You are asking why we need to update it here or why do we need the ConditionImageError in general?

carbonin: The image could not be ready, but also not be encountering an error as in this case and the cooldown case.
Or the image could not be ready and there could be an error that the user needs to handle.

We need to be able to express both situations clearly.

PR #4457, opened 2022-09-30
MGMT-12193: Add enhancement doc for collecting on-prem data

CrystalChun:
[MGMT-12193](https://issues.redhat.com/browse/MGMT-12193)
Epic: [MGMT-10308](https://issues.redhat.com/browse/MGMT-10308)

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/on-prem-data-collection.md
+++ b/docs/enhancements/on-prem-data-collection.md
@@ -1,0 +1,168 @@
+---
+title: on-prem-data-collection
+authors:
+  - "@CrystalChun"
+creation-date: 2023-01-18
+last-updated: 2023-01-18
+---
+
+# On-premises Data Collection
+
+## Summary
+
+The Assisted Installer ("AI") will provide a process to collect event data about cluster installations when it is deployed on-premises as part of either [Red Hat Advanced Cluster Management for Kubernetes](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.6) ("RHACM") or as a stand-alone Podman deployment. Data was already collected from clusters installed using [Red Hat's Hybrid Cloud Console](https://console.redhat.com/openshift/assisted-installer/clusters) (“SaaS”), but there was no method for obtaining the same data from on-premises deployments.
+
+## Motivation
+
+Data provides insight in order to improve the product for our users.
+
+### Goals
+
+1. Create a process to send [event](https://github.com/openshift/assisted-service/tree/master/docs/dev/events.md) data from on-premises deployments of AI.
+2. Analyze and process on-premises data (by use of visualizations, dashboards, etc.) at Red Hat similarly to SaaS data.
+3. Analyze on-premises data both separately and together with SaaS data.
+
+
+### Non-Goals
+
+1. Implementing new data types to collect is out of the scope of this epic. This epic defines how and when the data already collected will be processed and sent.
+2. Define how SaaS should redesign their data pipeline (taken care of in [MGMT-11244](https://issues.redhat.com/browse/MGMT-11244)) is out of the scope of this epic. This epic only refers to on-premises deployments of AI.
+3. Define the creation and/or management of an event streaming instance such as Kafka and the endpoint this data will be hosted at (both taken care of in [MGMT-11244](https://issues.redhat.com/browse/MGMT-11244)).
+4. Collecting must-gather data (and logs) for debugging failed cluster installation is outside the scope of this epic. This epic is only concerned with [event](https://github.com/openshift/assisted-service/blob/6b8401d3b0f0ceea5a837f33c17b9b1cd5ec52a3/internal/events/api/event.go) data.
+5. Data formatting and handling by the event streaming service is out of the scope of this epic. The event streaming service should be able to process the data from all versions of the on-premises environment.
+    - Similarly, the handling of data received from the event streaming service is out of the scope of this epic.
+
+## Proposal
+
+### Data Sending Flow
+
+All of the event data for a cluster will be queried from the DB, packaged, and sent to the data hosting endpoint when a cluster installation attempt finishes (regardless of success or failure).
+
+Sending data should not interfere with customer workloads and deployments.
+
+NOTE: Data from pre-existing (already installed) cluster(s) using previous/older versions of AI will not be sent when this feature is enabled.
+
+### Assisted-Service Changes 
+
+Changes that need to be made in the `Assisted-Service` to support this:
+- Needs to know what the data hosting service endpoint is and how to authenticate into it
+- Needs code changes when an install is attempted to query and send data
+- Needs to know how to format the data to send to the data hosting service
+    - Includes sanitizing the data such as hashing user's emails like in the [event scraper](https://github.com/openshift-assisted/assisted-events-scrape/blob/master/assisted-events-scrape/utils/anonymizer.py#L25)
+- Needs to expose an option for users to opt-out of this feature
+
+### Additional Changes
+
+- Documentation about how to opt-out
+- Documentation of requirements: non-disconnected installation
+
+### User Stories
+
+#### Story 1
+
+As an infrastructure operator, I can provision and manage my fleet of clusters knowing that data points are automatically collected and sent to the Assisted Installer team without having to do any additional configuration. I know my data will be safe and secure and the team will only collect data they need to improve the product. 
+
+#### Story 2
+
+As a developer on the Assisted Installer team, I can analyze the customer data to determine if a feature is worth implementing/keeping/improving. I know that the customer data is accurate and up-to-date. All of the data is parse-able and can be easily tailored to the graphs/visualizations that help my analysis.
+
+#### Story 3
+
+As a product owner, I can determine if the product is moving in the right direction based on the actual customer data. I can prioritize features and bug fixes based on the data.
+
+### Implementation Details/Notes/Constraints
+
+Assisted Service changes:
+1. New environment variables to be added:
+    - `ENABLE_DATA_COLLECTION`: If set to `false`, it will disable this feature and thereby allow a customer to opt-out, otherwise the feature is enabled by default.
+    - `DATA_UPLOAD_ENDPOINT`: The URL the data will be sent to.
+2. Add data collection and sending logic
+    - Collecting and sending: querying the DB for events of that cluster, sanitizing the data, and sending it as a tar.gz to the `DATA_UPLOAD_ENDPOINT` over HTTPS.
+    - Called when a cluster reaches one of the following states: `error`, `cancelled`, `installed`
+            - Each of these cluster stages represent an end state and that the cluster attempted installation.
+    - Sending data is non-blocking and sent as best effort
+3. Add logic for authenticating to the `DATA_UPLOAD_ENDPOINT` by using the user's OCM pull secret
+4. Add opt-out logic: 
+    - If the user has [opted out of Telemetry](https://docs.openshift.com/container-platform/4.11/support/remote_health_monitoring/opting-out-of-remote-health-reporting.html) (missing cloud.openshift.com from the global pull secret)
+    - If the user has set `ENABLE_DATA_COLLECTION` to `false`.
+
+### Risks and Mitigations
+
+#### Risk 1: Overusing Customer's Network & CPU
+
+On-premises users might have large-scale clusters so network usage and processor usage could be impacted depending on how large the data set is.
+
+##### Mitigation
+
+To avoid user workflow interruption, data collection, processing, and sending can happen as a concurrent thread. Data sending should be non-blocking and best effort.
+
+#### Risk 2: Storage Limits and Data Loss
+
+Large-scale clusters most likely means large data sets that are sent to the event streaming service. There is a risk that amount of data could overflow the storage limits of the server used.
+
+Furthermore, if the limits are reached, how will newer data be handled? Will the pre-existing data get overwritten or will the new data be rejected? There's potential for lost data here.
+
+##### Mitigation
+
+More research needs to be done here to see what the limits are and how we can overcome them. If using RH Ingress, there could be a daily job that would export data from here and import it into our own storage.
+
+#### Risk 3: DDoS-ing Insights Ingress Server
+
+Without any constraints, it's possible for the assisted service to send data to the data collection server repeatedly. This can prevent other clients from sending their data, and cause loss of data, or prevent us from reaching the data since the server will be busy with the repeated connections.
+
+##### Mitigation
+
+1. Clearly defining when events are sent (see `Data Sending Flow` under the `Proposal` section above)
+2. Data is sent as best effort
+3. Retry will be implemented as a future enhancement
+
+#### Risk 4: Duplicate Data
+
+With all the cases for sending data, there's a high chance for duplicated data to be sent to the data collection server. It can skew the analysis of data and it won't be representative of the actual data from the field.
+
+##### Mitigation
+
+A field can be used as a marker to indicate the data has been sent to the event streaming service or that the data already exists. One option is to use the timestamp, cluster ID, and user email to detect duplicate events.
+
+## Design Details
+
+### UI Impact
+
+No UI changes will need to be made.
+
+### Test Plan
+
+#### Test 1
+
+Add a check at the end of a cluster install attempt that will query the data hosting endpoint and verify the payload made it to the data hosting service.
+
+#### Test 2
+
+Check against a large-scale cluster with > 100 hosts to verify:
+1. All of the cluster's events are being stored, queried, and sent (within a reasonable time)
+2. That the event streaming can receive and process the large data set
+
+## Drawbacks
+
+- Dependency on the data hosting service selected in [MGMT-11244](https://issues.redhat.com/browse/MGMT-11244) could cause delays in this epic being completed.
+- Won't get data from disconnected environment users since this solution requires a connection to push the data
+
+## Alternatives
+
+### Alternative Solution 1: Customer Uploads Data
+
+#### Details
+
+The Assisted Service would provide an endpoint that the customer can query to get all data that they would upload to us.
+
+#### Benefits
+
+- Support disconnected deployments since the service will not be pushing the data.
+- Minimizes risk of DDoS-ing the data streaming service
+
+#### Drawbacks
+
+- Likely will not get data
+    - Due to the customer having to do extra work to upload the data
+- Data would be skewed since we'd only get data when the customer uploads it
+    - Customer is most likely only going to upload when they have an issue


docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+2. Add additional logic when a cluster reaches one of the following end stages: `error`, `cancelled`, `installed`
+    - Each of these cluster stages represent an end state and that the cluster attempted installation.
+    - Additional logic includes: querying the DB for events of that cluster and streaming this event data to the `DATA_COLLECTION_URL` over HTTPS.
+        - If the customer has large-scale clusters, would it be better to always compress the data in a separate thread?
+3. Add logic for authenticating to the `DATA_COLLECTION_URL` 
+    - Currently unknown since the service to host this data hasn't been decided on yet, but it could take the form of using a Kubernetes Secret
+
+### Risks and Mitigations
+
+On-premises users might have large-scale clusters so network usage and processor usage could be impacted depending on how large the data set is.
```
gamli75, line 81: what will happen if the on prem environment will continue to send the data over and over again and will cause to overload in the data flow? we need to have the ability to stop it

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Story 2
+
+As a developer on the assisted installer team, I can analyze the customer data to determine if a feature is worth implementing/keeping/improving. I know that the customer data is accurate and up-to-date. All of the data is parse-able and can be easily tailored to the graphs/visualizations that help my analysis.
+
+#### Story 3
+
+As a product owner, I can determine if the product is moving in the right direction based on the actual customer data. I can help determine which features and bug fixes to prioritize based on the data.
+
+### Implementation Details/Notes/Constraints
```
gamli75, line 66: Do we want to add a retry mechanism for sending the data in case the DATA_COLLECTION_URL is not available?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+### Risks and Mitigations
+
+On-premises users might have large-scale clusters so network usage and processor usage could be impacted depending on how large the data set is. 
+
+## Design Details
+
+### Open Questions
+
+1. Are we planning to support disconnected environments in the future?
+2. Does any of the event data we store contain Personally Identifiable Information? Does this need to be scrubbed or modified before we push this feature out?
```
gamli75, line 88: It's not only the event data, it's also the cluster, infra-env & host data. The only problematic field is the user email and we are hashing it inside the event-scrape so we will need to do the same in the service

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+
+### UI Impact
+
+No UI changes will need to be made.
+
+### Test Plan
+
+Add a check at the end of a cluster install attempt that will query the correct data made it to the data hosting service if the feature is enabled.
```
gamli75, line 96: check it with large-scale cluster - more than 100 hosts

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+To provide a better installation process and experience, the Assisted Installer ("AI") will collect data about installations when it is deployed on-premises as part of either [Red Hat Advanced Cluster Management for Kubernetes](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.6) ("RHACM") or as a stand-alone Podman deployment. Data was already collected from clusters deployed on [Red Hat's Hybrid Cloud Console](https://console.redhat.com/openshift/assisted-installer/clusters) (“SaaS”). The collection of data from on-premises deployments will give the team additional insight into the product's usage, customers' pain points, and analytics for how to best improve the product.
+
+## Motivation
+
+Data helps us see how the product is being used, helps us to analyze trends, and to determine ways we can improve the product for our users. 
+
+### Goals
+
+1. Create a process to collect and send [event](https://github.com/openshift/assisted-service/tree/master/docs/dev/events.md) data from on-premises deployments of AI.
+2. Analyze and process on-premises data (by use of visualizations, dashboards, etc.) at Red Hat like SaaS data.
```
gamli75, line 22: all the data will be stored in the same ELK instance so we should be able to differentiate between SaaS and each on-prem environment

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+    - `DISABLE_DATA_COLLECTION`: If set to `true`, it will disable this feature and thereby allow a customer to opt-out, otherwise the feature is on by default.
+    - `DATA_COLLECTION_URL`: The URL the data will be sent to.
+2. Add additional logic when a cluster reaches one of the following end stages: `error`, `cancelled`, `installed`
+    - Each of these cluster stages represent an end state and that the cluster attempted installation.
+    - Additional logic includes: querying the DB for events of that cluster and streaming this event data to the `DATA_COLLECTION_URL` over HTTPS.
+        - If the customer has large-scale clusters, would it be better to always compress the data in a separate thread?
+3. Add logic for authenticating to the `DATA_COLLECTION_URL` 
+    - Currently unknown since the service to host this data hasn't been decided on yet, but it could take the form of using a Kubernetes Secret
+
+### Risks and Mitigations
```
gamli75, line 79: what will happen if we will change the cluster/ host/ event data structure and the on-prem environment will contain old code? we need to think about backward support in our data flow

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+4. Collecting must-gather data (and logs) for debugging failed cluster installation is outside the scope of this epic. This epic is only concerned with [event](https://github.com/openshift/assisted-service/blob/6b8401d3b0f0ceea5a837f33c17b9b1cd5ec52a3/internal/events/api/event.go) data.
+
+
+## Proposal
+
+A new flow for sending data is to be added:
+When a cluster installation attempt finishes (regardless of success or failure), all of the event data for this cluster will be queried from the DB, packaged, and sent to the data hosting endpoint.
+
+Changes that need to be made in the `Assisted-Service` to support this:
+- Needs to know what the data hosting service endpoint is and how to authenticate into it
```
romfreiman, line 39: can be also local s3 / elk, right?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Story 3
+
+As a product owner, I can determine if the product is moving in the right direction based on the actual customer data. I can help determine which features and bug fixes to prioritize based on the data.
+
+### Implementation Details/Notes/Constraints
+
+Assisted Service changes:
+1. New environment variables to be added:
+    - `DISABLE_DATA_COLLECTION`: If set to `true`, it will disable this feature and thereby allow a customer to opt-out, otherwise the feature is on by default.
```
romfreiman, line 70: or we can piggy back on the Telemetry one?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+    - `DATA_COLLECTION_URL`: The URL the data will be sent to.
+2. Add additional logic when a cluster reaches one of the following end stages: `error`, `cancelled`, `installed`
+    - Each of these cluster stages represent an end state and that the cluster attempted installation.
+    - Additional logic includes: querying the DB for events of that cluster and streaming this event data to the `DATA_COLLECTION_URL` over HTTPS.
```
romfreiman, line 74: why? dont we have the donwload logs api that already collect the events?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+1. Implementing new data types to collect is out of the scope of this epic. This epic defines how and when the data already collected will be processed and sent.
+2. Define how SaaS should redesign their data pipeline (taken care of in [MGMT-11244](https://issues.redhat.com/browse/MGMT-11244)) is out of the scope of this epic. This epic only refers to on-premises deployments of AI.
+3. Define the creation and/or management of an event streaming instance such as Kafka and the endpoint this data will be hosted at (both taken care of in [MGMT-11244](https://issues.redhat.com/browse/MGMT-11244)).
+4. Collecting must-gather data (and logs) for debugging failed cluster installation is outside the scope of this epic. This epic is only concerned with [event](https://github.com/openshift/assisted-service/blob/6b8401d3b0f0ceea5a837f33c17b9b1cd5ec52a3/internal/events/api/event.go) data.
+
+
+## Proposal
+
+A new flow for sending data is to be added:
+When a cluster installation attempt finishes (regardless of success or failure), all of the event data for this cluster will be queried from the DB, packaged, and sent to the data hosting endpoint.
```
rccrdpccl, line 36: We will loose "deserted installations" insights. Is this acceptable?

docs/enhancements/on-prem-data-collection.md
rccrdpccl: It's a good point, but I think data format issues should be handled by services that will process the data, I think we can ignore shcema when sending it

docs/enhancements/on-prem-data-collection.md
CrystalChun: Good point! @rccrdpccl or @tsorya do you happen to know if there's a way to check for duplicate data in either the ingress (pipelines) or the kafka? Like if the cluster id & user email already exists?

---

If not then either a task to investigate this or an alternative solution needs to be added. 

---
An option for the alternative solution could be to include a new data field for each event to show whether or not it's been sent to streaming. The field could also only be filled if it's an on-prem install.

A problem could be that the field marks the data as sent but the connection or packets get dropped somewhere and never make it to the external service.

docs/enhancements/on-prem-data-collection.md
CrystalChun: That's a good idea! Perhaps the retry mechanism could be to retry X times after Y minutes has passed.
E.g. Retry sending 5 more times every 10 minutes until success or exceeded number of retries.

What do you think?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Sorry, could you clarify "deserted installations"? Does that mean that the customer created a cluster, but never starts trying to install it?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Goals
+
+1. Create a process to collect and send [event](https://github.com/openshift/assisted-service/tree/master/docs/dev/events.md) data from on-premises deployments of AI.
+2. Analyze and process on-premises data (by use of visualizations, dashboards, etc.) at Red Hat like SaaS data.
+
+
+### Non-Goals
+
+1. Implementing new data types to collect is out of the scope of this epic. This epic defines how and when the data already collected will be processed and sent.
```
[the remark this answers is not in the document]
CrystalChun: Thanks for bringing up this point! I'm not so sure data formatting should be handled in this epic, but it would definitely affect this epic. 

Is there already a discussion around this somewhere? Do you happen to have a link?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Thanks Rom! Wasn't aware there was a preexisting configuration. May I have a pointer to it? Or maybe @rccrdpccl or @gamli75 knows?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Good point! Added to the test plan - how does it sound?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Good point, added clarification! Does it sound ok?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Good point and agreed with Riccardo. I'll add that to the non-goals section

docs/enhancements/on-prem-data-collection.md
CrystalChun: I think events are collected, but I wasn't aware that they were automatically pushed. @rccrdpccl Do you know if events automatically pushed to ELK for SaaS now? Do you mind sharing a pointer to where that happens?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Thought it would need to be a remote/cloud s3/elk for us to be able to pull it? @rccrdpccl @tsorya do you know?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Good to know! May I have a pointer to how it's hashed right now in event-scrape?

docs/enhancements/on-prem-data-collection.md
rccrdpccl: We can handle duplication from the data pipeline (we should expose event_id though, that currently it isn't).
The main issue in my opinion is avoiding DDoS the http insights ingress service

docs/enhancements/on-prem-data-collection.md
rccrdpccl: https://github.com/openshift-assisted/assisted-events-scrape/blob/master/assisted-events-scrape/utils/anonymizer.py#L25

If we implement it before sending we can then remove it from the scraper (IMO it would be better doing it this way). In such case we'd need to make a plan to roll out without double-encoding or anything like that. WDYT @CrystalChun ?

docs/enhancements/on-prem-data-collection.md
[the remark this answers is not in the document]
rccrdpccl: No official discussion yet I am afraid... We can start one of course :)

docs/enhancements/on-prem-data-collection.md
rccrdpccl: Yes, something along these lines.
Our funnel goes like this:
1) Cluster creation
2) Adding host
3) Installing cluster
4) Installed

With the current proposal we are only covering point 4, but all the others will be uncovered.
By "deserted installations" we mean clusters that reach stage 2 but never stage 3, but I've just realized that we are also missing other cases.

docs/enhancements/on-prem-data-collection.md
rccrdpccl: Events are not pushed from the service to ELK (yet), but I believe the download api is leveraged (also) for TRIAGE tickets. I am afraid I don't know where it exactly happens.

docs/enhancements/on-prem-data-collection.md
celebdor: I was thinking of having a local target that allowed you to specify a directory visible to assisted service, where both events and logs would be kept.

Then, on podman assisted service (like _openshift-install agent_ installations), we'd mount some location to the pod. Continuing with the _openshift-install agent_ case, we'd then, after writing the image to disk in node0, we'd mount node0 and writing it there, so it can be retrieved after installation completes.

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,143 @@
...
+
+Changes that need to be made in the `Assisted-Service` to support this:
+- Needs to know what the data hosting service endpoint is and how to authenticate into it
+- Needs code changes when an install is attempted to query data
+- Needs to know how to format the data to send to the data hosting service
+- Needs to expose an option for users to opt out of this feature
+
+Additional changes needed:
+- Documentation regarding terms and conditions?
+- Documentation about how to opt out
```
celebdor, line 48: Isn't it just that if configuration for a destination is missing, it is opted out?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,143 @@
...
+When a cluster installation attempt finishes (regardless of success or failure), all of the event data for this cluster will be queried from the DB, packaged, and sent to the data hosting endpoint.
+
+Changes that need to be made in the `Assisted-Service` to support this:
+- Needs to know what the data hosting service endpoint is and how to authenticate into it
+- Needs code changes when an install is attempted to query data
+- Needs to know how to format the data to send to the data hosting service
+- Needs to expose an option for users to opt out of this feature
+
+Additional changes needed:
+- Documentation regarding terms and conditions?
```
celebdor, line 47: IIRC OpenShift terms and conditions already include data gathering, we should just check if it covers this part

celebdor (commented):
Added some perspective into the local collection for disconnected environments

docs/enhancements/on-prem-data-collection.md
CrystalChun: Oh I see, thanks for clarifying that!

Some ideas to prevent DDoS-ing is maybe:

1. Only sending the events when the cluster has reached installation or failure to install stages
2. Having a max retry for sending the events to the insights ingress service
3. Not retrying until a timeout has been reached and there's no response from the insights ingress service

What do you think? Are there better ways to prevent this?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Thanks for the link! Sounds like a good idea to implement the encoding before sending! Good point to make a plan to roll out without double-encoding.

docs/enhancements/on-prem-data-collection.md
CrystalChun: Thanks! Right, the download api is also used for TRIAGE tickets. For the additional logic mentioned here is referring to pushing this event data to the url when the cluster reaches an end stage.

docs/enhancements/on-prem-data-collection.md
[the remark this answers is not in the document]
CrystalChun: Good to know! Where would you recommend we start discussing this?

docs/enhancements/on-prem-data-collection.md
CrystalChun: I see what you mean. Do we currently collect data at all stages mentioned in SaaS now? Also would the data that hasn't surpassed stage 3 be relevant or interesting?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Great to know, thanks Toni!

docs/enhancements/on-prem-data-collection.md
CrystalChun: Interesting, I think I need some clarification. The local directory/mount would be an additional change to Assisted Service, but the service would still need to know the endpoint of the data hosting service, right?

docs/enhancements/on-prem-data-collection.md
rccrdpccl: Yes, at the moment in SaaS we collect all events (not really but that's the intention).
We are currently analyzing "deserted installations" (Reached stage 2 but not stage 3): these users are users that did interact with our application (not just creating a cluster, but bothered adding a host), but for some reason could not reach the installation stage.
If we can detect what are the reasons, in some cases we could help the users go through these hurdles (by enhancing or adding features or simply improving the UI).

I wouldn't say it's critical but it sure helps.

docs/enhancements/on-prem-data-collection.md
rccrdpccl: I think they are good points (point 1 is also discussed [here](https://github.com/openshift/assisted-service/pull/4457#discussion_r985767135)).
I would also consider retrying with exponential backoff and jitter, to prevent all clients retrying at the same time in case of downtime of our service and using remote configuration for these settings and possibly the service URL and circuit breaking, this way we have some control over client applications (i.e. possibility to configure `DATA_COLLECTION_URL` through a remote manifest  `https://example.com/data_collection.json` that the application would read first - this can be cached in a CDN or something so it can be retrieved many times). The advantage of doing this is that 1) we can send traffic to /dev/null 2) if we still want to try retrieve the data we'd have a few knobs to facilitate it.

CrystalChun: Great ideas! Agreed with exponential backoff and jitter.

For the remote manifest, how often should the assisted service try to read it? I would think:
1. When the assisted service initially starts
2. Every time the assisted service will send data it'll get this configuration first

If the manifest is unreachable, then we can default to the last configuration we got or the default if it was never able to reach it.

> The advantage of doing this is that 1) we can send traffic to /dev/null

By this, do you mean we can prevent DDoS-ing the http insights ingress service by directing `DATA_COLLECTION_URL` to `/dev/null`?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Gotcha, that makes sense! Thanks for explaining!

Trying to think of a better flow to capture this data without overloading the ingress http server, any ideas? If the customer reaches stage 2 but not stage 3, would they delete the cluster or start removing hosts? If so we could probably trigger a data send when they delete/remove?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Do you mean a procedure similar to the OCP health monitoring one where removing a JSON entry from the pull secret will opt the user out? https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html-single/support/index#insights-operator-new-pull-secret_opting-out-remote-health-reporting

docs/enhancements/on-prem-data-collection.md
rccrdpccl: > For the remote manifest, how often should the assisted service try to read it? I would think:
> 
> 1. When the assisted service initially starts
> 2. Every time the assisted service will send data it'll get this configuration first
>
> If the manifest is unreachable, then we can default to the last configuration we got or the default if it was never able to reach it.

Sounds good to me (both when to read it and behaviour when it's unreachable)

> By this, do you mean we can prevent DDoS-ing the http insights ingress service by directing `DATA_COLLECTION_URL` to `/dev/null`?

Yes, really by `/dev/null` I mean any URL that wouldn't overload the main one.
A secondary effect is that if we have a change in the ingress URL for any reason, we can update the URL without breaking older clients.

docs/enhancements/on-prem-data-collection.md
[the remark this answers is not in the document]
rccrdpccl: I have started a rough document with what I have in mind, maybe we can start the discussion there and see where it goes?  Should it eventually land in this very enhancement or a specific one? https://docs.google.com/document/d/1MUtjnhhGxKvY1pjnOV4caFR6FnO_5_qjrAHDhJs-Nrw

docs/enhancements/on-prem-data-collection.md
rccrdpccl: > Trying to think of a better flow to capture this data without overloading the ingress http server, any ideas? If the customer reaches stage 2 but not stage 3, would they delete the cluster or start removing hosts? If so we could probably trigger a data send when they delete/remove?

In some cases, yes. I think that a kind of shutdown trigger that would send all collected that it's a very good idea, although it won't work in the "unplugging the machine" situation but at least we cover a lot of cases.

Maybe we can also have a periodic check (long interval, maybe daily or something?) and if the cluster has not changed state (meaning the installation is abandoned/forgotten), just send the data we have until that point? WDYT?

docs/enhancements/on-prem-data-collection.md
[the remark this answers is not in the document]
CrystalChun: Awesome, thanks Riccardo! Maybe it could be its own? 

It's related to this epic, but I think restructuring the data might be its own effort. WDYT?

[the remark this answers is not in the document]
rccrdpccl: Sounds good to me!

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,197 @@
...
+
+Sending data should not interfere with customer workloads and deployment
+
+### Assisted-Service Changes 
+
+Changes that need to be made in the `Assisted-Service` to support this:
+- Needs to know what the data hosting service endpoint is and how to authenticate into it
+- Needs code changes when an install is attempted to query and send data (data sending flow event 1 above)
+- Needs code changes when a cluster is deleted or all hosts are removed from a cluster to query and send data (data sending flow event 2 above)
+- Needs code changes to periodically check if a cluster state hasn't changed in a given amount of time to query and send data (data sending flow event 3 above)
```
rccrdpccl, line 56: Maybe it's already covered, but should we also send all data when the API server is shutting down?

CrystalChun: Good point! Modified event flow 2 to include this as well!

docs/enhancements/on-prem-data-collection.md
filanov: isn't that just a different implementation of the client that is going to be used? 
for example send events is the interface and the implementation details can be file/s3/elk etc.
as long as the interface is clear there should be no difference no?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,215 @@
...
+
+Before the data is sent, Assisted Service will get a remote configuration from an endpoint we specify.
+
+Sending data should not interfere with customer workloads and deployment
+
+### Assisted-Service Changes 
+
+Changes that need to be made in the `Assisted-Service` to support this:
+- Needs to know what the data hosting service endpoint is and how to authenticate into it
+- Needs code changes when an install is attempted to query and send data (data sending flow event 1 above)
```
filanov, line 54: can't we create a wrapper to the exiting events?

docs/enhancements/on-prem-data-collection.md
filanov: suggesting to start with best effort first, retry can be an improvement and it cannot be a blocking operation.
meaning if we failed to send any data we cannot block the operation or the service because of it. 
As Crystal mentioned in `Risk 1: Overusing Customer's Network & CPU`

docs/enhancements/on-prem-data-collection.md
filanov: is it required if we have `DATA_COLLECTION_URL`? 
if `DATA_COLLECTION_URL` not set then it's disabled.

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,215 @@
...
+##### Mitigation
+
+A field can be used as a marker to indicate the data has been sent to the event streaming service or that the data already exists. One option is to use the timestamp, cluster ID, and user email to detect duplicate events.
+
+
+## Design Details
+
+### Open Questions
+
+1. Are we planning to support disconnected environments in the future?
```
filanov, line 172: @romfreiman correct me if i'm wrong but the plan was to have info from telcos as well no?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Agreed! Modified the doc to include this. 

> retry can be an improvement

Do you mean retry can be implemented later and not as part of this first pass at the design?

filanov: implemented later

docs/enhancements/on-prem-data-collection.md
CrystalChun: That makes sense, thank you for clarifying @filanov! FYI @rccrdpccl it looks like this is also targeted for disconnected environments

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,216 @@
...
+
+
+## Proposal
+
+### Data Sending Flow
+
+All of the event data for this cluster will be queried from the DB, packaged, and sent to the data hosting endpoint.
+
+Data will be sent when one of the following happens:
+1. A cluster installation attempt finishes (regardless of success or failure)
```
romfreiman, line 42: can we start with this one? i think the next 2 can be future enhancements

docs/enhancements/on-prem-data-collection.md
romfreiman: @celebdor how do envision it in case of billi - what will send it post installation? And even worth, what will send it in case of the installation failure - because this is when we'll need this data, but there wont be any control plane.

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,216 @@
...
+Before the data is sent, Assisted Service will get a remote configuration from an endpoint we specify.
+
+Sending data should not interfere with customer workloads and deployment
+
+### Assisted-Service Changes 
+
+Changes that need to be made in the `Assisted-Service` to support this:
+- Needs to know what the data hosting service endpoint is and how to authenticate into it
+- Needs code changes when an install is attempted to query and send data (data sending flow event 1 above)
+- Needs code changes when a cluster is deleted, all hosts are removed from a cluster, and the API server is shutting down to query and send data (data sending flow event 2 above)
```
romfreiman, line 55: i would not bother with this at this point. same about the next bullet. we can just decide that success / fail / pending user action are finite states.

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,216 @@
...
+- Host a remote configuration (using a CDN)
+
+### Remote Configuration
+
+The remote configuration will specify configuration data for the assisted-service such as `DATA_COLLECTION_URL`. This can be cached in a CDN.
+
+The assisted service will get this remote configuration:
+
+1. When the assisted service initially starts
+2. Every time the assisted service will send data it'll get this configuration first
```
romfreiman, line 77: why? how often it will be changed?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Implementation Details/Notes/Constraints
+
+Assisted Service changes:
+1. New environment variables to be added:
+    - `DISABLE_DATA_COLLECTION`: If set to `true`, it will disable this feature and thereby allow a customer to opt-out, otherwise the feature is on by default.
+    - `DATA_COLLECTION_URL`: The URL the data will be sent to.
+    - `REMOTE_CONFIG_URL`: The URL for the manifest with configuration data for on-prem data collection variables such as `DATA_COLLECTION_URL`
+2. Add data collection and sending logic
+    - Collecting and sending: querying the DB for events of that cluster, sanitizing the data, and streaming this event data to the `DATA_COLLECTION_URL` over HTTPS.
```
romfreiman, line 116: wdym streaming? even by event? tar.gz?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Sorry, may I get some clarification on this? Do you mean creating a wrapper on top of the events package? https://github.com/openshift/assisted-service/blob/master/internal/events/event.go

docs/enhancements/on-prem-data-collection.md
CrystalChun: Do you think it would be a good idea to add a `version` field to the data? It could be the assisted-service version. That way when it reaches the service that's processing the data, it'll know what fields to expect?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Sorry, wording is off, meant sending the data as a tar.gz
Modified the doc to reflect this

docs/enhancements/on-prem-data-collection.md
CrystalChun: Gotcha, thanks! Modified the doc to mention retry as a future enhancement

docs/enhancements/on-prem-data-collection.md
CrystalChun: Yes! Modified the doc to reflect this. Thanks!

docs/enhancements/on-prem-data-collection.md
CrystalChun: FYI Rom suggested for these to be future enhancements

docs/enhancements/on-prem-data-collection.md
slaviered: Isn't the data collection going to be independently managed from the install command trigger? A failure or retries can slow down the start of the installation itself. Also, what happens to the post install events?

docs/enhancements/on-prem-data-collection.md
slaviered: @celebdor isn't opting out done in "real time" (i.e. the user can at any time opt out)?

docs/enhancements/on-prem-data-collection.md
slaviered: Per the 3rd customer story: Are you discussing existing events or new ones with extra data? Because for our ELK the data in the event is processed first in order to provide a meaningful insight.

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Risks and Mitigations
+
+#### Risk 1: Overusing Customer's Network & CPU
+
+On-premises users might have large-scale clusters so network usage and processor usage could be impacted depending on how large the data set is.
+
+##### Mitigation
+
+To avoid user workflow interruption, data collection, processing, and sending can happen as a concurrent thread. Data sending should be non-blocking and best effort.
```
slaviered, line 126: Still the concurrent thread consumes resources out of the container's process. Perhaps we should do a POC on what scales are involved?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,220 @@
...
+## Proposal
+
+### Data Sending Flow
+
+All of the event data for this cluster will be queried from the DB, packaged, and sent to the data hosting endpoint.
+
+Data will be sent when:
+1. A cluster installation attempt finishes (regardless of success or failure)
+
+Before the data is sent, Assisted Service will get a remote configuration from an endpoint we specify.
```
carbonin, line 44: What do you mean by "get" here?
Isn't this something that would be baked into the deployment/operator?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,220 @@
...
+  - "@CrystalChun"
+creation-date: 2022-09-30
+last-updated: 2022-10-28
+---
+
+# On-premises Data Collection
+
+## Summary
+
+The Assisted Installer ("AI") will provide a process to collect data about installations when it is deployed on-premises as part of either [Red Hat Advanced Cluster Management for Kubernetes](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.6) ("RHACM") or as a stand-alone Podman deployment. Data was already collected from clusters deployed on [Red Hat's Hybrid Cloud Console](https://console.redhat.com/openshift/assisted-installer/clusters) (“SaaS”), but there was no method for obtaining data from on-premises deployments.
```
carbonin, line 13: The epic mentions BILLI, are we considering that here as well?

docs/enhancements/on-prem-data-collection.md
carbonin: This can be a configuration in the operator I'd imagine (or just a flag at the deployment level).
Not sure why it would need to be more complicated than that.

docs/enhancements/on-prem-data-collection.md
carbonin: IIUC I don't think this is necessary.
Mostly because I don't see this URL changing any more frequently than whatever mechanism you are planning to use to retrieve this URL.

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,220 @@
...
+1. Future-proof this feature
+    - The configuration is not reliant on the version of the assisted-service deployed
+    - E.g. if the data collection endpoint changes, older versions will be able to send it to the new endpoint without having to update their client
+2. Allows flexibility for our team
+    - We can adjust the configurations on the fly such as where data is sent to
+    - E.g. if the original data collection endpoint is unavailable or if we choose to use a different service later on, we only need to change the configuration endpoint 
+
+#### Drawbacks
+1. Another deployment artifact to update and maintain
+2. Doesn't fully future-proof against the remote configuration's URL changes
```
carbonin, line 86: This is my point above. I don't think it makes sense to solve this for the real endpoint when we'd have the same "problem" for the meta-endpoint.

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,220 @@
...
+The remote configuration will specify configuration data for the assisted-service such as `DATA_COLLECTION_URL`. This can be cached in a CDN.
+
+The assisted service will get this remote configuration:
+
+1. When the assisted service initially starts
+2. Every time the assisted service will send data it'll get this configuration first
+
+If the assisted service can't get the latest configuration, it'll fallback to the last configuration received or the hard-coded default it was deployed with.
+
+#### Benefits
```
carbonin, line 76: This has me thinking that maybe we want a way to control the granularity of the data we send?
It can be the same endpoint but maybe we want a way to tell the service that we don't need host inventory or something if we end up sending too much. That's just an example, but hopefully it illustrates the idea.

docs/enhancements/on-prem-data-collection.md
carbonin: Minor concern, but if we do keep this variable I think it's better to name things in the positive (i.e. make this `ENABLE_DATA_COLLECTION` defaulting to `true`)

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,220 @@
...
+As a product owner, I can determine if the product is moving in the right direction based on the actual customer data. I can help determine which features and bug fixes to prioritize based on the data.
+
+### Implementation Details/Notes/Constraints
+
+Assisted Service changes:
+1. New environment variables to be added:
+    - `DISABLE_DATA_COLLECTION`: If set to `true`, it will disable this feature and thereby allow a customer to opt-out, otherwise the feature is on by default.
+    - `DATA_COLLECTION_URL`: The URL the data will be sent to. (Data format env var?)
+    - `REMOTE_CONFIG_URL`: The URL for the manifest with configuration data for on-prem data collection variables such as `DATA_COLLECTION_URL`
+2. Add data collection and sending logic
```
carbonin, line 110: What process does this?

Is this going to be just some kind of fire-and-forget goroutine that gets spun off after we transition to an end state?
Or is it going to be another persistent "worker"-type thread that is constantly running, waiting for these events?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,220 @@
...
+
+##### Mitigation
+
+A field can be used as a marker to indicate the data has been sent to the event streaming service or that the data already exists. One option is to use the timestamp, cluster ID, and user email to detect duplicate events.
+
+## Design Details
+
+### Open Questions
+
+1. Is the remote configuration a good idea? A good amount of customers using this feature are Telcos and they would be disconnected
```
carbonin, line 164: In disconnected we're not going to get this data anyway right? Not sure how that relates to the remote config.

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,220 @@
...
+2. What is the Telemetry opt-out? Is it the pull secret not containing the auth for it? Would this be relevant for disconnected? Wouldn't it be better to use a specific opt-out variable to not run this service of pushing data to a server (local or remote)
+3. How would the configuration for writing to a local s3 storage work? Would it be something we define and tell the customer to set up? Would it be something the customer defines and set up? Ideally would the customer need to be involved at all? Wouldn't there be a big space requirement for this?
```
carbonin, line 166: Why are we ever writing to customer-local s3? How does that help us get this data?

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,220 @@
...
+#### Test 2
+
+Check against a large-scale cluster with > 100 hosts to verify:
+1. All of the cluster's events are being stored, queried, and sent (within a reasonable time)
+2. That the event streaming can receive and process the large data set
+
+## Drawbacks
+
+- Dependency on the data hosting service selected in [MGMT-11244](https://issues.redhat.com/browse/MGMT-11244) could cause delays in this epic being completed.
+- Disconnected environment users would still need to send data to us
```
carbonin, line 188: I don't think this is worth doing. I feel like it's incredibly unlikely that a customer goes out of their way to send us telemetry manually that isn't going to benefit them directly (solving a support issue or something).

If this involves no extra work then I guess it's fine to develop some kind of process for doing this, but I don't think we should design our code changes around this being something that might actually happen.

carbonin (commented):
I don't see much about auth or communication with the remote service here.

How are we authenticating to the service we're sending this data to?
What kind of protocol is this remote service using?

docs/enhancements/on-prem-data-collection.md
CrystalChun: I think it would include both existing and new events (all data from SaaS and the on-premises deployments)

docs/enhancements/on-prem-data-collection.md
CrystalChun: Discussed during the meeting with the conclusion that we'll expose an environment variable to opt-out that the user may toggle, but we'll also check if they opted-out of telemetry and respect that setting.

docs/enhancements/on-prem-data-collection.md
CrystalChun: Thanks for the perspective! Modified the doc to move remote configuration to a potential future enhancement.

docs/enhancements/on-prem-data-collection.md
CrystalChun: Good question, @celebdor @filanov is BILLI part of the scope of this epic?

docs/enhancements/on-prem-data-collection.md
CrystalChun: Agreed, but using it as a general configuration with other toggles (as you mentioned in another comment) sounds useful!

docs/enhancements/on-prem-data-collection.md
CrystalChun: Great idea! Added this to the future enhancement as an idea of what the remote configuration could contain!

docs/enhancements/on-prem-data-collection.md
CrystalChun: Good point! Thanks Nick! Modified the doc to reflect this.

docs/enhancements/on-prem-data-collection.md
CrystalChun: I think it'll be more similar to the fire-and-forget goroutine. 

I'm thinking of creating an event_uploader controller that has an upload function. 

So the flow would go something like this:
1. A cluster fails/finishes installation 
2. There's a separate function call to query the DB for events of that cluster
3. Then the event_uploader controller would be called to upload the events (including sanitization and packaging of data)


What do you think?

docs/enhancements/on-prem-data-collection.md
CrystalChun: That's right! The meeting helped clarified the end goal so this is not relevant at all. Thanks!

docs/enhancements/on-prem-data-collection.md
CrystalChun: That's right! The meeting helped clarified the end goal so this is not relevant at all. Thanks!

docs/enhancements/on-prem-data-collection.md
CrystalChun: That's right! The meeting helped clarified the end goal so this is not relevant at all. Thanks!

docs/enhancements/on-prem-data-collection.md
CrystalChun: That's right! Decided during meeting that remote configuration would not be part of the first iteration. Thanks!

docs/enhancements/on-prem-data-collection.md
filanov: something like https://github.com/openshift/assisted-service/blob/d7eb5d6c0857a0a20ef5017c7e2594805cd57b17/internal/controller/controllers/controller_event_wrapper.go

docs/enhancements/on-prem-data-collection.md
filanov: Yes

docs/enhancements/on-prem-data-collection.md
CrystalChun: I've added the remote manifest and retrying as a future enhancement

oourfali:
perhaps I missed it, but let's make sure that failure to send the data doesn't result in any failure of any kind, as sending this data is "best-effort" rather than "mandatory".

docs/enhancements/on-prem-data-collection.md
```diff
@@ -0,0 +1,221 @@
...
+#### Drawbacks
+
+- Likely will not get data
+    - Due to the customer having to do extra work to upload the data
+- Data would be skewed since we'd only get data when the customer uploads it
+    - Customer is most likely only going to upload when they have an issue
+
+---
+
+## Future Enancements
```
carbonin, line 172: I think it might be better for this to just be different epics in JIRA rather than a section in this doc.

docs/enhancements/on-prem-data-collection.md
carbonin: Did we agree that the only scope of this enhancement is sending data specifically to the SaaS ingress endpoint?

I don't see value in collecting the event data somewhere locally even if we wanted to deploy some separate component to store it it.

CrystalChun: Agreeing with Nick that there isn't much value collecting the event data locally.

docs/enhancements/on-prem-data-collection.md
CrystalChun: Thanks Nick! Removing this and creating new epics in JIRA

oourfali:
> 

Lgtm

carbonin (approved):

PR #4459, opened 2022-10-03
MGMT-11970: Add a validation to check for IP collissions.

paul-maidment:
Presently we do not have a way to report on IP collissions.

In conjunction with the changes in https://github.com/openshift/assisted-installer-agent/pull/433 this commit will add IP collission detection to assisted service.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
I will make sure to deploy assisted-installer in beaker and to use this version of the agent https://github.com/openshift/assisted-installer-agent/pull/433
I will then deliberately create a cluster that has IP collisions and check to see how these are reported.
- [x] No tests needed

## Checklist

- x] Title and description added to both, commit and PR.

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -1133,6 +1133,11 @@ func (b *bareMetalInventory) refreshAllHostsOnInstall(ctx context.Context, clust
 	if err != nil {
 		return err
 	}
+	err = b.detectAndStoreCollidingIPsForCluster(cluster.ID, b.db)
+	if err != nil {
+		b.log.WithError(err).Errorf("Failed to detect and store colliding IPs for cluster %s", cluster.ID.String())
+		return err
+	}
 	for _, chost := range cluster.Hosts {
 		if swag.StringValue(chost.Status) != models.HostStatusKnown && swag.StringValue(chost.Kind) == models.HostKindHost {
 			return common.NewApiError(http.StatusBadRequest, errors.Errorf("Host %s is in status %s and not ready for install",
@@ -1665,6 +1670,12 @@ func (b *bareMetalInventory) refreshClusterHosts(ctx context.Context, cluster *c
 	err := b.setMajorityGroupForCluster(cluster.ID, tx)
 	if err != nil {
 		log.WithError(err).Errorf("Failed to set cluster %s majority groups", cluster.ID.String())
+		return common.NewApiError(http.StatusInternalServerError, err)
+	}
+
+	err = b.detectAndStoreCollidingIPsForCluster(cluster.ID, tx)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to detect and store colliding IPs for cluster %s", cluster.ID.String())
 		return common.NewApiError(http.StatusInternalServerError, err)
 	}
 
@@ -3404,6 +3415,10 @@ func filterReply(expected interface{}, input string) (string, error) {
 
 func (b *bareMetalInventory) setMajorityGroupForCluster(clusterID *strfmt.UUID, db *gorm.DB) error {
 	return b.clusterApi.SetConnectivityMajorityGroupsForCluster(*clusterID, db)
+}
+
+func (b *bareMetalInventory) detectAndStoreCollidingIPsForCluster(clusterID *strfmt.UUID, db *gorm.DB) error {
+	return b.clusterApi.DetectAndStoreCollidingIPsForCluster(*clusterID, db)
 }
 
 func (b *bareMetalInventory) refreshClusterStatus(
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -111,6 +111,7 @@ type API interface {
 	PrepareClusterLogFile(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) (string, error)
 	SetUploadControllerLogsAt(ctx context.Context, c *common.Cluster, db *gorm.DB) error
 	SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID, db *gorm.DB) error
+	DetectAndStoreCollidingIPsForCluster(clusterID strfmt.UUID, db *gorm.DB) error
 	DeleteClusterLogs(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error
 	DeleteClusterFiles(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error
 	UpdateLogsProgress(ctx context.Context, c *common.Cluster, progress string) error
@@ -571,6 +572,10 @@ func (m *Manager) ClusterMonitoring() {
 				_ = m.autoAssignMachineNetworkCidr(cluster)
 				if err = m.setConnectivityMajorityGroupsForClusterInternal(cluster, m.db); err != nil {
 					log.WithError(err).Error("failed to set majority group for clusters")
+				}
+				err = m.detectAndStoreCollidingIPsForCluster(cluster, m.db)
+				if err != nil {
+					m.log.WithError(err).Errorf("Failed to detect and store colliding IPs for cluster %s", cluster.ID.String())
 				}
 				clusterAfterRefresh, err = m.refreshStatusInternal(ctx, cluster, m.db)
 				if err != nil {
@@ -1044,6 +1049,94 @@ func (m *Manager) IsReadyForInstallation(c *common.Cluster) (bool, string) {
 	return true, ""
 }
 
+func (m *Manager) detectAndStoreCollidingIPsForCluster(cluster *common.Cluster, db *gorm.DB) error {
+	if db == nil {
+		db = m.db
+	}
+	// We want to calculate ip collisions only when in pre-install states since it is needed for pre-install validations
+	allowedStates := []string{
+		models.ClusterStatusPendingForInput,
+		models.ClusterStatusInsufficient,
+		models.ClusterStatusReady,
+	}
+	if !funk.ContainsString(allowedStates, swag.StringValue(cluster.Status)) {
+		return nil
+	}
+
+	hosts := make([]*models.Host, len(cluster.Hosts))
+	copy(hosts, cluster.Hosts)
+	sort.Slice(hosts, func(i, j int) bool {
+		return hosts[i].ID.String() < hosts[j].ID.String()
+	})
+	// Note on ipCollisions map structure:
+	// The key is a remote IP for which collision has been detected.
+	// The map value is an array of macs found to be involved in the collision.
+	ipCollisions := make(map[string][]string)
+	collidingIPSWithMacs := make(map[string][]string)
+	for _, host := range hosts {
+		if len(host.Connectivity) > 0 {
+			connectivityReport, err := hostutil.UnmarshalConnectivityReport(host.Connectivity)
+			if err != nil {
+				// Let's not stop iterating over hosts but let's log this error.
+				m.log.WithError(err).Errorf("unable to unmarshall connectivity report for host %d due to error: %s", host.ID, err.Error())
+			} else {
+				collidingIPSWithMacs = getCollidingIPs(connectivityReport)
+				for k := range collidingIPSWithMacs {
+					ipCollisions[k] = collidingIPSWithMacs[k]
+				}
+			}
+
+		}
+	}
+
+	b, err := json.Marshal(&collidingIPSWithMacs)
+	if err != nil {
+		return common.NewApiError(http.StatusInternalServerError, err)
+	}
+
+	marshalledCollidingIPSWithMacs := string(b)
+	if marshalledCollidingIPSWithMacs != cluster.IPCollisions {
+		err = db.Model(&common.Cluster{}).Where("id = ?", cluster.ID.String()).Updates(&common.Cluster{
+			Cluster: models.Cluster{
+				IPCollisions: marshalledCollidingIPSWithMacs,
+			},
+			TriggerMonitorTimestamp: time.Now(),
+		}).Error
+		if err != nil {
+			return common.NewApiError(http.StatusInternalServerError, err)
+		}
+	}
+	return nil
+}
+
+func getCollidingIPs(connectivityReport *models.ConnectivityReport) map[string][]string {
+	collidingIPSWithMacs := make(map[string][]string)
+	collisionHistory := make(map[string]map[string]string)
+	for _, remoteHost := range connectivityReport.RemoteHosts {
+		for _, c := range remoteHost.L2Connectivity {
+			if collisionHistory[c.OutgoingNic] != nil {
+				if previousMac, ok := collisionHistory[c.OutgoingNic][c.RemoteIPAddress]; ok {
+					if previousMac != "" && previousMac != c.RemoteMac {
+						// Collision detected.
+						if collidingIPSWithMacs[c.RemoteIPAddress] == nil {
+							collidingIPSWithMacs[c.RemoteIPAddress] = []string{}
+						}
+						// For cache reasons, make sure that macs are in the same order every time
+						macs := []string{previousMac, c.RemoteMac}
+						sort.Strings(macs)
+						collidingIPSWithMacs[c.RemoteIPAddress] = macs
+					}
+				}
+			}
+			if collisionHistory[c.OutgoingNic] == nil {
+				collisionHistory[c.OutgoingNic] = make(map[string]string)
+			}
+			collisionHistory[c.OutgoingNic][c.RemoteIPAddress] = c.RemoteMac
+		}
+	}
+	return collidingIPSWithMacs
+}
+
 func (m *Manager) setConnectivityMajorityGroupsForClusterInternal(cluster *common.Cluster, db *gorm.DB) error {
 	if db == nil {
 		db = m.db
@@ -1101,6 +1194,18 @@ func (m *Manager) setConnectivityMajorityGroupsForClusterInternal(cluster *commo
 		}
 	}
 	return nil
+}
+
+func (m *Manager) DetectAndStoreCollidingIPsForCluster(clusterID strfmt.UUID, db *gorm.DB) error {
+	cluster, err := common.GetClusterFromDBWithHosts(db, clusterID)
+	if err != nil {
+		var statusCode int32 = http.StatusInternalServerError
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			statusCode = http.StatusNotFound
+		}
+		return common.NewApiError(statusCode, errors.Wrapf(err, "Getting cluster %s", clusterID.String()))
+	}
+	return m.detectAndStoreCollidingIPsForCluster(cluster, db)
 }
 
 func (m *Manager) SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID, db *gorm.DB) error {
--- a/internal/host/refresh_status_preprocessor.go
+++ b/internal/host/refresh_status_preprocessor.go
@@ -277,6 +277,10 @@ func newValidations(v *validator) []validation {
 			id:        NoSkipMissingDisk,
 			condition: v.noSkipMissingDisk,
 		},
+		{
+			id:        NoIPCollisionsInNetwork,
+			condition: v.noIPCollisionsInNetwork,
+		},
 	}
 }
 
--- a/internal/host/transition_test.go
+++ b/internal/host/transition_test.go
@@ -6032,6 +6032,7 @@ var allValidationIDs []validationID = []validationID{
 	CompatibleAgent,
 	NoSkipInstallationDisk,
 	NoSkipMissingDisk,
+	NoIPCollisionsInNetwork,
 }
 
 var allConditions []conditionId = []conditionId{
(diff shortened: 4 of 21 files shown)


internal/host/validations_test.go
```diff
@@ -1704,4 +1704,99 @@ var _ = Describe("Validations test", func() {
...
+			//Execute the validations.
+			mockAndRefreshStatus(&host)
+
+			// Retrieve results
+			host = hostutil.GetHostFromDB(*host.ID, host.InfraEnvID, db).Host
+			status, message, ok := getValidationResult(host.ValidationsInfo, NoIPCollisions)
+
+			// Validate expectations
+			Expect(ok).To(BeTrue())
+			Expect(message).To(Equal(fmt.Sprintf("IP collisions were detected on (192.168.1.2) for host %s", host.ID)))
```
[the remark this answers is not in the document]
paul-maidment: I'd suggest revising to `Collisions for IP 192.168.1.2 detected with mac addresses ("de:ad:be:ef:00:00", "c0:fe:e0:00:00") ` what do you think?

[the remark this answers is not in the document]
mkowalski: I like that. Could we write it in a more explicit way though, something like `Multiple MAC addresses (...) detected with the same IP (...)`? For a user who does not understand networking the message must be crystal clear

internal/host/validator.go
```diff
@@ -1714,3 +1714,58 @@ func (v *validator) serviceCanConnectToSpokeKubeAPI(c *validationContext) (Valid
...
+		return ValidationPending, fmt.Sprintf("Awaiting connectivity report for host %s", c.host.ID)
+	}
+	connectivityReport, err := hostutil.UnmarshalConnectivityReport(c.host.Connectivity)
+	if err != nil {
+		message := fmt.Sprintf("Unable to unmarshall host connectivity for %s: %s", c.host.ID, err)
+		v.log.Errorf(message)
+		return ValidationError, message
+	}
+
+	collidingIPSWithMacs := getCollidingIPs(connectivityReport)
```
vrutkovs, line 1731: My understanding is that this would check if node IPs are required to be unique, right? Do we want to make sure that API and Ingress VIPs are not yet assigned too?

internal/host/validations_test.go
```diff
@@ -1704,4 +1704,119 @@ var _ = Describe("Validations test", func() {
...
+			hostId, infraEnvId = strfmt.UUID(uuid.New().String()), strfmt.UUID(uuid.New().String())
+			host = hostutil.GenerateTestHostByKind(hostId, infraEnvId, &clusterID, models.HostStatusDiscovering, models.HostKindHost, models.HostRoleMaster)
+			Expect(db.Create(&host).Error).ShouldNot(HaveOccurred())
+			host = hostutil.GetHostFromDB(*host.ID, host.InfraEnvID, db).Host
+		})
+
+		It("Should mark the validation as pending if the connectivity report is unavailable yet", func() {
+			host.Connectivity = ""
+			mockAndRefreshStatus(&host)
+			host = hostutil.GetHostFromDB(*host.ID, host.InfraEnvID, db).Host
```
vrutkovs, line 1729: This seems to rewrite previous `host.Connectivity = ""`?

internal/host/validations_test.go
```diff
@@ -1704,4 +1704,119 @@ var _ = Describe("Validations test", func() {
...
+			status, message, ok := getValidationResult(host.ValidationsInfo, NoIPCollisions)
+			Expect(ok).To(BeTrue())
```
vrutkovs, line 1731: Lets compare this after `message` and `status` so that we'd get more info about the failure?

paul-maidment: That's fair, I'll reorder these statements

internal/host/validator.go
vrutkovs: Right, probably lets leave it like this. I've seen some installs using node address as API VIP, which will fail the install and was wondering if this validation can protect us from such cases.

paul-maidment: I think there should be another validation should take care of this, this is certainly something worth looking into as the focus of another ticket.

ori-amizur (changes requested):
There is a problem with this approach.
If there is a collision detected, the validation failure will be on the host perform the checks.  The real problem is on the remote host, which multiple MAC addresses were detected for its IP address.
Maybe a better approach will be to detect this problem during connectivity check on the agent.  If such problem happens, the connectivity check will send a failure to reach this host.  Since all the hosts in the cluster will experience the same problem, the host will not belong to the majority group.  In this case, no new validation is needed.

mkowalski:
I agree 100% with Ori on this one here (and that's also what we discussed offline) - our users will be confused by the fact that who failed the validation is not the one who is at fault.

>Maybe a better approach will be to detect this problem during connectivity check on the agent. If such problem happens, the connectivity check will send a failure to reach this host. Since all the hosts in the cluster will experience the same problem, the remote host will not belong to the majority group. In this case, no new validation is needed.

I wonder isn't this only shifting the error into another direction but not necessarily making it clearer? What I mean is that in case of a duplicated IP somewhere in the network the user would see a message saying that we couldn't make a majority group. But no clear reason why.

A yet another approach to consider (but I have no strong opinion on anything taken here) - make it a cluster-level validation named around "IP addressing uniqueness". If we failed the check in subject we wouldn't mark any host as ineligible but rather the cluster, with the message saying "duplicated IP addresses detected in the network". It's generic enough not to cause a confusion, but also detailed enough to point user in some direction

ori-amizur:
> I wonder isn't this only shifting the error into another direction but not necessarily making it clearer? What I mean is that in case of a duplicated IP somewhere in the network the user would see a message saying that we couldn't make a majority group. But no clear reason why.

I think in terms of validation. if there is problem with duplicated MAC address of host IP address - only this host has to marked as failing validation, and not any other host.  Cluster validation will not show it.
In addition, we need a way to explain why host does not belong to a majority which will give the exact reason for the error. This may be even a dedicated API call. I think it is needed regardless of this conflict detection.

mkowalski (approved):
The current message customer sees in case of IP collision in the network is `No connectivity to the majority of hosts in the cluster`. With this PR the message will be `[...] detected collisions [...]`. I agree that overall it may be not ideal where we place this message, but the current one is as helpful as saying "the sky is blue". Customers that see this message have no clue about what's going on. As the new message makes it better, the PR is good enough for me.

Like discussed offline, we may have ideas for the ideal solution to all the networking problems we may encounter, but at the end of the day we have customers that cannot wait for the ideal implementation. As long as this one can help at least one customer to solve the issue on his own without calling us, I'm very happy to see it landing in

paul-maidment:
@omertuc This does not need to wait for the agent change as the check will only fail if multiple unicast replies from arping suggest a collision and in the absence of a change to the agent, this check will simply pass.

This is a non breaking change in that respect.

internal/host/transition_test.go
```diff
@@ -6139,6 +6139,7 @@ var allValidationIDs []validationID = []validationID{
 	CompatibleAgent,
 	NoSkipInstallationDisk,
 	NoSkipMissingDisk,
+	NoIPCollisions,
```
omertuc, line 6142: Let's add a test that we actually move to insufficient when this validation fails.

See existing `"Moves from known to insufficient when disk skip validations fail - no skip installation disk"` test and its siblings, should be trivial to do the same for your validation

internal/host/refresh_status_preprocessor.go
```diff
@@ -277,6 +277,10 @@ func newValidations(v *validator) []validation {
 			id:        NoSkipMissingDisk,
 			condition: v.noSkipMissingDisk,
 		},
+		{
+			id:        NoIPCollisions,
```
omertuc, line 281: Can we include the word `Arping` in this validation name? No IP collisions sounds kinda like you're making sure all cluster hosts have a unique IP, while in reality we're doing something a little more specific - we're making sure arping doesn't indicate *some device* (as in, not necessarily a cluster host) on the network is using an IP that's being used by one of our cluster hosts

internal/host/validator.go
```diff
@@ -1680,3 +1680,59 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+	if len(collidingIPSWithMacs) > 0 {
+		message := "There was an IP collision detected.\n"
+		if len(collidingIPSWithMacs) > 1 {
+			message = "There were multiple IP collissions detected.\n"
+		}
+		for ipAddress, macInfo := range collidingIPSWithMacs {
+			message += fmt.Sprintf("Host %s detected collisions for IP %s, the following macs are in collision (%s)\n", c.host.ID, ipAddress, strings.Join(macInfo, ","))
+
+		}
+		return ValidationFailure, message
```
omertuc, line 1708: Is it really a failure when we notice an IP collision on a network interface that is not part of the cluster's machine network? I wouldn't really care for collisions on something like that, as it will not be used for cluster-critical networking

mkowalski: I cannot think of a customer that would have a topology that has duplicated IP addresses in a single IP segment and be still a valid network topology. Of course I agree that if this happens outside of the Machine Network subnet we will most probably never notice this, but nevertheless if our goal is to assist users, I would vote that this is an issue user should be told about.

I do not push for any direction, but from what we observe super-power-users run Podman on-prem anyway so it does not matter much. For a regular "please hold my hand" type of a user I believe such a message may help him in a long run (not necessarily just for the installation process)

internal/host/refresh_status_preprocessor.go
mkowalski: Or maybe `NoIPCollisionsInNetwork`? Is renaming the validation in the future something expensive? Because I foresee a scenario when we switch from `arping` to some other tool (or maybe even protocol)

internal/host/validator.go
omertuc: >I would vote that this is an issue user should be told about

Yes, but in this case we block the user from installing the cluster, we're not just telling them about it.

But you're correct, making this machine network specific is probably not worth it

internal/host/refresh_status_preprocessor.go
omertuc: Renaming in the future is expe

internal/host/validator.go
mkowalski: I didn't want to be as direct, but I meant to say "regular user will only thank us for blocking him; power user will need a few seconds to disable this validation". I know that blocking the installation is a fragile subject, but I personally think that we should stop users from using infrastructure that is wrong in some fundamental ways, even if those problems do not impact the installation process itself

paul-maidment:
I have made the requested changes and they are ready for review.

mkowalski (approved):

danielerez:
/override edge-e2e-ai-operator-ztp-hypershift-zero-nodes
(job should be fixed and isn't required)

internal/bminventory/inventory.go
```diff
@@ -1597,6 +1597,12 @@ func (b *bareMetalInventory) refreshClusterHosts(ctx context.Context, cluster *c
 		return common.NewApiError(http.StatusInternalServerError, err)
 	}
 
+	err = b.detectAndStoreCollidingIPsForCluster(cluster.ID, tx)
```
ori-amizur, line 1600: Why is it part of inventory.go ?

internal/bminventory/inventory.go
```diff
@@ -3252,6 +3258,10 @@ func (b *bareMetalInventory) setMajorityGroupForCluster(clusterID *strfmt.UUID,
 	return b.clusterApi.SetConnectivityMajorityGroupsForCluster(*clusterID, db)
 }
 
+func (b *bareMetalInventory) detectAndStoreCollidingIPsForCluster(clusterID *strfmt.UUID, db *gorm.DB) error {
```
ori-amizur, line 3261: I think we need to call it only from monitor.

internal/cluster/cluster.go
```diff
@@ -1039,6 +1040,88 @@ func (m *Manager) IsReadyForInstallation(c *common.Cluster) (bool, string) {
...
+	for _, remoteHost := range connectivityReport.RemoteHosts {
+		for _, c := range remoteHost.L2Connectivity {
+			if collisionHistory[c.OutgoingNic] != nil {
+				if previousMac, ok := collisionHistory[c.OutgoingNic][c.RemoteIPAddress]; ok {
+					if previousMac != "" && previousMac != c.RemoteMac {
+						// Collision detected.
+						if collidingIPSWithMacs[c.RemoteIPAddress] == nil {
+							collidingIPSWithMacs[c.RemoteIPAddress] = []string{}
+						}
+						collidingIPSWithMacs[c.RemoteIPAddress] = append(collidingIPSWithMacs[c.RemoteIPAddress], previousMac)
```
ori-amizur, line 1111: Maybe verify slice uniqueness ?

internal/cluster/cluster.go
```diff
@@ -1039,6 +1040,88 @@ func (m *Manager) IsReadyForInstallation(c *common.Cluster) (bool, string) {
...
+	for _, host := range hosts {
+		if len(host.Connectivity) > 0 {
+			connectivityReport, err := hostutil.UnmarshalConnectivityReport(host.Connectivity)
+			if err != nil {
+				// Let's not stop iterating over hosts but let's log this error.
+				m.log.WithError(err).Errorf("unable to unmarshall connectivity report for host %d due to error: %s", host.ID, err.Error())
+			} else {
+				collidingIPSWithMacs = getCollidingIPs(connectivityReport)
+				for k := range collidingIPSWithMacs {
+					ipCollisions[k] = collidingIPSWithMacs[k]
```
ori-amizur, line 1073: Many hosts are iterated.  Why can a result from current host overwrite result from previous host? Shouldn't results be appended ?

internal/cluster/cluster.go
```diff
@@ -1039,6 +1040,88 @@ func (m *Manager) IsReadyForInstallation(c *common.Cluster) (bool, string) {
...
+	b, err := json.Marshal(&collidingIPSWithMacs)
+	if err != nil {
+		return common.NewApiError(http.StatusInternalServerError, err)
+	}
+
+	marshalledCollidingIPSWithMacs := string(b)
+	if marshalledCollidingIPSWithMacs != cluster.IPCollisions {
+		err = db.Model(&common.Cluster{}).Where("id = ?", cluster.ID.String()).Updates(&common.Cluster{
+			Cluster: models.Cluster{
+				IPCollisions: marshalledCollidingIPSWithMacs,
```
ori-amizur, line 1089: Add trigger monitor timestamp, to trigger monitoring

internal/cluster/cluster.go
```diff
@@ -1039,6 +1040,88 @@ func (m *Manager) IsReadyForInstallation(c *common.Cluster) (bool, string) {
...
+		models.ClusterStatusPendingForInput,
+		models.ClusterStatusInsufficient,
+
+		models.ClusterStatusReady,
+	}
+	if !funk.ContainsString(allowedStates, swag.StringValue(cluster.Status)) {
+		return nil
+	}
+
+	hosts := cluster.Hosts
```
ori-amizur, line 1058: Maybe copy to a new slice without changing the original slice ?

internal/cluster/cluster.go
```diff
@@ -1098,6 +1181,18 @@ func (m *Manager) setConnectivityMajorityGroupsForClusterInternal(cluster *commo
 	return nil
 }
 
+func (m *Manager) DetectAndStoreCollidingIPsForCluster(clusterID strfmt.UUID, db *gorm.DB) error {
```
ori-amizur, line 1184: Should be done only in monitoring, so this function is not needed

internal/host/validator.go
```diff
@@ -1766,3 +1766,63 @@ func (v *validator) noSkipMissingDisk(c *validationContext) (ValidationStatus, s
...
+	}
+	if hasCollisions {
+		return ValidationFailure, collisionValidationText
+	}
+
+	return ValidationSuccess, fmt.Sprintf("No IP collisions were detected by host %s", c.host.ID)
+}
+
+func (v *validator) inventoryHasIP(inventory *models.Inventory, ipAddress string) (bool, error) {
+	ip := net.ParseIP(ipAddress)
```
ori-amizur, line 1813: Please verify that the result is not nil

internal/bminventory/inventory.go
slaviered: @ori-amizur what are the possible triggers for change in the collision status?

internal/cluster/cluster.go
slaviered: @ori-amizur there are quite a few places in the code where the hosts are sorted in memory just so the order will be predictable. I believe it would be much faster to read the hosts from the db ORDER BY the hostId and have them already sorted on the cluster object. WDYT?

internal/cluster/cluster.go
slaviered: I also agree that this should be done in monitoring. But please also note that in the current context of the inventory you are retrieving the cluster and hosts twice because they were already retrieved in the calling function. In addition, the calling function of this flow is invoked only in interactive mode but I think that the collision should be a general validation (not only in SaaS)

internal/cluster/cluster.go
slaviered: Please document the structure of the collision map with an example and what information we expect to retrieve from it.

internal/cluster/cluster.go
```diff
@@ -1039,6 +1040,88 @@ func (m *Manager) IsReadyForInstallation(c *common.Cluster) (bool, string) {
...
 
+func (m *Manager) detectAndStoreCollidingIPsForCluster(cluster *common.Cluster, db *gorm.DB) error {
+	if db == nil {
+		db = m.db
+	}
+	// We want to calculate ip collisions only when in pre-install states since it is needed for pre-install validations
+	allowedStates := []string{
+		models.ClusterStatusPendingForInput,
+		models.ClusterStatusInsufficient,
+
```
slaviered, line 1051: redundant space. Plus do we want to skip pending-for-input state because it may take quite some time?

internal/bminventory/inventory.go
paul-maidment: By the same logic, why is `setMajorityGroupForCluster` part of inventory.go?

filanov: I agree with Ori, it should not be included here. It need to be part of the monitor.

filanov: but because the order is that cluster collect the data and it will take affect in the host maybe it make sense to add it to https://github.com/openshift/assisted-service/blob/b8b1e4d624327361184ab39581321759e8eb9216/internal/bminventory/inventory.go#L1125

filanov: And if this is is the order then it can be called here as well?

filanov: To summarize it, because cluster is collecting the data and the host will be affected we need to
1. collect 
2. refresh host
3. refresh cluster 
If this is the case then we have 3 places that we need to run this function from
1. https://github.com/openshift/assisted-service/blob/b8b1e4d624327361184ab39581321759e8eb9216/internal/bminventory/inventory.go#L1125
2. https://github.com/openshift/assisted-service/blob/b8b1e4d624327361184ab39581321759e8eb9216/internal/bminventory/inventory.go#L1640
3. https://github.com/openshift/assisted-service/blob/b8b1e4d624327361184ab39581321759e8eb9216/internal/cluster/cluster.go#L524

ori-amizur: I think monitor is mandatory.  
This validation will not change as a result of user action.  Therefore, monitoring should be enough.

internal/bminventory/inventory.go
ori-amizur: Change in connectivity result received from host

internal/cluster/cluster.go
ori-amizur: I am not sure all places need the same order

internal/cluster/cluster.go
```diff
@@ -1039,6 +1040,88 @@ func (m *Manager) IsReadyForInstallation(c *common.Cluster) (bool, string) {
...
+	hosts := cluster.Hosts
+	sort.Slice(hosts, func(i, j int) bool {
+		return hosts[i].ID.String() < hosts[j].ID.String()
+	})
+	ipCollisions := make(map[string][]string)
+	collidingIPSWithMacs := make(map[string][]string)
+	for _, host := range hosts {
+		if len(host.Connectivity) > 0 {
+			connectivityReport, err := hostutil.UnmarshalConnectivityReport(host.Connectivity)
```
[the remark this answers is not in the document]
paul-maidment: I'm not really sure where this cache would be implemented, has this been done elsewhere in the code?
Surely this would depend on us knowing the previous checksum, I am not sure this is the situation here?

internal/cluster/cluster.go
slaviered: Most of them do

PR #4463, opened 2022-10-04
MGMT-11700: Validate storage requirements

jhernand:
Currently when the user creates an instance of the _AgentConfigService_
object the sizes of the volumes aren't validated. This patch changes the
controller so that it validates these sizes.

To avoid potential issues with environments that were created before
this change the validation will not generate an error and stop
reconciliation if the corresponding persistent volume claims and
stateful set have already been created.

Related: https://issues.redhat.com/browse/MGMT-11700

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Tested manually deploying the operator and trying to create _AgentServiceConfig_ objects with incorrect sizes.

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/go-openapi/swag"
 	"github.com/hashicorp/go-version"
@@ -97,6 +98,10 @@ var (
 	databasePort         = intstr.Parse("5432")
 	imageHandlerPort     = intstr.Parse("8080")
 	imageHandlerHTTPPort = intstr.Parse("8081")
+
+	minDatabaseStorage   = resource.MustParse("1Gi")
+	minFilesystemStorage = resource.MustParse("1Gi")
+	minImageStorage      = resource.MustParse("10Gi")
 )
 
 type AgentServiceConfigReconcileContext struct {
@@ -250,6 +255,48 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
 		}
 
 		return ctrl.Result{}, nil
+	}
+
+	// Validate the storage configuration. If that returns warnings then generate the
+	// corresponding events. If it returns errors then update the conditions and stop the
+	// reconciliation.
+	warnings, failures, err := r.validateStorage(ctx, log, instance)
+	if err != nil {
+		log.WithError(err).Error("Failed to validate storage configuration")
+		return ctrl.Result{Requeue: true}, err
+	}
+	for _, warning := range warnings {
+		r.Recorder.Event(instance, "Warning", aiv1beta1.ReasonStorageFailure, warning)
+	}
+	if len(failures) > 0 {
+		log.Error("Storage configuration isn't valid")
+		conditionsv1.SetStatusConditionNoHeartbeat(
+			&instance.Status.Conditions, conditionsv1.Condition{
+				Type:    aiv1beta1.ConditionReconcileCompleted,
+				Status:  corev1.ConditionFalse,
+				Reason:  aiv1beta1.ReasonStorageFailure,
+				Message: fmt.Sprintf("%s.", strings.Join(failures, ". ")),
+			},
+		)
+		err = r.Status().Update(ctx, instance)
+		if err != nil {
+			log.WithError(err).Error("Failed to update status")
+			return ctrl.Result{Requeue: true}, err
+		}
+		return ctrl.Result{}, err
+	}
+
+	// If we are here then the storage configuration validation succeeded, so we may need to
+	// remove a previous failure condition:
+	condition := conditionsv1.FindStatusCondition(
+		instance.Status.Conditions,
+		aiv1beta1.ConditionReconcileCompleted,
+	)
+	if condition != nil && condition.Reason == aiv1beta1.ReasonStorageFailure {
+		conditionsv1.RemoveStatusCondition(
+			&instance.Status.Conditions,
+			aiv1beta1.ConditionReconcileCompleted,
+		)
 	}
 
 	for _, component := range []component{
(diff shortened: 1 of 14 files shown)


pkg/webhooks/agentinstall/v1beta1/agent_service_config_admission_hook.go
```diff
@@ -0,0 +1,324 @@
...
+	agentServiceConfigResource         = "agentserviceconfigs"
+	agentServiceConfigAdmissionGroup   = "admission.agentinstall.openshift.io"
+	agentServiceConfigAdmissionVersion = "v1"
+)
+
+// Minimum sizes of the storage areas:
+var (
+	minDatabaseStorage   = resource.MustParse("50M")
+	minFilesystemStorage = resource.MustParse("50M")
+	minImageStorage      = resource.MustParse("10G")
```
jhernand, line 29: These sizes are just examples, I will adjust them to the real values before merging this.

vrutkovs: I think we want these configurable. 10G image storage, for instance, may be too large for some systems

jhernand:
@omertuc I changed the patch to use a validation in the controller instead of a webook, as you suggested.

@vrutkovs regarding making this configurable I think there is no easy way to add a configuration parameter to an operator, at least not one that users can easily change. It could be done adding new fields to the `AgentServiceConfig` CRD, but it would be weird if the user had to configure the minimum requirement and the actual value in the same place. To address your concern about environments where the minimum requirements may not make sense I think we can add support for an annotation in that CRD that disables the validation. Didn't do that in this version of the patch yet, will do in the next if we agree it is reasonable.

omertuc:
Why the go.mod/go.sum/vendor changes?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -2245,3 +2274,82 @@ func (r *AgentServiceConfigReconciler) getImageService(ctx context.Context, log
...
+			"Database storage %s is too small, it must be at least %s",
+			databaseStorage, &minDatabaseStorage,
+		)
+		warnings = append(warnings, message)
+		key := client.ObjectKey{
+			Namespace: r.Namespace,
+			Name:      databaseName,
+		}
+		var tmp corev1.PersistentVolumeClaim
+		err = r.Get(ctx, key, &tmp)
```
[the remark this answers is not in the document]
jhernand: Done, explained that in the documentation comment of the `validateStorage` method.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -220,6 +225,30 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
...
 
+	warnings, failures, err := r.validateStorage(ctx, log, instance)
+	if err != nil {
+		log.WithError(err).Error("Failed to validate storage configuration")
+		return ctrl.Result{Requeue: true}, err
+	}
+	for _, warning := range warnings {
+		r.Recorder.Event(instance, "Warning", aiv1beta1.ReasonStorageFailure, warning)
+	}
+	if len(failures) > 0 {
```
omertuc, line 236: else { clean the status condition? }

jhernand:
@omvertuc the changes in `go.mod` and `vendor` are to add the `github.com/bombsimon/logrusr`. That is a bridge between `logr` and `logrus` useful for sending the controller-runtime logs to the Ginkgo writer in the tests.

internal/controller/controllers/agentserviceconfig_controller.go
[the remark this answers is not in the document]
omertuc: Perfect, thank you

omertuc:
> @omertuc the changes in `go.mod` and `vendor` are to add the `github.com/bombsimon/logrusr`. That is a bridge between `logr` and `logrus` useful for sending the controller-runtime logs to the Ginkgo writer in the tests.

Ah okay, didn't notice the test added an additional dependency, so I assumed it was just a miscommit

internal/controller/controllers/agentserviceconfig_controller.go
jhernand: I think it isn't strictly needed because when the validation finishes successfully the rest of the controller will continue working, and it will eventually set another value for this condition, either true or false, doesn't really matter; it will replace the current value. However it is probably clearer it is explicitly removed as you suggest. I will do so.

jhernand: Done. Also added a test to verify that the condition is updated when the storage configuration is fixed.

internal/controller/controllers/agentserviceconfig_controller_storage_validation_test.go
```diff
@@ -0,0 +1,510 @@
...
+		logger     *logrus.Logger
+		client     clnt.WithWatch
+		watches    []watch.Interface
+		subject    *aiv1beta1.AgentServiceConfig
+		key        clnt.ObjectKey
+		request    ctrl.Request
+		reconciler *AgentServiceConfigReconciler
+	)
+
+	BeforeEach(func() {
```
omertuc, line 37: This is a really impressive test setup, can we refactor it out of this particular test so it can be reused in future AgentServiceConfig related tests?

I think it will also make the test clearer and more focused as all this setup isn't really specific to what is being tested

jhernand: @omertuc in the new version of the patch I moved that setup to a `FakeCluster` type inside the `internal/testing` package. Let me know what you think.

internal/controller/controllers/agentserviceconfig_controller_storage_validation_test.go
```diff
@@ -0,0 +1,432 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	clnt "sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/openshift/assisted-service/internal/testing"
+)
+
+var _ = FDescribe("Agent service config controller storage validation", func() {
```
omertuc, line 20:
```suggestion
var _ = Describe("Agent service config controller storage validation", func() {
```

jhernand: Done.

internal/controller/controllers/agentserviceconfig_controller_storage_validation_test.go
```diff
@@ -0,0 +1,431 @@
...
+	When("Objects haven't been created", func() {
+		It("Rejects small database storage", func() {
+			subject.Spec.DatabaseStorage.Resources.Requests = corev1.ResourceList{
+				corev1.ResourceStorage: resource.MustParse("1Mi"),
+			}
+			err := client.Create(ctx, subject)
+			Expect(err).ToNot(HaveOccurred())
+			Eventually(func(g Gomega) {
+				_, err := reconciler.Reconcile(ctx, request)
+				g.Expect(err).ToNot(HaveOccurred())
```
omertuc, line 113: When you don't care about the value returned, then gomega Expect looks only at the last parameter in the return value, so you can write more concise:

```suggestion
				g.Expect(reconciler.Reconcile(ctx, request)).To(Succeed())
```

omertuc: (this can be used in many other places in your code, not just this line)

internal/controller/controllers/agentserviceconfig_controller_storage_validation_test.go
```diff
@@ -0,0 +1,431 @@
...
+				err = client.Get(ctx, key, subject)
+				g.Expect(err).ToNot(HaveOccurred())
```
omertuc, line 115: Any reason to use the `g` instance here rather than just the global `Expect` function?

internal/controller/controllers/agentserviceconfig_controller_storage_validation_test.go
jhernand: I am aware of that, but I prefer to do it this way because I find nested function calls harder to read. Probably just a matter of taste. I will change it if you insist :-) .

omertuc: Nah I don't insist, just a personal preference. But don't do it if you don't prefer it as well

internal/controller/controllers/agentserviceconfig_controller_storage_validation_test.go
jhernand: Using the `g.Expect(...)` here is needed because using just `Expect(...)` here will indicate to the `gomega` library that the the test has failed completely. What we want to indicate here is that the current iteration inside `Eventually(...)` failed, so that the library will try again.

internal/controller/controllers/agentserviceconfig_controller_storage_validation_test.go
jhernand: I will leave it as it is then, thanks @omertuc. Note that I meant it: if you had insisted I would happily have changed it.

jhernand:
The new version of the patch updates the `setup_assisted_operator.sh` script so that it requests 10Gi for the image storage instead of 8Gi, as 8Gi is not valid and was making some tests fail.

PR #4464, opened 2022-10-04
MGMT-8890: Allow weak matching for GetReleaseImage

mkowalski:
This PR extends the GetReleaseImage function to allow for weak matching between requested and available OpenShift versions. This is to allow queries for e.g. `4.12` when only `4.12.1` is available.

The extensions allows to use the GetReleaseImage for validation purposes when the caller only cares about a non-empty output, but does not care which exactly release image is matched.

This has been discovered only after multi-arch clusters got the validation between InfraEnv (i.e. RHCOS image) and OCP release. With the ability to perform weak matching we don't need to align RHCOS image to the z-stream of OCP, so e.g. we can offer `4.12.0-ec.3-multi` to be served by `4.12` RHCOS (without that change we need to label RHCOS explicitly as `4.12.0-ec.3-multi`).

Single-arch clusters are not affected, as for such a cluster InfraEnv does not perform any validation whatsoever.

Closes: Bug-2131396
Closes: [MGMTBUGSM-567](https://issues.redhat.com//browse/MGMTBUGSM-567)
Caused-by: [MGMT-11500](https://issues.redhat.com//browse/MGMT-11500)
Caused-by: #4285

/cc @carbonin

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -37,7 +37,7 @@ type Versions struct {
 type Handler interface {
 	restapi.VersionsAPI
 	GetMustGatherImages(openshiftVersion, cpuArchitecture, pullSecret string) (MustGatherVersion, error)
-	GetReleaseImage(openshiftVersion, cpuArchitecture string) (*models.ReleaseImage, error)
+	GetReleaseImage(openshiftVersion, cpuArchitecture string, strongMatch bool) (*models.ReleaseImage, error)
 	GetDefaultReleaseImage(cpuArchitecture string) (*models.ReleaseImage, error)
 	GetOsImage(openshiftVersion, cpuArchitecture string) (*models.OsImage, error)
 	GetLatestOsImage(cpuArchitecture string) (*models.OsImage, error)
@@ -195,7 +195,7 @@ func (h *handler) GetMustGatherImages(openshiftVersion, cpuArchitecture, pullSec
 		return versions, nil
 	}
 	//if not, fetch it from the release image and add it to the cache
-	releaseImage, err := h.GetReleaseImage(openshiftVersion, cpuArchitecture)
+	releaseImage, err := h.GetReleaseImage(openshiftVersion, cpuArchitecture, true)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


mkowalski:
Confirmed to work via manual test using `quay.io/mkowalski/assisted-service:trey-some-bug-4`

internal/versions/versions.go
```diff
@@ -37,7 +37,7 @@ type Versions struct {
 type Handler interface {
 	restapi.VersionsAPI
 	GetMustGatherImages(openshiftVersion, cpuArchitecture, pullSecret string) (MustGatherVersion, error)
-	GetReleaseImage(openshiftVersion, cpuArchitecture string) (*models.ReleaseImage, error)
+	GetReleaseImage(openshiftVersion, cpuArchitecture string, strongMatch bool) (*models.ReleaseImage, error)
```
carbonin, line 40: This might be a time where we'd benefit from defining specific types for the matching strategies.
As is every time a see a call to this function I'm going to have to look at the definition to figure out what's going on.

internal/versions/versions_test.go
```diff
@@ -578,18 +578,55 @@ var _ = Describe("list versions", func() {
...
 			It("gets successfuly image using sub-architecture", func() {
-				releaseImage, err = h.GetReleaseImage("4.11.1", common.PowerCPUArchitecture)
+				releaseImage, err = h.GetReleaseImage("4.11.1", common.PowerCPUArchitecture, true)
 				Expect(err).ShouldNot(HaveOccurred())
 				Expect(*releaseImage.OpenshiftVersion).Should(Equal("4.11.1"))
 				Expect(*releaseImage.Version).Should(Equal("4.11.1-multi"))
 			})
 		})
+
+		Context("using weak match", func() {
```
carbonin, line 594: Can you include a case where we have a release that would weak match, but we specify strong match and we expect it to not find the release? Or is this covered with one of the existing tests in the other contexts?

carbonin (commented):
Do you think it would be better to parse the name further and be more explicit about this?

For example should we have releases be identified by arch, version (which is just x.y.z) and some "extra" field that would include these additional qualifiers?

Then we could either pass a specific extra field or not to indicate if we want a specific release.

carbonin:
/hold

@mkowalski and I talked about this and he's going to look at this in another way

mkowalski:
/close

Superseded by #4479

PR #4465, opened 2022-10-04
NO-ISSUE: Change `get_current_size_if_exist` to inspect PVC `.spec` and not `.status`

omertuc:
Users have encountered the following stack trace during CI runs:

```
[INFO] 2022-10-03 00:49:14,276 - deploy_postgres - Deploying /home/assisted/assisted-service/build/ocp-edge-qe/postgres-deployment.yaml", "deployment.apps/postgres unchanged", "service/postgres unchanged
[INFO] 2022-10-03 00:49:14,798 - deploy_postgres - Updating pvc size for postgres-pv-claim
  File \"./tools/deploy_postgres.py\", line 97, in <module>
    main()", "  File \"./tools/deploy_postgres.py\", line 19, in main
   deploy_postgres_storage(deploy_options)
  File \"./tools/deploy_postgres.py\", line 78, in deploy_postgres_storage
    docs=docs", "  File \"/home/assisted/assisted-service/tools/pvc_size_utils.py\", line 69, in update_size_in_yaml_docs
    size = determine_which_size_to_deploy(req, cur)
  File \"/home/assisted/assisted-service/tools/pvc_size_utils.py\", line 113, in determine_which_size_to_deploy
    cur_bytes = size_to_bytes(cur)
  File \"/home/assisted/assisted-service/tools/pvc_size_utils.py\", line 123, in size_to_bytes
    return BytesSuffix.to_bytes(s)
  File \"/home/assisted/assisted-service/tools/pvc_size_utils.py\", line 21, in to_bytes
    f'failed to convert size to bytes: {s}'
 ValueError: failed to convert size to bytes:
make: *** [deploy-postgres] Error 1
```

It seems that if for some reason the output we get from kubectl is
empty, then we eventually fail in this confusing manner.

This seems to be due to the `get_current_size_if_exist` function
returning an empty string.

The only situation I can imagine where the output for
`get_current_size_if_exist` is empty is when the PVC exists yet
`.status.capacity.storage` is not set. I guess this could happen when
the PVC gets created but didn't get assigned any PV yet.

In any case, it seems better if we instead inspect the `.spec` of the
PVC instead of its `.status`, because for determining whether a bigger
PVC is needed, this is enough. A `.status` is not guaranteed to exist
at the time we deploy the database, but a `.spec` is.

- Stop treating text in kubectl stderr output as a total failure, look at exit code
  instead, sometimes kubectl outputs warnings that are not fatal so this
  doesn't make sense

- If the output is somehow empty, immediately raise an exception rather
  than letting the code further down in the code fail unexpectedly. This
  is just in case the `.spec` solution above is not enough to solve this
  problem.

- Removed unncessary use of `shell=True`

- [ ] New Feature 
- [x] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/tools/pvc_size_utils.py
+++ b/tools/pvc_size_utils.py
@@ -1,5 +1,6 @@
-import subprocess as sp
+import subprocess
 import utils
+import shlex
 
 
 class BytesSuffix(object):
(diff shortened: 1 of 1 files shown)


adriengentil (approved):

adriengentil (approved):
:+1:

PR #4466, opened 2022-10-04
NO-ISSUE: reduce build image size and leverage buildkit cache

rccrdpccl:
This PR introduces the use of buildkit for build-image.
The objective is to reduce build time and image size, to have a more pleasant developer experience.
This is a first iteration, if this is successful we can apply the same concepts to other images and also refine the process further.

The main advantage is cache leverage. Most (all?) network-bound dependencies are being cached, so that if we run the build multiple times on the machine, only new dependencies will need networking, everything else will be found in the cache (until the cache is cleaned up of course). Probably this will not affect CI, as I am unsure if the buildkit cache can be shared between runs.
Speed for non-first builds improves massively, virtually immediate.

Another benefit for this PR is image size reduction:
Before:
```
assisted-service-build    2.79GB
```
After:
```
assisted-service-build     2.27GB
```

This can be further improved mainly by checking what we actually need and deleting what we don't, but also by further removing duplication among layers.

The main disadvantage of this approach is that we need to enable buildkit in order to build the image:
```
export DOCKER_BUILDKIT=1
```

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed
- build works, will run CI to see if it breask anywhere (it shouldn't, as functionality should not change and we have tried buldkit in other projects)

## Checklist

## Reviewers Checklist


hack/setup_env.sh
```diff
@@ -15,9 +19,20 @@ function print_help() {
   echo "Usage: bash ${0} (${ALL_FUNCS})"
 }
 
+function download() {
+    url=$1
+    output=$2
+    cached_filename=$(echo -n "$url" | sha256sum | awk '{ print $1 }')
+    cache_full_path="${CACHE_DIR}/${cached_filename}"
+    [ -f "${cache_full_path}" ] || curl --retry 5 -L "$url" -o "$cache_full_path"
```
rccrdpccl, line 27: By doing this we will leverage buildkit cache, and avoid downloading the file again if we already have a file with the same name in the cache directory

hack/setup_env.sh
```diff
@@ -89,9 +105,10 @@ function assisted_service() {
 
   OS=$(uname | awk '{print tolower($0)}')
   OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.10.1
-  curl --retry 5 -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH}
+  download ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH} operator-sdk_${OS}_${ARCH}
   chmod +x operator-sdk_${OS}_${ARCH}
   install operator-sdk_${OS}_${ARCH} /usr/local/bin/operator-sdk
+  rm -rf operator-sdk_${OS}_${ARCH}
```
rccrdpccl, line 111: This was forgotten, 75MB this alone

Dockerfile.assisted-service-build
```diff
@@ -13,23 +10,36 @@ ENV PATH="$VIRTUAL_ENV/bin:$GOROOT/bin:$GOPATH/bin:$TOOLS:$PATH"
...
+    --mount=type=cache,target=/tmp/.cache\
+    --mount=type=tmpfs,destination=/var/cache\
+    --mount=type=cache,target=/var/cache\
+    ./setup_env.sh podman_remote && \
     ./setup_env.sh assisted_service && \
-    dnf clean all && \
     chmod g+xw -R $GOPATH
 COPY ./data /tmp/data
+
+# merge layers
```
rccrdpccl, line 43: This takes an empty image and copies over everything from the previous building process.
The advantage of this, is that we are effectively squashing all layers. Won't be very effective in development, as we will have dangling layers, but if this is used in CI or in PROD image it could bring benefits by cutting down the size of the image without going crazy with deduplicating every single duplicated file among layers

adriengentil:
Have you tested with podman as well?

Dockerfile.assisted-service-build
adriengentil: How long this layer takes to be build? I think it's better to focus on developer experience as this image is mostly used along with skipper.

hack/setup_env.sh
adriengentil: Can you add what you said as a comment in the source file?

adriengentil:
> The main disadvantage of this approach is that we need to enable buildkit in order to build the image:
> 
> ```
> export DOCKER_BUILDKIT=1
> ```

Since it's used through skipper, we may supported it directly through it? via an option in the `skipper.yaml` maybe?

rccrdpccl:
> you mean `podman-remote`? I was thinking about the "plain" podman

no, plain podman, just not directly through my system:
```
docker run --privileged --entrypoint=/bin/sh -it -v $(pwd):/app mgoltzsche/podman:4-minimal
```

rccrdpccl:
> Since it's used through skipper, we may supported it directly through it? via an option in the `skipper.yaml` maybe?

AFAIU skipper image is built through the host, the env var must be enabled there. We could enable buildkit within skipper when applying this to the production image (if people build it through skipper and not from the host directly)

hack/setup_env.sh
```diff
@@ -15,9 +19,23 @@ function print_help() {
   echo "Usage: bash ${0} (${ALL_FUNCS})"
 }
 
+function download() {
+    # This function will leverage buildkit cache, and avoid downloading the file again
+    # if we already have a file with the same name in the cache directory
```
osherdp, line 24: can't we do the check in the target destination of the file?
I mean, we can consider the target path as a "cache" in a way, and remove the need of file duplications

osherdp: maybe it will lack the versioning impact in case someone changes the binary version, but seems good to go simpler at first than have a more complex logic

hack/setup_env.sh
```diff
@@ -89,9 +108,10 @@ function assisted_service() {
 
   OS=$(uname | awk '{print tolower($0)}')
   OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.10.1
-  curl --retry 5 -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH}
+  download ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH} operator-sdk_${OS}_${ARCH}
   chmod +x operator-sdk_${OS}_${ARCH}
   install operator-sdk_${OS}_${ARCH} /usr/local/bin/operator-sdk
+  rm -rf operator-sdk_${OS}_${ARCH}
```
osherdp, line 114: let's store it directly at ``/usr/local/bin`` and remove the ``rm``

osherdp (changes requested):

hack/setup_env.sh
osherdp: ack

PR #4475, opened 2022-10-07
Bug 2106110: Don't set ip=<nic>:dhcp kernel params with static networking

mkowalski:
This PR changes the behaviour of logic responsible for appending `ip=<nic>:dhcp` kernel parameters. Those must not be added when static network configuration is used as in some specific cases it causes loss of connectivity due to NetworkManager having to deal with two contradicting configurations (static config created by NMstate and DHCP as requested by kernel parameter).

With this change as soon as we detect that InfraEnv has network configuration provided by the user, we will not add any `ip` kernel param. This is not expected to cause any trouble as the behaviour will be as follows

* with no static configuration, `ip=<nic>:dhcp` will be set on the NIC that is used to provision the server (behaviour like today)
* with static configuration the NIC used to provision the server should have IP address assigned with NMstate

Closes: Bug-2106110
Closes: [MGMTBUGSM-508](https://issues.redhat.com//browse/MGMTBUGSM-508)

/cc @carbonin

### DIFF
--- a/internal/host/hostcommands/install_cmd.go
+++ b/internal/host/hostcommands/install_cmd.go
@@ -256,17 +256,27 @@ func (i *installCmd) getDisksToFormat(ctx context.Context, host *models.Host, in
 func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inventory *models.Inventory, infraEnv *common.InfraEnv, log logrus.FieldLogger) (string, error) {
 
 	var installerArgs []string
+	var err error
+	hasStaticNetwork := (infraEnv != nil && infraEnv.StaticNetworkConfig != "") || cluster.StaticNetworkConfigured
 
 	if host.InstallerArgs != "" {
-		err := json.Unmarshal([]byte(host.InstallerArgs), &installerArgs)
+		err = json.Unmarshal([]byte(host.InstallerArgs), &installerArgs)
 		if err != nil {
 			return "", err
 		}
 	}
 
-	installerArgs, err := appendDHCPArgs(cluster, host, inventory, installerArgs, log)
-	if err != nil {
-		return "", err
+	if !hasStaticNetwork {
+		// The set of ip=<nic>:dhcp kernel arguments should be added only if there is no static
+		// network configured by the user. This is because this parameter will configure RHCOS to
+		// try to obtain IP address from the DHCP server even if we provide a static addressing.
+		// As in majority of cases it's not an issue because of the priorities set in the config
+		// of NetworkManager, in some specific scenarios (e.g. BZ-2106110) this causes machines to
+		// lose their connectivity because priorities get mixed.
+		installerArgs, err = appendDHCPArgs(cluster, host, inventory, installerArgs, log)
+		if err != nil {
+			return "", err
+		}
 	}
 
 	for _, disk := range inventory.Disks {
@@ -275,7 +285,6 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
 		}
 	}
 
-	hasStaticNetwork := (infraEnv != nil && infraEnv.StaticNetworkConfig != "") || cluster.StaticNetworkConfigured
 	if hasStaticNetwork && !funk.Contains(installerArgs, "--copy-network") {
 		// network not configured statically or
 		// installer args already contain command for network configuration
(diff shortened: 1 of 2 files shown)


mkowalski:
/retitle Bug 2106110: Don't set ip=<nic>:dhcp kernel params with static networking

vrutkovs (approved):
/lgtm
/hold

for tests (to avoid regressions mostly)

mkowalski:
/retest

```
 + echo '### Copying test binaries'
### Copying test binaries
+ scp -o ConnectTimeout=5 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ServerAliveInterval=90 -o LogLevel=ERROR -i /var/run/secrets/ci.openshift.io/cluster-profile/packet-ssh-key /usr/bin/openshift-tests /usr/bin/kubectl root@147.75.84.177:/usr/local/bin
kex_exchange_identification: Connection closed by remote host
lost connection 
```

omertuc:
>with static configuration the NIC used to provision the server should have IP address assigned with NMstate

This assumes the user uses nmstate for set a static IP address - which is not a necessity, nmstate could be used for other things, e.g. setting a different DNS server

internal/host/hostcommands/install_cmd.go
```diff
@@ -256,17 +256,27 @@ func (i *installCmd) getDisksToFormat(ctx context.Context, host *models.Host, in
...
+		err = json.Unmarshal([]byte(host.InstallerArgs), &installerArgs)
 		if err != nil {
 			return "", err
 		}
 	}
 
-	installerArgs, err := appendDHCPArgs(cluster, host, inventory, installerArgs, log)
-	if err != nil {
-		return "", err
+	if !hasStaticNetwork {
```
carbonin, line 269: Can/should we do something more clever here?

I remember people mentioning that there are many cases where users configure static networking only for a subset of the interfaces on a host.

Should we only be omitting the ones they have specifically set static networking for?

omertuc:
> > with static configuration the NIC used to provision the server should have IP address assigned with NMstate
> 
> This assumes the user uses nmstate for setting static IP addresses - which is not a necessity, nmstate could be used (potentially exclusively) for other things as well, such as setting a different DNS server, disabling interfaces, etc

We may decide we don't care aobut those users because they are rare, but we should probably be explicit about this and reject nmstate configurations that use DHCP

mkowalski:
Your discussion is orthogonal, not having `ip=eno1:dhcp` does *not* mean you can't have NMstate with DHCP. It only configures what happens in the absence of any NMstate whatsoever.

So

>but we should probably be explicit about this and reject nmstate configurations that use DHCP, otherwise we're just gonna run into confusing bugs with those users

No, if you configure DHCP explicitly, nothing changes

>nmstate could be used (potentially exclusively) for other things as well, such as setting a different DNS server, disabling interfaces

We don't support this. You cannot use NMstate for setting only DNS; also you can't use it to exclusively take the interface down

>I remember people mentioning that there are many cases where users configure static networking only for a subset of the interfaces on a host.

Yes. With this change if you think that you can use static nmstate for some interfaces and then have DHCP automagically for the others, it will not score

carbonin:
> Yes. With this change if you think that you can use static nmstate for some interfaces and then have DHCP automagically for the others, it will not score

Okay, so the intention then is to have these users now explicitly configure dhcp on the interfaces that previously would have gotten it because of this kernel param?

omertuc:
> > nmstate could be used (potentially exclusively) for other things as well, such as setting a different DNS server, disabling interfaces
> 
> We don't support this. You cannot use NMstate for setting only DNS; also you can't use it to exclusively take the interface down

Source?

mkowalski:
> > We don't support this. You cannot use NMstate for setting only DNS; also you can't use it to exclusively take the interface down
> 
> Source?

https://coreos.slack.com/archives/CUPJTHQ5P/p1663597152614129?thread_ts=1663593059.601929&cid=CUPJTHQ5P

mkowalski:
/test edge-e2e-ai-operator-ztp-ipv4v6-3masters-ocp-411

Issue with Hub cluster

mkowalski:
@vrutkovs `edge-e2e-metal-assisted-kube-api-net-suite` seems to be broken and unused (based on the history); anything else that seemed relevant is passing. Let's unhold?

vrutkovs:
Agree

/hold cancel

PR #4479, opened 2022-10-10
MGMT-8890: Create validator for RHCOS and OCP versions

mkowalski:
This PR creates the `ValidateReleaseImageForRHCOS` function responsible for validating whether a specified RHCOS version has a matching OCP release available, as otherwise it makes no sense to create such an InfraEnv as it would be effectively unusable until a valid OCP version appears.

This has been discovered only after multi-arch clusters got the validation between InfraEnv (i.e. RHCOS image) and OCP release. With the ability to perform weak matching we don't need to align RHCOS image to the z-stream of OCP, so e.g. we can offer `4.12.0-ec.3-multi` to be served by `4.12` RHCOS (without that change we need to label RHCOS explicitly as `4.12.0-ec.3-multi`).

Single-arch clusters are not affected, as for such a cluster InfraEnv does not perform any validation whatsoever.

Closes: Bug-2131396
Closes: [MGMTBUGSM-567](https://issues.redhat.com//browse/MGMTBUGSM-567)
Caused-by: [MGMT-11500](https://issues.redhat.com//browse/MGMT-11500)
Caused-by: #4285
Supersedes: #4464

/cc @carbonin

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -46,6 +46,7 @@ type Handler interface {
 	GetOpenshiftVersions() []string
 	AddReleaseImage(releaseImageUrl, pullSecret, ocpReleaseVersion string, cpuArchitectures []string) (*models.ReleaseImage, error)
 	ValidateAccessToMultiarch(ctx context.Context, authzHandler auth.Authorizer) error
+	ValidateReleaseImageForRHCOS(rhcosVersion, cpuArch string) error
 }
 
 func NewHandler(log logrus.FieldLogger, releaseHandler oc.Release,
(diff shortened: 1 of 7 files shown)


internal/versions/versions.go
```diff
@@ -327,6 +328,43 @@ func (h *handler) GetReleaseImage(openshiftVersion, cpuArchitecture string) (*mo
...
+				if err != nil {
+					return err
+				}
+				ocpVersions = append(ocpVersions, minorVersion)
+			}
+		}
+	}
+	h.log.Debugf("For the architecture %s found the following OCP versions: %s", cpuArchitecture, ocpVersions)
+
+	for _, version := range ocpVersions {
```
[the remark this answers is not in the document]
mkowalski: Yeah that's fair. I was trying to overengineer by having the whole list of potential matches available, but in the end this is not at all needed for what we want to achieve here. Thanks!

mkowalski:
/retest

LSO strikes again

internal/versions/versions_test.go
```diff
@@ -592,6 +592,62 @@ var _ = Describe("list versions", func() {
...
+		})
+		It("validates successfuly using major.minor", func() {
+			err = h.ValidateReleaseImageForRHCOS("4.11", common.X86CPUArchitecture)
+			Expect(err).ShouldNot(HaveOccurred())
+		})
+		It("validates successfuly using major.minor using default architecture", func() {
+			err = h.ValidateReleaseImageForRHCOS("4.11", "")
+			Expect(err).ShouldNot(HaveOccurred())
+		})
+		It("validates successfuly using major.minor-something", func() {
```
carbonin, line 631: Did you mean this name to by `major.minor.patch-something` or did we actually want to test `4.12-chocobomb`?

mkowalski: Honestly, not a difference because the core of the test is that both `4.12.2-chocobomb` and `4.12-chocobomb` fold into `4.12` and we check if we have release image for `4.12` available

mkowalski: Let me fix the name then. `4.12-chocobomb` is not natural and we should not encounter it anywhere in the wild

carbonin (approved):

PR #4485, opened 2022-10-11
MGMT-12056: changing default network type kube api

danmanor:

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/clusterdeployments_controller.go
+++ b/internal/controller/controllers/clusterdeployments_controller.go
@@ -95,6 +95,8 @@ type ClusterDeploymentsReconciler struct {
 	AuthType auth.AuthType
 }
 
+const minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes = "4.12.0-0.0"
+
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;update;create
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
 // +kubebuilder:rbac:groups=hive.openshift.io,resources=clusterdeployments,verbs=get;list;watch;create;update;patch;delete
@@ -779,7 +781,7 @@ func (r *ClusterDeploymentsReconciler) updateIgnitionInUpdateParams(ctx context.
 
 func (r *ClusterDeploymentsReconciler) updateNetworkParams(clusterDeployment *hivev1.ClusterDeployment,
 	clusterInstall *hiveext.AgentClusterInstall,
-	cluster *common.Cluster, params *models.V2ClusterUpdateParams) bool {
+	cluster *common.Cluster, params *models.V2ClusterUpdateParams) (*bool, error) {
 	update := false
 	updateString := func(new, old string, target **string) {
 		if new != old {
@@ -837,8 +839,12 @@ func (r *ClusterDeploymentsReconciler) updateNetworkParams(clusterDeployment *hi
 
 	if clusterInstall.Spec.Networking.NetworkType != "" {
 		updateString(clusterInstall.Spec.Networking.NetworkType, swag.StringValue(cluster.NetworkType), &params.NetworkType)
-	} else {
-		updateString(selectClusterNetworkType(params, cluster), swag.StringValue(cluster.NetworkType), &params.NetworkType)
+	} else if !clusterDeployment.Spec.Installed {
+		desiredNetworkType, err := selectClusterNetworkType(params, cluster)
+		if err != nil {
+			return nil, err
+		}
+		updateString(swag.StringValue(desiredNetworkType), swag.StringValue(cluster.NetworkType), &params.NetworkType)
 	}
 
 	// Update APIVIP and IngressVIP only if cluster is not SNO or VipDhcpAllocation is not enabled
@@ -854,7 +860,7 @@ func (r *ClusterDeploymentsReconciler) updateNetworkParams(clusterDeployment *hi
 	if userManagedNetwork := isUserManagedNetwork(clusterInstall); userManagedNetwork != swag.BoolValue(cluster.UserManagedNetworking) {
 		params.UserManagedNetworking = swag.Bool(userManagedNetwork)
 	}
-	return update
+	return swag.Bool(update), nil
 }
 
 func (r *ClusterDeploymentsReconciler) updateIfNeeded(ctx context.Context,
@@ -877,8 +883,11 @@ func (r *ClusterDeploymentsReconciler) updateIfNeeded(ctx context.Context,
 	updateString(spec.ClusterName, cluster.Name, &params.Name)
 	updateString(spec.BaseDomain, cluster.BaseDNSDomain, &params.BaseDNSDomain)
 
-	update = r.updateNetworkParams(clusterDeployment, clusterInstall, cluster, params) || update
-
+	shouldUpdateNetworkParams, err := r.updateNetworkParams(clusterDeployment, clusterInstall, cluster, params)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to update network params")
+	}
+	update = swag.BoolValue(shouldUpdateNetworkParams) || update
 	// Trim key before comapring as done in RegisterClusterInternal
 	sshPublicKey := strings.TrimSpace(clusterInstall.Spec.SSHPublicKey)
 	updateString(sshPublicKey, cluster.SSHPublicKey, &params.SSHPublicKey)
@@ -956,7 +965,7 @@ func (r *ClusterDeploymentsReconciler) updateIfNeeded(ctx context.Context,
 	return clusterAfterUpdate, nil
 }
 
-func selectClusterNetworkType(params *models.V2ClusterUpdateParams, cluster *common.Cluster) string {
+func selectClusterNetworkType(params *models.V2ClusterUpdateParams, cluster *common.Cluster) (*string, error) {
 	clusterWithNewNetworks := &common.Cluster{
 		Cluster: models.Cluster{
 			ClusterNetworks: cluster.ClusterNetworks,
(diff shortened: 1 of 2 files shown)


swagger.yaml
```diff
@@ -4294,7 +4294,7 @@ definitions:
         type: string
         description: "The desired network type used."
         enum: ['OpenShiftSDN', 'OVNKubernetes']
-        default: 'OpenShiftSDN'
+        x-nullable: true
```
omertuc, line 4297: Not sure how I feel about this. This kinda feels like a breaking change - previously the default value was explicitly `OpenShiftSDN` and users could rely on it always being `OpenShiftSDN` when they omit this parameter. Suddenly it can take an entirely different value.

But actually we've already broke that contract once we've started forcing OVNKubernetes on Single Node installations, so it's not really this PR's fault, this PR is just being explicit about us breaking it instead of having the swagger lie...

internal/bminventory/inventory.go
```diff
@@ -340,9 +341,35 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
 
+	params.NewClusterParams.NetworkType, err = getDefaultNetworkType(params)
+	if err != nil {
+		return params, err
+	}
+
 	return params, nil
 }
 
+// If the cluster is SNO or the OpenShift version >= 4.12.0-0.0,
```
omertuc, line 352: Please adhere to [godoc](https://go.dev/blog/godoc) documentation format, specifically `Notice this comment is a complete sentence that begins with the name of the element it describes`

internal/bminventory/inventory.go
```diff
@@ -340,9 +341,35 @@ func (b *bareMetalInventory) setDefaultRegisterClusterParams(ctx context.Context
...
+	if params.NewClusterParams.NetworkType != nil {
+		return params.NewClusterParams.NetworkType, nil
+	}
+
+	isOpenShiftVersionRecentEnough, err := common.VersionGreaterOrEqual(swag.StringValue(params.NewClusterParams.OpenshiftVersion), minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes)
+	if err != nil {
+		return nil, err
+	}
+
+	isSingleNodeCluster := swag.StringValue(params.NewClusterParams.HighAvailabilityMode) == models.ClusterCreateParamsHighAvailabilityModeNone
```
omertuc, line 364: I believe we already had logic (somewhere) to default to OVN when the cluster is a single node cluster, I would expect that this PR would remove / modify said code, just as you did with the controller, but I see you're only adding new code here. Am I missing something?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -877,8 +883,11 @@ func (r *ClusterDeploymentsReconciler) updateIfNeeded(ctx context.Context,
 	updateString(spec.ClusterName, cluster.Name, &params.Name)
 	updateString(spec.BaseDomain, cluster.BaseDNSDomain, &params.BaseDNSDomain)
 
-	update = r.updateNetworkParams(clusterDeployment, clusterInstall, cluster, params) || update
-
+	shouldUpdateNetworkParams, err := r.updateNetworkParams(clusterDeployment, clusterInstall, cluster, params)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse cluster OpenShift version")
```
omertuc, line 888: This error seems overly specific and includes information that I would expect to be present in the error being wrapped, the wrapping error should instead provide context on what was being done while the underlying error occurred, in this case something like "failed to update network params"

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -975,15 +984,20 @@ func selectClusterNetworkType(params *models.V2ClusterUpdateParams, cluster *com
 		clusterWithNewNetworks.MachineNetworks = params.MachineNetworks
 	}
 
+	isOpenShiftVersionRecentEnough, err := common.VersionGreaterOrEqual(cluster.OpenshiftVersion, minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes)
+	if err != nil {
+		return nil, err
```
omertuc, line 989: Here it would be nice to wrap the error `return nil, errors.Wrap(err, "failed to parse cluster OpenShift version")`

internal/bminventory/inventory.go
omertuc: Or perhaps that defaulting was only done in the UI and not in the backend?

Could you please confirm?

internal/bminventory/inventory_test.go
```diff
@@ -10018,6 +10018,110 @@ var _ = Describe("TestRegisterCluster", func() {
...
 
+	// Testing the default network type setting during register flow, with various combinations of platforms,
+	// OpenShift version and high availability modes
+	// Note: SNO is supported starting OpenShift v4.8, and currently only supports Platform "none"
+	It("Default Network Type Setting", func() {
+		type clusterParamsForTest struct {
+			OpenShiftVersion     *string
+			HighAvailabilityMode *string
+			DesiredNetworkType   string
+			Platform             models.Platform
```
omertuc, line 10029: I don't see how platform is involved in `getDefaultNetworkType`, why is it being varied here in the test vectors?

internal/bminventory/inventory_test.go
```diff
@@ -10018,6 +10018,110 @@ var _ = Describe("TestRegisterCluster", func() {
...
+			// OpenShift version >= 4.12.0-0.0, SNO => OVNKubernetes
+			{
+				OpenShiftVersion:     swag.String("4.12-rc.1"),
+				HighAvailabilityMode: swag.String(models.ClusterCreateParamsHighAvailabilityModeNone),
+				DesiredNetworkType:   models.ClusterCreateParamsNetworkTypeOVNKubernetes,
+				Platform:             models.Platform{Type: models.NewPlatformType(models.PlatformTypeNone)},
+			},
+		}
+
+		for _, t := range tests {
```
omertuc, line 10105: If you add a `Name` parameter to `clusterParamsForTest` (initialized with what is currently simple `//` comments) and put the `It(...)` call inside this loop body, you can pass `t.Name` to `It` and so it would show up as nicer separate tests in the unit test reports

internal/bminventory/inventory_test.go
```diff
@@ -10018,6 +10018,110 @@ var _ = Describe("TestRegisterCluster", func() {
...
 	})
 
+	// Testing the default network type setting during register flow, with various combinations of platforms,
+	// OpenShift version and high availability modes
+	// Note: SNO is supported starting OpenShift v4.8, and currently only supports Platform "none"
+	It("Default Network Type Setting", func() {
+		type clusterParamsForTest struct {
+			OpenShiftVersion     *string
+			HighAvailabilityMode *string
+			DesiredNetworkType   string
```
omertuc, line 10028: Personal preference: keep desired parameters at the end of the struct, to make it more clear what is the test input and what is the test's expected output

internal/bminventory/inventory_test.go
```diff
@@ -11548,7 +11659,7 @@ var _ = Describe("TestRegisterCluster", func() {
 				reply := bm.V2RegisterCluster(authCtx, installer.V2RegisterClusterParams{
 					NewClusterParams: &models.ClusterCreateParams{
 						Name:             swag.String("some-cluster-name"),
-						OpenshiftVersion: swag.String("some-version-doesnt-matter-because-we-mock"),
+						OpenshiftVersion: swag.String("4.12.0-someFakeFlavour"),
```
omertuc, line 11662: Can we maybe have a global test var called `var arbitraryOCPVersion := swag.String("4.12.0-someFakeFlavour")`? for all those tests you modified that don't particularly care about the OpenShift version but must specify it because it's required. It reduces mental overhead when reading those tests, avoiding questions like "Why are they using 4.12.0 specifically here?" etc

omertuc:
Please add a detailed commit message explaining:

1) in short, what you're changing (e.g. the defaulting behavior of the network type parameter in the REST API cluster create params and the defaulting behavior of the network type for the CD/ACI controller)

2) And if it's not obvious, go into detail about *why* you're changing that. e.g. what's so special about 4.12 that requires default OVN network type? Assume the person needing to understand this doesn't have access to Jira, feel free to copy that information from Jira if it's actually still relevant (lots of time Jira tickets are outdated since we figure out a lot of things / new information during implementation so be careful and don't just copy as-is)

EDIT: While writing this message I see you've removed the REST API changes and left only the controller changes. Why? Please don't disregard my comments from the portions you removed if you're gonna move them to a new PR.

If your PR is not in a final / presentable state, please convert it to draft and prefix the title with `WIP: ` so reviewers know this is just a rough draft

swagger.yaml
danmanor: I only added changes to internal/controller/controllers/clusterdeployments_controller.go and internal/controller/controllers/clusterdeployments_controller_test.go, sorry for that

osherdp:
/retest 
failures seems like infra flakes

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -95,6 +95,8 @@ type ClusterDeploymentsReconciler struct {
 	AuthType auth.AuthType
 }
 
+const minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes = "4.12.0-0.0"
```
osherdp, line 98: I think we have this exact variable defined somewhere else

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -835,10 +837,14 @@ func (r *ClusterDeploymentsReconciler) updateNetworkParams(clusterDeployment *hi
 		}
 	}
 
-	if clusterInstall.Spec.Networking.NetworkType != "" {
+	if swag.StringValue(cluster.Status) != "installed" && clusterInstall.Spec.Networking.NetworkType != "" {
```
osherdp, line 840: you can use
https://github.com/openshift/assisted-service/blob/1f7410a2d43220908c4ef1b136d2075ac284ec96/models/cluster.go#L1077

osherdp (changes requested):
partial review

eranco74:
/test edge-e2e-metal-assisted-capi
This test shold pass now

osherdp (approved):

osherdp:
capi failure seems legit (installation succeed, conformance suite fails because of DNS errors which is an existing bug)

PR #4501, opened 2022-10-14
Bug 2107715: Add validation for Agent's role

mkowalski:
This PR adds validation for the Agent's role whenever an update is requested. This is in order to prevent configuration where an incorrect role causes the Node to try to download a non-existing ignition file and fail the installation without a clear error returned to the end user.

With this change, as soon as an attempt to set an invalid role is detected, an API error will be returned preventing the change to be applied (in the REST API flow) or causing Agent CR to return a SpecSync error (in the Kube API flow).

Closes: [MGMTBUGSM-473](https://issues.redhat.com//browse/MGMTBUGSM-473)

/cc @omertuc

### DIFF
--- a/internal/host/host_test.go
+++ b/internal/host/host_test.go
@@ -74,7 +74,7 @@ var _ = Describe("update_role", func() {
 		common.DeleteTestDB(db, dbName)
 	})
 
-	Context("update role by src state", func() {
+	Context("by src state", func() {
 		success := func(srcState string) {
 			host = hostutil.GenerateTestHost(id, infraEnvID, clusterID, srcState)
 			Expect(db.Create(&host).Error).ShouldNot(HaveOccurred())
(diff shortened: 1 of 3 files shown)


internal/host/host_test.go
```diff
@@ -192,17 +192,73 @@ var _ = Describe("update_role", func() {
...
+				Expect(state.UpdateRole(ctx, &h.Host, models.HostRoleWorker, nil)).NotTo(HaveOccurred())
+				h = hostutil.GetHostFromDB(id, infraEnvID, db)
+				Expect(h.Role).To(Equal(models.HostRoleWorker))
+			})
+			It("to bootstrap successfuly", func() {
+				Expect(state.UpdateRole(ctx, &h.Host, models.HostRoleBootstrap, nil)).NotTo(HaveOccurred())
+				h = hostutil.GetHostFromDB(id, infraEnvID, db)
+				Expect(h.Role).To(Equal(models.HostRoleBootstrap))
+			})
+			It("to chocobomb with failure", func() {
```
omertuc, line 222: this test is not future-proof, what if someone adds a chocobomb role in the future? this test will break!

omertuc: Your test should instead iterate all possible roles using reflection, then dynamically generate a role that necessarily doesn't exist, this way you can save this theoretical future chocobomb dev a few seconds of fixing unit tests

omertuc: (i sure hope you don't think I was serious in any of my messages)

But in this case the unit-test is breaking for irrelevant reasons. There's nothing wrong with adding a chocobomb role.Yet the unit-test failed. The developer has become aware of the test, but should they have been aware of it?

mkowalski: Let me step in the shoes of a theoretical dev, Joshua, from the future who tries to add a chocobomb role. We will walk through a proof that existence of this test is only of a benefit...

Joshua decides that chocobomb is a valid role for all the flows, so he goes to the function IsRoleValid and does his stuff. At this moment if Joshua is not experienced enough, he just commits the code and prays he gets LTGM without anyone noticing he missed the unit test.

If Joshua is an experienced developer, he will immediately start writing a unit test to cover the fact that chocobomb is now allowed role. At this moment he notices that I created a test for the same role and just piggybacks on it by modifying the condition. In this case my test helped Joshua because he doesn't need to write a new case from scratch, he can just edit the test that I have created.

So by the reasoning above I have proven that existence of the chocobomb test can be only a benefit. For inexperienced Joshua it will be a learning experience as his build will fail and force him to at least look at unit test (and hopefully remember for the future that those are a must). For experienced Joshua it will save him something between 1 up to 2 lines of code. By some quick calculation this is a save of around 3 kg of CO2 emission but I leave the calculation to the reader as it's not relevant to the proof.

To sum up, we have proven that existence of a test for chocobomb role can be only of a help.

mkowalski:
The PR has been manually tested using Infrastructure Operator

```
# AGENT=1fc01fac-2030-4121-b3bc-dc91595aba01; oc patch agent -n assisted-installer $AGENT -p '{"spec":{"role":"chocobomb"}}' --type merge
agent.agent-install.openshift.io/1fc01fac-2030-4121-b3bc-dc91595aba01 patched

# oc get agent -n assisted-installer 1fc01fac-2030-4121-b3bc-dc91595aba01 -oyaml
[...]
status:
  conditions:
  - lastTransitionTime: "2022-10-14T10:59:24Z"
    message: 'The Spec could not be synced due to backend error: Requested role (chocobomb)
      is invalid for host 1fc01fac-2030-4121-b3bc-dc91595aba01 from infraEnv 3d4bbdbf-5cee-4ec7-8038-972eaa830a9f'
    reason: BackendError
    status: "False"
    type: SpecSynced
```

PR #4513, opened 2022-10-18
MGMT-12261: Respect `InstallConfigOverrides` when `getCAContents` is non-empty

omertuc:
# Background

Users can influence the install-config `AdditionalTrustBundle` through `InstallConfigOverrides`.

This generally works. Except for in the situation described below.

# Issue

`getCAContents` is a function that in some circumstances returns
internal CA certificates that need to be set in the install-config
`AdditionalTrustBundle`.

In most cases, `getCAContents` returns an empty string and everything
works fine.

But when `getCAContents` actually returns certificates, then we
completely overwrite the install-config `AdditionalTrustBundle`,
even when the user modified it through `InstallConfigOverrides`.

This is bad.

The situations in which `getCAContents` is non-empty is when the
Assisted Service admin deploys it with either:

- Red Hat Root CA (`INSTALL_RH_CA` env)
    - This is the case in our staging environment and this is the cause
      for the issues seen in MGMT-12261
- Mirror registry configuration mounts, as detected by our mirrorregistries package
    - I'm not aware of anyone that ran into this particular issue, but
      it theoretically could happen

# Solution

We should instead merge the `InstallConfigOverrides` certs with the certs we
receive from `getCAContents`, rather than overwrite them.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [x] Unit tests

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/installcfg/builder/builder.go
+++ b/internal/installcfg/builder/builder.go
@@ -187,7 +187,13 @@ func (i *installConfigBuilder) getInstallConfig(cluster *common.Cluster, addRhCa
 	}
 	caContent := i.getCAContents(cluster, ca, addRhCa)
 	if caContent != "" {
-		cfg.AdditionalTrustBundle = fmt.Sprintf(` | %s`, caContent)
+		// TODO: This | symbol is actually useless, it'll be removed in a future separate PR that's already WIP
+		if cfg.AdditionalTrustBundle == "" {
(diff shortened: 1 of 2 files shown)


osherdp:
/test edge-subsystem-aws
known flake

osherdp (approved):

PR #4523, opened 2022-10-20
NO-ISSUE: Update doc for per-customer feature enablement

mkowalski:
/cc @jhernand 
/cc @osherdp

### DIFF
--- a/docs/dev/feature-per-customer.md
+++ b/docs/dev/feature-per-customer.md
@@ -88,3 +88,7 @@ if !hasMultiarchAuthorization {
 ```
 
 The full PR containing this example and implementing all the features mentioned above is [Allow org-based filtering for multiarch images](https://github.com/openshift/assisted-service/pull/4368).
+
+## Enabling feature gates in SaaS
+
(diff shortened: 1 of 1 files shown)


docs/dev/feature-per-customer.md
```diff
@@ -88,3 +88,7 @@ if !hasMultiarchAuthorization {
 ```
 
 The full PR containing this example and implementing all the features mentioned above is [Allow org-based filtering for multiarch images](https://github.com/openshift/assisted-service/pull/4368).
+
+## Enabling feature gates in SaaS
+
+In order for the service to use all the logic described above, it needs to run with the setting `ENBALE_ORG_BASED_FEATURE_GATES` enabled. This is controlled by the environment variables passed to the container. For our SaaS offering that would mean you need to make sure that `data/services/assisted-installer/cicd/saas.yaml` in the [app-interface](https://gitlab.cee.redhat.com/service/app-interface) repository has the variable enabled (as it defaults to disabled).
```
osherdp, line 94:
```suggestion
In order for the service to use all the logic described above, it needs to run with the setting `ENABLE_ORG_BASED_FEATURE_GATES` enabled. This is controlled by the environment variables passed to the container. For our SaaS offering that would mean you need to make sure that `data/services/assisted-installer/cicd/saas.yaml` in the [app-interface](https://gitlab.cee.redhat.com/service/app-interface) repository has the variable enabled (as it defaults to disabled).
```

osherdp (changes requested):

jhernand:
Very nice @mkowalski, but you will have to fix my spelling error as suggested by @osherdp, sorry for that.

osherdp (approved):

PR #4529, opened 2022-10-20
MGMT-12312: Adds node-boot log type

nmagnezi:
Per https://github.com/openshift/assisted-service/pull/4444, add a new log type named: `node-boot`.

This change updates both `V2UploadLogs` and `V2DownloadClusterLogs` to accept the new log type.
In addition:
1. A new event was added: `host_boot_logs_uploaded`.
2. Defined download name: `<cluster_name>_<host_role>_boot_<host_id>.tar.gz`.
3. Note that `V2UploadLogs` with log type `node-boot` will not update the host progress.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/vendor/github.com/openshift/assisted-service/models/logs_type.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/logs_type.go
@@ -33,6 +33,9 @@ const (
 	// LogsTypeHost captures enum value "host"
 	LogsTypeHost LogsType = "host"
 
+	// LogsTypeNodeBoot captures enum value "node-boot"
+	LogsTypeNodeBoot LogsType = "node-boot"
+
 	// LogsTypeController captures enum value "controller"
 	LogsTypeController LogsType = "controller"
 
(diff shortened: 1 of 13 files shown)


nmagnezi:
this is WIP

carbonin:
Looks like there's still quite a few questions in the enhancement.
The code looks good though. I'm curious about "boot logs" as a name though since we boot the host at least twice.

Maybe we need a more specific name since we're putting it in a user-facing event.
Maybe just saying `node boot` rather than `host` would be better?
Something like `post install boot logs` might work but is a bit confusing because we're done installing coreos, but the node isn't quite done yet.

@avishayt what do you think?

avishayt:
Agree with @carbonin .  node-boot-logs is better than host-boot-logs, maybe boot-from-disk-logs is more explicit?

nmagnezi:
> Agree with @carbonin . node-boot-logs is better than host-boot-logs, maybe boot-from-disk-logs is more explicit?

@carbonin, @avishayt,
I changed the log name to `node-boot`

avishayt:
> 

Excellent,  thanks

PR #4530, opened 2022-10-21
MGMT-12314: Fix API IP detection for day2 ZTP CI

mkowalski:
This commit fixes the logic of API IP detection and libvirt DNS entries that are created in a scenario when spoke is a multi-node without UMN.

Currently for such a scenario the DNS entry for Spoke's API uses the default value of 192.168.111.1 which is incorrect. Because of that, the kubeconfig for such a Spoke cluster is unusable.

With this change, if there is no dedicated loadbalancer deployed, the API IP will be detected based on the `.status.apiVIP` of the AgentClusterInstall. This will help us to get the day2 CI back on track.

Closes: [MGMT-12314](https://issues.redhat.com//browse/MGMT-12314)

/test edge-e2e-ai-operator-ztp-compact-day2-workers
/cc @vrutkovs


vrutkovs:
LGTM

vrutkovs (approved):
/lgtm

PR #4534, opened 2022-10-24
MGMT-11750: ensure namespace exists on spoke cluster (HASC)

danielerez:
This PR handles the following:
 * Create namespace on spoke cluster to ensure it exists (the namespace of HASC resource).
 * Create assisted-service service-account on both hub and spoke clusters.
 * Restore client on AgentServiceConfigReconcileContext upon finishing reconcile (as there's a single context).

Note: to avoid switching the client property between hub and spoke, client should be extracted from AgentServiceConfigReconcileContext. That way, we could avoid sharing it and just pass it as needed.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/hypershiftagentserviceconfig_controller.go
+++ b/internal/controller/controllers/hypershiftagentserviceconfig_controller.go
@@ -31,7 +31,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -40,7 +39,6 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 const (
@@ -65,6 +63,15 @@ func (asc *ASC) initHASC(r *HypershiftAgentServiceConfigReconciler, instance *ai
 	asc.Object = instance
 	asc.spec = &instance.Spec.AgentServiceConfigSpec
 	asc.conditions = &instance.Status.Conditions
+	context := r.AgentServiceConfigReconcileContext
+	asc.rec = &AgentServiceConfigReconcileContext{
+		Client:       context.Client,
+		Log:          context.Log,
+		Scheme:       context.Scheme,
+		NodeSelector: context.NodeSelector,
+		Tolerations:  context.Tolerations,
+		Recorder:     context.Recorder,
+	}
 }
 
 var assistedServiceRBAC_l0 = []component{
@@ -77,7 +84,6 @@ var assistedServiceRBAC_l1 = []component{
 	{"AssistedServiceRoleBinding_l1", aiv1beta1.ReasonRBACConfigurationFailure, newAssistedServiceRoleBinding},
 	{"AssistedServiceClusterRole_l1", aiv1beta1.ReasonRBACConfigurationFailure, newAssistedServiceClusterRole},
 	{"AssistedServiceClusterRoleBinding_l1", aiv1beta1.ReasonRBACConfigurationFailure, newAssistedServiceClusterRoleBinding},
-	{"AssistedServiceServiceAccount_l1", aiv1beta1.ReasonRBACConfigurationFailure, newAssistedServiceServiceAccount},
 }
 
 // Adding required resources to rbac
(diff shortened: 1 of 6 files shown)


internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -128,8 +122,44 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
...
+	// Ensure instance namespace exists on spoke cluster
+	if err := hr.ensureSpokeNamespace(ctx, spokeClient, instance.Namespace); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Switch to spoke client in the reconciler context.
+	// We need to store the hub client as ASC stores a reference to AgentServiceConfigReconcileContext.
+	hubClient := asc.rec.Client
+	defer func() {
+		asc.rec.Client = hubClient
```
carbonin, line 142: Why does this matter? `asc` is a local variable scoped to this `Reconcile` function call. Why would we care about the state after this call?

danielerez: asc is indeed a local env, but we change Client in rec - which is a pointer to AgentServiceConfigReconcileContext in the reconciler.

slaviered: Then this would not be changed because reconcile can be called in parallel

danielerez: Do we really need to worry about this if we're using the default 1 concurrent Reconciles?
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controller.go#L39

danielerez: ok, created a new instance of AgentServiceConfigReconcileContext in initHASC, so it should mitigate this issue.

slaviered: Yes. the defensive copy should be a good solution path

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -267,6 +309,18 @@ func (hr *HypershiftAgentServiceConfigReconciler) getAgentInstallCRDs(ctx contex
 	return crds, nil
 }
 
+func (hr *HypershiftAgentServiceConfigReconciler) ensureSpokeNamespace(ctx context.Context, c client.Client, namespace string) error {
```
carbonin, line 312: Strange that this has a client as a parameter, but we have the spoke client cache as a field on `HypershiftAgentServiceConfigReconciler`.

I'd say either use the spoke client cache in `hr` or pass the log and make this a function rather than a method.

internal/controller/controllers/hypershiftagentserviceconfig_controller_test.go
```diff
@@ -366,3 +403,27 @@ var _ = Describe("HypershiftAgentServiceConfig reconcile", func() {
 		})
 	})
 })
+
+type fakeSpokeK8sClient struct {
```
carbonin, line 407: This is unfortunate. Is there no way to avoid this?

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -128,8 +131,41 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
...
+	}
+
+	// Ensure instance namespace exists on spoke cluster
+	if err := hr.ensureSpokeNamespace(ctx, spokeClient, instance.Namespace); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Switch to spoke client in the reconciler context
+	// TODO: extract client from AgentServiceConfigReconcileContext to avoid sharing
+	//       it with both hub and spoke clients.
```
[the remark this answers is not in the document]
slaviered: However, please note that we are not working with 2 clients simultaneously. alternating between clients enables us to re-use most of the AC code. All we need to check is that we are not saving state between calls

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -128,8 +131,41 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
...
+	if result, err := hr.reconcileSpokeComponents(ctx, log, asc); err != nil {
+		return result, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (hr *HypershiftAgentServiceConfigReconciler) reconcileHubComponents(ctx context.Context, log *logrus.Entry, asc ASC) (ctrl.Result, error) {
+	hubComponents := []component{}
+	hubComponents = append(hubComponents, assistedServiceRBAC_l0...)
```
slaviered, line 162: This kind of component aggregation will not scale well with the additional rbac and webhook mechanism that I will soon add and with additional components that we will add in the future. I suggest having a clear component separation based on categories and then aggregate them once.

internal/controller/controllers/hypershiftagentserviceconfig_controller_test.go
slaviered: Is there a way to generate a mock like we do with our own code? (with mockgen?)

danielerez: Haven't found such :(

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
danielerez: sure, will handle in consecutive PRs then, it's a bit out of the scope here...

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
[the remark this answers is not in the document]
danielerez: well, the alternative is to pass the needed client separately. but that's just a todo to rethink later on.

slaviered:
LGTM-ing on the grounds that we need to submit a PR after this one that fix the component comment
/lgtm

danielerez:
/override ci/prow/images
(not related to changes)

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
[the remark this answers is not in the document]
carbonin: > we are not working with 2 clients simultaneously

You are if you reconcile in parallel and share the state in the reconciler.

> alternating between clients enables us to re-use most of the AC code

My point is that you can still reuse the code without using a single client reference in this shared context.

PR #4541, opened 2022-10-25
MGMT-12053: Add support for additional kernel arguments when serving iPXE script

ori-amizur:

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @filanov

### DIFF
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -9253,6 +9253,27 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
 		return body
 	}
 
+	strArrayToKernelArguments := func(args []string) (kargs models.KernelArguments) {
+		for _, s := range args {
+			kargs = append(kargs, &models.KernelArgument{
+				Operation: models.KernelArgumentOperationAppend,
+				Value:     s,
+			})
+		}
+		return
+	}
+
+	updateKernelArgs := func(kernelArgs ...string) {
+		params := installer.UpdateInfraEnvParams{
+			InfraEnvUpdateParams: &models.InfraEnvUpdateParams{},
+			InfraEnvID:           infraEnvID,
+		}
+		params.InfraEnvUpdateParams.DiscoveryKernelArguments = append(params.InfraEnvUpdateParams.DiscoveryKernelArguments, strArrayToKernelArguments(kernelArgs)...)
+		response := bm.UpdateInfraEnv(ctx, params)
+		_, ok := response.(*installer.UpdateInfraEnvCreated)
+		Expect(ok).To(BeTrue())
+	}
+
 	It("should ensure the correct discovery iso type is passed to the ignition builder", func() {
 		mockIgnitionBuilder.EXPECT().FormatDiscoveryIgnitionFile(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), "").Return(discovery_ignition_3_1, nil).Times(1)
 		body := getResponseData("discovery.ign", false, nil, "")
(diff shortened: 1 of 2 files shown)


internal/bminventory/inventory_test.go
```diff
@@ -9313,6 +9324,79 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
...
+		mockVersions.EXPECT().GetOsImageOrLatest(common.TestDefaultConfig.OpenShiftVersion, gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).Times(3)
+		content := getResponseData("ipxe-script", false, nil, "")
+		lines := strings.Split(string(content), "\n")
+
+		Expect(lines[0]).To(Equal("#!ipxe"))
+
+		By("validating the kernel line")
+		kernelRegex := regexp.MustCompile(`^kernel (\S+) initrd=initrd coreos.live.rootfs_url=(\S+) (.+)`)
+		match := kernelRegex.FindStringSubmatch(lines[2])
+		fmt.Fprintf(GinkgoWriter, "lines[2]: %s\n", lines[2])
```
filanov, line 9342: is it for debug?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -672,6 +673,28 @@ func (b *bareMetalInventory) canServeHostIPXEScript(infraEnv *common.InfraEnv, m
...
+	return ret, nil
+}
+
+func kernelArgsAppendStr(infraEnv *common.InfraEnv) (string, error) {
+	kernelArguments, err := kernelArgsToSlice(infraEnv)
+	if err != nil {
+		return "", err
+	}
+	if len(kernelArguments) > 0 {
+		return " " + strings.Join(kernelArguments, " "), nil
```
filanov, line 693: why do you need the space at the start?

ori-amizur: To separate from the existing line

internal/bminventory/inventory_test.go
```diff
@@ -9340,6 +9351,46 @@ var _ = Describe("V2DownloadInfraEnvFiles", func() {
...
+		lines := strings.Split(string(content), "\n")
+
+		Expect(lines[0]).To(Equal("#!ipxe"))
+
+		By("validating the kernel line")
+		kernelRegex := regexp.MustCompile(`^kernel (\S+) initrd=initrd coreos.live.rootfs_url=(\S+) (.+)`)
+		match := kernelRegex.FindStringSubmatch(lines[2])
+		Expect(match).NotTo(BeNil())
+
+		Expect(match[3]).To(Equal(initialKernelArguments))
```
carbonin, line 9372: I don't think you need to test the initial args as we're already testing that in other places.

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -672,6 +672,28 @@ func (b *bareMetalInventory) canServeHostIPXEScript(infraEnv *common.InfraEnv, m
 	return nil
 }
 
+func kernelArgsToSlice(infraEnv *common.InfraEnv) ([]string, error) {
```
carbonin, line 675: Are we going to need this elsewhere? Maybe this would be a good method to define on `common.InfraEnv` directly?

Marbe `infraEnv.UnmarshalKernelArgs()` or something.

internal/common/common.go
```diff
@@ -437,3 +438,32 @@ func ApplyYamlPatch(src []byte, ops []byte) ([]byte, error) {
...
+	var args []string
+	for _, arg := range kargs {
+		if arg.Operation != models.KernelArgumentOperationAppend {
+			return nil, errors.Errorf("only '%s' operation is allowed.  got '%s'", models.KernelArgumentOperationAppend,
+				arg.Operation)
+		}
+		args = append(args, arg.Value)
+	}
+	return args, nil
+}
```
carbonin, line 469: These are all only used in one place. And some are only used in tests.
That doesn't seem like they're generally useful enough to add to `common`.

Maybe if/when we end up with more callers this will make sense, but let's not add to `common` unless it's very necessary.

internal/bminventory/inventory_v2_handlers.go
filanov: then why not have the space there? it looks strange to have it in this function, or at least not clear why it's needed

ori-amizur: Because I didn't want to add an extra space if no parameter follows it.

ori-amizur: This test https://github.com/openshift/assisted-service/blob/938229b5f6f77c3d0331c18324e2e73534ff9604/internal/bminventory/inventory_test.go#L9325 will probably fail if it will be added unconditionally.

internal/bminventory/inventory_test.go
ori-amizur: OK

internal/bminventory/inventory_v2_handlers.go
ori-amizur: Not sure.  Currently this is the only place it is used.

internal/common/common.go
ori-amizur: Actually I wasn't I wasn't sure where is the best location to put them.
I wanted them to be together in a single place (not spread in several location), although their usage is from several files.
Where do you think is the best location?

internal/bminventory/inventory_test.go
carbonin: It just doesn't seem helpful to me to have two tests fail for exactly the same thing if/when we change the default set of params, but it's fine if you want to keep it.

internal/bminventory/inventory_v2_handlers.go
carbonin: :man_shrugging: maybe just leave it as a part of `kernelArgsAppendStr` then and don't have it as a separate function at all.

internal/common/common.go
carbonin: I think `StrArrayToKernelArguments` should be in `inventory_test.go` as that's the only place it's used.

I think `KernelArgumentsToStrSlice` and `UnmarshalKernelArguments` shouldn't be standalone functions at all.
Just put the content into `kernelArgsToSlice` directly as that's the only place they are used.

internal/bminventory/inventory_v2_handlers.go
ori-amizur: Changed.  Put all the functionality in this function.

internal/bminventory/inventory_test.go
ori-amizur: OK

internal/bminventory/inventory_v2_handlers.go
carbonin: What do you think about going one more step and combining this with `kernelArgsAppendStr`?
`kernelArgsAppendStr` is really only one more call in addition to what's here.

ori-amizur: Seemed to me more logical to separate them because they are doing different things.
Many times I am for function separation.  It is probably a matter for style.
Anyway, I can change if you think it is essential.

carbonin: It's definitely not essential, but my heuristic for separating out a function is usually something like:

- Will the logic be reused in multiple places?
  - No, this function is only called in one place
- Is the logic complicated enough to need to be named to be understood?
  - No, the other function is a single conditional and an append
- Does the logic need its own unit-tests?
  - No, there are no direct unit tests for either of these
  
So in this case it seemed to me that one function would be better.
In fact, this is how I considered all of the functions I commented on in this PR.

But again, it's far from necessary. Clearly everything will work fine without making this change - it's just a suggestion along with some insight into my thought process.

ori-amizur: OK. So I prefer to leave it as is.

carbonin (approved):

PR #4543, opened 2022-10-26
MGMT-12329: Implements assisted boot reporter and add to ignition

nmagnezi:
Implements: https://github.com/openshift/assisted-service/pull/4444

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -395,7 +395,7 @@ func main() {
 	failOnError(err, "failed to create valid bm config S3 endpoint URL from %s", Options.BMConfig.S3EndpointURL)
 	Options.BMConfig.S3EndpointURL = newUrl
 
-	generator := generator.New(log, objectHandler, Options.GeneratorConfig, Options.WorkDir, operatorsManager, providerRegistry, Options.ClusterTLSCertOverrideDir)
+	generator := generator.New(log, objectHandler, Options.GeneratorConfig, Options.WorkDir, operatorsManager, providerRegistry, Options.ClusterTLSCertOverrideDir, authHandler)
 	var crdUtils bminventory.CRDUtils
 	if ctrlMgr != nil {
 		crdUtils = controllers.NewCRDUtils(ctrlMgr.GetClient(), hostApi)
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -3555,6 +3555,7 @@ func (b *bareMetalInventory) getLogFileForDownload(ctx context.Context, clusterI
 	if err != nil {
 		return "", "", err
 	}
+	b.log.Debugf("log type to download: %s", logsType)
 	switch logsType {
 	case string(models.LogsTypeHost), string(models.LogsTypeNodeBoot):
 		if hostId == nil {
--- a/internal/cluster/auth.go
+++ b/internal/cluster/auth.go
@@ -1,6 +1,9 @@
 package cluster
 
 import (
+	"fmt"
+	"reflect"
+
 	"github.com/openshift/assisted-service/internal/cluster/validations"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/gencrypto"
@@ -8,12 +11,27 @@ import (
 	"github.com/pkg/errors"
 )
 
-func AgentToken(infraEnv *common.InfraEnv, authType auth.AuthType) (token string, err error) {
+func AgentToken(resource interface{}, authType auth.AuthType) (token string, err error) {
+	var (
+		pullSecret string
+		resId      string
+	)
+	switch res := resource.(type) {
+	case *common.InfraEnv:
+		resId = res.ID.String()
+		pullSecret = res.PullSecret
+	case *common.Cluster:
+		resId = res.ID.String()
+		pullSecret = res.PullSecret
+	default:
+		return "", fmt.Errorf("unsupported type, expected InfraEnv or Cluster got %s", reflect.TypeOf(resource))
+	}
+
 	switch authType {
 	case auth.TypeRHSSO:
-		token, err = cloudPullSecretToken(infraEnv.PullSecret)
+		token, err = cloudPullSecretToken(pullSecret)
 	case auth.TypeLocal:
-		token, err = gencrypto.LocalJWT(infraEnv.ID.String(), gencrypto.InfraEnvKey)
+		token, err = gencrypto.LocalJWT(resId, gencrypto.InfraEnvKey)
 	case auth.TypeNone:
 		token = ""
 	default:
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -1059,7 +1059,11 @@ func (m *Manager) PrepareClusterLogFile(ctx context.Context, c *common.Cluster, 
 					if hostObject.Bootstrap {
 						role = string(models.HostRoleBootstrap)
 					}
-					tarredFilename = fmt.Sprintf("%s_%s_%s.tar.gz", sanitize.Name(c.Name), role, sanitize.Name(hostutil.GetHostnameForMsg(hostObject)))
+					name := sanitize.Name(hostutil.GetHostnameForMsg(hostObject))
+					if strings.Contains(file, "boot_") {
+						name = fmt.Sprintf("boot_%s", name)
+					}
+					tarredFilename = fmt.Sprintf("%s_%s_%s.tar.gz", sanitize.Name(c.Name), role, name)
 				}
 			} else {
 				tarredFilename = fmt.Sprintf("%s_%s", fileNameSplit[len(fileNameSplit)-2], fileNameSplit[len(fileNameSplit)-1])
--- a/internal/ignition/boot-reporter/assisted-boot-reporter.sh
+++ b/internal/ignition/boot-reporter/assisted-boot-reporter.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+
+export LOG_SEND_FREQUENCY_IN_MINUTES=5
+export SERVICE_TIMEOUT_MINUTES=60
+
+function log() {
+  echo "$(date '+%F %T') ${HOSTNAME} $2[$$]: level=$1 msg=\"$3\""
+}
+
+function log_info() {
+  log "info" "$1" "$2"
+}
+
+function log_error() {
+  log "error" "$1" "$2"
+}
+
+function init_variables() {
+  func_name=${FUNCNAME[0]}
+  init_failed="false"
+
+  if [ "$LOG_SEND_FREQUENCY_IN_MINUTES" == "" ]; then
+    init_failed="true"
+    log_error "${func_name}" "LOG_SEND_FREQUENCY_IN_MINUTES is empty."
+  fi
+
+  if [ "$SERVICE_TIMEOUT_MINUTES" == "" ]; then
+    init_failed="true"
+    log_error "${func_name}" "SERVICE_TIMEOUT_MINUTES is empty."
+  fi
+
+  if [ "$ASSISTED_SERVICE_URL" == "" ]; then
+    init_failed="true"
+    log_error "${func_name}" "ASSISTED_SERVICE_URL is empty."
+  elif [ "${ASSISTED_SERVICE_URL: -1}" == "/" ]; then
+    export ASSISTED_SERVICE_URL="${ASSISTED_SERVICE_URL::-1}"
+  fi
+
+  if [ "$CLUSTER_ID" == "" ]; then
+    init_failed="true"
+    log_error "${func_name}" "CLUSTER_ID is empty."
+  fi
+
+  if [ "$HOST_ID" == "" ]; then
+    init_failed="true"
+    log_error "${func_name}" "HOST_ID is empty."
+  fi
+
+  if [ "$init_failed" == "true" ]; then
+    log_error "${func_name}" "Failed to initialize variables. Exiting."
+    exit 1
+  fi
+}
+
+function collect_and_upload_logs() {
+  func_name=${FUNCNAME[0]}
+
+  log_info "${func_name}" "Collecting logs."
+  logs_dir_name=boot_logs_$HOST_ID
+  logs_path=/tmp/$logs_dir_name
+
+  rm -rf "$logs_path"
+  mkdir -p "$logs_path"
+
+  log_info "${func_name}" "Copying journalctl to $logs_path/"
+  journalctl > "$logs_path"/journalctl.log
+  log_info "${func_name}" "Capturing the output of 'ip a' to $logs_path/"
+  ip a > "$logs_path"/ip_a.log
+  log_info "${func_name}" "Copying /etc/resolv.conf to $logs_path/"
+  cp /etc/resolv.conf "$logs_path"
+  log_info "${func_name}" "Copying /var/log/pods to $logs_path/"
+  cp -r /var/log/pods "$logs_path"
+  log_info "${func_name}" "Compressing logs to $logs_path.tar.gz"
+  tar -czvf "$logs_path".tar.gz -C /tmp "$logs_dir_name"
+
+  log_info "${func_name}" "Uploading logs."
+
+  if curl -X POST  -H "X-Secret-Key: ${PULL_SECRET_TOKEN}" \
+       -F upfile=@"$logs_path".tar.gz \
+       "$ASSISTED_SERVICE_URL/api/assisted-install/v2/clusters/$CLUSTER_ID/logs?logs_type=node-boot&host_id=$HOST_ID" ; then
+      log_info "${func_name}" "Successfully uploaded logs."
+  else
+      log_error "${func_name}" "Failed to upload logs."
+  fi
+}
+
+function main() {
+  func_name=${FUNCNAME[0]}
+  count=$((SERVICE_TIMEOUT_MINUTES/LOG_SEND_FREQUENCY_IN_MINUTES))
+
+  for i in $(seq $count)
+  do
+      log_info "${func_name}" "Upload logs attempt ${i}/${count}"
+      collect_and_upload_logs
+      if [ "$i" != "$count" ]; then # don't sleep at the last iteration.
+        log_info "${func_name}" "Sleeping for ${LOG_SEND_FREQUENCY_IN_MINUTES} minutes until the next attempt."
+        sleep $((LOG_SEND_FREQUENCY_IN_MINUTES*60))
+      fi
+  done
+}
+
+log_info assisted-boot-reporter "assisted-boot-reporter start"
+init_variables
+main
+log_info assisted-boot-reporter "assisted-boot-reporter end"
+exit 0
--- a/internal/ignition/ignition.go
+++ b/internal/ignition/ignition.go
@@ -3,6 +3,7 @@ package ignition
 import (
 	"bytes"
 	"context"
+	"embed"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
@@ -56,6 +57,9 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 	k8syaml "sigs.k8s.io/yaml"
 )
+
+//go:embed boot-reporter
+var reporter embed.FS
 
 const (
 	masterIgn = "master.ign"
@@ -191,6 +195,29 @@ const highlyAvailableInfrastructureTopologyPatch = `---
 
 const tempNMConnectionsDir = "/etc/assisted/network"
 
+const bootReporterPath = "boot-reporter/assisted-boot-reporter.sh"
+
+var assistedBootReporterunitTemplate = `[Unit]
+Description=Collect and upload host boot logs to assisted-service
+Wants=network-online.target
+After=network-online.target
+DefaultDependencies=no
+[Service]
+Environment=ASSISTED_SERVICE_URL=%s
+Environment=PULL_SECRET_TOKEN=%s
+Environment=CLUSTER_ID=%s
+Environment=HOST_ID=%s
+Environment=LOG_SEND_FREQUENCY_IN_MINUTES=%d
+Environment=SERVICE_TIMEOUT_MINUTES=%d
+User=root
+Type=oneshot
+ExecStart=/bin/bash /usr/local/bin/assisted-boot-reporter.sh
+PrivateTmp=true
+RemainAfterExit=no
+[Install]
+WantedBy=multi-user.target
+`
+
 var fileNames = [...]string{
 	"bootstrap.ign",
 	masterIgn,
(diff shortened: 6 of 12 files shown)


internal/ignition/ignition.go
```diff
@@ -191,6 +192,28 @@ const highlyAvailableInfrastructureTopologyPatch = `---
 
 const tempNMConnectionsDir = "/etc/assisted/network"
 
+const bootReporterPath = "/assisted-boot-reporter.sh"
```
omertuc, line 195: We can use https://pkg.go.dev/embed instead of including it in the service image

nmagnezi: What makes it preferable here?

omertuc: Consistency. We already use it all over the service (most notably all the files included by `internal/ignition/templates.go` including files in ignition specifically too)

Also it avoids a change to the Dockerfile. We have copies of this Dockerfile everywhere maintained by the Agent team, our team, ACM pipelines, MCE, downstream SaaS, etc etc. So changes to the Dockerfile are super expensive from an engineering perspective because we have to make sure we do them everywhere, and do them right. We often get bug reports and complaints when those Dockerfiles don't get updated correctly.

nmagnezi: Done

internal/bminventory/inventory.go
```diff
@@ -1323,7 +1323,7 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params
 			}
 		}()
 
-		if err = b.generateClusterInstallConfig(asyncCtx, *cluster, clusterInfraenvs); err != nil {
+		if err = b.generateClusterInstallConfig(asyncCtx, *cluster, clusterInfraenvs, b.ServiceBaseURL); err != nil {
```
filanov, line 1326: why do you need to pass `ServiceBaseURL` if it's part of the same object (`bareMetalInventory`)

internal/bminventory/inventory.go
```diff
@@ -1694,7 +1694,7 @@ func (b *bareMetalInventory) generateClusterInstallConfig(ctx context.Context, c
 		installerReleaseImageOverride = *defaultArchImage.URL
 	}
 
-	if err := b.generator.GenerateInstallConfig(ctx, cluster, cfg, *releaseImage.URL, installerReleaseImageOverride); err != nil {
+	if err := b.generator.GenerateInstallConfig(ctx, cluster, cfg, *releaseImage.URL, installerReleaseImageOverride, serviceBaseURL, b.authHandler.AuthType()); err != nil {
```
filanov: base url and auth type both are set once on startup, maybe it should be part of the configuration to the generator?

internal/ignition/ignition.go
```diff
@@ -276,12 +307,14 @@ func NewBuilder(log logrus.FieldLogger, staticNetworkConfig staticnetworkconfig.
 }
 
 // NewGenerator returns a generator that can generate ignition files
-func NewGenerator(workDir string, installerDir string, cluster *common.Cluster, releaseImage string, releaseImageMirror string,
-	serviceCACert, installInvoker string, s3Client s3wrapper.API, log logrus.FieldLogger, operatorsApi operators.API,
+func NewGenerator(db *gorm.DB, serviceBaseURL string, workDir string, installerDir string, cluster *common.Cluster, releaseImage string, releaseImageMirror string,
```
filanov, line 310: with so many parameters maybe we should consider having a configuration struct?

nmagnezi: Indeed a lot of parameters, however, all this function is doing is constructing those parameters to a `generator` struct and returning that struct.

nmagnezi: Therefore I'm not sure we want to create an interim struct just so we can feed that function and return a new generator struct, which will be identical.
Let me know if I missed something here.

internal/bminventory/inventory.go
nmagnezi: done

internal/ignition/ignition.go
```diff
@@ -191,6 +196,30 @@ const highlyAvailableInfrastructureTopologyPatch = `---
...
+const bootReporterPath = "boot-reporter/assisted-boot-reporter.sh"
+
+var assistedBootReporterunitTemplate = `[Unit]
+Description=Collect and upload host boot logs to assisted-service
+Wants=network-online.target
+After=network-online.target
+DefaultDependencies=no
+[Service]
+Environment=ASSISTED_SERVICE_URL=%s
+Environment=PULL_SECRET_TOKEN=%s
```
eranco74, line 208: @tsorya why do we need this? how is the controller talking to the service? 
Can we use the same auth here?

internal/ignition/boot-reporter/assisted-boot-reporter.sh
```diff
@@ -0,0 +1,114 @@
...
+  log_info "${func_name}" "Copying /etc/resolv.conf to $logs_path/"
+
+  pushd /tmp
+  log_info "${func_name}" "Compressing logs to $logs_dir_name.tar.gz"
+  tar -czvf "$logs_dir_name".tar.gz "$logs_dir_name"
+  popd
+
+  log_info "${func_name}" "Uploading logs."
+
+  curl -X POST  -H "X-Secret-Key: ${PULL_SECRET_TOKEN}" \
```
eranco74, line 84: According to the swagger you don't need the infra_env_id in the query params when POSTing logs.
https://github.com/openshift/assisted-service/blob/64897a310a4b5f7eae7d115e159e8f7303b1ceab/swagger.yaml#L3084
By removing the infra_env_id you can avoid the db addition to the generator struct, less code changes +  reduce DB calls.

internal/ignition/boot-reporter/assisted-boot-reporter.sh
```diff
@@ -0,0 +1,114 @@
...
+  fi
+
+  if [ "$init_failed" == "true" ]; then
+    log_error "${func_name}" "Failed to initialize variables. Exiting."
+    exit 1
+  fi
+}
+
+function collect_and_upload_logs() {
+  func_name=${FUNCNAME[0]}
```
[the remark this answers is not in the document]
nmagnezi: `$FUNCNAME` is an array; it does without `[0]`, but the IDE linter points at [SC2128]( https://github.com/koalaman/shellcheck/wiki/SC2128)

internal/ignition/boot-reporter/assisted-boot-reporter.sh
```diff
@@ -0,0 +1,114 @@
...
+
+  log_info "${func_name}" "Collecting logs."
+  logs_dir_name=boot_logs_$HOST_ID
+  logs_path=/tmp/$logs_dir_name
+
+  rm -rf $logs_path
+  mkdir -p $logs_path
+
+  journalctl > "$logs_path"/journalctl.log
+  log_info "${func_name}" "Copying journalctl to $logs_path/"
```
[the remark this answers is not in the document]
nmagnezi: Fixed for this and the other two commands below.

internal/ignition/boot-reporter/assisted-boot-reporter.sh
```diff
@@ -0,0 +1,114 @@
...
+  ip a > $logs_path/ip_a.log
+  log_info "${func_name}" "Capturing the output of 'ip a' to $logs_path/"
+  cp /etc/resolv.conf $logs_path
+  log_info "${func_name}" "Copying /etc/resolv.conf to $logs_path/"
+
+  pushd /tmp
```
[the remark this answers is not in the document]
nmagnezi: In this case, it is [basically the same](https://opensource.com/article/19/8/navigating-bash-shell-pushd-popd).
btw we use `pushd` and `popd` in assisted-test-infra as well.

internal/ignition/boot-reporter/assisted-boot-reporter.sh
nmagnezi: That is a fair point. Indeed it should not be a must. I'll test it without `infra_env_id`.
https://github.com/openshift/assisted-service/blob/64897a310a4b5f7eae7d115e159e8f7303b1ceab/swagger.yaml#L3173-L3178

nmagnezi: > That is a fair point. Indeed it should not be a must. I'll test it without `infra_env_id`.
> 
> https://github.com/openshift/assisted-service/blob/64897a310a4b5f7eae7d115e159e8f7303b1ceab/swagger.yaml#L3173-L3178

/hold
testing this in CI.

omertuc (changes requested):

internal/ignition/boot-reporter/assisted-boot-reporter.sh
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env bash
```
omertuc, line 1: Let's run lint this file with shellcheck to avoid common mistakes

internal/ignition/boot-reporter/assisted-boot-reporter.sh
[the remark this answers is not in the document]
omertuc: I'm also in favor of pushd popd, as it scales when you want to add even more changes of directory, you just `popd; popd; popd;` at the end. While `cd -` only works when there was just a single change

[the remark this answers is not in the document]
omertuc: However, tar has this:

>       -C, --directory=DIR
              Change to DIR before performing any operations.  This option is order-sensitive, i.e. it affects all options that follow.
              
Which I think is better than changing the directory at the script level

internal/ignition/boot-reporter/assisted-boot-reporter.sh
```diff
@@ -0,0 +1,109 @@
...
+
+  curl -X POST  -H "X-Secret-Key: ${PULL_SECRET_TOKEN}" \
+       -F upfile=@$logs_path.tar.gz \
+       "$ASSISTED_SERVICE_URL/api/assisted-install/v2/clusters/$CLUSTER_ID/logs?logs_type=node-boot&host_id=$HOST_ID"
+
+  if [ $? -eq 0 ]; then
+      log_info "${func_name}" "Successfully uploaded logs."
+  else
+      log_error "${func_name}" "Failed to upload logs."
+  fi
```
omertuc, line 87: Can put the call directly inside the `if` statement:

```suggestion
    if curl -X POST  -H "X-Secret-Key: ${PULL_SECRET_TOKEN}" \
       -F upfile=@$logs_path.tar.gz \
       "$ASSISTED_SERVICE_URL/api/assisted-install/v2/clusters/$CLUSTER_ID/logs?logs_type=node-boot&host_id=$HOST_ID"; then
      log_info "${func_name}" "Successfully uploaded logs."
  else
      log_error "${func_name}" "Failed to upload logs."
  fi
```

internal/ignition/ignition.go
tsorya: Nope, controller is getting pull secret from secret.

internal/ignition/boot-reporter/assisted-boot-reporter.sh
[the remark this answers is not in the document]
tsorya: Totally agree, there is no need to cd into temp for running tar

internal/ignition/boot-reporter/assisted-boot-reporter.sh
```diff
@@ -0,0 +1,109 @@
...
+function log_info() {
+  log "info" "$1" "$2"
+}
+
+function log_error() {
+  log "error" "$1" "$2"
+}
+
+function init_variables() {
+  func_name=${FUNCNAME[0]}
```
tsorya, line 19: What is this FUNCNAME ? I don't see where we set it, sorry :(

internal/ignition/boot-reporter/assisted-boot-reporter.sh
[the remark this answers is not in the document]
tsorya: I think better to use $1 and just provide params to this function.

internal/ignition/boot-reporter/assisted-boot-reporter.sh
nmagnezi: no need to set it ourselves: https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-FUNCNAME

internal/ignition/boot-reporter/assisted-boot-reporter.sh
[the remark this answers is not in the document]
nmagnezi: @tsorya not sure I get you. this if logging the function name, not the process name.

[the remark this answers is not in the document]
tsorya: Now I get what it is. To say the truth I am not sure how it is helpful and why we really need this func name. 
As for me I would drop function name in order to remove this parameter from all the function. It doesn't worth it

internal/ignition/ignition.go
filanov: the difference between a function parameters and a struct is that it's clear to see what is set in a struct. 
it's a common pattern in apis that can have a lot of parameters. 
It's out of context for this PR so you can resolve this comment

internal/ignition/boot-reporter/assisted-boot-reporter.sh
nmagnezi: IMO, the existing convention is more readable.

nmagnezi: well, I see this is failing in shellcheck so I might change it after all

internal/ignition/boot-reporter/assisted-boot-reporter.sh
nmagnezi: I checked it, and it looks like infra_env should be marked as `required: true`.
Looking at:
https://github.com/openshift/assisted-service/blob/8b8e06e2eaf14d6497364d7fc0d4697d256cfe89/internal/bminventory/inventory_v2_handlers.go#L463-L487

We use `infra_env_id` to fetch the host from DB.
Even if I remove it, we also use it down the line with events.

This was the case for the existing host logs before I added node-boot logs, so this is not new.

internal/ignition/boot-reporter/assisted-boot-reporter.sh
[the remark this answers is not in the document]
nmagnezi: @tsorya, this is on par with the go-based logging we have. 
It helps to understand in which part of the script we had an issue, and it is already in place and working. So why drop?

internal/ignition/ignition.go
```diff
@@ -203,7 +232,7 @@ var fileNames = [...]string{
 
 // Generator can generate ignition files and upload them to an S3-like service
 type Generator interface {
-	Generate(ctx context.Context, installConfig []byte, platformType models.PlatformType) error
+	Generate(ctx context.Context, installConfig []byte, platformType models.PlatformType, serviceBaseURL string, authType auth.AuthType) error
```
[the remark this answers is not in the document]
nmagnezi: done

internal/ignition/boot-reporter/assisted-boot-reporter.sh
nmagnezi: done

internal/ignition/boot-reporter/assisted-boot-reporter.sh
nmagnezi: done

nmagnezi: done

internal/cluster/cluster.go
```diff
@@ -1055,7 +1055,12 @@ func (m *Manager) PrepareClusterLogFile(ctx context.Context, c *common.Cluster,
 					if hostObject.Bootstrap {
 						role = string(models.HostRoleBootstrap)
 					}
-					tarredFilename = fmt.Sprintf("%s_%s_%s.tar.gz", sanitize.Name(c.Name), role, sanitize.Name(hostutil.GetHostnameForMsg(hostObject)))
+					name := sanitize.Name(hostutil.GetHostnameForMsg(hostObject))
+					if strings.Contains(file, "boot_") {
+						name = fmt.Sprintf("boot_%s", name)
+
```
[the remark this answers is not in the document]
nmagnezi: removed.

internal/ignition/boot-reporter/assisted-boot-reporter.sh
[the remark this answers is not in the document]
nmagnezi: done

internal/ignition/boot-reporter/assisted-boot-reporter.sh
nmagnezi: Based on test results, it will force users to specify infra_env_id even when they upload cluster logs. I'll revert back to `required: false`. It looks out of scope for this PR.

nmagnezi: done

internal/ignition/boot-reporter/assisted-boot-reporter.sh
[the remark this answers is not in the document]
eranco74: @tsorya FUNCNAME is a bash Internal variable, you get if for free.

internal/ignition/boot-reporter/assisted-boot-reporter.sh
omertuc: /hold

omertuc: I don't see a shellcheck linter being called anywhere so holding

internal/ignition/boot-reporter/assisted-boot-reporter.sh
eranco74: I see, sorry for all the truble this put you through

eranco74: Note that you could update:
`dbHost, err := common.GetHostFromDB(b.db, params.InfraEnvID.String(), params.HostID.String()` 
to:
`dbHost, err := common.GetHostFromDBByHostId(b.db, params.InfraEnvID.String(), params.HostID.String()`

internal/ignition/templates/node.ign
```diff
@@ -15,4 +15,4 @@
       }
     }{{end}}
   }
-}
+}
```
tsorya, line 18: File changed without any reason

internal/ignition/boot-reporter/assisted-boot-reporter.sh
tsorya: Sec, i agree with @eranco74 actually, if you don't need infraenv why we need to add  it and complicate so many stuff.
There is no  need for it at all. 
If you not must use it please remove

internal/ignition/templates/node.ign
nmagnezi: fixed.

omertuc:
/hold

https://github.com/openshift/assisted-service/pull/4543#discussion_r1087603540

omertuc:
Will add shellcheck linting in a separate PR

/unhold

nmagnezi:
/hold
testing another attempt to get rid of infra_env_id

cmd/main.go
```diff
@@ -395,7 +395,7 @@ func main() {
 	failOnError(err, "failed to create valid bm config S3 endpoint URL from %s", Options.BMConfig.S3EndpointURL)
 	Options.BMConfig.S3EndpointURL = newUrl
 
-	generator := generator.New(log, objectHandler, Options.GeneratorConfig, Options.WorkDir, operatorsManager, providerRegistry, Options.ClusterTLSCertOverrideDir)
+	generator := generator.New(log, db, objectHandler, Options.GeneratorConfig, Options.WorkDir, operatorsManager, providerRegistry, Options.ClusterTLSCertOverrideDir, authHandler)
```
eranco74, line 398: Why do we need the DB?

internal/ignition/ignition.go
```diff
@@ -1249,15 +1281,57 @@ func setCACertInIgnition(role models.HostRole, path string, workDir string, caCe
...
-	g := new(errgroup.Group)
+func getBootReporterFileContent() (string, error) {
+	data, err := reporter.ReadFile(bootReporterPath)
+	if err != nil {
+		return "", err
+	}
+	return base64.StdEncoding.EncodeToString(data), nil
+}
+
+func (g *installerGenerator) getHostsInfraEnvs(hosts []*models.Host) (map[string]*common.InfraEnv, error) {
```
eranco74, line 1292: This should be removed

internal/ignition/ignition.go
```diff
@@ -1249,15 +1281,57 @@ func setCACertInIgnition(role models.HostRole, path string, workDir string, caCe
...
 	for i := range hosts {
 		host := hosts[i]
-		g.Go(func() error {
+		errGroup.Go(func() error {
 			config, err := parseIgnitionFile(filepath.Join(workDir, baseFile))
 			if err != nil {
 				return err
 			}
+			infraEnv := infraEnvs[host.InfraEnvID.String()]
+			pullSecretToken, err := clusterPkg.AgentToken(infraEnv, authType)
```
eranco74, line 1329: pullSecretToken := g.cluster.PullSecret

nmagnezi: done

internal/ignition/ignition.go
nmagnezi: done

cmd/main.go
nmagnezi: done

internal/cluster/auth.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func AgentToken(infraEnv *common.InfraEnv, authType auth.AuthType) (token string, err error) {
+func AgentToken2(infraEnv *common.InfraEnv, authType auth.AuthType) (token string, err error) {
```
nmagnezi, line 11: this should go away..

internal/cluster/auth.go
```diff
@@ -8,12 +8,27 @@ import (
...
+	)
+	switch res := resource.(type) {
+	case *common.InfraEnv:
+		resId = res.ID.String()
+		pullSecret = res.PullSecret
+	case *common.Cluster:
+		resId = res.ID.String()
+		pullSecret = res.PullSecret
+	default:
+		return "", errors.New("internal error")
```
eranco74, line 24:
```suggestion
		return "", fmt.Errorf("Unsupported type, expected InfraEnv or Cluster got %s", reflect.TypeOf(resource))
```

eranco74 (commented):
looks good, just a nit

internal/cluster/auth.go
nmagnezi: Done

eranco74 (commented):
/lgtm

PR #4545, opened 2022-10-26
NO-ISSUE: annotate lso namespace instead of project

adriengentil:
The annotation on project is currently failing with:
```
$ oc annotate project openshift-local-storage openshift.io/node-selector="" --overwrite=true
The Project "openshift-local-storage" is invalid:
* metadata.namespace: Invalid value: "openshift-local-storage": field is immutable
* metadata.namespace: Forbidden: not allowed on this type
```

while the annotation on the namespace seems to work:
```
$ oc annotate namespace openshift-local-storage openshift.io/node-selector="" --overwrite=true
namespace/openshift-local-storage annotated
```

### DIFF
--- a/deploy/operator/setup_lso.sh
+++ b/deploy/operator/setup_lso.sh
@@ -18,7 +18,7 @@ function print_help() {
 function install_lso() {
   oc adm new-project openshift-local-storage || true
 
-  retry -- oc annotate project openshift-local-storage openshift.io/node-selector='' --overwrite=true
+  retry -- oc annotate namespace openshift-local-storage openshift.io/node-selector='' --overwrite=true
 
   catalog_source_name="redhat-operators"
 


carbonin:
Looks like some kind of package installation issue?

```
Depsolve Error occurred: 
 Problem 1: cannot install the best update candidate for package gnutls-dane-3.6.16-4.el8.x86_64
  - nothing provides gnutls(x86-64) = 3.6.16-5.el8_6 needed by gnutls-dane-3.6.16-5.el8_6.x86_64
 Problem 2: cannot install the best update candidate for package gnutls-utils-3.6.16-4.el8.x86_64
  - nothing provides gnutls(x86-64) = 3.6.16-5.el8_6 needed by gnutls-utils-3.6.16-5.el8_6.x86_64
```

adriengentil:
I think we can go ahead, `ztp-capi` failed later in the testing and `ztp` failed to build the dev-script image which is not related to this change.

PR #4547, opened 2022-10-26
MGMT-12368: Verify that Kernel Arguments API does not change in the future

ori-amizur:
Since Coreos Installer supports [append, replace, delete] operations on kernel arguments,  this API is modified accordingly.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @carbonin 
/cc @avishayt


carbonin:
Should we have an explicit reset operation? It's hard to understand that passing an empty list of args is the intended way to do that.

ori-amizur:
> Should we have an explicit reset operation? It's hard to understand that passing an empty list of args is the intended way to do that.

I think reset is meaningless since we are not editing and saving ISO.  The ISO is always in its default state.  We are modifying and streaming it without saving.

carbonin:
> I think reset is meaningless since we are not editing and saving ISO. The ISO is always in its default state. We are modifying and streaming it without saving.

I'm aware of this, but we do need a way for users to clear the value from the DB. I'm wondering if an update with an empty list is enough or if we should consider something more explicit? I'm just trying to ensure we get this right this time and don't have to change the API again.

Curious what @avishayt thinks.

avishayt:
Thanks for this PR.  Let's think about usability.  A few questions since I don't have time right now to try it out:
1. Do we know what the default kernel arguments are at this point, or is that knowledge only in the image service?
2. Why does the "replace" action exist as opposed to using "delete" and "append"?
3. To confirm my understanding - "delete" would be used only for removing a default kernel arg, "replace" would be used for changing the value of default kernel arg, and "append" would be for adding new args.  Correct?

ori-amizur:
> Do we know what the default kernel arguments are at this point, or is that knowledge only in the image service?

Even the image service is not aware of the kernel arguments, although it is possible to query them.  In addition, these arguments may vary between different ISO files.

> Why does the "replace" action exist as opposed to using "delete" and "append"?

That's Coreos Installer terminology

> To confirm my understanding - "delete" would be used only for removing a default kernel arg, "replace" would be used for changing the value of default kernel arg, and "append" would be for adding new args. Correct?

Yes

avishayt (approved):
I'm OK with this API, thanks for the change

PR #4548, opened 2022-10-27
MGMT-12353: Add dual-stack VIP fields to the Rest API

mkowalski:
This PR extends the Rest API to allow storing API and Ingress VIPs as an array of IP addresses. This is because with introduction of dual-stack VIPs it is no longer feasible to have a singular field to store aforementioned VIP.

With this change we are adding a new set of fields but we do not remove the existing ones. This is because we need to comply with the phase-out process so for some releases both singular and plural fields are required to co-exist.

Contributes-to: [MGMT-12353](https://issues.redhat.com//browse/MGMT-12353)
Implements: #4245

/cc @nmagnezi 
/cc @avishayt


nmagnezi (changes requested):
What about tests? Unit and subsystem.

mkowalski:
> What about tests? Unit and subsystem.

Given that there is no any logic nor behaviour implemented in this PR (it's just change of API schema), I implicitly assumed that tests will come once we start working with those fields. If not, let me know what you would see here; today a test would look like

```
cluster := createcluster{
[...]
api_vips:
  - someip1
  - someip2
}

Expect(cluster.api_vips.ToBe(nil))
```

which may be confusing as in your payload you did set them.

We could add those, but let me know if you really want that path

avishayt (approved):

avishayt (approved):

mkowalski:

Race

mkowalski:
/hold

Some issues with the DB naming schema, `api_vip` vs `api_vips` as DB table name

mkowalski:
@avishayt @nmagnezi conflict with Ori's PR, please re-lgtm

PR #4549, opened 2022-10-27
MGMT-12353: Add dual-stack VIP fields to the KubeAPI

mkowalski:
This PR extends the KubeAPI to allow storing API and Ingress VIPs as an array of IP addresses. This is because with introduction of dual-stack VIPs it is no longer feasible to have a singular field to store aforementioned VIP.

With this change we are adding a new set of fields but we do not remove the existing ones. This is because we need to comply with the phase-out process so for some releases both singular and plural fields are required to co-exist.

Contributes-to: MGMT-12353
Contributes-to: MGMT-9915
Implements: #4245

/cc @nmagnezi
/cc @avishayt


nmagnezi (changes requested):
1. Missing tests.
2. Should bump the API version per https://github.com/openshift/assisted-service/pull/4245: 
`Given that all the newly introduced fields are optional, there is no need to release a new version of the API.`

mkowalski:
Re 1, copy-pasta from the other PR

Given that there is no any logic nor behaviour implemented in this PR (it's just change of API schema), I implicitly assumed that tests will come once we start working with those fields. If not, let me know what you would see here; today a test would look like

```
cluster := createcluster{
[...]
api_vips:
  - someip1
  - someip2
}

Expect(cluster.api_vips.ToBe(nil))
```

which may be confusing as in your payload you did set them.

We could add those, but let me know if you really want that path

--------

Re 2:

We do not bump the API version now, as the new fields are only optional. We will bump it in the Phase 3 of the enhancement, when we deprecate something that already exists

nmagnezi:
> as the new fields are only optional. We will bump it in the Phase 3 of the enhancement, when we deprecate something that already exis

Ack. Approving.

nmagnezi (approved):
/lgtm

avishayt (approved):

mkowalski:
/hold

Some issues with the DB naming schema, `api_vip` vs `api_vips` as DB table name

PR #4551, opened 2022-10-28
MGMT-12361: Sync install-config structure with upstream

mkowalski:
With https://github.com/openshift/installer/pull/5798 o/installer allows
to use a new install-config syntax in order to support dual-stack VIPs.
Because of this there are now allowed both `apiVIP` and `apiVIPs` with
the former marked as deprecated internally.

This PR adapts to this change by marking singular field internally as
deprecated but still mapping it to the backwards-compatible field in the
generated yaml.

The change does not introduce any breakage to the current flows but
transparently allows to decide on whether the singular or plural struct
is to be used in the generated install-config. The ability to choose
whether singular or plural is used comes from the `omitempty` property
which will not place the key in the generated yaml if the value has not
been set. Without the property we would be generating yaml with key with
empty value and such a combination would not be passing o/installer
sanity checks.

Closes: [MGMT-12361](https://issues.redhat.com//browse/MGMT-12361)
Caused-by: https://github.com/openshift/installer/pull/5798

/cc @omertuc


omertuc:
Can you edit the commit message to include an explanation of why the added omitempty? I feel it's not trivial

omertuc:
Looks good otherwise

omertuc:
Do we have 4.12 jobs we can test this against?

mkowalski:
> Do we have 4.12 jobs we can test this with?

Yes, we have recently moved lots of our CI jobs to use 4.12 as the default latest

mkowalski:
> Can you edit the commit message to include an explanation of why the added omitempty? I feel it's not trivial

Of course, just edited

mkowalski:
/test edge-subsystem-aws

```
 RuntimeError: command=kubectl --namespace ci-op-11l0t20r get service assisted-image-service | grep assisted-image-service exited with an error=The connection to the server api.ci-ocp-4-10-amd64-aws-us-east-1-xm2sr.hive.aws.ci.openshift.org:6443 was refused - did you specify the right host or port? code=1 
```

PR #4562, opened 2022-11-02
MGMT-12354: Migrate single VIP values to the new data structure

mkowalski:
With the new way of storing the data, we need to pre-populate database so that the new tables `api_vips` and `ingress_vips` get the data for all the existing clusters filled.

This PR is resposible for going over all the existing clusters and populating the newly created DB tables with already known IP addresses.

Closes: [MGMT-12354](https://issues.redhat.com//browse/MGMT-12354)
Implements: #4245

/cc @nmagnezi

### DIFF
--- a/internal/common/db.go
+++ b/internal/common/db.go
@@ -138,13 +138,32 @@ const (
 	ClusterNetworksTable    = "ClusterNetworks"
 	ServiceNetworksTable    = "ServiceNetworks"
 	MachineNetworksTable    = "MachineNetworks"
+	APIVIPsTable            = "APIVips"
+	IngressVIPsTable        = "IngressVips"
 )
 
-var ClusterSubTables = [...]string{HostsTable, MonitoredOperatorsTable, ClusterNetworksTable, ServiceNetworksTable, MachineNetworksTable}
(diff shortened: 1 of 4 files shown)


internal/common/db.go
```diff
@@ -138,13 +138,32 @@ const (
...
+	ServiceNetworksTable,
+	MachineNetworksTable,
+	APIVIPsTable,
+	IngressVIPsTable,
+}
 
 func AutoMigrate(db *gorm.DB) error {
-	return db.AutoMigrate(&models.MonitoredOperator{}, &Host{}, &Cluster{}, &Event{}, &InfraEnv{},
-		&models.ClusterNetwork{}, &models.ServiceNetwork{}, &models.MachineNetwork{})
+	return db.AutoMigrate(&models.MonitoredOperator{},
```
filanov, line 156: i like that you broke this line, it will be clearer to see in git log when someone add a new field.

nmagnezi (approved):
/lgtm

internal/common/db.go
mkowalski: Yes indeed, `git blame` becomes my daily driver ;-)

PR #4564, opened 2022-11-03
MGMT-12422: Fix bm platform update failure when cluster platform different than bm/none

eliorerz:
We missed the case where the cluster has platform different than baremetal/none and the user set the platform to baremetal.
Furthermore, UMN current value should be taken into account when updating the platform, for example if cluster platform is vsphere and UMN is enabled the user should not set the platform to BM (because `platform=baremetal and UMN=true` is not a valid state).
This PR handles those invalid cases and return the correct update value if it wasn't met to any condition

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist
/cc @eranco74 
/cc @gamli75

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -45,6 +45,7 @@ import (
 	"github.com/openshift/assisted-service/internal/metrics"
 	"github.com/openshift/assisted-service/internal/network"
 	"github.com/openshift/assisted-service/internal/operators"
+	"github.com/openshift/assisted-service/internal/provider"
 	"github.com/openshift/assisted-service/internal/provider/registry"
 	"github.com/openshift/assisted-service/internal/usage"
 	"github.com/openshift/assisted-service/internal/versions"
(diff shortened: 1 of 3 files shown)


eliorerz:
We can ignore `ci/prow/edge-e2e-nutanix-assisted` at the moment, apparently it is broken
/override ci/prow/edge-e2e-nutanix-assisted

internal/bminventory/inventory.go
```diff
@@ -1729,6 +1729,16 @@ func checkPlatformWrongParamsInput(platform *models.Platform, userManagedNetwork
 		}
 	}
 
+	if cluster != nil && *cluster.Platform.Type != models.PlatformTypeBaremetal && *cluster.Platform.Type != models.PlatformTypeNone {
```
eranco74, line 1732: Consider extracting this logic to a function e.g. `! (isPlatfromBM || isPlatfromNone)`

eranco74, line 1732: Add a comment describing why we need this

internal/provider/common.go
```diff
@@ -0,0 +1,146 @@
+package provider
```
eliorerz, line 1: This logic moved from `bminventory/inventory.go` 
The only changes are extracting the some logic to functions to make it more clean and easy to read.
new methods are:

1. `isPlatformBM`
2. `isPlatformNone`
3. `isClusterPlatformNone`
4. `isClusterPlatformBM`

Using those methods simplify the already complex cases.

internal/provider/common.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func doesPlatformDifferentThanBaremetalOrNone(platform *models.Platform, cluster *common.Cluster) bool {
+	if platform != nil && !isPlatformBM(platform) && !isPlatformNone(platform) {
+		return true
+	}
+
+	if platform == nil && !isClusterPlatformBM(cluster) && !isClusterPlatformNone(cluster) {
+		return true
+	}
```
eranco74, line 64: I think you can remove this part, it seems that this scenario is already covered in  `GetActualUpdateClusterPlatformParams` (lines 82 and 91)

eranco74 (commented):
/lgtm

PR #4565, opened 2022-11-03
MGMT-12389: Create flag for per-version support of dual-stack VIPs

mkowalski:
This PR introduces an internal flag to track support of dual-stack VIPs as a function of the OCP version. This will make it easier to implement various checks as they won't need to hardcode a version, but they will just be able to refer to the `SupportLevel`.

Contributes-to: [MGMT-12389](https://issues.redhat.com//browse/MGMT-12389)
Implements: #4245

/cc @nmagnezi 
/cc @avishayt


avishayt:
The code looks fine but doesn't seem to match the commit message / PR description.  It looks like you implemented the SupportLevel - I'm not sure what the internal flag is.  Did I miss something?

mkowalski:
I think we are talking about the same thing in fact - I need something that answers (in a manner yes/no) a question "is the feature X supported with OpenShift in version Y" and this is it. Yes, internally it's called `SupportLevelsList` but at the end of the day it's yet another flag amongst Feature Flag, Feature Usage Flag and maybe some others that I am not aware of.

Anyway if you feel like renaming anything here, let me know

PR #4566, opened 2022-11-03
MGMT-12068: kube api support for discovery kernel arguments

ori-amizur:
Add kube API support for discovery kernel arguments.  This API adds in the InfraEnv Spec section a field called "DiscoveryKernelArguments". The field is a collection of KernelArgument objects.  Each such object has attribute "operation", which defines the operation to be done regarding the other attribute "value".
There are 3 operations defined in the API: [append, replace, delete]. The Reconcile function just vrifies that the discovery kernel arguments field in the InfraEnv database record is in sync with the spec, where the spec is considered as "source of truth".

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @avishayt 
/cc @carbonin


internal/controller/controllers/infraenv_controller.go
```diff
@@ -183,10 +184,46 @@ func (r *InfraEnvReconciler) updateInfraEnv(ctx context.Context, log logrus.Fiel
 
 	updateParams.InfraEnvUpdateParams.ImageType = r.Config.ImageType
 
+	existingKargs, err := existingDiscoveryKernelArguments(internalInfraEnv)
+	if err != nil {
+		return nil, err
+	}
+	if !funk.Equal(infraEnv.Spec.DiscoveryKernelArguments, existingKargs) {
```
carbonin, line 191: Does order matter or should this do some kind of set comparison?

api/v1beta1/infraenv_types.go
```diff
@@ -98,6 +98,24 @@ type InfraEnvSpec struct {
 	// +kubebuilder:default=DiscoveryImageAlways
 	// +optional
 	IPXEScriptType IPXEScriptType `json:"ipxeScriptType"`
+
+	// The discovery kernel arguments to be passed during boot time.
+	// Applicable for both iPXE, and ISO streaming from Image Service.
+	// +optional
+	DiscoveryKernelArguments []KernelArgument `json:"discoveryKernelArguments,omitempty"`
```
carbonin, line 105: I don't think we need `Discovery` here as everything in the infra env is related to discovery.
For example, we not writing `DiscoveryIgnitionConfigOverride` here.

internal/controller/controllers/infraenv_controller.go
```diff
@@ -183,10 +184,46 @@ func (r *InfraEnvReconciler) updateInfraEnv(ctx context.Context, log logrus.Fiel
...
+func discoveryKernelArgumentsParams(kargs []aiv1beta1.KernelArgument) models.KernelArguments {
+	var ret models.KernelArguments
+	for _, arg := range kargs {
+		ret = append(ret, &models.KernelArgument{
+			Operation: arg.Operation,
+			Value:     arg.Value,
+		})
+	}
+	return ret
+}
```
carbonin, line 225: The meat of these functions is really transitioning between the two types, right?
The names make it hard to see that.

Maybe something like:
`func kubeKernelArgs(models.KernelArguments) []aiv1beta1KernelArguments`
 and
`func internalKernelArgs([]aiv1beta1.KernelArgument) models.KernelArguments`?

carbonin (commented):
Any comments on the API @avishayt @mhrivnak ?

internal/controller/controllers/infraenv_controller.go
ori-amizur: Seems to me yes.  That is the order that will be applied to iPXE and ISO.  Do you think differently?

api/v1beta1/infraenv_types.go
ori-amizur: Couldn't someone think that these arguments will be used after reboot?

internal/controller/controllers/infraenv_controller.go
ori-amizur: I am open about the names ...
But please note that these functions are not the reverse of each other.

internal/controller/controllers/infraenv_controller.go
carbonin: No, I'm actually asking if the order matters when you provide kernel arguments.

If it doesn't then we might be able to skip some unnecessary updates, but if it does then this looks fine to me.
I suppose keeping it as is is safer in case order does indeed matter :man_shrugging:

api/v1beta1/infraenv_types.go
carbonin: No? We already have an existing mechanism for that.
And if they do then they should try to understand how this system works a bit more before using it :sweat_smile:

internal/controller/controllers/infraenv_controller.go
carbonin: > But please note that these functions are not the reverse of each other.

I know, but they could be made to be if the unmarshal is moved out of the first one. Since the first function is only called in one place I'm suggesting the parity between the two functions will make them easier to understand and easier to name.

api/v1beta1/infraenv_types.go
ori-amizur: > No? We already have an existing mechanism for that.

Regardless, name should provide best hint to the usage.

> And if they do then they should try to understand how this system works a bit more before using it

But when someone tries to understand the code before using it, clearer names will help understanding. 

I think we should try to keep the same names between db and kube entities.  It helps during searches and it provides a hint that they represent the same thing.  I think it is clearer.

internal/controller/controllers/infraenv_controller.go
ori-amizur: OK.  Will change

api/v1beta1/infraenv_types.go
avishayt: I agree with both of you :)
Let's go with KernelArguments, and change the swagger to match (it's also in the context of infra-env there).

ori-amizur: @avishayt I am not sure we want to do that.
discovery_kernel_arguments is also the name of the column in the database.  Do we want to change ?  It will require migration.

avishayt: Should be straightforward, no?

ori-amizur: It means we keep the migration forever.
BTW, I think the name DiscoveryKernelArguments is better. 
Do you think the change is essential ?

ori-amizur: Change

api/v1beta1/infraenv_types.go
```diff
@@ -98,6 +98,24 @@ type InfraEnvSpec struct {
 	// +kubebuilder:default=DiscoveryImageAlways
 	// +optional
 	IPXEScriptType IPXEScriptType `json:"ipxeScriptType"`
+
+	// The discovery kernel arguments to be passed during boot time.
```
avishayt, line 102: The k8s API convention is to start each of these comments with the property name.  So this would be "KernelArguments is a list of kernel arguments to pass..."

Same for Operation and Value below

internal/controller/controllers/infraenv_controller.go
```diff
@@ -183,10 +184,46 @@ func (r *InfraEnvReconciler) updateInfraEnv(ctx context.Context, log logrus.Fiel
...
+		updateParams.InfraEnvUpdateParams.KernelArguments = internalKernelArgs(infraEnv.Spec.KernelArguments)
+	}
+
 	// UpdateInfraEnvInternal will generate an ISO only if there it was not generated before,
 	return r.Installer.UpdateInfraEnvInternal(ctx, updateParams, nil)
 }
 
+func kubeKernelArgs(internalInfraEnv *common.InfraEnv) ([]aiv1beta1.KernelArgument, error) {
+	var args models.KernelArguments
+	if internalInfraEnv.KernelArguments != nil {
```
mhrivnak, line 201: This is a good opportunity to also make sure `internalInfraEnv != nil`

internal/controller/controllers/infraenv_controller.go
```diff
@@ -183,10 +184,46 @@ func (r *InfraEnvReconciler) updateInfraEnv(ctx context.Context, log logrus.Fiel
...
+		if err := json.Unmarshal([]byte(swag.StringValue(internalInfraEnv.KernelArguments)), &args); err != nil {
+			return nil, errors.Wrapf(err, "failed to unmarshal discovery kernel arguments of infra-env %s", internalInfraEnv.ID.String())
+		}
+	}
+	var ret []aiv1beta1.KernelArgument
```
mhrivnak, line 206: just an observation: generally when transforming a slice of X to a slice of Y, we know the length, so it's helpful to `make()` the slice the correct length up-front and assign values as you go.

```
ret := make([]aiv1beta1.KernelArgument, len(args))
for i := range args {
  ret[i] = ...
}
```

This helps avoid re-sizing and fragmentation of the slice's underlying data, though for cases like this it probably doesn't make much difference.

api/v1beta1/infraenv_types.go
```diff
@@ -98,6 +98,24 @@ type InfraEnvSpec struct {
 	// +kubebuilder:default=DiscoveryImageAlways
 	// +optional
 	IPXEScriptType IPXEScriptType `json:"ipxeScriptType"`
+
+	// KernelArguments is the kernel arguments to be passed during boot time of the discovery image.
```
mhrivnak, line 102: These are in addition to some default kernel args?

If so, what happens if someone specifies a different value here than the default? Does this API let you override the value of a default kernel arg?

ori-amizur: Our implementation supports only additional arguments (for now).
The API also supports **replace** and **delete** operations.  This is meant for existing arguments manipulation.

internal/controller/controllers/infraenv_controller.go
ori-amizur: Right, usually the slice size for kernel arguments is small.

api/v1beta1/infraenv_types.go
mhrivnak: It would help to clarify in the comment then that these are "additional kernel arguments to be passed..."

internal/controller/controllers/infraenv_controller.go
ori-amizur: OK

avishayt (approved):

PR #4347, opened 2022-09-02
MGMTBUGSM-547: Add UnsyncedAgents error in ClusterDeployment and ACI

mkowalski:
This PR adds a new "UnsyncedAgents" condition for marking clusters where
at least one of Agents has SpecSynced failure. Currently if this happens
we are showing a misleading error "UnapprovedAgents" coming from the
fact that data in the DB is not in sync with K8s CRs.

Cherry-picks: Bug-2076333
Cherry-picks: MGMTBUGSM-307
Closes: MGMTBUGSM-547

/cc @avishayt 
/cc @filanov

PR #4571, opened 2022-11-04
MGMT-12423: Add debug logs for counting unsynced agents

mkowalski:
In order to make it easier to debug race conditions in the KubeAPI, this PR adds additional debug logs that will be returning results of the intermediate steps that lead to marking cluster as `UnsyncedAgents` or `UnapprovedAgents`.

Contributes-to: [MGMT-12423](https://issues.redhat.com//browse/MGMT-12423)

/cc @omertuc


omertuc:
Yikes

mkowalski:
OCM 2.5 will not work automatically, but I'll handle it manually by combining together with https://github.com/openshift/assisted-service/pull/4347

PR #4568, opened 2022-11-03
MGMT-12355: API and Ingress VIPs CRUD implementation

nmagnezi:
This PR adds support for plural API and Ingress VIPs, both to
`V2RegisterCluster` and `V2UpdateCluster`.

Note that API backward compatibility is currently kept, and thus singular API VIP
and Ingress VIP were not removed.
However, please look at https://github.com/openshift/assisted-service/pull/4245 to understand the expected behavior when old and new API parameters.

As this PR is big enough on its own, a follow-up PR will handle the
plural versions of `isApiVipValid` and `isIngressVipValid` state-machine validations.

Contributes-to: [MGMT-12355](https://issues.redhat.com//browse/MGMT-12355)
Implements: https://github.com/openshift/assisted-service/pull/4245

## List all the issues related to this PR

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [x] Automation (CI, tools, etc)
- [x] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #4573, opened 2022-11-04
MGMT-12360: Pass multiple VIPs down to install-config on OCP >=4.12

mkowalski:
After the change introduced in
https://github.com/openshift/installer/pull/5798, we can now pass multiple VIPs to the install-config.yaml

This PR adds a logic that checks the version of the cluster that is about to be installed, compares it with a set of allowed fields in install-config struct and in case we are using OCP version with dual-stack VIPs allowed, all those will be passed down to install-config.

In case of older versions the old syntax is used and only single value is allowed for any VIP.

Closes: MGMT-12360
Contributes-to: MGMT-9915
Implements: #4245

/cc @nmagnezi


mkowalski:
/hold

Requires: #4565
Requires: #4568

mkowalski:
Update on prerequisites for this PR:

Requires: #4575
Requires: #4568

nmagnezi (commented):
Code LGTM.
I'll vote as soon as we merge https://github.com/openshift/assisted-service/pull/4568 and https://github.com/openshift/assisted-service/pull/4643

nmagnezi (approved):
/lgtm

mkowalski:
/unhold

All dependencies are merged by now

nmagnezi (approved):
/lgtm

PR #4574, opened 2022-11-04
NO-ISSUE: Rename misleading network utility function

mkowalski:
This PR renames the function `GetClusterNetworks` to `GetInventoryNetworks` in order to remove confusion about what Cluster Network is.

The logic of the function goes over the inventory of all discovered hosts and returns all the CIDRs that have been found.

On the other hand, Cluster Network in the OpenShift universe is a network that is used for Pods to talk with each other and is not related at all to the physical network where Nodes belong to.

In order to remove a confusion (because we already have `GetClusterNetworkCidrs` which does something completely different than `GetClusterNetworks`) we are renaming it and use Inventory Networks.

/cc @nmagnezi


nmagnezi (approved):
/lgtm

PR #4575, opened 2022-11-04
MGMT-9915: Create utility functions to get cluster's VIPs

mkowalski:
This PR creates a set of utility functions that are returning cluster's VIPs as a simple `[]string` or `string` instead of a custom data structures. This is to make it easier to perform parsing and validations when only the value of the VIP is required and not the whole context it carries.

Implements: #4245

/cc @nmagnezi

### DIFF
--- a/internal/network/utils.go
+++ b/internal/network/utils.go
@@ -177,6 +177,34 @@ func GetClusterNetworkCidrs(cluster *common.Cluster) (ret []string) {
 	return
 }
 
+func GetApiVips(cluster *common.Cluster) (ret []string) {
+	ret = funk.Map(cluster.APIVips, func(x *models.APIVip) string { return string(x.IP) }).([]string)
+	return
+}
+
+func GetIngressVips(cluster *common.Cluster) (ret []string) {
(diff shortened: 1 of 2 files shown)


internal/network/utils.go
```diff
@@ -177,6 +177,34 @@ func GetClusterNetworkCidrs(cluster *common.Cluster) (ret []string) {
...
+	ret = funk.Map(cluster.APIVips, func(x *models.APIVip) string { return string(x.IP) }).([]string)
+	return
+}
+
+func GetIngressVips(cluster *common.Cluster) (ret []string) {
+	ret = funk.Map(cluster.IngressVips, func(x *models.IngressVip) string { return string(x.IP) }).([]string)
+	return
+}
+
+func GetApiVipById(cluster *common.Cluster, index int) string {
```
nmagnezi, line 190: A proposal that we can do later if we agree that it would be a good idea: 
We agree that both `GetApiVipById` and `GetIngressVipById` should be failing silently by returning an empty string. However, it might be good to add a debug-level log for this.

nmagnezi (approved):
/lgtm

PR #4580, opened 2022-11-06
MGMT-12462: Fix assisted-service swagger file

osherdp:
There are a couple of problems with current version of swagger.yaml file:

* We're sometimes using ``format: datetime`` instead of ``format: date-time``, while datetime isn't a built-in format. E.g.: https://github.com/openshift/assisted-service/blob/f267d0a7caa5a3f1e3f981dab345e1408845decf/swagger.yaml#L3973
* We're not annotating optional fields as nullable. E.g.: https://github.com/openshift/assisted-service/blob/f267d0a7caa5a3f1e3f981dab345e1408845decf/swagger.yaml#L4645-L4651
* required fields list is not under properties: https://github.com/openshift/assisted-service/blob/cd50460292ee175db28ec18f20f651140a59a5ec/swagger.yaml#L3578-L3584
* It is probably redundant to specify both ``type: object`` and referencing it to a type definition: https://github.com/openshift/assisted-service/blob/cd50460292ee175db28ec18f20f651140a59a5ec/swagger.yaml#L4250-L4252

For most cases it probably doesn't matter for the Python client generator, but it's better fixing those issues to be more compatible to other generators / clients.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

I did a bit of manual testing to see ``bravado`` client being able to consume the new spec with less warnings than before (still got some, but it doesn't seem to limit too much)

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/vendor/github.com/openshift/assisted-service/models/feature_support_level.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/feature_support_level.go
@@ -126,12 +126,14 @@ func (m *FeatureSupportLevel) UnmarshalBinary(b []byte) error {
 type FeatureSupportLevelFeaturesItems0 struct {
 
 	// The ID of the feature
+	// Required: true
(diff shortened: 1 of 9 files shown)


avishayt:
Great PR, thanks for catching this.  Do any of these changes break the API?

avishayt:
Bonus points if you add rules to .spectral.yml to make sure we don't have these issues in the future :)

osherdp:
> Great PR, thanks for catching this. Do any of these changes break the API?

I think none of them is an API breakage. Explanation:
* ``datetime`` vs ``date-time`` format: part of a model we're exposing outside without interfering with any kind of request's validation. It makes sense for most clients to support the ``date-time`` format rather than the ``datetime`` non-official format
* foreign-key values that weren't nullable: I can see ``api_vips`` coming up as null with existing code, so it should be more correct to explicitly specify it in the spec
* required field list not in the right place: those models that are only being exposed and not given, it's the original intent to have those fields required as it doesn't make sense to have a feature_id mentioned without its support_level and via versa
* redundant object and ref: doesn't seem to break. The openapi client I'm using says the ref should override its siblings for those cases

osherdp:
/retest
seems like flakes

PR #4583, opened 2022-11-07
WIP: Add a way to apply patches to core manifests

flaper87:

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


flaper87:
/hold 

Marking this as ready for review now that subsystem tests and the rest of the CI passed. I'm holding it until we have properly tested e2e. We won't be releasing this PR until next week anyways.

internal/ignition/ignition.go
```diff
@@ -391,6 +391,12 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte,
 		}
 	}
 
+	err = g.applyManifestPatches(ctx)
+	if err != nil {
+		log.WithError(err).Errorf("failed to apply manifests' patches for cluster '%s'", g.cluster.ID)
```
omertuc, line 396: We probably want to clean up the mess just as we do above (also in the conditions below)

Maybe it's best to just defer a conditional cleanup when err != nil 

```suggestion
		_ = os.Remove(filepath.Join(g.workDir, "manifests"))
		_ = os.Remove(filepath.Join(g.workDir, "openshift"))
		log.WithError(err).Errorf("failed to apply manifests' patches for cluster '%s'", g.cluster.ID)
```

internal/ignition/ignition.go
```diff
@@ -468,6 +474,66 @@ func (g *installerGenerator) addBootstrapKubeletIpIfRequired(log logrus.FieldLog
...
+
+		manifestPatchPath := filepath.Join(manifestsOpenShiftDir, info.Name())
+		log.Debugf("Applying the following patch: %s", manifestPatchPath)
+		manifestPatch, err := os.ReadFile(manifestPatchPath)
+		if err != nil {
+			return errors.Wrapf(err, "failed to read manifest patch %s", manifestPatchPath)
+		}
+		log.Debugf("read the manifest at %s", manifestPatchPath)
+
+		manifestPath := filepath.Join(manifestsOpenShiftDir, strings.TrimSuffix(info.Name(), ".patch"))
```
omertuc, line 498: For patches you allow both `.yml` and `.yaml`, yet when it comes time to find the actual YAML to be patched - you insist the the extension will match the original file's extension. 

Should we offer some flexibility here - should we try to find both `.yml` and `.yaml` installer manifests that match our `<x>.y(a)ml.patch` ?

internal/manifests/manifests.go
```diff
@@ -91,6 +91,11 @@ func (m *Manifests) CreateClusterManifestInternal(ctx context.Context, params op
 		if !json.Valid(manifestContent) {
 			return nil, common.NewApiError(http.StatusBadRequest, errors.New("Manifest content has an illegal JSON format"))
 		}
+	} else if strings.HasSuffix(fileName, ".yaml.patch") || strings.HasSuffix(fileName, ".yml.patch") {
```
omertuc, line 94: IIUC, users have no way to give us 2 or more patch files that patch the same file, is that right? Can we come up with a design that will allow this? Maybe allow arbitrary descriptions after `_patch`? Something like `<original_name>.yaml.patch_fix_schedulable_masters`

omertuc: These can then be sorted alphabetically to enforce ordering, which would be a bonus

internal/ignition/ignition.go
flaper87: > Should we offer some flexibility here - should we try to find both .yml and .yaml installer manifests that match our <x>.y(a)ml.patch ?

Yes, good point. Let me add this. thanks.

internal/ignition/ignition.go
flaper87: mmh, I think we can add this too. I added a `Remove` in the `applyManifestPatches` function to remove the `.patch` files and prevent from injecting those in the ignition file. I think adding this remove (with a defer) here also makes sense since at the end of this call we should've already generated the ignition file.

internal/manifests/manifests.go
flaper87: I like this, thanks for the suggestion, thanks. Let me implement this on this PR.

internal/ignition/ignition.go
flaper87: I was looking at this and decided not to look for both `yaml` and `yml` but rather expect the user to use the same extension as the manifest that should be patched.

For example, for `cluster-scheduler-02-config.yml` then `cluster-scheduler-02-config.yml.patch(_something)` should be used. A couple of reasons why I think this is probable a better approach:

1. It keeps consistency across the names of the manifests that will be patched
2. It makes the code simpler and easier to maintain.

The downside is that it may not super obvious to users that this is the case and they may push a `.yaml.patch` to patch a `.yml` manifest. 

Another option is to only allow `.yml.patch` but I'm not 100% sure that OpenShift creates only `.yml` files and this would limit us in the future. If we can confirm that OCP only generates `.yml` files, then we may be able to go ahead with this strategy.

internal/ignition/ignition.go
```diff
@@ -384,13 +384,21 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte,
 		log.Infof("adding manifest %s to working dir for cluster %s", manifest, g.cluster.ID)
 		err = g.downloadManifest(ctx, manifest)
 		if err != nil {
-			_ = os.Remove(filepath.Join(g.workDir, "manifests"))
-			_ = os.Remove(filepath.Join(g.workDir, "openshift"))
+			defer os.Remove(filepath.Join(g.workDir, "manifests"))
```
omertuc, line 387: `defer` doesn't do anything here

What I had in mind  was to run:

```go
defer func() {
	if err != nil {
		os.Remove(filepath.Join(g.workDir, "manifests"))
		os.Remove(filepath.Join(g.workDir, "openshift"))
	}
}
```

In the beginning of the function. This way if any error happens we get automatic cleanup in all error branches

flaper87: :facepalm:  my bad, now I see what you meant.

internal/ignition/ignition.go
omertuc: >may not super obvious to users

We can solve that by failing an installation in which we don't find an installer manifest matching one of the user patches. It's a good thing anyway, to catch user typos in file names. Tell the user "Either remove this patch or we will not continue installation, because we can't find a file matching it". Bonus points if you implement a handy little "did you mean: <file name>" with some off-the-shelf text distance library

>but I'm not 100% sure that OpenShift creates only .yml files

Yeah I'm not willing to bet on that, let's not make this assumption

flaper87: I like the proposal. 

Considering the urgency of this fix, would you be ok with me implementing your latest suggestions in a follow-up PR?

omertuc: Sure, it's on you to create the ticket/follow up on it, I won't have it in my TODO

internal/manifests/manifests.go
omertuc, line 94: This still looks like it wouldn't match `patch_*`

internal/network/manifests_generator.go
```diff
@@ -213,8 +207,8 @@ func (m *ManifestsGenerator) AddChronyManifest(ctx context.Context, log logrus.F
 }
 
 func (m *ManifestsGenerator) AddSchedulableMastersManifest(ctx context.Context, log logrus.FieldLogger, cluster *common.Cluster) error {
-	content := []byte(schedulableMastersManifest)
-	schedulableMastersManifestFile := "50-schedulable_masters.yaml"
+	content := []byte(schedulableMastersManifestPatch)
+	schedulableMastersManifestFile := "cluster-scheduler-02-config.yml.patch"
```
omertuc, line 211: Let's add a `_description`?

internal/common/common.go
```diff
@@ -120,14 +120,7 @@ func IsImportedCluster(cluster *Cluster) bool {
...
 func AreMastersSchedulable(cluster *Cluster) bool {
-	// If the topology forces schedulable masters (i.e. there are no workers),
-	// SchedulableMastersForcedTrue will be true, but also it will be enabled
-	// by default in the installer. Since overriding it currently causes a
-	// failure due to a conflict starting with 4.12, we prefer to avoid it. We
-	// only need to override the installer when the user has set
-	// SchedulableMasters explicitly and that is not already the default in the
-	// installer.
-	return !swag.BoolValue(cluster.SchedulableMastersForcedTrue) && swag.BoolValue(cluster.SchedulableMasters)
+	return swag.BoolValue(cluster.SchedulableMastersForcedTrue) || swag.BoolValue(cluster.SchedulableMasters)
```
flaper87, line 123: @zaneb FYI, with this PR I'm now reverting the change you applied since we can apply partial patches to manifests.

osherdp:
Not sure what's going on with the jira-lifecycle-plugin, anyway removing the label manually for now

gamli75 (approved):

zaneb:
Let's see how the bot handles this...
/cherry-pick release-4.12

PR #4589, opened 2022-11-07
NO-ISSUE: Make code generation faster

omertuc:
# Overview

These changes improve the time it takes to perform `skipper make generate`.
On my machine, the improvement is from 93 seconds down to just 30 seconds.

This is achieved mainly through introducing concurrency to the
generation flow. Another thing that caused slowness is that both
`generate-go-client` and `generate-go-server` validated the swagger
file, which takes quite a bit of time. Instead I changed it so that
none of them validate, and I introduced a new separate swagger
validation step (that runs in parallel to them).

# Concurrency

## Make `-j`

To achieve concurrency, I have simply made it so that `make generate`
re-invokes the Makefile but with the `-j30` flag, which tells make to
use 30 separate threads to process the make targets. (I'm not sure how
much of this workload is CPU-bound, so I wasn't sure whether to make it
the same as the number of cores or simply just constant 30. I guess this
can be adjusted in the future).

## Tree

But it was not so simple, because some of our generate targets depend on
other generate targets to finish before they themselves can be invoked
without issues. Simply adding `-j30` is insufficient because it means
all targets would run at the same time, and some of them with
dependencies on the others would fail.

To fix this, I've re-organized the targets by making some of them
prerequisites of others, so they form a sort of a tree. At the root of
that tree we have `generate-bundle` which directly and indirectly
depends on all others. That means that simply invoking `generate-bundle`
will now cause make to generate everything. This is not ideal, because
it means users can't just generate a single thing in isolation. But on
the other hand, generating everything is now so fast that it's not that
big of a deal. This can be improved in the future by making fake
targets.

## Go Generate

We used to simply call `go generate <pkg> <pkg> <pkg>...` and pass it
all of our packages in just one call. Instead I've made it so that `go generate <pkg>` for each package is its own Makefile target. This means
Makefile will now run `go generate <pkg>` for each of our packages in
parallel.

### But `mockgen` doesn't like it...

As you can see in the diff, I've changed the `mockgen` output path
extension to `.generated_go` instead of `.go`. This is because in order
to generate mocks, `mockgen` actually creates a temporary go package
that imports the mocked `interface`'s package and uses reflection to
generate the actual mock. `mockgen` then `go build`s this temporary
package, and runs it to produce the mock go file. The issue with this
approach is that since that temporary package imports the mocked
`interface`'s package, which is really just a simple Assisted package,
and that Assisted package in turn imports other Assisted packages, it
means every instance of `mockgen` (we run one for each of our packages)
has to also build and read code from other packages. But since other
instances of `mockgen` are actively *writing code* in those packages
(remember, we're generating `mock_*.go` files) at the same time, what
happens in practice is that `mockgen` is actually trying to build code
that's actively being changed! This causes `mockgen` (or more precisely
`go build`) to crash and burn ungracefully.

The solution is renaming the `mockgen` output path extension to
`.generated_go`, this way the files being modified by `mockgen` are not
actual Go files, so they are ignored by the `mockgen` `go build`
processs. After all generation is done we simply batch rename all of
them to have the correct `.go` extension.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


hack/Makefile
```diff
@@ -1,21 +1,33 @@
...
 	rm -rf restapi/
-	goswagger generate server --template=stratoscale -f swagger.yaml
+	goswagger generate server --skip-validation --template=stratoscale -f swagger.yaml
 
 generate-python-client:
 	mkdir -p ${BUILD_FOLDER}
 	./hack/generate.sh generate_python_client
 
-generate-mocks:
+packages-for-generate = $(addsuffix -go-generate-package, $(shell go list ./... | grep -v 'assisted-service/models\|assisted-service/client\|assisted-service/restapi'))
```
filanov, line 19: will it work on submodels? 
I know that right now we don't have anything to mock there but if it will be in the future then no one will catch it.
`go list ./...` will skip `api` directory because it contain go.mod

omertuc: This was already the case, I didn't change anything in that regard

eranco74:
@omertuc is this still relevant?

omertuc:
Yes but the CI keeps ooming it :(

omertuc:
sadly this uses too much memory even though it makes things faster. dropping it

PR #4613, opened 2022-11-14
MGMT-8357: Support kind hub-cluster for e2e tests

osherdp:
This change is about:
* Enable running e2e tests via assisted-test-infra when hub-cluster is a
  kind cluster.
* There's another side of this automation in assisted-test-infra
  repository, which depends on this change.
* A bit of cleaning-up, as some stuff on those automation scripts are
  pretty yucky. For example: removing confusing k3d binary mount,
  refactor code to have a less-nested indentation, etc.
* Kind is able to mirror ports from the kubernetes node to the external
  host. We have chosen to mirror 80->80 (I can accept other suggestions
  if anyone rejects that). In addition, we have an ingress-controller,
  meaning we just use <host>:80 to access
  assisted-service/assisted-image-service. This should explain most of
  the changes here. I preferred usage of ingress instead of
  load-balancer services strategies like Metallb to simplify the
  networking architecture and also to make it work smoothly on non-Linux
  machines.

Those things are out of scope:
* Removing 'minikube' code or using 'kind' by default. To make things
  easier we'd like to transition slowly and only remove 'minikube'
  support support when we're fully ready for it.
* As a consequence of point 1, code will still be yucky for now (but I
  do want to continue and simplify it in time. Any easy change in the
  scope of this PR will be accepted gladly.
* Supporting subsystem or debug mode. Those will be considered next.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] Reviewer's test appreciated - I'll need someone with mac to give this whole setup a try
- [x] Waiting for CI to do a full test run - mainly a validation
- [x] Manual (Elaborate on how it was tested) - tested together with the assisted-test-infra code changes
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/skipper.yaml
+++ b/skipper.yaml
@@ -5,9 +5,7 @@ containers:
   assisted-service-build: Dockerfile.assisted-service-build
 volumes:
   # programs
-  - /usr/bin/minikube:/usr/bin/minikube
-  - /usr/local/bin/minikube:/usr/local/bin/minikube
-  - /usr/local/bin/k3d:/usr/local/bin/k3d
+  - $(which minikube || echo /usr/local/bin/minikube):/usr/local/bin/minikube
 
   # config
   - $MINIKUBE_HOME:$MINIKUBE_HOME
--- a/tools/deploy_assisted_installer_configmap.py
+++ b/tools/deploy_assisted_installer_configmap.py
@@ -2,6 +2,7 @@ import argparse
 import json
 import os
 import yaml
+import socket
 import subprocess
 
 import deployment_options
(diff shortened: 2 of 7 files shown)


osherdp:
```
error: Requested operation is not valid: Domain already contains a disk with that address
```
seems like an existing filed flake

adriengentil:
> Kind is able to mirror ports from the kubernetes node to the external
> host
That means we could get rid of xined?

adriengentil (approved):

osherdp:
> > Kind is able to mirror ports from the kubernetes node to the external
> > host
> 
> That means we could get rid of xined? (and all the tunnel and port-forward we do)

correct. Both of xinetd and the tunnel won't be needed when we move to kind. you can see https://github.com/openshift/assisted-test-infra/pull/1915 for the port-mirroring configuration
there's another thing I'm currently not handling which is the registry, and I'm not sure yet if we'd need to have a registry because kind have a cool command to copy images to the cluster

PR #4616, opened 2022-11-14
MGMT-12332: Subsystem tests for discovery kernel arguments

ori-amizur:
REST API only
This commit also adds "x-omitempty: false" in the swagger for kernel
arguments.  This causes the kernel arguments to always be part of the
infra-env register and update payloads.  This was needed to enable clear
operation on kernel arguments.
When clearing kernel arguments, empty slice is sent to indicate the request to clear them.
This is opposed to null value when the kernel arguments should not be manipulated.
When omitempty is present, empty slice and null slice are ignored and
not sent in the payload.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @filanov

### DIFF
--- a/subsystem/infra_env_test.go
+++ b/subsystem/infra_env_test.go
@@ -3,6 +3,7 @@ package subsystem
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"io"
 	"net/http"
 	"os"
@@ -14,6 +15,7 @@ import (
 	"github.com/go-openapi/swag"
 	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	"github.com/openshift/assisted-service/client/installer"
 	"github.com/openshift/assisted-service/internal/bminventory"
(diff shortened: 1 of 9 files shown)


subsystem/infra_env_test.go
```diff
@@ -46,6 +48,65 @@ var internalRegisterInfraEnv = func(clusterID *strfmt.UUID, imageType models.Ima
...
+			kargs := models.KernelArguments{
+				{
+					Operation: operation,
+					Value:     value,
+				},
+			}
+			_, err := register(kargs)
+			Expect(err).To(HaveOccurred())
+		},
+		Entry("operation is replace", models.KernelArgumentOperationReplace, "p1"),
```
filanov, line 103: from the name it's not clear what is not supported in this case? 
is it the operator or the argument? 
same question for the next entry

subsystem/infra_env_test.go
```diff
@@ -83,6 +144,90 @@ var _ = Describe("Infra_Env", func() {
...
+								Operation: operation,
+								Value:     value,
+							},
+						},
+					},
+				}
+				_, err := userBMClient.Installer.UpdateInfraEnv(ctx, updateParams)
+				Expect(err).To(HaveOccurred())
+			},
+			Entry("operation is replace", models.KernelArgumentOperationReplace, "p1"),
```
filanov, line 224: the name of the case is not clear, cannot understand what is not supported.

subsystem/infra_env_test.go
filanov: maybe `unsupported replace operation` it wasn't clear because i was thinking that replace of p1 is invalid because it doesn't exist in the db

swagger.yaml
```diff
@@ -6324,6 +6324,7 @@ definitions:
 
   kernel_arguments:
     description: List of kernel arugment objects that define the operations and values to be applied.
+    x-omitempty: false
```
filanov, line 6327: please add an explanation in the commit message

subsystem/infra_env_test.go
ori-amizur: OK

filanov (approved):

PR #4617, opened 2022-11-14
NO-ISSUE: Explicitly finish mock controllers in cluster transition tests

jhernand:
Currently some of the mock controllers in the cluster transition tests aren't finished. As a result some of the expectations aren't verified, and some of the tests are incorrect. With the introduction of Ginkgo 2 the controllers will be automatically finished, and that will make these tests fail. In order to reduce the size of that migration to Gingo 2 this patch fixes those tests and changes them to explicitly finish the mock controllers.

The patch also brings the `BeforeEach` and `AfterEach` blocks that create the mock controllers closer, so that it will hopefully be harder to forget to add the cleanup code.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/transition_test.go
+++ b/internal/cluster/transition_test.go
@@ -58,6 +58,11 @@ var _ = Describe("Transition tests", func() {
 		clusterId = strfmt.UUID(uuid.New().String())
 	})
 
+	AfterEach(func() {
+		ctrl.Finish()
+		common.DeleteTestDB(db, dbName)
+	})
+
 	Context("cancel_installation", func() {
 		BeforeEach(func() {
 			capi = NewManager(getDefaultConfig(), common.GetTestLog(), db, eventsHandler, nil, mockMetric, nil, nil, operatorsManager, nil, nil, nil, nil)
@@ -80,7 +85,6 @@ var _ = Describe("Transition tests", func() {
 				Cluster: models.Cluster{ID: &clusterId, Status: swag.String(models.ClusterStatusInsufficient)},
 			}
 			Expect(db.Create(&c).Error).ShouldNot(HaveOccurred())
-			mockMetric.EXPECT().ClusterInstallationFinished(gomock.Any(), models.ClusterStatusCancelled, models.ClusterStatusInsufficient, c.OpenshiftVersion, *c.ID, c.EmailDomain, c.InstallStartedAt)
 			replay := capi.CancelInstallation(ctx, &c, "", db)
 			Expect(replay).Should(HaveOccurred())
 			Expect(int(replay.StatusCode())).Should(Equal(http.StatusConflict))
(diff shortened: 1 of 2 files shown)


internal/cluster/transition_test.go
```diff
@@ -175,7 +173,7 @@ var _ = Describe("Transition tests", func() {
 			{
 				name:                       "available builtin operators, progress OLM -> finalizing",
 				uploadKubeConfig:           true,
-				updateSuccessfullyFinished: true,
+				updateSuccessfullyFinished: false,
```
carbonin, line 176: This feels like it's significantly changing the expectations of this test.
Should we look at if this is a bug?

jhernand: No objection to handle this as a bug if needed. But I believe that the test is incorrect: a cluster that is in "finalizing" isn't completely installed. Is that wrong?

filanov: as far as i know OLM is not a blocker for installation finalizing mean is waiting for required operators. 
@gamli75 maybe you are familiar with that more them me

internal/cluster/transition_test.go
```diff
@@ -3891,6 +3909,29 @@ var _ = Describe("Log Collection - refresh cluster", func() {
 		clusterId = strfmt.UUID(uuid.New().String())
 	})
 
+	AfterEach(func() {
```
filanov, line 3912: after each a new cluster is created? why?
I know that you just moved the code but i'm not sure why it's need to be done in after each

jhernand: That is how these tests are written: they run part of the setup and all the verifications inside the shared "AfterEach" block instead of inside the "It" block. I'd say it is a bit surprising, and it would probably be better to do that inside "It" and leave "AfterEach" only for cleanups. But I didn't want to change more things inside this patch.

internal/cluster/transition_test.go
jhernand: @filanov note that this test is checking a transition from "finalizing" to "finalizing", and I think that is a nop transition, so it doesn't update metrics.

gamli75: as @filanov wrote - OLM is not a blocker for successful installation

jhernand: @gamli75 you are right, I misunderstood the reason why this test is not updating the metrics. The reason is that the test transitions from the "finalizing" state to the same "finalizing" state, and that is a no-op transition, so nothing is done, in particular no metric is updated. So I still think this change is correct.

filanov (approved):

PR #4621, opened 2022-11-15
MGMT-11266: nutanix: remove worker MachineSet too

vrutkovs:
Worker machineset needs to be removed so that MachineAPI operator would not go degraded due to placeholder credentials on install


romfreiman:
do we have it covered in the CI - 3M + 2W ?

vrutkovs:
https://github.com/openshift/release/pull/34023 would cover that

carbonin (commented):
Looks good, but do we have unit tests?

vrutkovs:
Sadly no, same as for vsphere.

I'd rather add them in the follow up so that we wouldn't block an important bugfix for Nutanix

carbonin (approved):

PR #4624, opened 2022-11-15
MGMT-12555: Try avoid deadlocks when updating cluster or host

ori-amizur:
Daedlock occurs when there are 2 resources (a,b)  to be locked by 2 users.  If user 1 locks resource "a", then user 2 locks resource "b", then user 1 tries to lock resource "b" and user 2 tries to lock resource "a". This may occure between 2 different database transactions.  In case of database deadlock, the database engine selects a victim and terminates its transaction.
One of the ways to avoid deadlock is to keep always the same locking order.  This is what was done here.  The 2 resources are cluster and host.  The change verifies that in case cluster and host need to be locked in a transaction, the cluster is locked before the host.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @filanov

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4684,7 +4684,23 @@ func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installe
 		}
 	}()
 
-	dbHost, err := common.GetHostFromDB(transaction.AddForUpdateQueryOption(tx), params.InfraEnvID.String(), params.NewHostParams.HostID.String())
+	infraEnv, err := common.GetInfraEnvFromDB(tx, params.InfraEnvID)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get infra env: %s", params.InfraEnvID)
+		return common.GenerateErrorResponder(err)
+	}
+
+	var cluster *common.Cluster
+	var c *models.Cluster
+
+	// The query for cluster must appear before the host query to avoid potential deadlock
+	cluster, err = b.getBoundClusterForUpdate(tx, infraEnv, params.InfraEnvID, *params.NewHostParams.HostID)
+	if err != nil {
+		log.WithError(err).Errorf("Bound Cluster get")
+		return common.NewApiError(http.StatusInternalServerError, err)
+	}
+
+	_, err = common.GetHostFromDB(transaction.AddForUpdateQueryOption(tx), params.InfraEnvID.String(), params.NewHostParams.HostID.String())
 	if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
 		log.WithError(err).Errorf("failed to get host %s in infra-env: %s",
 			*params.NewHostParams.HostID, params.InfraEnvID.String())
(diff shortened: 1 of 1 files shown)


internal/bminventory/inventory.go
```diff
@@ -4667,9 +4667,15 @@ func (b *bareMetalInventory) GetInfraEnvByKubeKey(key types.NamespacedName) (*co
 }
 
 func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installer.V2RegisterHostParams) middleware.Responder {
+	var clusterID strfmt.UUID
+	var err error
+
 	log := logutil.FromContext(ctx, b.log)
 	log.Infof("Register host: %+v", params)
 
+	if clusterID, err = b.getClusterIDFromHost(*params.NewHostParams.HostID, params.InfraEnvID); err != nil {
```
filanov, line 4676: how will it work? the host is not registered yet, you need to check the infra-env here not the hosts table

filanov: `getBoundCluster` is looking first at the infra env 
`getClusterIDFromHost` is looking at the hosts table
what did i miss?

ori-amizur: Changed to make it clearer.
Anyway, after the change getClusterIDFromHost will be called only if infra-env doesn't have cluster-id.
Before the getClusterIDFromHost  was called, but it was used only if infra-env doesn't have cluster-id (but it is irrelevant now).

internal/bminventory/inventory.go
```diff
@@ -4670,6 +4670,9 @@ func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installe
 	log := logutil.FromContext(ctx, b.log)
 	log.Infof("Register host: %+v", params)
 
+	if params.NewHostParams.HostID == nil {
```
filanov, line 4673: isn't that validated on the swagger level? https://github.com/openshift/assisted-service/blob/96f0f2434fbadebf15b7edb697ec035488c3c872/swagger.yaml#L3820-L3823

internal/bminventory/inventory.go
```diff
@@ -5437,6 +5456,20 @@ func (b *bareMetalInventory) V2UpdateHostInternal(ctx context.Context, params in
 		}
 	}()
 
+	// Get the bound cluster id out of the transaction.  Note: transaction is not used to query the cluster id
+	if clusterID, err = b.getClusterIDFromHost(params.HostID, params.InfraEnvID); err != nil {
```
filanov, line 5460: this function is using `b.db` wouldn't that be a problem because you are in a middle of transaction?

internal/bminventory/inventory.go
ori-amizur: OK

internal/bminventory/inventory.go
ori-amizur: Changed

internal/bminventory/inventory.go
ori-amizur: Removed

internal/bminventory/inventory.go
```diff
@@ -5437,6 +5455,20 @@ func (b *bareMetalInventory) V2UpdateHostInternal(ctx context.Context, params in
 		}
 	}()
 
+	// Get the bound cluster id out of the transaction.  Note: transaction is not used to query the cluster id
```
filanov, line 5458: this comment is incorrect now

ori-amizur: OK

filanov (approved):

PR #4630, opened 2022-11-15
Remove ValidateAccessToMultiarch function

carbonin:
This function is causing more problems than it is solving.
It doesn't really belong in `versions` and the caller in versions is having to match the error message string to work properly.

Using the defined `Authorizer` interface directly makes more sense.

Doing this as a part of cleaning up and refactoring the `versions` package.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


pkg/auth/authz.go
```diff
@@ -55,3 +58,17 @@ func NewAuthzHandler(cfg *Config, ocmCLient *ocm.Client, log logrus.FieldLogger,
...
+func ValidateAccessToMultiarch(ctx context.Context, authzHandler Authorizer) error {
+	var err error
+	var multiarchAllowed bool
+
+	multiarchAllowed, err = authzHandler.HasOrgBasedCapability(ctx, ocm.MultiarchCapabilityName)
+	if err != nil {
+		return common.NewApiError(http.StatusInternalServerError, fmt.Errorf("error getting user %s capability, error: %w", ocm.MultiarchCapabilityName, err))
+	}
+	if !multiarchAllowed {
+		return common.NewApiError(http.StatusBadRequest, errors.Errorf("%s", "multiarch clusters are not available"))
```
slaviered, line 71: 404 will be more appropriate here

slaviered: How about making it part of hasAccessTo with an object that represent a multiarch. and the application level can then invoke it pretty much as it invokes ValidateAccessToMultiarch.

pkg/auth/authz_test.go
```diff
@@ -0,0 +1,77 @@
...
+			Role:         ocm.UserRole,
+		}
+		ctx = context.WithValue(context.Background(), restapi.AuthKey, payload)
+
+		mockAccountsMgmt.EXPECT().CreateSubscription(ctx, gomock.Any(), gomock.Any()).Return(&amgmtv1.Subscription{}, nil)
+		authzHandler = NewAuthzHandler(cfg, mockOcmClient, common.GetTestLog().WithField("pkg", "auth"), db)
+	})
+
+	AfterEach(func() {
+		common.DeleteTestDB(db, dbName)
```
jhernand, line 54: Consider moving `ctrl` to the `var (...)` block above, and call `ctrl.Finish()` here, otherwise expectations will not be completely validated.

pkg/auth/authz.go
carbonin: I can change it, but this is just a copy and paste.

pkg/auth/authz_test.go
carbonin: Thanks for the reminder

pkg/auth/authz.go
carbonin: I think I'd rather just remove the function entirely. It's just some error handling wrapping a single call.

One caller is matching based on the error string so it would be better for it to be able to define how it handles errors. Then the other caller could implement this function or not.

carbonin:
@slaviered can you take another look at this one?

mkowalski:
Of course this makes sense, thanks! Just please note that there is a doc https://github.com/openshift/assisted-service/blob/master/docs/dev/feature-per-customer.md that refers to the old way of implementing, so this PR should make sure that the doc is not outdated

carbonin:
Yeah, I saw that doc, but it's just using that as an example I'm not sure it matters that the function doesn't actually exist any more in source.

internal/versions/versions.go
```diff
@@ -119,14 +118,11 @@ func (h *handler) V2ListSupportedOpenshiftVersions(ctx context.Context, params o
...
-				if err := h.ValidateAccessToMultiarch(ctx, h.authzHandler); err != nil {
-					if strings.Contains(err.Error(), "multiarch clusters are not available") {
-						continue
-					} else {
-						return common.GenerateErrorResponder(err)
-					}
+				var err error
+				hasMultiarchAuthorization, err = h.authzHandler.HasOrgBasedCapability(ctx, ocm.MultiarchCapabilityName)
+				if err != nil {
+					return common.GenerateErrorResponder(err)
```
slaviered, line 124: Do not return the original error to the user because it will expose the inner details of the security related call. I would replace the error with a message that states that you can not access multiarch cluster and has a 404 code

internal/bminventory/inventory.go
```diff
@@ -497,7 +497,14 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 		cpuArchitecture = common.MultiCPUArchitecture
 		// (MGMT-11859) Additional check here ensures that the customer cannot just guess a version
 		//              with multiarch in order to get access to that release payload.
-		if err = b.versionsHandler.ValidateAccessToMultiarch(ctx, b.authzHandler); err != nil {
+		var multiarchAllowed bool
+		multiarchAllowed, err = b.authzHandler.HasOrgBasedCapability(ctx, ocm.MultiarchCapabilityName)
+		if err != nil {
+			err = common.NewApiError(http.StatusInternalServerError, fmt.Errorf("error getting user %s capability, error: %w", ocm.MultiarchCapabilityName, err))
```
slaviered, line 503: It is not advisable to expose the details of the auth capability we use to the user. I would change the error message here to something more generic .

slaviered: Also I do think that a utility function that hide this code will be more readable. The former ValidateAccessToMultiarch was very clear. If you like a different error handling you can inject an error message but personally I don't see why we need a different errors here and in versionhandler.

carbonin: Here you're probably right, we likely don't need different errors, but in the version handler we have different behavior for error vs if multi arch is not available. This is what made sharing the implementation rather strange.

internal/versions/versions.go
carbonin: Is there any reason to return an error here in this case? Or can we just skip this image if the ocm call fails (as we do for if the authorization comes back `false`?

If we're changing behavior we might as well think about everything while we're here :sweat_smile: 
@mkowalski ?

carbonin: I ask because no one was specifically asking for a multiarch cluster in this case, they were just asking what versions are available to them. So it seems a bit strange to error with something like "multiarch not allowed" if they don't actually care about multiarch.

mkowalski: It’s correct here to say that we mustn’t return an error but rather transparently hide the image that user has no access to

carbonin:
/hold

Working on some comments still

carbonin:
@mkowalski @slaviered any other issues here?

PR #4645, opened 2022-11-18
Refactor the versions handler into a proper cache

carbonin:
As it is today the versions handler does too much and doesn't function well as a cache when we need to add release images from `ClusterImageSets`.

This has caused a few issues such as https://issues.redhat.com/browse/MGMT-8279 and https://issues.redhat.com/browse/MGMT-11956. The solution to these has been to add the release image from more places, but that means the issue will continue to occur.

Additionally the way the cluster deployment controller added to the cache, but the main inventory code read from it was going to make it difficult to separate out the controllers from the API in the future.

This effort began by splitting unrelated functionality off of the existing `versions.Handler` interface.
Really it did several unrelated things that are split into new structs/interfaces:
1. Track and cache a list of release images (`versions.Handler`)
2. ~~Track OS images (`versions.OSImages`)~~ - https://github.com/openshift/assisted-service/pull/4737
3. ~~Implement API controllers for serving currently used container images and release versions (`versions.apiHandler`)~~ - https://github.com/openshift/assisted-service/pull/4728
4. ~~Check authorization for multiarch release images~~ - https://github.com/openshift/assisted-service/pull/4630

This PR implements a better caching mechanism where the versions cache handles refreshing itself from the kubeAPI when a `get` operation misses. This means that we don't have to rely on a controller to specifically add an image just before we install to ensure it will be present when the install actually runs while also not increasing the amount of times we query for `ClusterImageSets`.

Each commit in this PR is individually reviewable (I recommend looking at them individually and setting up the diff viewer to ignore whitespace changes).

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

1. Deployed assisted with kube-api
2. Created an SNO cluster without late-binding (infraenv, agentclusterinstall, clusterdeployment, clusterimageset, pullsecret)
3. Deleted the assisted-service pod (to ensure release image cache was empty)
4. Booted and approved an agent
5. Installed cluster

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -24,7 +24,7 @@ type MustGatherVersions map[string]MustGatherVersion
 type Handler interface {
 	GetReleaseImage(ctx context.Context, openshiftVersion, cpuArchitecture, pullSecret string) (*models.ReleaseImage, error)
 	GetDefaultReleaseImage(cpuArchitecture string) (*models.ReleaseImage, error)
-	AddReleaseImage(releaseImageUrl, pullSecret string) (*models.ReleaseImage, error)
+	GetReleaseImageByURL(ctx context.Context, url, pullSecret string) (*models.ReleaseImage, error)
 	GetMustGatherImages(openshiftVersion, cpuArchitecture, pullSecret string) (MustGatherVersion, error)
 	ValidateReleaseImageForRHCOS(rhcosVersion, cpuArch string) error
 }
@@ -123,7 +123,7 @@ func (h *handler) GetReleaseImage(ctx context.Context, openshiftVersion, cpuArch
 			}
 		}
 		if !existsInCache {
-			_, err := h.AddReleaseImage(clusterImageSet.Spec.ReleaseImage, pullSecret)
+			_, err := h.addReleaseImage(clusterImageSet.Spec.ReleaseImage, pullSecret)
 			if err != nil {
 				h.log.WithError(err).Warnf("Failed to add release image %s", clusterImageSet.Spec.ReleaseImage)
 			}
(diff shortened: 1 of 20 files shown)


internal/versions/osimages.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/go-openapi/swag"
+	"github.com/hashicorp/go-version"
+	"github.com/openshift/assisted-service/internal/common"
+	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
+	"github.com/thoas/go-funk"
+)
+
+//go:generate mockgen --build_flags=--mod=mod -package versions -destination mock_osimages.go -self_package github.com/openshift/assisted-service/internal/versions . OSImages
+type OSImages interface {
```
danielerez, line 16: Nice:) It's a good a idea to extract these from Version. So maybe later we should rename 'Versions' to 'ReleaseImages' interface, to make it consistent.

internal/versions/versions.go
```diff
@@ -206,37 +206,30 @@ func (h *handler) ValidateReleaseImageForRHCOS(rhcosVersion, cpuArchitecture str
...
-		// that the image references to.
-		cpuArchitectures, err = h.releaseHandler.GetReleaseArchitecture(h.log, releaseImageUrl, "", pullSecret)
-		if err != nil {
-			return nil, err
-		}
-		h.log.Debugf("For release image %s detected architecture: %s", releaseImageUrl, cpuArchitectures)
+	// Get CPU architecture from release image. For single-arch image the returned list will contain
+	// as single entry with the architecture. For multi-arch image the list will contain all the architectures
+	// that the image references to.
+	cpuArchitectures, err := h.releaseHandler.GetReleaseArchitecture(h.log, releaseImageUrl, "", pullSecret)
```
danielerez, line 230: So shouldn't we now check if the releaseImageUrl exists in cache first?
I.e. To avoid calling GetReleaseArchitecture every time, as this func is not very optimized (might execute two calls, for both skopeo and oc)

PR #4650, opened 2022-11-21
OKD-90: Support automated extraction of OKD RPMs

vrutkovs:
If payload has `okd-rpms` image Assisted Service should automatically apply OKD Ignition overrides. This would enable us to add OKD payload to AI without additional configuration option

- [x] Enhancement 

- [x] Cloud

## How was this code tested?

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/ignition/ignition.go
+++ b/internal/ignition/ignition.go
@@ -35,8 +35,10 @@ import (
 	"github.com/openshift/assisted-service/internal/installercache"
 	"github.com/openshift/assisted-service/internal/manifests"
 	"github.com/openshift/assisted-service/internal/network"
+	"github.com/openshift/assisted-service/internal/oc"
 	"github.com/openshift/assisted-service/internal/operators"
 	"github.com/openshift/assisted-service/internal/provider/registry"
+	"github.com/openshift/assisted-service/internal/versions"
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/assisted-service/pkg/auth"
 	logutil "github.com/openshift/assisted-service/pkg/log"
@@ -148,10 +150,6 @@ do
     sleep 5
 done
 mnt=$(podman image mount "${RPMS_IMAGE}")
-# Extract machine-config-daemon binary
-cp -rvf ${mnt}/binaries/machine-config-daemon /usr/local/bin/machine-config-daemon
-chmod a+x /usr/local/bin/machine-config-daemon
-restorecon -Rv /usr/local/bin/machine-config-daemon
 # Install RPMs in overlayed FS
 mkdir /tmp/rpms
 cp -rvf ${mnt}/rpms/* /tmp/rpms
@@ -168,6 +166,9 @@ fi
 losetup -c /dev/loop0
 xfs_growfs /var
 mount -o remount,size=6G /run
+# Symlink kubelet pull secret
+mkdir -p /var/lib/kubelet
+ln -s /root/.docker/config.json /var/lib/kubelet/config.json
 `
 
 const okdHoldAgentUntilBinariesLanded = `[Unit]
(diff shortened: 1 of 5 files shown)


vrutkovs:
Still doing manual testing, but that should allow us to avoid special OKD config for Assisted Service starting OKD 4.12
cc @gmarcy  - could you try it out too?

internal/ignition/ignition.go
```diff
@@ -1435,6 +1441,22 @@ func SetHostnameForNodeIgnition(ignition []byte, host *models.Host) ([]byte, err
...
 }
 
+func (ib *ignitionBuilder) shouldAppendOKDFiles(ctx context.Context, infraEnv *common.InfraEnv, cfg IgnitionConfig) bool {
+	// OKD_RPMS_IMAGE explicitly set in config
+	if cfg.OKDRPMsImage != "" {
+		return true
+	}
+	// Selected payload contains `okd-rpms` image
+	releaseImage, err := ib.versionHandler.GetReleaseImage(infraEnv.OpenshiftVersion, infraEnv.CPUArchitecture)
+	if err != nil {
```
osherdp, line 1451: Does this case relevant to a "happy-flow" use-case, or is it a real indication of an unexpected error?

osherdp (changes requested):

internal/ignition/ignition.go
vrutkovs: Its unexpected error. If we can't inspect the release image for some reason we should not embed OKD Ignition bits and complicate troubleshooting (also, unittests are making use of it)

internal/ignition/ignition.go
```diff
@@ -1435,6 +1441,22 @@ func SetHostnameForNodeIgnition(ignition []byte, host *models.Host) ([]byte, err
 	return configBytes, nil
 }
 
+func (ib *ignitionBuilder) shouldAppendOKDFiles(ctx context.Context, infraEnv *common.InfraEnv, cfg IgnitionConfig) bool {
+	// OKD_RPMS_IMAGE explicitly set in config
+	if cfg.OKDRPMsImage != "" {
+		return true
+	}
+	// Selected payload contains `okd-rpms` image
```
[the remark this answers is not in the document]
vrutkovs: It kinda adds a note above the following lines where release payload is inspected and okd-rpms image pullspec is being extracted.

internal/ignition/ignition.go
osherdp: maybe we should log it or propagate the error?

vrutkovs:
/hold

For manual testing

osherdp (approved):

internal/ignition/ignition.go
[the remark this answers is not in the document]
omertuc: Oh okay. The current grammar in that comment implies that the comment is stating a fact, something that is already true at that point in the code, when in reality, IIUC, it's just stating what we're going to check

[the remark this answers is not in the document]
vrutkovs: Actually, a good point. Updated the comment so that its clearly a check about to be performed, not a fact yet

internal/ignition/ignition.go
```diff
@@ -168,6 +166,9 @@ fi
 losetup -c /dev/loop0
 xfs_growfs /var
 mount -o remount,size=6G /run
+# Symlink kubelet pull secret
+mkdir /var/lib/kubelet
```
omertuc, line 170: Maybe `-p` just in case?

vrutkovs:
/hold cancel

Verified manually that it works

internal/ignition/ignition.go
```diff
@@ -1435,6 +1440,24 @@ func SetHostnameForNodeIgnition(ignition []byte, host *models.Host) ([]byte, err
 	return configBytes, nil
 }
 
+func (ib *ignitionBuilder) shouldAppendOKDFiles(ctx context.Context, infraEnv *common.InfraEnv, cfg IgnitionConfig) (string, bool) {
+	// Use OKD override if OKD_RPMS_IMAGE explicitly set in config
+	if cfg.OKDRPMsImage != "" {
+		return cfg.OKDRPMsImage, true
+	}
+	// Check if selected payload contains `okd-rpms` image
+	releaseImage, err := ib.versionHandler.GetReleaseImage(infraEnv.OpenshiftVersion, infraEnv.CPUArchitecture)
```
carbonin, line 1449: It's not really correct to get a release image from an infraenv version.
This version attribute is poorly named, but it is the coreos version, not an openshift version.
It's likely that this will, at some point, not match a version in the release image.

Is there a better way to detect what you need? Or is it something we can only do if we're not using late-binding?

See my comment in a very similar bug [here](https://issues.redhat.com/browse/MGMT-12482?focusedCommentId=21220244&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21220244) for details. Essentially there's no reason for this to match any configured release image. Additionally it's valid for a user to discover hosts before adding any release image at all.

carbonin: Found this while trying to get https://github.com/openshift/assisted-service/pull/4645 ready for review FWIW

vrutkovs: Discussed in Slack. Its fine to leave as is, potentially late binding usecase would be broken for OKD - but this is acceptable for now

PR #4657, opened 2022-11-22
MGMT-12552: Add UT for day-2 agent status updates

eranco74:


internal/host/host_test.go
```diff
@@ -3794,6 +3794,121 @@ var _ = Describe("Installation stages", func() {
 			Expect(*h.Status).To(Equal(models.HostStatusAddedToExistingCluster))
 		})
 	})
+	It("UpdateInstallProgress test - day2 hosts with kubeapi enabled", func() {
```
filanov, line 3797: usually unit test don't have so many steps in a single test. 
an alternative is to use table flow https://onsi.github.io/ginkgo/#table-specs 
so you can make a context of this test, have a table for all the common cases an a different test in the context for the last case

PR #438, opened 2020-09-29
MGMT-2283 change unit-tests db into quay.io/ocpmetal/postgresql-12-ce…

filanov:
…ntos7

PR #4665, opened 2022-11-23
MGMT-12616: Reject register after install

jhernand:
This patch changes the service so that it returns a 403 error response when an agent tries to register when the cluster is already installed, as otherwise the agent will keep trying to register and spam both the log and the events.

Related: https://issues.redhat.com/browse/MGMT-12616
Related: https://issues.redhat.com/browse/MGMT-2359
Related: https://issues.redhat.com/browse/MGMT-1041

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Tested manuall installing a cluster and then rebooting one of the nodes with the installation ISO.

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/transition.go
+++ b/internal/host/transition.go
@@ -55,6 +55,7 @@ type TransitionHandler interface {
 	PostRefreshLogsProgress(progress string) stateswitch.PostTransition
 	PostRefreshReclaimTimeout(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error
 	PostRegisterDuringInstallation(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error
+	PostRegisterAfterInstallation(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error
 	PostRegisterDuringReboot(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error
 	PostRegisterHost(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error
 	PostResettingPendingUserAction(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error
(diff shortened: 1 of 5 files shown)


jhernand:
> We discussed exponential backoff for retries. Wouldn't that be safer? What if the user resets the cluster to try again?

If I understand correctly we discussed that for the loop that requests next instructions. That was already implemented and merged in #438. This is for registration, a different loop. I agree that it would be better to also retry for these registration errors, but I think this is convenient anyhow. I opened https://issues.redhat.com/browse/MGMT-12673 to track the change to the registration loop.

internal/host/transition.go
```diff
@@ -138,6 +139,28 @@ func (th *transitionHandler) PostRegisterDuringInstallation(sw stateswitch.State
...
+	}
+	_, ok = args.(*TransitionArgsRegisterHost)
+	if !ok {
+		return errors.New("PostRegisterAfterInstallation invalid argument")
+	}
+
+	return common.NewApiError(
+		http.StatusForbidden,
+		errors.New(
+			"Host is trying to register after the cluster has already been installed. "+
```
gamli75, line 155: @jhernand if the user will see this message in the UI - @nirfarkas should review it

avishayt:
1. 403 isn't the correct code, there is no issue with access rights here.
2. Why not implement the exponential backoff as a correct solution?

internal/host/transition.go
jhernand: Thanks @gamli75, I forgot that. The message will be part of an error event, so it is displayed to the user. @nirfarkas please review.

gamli75: @jhernand what do you mean "error event"? Did you verify that we don't spam user events?

jhernand:
> 1. 403 isn't the correct code, there is no issue with access rights here.

I agree with that, however it was reviewed and approved for another similar situation in #4665, so I assumed the same logic can be used here for a quick fix.

>     2. Why not implement the exponential backoff as a correct solution?

I have no objection to use retry with exponential backoff as a better solution, just rather not in this ticket/patch.

internal/host/transition.go
jhernand: I mean this:

image

That event will be generated only once because the agent will then freeze.

eranco74 (commented):
/lgtm

eranco74:
24 hours in merge pool?
Something isn't right.
@osherdp @adriengentil any idea what'sgoing on?

PR #4666, opened 2022-11-23
MGMT-11536: refresh cluster after host deletion

slaviered:
When a user deletes a host, the cluster validations do not run immediately. Therefore the UI presents an incorrect view of the validation status. For example, master count validation shows erroneous status.

That may lead to a user starting an installation when the cluster is actually not ready to be installed.

This PR adds a refresh cluster after host deletion so the validation will run before the result is returned to the UI.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -146,7 +146,7 @@ type InstallerInternals interface {
 	GetHostByKubeKey(key types.NamespacedName) (*common.Host, error)
 	InstallClusterInternal(ctx context.Context, params installer.V2InstallClusterParams) (*common.Cluster, error)
 	DeregisterClusterInternal(ctx context.Context, cluster *common.Cluster) error
-	V2DeregisterHostInternal(ctx context.Context, params installer.V2DeregisterHostParams) error
+	V2DeregisterHostInternal(ctx context.Context, params installer.V2DeregisterHostParams, interactivity Interactivity) error
 	GetCommonHostInternal(ctx context.Context, infraEnvId string, hostId string) (*common.Host, error)
 	UpdateHostApprovedInternal(ctx context.Context, infraEnvId string, hostId string, approved bool) error
 	V2UpdateHostInstallerArgsInternal(ctx context.Context, params installer.V2UpdateHostInstallerArgsParams) (*models.Host, error)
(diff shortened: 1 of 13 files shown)


subsystem/cluster_test.go
```diff
@@ -218,6 +219,22 @@ var _ = Describe("Cluster", func() {
 		Expect(c.Hosts[0].ID.String()).Should(Equal(h.ID.String()))
 	})
 
+	It("Deregister host triggers cluster validations", func() {
```
nmagnezi, line 222: Please also add a test for a negative case, where `refreshClusterStatus` returns an error.

internal/bminventory/inventory.go
```diff
@@ -2872,6 +2872,9 @@ func (b *bareMetalInventory) V2DeregisterHostInternal(ctx context.Context, param
 		hostutil.GetHostnameForMsg(&h.Host))
 
 	if h.ClusterID != nil {
+		if _, err = b.refreshClusterStatus(ctx, h.ClusterID, b.db); err != nil {
```
nmagnezi, line 2875: Won't that make the flow synchronous?
Did you check how long it takes for `refreshClusterStatus` to run?

nmagnezi (changes requested):

internal/bminventory/inventory.go
slaviered: This was exactly the intention of this PR. And yes this solution was approved. The Ux experience in SaaS takes precedence here.

subsystem/cluster_test.go
slaviered: It is not an interesting case since the update process just fails. What important here was to verify that the validations are returned synchronously.

nmagnezi: I disagree.
Unit tests are meant to assert possible outcomes.
We should assert that if cluster sync fails, we return an error here.

slaviered: Good point on both accounts. Adding a unit test

internal/bminventory/inventory.go
```diff
@@ -5239,7 +5242,7 @@ func (b *bareMetalInventory) BindHostInternal(ctx context.Context, params instal
 	return host, nil
 }
 
-func (b *bareMetalInventory) UnbindHostInternal(ctx context.Context, params installer.UnbindHostParams, reclaimHost bool) (*common.Host, error) {
+func (b *bareMetalInventory) UnbindHostInternal(ctx context.Context, params installer.UnbindHostParams, reclaimHost bool, interactivity Interactivity) (*common.Host, error) {
```
ori-amizur, line 5245: The interactivity is not used

internal/bminventory/inventory.go
```diff
@@ -2978,6 +2978,11 @@ func (b *bareMetalInventory) V2DeregisterHostInternal(ctx context.Context, param
 		hostutil.GetHostnameForMsg(&h.Host))
 
 	if h.ClusterID != nil {
+		if interactivity == Interactive {
+			if _, err = b.refreshClusterStatus(ctx, h.ClusterID, b.db); err != nil {
+				log.WithError(err).Warnf("Failed to refresh cluster after de-registerating host <%s>", params.HostID)
```
ori-amizur, line 2983: Only log ? Why not return error ?

slaviered: Per discussion with @avishayt we decided to fail gracefully in such case (which should be rare).

PR #4670, opened 2022-11-24
MGMT-12684 add envoy as sidecar to produce access log

rccrdpccl:
This PR will allow us to produce access logs without changing the service's code.

It will produce lines similar to:
```
{"response_flags":"-","user_agent":"curl/7.61.1","authority":"localhost:8090","x_forwarded_for":"-","protocol":"HTTP/1.1","path":"/api/assisted-install/v2/clusters/foobar","upstream_host":"127.0.0.1:8090","route_name":"-","response_code":"401","x_agent_name":"-"}
```

Signed-off-by: Riccardo piccoli <rpiccoli@redhat.com>

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

Run in integration environment.

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/openshift/template.yaml
+++ b/openshift/template.yaml
@@ -144,6 +144,20 @@ parameters:
 - name: LIVENESS_PROBE_INITIAL_DELAY_SECONDS
   value: "30"
   required: false
+- name: ENVOY_IMAGE
+  value: "quay.io/app-sre/envoyproxy"
+  required: false
+- name: ENVOY_TAG
+  value: "v1.24.0"
+  required: false
+- name: ENVOY_MEMORY_REQUEST
+  value: 256Mi
+- name: ENVOY_CPU_REQUEST
+  value: 10m
+- name: ENVOY_MEMORY_LIMIT
+  value: 512Mi
+- name: SVC_TARGET_PORT
+  value: "8900"
 apiVersion: v1
 kind: Template
 metadata:
@@ -165,6 +179,39 @@ objects:
       spec:
         serviceAccountName: assisted-service
         containers:
+          - name: envoy-sidecar
+            image: ${ENVOY_IMAGE}:${ENVOY_TAG}
+            imagePullPolicy: IfNotPresent
+            command:
+            - envoy
+            - --config-path
+            - /configs/envoy/main.yaml
+            ports:
+            - name: api-envoy
+              protocol: TCP
+              containerPort: 8900
+            - name: metrics-envoy
+              protocol: TCP
+              containerPort: 9000
+            volumeMounts:
+            - name: envoy-config
+              mountPath: /configs/envoy
+            livenessProbe:
+              httpGet:
+                path: /health
+                port: 8900
+              initialDelaySeconds: ${{LIVENESS_PROBE_INITIAL_DELAY_SECONDS}}
+            readinessProbe:
+              httpGet:
+                path: /ready
+                port: 8900
+              initialDelaySeconds: ${{READINESS_PROBE_INITIAL_DELAY_SECONDS}}
+            resources:
+              requests:
+                memory: ${ENVOY_MEMORY_REQUEST}
+                cpu: ${ENVOY_CPU_REQUEST}
+              limits:
+                memory: ${ENVOY_MEMORY_LIMIT}
           - name: assisted-service
             image: ${ASSISTED_SERVICE_IMAGE}:${IMAGE_TAG}
             imagePullPolicy: Always
@@ -348,6 +395,9 @@ objects:
             secret:
               secretName: route53-creds
               optional: true
+          - name: envoy-config
+            configMap:
+              name: assisted-service-envoy-config
 - apiVersion: v1
   kind: Service
   metadata:
@@ -359,7 +409,7 @@ objects:
       - name: assisted-svc
         port: 8090
         protocol: TCP
-        targetPort: 8090
+        targetPort: ${{SVC_TARGET_PORT}}
     selector:
       app: assisted-service
 - apiVersion: policy/v1beta1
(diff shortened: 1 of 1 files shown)


openshift/template.yaml
```diff
@@ -144,6 +144,25 @@ parameters:
...
   value: "30"
   required: false
+- name: ENVOY_IMAGE
+  value: "docker.io/envoyproxy/envoy"
+  required: false
+- name: ENVOY_TAG
+  value: "v1.24.0"
+  required: false
+- name: ENVOY_IMAGE_PULL_POLICY
+  value: IfNotPresent
```
jhernand, line 154: I'd say this doesn't need to be configurable.

openshift/template.yaml
```diff
@@ -144,6 +144,25 @@ parameters:
...
+- name: ENVOY_MEMORY_REQUEST
+  value: 256Mi
+- name: ENVOY_CPU_REQUEST
+  value: 10m
+- name: ENVOY_MEMORY_LIMIT
+  value: 512Mi
+- name: ENVOY_PORT
+  value: 8900
+- name: SVC_TARGET_PORT
+  value: 8900
```
jhernand: I don't see the need to make the ports configurable. If you decide to do so then use the variables instead of the constants everywhere below.

openshift/template.yaml
```diff
@@ -373,3 +428,70 @@ objects:
...
+                    route: { cluster: assisted_service }
+              http_filters:
+              - name: envoy.filters.http.router
+                typed_config:
+                  "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
+      clusters:
+      - name: assisted_service
+        connect_timeout: 1s
+        type: STATIC
+        lb_policy: ROUND_ROBIN
```
jhernand, line 488: You will probably want to use HTTP/2 here:

```
http2_protocol_options: {}
```

It needs to be enabled explicitly, otherwise all the communication with the backend will use HTTP/1.1, which is less efficient and requires more connections.

openshift/template.yaml
```diff
@@ -373,3 +428,70 @@ objects:
...
+          - name: envoy.filters.network.http_connection_manager
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
+              access_log:
+              - name: envoy.access_loggers.file
+                typed_config:
+                  "@type": type.googleapis.com/envoy.extensions.access_loggers.file.v3.FileAccessLog
+                  path: /dev/stdout
+                  jsonFormat:
+                    authority: "%REQ(:AUTHORITY)%"
```
jhernand, line 459: I miss the date/time here: `"whatever": %START_TIME%`

openshift/template.yaml
```diff
@@ -373,3 +428,70 @@ objects:
...
+                    duration: "%DURATION%"
+                    path: "%REQ(X-ENVOY-ORIGINAL-PATH?:PATH)%"
+                    protocol: "%PROTOCOL%"
+                    response_code: "%RESPONSE_CODE%"
+                    response_flags: "%RESPONSE_FLAGS%"
+                    route_name: "%ROUTE_NAME%"
+                    upstream_host: "%UPSTREAM_HOST%"
```
jhernand, line 466: `route_name` and `upstream_host` aren't really interesting here, the route name will always be the same as there is only one route (and it hasn't a name) and the upstream host will also always be `127.0.0.1`.

openshift/template.yaml
```diff
@@ -373,3 +428,70 @@ objects:
...
+                    user_agent: "%REQ(USER-AGENT)%"
+                    x_agent_name: "%REQ(X-AGENTNAME)%"
```
jhernand, line 468: What is the `x-agentname` header?

jhernand (commented):
Looks great @rccrdpccl. I have some minor comments that you want may to consider before merging.

For the future, not in this pull request, I'd also suggest to add to the backend support for Unix sockets and HTTP/2 without TLS. That reduces the resources needed for the communication between Envoy and the backend.

openshift/template.yaml
rccrdpccl: :+1: makes sense, however I'd like to keep the svc target configurable, in case something goes wrong when deploying envoy we can easily switch to the original port and bypass envoy by overriding this parameter (no need to backport changes in the git repo). WDYT?

openshift/template.yaml
rccrdpccl: Oops, leftover from the example I took for the json formatting :unamused:

openshift/template.yaml
jhernand: That is a good reason, yes.

romfreiman:
Finally!!!!!!
What about prod? :)

openshift/template.yaml
rccrdpccl: It looks like the assisted-service is not handling http2 without TLS (golang http library).
However I will try to configure it so that envoy can handle http2 (and terminate it in this case), so it will be easier to switch full http2 from the app.

jhernand: Yes, that is what I mentioned in the overall review comment: currently we don't support Unix sockets or HTTP/2 without TLS; would be nice to have in the future.

jhernand: In the listener side you don't need to configure HTTP/2 explicitly: the default connection manager supports both HTTP/1.1 and HTTP/2 out of the box. But the client (the OpenShift router in this case) will not even try connect using HTTP/2 because it doesn't support _h2c_ (HTTP/2 without TLS) and our service doesn't use TLS.

osherdp:
> /cc @osherdp WDYT about all other environments? CI/local etc, should they also have envoy as sidecar?

Good question. I don't know if there's any need in understanding the source of requests so the main advantage of doing so is uniformity with behavior in saas envs

All in all, I think we don't need it at the moment in any of those envs

openshift/template.yaml
```diff
@@ -373,3 +428,70 @@ objects:
...
+        - filters:
+          - name: envoy.filters.network.http_connection_manager
+            typed_config:
+              "@type": type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager
+              access_log:
+              - name: envoy.access_loggers.file
+                typed_config:
+                  "@type": type.googleapis.com/envoy.extensions.access_loggers.file.v3.FileAccessLog
+                  path: /dev/stdout
+                  jsonFormat:
```
osherdp, line 458: I took a look in integration env, and I'd like to check something
Is there a way to order fields in this JSON by alphabetical order? It can come in handy when reading the output directly, even if we plan to let other tools read this stuff

osherdp (changes requested):

openshift/template.yaml
```diff
@@ -144,6 +144,25 @@ parameters:
 - name: LIVENESS_PROBE_INITIAL_DELAY_SECONDS
   value: "30"
   required: false
+- name: ENVOY_IMAGE
+  value: "docker.io/envoyproxy/envoy"
```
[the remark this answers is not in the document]
rccrdpccl: Good catch, I've used this for testing but was planning to switch to app-sre mirror

openshift/template.yaml
rccrdpccl: I am not aware of any option to force envoy to output ordered logs. I was also confused by the seemingly random order for log fields

osherdp: then no worries about it. it's just a micro-optimization

openshift/template.yaml
rccrdpccl: Right, so I take it that once we change the service to handle http2, it will work with http2 out of the box (once added the above mentioned settings), no need to force upgrade right?
If I add such settings right now it would break with "protocol error".

openshift/template.yaml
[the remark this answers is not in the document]
jhernand: The APP-SRE team is already mirroring the Envoy repository from _docker.io_ to _quay.io_. All you need is include something like this in the app-interface configuration:

```yaml
ENVOY_IMAGE: quay.io/app-sre/envoyproxy:v1.14.0
```

openshift/template.yaml
jhernand: Yes, correct.

openshift/template.yaml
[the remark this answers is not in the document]
rccrdpccl: Yes that's what I meant, I am now trying to configure this deployment to be able to pull from that repository

osherdp (approved):

PR #4685, opened 2022-11-28
MGMT-12697: Reject requests with unknown fields

jhernand:
Currently the API accepts requests with unknown fields and silently ignores them. This patch changes it so that it rejects those requests. For example a request like this:

```
POST /api/assisted-install/v2/clusters/a67d9ab8-ef3f-4176-ac69-337bb26ae135 <<.
{
  "base_dns_doman": "home.arpa"
}
```

Will result in the following response:

```
HTTP/1.1 400 Bad Request
{
  "code": 400,
  "message": "parsing clusterUpdateParams body from \"\" failed, because json: unknown field \"base_dns_doman\""
}
```

The error message isn't very nice, but similar to error messages generated by other wrong JSON documents. Should probably be improved in the future.

Note that in order to do this it is necessary to change the code that creates the JSON decoder used to parse request bodies, and currently that requires changing the templates used by the code generators. The templates in the new `templates` directory have been copied from here and then modified:

https://github.com/go-swagger/go-swagger/tree/v0.28.0/generator/templates/contrib/stratoscale/server

Related: https://issues.redhat.com/browse/MGMT-12697

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


carbonin:
I believe this team (or a good bit of this team while at a different company) contributed that template.
I think you should be able to submit changes to it if we'd rather not maintain it here (although I'm not sure why we wouldn't).

cc @filanov

filanov:
Is that a temporary solution until your changes are getting merged into go-swagger?

jhernand:
> Is that a temporary solution until your changes are getting merged into go-swagger?

Yes, this is an alternative to waiting till the change in _go-swagger_ is merged (it is already, see [here](https://github.com/go-swagger/go-swagger/pull/2863)), a new version containing it is released and we update our code to use it (it may require other unrelated changes).

I'd say that this feature isn't important enough to not wait for the new _go-swagger_ release. But if we don't want to wait we can merge this and revert it once the new version has been released and we have updated our code to use it.

filanov:
> > Is that a temporary solution until your changes are getting merged into go-swagger?
> 
> Yes, this is an alternative to waiting till the change in _go-swagger_ is merged (it is already, see [here](https://github.com/go-swagger/go-swagger/pull/2863)), a new version containing it is released and we update our code to use it (it may require other unrelated changes).
> 
> I'd say that this feature isn't important enough to not wait for the new _go-swagger_ release. But if we don't want to wait we can merge this and revert it once the new version has been released and we have updated our code to use it.

hoped to avoid having in between solution and wait for go-swagger to be merged. as you said it's not urgent so we can wait for the release

jhernand:
> hoped to avoid having in between solution and wait for go-swagger to be merged. as you said it's not urgent so we can wait for the release

OK, I am closing this, we can always reopen if we change our minds or if the release of _go-swagger_ takes too long.

PR #4694, opened 2022-11-30
MGMT-12388: Forbid dual-stack VIPs for OpenShift pre-4.12

mkowalski:
The feature is available upstream starting only from 4.12; given that Assisted's API is version-agnostic, we are adding a validator for cluster creation as well as cluster update logic that will make sure we raise an error whenever an user tries to pass multiple VIPs to a cluster that uses OpenShift older than 4.12

Closes: [MGMT-12388](https://issues.redhat.com//browse/MGMT-12388)
Implements: #4245

/cc @nmagnezi


internal/cluster/validations/validations.go
```diff
@@ -398,6 +399,12 @@ func ValidateClusterCreateIPAddresses(ipV6Supported bool, clusterId strfmt.UUID,
 		return common.NewApiError(http.StatusBadRequest, err)
 	}
 
+	if (len(params.APIVips) > 1 || len(params.IngressVips) > 1) && !featuresupport.IsFeatureSupported(*params.OpenshiftVersion,
+		models.FeatureSupportLevelFeaturesItems0FeatureIDDUALSTACKVIPS) {
+
+		return common.NewApiError(http.StatusBadRequest, errors.Errorf("%s %s", "dual-stack VIPs are not supported in OpenShift", *params.OpenshiftVersion))
```
nmagnezi, line 405: A suggestion, should this response also include the OpenShift version from which dual stack is supported?

mkowalski: It's not trivial because I want to leverage `featuresupport.IsFeatureSupported()` function without hardcoding anything. I could write manually "use 4.12 or later" but then I defeat the purpose of using the `featuresupport` package.

I'm happy to do what you suggest, but I'd need some direction how to do it without any hardcoding here

nmagnezi (approved):
/lgtm

PR #4703, opened 2022-12-01
MGMT-12743: adds enhancement proposal for shorter image URLs

mhrivnak:

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/short-url-enhancement.md
+++ b/docs/enhancements/short-url-enhancement.md
@@ -3,7 +3,7 @@ title: short-image-urls
 authors:
   - "@mhrivnak"
 creation-date: 2022-11-30
-last-updated: 2022-11-30
+last-updated: 2022-12-22
 ---
 
 # Short image URLs
@@ -81,6 +81,32 @@ That should change to:
 ```
 http://api.openshift.com/api/assisted-images/images/881dd303-940e-4c1a-a844-dc3c64da0e95/4.11/x86_64/full.iso
 ```
+
+#### Token Parsing
+
+Unfortunately the JWT payload uses different keys to identify the InfraEnv
+depending on whether the service is deployed as part of OCM
+(console.redhat.com) or locally using "local auth".
+
+Tokens served by OCM have a JWT payload with two fields:
+
+```
+{
+  "exp": 1671661677,                             # expiration in seconds since epoch
+  "sub": "e1bfd718-7ce4-4338-a123-07e68214734a"  # infraenv ID
+}
+```
+
+Tokens served when using local auth have a JWT payload with one self-evident field:
+
+```
+{
+  "infra_env_id": "5b6e5a9e-9fe2-45a0-a95f-976ee62169aa"
+}
+```
+
+The image service will need to parse the payload and look for the value under
+both keys.
 
 #### Compatibility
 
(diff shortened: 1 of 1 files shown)


docs/enhancements/short-url-enhancement.md
```diff
@@ -0,0 +1,118 @@
...
+Today's unsigned URLs look like this:
+
+```
+http://api.openshift.com/api/assisted-images/images/881dd303-940e-4c1a-a844-dc3c64da0e95?arch=x86_64&type=full-iso&version=4.11
+```
+
+That should change to:
+
+```
+http://api.openshift.com/api/assisted-images/images/881dd303-940e-4c1a-a844-dc3c64da0e95/4.11/x86_64/full.iso
```
omertuc, line 82: Shouldn't this be:
```suggestion
http://api.openshift.com/api/assisted-images/bytoken/images/<TOKEN>/4.11/x86_64/full.iso
```

carbonin: In this case I'm not sure `bytoken` makes sense when we have no token since this is unsigned.

omertuc: sorry ignore my comment it doesn't make sense

carbonin:
I'm okay with this.

@mhrivnak did you want to add something about the ACM vs SaaS tokens or leave that to sort out during implementation?

docs/enhancements/short-url-enhancement.md
```diff
@@ -0,0 +1,118 @@
...
+
+That should change to:
+
+```
+http://api.openshift.com/api/assisted-images/byid/881dd303-940e-4c1a-a844-dc3c64da0e95/4.11/x86_64/full.iso
+```
+
+#### Compatibility
+
+The previous URL layouts should continue to work. But the default URLs returned
```
nmagnezi, line 87: Do we keep the older format for backward compatibility? 
If yes, can we somehow mark (or just document) that the old format is deprecated?

nmagnezi (commented):
Looks good.
I have one question that I posted as a comment.

docs/enhancements/short-url-enhancement.md
carbonin: I don't think we can mark it as deprecated, but at the same time no one should be relying on the format.

The URL is always given to the user in full so they shouldn't care about how it's formatted. I see it as the same as when we moved from s3 to the image service, the URL will be different, but it should be transparent to the user.

mhrivnak: I think we need compatibility just long enough to make sure that any URL handed out recently continue to work after the service gets upgraded.

mhrivnak:
> @mhrivnak did you want to add something about the ACM vs SaaS tokens or leave that to sort out during implementation?

I'd like to nail down how ACM handles tokens before accepting this proposal.

carbonin (approved):

PR #4708, opened 2022-12-01
MetalLB Enhancement

flaper87:
Propose an enhancement to add a metallb opreator to Assisted Installer

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


docs/enhancements/metallb-operator.md
```diff
@@ -0,0 +1,205 @@
...
+to act as an appliance.
+
+Some of this tasks could be made part of components that are closer to the core so that they can be
+leveraged in other environments like SaaS. This enhancement proposes moving one of these tasks to
+Assisted Installer. Specifically, this enhancement proposes to deploy the MetalLB operator using
+Assisted Installer so that it can create clusters that can be relocated.
+
+### Goals
+
+- To add a MetalLB operator to Assiste Installer
```
flaper87, line 33: I have a draft PR here (not complete since it's waiting on the discussion in this enhancement): 

https://github.com/openshift/assisted-service/pull/4792

flaper87 (commented):
Would also love to hear feedback from /cc @jelkosz @eranco74 @tsorya

docs/enhancements/metallb-operator.md
```diff
@@ -0,0 +1,205 @@
...
```
tsorya, line 33: nit : Assiste -> Assisted

docs/enhancements/metallb-operator.md
```diff
@@ -0,0 +1,205 @@
...
+external network from the External API VIP and the External Ingress IP provided by the user. There
+are utilities in Assisted Installer to do this already.
+
+Some requirements:
+
+- The **external** API VIP and Ingress IP must be on the same subnet.
+- The host must be connected to this network and the network must be routable
+- These 2 IPs must be optional. The IPAddressPool and Service won't be configured if this information is not provided
+- Either both are IPs are provided or none.
+- The *internal* network must be configured as a static network to prevent the MachineNetwork to
```
tsorya, line 133: Why it should be static? I mean there is a requirement that ip of the nodes can't change as etcd will break so why it is required to be "static" here?

flaper87: This is because we don't want this IP to change during the relocation. As you said, everything will break if this IP changes. We can prevent this by having this internal IP as static and isolated. That is, the internal network is not expected to be connected to anything.

This is how we achieved relocation for ZTPFW. The big difference is that, in ZTPFW, there's a specific IP hardcoded. In this enhancement, the idea is to let the user's pick the IP they want.

Hope it makes sense

docs/enhancements/metallb-operator.md
```diff
@@ -0,0 +1,205 @@
...
+different location and reconnected.
+
+#### Story 2
+
+As a datacenter administrator, I want to be able to install an OCP cluster whose IP can be
+reconfigured as well as its domain.
+
+### Implementation Details/Notes/Constraints [optional]
+
+1. It's currently not possible to enable operators when using KubeAPI. Implementing this feature,
```
jelkosz, line 150: ACM has its own ways of interacting with clusters which have been installed from it, including assisted clusters (e.g. using policies). Do you think you could utilize this to achieve the same?

docs/enhancements/metallb-operator.md
```diff
@@ -0,0 +1,205 @@
...
+to be familiar with the 2 networks that are needed.
+
+## Design Details [optional]
+
+### Open Questions
+
+### UI Impact
+
+It will be required to adapt the UI to interact with this feature. No breaking changes will be
+introduced by this enhancement.
```
jelkosz, line 185: The api will be similar to the one for adding CNV or ODF, right?
@jkilzi FYI

tsorya: Should be at least, if we already have all the relevant parts why to implement something new

docs/enhancements/metallb-operator.md
tsorya: The question is if we are going to block installation if user don't set static network or we will allow to spawn metalb when ever user wants.
I think we should just allow spawning metalb operator without checking if static network was configured or not

tsorya: In that case we can decouple "internal network" configuration from metalb and say that we want to allow installing metalb from assisted on user demand. This will allow us to minimize validations that will be required.

docs/enhancements/metallb-operator.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+### Non-Goals
+
+- Change the way network configuration is managed today
+
+
+## Premises
+
+An Edgecluster must be connected to 2 networks. We will refer to these networks as Insternal network
```
romfreiman, line 45: During the installation phase?

docs/enhancements/metallb-operator.md
```diff
@@ -0,0 +1,205 @@
...
+
+People using ZTPFW have solved this issue by providing an isolated, routable, network as the
+MachineNetwork. This is done either by using a top of the rack switch or some other mechanism. This
+is particularly relevant for multinode deployments where hosts need to be able to talk to each
+other.
+
+Finally, the UX won't be as straightforward when building a relocatable cluster as users will have
+to be familiar with the 2 networks that are needed.
+
+## Design Details [optional]
```
romfreiman, line 178: i dont see any discussions about pre flight validations - what will change? new validations that will be added? 
I think this is the core of this feature so lets think it through

docs/enhancements/metallb-operator.md
```diff
@@ -0,0 +1,205 @@
...
+
+### Open Questions
+
+### UI Impact
```
romfreiman, line 182: also to d/s the unboxing UI? and include it here?

docs/enhancements/metallb-operator.md
avishayt, line 45:
```suggestion
An Edgecluster must be connected to 2 networks. We will refer to these networks as Internal network
```

docs/enhancements/metallb-operator.md
```diff
@@ -0,0 +1,205 @@
...
+### Implementation Details/Notes/Constraints [optional]
+
+1. It's currently not possible to enable operators when using KubeAPI. Implementing this feature,
+without enabling it through KubeAPI will limit the scope of it for environments that use ZTP. It
+should be considered to allow enabling operators through KubeAPI either as part of this enhancement
+or a separate one.
+2. Assisted Installer doesn't have a way to configure an operator. Operators can be enabled or
+disabled. The rest of the information they take from the infrastructure data that has been
+collected. For this enhancement, we need to either introduce the concept of External API VIP and
+External Ingress IP to the cluster or allow for specifying operator-specific configurations.
```
avishayt, line 157: The API has a way to define and pass operator-specific parameters, but none of the currently-supported operators use this.  I'm wondering if we should expose this feature as "install metalLB" or "make this cluster relocatable".  I'm leaning toward the second option.

docs/enhancements/metallb-operator.md
avishayt: I'm not so sure.  This enhancement isn't really about installing MetalLB with whatever configuration the user wants, but about making the cluster relocatable.  This has a very different UX - we can expect different parameters and validate the configuration more accurately.

docs/enhancements/metallb-operator.md
flaper87: > I think we should just allow spawning metalb operator without checking if static network was configured or not

mmh, not sure I agree with this. The reason is that the whole point of this feature is to build clusters that can be relocated. If the internal network is not static, the cluster will break if/once it is relocated. We would allow the user to build a cluster that will break on the first move.

We could argue that installing MetalLB should not mean "relocatable" but the configuration we are applying is based on that assumption

docs/enhancements/metallb-operator.md
flaper87: > I'm wondering if we should expose this feature as "install metalLB" or "make this cluster relocatable". I'm leaning toward the second option.

Agreed! Big part of the reason is that the configurations and validations we would expose work under the assumption that we are building a relocatable cluster. Otherwise, we would not need any deployment-time configurations but rather just install MetalLB

docs/enhancements/metallb-operator.md
flaper87: Agreed! I didn't detail this part but I will incorporate it once we discuss some of the initial details of the enhancement (like requiring static IP  for internal network vs DHCP).

docs/enhancements/metallb-operator.md
flaper87: mmmh, I like this idea. If we are building a relocatable cluster, we might as well ship the UI that allows to reconfigure things. But, at this point, I wonder if it should/could be an extension on the actual OpenShift console rather than a completely independent container/ui. IIRC, we have a way to do this @jelkosz ?

docs/enhancements/metallb-operator.md
tsorya: I think we should not argue, assisted should not support something specific for "relocatable", it should support "generic" option and if you want to make "relocatable" just apply relevant configuration to make it such. 
I mean that one day it will be in SAAS and people will want it and they are not going to relocate their cluster, they just want to use this functionality, why not to give to them. 
IMO if user needs static network he should configure it , if he doesn't why to block him.

tsorya: In case we are talking only about "relocatable" (not sure i understand why we need to implement this very specific scenario and why external tool is not enough for this as it can apply relevant configuration) we should change name of the enhancement and rephrase it.
Though we were already talked about MetalLb for agent capi and it is already used there (not sure how we apply it). I think will nicer to support all the options and make it generic as we will move there for sure one day

PR #4720, opened 2022-12-05
NO-ISSUE: extract client from reconcile context

danielerez:
In HypershiftAgentServiceConfigReconciler, we currently need to switch between clients (hub and spoke), since a single client property exists in AgentServiceConfigReconcileContext.

Hence, to avoid sharing the property with both clients in order to have a clear notion of the client in use; moved it to ASC struct for explicitly specifying the required client.
I.e. instead of a single ASC with alternating clients, created a separate one for the spoke using the generated client.

Also, moved the 'properties' from ASC to AgentServiceConfigReconcileContext as the map is used for sharing values between clients.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -107,7 +107,6 @@ var (
 )
 
 type AgentServiceConfigReconcileContext struct {
-	client.Client
 	Log    logrus.FieldLogger
 	Scheme *runtime.Scheme
 
@@ -120,6 +119,8 @@ type AgentServiceConfigReconcileContext struct {
 
 // AgentServiceConfigReconciler reconciles a AgentServiceConfig object
 type AgentServiceConfigReconciler struct {
+	client.Client
+
 	AgentServiceConfigReconcileContext
 
 	// Namespace the operator is running in
@@ -139,6 +140,11 @@ type ASC struct {
 	/* reference to the reconciler state */
 	rec *AgentServiceConfigReconcileContext
 
+	/* The client for cluster API, which is used for either hub or spoke cluster client.
+	   I.e. it should be set according to the required cluster communication (hub/spoke).
+	*/
+	Client client.Client
+
 	/* the instance itself */
 	Object client.Object
 
@@ -152,12 +158,15 @@ type ASC struct {
 	properties map[string]interface{}
 }
 
-func (asc *ASC) init(r *AgentServiceConfigReconciler, instance *aiv1beta1.AgentServiceConfig) {
+func initASC(r *AgentServiceConfigReconciler, instance *aiv1beta1.AgentServiceConfig) ASC {
+	var asc ASC
 	asc.namespace = r.Namespace
 	asc.rec = &r.AgentServiceConfigReconcileContext
+	asc.Client = r.Client
 	asc.Object = instance
 	asc.spec = &instance.Spec
 	asc.conditions = &instance.Status.Conditions
+	return asc
 }
 
 type NewComponentFn func(context.Context, logrus.FieldLogger, ASC) (client.Object, controllerutil.MutateFn, error)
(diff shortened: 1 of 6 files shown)


internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -164,7 +159,7 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
 	}
 
 	// Initialize ASC to reconcile components according to instance context
-	asc.initHASC(hr, instance)
+	asc.initHASC(hr, instance, hr.Client)
```
carbonin, line 162: I know this isn't really related to this change directly, but why not have this function return a new `ASC` instance?

Initializing it here (rather than creating it) makes me feel like it is shared when the whole premise of this PR is that is it not.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -107,19 +107,23 @@ var (
...
-	client.Client
 	Log    logrus.FieldLogger
 	Scheme *runtime.Scheme
 
 	// selector and tolerations the Operator runs in and propagates to its deployments
 	NodeSelector map[string]string
 	Tolerations  []corev1.Toleration
 
+	/* properties. use this field for cross cluster communication */
+	Properties map[string]interface{}
```
carbonin, line 118: I'm not really sure what this is, but I don't think maps can be safely used concurrently without locking.
Maybe make this a separate PR with more clear justification?

slaviered: why does the map moved from HASC to the global scope?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -149,15 +155,18 @@ type ASC struct {
 	conditions *[]conditionsv1.Condition
 
 	/* properties. use this field for cross cluster communication */
-	properties map[string]interface{}
+	properties *map[string]interface{}
```
carbonin, line 158: The zero-value of a map is `nil` and they are generally treated as references. Why do you need a pointer here?

Here's an example if it helps: https://go.dev/play/p/l5YfudQhWT1

danielerez: Nice, wasn't sure about it. Fixed.

internal/controller/controllers/hypershiftagentserviceconfig_controller_test.go
```diff
@@ -570,31 +570,31 @@ var _ = Describe("HypershiftAgentServiceConfig reconcile", func() {
 		}
 
 		It("successfully for leader election role", func() {
-			asc.initHASC(hr, hsc)
+			asc = initHASC(hr, hsc, hr.Client, ascProperties)
```
carbonin, line 573: Shouldn't the `asc` instance be scoped to the example (i.e. remove the variable declaration and use `:=` here)? I'm imagining a test not calling `initHASC` and getting the value from the previous test run.

internal/controller/controllers/hypershiftagentserviceconfig_controller_test.go
```diff
@@ -259,6 +259,7 @@ var _ = Describe("HypershiftAgentServiceConfig reconcile", func() {
 		mockSpokeClient = spoke_k8s_client.NewMockSpokeK8sClient(mockCtrl)
 		mockSpokeClientCache = NewMockSpokeClientCache(mockCtrl)
 
+		ascProperties = make(map[string]interface{})
```
carbonin, line 262: If this is only relevant in an inner context then it should be defined and set in that context.

Moving it further away from where it's used makes the tests harder to understand.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -163,8 +161,11 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	// Create a map for properties that are shared between ASC objects
+	ascProperties := make(map[string]interface{})
```
carbonin, line 165: Is moving this out solving a bug of some sort? This really feels like it deserves its own PR and an issue, or at least a separate commit.

carbonin (approved):

PR #4755, opened 2022-12-08
MGMT-12870: Add missing dual-stack VIP CRUD for KubeAPI

mkowalski:
This PR implements a missing CRUD operations to handle dual-stack VIPs in case when KubeAPI is used.

Closes: [MGMT-12870](https://issues.redhat.com//browse/MGMT-12870)
Implements: #4245

/cc @nmagnezi

### DIFF
--- a/internal/controller/controllers/common.go
+++ b/internal/controller/controllers/common.go
@@ -315,6 +315,30 @@ func machineNetworksEntriesToArray(entries []hiveext.MachineNetworkEntry) []*mod
 	}).([]*models.MachineNetwork)
 }
 
+func apiVipsArrayToStrings(vips []*models.APIVip) []string {
+	return funk.Map(vips, func(vip *models.APIVip) string {
+		return string(vip.IP)
+	}).([]string)
+}
+
+func apiVipsEntriesToArray(entries []string) []*models.APIVip {
+	return funk.Map(entries, func(entry string) *models.APIVip {
+		return &models.APIVip{IP: models.IP(entry)}
+	}).([]*models.APIVip)
+}
+
+func ingressVipsArrayToStrings(vips []*models.IngressVip) []string {
+	return funk.Map(vips, func(vip *models.IngressVip) string {
+		return string(vip.IP)
+	}).([]string)
+}
+
+func ingressVipsEntriesToArray(entries []string) []*models.IngressVip {
+	return funk.Map(entries, func(entry string) *models.IngressVip {
+		return &models.IngressVip{IP: models.IP(entry)}
+	}).([]*models.IngressVip)
+}
+
 func signURL(urlString string, authType auth.AuthType, id string, keyType gencrypto.LocalJWTKeyType) (string, error) {
 	if authType != auth.TypeLocal {
 		return urlString, nil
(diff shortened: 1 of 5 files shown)


internal/controller/controllers/common.go
```diff
@@ -315,6 +315,30 @@ func machineNetworksEntriesToArray(entries []hiveext.MachineNetworkEntry) []*mod
 	}).([]*models.MachineNetwork)
 }
 
+func apiVipsArrayToStrings(vips []*models.APIVip) []string {
+	return funk.Map(vips, func(vip *models.APIVip) string {
```
omertuc, line 319: We have go 1.18 now we don't need funk anymore, we have https://github.com/samber/lo 

But you may keep using `funk` for consistency with the rest of this file if you want

omertuc:
Could you give some short background about:

- What is "autocalculation" - calculation of what exactly, based on what exactly
- What does VIP have to do with SNO - SNO doesn't have a concept of VIPs, so what's up with all this VIP talk? What does it mean?

internal/controller/controllers/common.go
mkowalski: Thanks for the suggestion! I had a look and it seems like instead of `funk.Map` I would be doing

```
import "github.com/samber/lo"

lo.Map[int64, string]([]int64{1, 2, 3, 4}, func(x int64, index int) string {
    return strconv.FormatInt(x, 10)
})
// []string{"1", "2", "3", "4"}
```

so I'd still need to import additional library and syntax-wise it's not much more clear than `funk.Map`. If there is no other push towards moving out of `funk`, I'd prefer to stay with it

omertuc: Actually I don't know why their README example is so ugly, if you use it properly it's simpler and gets rid of the cast at the end:

```go
func apiVipsArrayToStrings(vips []*models.APIVip) []string {
	return lo.Map(vips, func(vip *models.APIVip, _ int) string {
		return string(vip.IP)
	})
}
```

Not that performance matters here, but it's also much faster since it doesn't use ugly runtime `Interfaces{}` for everything

mkowalski:
/hold

After discussing with Omer offline we have realized there is no any real need for users of SNO to define API VIPs explicitly in order to consume this feature. Therefore I will remove the part of the code that allows doing so.

internal/cluster/validations/validations.go
```diff
@@ -442,6 +442,7 @@ func ValidateClusterCreateIPAddresses(ipV6Supported bool, clusterId strfmt.UUID,
 	targetConfiguration.IngressVips = params.IngressVips
 	targetConfiguration.UserManagedNetworking = params.UserManagedNetworking
 	targetConfiguration.VipDhcpAllocation = params.VipDhcpAllocation
+	targetConfiguration.HighAvailabilityMode = params.HighAvailabilityMode
```
mkowalski, line 445: todo myself: extract this part to separate commit

subsystem/kubeapi_test.go
```diff
@@ -3917,12 +3917,24 @@ var _ = Describe("[kube-api]cluster installation", func() {
 			return aci.Status.APIVIP
 		}, "30s", "1s").Should(Equal(aciSpec.APIVIP))
 
+		By("Ensure APIVIPs exist in status")
+		Eventually(func() int {
+			aci := getAgentClusterInstallCRD(ctx, kubeClient, installkey)
+			return len(aci.Status.APIVIPs)
+		}, "30s", "1s").ShouldNot(Equal(0))
```
mkowalski, line 3924: todo myself: add a test case to check apivips[0] for a value of the IP address

nmagnezi (approved):
/lgtm

PR #4767, opened 2022-12-12
OCPBUGS-2953: Fix regex to filter images by tag and by sha

mkowalski:
This PR modifies the regex expression that we use to extract the name of the image. The previous version worked only with images passed via tag, but did not support images passed by SHA. With this modification we now support both naming schemas.

/cc @omertuc

### DIFF
--- a/internal/ignition/ignition.go
+++ b/internal/ignition/ignition.go
@@ -136,7 +136,7 @@ const agentFixBZ1964591 = `#!/usr/bin/sh
 # In such a scenario agent.service will detect the image is not present and pull it again. In case
 # the image is present and can be detected correctly, no any action is required.
 
-IMAGE=$(echo $1 | sed 's/:.*//')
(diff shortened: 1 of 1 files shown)


mkowalski:
/test e2e-agent-compact

`{  failed to acquire lease for "equinix-ocp-metal-quota-slice": status 503 Service Unavailable, status code 503}`

filanov (approved):

omertuc:
If [bugzilla 1964591](https://bugzilla.redhat.com/show_bug.cgi?id=1964591) was resolved, shouldn't we just remove this script?

omertuc:
Unless this "workaround" is what resolved it, and it's not really a workaround?

mkowalski:
https://bugzilla.redhat.com/show_bug.cgi?id=1993243 is what was hitting us

PR #4768, opened 2022-12-12
MGMT-12886: fetch CRDs by group in HASC ctrl

danielerez:
In HypershiftAgentServiceConfigReconciler, we fetch the agent-install CRDs by filtering using the label added by OLM ('operators.coreos.com/assisted-service-operator.[ns]').

However, MCE are [applying](https://github.com/stolostron/backplane-operator/blob/main/main.go#L148-L164) the [CRDs](https://github.com/stolostron/backplane-operator/tree/main/pkg/templates/crds/assisted-service) (scraped from [community-operators-prod](https://github.com/redhat-openshift-ecosystem/community-operators-prod/tree/main/operators/assisted-service-operator/0.4.50)) manually without using OLM.
Hence, no labels exist in the CRDs. So instead of relying on the label, search for relevent CRDs according to 'agent-install.openshift.io' group.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/hypershiftagentserviceconfig_controller.go
+++ b/internal/controller/controllers/hypershiftagentserviceconfig_controller.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	routev1 "github.com/openshift/api/route/v1"
+	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
 	aiv1beta1 "github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/config"
 	"github.com/openshift/assisted-service/internal/spoke_k8s_client"
@@ -69,9 +70,6 @@ type HypershiftAgentServiceConfigReconciler struct {
 
 	// A cache for the spoke clients
 	SpokeClients SpokeClientCache
-
-	// Namespace the operator is running in
-	Namespace string
 }
 
 func initHASC(r *HypershiftAgentServiceConfigReconciler, instance *aiv1beta1.HypershiftAgentServiceConfig,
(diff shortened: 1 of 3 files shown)


internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -463,12 +460,17 @@ func (hr *HypershiftAgentServiceConfigReconciler) cleanStaleSpokeAgentInstallCRD
...
-	}
-	if err := c.List(ctx, crds, listOpts...); err != nil {
+	if err := c.List(ctx, crds); err != nil {
 		return nil, pkgerror.Wrap(err, "Failed to list CRDs")
 	}
+
+	// Filter CRDs by 'agent-install.openshift.io' and 'extensions.hive.openshift.io' groups
+	// (to include our agentclusterinstalls hive extension)
+	agentInstallCRDs := funk.Filter(crds.Items, func(crd apiextensionsv1.CustomResourceDefinition) bool {
+		return crd.Spec.Group == "agent-install.openshift.io" || crd.Spec.Group == "extensions.hive.openshift.io"
```
filanov, line 470: i think that we already have a const for that https://github.com/openshift/assisted-service/blob/3d6d97472d74c37a14fe0a9d0a4cc538dd28f004/api/v1beta1/groupversion_info.go#L28
and 
https://github.com/openshift/assisted-service/blob/3d6d97472d74c37a14fe0a9d0a4cc538dd28f004/api/hiveextension/v1beta1/groupversion_info.go#L28

slaviered: What guarantees that this logic will be always aligned with OLM's? 
I would keep the old code, and only if there are no labels present, take the other route.

In addition, take the implementation to a well documented function, so future maintainers will be able to quickly identify any issues of OLM-based installations start to behave differently than OCM's

danielerez: We are defining the group when creating the CRDs (e.g. [here](https://github.com/openshift/assisted-service/blob/3d6d97472d74c37a14fe0a9d0a4cc538dd28f004/config/crd/resources.yaml#L9)), so no need to fetch CRDs twice, it would work regardless of OLM.
I can indeed extract the filter logic to another func if it makes the code clearer.

PR #4770, opened 2022-12-12
MGMT-6536: select default cidr in SNO based on default route metrics

slaviered:
## Desciprtion
To enable selecting a recommended CIDR when there are multiple options, we will use the following method:
    
    1. Select the default route(s). Use the Metric field to choose the route with the least cost
    2. Check if the gateway IP is part of any known machine CIDR
    3. Choose this CIDR as the machine network
    
    The network for the default route will also be selected by the OVN after
    rebooting of the bootstrap node regardless of the selection of the user,
    so this mechanism informs the user what network is the best to use
    
#### Example of Route data in the DB:</u>
    routes":[{"destination":"0.0.0.0","family":2,"gateway":"192.168.127.1","interface":"ens3","metric
    ":100},{"destination":"0.0.0.0","family":2,"gateway":"192.168.145.1","interface":"e
    ns4","metric":101},{"destination":"10.88.0.0","family":2,"interface":"cni-podman0"}
    ,{"destination":"192.168.127.0","family":2,"interface":"ens3","metric":100},{"desti
    nation":"192.168.145.0","family":2,"interface":"ens4","metric":101},{"destination":
    "::1","family":10,"interface":"lo","metric":256},{"destination":"fe80::","family":1
    0,"interface":"cni-podman0","metric":256},{"destination":"fe80::","family":10,"inte
    rface":"ens3","metric":1024},{"destination":"fe80::","family":10,"interface":"ens4"
    ,"metric":1024}]
    
    In this case the selected CIDR will be: 192.168.127.0/24 which was
    calculated from the interface {"name": "ens3, "ipv4_addresses":["192.168.127.10/24"],"ipv6_addresses":[] ...}
    
#### Manual Tests
    Delete the machine network from the db and see that the correct route is chosen

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/common.go
+++ b/internal/cluster/common.go
@@ -272,7 +272,7 @@ func UpdateMachineCidr(db *gorm.DB, cluster *common.Cluster, machineCidr []strin
 		previousSecondaryMachineCidr = network.GetMachineCidrById(cluster, 1)
 	}
 
-	if machineCidr[0] != previousPrimaryMachineCidr || (len(machineCidr) > 1 && machineCidr[1] != previousSecondaryMachineCidr) {
+	if (len(machineCidr) > 0 && machineCidr[0] != previousPrimaryMachineCidr) || (len(machineCidr) > 1 && machineCidr[1] != previousSecondaryMachineCidr) {
 		err := db.Transaction(func(tx *gorm.DB) error {
 			if err := db.Where("cluster_id = ?", *cluster.ID).Delete(&models.MachineNetwork{}).Error; err != nil {
 				err = errors.Wrapf(err, "failed to delete machine networks of cluster %s", *cluster.ID)
--- a/internal/network/machine_network_cidr.go
+++ b/internal/network/machine_network_cidr.go
@@ -3,6 +3,7 @@ package network
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"net"
 	"strings"
 
@@ -415,6 +416,63 @@ func GetInventoryNetworksByFamily(hosts []*models.Host, log logrus.FieldLogger) 
 		ret[family] = append(ret[family], n)
 	}
 	return ret, nil
+}
+
+func GetDefaultRouteNetworkByFamily(h *models.Host, networks map[AddressFamily][]string, log logrus.FieldLogger) (map[AddressFamily]string, error) {
+	var err error
+	ret := make(map[AddressFamily]string)
+
+	//start with dummy route
+	defaultRoutev4 := &models.Route{Metric: math.MaxInt32}
+	defaultRoutev6 := &models.Route{Metric: math.MaxInt32}
+
+	if h.Inventory != "" {
+		var inventory models.Inventory
+		err = json.Unmarshal([]byte(h.Inventory), &inventory)
+		if err != nil {
+			log.WithError(err).Warnf("Unmarshal inventory %s", h.Inventory)
+			return ret, err
+		}
+		//find the minimal default route
+		var isDefault, found bool
+		for _, route := range inventory.Routes {
+			if isDefault, err = IsDefaultRoute(route); err != nil {
+				log.WithError(err).Errorf("fail to find default route")
+				return ret, err
+			}
+			if isDefault {
+				if IsIPv4Addr(route.Gateway) && route.Metric < defaultRoutev4.Metric {
+					defaultRoutev4 = route
+					found = true
+				} else if IsIPv6Addr(route.Gateway) && route.Metric < defaultRoutev6.Metric {
+					defaultRoutev6 = route
+					found = true
+				}
+			}
+		}
+
+		if !found {
+			return ret, fmt.Errorf("could not find default route")
+		}
+
+		//if the default route gateway is located in one of the cidrs
+		//in the inventory networks then return this network (per family)
+		for _, cidr := range networks[IPv4] {
+			if ipInCidr(defaultRoutev4.Gateway, cidr) {
+				ret[IPv4] = cidr
+				break
+			}
+		}
+		for _, cidr := range networks[IPv6] {
+			if ipInCidr(defaultRoutev6.Gateway, cidr) {
+				ret[IPv6] = cidr
+				break
+			}
+		}
+		log.Infof("available default route CIDRs: %+v", ret)
+		return ret, nil
+	}
+	return ret, fmt.Errorf("can not find cidr by route: no inventory for host %s", h.ID.String())
 }
 
 func IsHostInPrimaryMachineNetCidr(log logrus.FieldLogger, cluster *common.Cluster, host *models.Host) bool {
--- a/internal/network/utils.go
+++ b/internal/network/utils.go
@@ -1,6 +1,7 @@
 package network
 
 import (
+	"fmt"
 	"net"
 	"net/http"
 	"sort"
@@ -13,6 +14,20 @@ import (
 	"github.com/thoas/go-funk"
 	"gorm.io/gorm"
 )
+
+func IsDefaultRoute(r *models.Route) (bool, error) {
+	gw := net.ParseIP(r.Gateway)
+	if gw == nil {
+		return false, nil //gateway is empty for non default routes
+	}
+
+	dst := net.ParseIP(r.Destination)
+	if dst == nil {
+		return false, fmt.Errorf("unable to parse destination IP: %s", r.Destination)
+	}
+
+	return dst.IsUnspecified() && !gw.IsUnspecified(), nil
+}
 
 // Obtains the IP addresses used by a host
 func GetInventoryIPAddresses(inventory *models.Inventory) ([]string, []string) {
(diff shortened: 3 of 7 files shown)


internal/network/machine_network_cidr.go
```diff
@@ -417,6 +450,57 @@ func GetInventoryNetworksByFamily(hosts []*models.Host, log logrus.FieldLogger)
...
+	cidr   string
+	metric int32
+}
+
+func GetDefaultRouteNetworkByFamily(hosts []*models.Host, log logrus.FieldLogger) map[AddressFamily]string {
+	ret := make(map[AddressFamily]string)
+	cidrmap := make(map[AddressFamily][]CidrMetric)
+	var err error
+
+	for _, h := range hosts {
```
ori-amizur, line 463: Maybe it is better to find first the selected route, and later find appropriate CIDR for it?

internal/network/utils.go
```diff
@@ -14,6 +15,18 @@ import (
 	"gorm.io/gorm"
 )
 
+func IsDefaultRoute(r *models.Route) (bool, error) {
+	dst := net.ParseIP(r.Destination)
```
ori-amizur, line 19: Isn't destination a CIDR ?

internal/network/machine_network_cidr.go
```diff
@@ -417,6 +450,57 @@ func GetInventoryNetworksByFamily(hosts []*models.Host, log logrus.FieldLogger)
...
+				}
+			}
+		}
+	}
+
+	//per family choose the cidr with the minimal metric
+	for family, cidrWithRoutes := range cidrmap {
+		cidrs := cidrWithRoutes
+		sort.SliceStable(cidrs, func(i, j int) bool {
+			return cidrs[i].metric < cidrs[j].metric
```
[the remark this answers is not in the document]
slaviered: Since it this function is used for SNO exclusively how about I will change its signature to receive a  single host? In that case it will be clear that the cidrmap holds only data relevant to a single source.

internal/network/utils.go
slaviered: It is a CIDR when the route is not default. At any rate this logic need to change to support both options

internal/network/machine_network_cidr.go
```diff
@@ -417,6 +450,57 @@ func GetInventoryNetworksByFamily(hosts []*models.Host, log logrus.FieldLogger)
 	return ret, nil
 }
 
+type CidrMetric struct {
+	cidr   string
+	metric int32
+}
+
+func GetDefaultRouteNetworkByFamily(hosts []*models.Host, log logrus.FieldLogger) map[AddressFamily]string {
```
tsorya, line 458: Should not it be only for sno? why to pass hosts?

internal/network/machine_network_cidr.go
```diff
@@ -417,6 +450,57 @@ func GetInventoryNetworksByFamily(hosts []*models.Host, log logrus.FieldLogger)
...
+	ret := make(map[AddressFamily]string)
+	cidrmap := make(map[AddressFamily][]CidrMetric)
+	var err error
+
+	for _, h := range hosts {
+		if h.Inventory != "" {
+			var inventory models.Inventory
+			err = json.Unmarshal([]byte(h.Inventory), &inventory)
```
tsorya, line 466: maybe to use UnmarshalInventory from common?

internal/network/machine_network_cidr.go
```diff
@@ -417,6 +450,57 @@ func GetInventoryNetworksByFamily(hosts []*models.Host, log logrus.FieldLogger)
...
+			if err != nil {
+				log.WithError(err).Warnf("Unmarshal inventory %s", h.Inventory)
+				continue
+			}
+			for _, route := range inventory.Routes {
+				if isDefault, err := IsDefaultRoute(route); !isDefault {
+					if err != nil {
+						log.Error(err)
```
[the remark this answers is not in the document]
tsorya: Lets log  error with relevant message in IsDefaultRoute, this i function is already too large

internal/network/machine_network_cidr.go
tsorya: Agree with Ori

internal/network/utils.go
tsorya: From what i see default route is the one that has 0.0.0.0 as destination no?

internal/network/machine_network_cidr.go
slaviered: It can be done but really there is usually a single default route (or 2 at most). This map already contains only default routes and then I am sorting them according to metrics. The matching between the route and the CIDR was done above

internal/network/utils.go
slaviered: I added the structure as appear in the DB into the PR text. The components of the routes are IPs.

internal/network/machine_network_cidr.go
slaviered: fixed

internal/network/machine_network_cidr.go
slaviered: I aligned with the code in the module. The change can be done in a separate PR across all the network utils

osherdp:
/uncc osherdp
I guess original reviewers are enough here, I wonder why prow tags more people when returning from draft :thinking:

internal/cluster/common.go
```diff
@@ -272,7 +272,7 @@ func UpdateMachineCidr(db *gorm.DB, cluster *common.Cluster, machineCidr []strin
 		previousSecondaryMachineCidr = network.GetMachineCidrById(cluster, 1)
 	}
 
-	if machineCidr[0] != previousPrimaryMachineCidr || (len(machineCidr) > 1 && machineCidr[1] != previousSecondaryMachineCidr) {
+	if (len(machineCidr) > 0 && machineCidr[0] != previousPrimaryMachineCidr) || (len(machineCidr) > 1 && machineCidr[1] != previousSecondaryMachineCidr) {
```
ori-amizur, line 275: Maybe it is a good opportunity to call it machineNetworks and nor machineCidr ?

ori-amizur, line 275: Not related to your code - what is the point to open transaction and not use the transaction object ?

internal/network/machine_network_cidr.go
```diff
@@ -417,6 +418,64 @@ func GetInventoryNetworksByFamily(hosts []*models.Host, log logrus.FieldLogger)
...
+		//find the minimal default route
+		var isDefault, found bool
+		for _, route := range inventory.Routes {
+			if isDefault, err = IsDefaultRoute(route); err != nil {
+				log.WithError(err).Errorf("fail to find default route")
+				return ret, err
+			}
+			if isDefault {
+				if IsIPv4Addr(route.Gateway) && route.Metric < defaultRoutev4.Metric {
+					log.Info("==> found ipv4 route")
```
ori-amizur, line 445: Is this staying ?

internal/network/machine_network_cidr.go
```diff
@@ -417,6 +418,64 @@ func GetInventoryNetworksByFamily(hosts []*models.Host, log logrus.FieldLogger)
...
+					defaultRoutev4 = route
+					found = true
+				} else if IsIPv6Addr(route.Gateway) && route.Metric < defaultRoutev6.Metric {
+					log.Info("==> found ipv6 route")
```
ori-amizur, line 449: Is this staying ?

internal/network/machine_network_cidr.go
```diff
@@ -417,6 +418,64 @@ func GetInventoryNetworksByFamily(hosts []*models.Host, log logrus.FieldLogger)
...
+				break
+			}
+		}
+		for _, cidr := range networks[IPv6] {
+			if ipInCidr(defaultRoutev6.Gateway, cidr) {
+				ret[IPv6] = cidr
+				break
+			}
+		}
+		return ret, nil
```
ori-amizur, line 474: If default route was found for only 1 family, is it OK to return the map for both ?

internal/network/utils.go
ori-amizur: So it can either be CIDR or IP depending if it is default or not ?

internal/cluster/common.go
slaviered: It is an existing code. I'd like not to change it in this PR. We can open a different one for fixing this function including the transaction issue

slaviered: see my comment above on fixing it in a separate PR

internal/network/machine_network_cidr.go
slaviered: removed and replaced by a log at the end of the function

internal/network/machine_network_cidr.go
slaviered: see above comment

internal/network/machine_network_cidr.go
slaviered: It is necessary because the calling function is working per family. in the calling function we check per family if we have an input and proceed accordingly

internal/network/utils.go
ori-amizur: This is a bug.  Should be fixed in the agent and all places that use it.

slaviered: But not in this PR but in a separate task

PR #4774, opened 2022-12-13
NO-ISSUE: Pass HA mode of the cluster to network validations

mkowalski:
This PR adds an ability for network validation functions to access the HA mode of the cluster. This will be useful when those validations will have a different logic depending on whether the cluster is SNO or multi-node.

/cc @nmagnezi


nmagnezi (approved):
/lgtm

PR #4786, opened 2022-12-14
MGMT-12237: Limit upgrade agent events

jhernand:
When agent fails to upgrade the generated events will be generated each
time that the agent asks for next steps, approxy once per minute. This
can fill the collection of events with useless repeated information. To
avoid that this patch changes the event handling so that it will be
possible to define a minimum distance (in time) between those events.
Events that are closer than that minimum distance will be discarded.
    
Related: https://issues.redhat.com/browse/MGMT-12237

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/events/event.go
+++ b/internal/events/event.go
@@ -36,48 +36,6 @@ func New(db *gorm.DB, authz auth.Authorizer, log logrus.FieldLogger) eventsapi.H
 	}
 }
 
-func (e *Events) saveEvent(ctx context.Context, clusterID strfmt.UUID, hostID *strfmt.UUID, category string, severity string, message string, t time.Time, requestID string, props ...interface{}) error {
-	log := logutil.FromContext(ctx, e.log)
-	tt := strfmt.DateTime(t)
-	uid := clusterID
-	rid := strfmt.UUID(requestID)
-
-	additionalProps, err := toProps(props...)
-	if err != nil {
-		log.WithError(err).Error("failed to parse event's properties field")
-	}
-	event := common.Event{
-		Event: models.Event{
-			EventTime: &tt,
-			ClusterID: &uid,
-			Severity:  &severity,
-			Category:  category,
-			Message:   &message,
-			RequestID: rid,
-			Props:     additionalProps,
-		},
-	}
-	if hostID != nil {
-		event.HostID = hostID
-	}
-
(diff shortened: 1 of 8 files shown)


internal/events/event.go
```diff
@@ -125,9 +84,59 @@ func (e *Events) v2SaveEvent(ctx context.Context, clusterID *strfmt.UUID, hostID
 			tx.Commit()
 		}
 	}()
+
+	// Check and if the event exceeds the limits:
+	exceedsLimits, dberr := e.exceedsLimits(ctx, tx, &event)
+	if dberr != nil || exceedsLimits {
+		return
```
rccrdpccl, line 91: Should we fire a warning if this condition is tripped?

jhernand: If we do that then we will probably need another mechanism to limit the number of such warnings.

rccrdpccl: IMO we need to realise when something bad is happening, otherwise this kind of behaviour can go on undetected, which I believe we'd be worse off than generating millions of events.
One logging line is definitely less expensive than a row in the database, but as it stands we cannot generate alerts on logs (and we could generate them on database data)

jhernand: Done. The log message will look like this:

```
WARN[0446].../internal/events/event.go:176
github.com/openshift/assisted-service/internal/events.(*Events).reportDiscarded()
Event will be discarded
category=user
cluster_id=cb284f33-6c0e-4ed5-85c3-06ce366a4183
count=1
fields.time="2022-12-15T10:33:35.389+01:00" go-id=506 host_id=4c7bf9cc-da2f-4257-9c48-4f4cf4aa567c
infra_env_id=9d5d2cc0-8801-4153-893f-3b1358e1cc22
limit=1h0m0s
message="Host sno: Agent has been instructed to download image 'quay.io/jhernand123/agent:15' and restart"
name=upgrade_agent_started
pkg=events
props=
request_id=37ff57a9-ddf2-44aa-8631-7f2a119e5718 severity=info
```

PR #4787, opened 2022-12-14
MGMT-12445: Update host install progress response

CrystalChun:
[MGMT-12445](https://issues.redhat.com/browse/MGMT-12445)
When a host updates its install progress from an invalid state, the service
would initially respond with a 500 internal server error. This caused a lot
of noise that made it seem as if the service was down, when it wasn't.

This PR modifies the response to send a 409 conflict when updating
a host's install progress fails when updating the host from an invalid state.

Additionally, it modifies the error response sent to be the appropriate
code depending on the error that occurred.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

1. Deploy assisted-service with these changes
2. Install a cluster using assisted-test-infra and let it timeout (fail), ensure the hosts have reached an error state
3. Call the update host install progress API, ensure it's a 409 error returned with the correct error message
```bash
$ curl -X PUT ${SERVICE_URL}/api/assisted-install/v2/infra-envs/${INFRA_ENV_ID}/hosts/${HOST_ID}/progress -d '{"current_stage":"Rebooting", "progress_info":"test"}' -H "X-Secret-Key: ${TOKEN}" -H "Content-Type: application/json"
```

## Checklist

## Reviewers Checklist

/cc @carbonin

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -176,7 +176,7 @@ type InstallerInternals interface {
 	GetKnownApprovedHosts(clusterId strfmt.UUID) ([]*common.Host, error)
 	ValidatePullSecret(secret string, username string) error
 	GetInfraEnvInternal(ctx context.Context, params installer.GetInfraEnvParams) (*common.InfraEnv, error)
-	V2UpdateHostInstallProgressInternal(ctx context.Context, params installer.V2UpdateHostInstallProgressParams) (*common.Host, error)
+	V2UpdateHostInstallProgressInternal(ctx context.Context, params installer.V2UpdateHostInstallProgressParams) error
 }
 
 //go:generate mockgen --build_flags=--mod=mod -package bminventory -destination mock_crd_utils.go . CRDUtils
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -5234,29 +5234,25 @@ func (b *bareMetalInventory) V2GetHost(ctx context.Context, params installer.V2G
 }
 
 func (b *bareMetalInventory) V2UpdateHostInstallProgress(ctx context.Context, params installer.V2UpdateHostInstallProgressParams) middleware.Responder {
-	h, err := b.V2UpdateHostInstallProgressInternal(ctx, params)
+	err := b.V2UpdateHostInstallProgressInternal(ctx, params)
 	if err != nil {
-		if h == nil {
-			return installer.NewV2UpdateHostInstallProgressNotFound().
-				WithPayload(common.GenerateError(http.StatusNotFound, err))
-		}
-		return installer.NewV2UpdateHostInstallProgressInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+		return common.GenerateErrorResponder(err)
 	}
 	return installer.NewV2UpdateHostInstallProgressOK()
 }
-func (b *bareMetalInventory) V2UpdateHostInstallProgressInternal(ctx context.Context, params installer.V2UpdateHostInstallProgressParams) (*common.Host, error) {
+func (b *bareMetalInventory) V2UpdateHostInstallProgressInternal(ctx context.Context, params installer.V2UpdateHostInstallProgressParams) error {
 	log := logutil.FromContext(ctx, b.log)
 	log.Infof("Update host %s install progress", params.HostID)
 	host, err := common.GetHostFromDB(b.db, params.InfraEnvID.String(), params.HostID.String())
 	if err != nil {
 		log.WithError(err).Errorf("failed to find host %s", params.HostID)
-		return nil, err
+		return err
 	}
 
 	if host.ClusterID == nil {
 		err = fmt.Errorf("host %s is not bound to any cluster, cannot update progress", params.HostID)
 		log.WithError(err).Error()
-		return host, err
+		return common.NewApiError(http.StatusBadRequest, err)
 	}
 
 	stageChanged := params.HostProgress.CurrentStage != host.Progress.CurrentStage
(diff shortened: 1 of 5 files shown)


internal/bminventory/inventory.go
```diff
@@ -5263,7 +5263,7 @@ func (b *bareMetalInventory) V2UpdateHostInstallProgress(ctx context.Context, pa
 			return installer.NewV2UpdateHostInstallProgressNotFound().
 				WithPayload(common.GenerateError(http.StatusNotFound, err))
 		}
-		return installer.NewV2UpdateHostInstallProgressInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+		return installer.NewV2UpdateHostInstallProgressConflict().WithPayload(common.GenerateError(http.StatusConflict, err))
```
carbonin, line 5266: I think this is too wide. There are a lot of things that could go wrong in `V2UpdateHostInstallProgressInternal` that are worthy of 500 errors rather than 409.

I'm hoping that you can find the state incompatibility issues and return an [`ApiErrorResponse`](https://github.com/openshift/assisted-service/blob/5e951a16d3446d5c0188f3d4454687757537b7d5/internal/common/error_utils.go#L52) up from those then change this to use [`common.GenerateErrorResponder`](https://github.com/openshift/assisted-service/blob/5e951a16d3446d5c0188f3d4454687757537b7d5/internal/common/error_utils.go#L118-L133) which will either return the error it was given or create a new 500 error.

CrystalChun: Gotcha, thanks for the pointers! I went ahead and modified the error responses and used the common.GenerateErrorResponder. How does it look now?

internal/bminventory/inventory.go
```diff
@@ -5278,12 +5274,12 @@ func (b *bareMetalInventory) V2UpdateHostInstallProgressInternal(ctx context.Con
 		if stageChanged {
 			if err := b.clusterApi.UpdateInstallProgress(ctx, *host.ClusterID); err != nil {
 				log.WithError(err).Errorf("failed to update cluster %s progress", host.ClusterID)
-				return host, err
+				return common.NewApiError(http.StatusInternalServerError, err)
```
carbonin, line 5277: I think for something like this you can just return the error because we're using `GenerateErrorResponder` in the caller.

internal/host/host.go
```diff
@@ -548,14 +548,19 @@ func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, pro
 			"progress_progress_info":    progress.ProgressInfo,
 			"progress_stage_updated_at": strfmt.DateTime(time.Now()),
 		}
-		return m.db.Model(h).UpdateColumns(updates).Error
+		if err := m.db.Model(h).UpdateColumns(updates).Error; err != nil {
+			return common.NewApiError(http.StatusInternalServerError, err)
```
carbonin, line 552: I'd say just return the error here because we map the gorm not found errors to 404 in `GenerateErrorResponder`.

internal/host/host.go
```diff
@@ -626,7 +633,10 @@ func (m *Manager) UpdateInstallProgress(ctx context.Context, h *models.Host, pro
 			previousProgress.CurrentStage, progress.CurrentStage, progress.ProgressInfo, extra...)
 	}
 	m.reportInstallationMetrics(ctx, h, previousProgress, progress.CurrentStage)
-	return err
+	if err != nil {
+		return common.NewApiError(http.StatusInternalServerError, err)
```
carbonin, line 637: I think just returning the error here is fine. I don't think we can assert this is definitely a 500 here.

internal/bminventory/inventory.go
CrystalChun: Gotcha, modified as such! Thanks!

internal/host/host.go
CrystalChun: Gotcha, modified as such! Thanks!

internal/host/host.go
CrystalChun: Gotcha, modified as such! Thanks!

internal/bminventory/inventory.go
```diff
@@ -5234,29 +5234,25 @@ func (b *bareMetalInventory) V2GetHost(ctx context.Context, params installer.V2G
...
 }
-func (b *bareMetalInventory) V2UpdateHostInstallProgressInternal(ctx context.Context, params installer.V2UpdateHostInstallProgressParams) (*common.Host, error) {
+func (b *bareMetalInventory) V2UpdateHostInstallProgressInternal(ctx context.Context, params installer.V2UpdateHostInstallProgressParams) error {
 	log := logutil.FromContext(ctx, b.log)
 	log.Infof("Update host %s install progress", params.HostID)
 	host, err := common.GetHostFromDB(b.db, params.InfraEnvID.String(), params.HostID.String())
 	if err != nil {
 		log.WithError(err).Errorf("failed to find host %s", params.HostID)
-		return nil, err
+		return common.NewApiError(http.StatusNotFound, err)
```
carbonin, line 5249: I don't think this is definitely a 404. For example if the DB happens to be down or if we have some network error this should be 500. Since we use `GenerateErrorResponder` in the caller this can just return the error directly.

carbonin (commented):
Looking good. One more error code comment :sweat_smile:

internal/bminventory/inventory.go
CrystalChun: Thanks for catching all these! Modified it!

carbonin (approved):

PR #4791, opened 2022-12-15
MGMT-12950: Disable compatible agent validation while installing

jhernand:
Currently the agent compatibility validation is always enabled, even when the cluster is installing. If the service is restarted in the middle of the installation with a different version of the agent (with a different value for the `AGENT_DOCKER_IMAGE`) that results in warning event explaining that the agent isn't compatible with the service. That has no effect because the service will not be sending the upgrade agent step to the server, but it is confusing. This patch changes the service so that the validation is disabled when the cluster is installing.

Related: https://issues.redhat.com/browse/MGMT-12950

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/common.go
+++ b/internal/host/common.go
@@ -74,6 +74,21 @@ var manualRebootStages = []models.HostStage{
 	models.HostStageDone,
 }
 
+var allStages = []models.HostStage{
+	models.HostStageConfiguring,
+	models.HostStageDone,
+	models.HostStageFailed,
+	models.HostStageInstalling,
+	models.HostStageJoined,
+	models.HostStageRebooting,
+	models.HostStageStartingInstallation,
+	models.HostStageWaitingForBootkube,
+	models.HostStageWaitingForControlPlane,
+	models.HostStageWaitingForController,
+	models.HostStageWaitingForIgnition,
+	models.HostStageWritingImageToDisk,
+}
+
 var hostStatusesBeforeInstallation = [...]string{
 	models.HostStatusDiscovering, models.HostStatusKnown, models.HostStatusDisconnected,
 	models.HostStatusInsufficient, models.HostStatusPendingForInput,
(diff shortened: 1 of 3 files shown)


internal/host/common.go
```diff
@@ -74,6 +74,21 @@ var manualRebootStages = []models.HostStage{
 	models.HostStageDone,
 }
 
+var allStages = []models.HostStage{
```
flaper87, line 77: Should we call this `installationStages` ?

IIUC, this slice does not contain *all* the stages that a host can be in, does it?

jhernand: It does contains all the values of the `models.HostStage` enum. If I understand correctly all the values of that enum are installation stages, and there are no other stages. Note that there is not the host _state_, but the _stage_ which is only used to report the progress of the installation.

eranco74:
Seems that the issue isn't just during installation, we see the validation failure events in clusters that are already installed
```
validation 'compatible-agent' that used to succeed is now failing
```

internal/host/common.go
filanov: Maybe it will be worth looking at the state?
if the agent is the one that reporting the stage then we probably don't want to update when host is preparing for installation is well no?

eranco74: Hmmm, tricky one.
On the one hand, it's true that all stages are actually installation stages (we report the stage only during the installation), so the name is accurate. 
On the other hand, when you see the validation is marked with `skippedStates: allStages` it seems like it's always skipped (and not just during installations).
I guess a comment explaining that the stage is set only during installation should resolve it.

eranco74: > Maybe it will be worth looking at the state?
True

filanov: as you said it will resolve most of the cases, and we can fix the other cases later

PR #4793, opened 2022-12-15
MGMT-12423: Watch for Agent changes in ClusterDeployment

mkowalski:
In order to catch cases where changes to the Agent CR affect AgentClusterInstall but they do not go over the AI DB, we are adding a watch over Agents that reference the ACI in subject.

Because it's ClusterDeployment reconciler that in reality modifies the ACI CR, we are placing a watch in the CD CR and not the ACI CR.

Contributes-to: [MGMT-12423](https://issues.redhat.com//browse/MGMT-12423)

/cc @tsorya

### DIFF
--- a/internal/controller/controllers/clusterdeployments_controller.go
+++ b/internal/controller/controllers/clusterdeployments_controller.go
@@ -62,10 +62,13 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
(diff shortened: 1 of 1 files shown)


filanov:
/hold 
can we check the performance for this change? 
And as Mat suggested maybe we could look into other solutions? https://sdk.operatorframework.io/docs/building-operators/golang/references/event-filtering/

filanov:
/assign @mhrivnak 
@mhrivnak what do you think watch vs event filtering?

filanov:
Once it's merged lets backport into 2.6 we saw a similar case in a subsystem that failed when agent cluster install didn't catch this sync change in the agent 
cc: @slaviered

mhrivnak:
Can you provide an example of an Agent status change that requires reconciling the ClusterDeployment?

nmagnezi (changes requested):
What about adding tests to ensure it behaves as expected?
I think we need both UTs and subsystem tests.

mkowalski:
> Can you provide an example of an Agent status change that requires reconciling the ClusterDeployment?

Sure thing. Imagine the following steps

1) you create Agent with no ignition overrides; Agent has SpecSync OK and ClusterDeployment is Ready
2) you modify Agent by adding ignition override which has a wrong syntax; Agent goes into SpecSync Error

After step (2) you would expect ClusterDeployment to go into NotReady with the reason "You have 1 Agent with Error". This is not happening because change in Agent never triggered reconciliation. With this PR I am submitting here, a change in Agent's SpecSync will trigger a reconciliation.

mkowalski:
> What about adding tests to ensure it behaves as expected? I think we need both UTs and subsystem tests.

I agree 100% with this, but given this is some amount of work, I need to first have an agreement that the direction I take here (i.e. Watch with Predicate) is an approved design. Once this is settled, I will make this PR a proper piece of code with tests&co.

slaviered:
We already have a subsystem test that can benefit from this logic because it has a race condition with agent changes and aci changes: see '`deploy clusterDeployment with agent and invalid ignition config`'

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1528,11 +1531,72 @@ func (r *ClusterDeploymentsReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+			return []reconcile.Request{
+				{
+					NamespacedName: types.NamespacedName{
+						Namespace: agent.Spec.ClusterDeploymentName.Namespace,
+						Name:      agent.Spec.ClusterDeploymentName.Name,
+					},
+				},
+			}
+		} else {
+			return []reconcile.Request{}
```
paul-maidment, line 1556: I'm guessing the reason we don't warn about the lack of `agent.Spec.ClusterDeploymentName` in the log is that this is common and to report on it would generate too much noise?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1528,11 +1531,72 @@ func (r *ClusterDeploymentsReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+		}
+	}
+
+	agentSpecStatusChangedPredicate := builder.WithPredicates(predicate.Funcs{
+		UpdateFunc: func(updateEvent event.UpdateEvent) bool {
+			oldAgent, ok := updateEvent.ObjectOld.(*aiv1beta1.Agent)
+			if !ok {
+				return false
```
paul-maidment, line 1564: Should we log anything if we are having trouble obtaining the agent objects?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1528,11 +1531,72 @@ func (r *ClusterDeploymentsReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+			}
+
+			newAgent, ok := updateEvent.ObjectNew.(*aiv1beta1.Agent)
+			if !ok {
+				return false
```
paul-maidment, line 1569: Should we log anything if we are having trouble obtaining the agent objects?

internal/controller/controllers/clusterdeployments_controller.go
mkowalski: We don't warn about it, because this is not really a reason for a warning. Agent without a respective ClusterDeployment is still a valid agent (think about late-binding or some other scenarios).

We could at most `log.Debug` about it, but I don't personally see a big benefit

internal/controller/controllers/clusterdeployments_controller.go
mkowalski: Nope, because this is not a trouble *obtaining* the object, it would be a trouble with converting the type.

What we do here is we receive `updateEvent.ObjectOld` which we need to map into Agent object. It's purely implementation detail that we need this cast, because even though `updateEvent.ObjectOld` comes from the Watch over `aiv1beta1.Agent`, it doesn't have its type. That's because `updateEvent` itself has `event.UpdateEvent` type. We would need generics or something to have it more (or less) elegant, but as this comes from the framework, I just consume it "as is"

internal/controller/controllers/clusterdeployments_controller.go
mkowalski: As above, no because it's a framework

filanov (approved):

mkowalski:
/unhold

Trivial to revert if any perf issues are observed

PR #4777, opened 2022-12-13
NO-ISSUE: Add multiarch OCP 4.12 release images

mkowalski:
/cc @osherdp

PR #4795, opened 2022-12-15
NO-ISSUE: Add multiarch OCP 4.12 release images

mkowalski:
In this PR we are simplifying the way of serving multi-arch release images. We will expose those simply as `4.12-multi` and not by using the whole `X.Y.Z-multi`. Thanks to this we will be able to bump z-streams transparently without requiring users of the API/AICLI to adapt every time an update is available.

/cc @osherdp 

Supersedes: #4777


osherdp (approved):

osherdp:
So we can close https://github.com/openshift/assisted-service/pull/4777?

PR #4808, opened 2022-12-19
MGMT-12978: Allow version of the release image to be arbitrary

mkowalski:
Contributes-to: [MGMT-12978](https://issues.redhat.com//browse/MGMT-12978)

/cc @osherdp 
/cc @danielerez

### DIFF
--- a/tools/handle_ocp_versions.py
+++ b/tools/handle_ocp_versions.py
@@ -32,27 +32,21 @@ def verify_images(release_images: List[str]):
 def verify_release_version(ocp_version: str, release_image: str, release_version: str):
     """
     This function takes the custom definition of release image as expected by the Assisted Service
-    and performs the following checks
+    and performs the following check
 
-    1) 'openshift_version' field must match at least "major.minor" version as extracted from the
-       release image's "metadata.version"
-    2) 'version' field must fully match the "metadata.version" field of the release image
+    * 'openshift_version' field must match at least "major.minor" version as extracted from the
+      release image's "metadata.version"
 
     This logic allows us to serve release images in a simplified way so that e.g. image with
     "4.12.0-0.nightly-multi-2022-09-08-131900" as metadata can be offered as simply "4.12-nightly".
-
-    The requirement of the full match of the 'version' field ensures awareness of what is the
-    real version of the image that is being served.
     """
-
-    oc_version = get_oc_version(release_image)
-    assert oc_version == release_version, (f"{release_image} full version is {oc_version} not {release_version}")
 
     # Valid delimiters for versions are "." as well as "-". This is in order to cover extraction
     # for all the following combinations
     #   * "4.11.0"
     #   * "4.11-multi"
     #   * "4.12.0-0.nightly-multi-2022-09-08-131900"
+    oc_version = get_oc_version(release_image)
     oc_version = oc_version.replace('-', '.')
     ocp_version = ocp_version.replace('-', '.')
 


danielerez:
Is it actually needed to make the 'version' property arbitrary? I thought we can just change the 'DisplayName' assignment [here](https://github.com/openshift/assisted-service/blob/master/internal/versions/api.go#L112). I.e. can't we just generate the display name as required on V2ListSupportedOpenshiftVersions?

mkowalski:
> Is it actually needed to make the 'version' property arbitrary? I thought we can just change the 'DisplayName' assignment [here](https://github.com/openshift/assisted-service/blob/master/internal/versions/api.go#L112). I.e. can't we just generate the display name as required on V2ListSupportedOpenshiftVersions?

Generate? Based on what exactly? The whole point here is that CVO does not return any `-multi` suffix, it has to come from the human. I'm open to suggestions though

danielerez:
> > Is it actually needed to make the 'version' property arbitrary? I thought we can just change the 'DisplayName' assignment [here](https://github.com/openshift/assisted-service/blob/master/internal/versions/api.go#L112). I.e. can't we just generate the display name as required on V2ListSupportedOpenshiftVersions?
> 
> Generate? Based on what exactly? The whole point here is that CVO does not return any `-multi` suffix, it has to come from the human. I'm open to suggestions though

Oh, I assumed we can use the indication we have in [GetReleaseArchitecture](https://github.com/openshift/assisted-service/blob/fdb0f5f2a59aa502ac1bc8d401bdec4ebd164f8b/internal/oc/release.go#L213) for that. If it's not enough, then makes sense.

mkowalski:
> Oh, I assumed we can use the indication we have in [GetReleaseArchitecture](https://github.com/openshift/assisted-service/blob/fdb0f5f2a59aa502ac1bc8d401bdec4ebd164f8b/internal/oc/release.go#L213) for that. If it's not enough, then makes sense.

Yep that's a solution we could embed here. But then we need to keep in mind, that multi-arch images will *always* be suffixed with no way of omitting it (i.e. you will not be able to transparently expose multi-arch image without any suffix). If we are okay with this, we can go this direction.

But at the same time, if we take this path, I will start asking a question why do we keep 2 separate fields to mark version in the JSON map, given that we start autocalculating the "display value" in the backend. Maybe @osherdp can shed some light - my understanding was always that `openshift_version` is a real value that backend uses and `version` is just a handicapped name for a field that has been created for the UI purposes. But as it's undocumented, it's only my assumption.

If the assumption is correct and we take the path of calculating display value in the backend, can we just remove the field from the JSON ?

danielerez:
> > Oh, I assumed we can use the indication we have in [GetReleaseArchitecture](https://github.com/openshift/assisted-service/blob/fdb0f5f2a59aa502ac1bc8d401bdec4ebd164f8b/internal/oc/release.go#L213) for that. If it's not enough, then makes sense.
> 
> Yep that's a solution we could embed here. But then we need to keep in mind, that multi-arch images will **always** be suffixed with no way of omitting it (i.e. you will not be able to transparently expose multi-arch image without any suffix). If we are okay with this, we can go this direction.
> 
> But at the same time, if we take this path, I will start asking a question why do we keep 2 separate fields to mark version in the JSON map, given that we start autocalculating the "display value" in the backend. Maybe @osherdp can shed some light - my understanding was always that `openshift_version` is a real value that backend uses and `version` is just a handicapped name for a field that has been created for the UI purposes. But as it's undocumented, it's only my assumption.
> 
> If the assumption is correct and we take the path of calculating display value in the backend, can we just remove the field from the JSON ?

So the `openshift_version` is actually the 'key' of the release image ([OpenshiftVersions](https://github.com/openshift/assisted-service/blob/1810f337f8dc5b55644a13166bd089db0c5129a6/vendor/github.com/openshift/assisted-service/models/openshift_versions.go#L18-L19) is structured as a map). So `openshift_version` can either be an x.y or x.y.z version. E.g. if a user specify 4.12.1 in OpenshiftVersion when creating a cluster, we have logic to fallback to the 4.12 entry if the 4.12.1 doesn't exist. So the 'version' property always refers to the full version (x.y.z-*), and that's the property we use when [creating](https://github.com/openshift/assisted-service/blob/be2c5764fd0f2d90750d650aff267cda579fa4ef/internal/bminventory/inventory.go#L543) a cluster, and also on [getSupportLevel](https://github.com/openshift/assisted-service/blob/4d7f5ba7253823d61243279f4d368cab974d9b0d/internal/versions/api.go#L146). That's the reason I'm a bit hassistant of making it a 'free-text' property. Just need to be careful if taking this route.
Anyway, I guess the question is indeed whether always having the 'multi' suffix is a problem...

mkowalski:
Right, that sounds reasonable. So I'll take the path of appending `-multi` to the code you mentioned whenever we detect we are working with multi-arch release image. Thanks for suggestions!

mkowalski:
/close

Superseded by https://github.com/openshift/assisted-service/pull/4811

PR #4817, opened 2022-12-20
MGMT-12978: Handle multiple images with the same OCP version

mkowalski:
This PR fixes the behaviour of filtering release images by architecture in case there are multiple images that have exactly the same version as reported by the CVO.

As a consequence, it was no longer possible to correctly handle single- and multi-arch release payloads that correspond to the same OCP version. With this change, we again have ability to support this scenario.

Contributes-to: [MGMT-12978](https://issues.redhat.com//browse/MGMT-12978)

/cc @danielerez 
/cc @carbonin

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -4,6 +4,7 @@ import (
 	context "context"
 	"fmt"
 	"runtime/debug"
+	"strings"
 
 	"github.com/go-openapi/swag"
 	"github.com/hashicorp/go-version"
@@ -172,7 +173,11 @@ func (h *handler) getReleaseImageFromCache(openshiftVersion, cpuArchitecture str
 			return nil, err
 		}
 		releaseImage = funk.Find(releaseImages, func(releaseImage *models.ReleaseImage) bool {
-			return *releaseImage.OpenshiftVersion == versionKey
+			// Starting from OCP 4.12 multi-arch release images do not have "-multi" suffix
+			// reported by the CVO, but we still offer the sufix internally to allow for explicit
+			// selection or single- or multi-arch payload.
+			version := strings.TrimSuffix(*releaseImage.OpenshiftVersion, "-multi")
+			return version == versionKey
 		})
 	}
 
--- a/internal/versions/versions_test.go
+++ b/internal/versions/versions_test.go
@@ -75,6 +75,24 @@ var defaultOsImages = models.OsImages{
 }
 
 var defaultReleaseImages = models.ReleaseImages{
+	// Two images below represent a scenario when the same OpenShift Version (as reported by the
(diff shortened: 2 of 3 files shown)


internal/versions/versions_test.go
```diff
@@ -75,6 +75,24 @@ var defaultOsImages = models.OsImages{
 }
 
 var defaultReleaseImages = models.ReleaseImages{
+	// Two images below represent a scenario when the same OpenShift Version (as reported by the
+	// CVO) is provided by more than a single release image. This is a scenario when for the same
+	// OCP version we have single-arch and multi-arch image. This happens because starting from
```
carbonin, line 80: Why would we do this? Shouldn't we only provide the multi-arch image in this case? Or do some multi-arch images not include all the possible architectures?

mkowalski: No, because multi-arch is not supported. For 4.12 we offer both multi-arch and single-arch. But it must be always the user's conscious decision to pick the multi-arch one

internal/versions/versions.go
```diff
@@ -172,7 +173,11 @@ func (h *handler) getReleaseImageFromCache(openshiftVersion, cpuArchitecture str
 			return nil, err
 		}
 		releaseImage = funk.Find(releaseImages, func(releaseImage *models.ReleaseImage) bool {
-			return *releaseImage.OpenshiftVersion == versionKey
+			// Starting from OCP 4.12 multi-arch release images do not have "-multi" suffix
+			// reported by the CVO, but we still offer the sufix internally to allow for explicit
+			// selection or single- or multi-arch payload.
+			version := strings.TrimSuffix(*releaseImage.OpenshiftVersion, "-multi")
```
carbonin, line 179: Is this really the field we should be using to filter?

I know that's how it was before, but if this is intended to be a user-facing field then shouldn't we use `Version` instead which is the same for single and multi-arch?

mkowalski: I think "that's how it was before" is the best answer here. If you asked me, I would be in favour of taking the time and rewriting the struct used to store the release image so that it does not hold `OpenshiftVersion` and `Version` fields which both don't have a well-defined usage (e.g. for some time I thought that only one of them is used by the backend and the other is exclusively for UI, but that's not true).

We should have only 1 version field. We should not have a field that artificially duplicates content reported by the CVO, because we have ability to extract this value ourselves. IMHO the simplest ReleaseImage type should hold only `URL` and `DisplayName`. Even architecture shouldn't be needed, because once you have the URL, you can get all the other info.

But those are big changes. It does not make our life easier that lots of them are only affecting UI which is not testable and you only see the effect once you manually rollout (because no CI will stop you from merging change that makes UI show silly stuff).

carbonin (approved):

PR #4822, opened 2022-12-21
MGMT-13038: Git fails to trust git repository because of mismatch with files

adriengentil:
git was updated in centos 9 stream repository and now checks if the owner of the files on the filesystem match the current user.

This change disable the check because Prow runs jobs with a random user.


adriengentil:
/retitle MGMT-13038: Git fails to trust git repository because of mismatch with files

osherdp (approved):

PR #4823, opened 2022-12-21
MGMT-12292: stream events to kafka stream

rccrdpccl:
Implementation for pushing events to kafka streams.

Instead of polling events from assisted-events-scraper, we will be pushing them from the service.

We will be sending a few types of events:
- current events, signaling that something happened
- cluster state changes: to capture important changes in clusters objects
- host state changes: to capture changes in host state
- infraenv changes: to capture changes in infraenv

Additionally, we will attach metadata to each event (for now only version metadata). Although this duplicates information, it simplifies the process as we do not need to relate this data.

Signed-off-by: Riccardo piccoli <rpiccoli@redhat.com>

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

This has been tested manually by running some basic operations (and unit tests), waiting for CI for more complex testing.

## Checklist

## Reviewers Checklist


rccrdpccl:
I've split the work in two commits for readability.

rccrdpccl:
FYI @gamli75 @adriengentil

jhernand:
@rccrdpccl I didn't look deeply into the changes yet, but I wanted to ask if you took into account that the Kafka library that you selected uses CGO and it statically links a version of librdkafka and its dependencies. In particular it includes static versions of gzip, curl and of OpenSSL. That means that any fix for those libraries will force us to rebuild our binary with a version of the Kafka library that in turn contains the required fixes. That may be challenging. I think that we already gave up on pure static binaries with the recent introduction of nmstate. If we are going to follow on that path I'd suggest to see if we can dynamically link libdrkafka, and therefore use the underlying distribution packages. That way it would be easier update.

internal/events/stream/event_test.go
```diff
@@ -0,0 +1,60 @@
...
+	)
+	BeforeEach(func() {
+		ctrl = gomock.NewController(GinkgoT())
+		stream = eventstream.NewMockEventStreamWriter(ctrl)
+	})
+
+	It("return error if stream is not configured", func() {
+
+		res := StreamEvent(nil, "TestEvent", "mykey", nil)
+		Expect(res).ShouldNot(BeNil())
```
osherdp, line 30:
```suggestion
		err := StreamEvent(nil, "TestEvent", "mykey", nil)
		Expect(err).ShouldNot(BeNil())
```

internal/events/stream/event_test.go
```diff
@@ -0,0 +1,60 @@
...
+	})
+
+	It("sends event with regular payload", func() {
+
+		payload := &TestStruct{
+			Foo: "bar",
+			Bar: "foo",
+		}
+		expectedPayload := []byte(`{"name":"TestEvent","payload":{"foo":"bar","bar":"foo"},"metadata":{"versions":{"assisted-installer":"quay.io/edge-infrastructure/assisted-installer:latest","assisted-installer-controller":"quay.io/edge-infrastructure/assisted-installer-controller:latest","assisted-installer-service":"Unknown","discovery-agent":"quay.io/edge-infrastructure/assisted-installer-agent:latest"}}}`)
```
osherdp, line 39: can we indent it for readability?
maybe even construct it as a map and marshal afterwards

internal/events/stream/event_test.go
osherdp: here and elsewhere

internal/events/stream/event_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+		expectedPayload := []byte(`{"name":"TestEvent","payload":null,"metadata":{"versions":{"assisted-installer":"quay.io/edge-infrastructure/assisted-installer:latest","assisted-installer-controller":"quay.io/edge-infrastructure/assisted-installer-controller:latest","assisted-installer-service":"Unknown","discovery-agent":"quay.io/edge-infrastructure/assisted-installer-agent:latest"}}}`)
+		stream.EXPECT().Write([]byte("mykey"), expectedPayload).Times(1).Return(nil)
+		res := StreamEvent(stream, "TestEvent", "mykey", payload)
+		Expect(res).To(BeNil())
+	})
+
+})
+
+func TestStream(t *testing.T) {
```
osherdp, line 57: don't we miss here validating the metadata for versions-info?

internal/bminventory/inventory_test.go
```diff
@@ -16371,3 +16376,93 @@ var _ = Describe("UpdateIgnitionEndpointIfHasMCSCert", func() {
...
+				ID: &infraEnvID,
+			}
+			mockStream.EXPECT().Write(gomock.Any(), gomock.Any()).Times(0)
+			response := bm.updateInfraEnv(db, &infraEnv, updates, true)
+			Expect(response.Error).To(BeNil())
+		})
+	})
+
+	When("Updating inventory", func() {
+		It("Should succeed even when we fail sending and event", func() {
```
osherdp, line 16434:
```suggestion
		It("Should succeed even when we fail sending an event", func() {
```

internal/bminventory/inventory.go
```diff
@@ -279,6 +284,28 @@ func NewBareMetalInventory(
...
+		key := infra.ClusterID.String()
+		streamErr := eventstream.StreamEvent(
+			b.stream,
+			"InfraEnv",
+			key,
+			infraEnv,
+		)
+
+		if streamErr != nil {
+			b.log.Warnf("Error while streaming infra_env update: %s", streamErr)
```
osherdp, line 303:
```suggestion
			b.log.WithError(streamErr).Warnf("Error while streaming infra_env update")
```

osherdp (changes requested):
partial review

pkg/stream/event_stream.go
```diff
@@ -0,0 +1,98 @@
...
+		_ = cfg.SetKey("sasl.password", config.ClientSecret)
+	}
+	if config.SecurityProtocol != "" {
+		_ = cfg.SetKey("security.protocol", config.SecurityProtocol)
+	}
+	if config.SASLMechanism != "" {
+		_ = cfg.SetKey("sasl.mechanism", config.SASLMechanism)
+	}
+	return cfg
+
```
osherdp, line 60: nit: delete blank line

pkg/stream/event_stream.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func NewKafkaWriter() (*KafkaWriter, error) {
+	config := &KafkaConfig{}
+	err := envconfig.Process("", config)
+	if err != nil {
+		return nil, err
+	}
+	cfg := getKafkaCfg(config)
```
osherdp, line 69:
```suggestion
	configMap := getKafkaConfigMap(config)
```
It took me a while to understand why we need ``config`` and ``cfg`` :sweat_smile:

pkg/stream/event_stream_test.go
```diff
@@ -0,0 +1,61 @@
...
+				Partition: kafka.PartitionAny,
+			},
+			Key:   key,
+			Value: value,
+		}
+		mockProducer.EXPECT().Produce(expectedMessage, nil).Times(1).Return(nil)
+
+		result := writer.Write(key, value)
+
+		Expect(result).To(BeNil())
```
osherdp, line 45:
```suggestion
		err := writer.Write(key, value)

		Expect(err).To(BeNil())
```

osherdp: here and elsewhere. ``result`` is a very generic term that hides-away the true meaning of the returned value

cmd/main.go
```diff
@@ -851,3 +854,19 @@ func createVersionHandlers(log logrus.FieldLogger, ctrlMgr manager.Manager, rele
 
 	return versionsHandler, versionsAPIHandler, nil
 }
+
+func getKafkaWriter(log *logrus.Logger) *stream.KafkaWriter {
+	kafkaWriter, err := stream.NewKafkaWriter()
+	if err != nil {
+		// We do not want to fail if connection with kafka is not possible.
+		// We will loose events, but the service will keep running
```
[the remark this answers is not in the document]
rccrdpccl: Can you please expand on how you think this could be an issue with logs?

I agree that it is an error, but this should not fail the service.

deploy/assisted-service.yaml
```diff
@@ -39,6 +39,10 @@ spec:
             - configMapRef:
                 name: assisted-service-config
           env:
+            - name: KAFKA_BOOTSTRAP_SERVER
```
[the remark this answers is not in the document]
rccrdpccl: Bootstrap server, in kafka context, is where the client connects to retrieve metadata about the cluster (what broker it should connect to, what topics are there, etc)

internal/events/stream/event_test.go
rccrdpccl: We are validating it right? Are you suggesting to change the version and validate that it actually changes in the metadata? Wouldn't this better be tested inside the versions package instead?

internal/bminventory/inventory.go
```diff
@@ -1083,7 +1112,11 @@ func (b *bareMetalInventory) GenerateInfraEnvISOInternal(ctx context.Context, in
 	updates := map[string]interface{}{}
 	updates["generated_at"] = strfmt.DateTime(now)
 	updates["image_expires_at"] = strfmt.DateTime(now.Add(b.Config.ImageExpirationTime))
-	dbReply := b.db.Model(&common.InfraEnv{}).Where("id = ?", infraEnv.ID.String()).Updates(updates)
+	i := models.InfraEnv{
```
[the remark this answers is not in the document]
rccrdpccl: I was really running out of choices... is it better `i` or having `infraenv` and `infraEnv`? :sweat_smile: 
The trouble is that `common.InfraEnv` it's used, while I need `models.InfraEnv`

internal/events/stream/event_test.go
rccrdpccl: Although I would love to have better reliability here, I am afraid if we serialize it from a map it would lose the order.
If we populate the same data structure that we have in the productive code, we would be just replicating the same code in the test...
I'll try to think of something to improve readability but if you have other ideas please let me know.

cmd/main.go
```diff
@@ -851,3 +854,19 @@ func createVersionHandlers(log logrus.FieldLogger, ctrlMgr manager.Manager, rele
 
 	return versionsHandler, versionsAPIHandler, nil
 }
+
+func getKafkaWriter(log *logrus.Logger) *stream.KafkaWriter {
```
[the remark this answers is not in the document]
osherdp: I more thought about a boolean-field, because removing the entire configuration to enable and disable features is a bit messy

internal/events/stream/event_test.go
osherdp: no you're right, it's already validated 
I missed it because ``expectedPayload`` is such a blob...

internal/bminventory/inventory.go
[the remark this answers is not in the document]
osherdp: ``infraEnv`` is totally fine. what not?
I just don't like naming variables with one or two characters. it requires me as a reader to go back to the original definition and see what's the object's purpose

internal/events/stream/event_test.go
osherdp: too bad golang doesn't allow to preserve order in json marshaling

cmd/main.go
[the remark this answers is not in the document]
osherdp: If someone adds a bug somewhere around those lines, it gets unnoticed (unless somehow being spotted in the review, or if we somehow inspect the logs)
I prefer failing the service over this situation, especially that we only talk about the initial connection to kafka (and not a constant interaction with it, which can and should mask errors)

tools/deploy_kafka.py
```diff
@@ -0,0 +1,44 @@
...
+    utils.verify_build_directory(deploy_options.namespace)
+
+    deploy_kafka(deploy_options)
+
+    log.info('Completed kafka deployment')
+
+
+def deploy_kafka(deploy_options):
+    kafka_dep_file = 'deploy/kafka/kafka.yaml'
+    docs = utils.load_yaml_file_docs(kafka_dep_file)
```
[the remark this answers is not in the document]
osherdp: ahh, docs as in documents
I thought it's something like documentation..

internal/bminventory/inventory.go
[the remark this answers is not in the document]
rccrdpccl: Yes, I was mentioning that I already had an `infraEnv` variable, so naming another `infraenv` would be confusing (note casing difference).

I have found another way though :)

internal/events/stream/event_test.go
jhernand: If you really need to preserve order in JSON documents you could use this library (or any other similar one):

https://gitlab.com/c0b/go-ordered-json

But I'd say that in this context it would be preferable to compare the expected and actual JSON documents ignoring order. For that I'd suggest to create a new gomock matcher similar to the gomega `MatchJSON`:

https://onsi.github.io/gomega/#matchjsonjson-interface

Could be something like this:

```go
func JSON(expected string) gomock.Matcher {
	var expectedAny any
	err := json.Unmarshal([]byte(expected), &expectedAny)
	if err != nil {
		panic(err)
	}
	buffer := &bytes.Buffer{}
	encoder := json.NewEncoder(buffer)
	encoder.SetIndent("", "\t")
	err = encoder.Encode(expectedAny)
	if err != nil {
		panic(err)
	}
	expectedText := buffer.String()
	return &jsonMatcher{
		expectedText: expectedText,
		expectedAny:  expectedAny,
	}
}

type jsonMatcher struct {
	expectedText string
	expectedAny  any
}

func (m *jsonMatcher) Matches(actual any) bool {
	actualText := actual.(string)
	var actualAny any
	err := json.Unmarshal([]byte(actualText), &actualAny)
	if err != nil {
		panic(err)
	}
	return reflect.DeepEqual(actualAny, m.expectedAny)
}

func (m *jsonMatcher) String() string {
	return m.expectedText
}
```

Then you should be able to do something like this in your test:

```go
stream.EXPECT().Write(
        []byte("mykey"),
        JSON(`{
                "name": "TestEvent",
                "payload": ...
        }`),
)
```

The matcher will ignore what isn't relevant in a JSON document, like extra space and ordering of fields.

jhernand:
@rccrdpccl what concerns me is the statically linked version of OpenSSL in particular. Imagine a CVE that affects OpenSSL. To address it with this approach we would have to wait for the team that develops librdkafka to rebuild and release it with the fixed version of OpenSSL. Then we would have to wait for the team that develops the Go library to update to the new version of librdkafka, and finally we would have to update our binary to use the new version of that Go library. That is a lot of time, and those CVE are unfortunately frequent. If the CVE is severe enough we would probably be pressed to speed things up and we could be forced to fork and rebuild those libraries ourselves.

Would it make sense to use a Kafka library that doesn't use CGO? For example, I have used https://github.com/segmentio/kafka-go in the past and it seems to work fine. Would be too much of a trouble to use it for this? Not that that will be free of CVEs and issues, but at least if we have to update it it would be like any other Go library that we use.

rccrdpccl:
> @rccrdpccl what concerns me is the statically linked version of OpenSSL in particular. Imagine a CVE that affects OpenSSL. To address it with this approach we would have to wait for the team that develops librdkafka to rebuild and release it with the fixed version of OpenSSL. Then we would have to wait for the team that develops the Go library to update to the new version of librdkafka, and finally we would have to update our binary to use the new version of that Go library. That is a lot of time, and those CVE are unfortunately frequent. If the CVE is severe enough we would probably be pressed to speed things up and we could be forced to fork and rebuild those libraries ourselves.
> 
> Would it make sense to use a Kafka library that doesn't use CGO? For example, I have used https://github.com/segmentio/kafka-go in the past and it seems to work fine. Would be too much of a trouble to use it for this? Not that that will be free of CVEs and issues, but at least if we have to update it it would be like any other Go library that we use.

Fair enough, I see your point, and although I think the way we use kafka we are not exposing a lot of surface for exploiting potential vulnerabilities, I understand the concern.

I had a first prototype with the above mentioned library (segmentio), and I've switched to confluent's mainly because it supported OAUTH auth mechanism... it turns out we don't need it though. Overall I think confluent library has a marginally better API, documentation and kafka support, but we can forgo those marginal advantages and reduce the risk, it's not a big change (and it won't be even in the case we'll need some of the confluent features in the future).

internal/events/stream/event_test.go
rccrdpccl: That is a great idea @jhernand

internal/events/event.go
```diff
@@ -82,6 +86,23 @@ func (e *Events) v2SaveEvent(ctx context.Context, clusterID *strfmt.UUID, hostID
...
+				"event": event,
+			}).Debugf("sending event")
+			err := eventstream.StreamEvent(
+				ctx,
+				e.stream,
+				"Event",
+				key,
+				event.Event,
+			)
+			if err != nil {
```
gamli75, line 103: what will happen if there is no connection to the Kafka instance? Does it mean that all the service logs will be full of those error messages?

rccrdpccl:
Note that the client has lazy connection, therefore:
* if we misconfigure kafka client settings to something invalid, main will break on startup (fatal)
* if we don't have connection to configured host (either broken server or wrong url), the app will start but will send a warning each and every event
* if we loose connection after starting, the same as above happens until service is restored
* if we disable kafka at the start, everything works as before this change

adriengentil:
> Note that the client has lazy connection, therefore:
> 
> * if we misconfigure kafka client settings to something invalid, main will break on startup (fatal)
> * if we don't have connection to configured host (either broken server or wrong url), the app will start but will send a warning each and every event
> * if we loose connection after starting, the same as above happens until service is restored
> * if we disable kafka at the start, everything works as before this change

would it make sense to expose a metric and have an alert to notify us that we fail to write events into kafka (and lose events)?

rccrdpccl:
> would it make sense to expose a metric and have an alert to notify us that we fail to write events into kafka (and lose events)?

Yes it would, totally, the plan is to wait for log-based alert to see the client's perspective, and have server-side exported metrics.
Unfortunately at the moment exposing client-perspective metrics is non-trivial (at least to do it tidily) due to a dependency issue (metrics manager depends on events handler - the main point from where we stream events).
I think it could be a good addition, but I would do it as a separate stage, once we sort out this dependencies issue. Happy to discuss it further if you don't agree :)

internal/bminventory/inventory.go
```diff
@@ -279,6 +284,39 @@ func NewBareMetalInventory(
...
 }
 
+func (b *bareMetalInventory) updateInfraEnvByID(ctx context.Context, db *gorm.DB, infraEnvID *strfmt.UUID, updates interface{}, streamEvents bool) *gorm.DB {
+	infraenv := models.InfraEnv{
+		ID: infraEnvID,
+	}
+	return b.updateInfraEnv(ctx, db, &infraenv, updates, streamEvents)
+}
+
+func (b *bareMetalInventory) updateInfraEnv(ctx context.Context, db *gorm.DB, infraEnv *models.InfraEnv, updates interface{}, streamEvents bool) *gorm.DB {
```
carbonin, line 294: I think this wrapper might not be needed.

It might be better to create a function for sending the event to the stream rather than one for updating the database and sending the event.

carbonin (commented):
Generally I think we should watch out for situations where we might by sending data to kafka before the database transaction is committed.

pkg/stream/event_stream.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	if config.ClientID != "" && config.ClientSecret != "" {
+		mechanism := &plain.Mechanism{
+			Username: config.ClientID,
+			Password: config.ClientSecret,
+		}
+		writer.Transport = &kafka.Transport{
+			SASL: mechanism,
+			// let config pick default root CA, but define it to force TLS
+			TLS: &tls.Config{},
```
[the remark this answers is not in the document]
rccrdpccl: yes, we are using TLS to transport this, however as the root CA is available in our underlying distribution there is no need to explicitly define which one to use, by just defining empty tls config it would pick it up from well-known system directories.
If we didn't want to encrypt, tls Config must be nil.

internal/events/stream/event.go
```diff
@@ -0,0 +1,52 @@
...
+	streamPayload, err := json.Marshal(envelope)
+	if err != nil {
+		return err
+	}
+	return stream.Write(ctx, []byte(key), streamPayload)
+}
+
+func getVersions() (*models.Versions, error) {
+	v := versions.Versions{}
+	err := envconfig.Process("", &v)
```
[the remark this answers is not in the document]
rccrdpccl: Good point, as discussed somewhere else it would be best to make it more generic and inject it in the entrypoint instead - thus freeing this package from the `internal` bit.

deploy/kafka/kafka.yaml
```diff
@@ -0,0 +1,237 @@
...
+            - name: KAFKA_ENABLE_KRAFT
+              value: "yes"
+            - name: KAFKA_INTER_BROKER_LISTENER_NAME
+              value: "PLAINTEXT"
+            - name: KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP
+              value: "CONTROLLER:PLAINTEXT,PLAINTEXT:PLAINTEXT"
+            - name: KAFKA_CFG_LISTENERS
+              value: "PLAINTEXT://:9092,CONTROLLER://:9093"
+            - name: KAFKA_CFG_ADVERTISED_LISTENERS
+              value: "PLAINTEXT://$(MY_POD_NAME).ai-kafka-headless.assisted-installer.svc.cluster.local:9092"
```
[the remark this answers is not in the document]
rccrdpccl: Regular service would load-balance connections among pods, but for Kafka we need to talk to specific instances (as happens with many stateful applications, the client fetches data from a specific host). Headless service allows us to talk directly with the needed instance (pod).

rccrdpccl:
Cluster claim timed out

rccrdpccl:
Turned off by default kafka deployment for integration tests.

rccrdpccl:
/test edge-subsystem-aws  edge-subsystem-kubeapi-aws
flaky quay

rccrdpccl:
/override edge-e2e-ai-operator-ztp-capi

Overridden as the job is currently broken

rccrdpccl:
@gamli75 @osherdp could you please review again :pray: ? All comments should have been addressed (if not the reason should be in the comments)

Makefile
```diff
@@ -345,6 +351,12 @@ deploy-postgres: deploy-namespace
 	python3 ./tools/deploy_postgres.py --namespace "$(NAMESPACE)" --target "$(TARGET)" \
 		--apply-manifest $(APPLY_MANIFEST) --persistent-storage $(PERSISTENT_STORAGE)
 
+deploy-kafka: deploy-namespace
+	python3 ./tools/deploy_kafka.py --namespace "$(NAMESPACE)" --target "$(TARGET)" \
+		--apply-manifest $(APPLY_MANIFEST)
+
+deploy-all-event-streaming: deploy-kafka deploy-all
```
osherdp, line 358: I want to limit the amount of makefile targets if that's possible
Can we deploy kafka as part of ``deploy-all`` based on the value of ``ENABLE_EVENT_STREAMING``?

cmd/main.go
```diff
@@ -851,3 +856,33 @@ func createVersionHandlers(log logrus.FieldLogger, ctrlMgr manager.Manager, rele
...
 }
+
+func getKafkaWriter(log *logrus.Logger) (*stream.KafkaWriter, error) {
+
+	metadata := map[string]interface{}{
+		"versions": Options.Versions,
+	}
+	eventStreamWriter, err := stream.NewKafkaWriterWithMetadata(metadata)
+	if err != nil {
+		return nil, errors.New("error instantiating kafka client")
```
osherdp, line 867: I think we should wrap the original error to make sure we don't lose this information

osherdp (changes requested):

osherdp (approved):

rccrdpccl:
/override ci/prow/edge-e2e-metal-assisted-ocs ci/prow/edge-e2e-metal-assisted
They passed many times in the past - they are flaky ATM

rccrdpccl:
/override ci/prow/e2e-agent-compact ci/prow/edge-e2e-ai-operator-ztp

PR #4826, opened 2022-12-22
MGMT-13045: Enhancement document for node labeling

ori-amizur:

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @filanov 
/cc @avishayt

### DIFF
--- a/docs/enhancements/support-node-labeling.md
+++ b/docs/enhancements/support-node-labeling.md
@@ -1,0 +1,167 @@
+---
+title: node labeling support
+authors:
+- "@oamizur"
+  creation-date: 2022-12-20
+  last-updated: 2023-01-30
+---
+
+# Support node labeling during installation
+
+## Summary
+
+There was a requirement to support node roles during installation.  These roles 
+are extra roles, which are used by the user and not by any Openshift components.
+These labels should be applied on a node in a spoke cluster.
+
+Although the requirement is specifically for roles, the implementation should support setting 
+any label on a node in a spoke cluster, regardless if the label represents a role.
+
+## Motivation
+
+Currently, nodes can be assigned with a single role which can be one of two values [ **master**, **worker**].
+The may want to add additional node labels that can be used as custom roles.  
+
+Each role in OCP are associated with machine config pool (MCP).  When a node reboots itself
+from the pointer ignition, it uses specific MCP that contains the specific configuration
+for the role.  This MCP has the format `rendered-<role>-*` (* stands for a hash string).  So there is MCP for master 
+role `rendered-master-*` and for worker `rendered-worker-*`.
+
+When adding custom role, it can be associated with custom MCP which references specific configuration
+for that role.  Setting appropriate label a node, causes the node to reboot and apply the 
+MCP configuration on that node.
+
+Setting node labels should support this flow.
+
+### Goals
+- For ZTP flow: add capability to add and update annotations on BMH which will be transferred 
+to agent CR.
+- For boot it yourself flow: add capability to add and update labels to agent CR after agent
+  has been created.
+- Transfer the labels from the agent CR to the corresponding node on the spoke cluster.
+
+### Non goals
+- Single way to set node labels before agent creation both for ZTP flow, and boot it yourself flow.
+
+## Proposal
+
+Additional field **nodeLabels** will be added to the spec
+section of the agent CR.  This field will be a **map[string]string**  
+### ZTP flow
+Labels can be applied on every CR.  So there is no known limitation to add a label on agent.
+
+For the ZTP flow, there should a pattern for transferring an annotation from the BMH to an
+entry in the field **nodeLabels** of agent CR spec.
+The pattern should be **"bmac.agent-install.openshift.io.node-label."** prefix of the annotation key.  
+The key suffix of the annotation will be the key in the **nodeLabels** field in the agent CR. 
+The value of the annotation will be the value in the **nodeLabels** map.
+
+For example assuming we want to apply a label **node-role.kubernetes.io/infra=** on an
+a node.  Then the following annotation should be added to the BMH: **bmac.agent-install.openshift.io.node-label.node-role.kubernetes.io/infra=**.
+
+### Node labels on agent CR
+
+If boot it yourself flow is in use, the spec field **nodeLabels**  should be set 
+directly on the agent CR.
+
+### Moving labels to nodes
+
+#### Day 1 cluster
+When installing day 1 cluster, the implementation should make use of the **nodeLabels** field
+which is part of host DB record. Once the node becomes ready, the label is applied on the node.
+This functionality already exists in the assisted controller which runs as a pod on the 
+spoke cluster during installation.
+
+#### Day 2 nodes
+The flow can be supported only when working with Kubernetes API, since direct access to
+the cluster is required.  In this case, setting node labels should be done by a 
+Kuberenetes controller (probably agent controller).
+
+In addition, for day 2 nodes we need to support setting custom manifests.  This functionality is
+already supported for day 1 clusters by the assisted controller.  In order to allow adding new
+MCP to an existing day 2 cluster, the user will have to apply them on the existing cluster.
+
+### User Stories
+
+#### Story 1
+
+Let the user define roles specific to the business domain in addition to 
+the role that is currently assigned to nodes.  For example, storage, rt, monitoring.
+
+### Implementation Details/Notes/Constraints
+
+In order to apply custom role on a worker node using AI the following operations should be done:
+- Add node label specifying the custom role to be associated with the node. The custom role should 
+be defined using the the machinsm specified in this document.
+- Optional: Add custom machine config CRs specifying additional configurations that hace to be applied on 
+the node with the custom role. These manifests should be applied using the manifest API that already exists
+in AI.
+- Add custom machine config pool which unifies the worker machine config and the additional machine config CRs defined in the 
+previous step.  Additionally, this MCP should also be associated with the custom role.  As with MC,
+the MCP should be defined using the manifests API.
+
+#### ZTP flow 
+
+When using ZTP, the annotations in the BMH are set in the following manner: 
+
+```yaml
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: ostest-worker-0
+  namespace: assisted-installer
+  annotations:
+    bmac.agent-install.openshift.io.node-label.node-role.kubernetes.io/infra: ""
+  labels:
+    infraenvs.agent-install.openshift.io: "myinfraenv"
+spec:
+  online: true
+  bootMACAddress: 00:ec:ee:f8:5a:ba
+  automatedCleaningMode: disabled
+  bmc:
+    address: ....
+    credentialsName: bmc-secret
+```
+
+#### Boot it yourself flow
+
+Example that can be used to patch existing CR:
+
+```yaml
+apiVersion: agent-install.openshift.io/v1beta1
+kind: Agent
+metadata:
+  name: 51a7c23e-00d7-4b27-b2e4-053bf5ee9cf8
+  namespace: assisted-spoke-cluster
+spec:
+  nodeLabels:
+     "node-role.kubernetes.io/infra": ""
+
+```
+
+### Risks and Mitigations
+Currently, single MCP can be active on a specific node. Assuming that multiple roles
+are needed on a specific node, it implies that a role and a superset MCP need to be created. 
+The MCP should contain all machine configs that are needed be applied on the node.
+In addition, only single custom should be used on the node.  This role corresponds to 
+the superset MCP.
+## Design Details [optional]
+When custom role is applied to an existing node, the node is rebooted in order to apply
+the corresponding MCP.  The extra reboot is undesirable.
+
+There is a possibility to avoid the extra reboot by modifying the worker pointer 
+ignition for a specific host.  This modification will cause the node to be attached 
+to the desired MCP instead of the worker MCP.  When the node label which represents 
+the custom role will be applied, the node will already be attached to the desired
+MCP, so no extra reboot will be needed.
+Please see https://issues.redhat.com/browse/MGMT-13186 for more details.
+The effort to avoid extra reboot will be done as part of another epic: https://issues.redhat.com/browse/MGMT-13544.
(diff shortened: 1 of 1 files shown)


docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Goals
+- For ZTP flow: add capability to add and update labels on BMH which will be transferred 
+to agent CR.
+- For boot it yourself flow: add capability to add and update labels to agent CR after agent
+  has been created.
+
+### Non goals
+- Single way to set node labels before agent creation both for ZTP flow, and boot it yourself flow.
+- REST API support.  Node labeling will not be supported under REST API.
```
filanov, line 30: It's already supported, but only in day1 https://github.com/openshift/assisted-service/blob/2b0f66c4cdcdd2dc8556d7f65b7f3e0eac0c958a/swagger.yaml#L4116-L4119

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Proposal
+
+There are 2 implementation options: Using labels, or adding dedicated spec field.
+
+### Using labels
```
filanov, line 36: because we are talking about a specific case of adding labels to nodes i don't think that we should use the labels on the agent for that.

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,128 @@
...
+Labels can be applied on every CR.  So there is no known limitation to add a label on agent.
+
+For the ZTP flow, there should a pattern for transferring an annotation from the BMH to a label  
+in the corresponding agent.  The pattern is actually the prefix of the annotation.
+It should be **agent-install.openshift.io.user.role**.  These annotations will be transferred to 
+to labels on the corresponding agent CR.
+
+### Adding dedicated spec field in agent CR
+Additional field **userRoles** will be added to the spec
```
filanov, line 45: suggesting to be more generic and call it node labels
Roles is just one of the use cases here.

docs/enhancements/support-node-labeling.md
ori-amizur: OK, so it is non goal.

docs/enhancements/support-node-labeling.md
ori-amizur: OK

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,128 @@
...
+There was a requirement to support node roles during installation.  These roles 
+are extra roles, which are used by the user and not by any Openshift components.
+These labels should be applied on agent CR which represents a node in a spoke cluster.
+
+## Motivation
+
+Currently, nodes can be assigned with a single role which can be one of two values [ master, worker].
+The may want to add additional roles.  These roles are for the user usage only.
+
+### Goals
```
romfreiman, line 22: what about day2?

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,128 @@
...
+- "@oamizur"
+  creation-date: 2022-12-20
+  last-updated: 2022-12-20
+---
+
+# Support node labeling during installation
+
+## Summary
+
+There was a requirement to support node roles during installation.  These roles
```
filanov, line 13: roles is just a use case here, i suggest to have summary talk about the general use case, which is defining labels on the agent that will result in the nodes on the spoke cluster after the installation

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,128 @@
...
+section of the agent CR.  This field will be a string list.
+
+For the ZTP flow, the annotations having the format **agent-install.openshift.io.user.role/[key]=[value]**,
+the value part will be transferred to the **userRoles** field as additional entry. 
+
+### User Stories
+
+#### Story 1
+
+Let the user define roles specific to the business domain in addition to
```
romfreiman, line 55: what will it transfer to in OCP? what will it affect?

romfreiman:
@ori-amizur im looking into ROSA demo (2:45 min) - do see there node labels. Lets understand how it's being implemented with hive?
https://www.youtube.com/watch?v=gAMr3sI5bdY&t=169s

avishayt:
I agree with @filanov .  We have node_labels in the REST API, let's add it to the ACI Spec and a corresponding annotation on the BMH.

romfreiman:
I dont understand who/how this will help? It wont reduce any reboot, neither wont control mcp that the host will be part of.

romfreiman:
There is a work that was dont by @tsorya in OCP 4.12 that can help, but it does not really need a new API i beleive.

https://github.com/openshift/machine-config-operator/blob/master/templates/worker/01-worker-kubelet/on-prem/units/kubelet.service.yaml#L29

docs/enhancements/support-node-labeling.md
tsorya: You can't set custom role before node joins

docs/enhancements/support-node-labeling.md
tsorya: There is already field for node_labels, please just use it

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,128 @@
...
+For the ZTP flow, there should a pattern for transferring an annotation from the BMH to a label  
+in the corresponding agent.  The pattern is actually the prefix of the annotation.
+It should be **agent-install.openshift.io.user.role**.  These annotations will be transferred to 
+to labels on the corresponding agent CR.
+
+### Adding dedicated spec field in agent CR
+Additional field **userRoles** will be added to the spec 
+section of the agent CR.  This field will be a string list.
+
+For the ZTP flow, the annotations having the format **agent-install.openshift.io.user.role/[key]=[value]**,
```
tsorya, line 48: Not specific for role

tsorya:
> There is a work that was dont by @tsorya in OCP 4.12 that can help, but it does not really need a new API i beleive.
> 
> https://github.com/openshift/machine-config-operator/blob/master/templates/worker/01-worker-kubelet/on-prem/units/kubelet.service.yaml#L29

It allows to set labels when kubelet start though it doesn't allow to set roles and some more labels that are blocked by kubelet code

romfreiman:
> > There is a work that was dont by @tsorya in OCP 4.12 that can help, but it does not really need a new API i beleive.
> > https://github.com/openshift/machine-config-operator/blob/master/templates/worker/01-worker-kubelet/on-prem/units/kubelet.service.yaml#L29
> 
> It allows to set labels when kubelet start though it doesn't allow to set roles and some more labels that are blocked by kubelet code

right, but this api wont help neither

oourfali:
@ori-amizur regarding ROSA, iirc it is done via syncsets OCM defined in hive. Perhaps that can be leveraged via assisted in ACM as well.

docs/enhancements/support-node-labeling.md
oourfali: iiuc the RFE says they are doing it today at day-2 which is something they prefer to avoid.
Let's sit with Jose and Ian to understand the details.

docs/enhancements/support-node-labeling.md
filanov: but if it will exist in the DB we will need to take into account that the controller will use that

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,128 @@
...
+the value part will be transferred to the **userRoles** field as additional entry. 
+
+### User Stories
```
filanov, line 51: @imiller0 maybe you can provide additional user stories? 
it will make the requirement clearer. 
For now i'm not sure that fully understand the requirement or if we are even able to execute it with the existing tools.

docs/enhancements/support-node-labeling.md
ori-amizur: It will be done through the Kube API, so the controller has access to the spoke nodes.

docs/enhancements/support-node-labeling.md
ori-amizur: These roles do not have operational meaning from our POV.  These will be user defined attributes for their own use.

docs/enhancements/support-node-labeling.md
ori-amizur: OK.  We will discuss with @imiller0 to clarify the requirements.

docs/enhancements/support-node-labeling.md
tsorya: This is depends on what you mean by roles. I talked about k8s role and you can set them only after node joins k8s cluster

docs/enhancements/support-node-labeling.md
tsorya: I assume you mean service will have and not controller right?

docs/enhancements/support-node-labeling.md
tsorya: i assume you need to set them in cr and then in db as all other things we do, right? I just mean that field in db already exists and we need to set it in kubeapi flow

docs/enhancements/support-node-labeling.md
tsorya: It doesn't really mean what are specific requirements as we are going to set labels, what user means but this label doesn't really matter, it is still called label

docs/enhancements/support-node-labeling.md
ori-amizur: It will be transferred to labels on nodes

docs/enhancements/support-node-labeling.md
ori-amizur: Yes

docs/enhancements/support-node-labeling.md
ori-amizur: to labels on nodes.
It may effect the user role which may cause attaching to different MCP

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Proposal
+
+### ZTP flow
+Labels can be applied on every CR.  So there is no known limitation to add a label on agent.
+
+For the ZTP flow, there should a pattern for transferring an annotation from the BMH to a label  
+in the corresponding agent.  The pattern is actually the prefix of the annotation.
+It should be **node-label**.  The suffix of the annotation will be transferred 
+to labels on the corresponding agent CR.
```
avishayt, line 54: This section contradicts the next section ("Node labels on an agent CR").  Here you say that we are "transferring an annotation from the BMH to a label in the corresponding agent" but below you say that annotations "will be transferred to the dedicated field nodeLabels".

I think the next section is correct and you can erase this one.  We should not automatically copy labels between resources but rather have it explicit in the API.

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,159 @@
...
+
+#### Day 2 nodes
+The flow can be supported only when working with Kubernetes API, since direct access to
+the cluster is required.  In this case, setting node labels should be done by a 
+Kuberenetes controller (probably agent controller).
+
+In addition, for day 2 nodes we need to support setting custom manifests.  This functionality is
+already supported for day 1 clusters by the assisted controller.  In order to allow adding new
+MCP to existing day 2 clusters, this functionality needs to be supported by a Kubernetes controller - 
+probably the cluster-deployment controller.
```
avishayt, line 84: Adding MCPs to existing clusters feels out of scope for Assisted.

docs/enhancements/support-node-labeling.md
```diff
@@ -0,0 +1,159 @@
...
+
+For the ZTP flow, the annotations having the prefix **node-label** in BMH will be transferred
+to the dedicated field **nodeLabels** without the **node-label** prefix.
+
+If boot it yourself flow is in use, nodeLabels should be set directly on the agent CR.
+
+### Moving labels to nodes
+
+#### Day 1 cluster
+When installing day 1 cluster, the implementation should make use of the **nodeLabels** field
```
avishayt, line 71: Why nodeLabels and not machineConfigPoolName?

tsorya: I assume both. User will need to set mcp + role in nodelables

avishayt: @ori-amizur , this enhancement needs more details.
- The title indicates that the end goal is labeling nodes, but that's different from what the use case is.
- It would help to have a description of the steps required to do this with UPI and then what the steps would be with CIM after this proposal.  Here is a partial list of questions in my head after reading this: Does the user need to create an MCP?  How?  Does the user also need to use the machineConfigPoolName API?  How?  What is that API for and what is this label for?  If we need both machineConfigPoolName and label, can the existing API just add the label?

ori-amizur: Node labels can be used for many purposes - not necessarily roles.  This enhancement is for node labels, which custom roles are specific use-case for node labels.

ori-amizur: @avishayt although in terms of user story, MCP and MC are related to node labels, creation of MC and MCP are done by creating manifests.  This is already supported by AI for day1 installation.
As I indicated in the document, we should also support it for day2

docs/enhancements/support-node-labeling.md
ori-amizur: OK.

docs/enhancements/support-node-labeling.md
ori-amizur: OK

docs/enhancements/support-node-labeling.md
ori-amizur: MCP name is another field that already exists. No need to add it.

avishayt:
The document is missing an end-to-end flow.  I'm a user that wants to use a custom role - what APIs do I need to use?

ori-amizur:
> The document is missing an end-to-end flow. I'm a user that wants to use a custom role - what APIs do I need to use?

1.  Add node label
2. Add manifests: MachineConfigPool and possibly MachineConfig

PR #4798, opened 2022-12-16
MGMT-12894: Use the hub release image when determining the ironic agent image

carbonin:
Previously, when using the converged flow, the service attempted to extract the ironic agent image using the spoke release. This was not ideal for a few reasons:

1. The agent version should match that of the hub ironic service
    - While it was likely the agent from the spoke version would be compatible with the hub service it's objectively better for them to actually match.
2. The default image was used for late-binding cases
    - This meant that a rather old and seemingly random image was used when the service didn't have a cluster from which to pull a release image
3. It's difficult to communicate to a user what to mirror in disconnected cases

With this change the agent will always match the hub service, the hub agent will be used regardless of late-binding or not, and the user just needs to know to mirror the hub release into their spoke environment when disconnected.

The only case when we should use the default image now is if the hub release image doesn't support the CPU architecture of the spoke infraEnv, but with multi-arch images coming soon and differing hub/spoke architectures exceedingly rare already this shouldn't be a problem.

## List all the issues related to this PR

Resolves https://issues.redhat.com/browse/MGMT-12894

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

1. Deployed this change in a dev-scripts environment.
2. Created a bmh referencing an infraenv (late binding)
3. Downloaded discovery ignition for the infraenv
4. Validated the ironic agent image in the ignition matched the one from the hub release image

- [ ] assisted-test-infra environment
- [x] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

PR #4827, opened 2022-12-22
MGMT-10534: Add infra-env override for ironic agent image

carbonin:
Adds a field to the infra-env CR for overriding the ironic agent image when using the ZTP converged flow.
This also changes the behavior of the preprovisioning image reconciler so that we update the ironic ignition whenever it changes rather than only updating the ignition when the annotation is missing.

~~Built on top of https://github.com/openshift/assisted-service/pull/4798 (WIP until hub image PR is merged)~~ Merged
Resolves https://issues.redhat.com/browse/MGMT-10534

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?
1. Deployed a dev-scripts cluster
2. Created an infraenv + BMH
3. Observed the image from the hub cluster in the discovery ignition
4. Set the override image
5. Observed the override image used in the discovery ignition
6. Changed/removed override - got expected behavior

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/v1beta1/infraenv_types.go
+++ b/api/v1beta1/infraenv_types.go
@@ -103,6 +103,12 @@ type InfraEnvSpec struct {
 	// Applicable for both iPXE, and ISO streaming from Image Service.
 	// +optional
 	KernelArguments []KernelArgument `json:"kernelArguments,omitempty"`
+
+	// IronicAgentImageOverride is the pull spec for the ironic agent image to use when discovering hosts.
+	// Only applies when provisioning using a BareMetalHost.
+	// If unspecified, the ironic agent image from the hub cluster release is used.
+	// +optional
+	IronicAgentImageOverride string `json:"ironicAgentImageOverride,omitempty"`
 }
 
 type KernelArgument struct {
--- a/docs/hive-integration/README.md
+++ b/docs/hive-integration/README.md
@@ -55,7 +55,7 @@ More details on conditions is available [here](kube-api-conditions.md)
 #### Debug Information
 
 The `DebugInfo` field under `Status` provides additional information for debugging installation process:
-- `EventsURL` specifies an HTTP/S URL that contains events occured during cluster installation process
+- `EventsURL` specifies an HTTP/S URL that contains events occurred during cluster installation process
 
 
 
@@ -198,7 +198,7 @@ It will continue to:
 The assisted agent will not reboot the machine at the end of the installation, instead it will stop
 the assisted agent service and let the ironic agent to manage the machine power state
 
-The converged flow is disalbed by default, you can enable the converged flow by setting the `ALLOW_CONVERGED_FLOW` env to true [here](../operator.md##specifying-environmental-variables-via-configmap)
+The converged flow is disabled by default, you can enable the converged flow by setting the `ALLOW_CONVERGED_FLOW` env to true [here](../operator.md##specifying-environmental-variables-via-configmap)
 
 ### Ironic Agent Image
 
@@ -213,6 +213,9 @@ The environment var for the ironicAgent image to be used on X86_64 CPU architect
 `IRONIC_AGENT_IMAGE`
 The environment var for the ironicAgent image to be used on arm64 CPU architecture:
 `IRONIC_AGENT_IMAGE_ARM`
+
+The ironic agent image can also be overridden using the InfraEnv spec field `ironicAgentImageOverride`
+If this field is set this image will be used instead of the hub or default image.
 
 **NOTE**
 Ensure the correct images are mirrored if installing in a disconnected environment.
--- a/internal/controller/controllers/preprovisioningimage_controller.go
+++ b/internal/controller/controllers/preprovisioningimage_controller.go
@@ -356,15 +356,12 @@ func (r *PreprovisioningImageReconciler) mapInfraEnvPPI() func(a client.Object) 
 		reconcileRequests := []reconcile.Request{}
 
 		for i := range images {
-			// Don't queue if the Image URL and InfraEnv's URL is the same.
-			if images[i].Status.ImageUrl != infraEnv.Status.ISODownloadURL {
-				reconcileRequests = append(reconcileRequests, reconcile.Request{
-					NamespacedName: types.NamespacedName{
-						Namespace: images[i].Namespace,
-						Name:      images[i].Name,
-					},
-				})
-			}
+			reconcileRequests = append(reconcileRequests, reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: images[i].Namespace,
+					Name:      images[i].Name,
+				},
+			})
 		}
 		return reconcileRequests
 	}
--- a/internal/controller/controllers/preprovisioningimage_controller.go
+++ b/internal/controller/controllers/preprovisioningimage_controller.go
@@ -19,7 +19,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"strings"
 	"time"
 
@@ -138,8 +137,9 @@ func (r *PreprovisioningImageReconciler) Reconcile(origCtx context.Context, req 
 		return ctrl.Result{}, err
 	}
 
-	if !IronicAgentEnabled(log, infraEnv) {
-		return r.AddIronicAgentToInfraEnv(ctx, log, infraEnv)
+	infraEnvUpdated, err := r.AddIronicAgentToInfraEnv(ctx, log, infraEnv)
+	if infraEnvUpdated || err != nil {
+		return ctrl.Result{}, err
 	}
 
 	if infraEnv.Status.CreatedTime == nil {
@@ -371,8 +371,9 @@ func (r *PreprovisioningImageReconciler) mapInfraEnvPPI() func(a client.Object) 
 	return mapInfraEnvPPI
 }
 
-func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Context, log logrus.FieldLogger, infraEnv *aiv1beta1.InfraEnv) (ctrl.Result, error) {
-	// Retrieve infraenv from the database
+// AddIronicAgentToInfraEnv updates the infra-env in the database with the ironic agent ignition config if required
+// returns true when the infra-env was updated, false otherwise
+func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Context, log logrus.FieldLogger, infraEnv *aiv1beta1.InfraEnv) (bool, error) {
 	key := types.NamespacedName{
 		Name:      infraEnv.Name,
 		Namespace: infraEnv.Namespace,
@@ -380,11 +381,16 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
 	infraEnvInternal, err := r.Installer.GetInfraEnvByKubeKey(key)
 	if err != nil {
 		log.WithError(err).Error("failed to get corresponding infraEnv")
-		return ctrl.Result{}, err
+		return false, err
 	}
-	ironicAgentImage, err := r.getIronicAgentImageByRelease(ctx, log, infraEnvInternal)
-	if err != nil {
-		log.WithError(err).Warningf("Failed to get ironic agent image by release for infraEnv: %s", infraEnv.Name)
+	ironicAgentImage := infraEnv.Spec.IronicAgentImageOverride
+	if ironicAgentImage == "" {
+		ironicAgentImage, err = r.getIronicAgentImageByRelease(ctx, log, infraEnvInternal)
+		if err != nil {
+			log.WithError(err).Warningf("Failed to get ironic agent image by release for infraEnv: %s", infraEnv.Name)
+		}
+	} else {
+		log.Infof("Using override ironic agent image (%s) for infraEnv %s", ironicAgentImage, infraEnv.Name)
 	}
 
 	// if ironicAgentImage can't be found by version use the default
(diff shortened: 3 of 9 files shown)


internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -367,20 +371,21 @@ func (r *PreprovisioningImageReconciler) mapInfraEnvPPI() func(a client.Object)
...
 	infraEnvInternal, err := r.Installer.GetInfraEnvByKubeKey(key)
 	if err != nil {
 		log.WithError(err).Error("failed to get corresponding infraEnv")
-		return ctrl.Result{}, err
+		return false, err
 	}
-	var ironicAgentImage string
-	if infraEnvInternal.ClusterID != "" {
-		ironicAgentImage, err = r.getIronicAgentImageByRelease(ctx, log, *infraEnvInternal)
+	ironicAgentImage := infraEnv.Spec.IronicAgentImageOverride
```
eranco74, line 386: Add logging If there was an override.

eranco74 (commented):
It's doing more than what the PR title is saying.
Looks good to me

carbonin:
>It's doing more than what the PR title is saying.

Built on top of https://github.com/openshift/assisted-service/pull/4798 (WIP until hub image PR is merged)

carbonin:
@eranco74 this one is ready for review

carbonin:
I don't see why this PR would break the capi job, retesting.

/test edge-e2e-ai-operator-ztp-capi

docs/hive-integration/README.md
```diff
@@ -214,6 +214,9 @@ The environment var for the ironicAgent image to be used on X86_64 CPU architect
 The environment var for the ironicAgent image to be used on arm64 CPU architecture:
 `IRONIC_AGENT_IMAGE_ARM`
 
+The ironic agent image can also be overridden using the InfraEnv spec field `ironicAgentImageOverride`
+If this field is set this image will be used instead of the hub or defaul image.
```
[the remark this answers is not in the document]
eranco74: Though I think that after this change we can just remove the default image, and if someone wants to use the converged flow with late-binding they will need to set the ironicAgentImageOverride for that to work.

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -356,35 +356,38 @@ func (r *PreprovisioningImageReconciler) mapInfraEnvPPI() func(a client.Object)
...
-			// Don't queue if the Image URL and InfraEnv's URL is the same.
-			if images[i].Status.ImageUrl != infraEnv.Status.ISODownloadURL {
-				reconcileRequests = append(reconcileRequests, reconcile.Request{
-					NamespacedName: types.NamespacedName{
-						Namespace: images[i].Namespace,
-						Name:      images[i].Name,
-					},
-				})
-			}
+			reconcileRequests = append(reconcileRequests, reconcile.Request{
```
eranco74, line 359: This seems like an overkill and I think we can optimize it, not sure if it's worth the trouble though.
(we care about the imageURL, irnoicAgentImageOverride and the internalIgnitionOverride everything else is just noise).

eranco74 (commented):
/lgtm

docs/hive-integration/README.md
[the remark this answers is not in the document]
carbonin: But yeah, I forgot we talked about removing the defaults with this. I'll do that

internal/controller/controllers/preprovisioningimage_controller.go
carbonin: But we don't know the current state of those overrides. That's why we have to render the entire ignition to compare in the reconcile step.

It was easy with the url because we could compare them directly, but it's not so simple now.

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -400,25 +403,29 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
...
+		if infraEnv.ObjectMeta.Annotations == nil {
+			infraEnv.ObjectMeta.Annotations = make(map[string]string)
+		}
+		infraEnv.Annotations[EnableIronicAgentAnnotation] = "true"
+		if err := r.Client.Update(ctx, infraEnv); err != nil {
+			// Just warn here if the update fails as this annotation is just informational
+			log.WithError(err).Warnf("failed to set %s annotation on infraEnv %s", EnableIronicAgentAnnotation, infraEnv.Name)
+		}
+
+		r.CRDEventsHandler.NotifyInfraEnvUpdates(infraEnv.Name, infraEnv.Namespace)
```
filanov, line 426: doesn't adding annotation trigger reconcile?

carbonin: Maybe but for one, we may already have this annotation so the update may be a noop. This was also there before, just the indentation changed.

I can leave the comment too, if you want :man_shrugging:

filanov: The code is confusing because there is not check that the annotation is actually set or not, it may already exist. 
So the notification is actually depend on the `updated` field

carbonin:
/unhold

I'm going to remove the defaults in a follow-up since this is pretty much ready.

api/v1beta1/infraenv_types.go
```diff
@@ -103,6 +103,11 @@ type InfraEnvSpec struct {
 	// Applicable for both iPXE, and ISO streaming from Image Service.
 	// +optional
 	KernelArguments []KernelArgument `json:"kernelArguments,omitempty"`
+
+	// IronicAgentImageOverride is the pull spec for the ironic agent image to use when discovering hosts
+	// Only applies when provisioning using a BareMetalHost
+	// +optional
+	IronicAgentImageOverride string `json:"ironicAgentImageOverride,omitempty"`
```
avishayt, line 110: Do we want this directly in the spec, or would an annotation be more appropriate?

mhrivnak:
Can we get a bit more detail on the use case? I see in the issue:

> We might need it in case our default image will not match the infraEnv paltform.

What exactly might not "match the platform"?

How will a user know what value is in use currently, and how will they know what other value to put in this field and why?

How is this managed over time? Does the user have to update this as new releases come out? Are we giving any guidance on that?

eranco74:
> Can we get a bit more detail on the use case? I see in the issue:
> 
> > We might need it in case our default image will not match the infraEnv paltform.
> 
> What exactly might not "match the platform"?
Platform in this regard is the CPU architecture,  the nodes architecture might not match the hub nodes architecture and the Ironic agent image from the release might not work.

> How will a user know what value is in use currently, and how will they know what other value to put in this field and why?
The infraEnv spec should reflect that (once we remove the default images)
And what value the user should use in case they need to should be documented. 
> How is this managed over time? Does the user have to update this as new releases come out? Are we giving any guidance on that?
In essence it should just work seamlessly, the Ironic agent should be backward compatible in the functionalitywe require (and thats why we started with default images). so even when the user provided an override (which shouldn't be the common case) no action should be required.

This doc covers the reasons behind this PR
https://docs.google.com/document/d/1inBEXtQnJ1wZdxDdKxV4MoqeZUL28Ltb40vt223BcdI/edit?usp=drivesdk

carbonin:
Updated the JIRA with more details about the reasoning for this change and the use cases.

I'll copy the user stories I wrote here to save a click:
1. As a user I want to install a spoke cluster with a different CPU architecture than is supported by my hub cluster release image.
2. As a user I want to consistently mirror a single ironic agent image to my disconnected environments regardless of the version of my hub cluster.
3. As an assisted installer developer I want to easily override the ironic agent used during discovery for testing or debugging.

These use cases could be achieved by using deployment configuration (configure an image per architecture), but if we want this to be a supported, user-facing API we should not require them to configure the "unsupported" configmap to alter the assisted-service's behavior.

carbonin:
Any other thoughts on this?

@mhrivnak @avishayt ?

api/v1beta1/infraenv_types.go
mhrivnak: I made the case elsewhere for putting it in the Spec, and I think @avishayt agreed.

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -356,35 +356,38 @@ func (r *PreprovisioningImageReconciler) mapInfraEnvPPI() func(a client.Object)
...
+		return false, err
 	}
-	ironicAgentImage, err := r.getIronicAgentImageByRelease(ctx, log, infraEnvInternal)
-	if err != nil {
-		log.WithError(err).Warningf("Failed to get ironic agent image by release for infraEnv: %s", infraEnv.Name)
+	ironicAgentImage := infraEnv.Spec.IronicAgentImageOverride
+	if ironicAgentImage == "" {
+		ironicAgentImage, err = r.getIronicAgentImageByRelease(ctx, log, infraEnvInternal)
+		if err != nil {
+			log.WithError(err).Warningf("Failed to get ironic agent image by release for infraEnv: %s", infraEnv.Name)
```
mhrivnak, line 387: Should this return with the error here? I'm not sure why it would make sense for the function to keep going.

api/v1beta1/infraenv_types.go
```diff
@@ -103,6 +103,11 @@ type InfraEnvSpec struct {
 	// Applicable for both iPXE, and ISO streaming from Image Service.
 	// +optional
 	KernelArguments []KernelArgument `json:"kernelArguments,omitempty"`
+
+	// IronicAgentImageOverride is the pull spec for the ironic agent image to use when discovering hosts
```
mhrivnak, line 107: It sounds like the core use case for end users is when they want to install a spoke using BareMetalHosts for hardware that's a different CPU architecture than the hub. I think that, or some similar guidance and context, is worth documenting here. When someone reads the API docs, they should understand when/why this feature is useful and applicable.

mhrivnak (changes requested):

api/v1beta1/infraenv_types.go
mhrivnak: Re-reading, maybe I spoke too soon. :) My basic assertion is that putting it in the spec:
* lets us clearly document what the feature is and how/when to use it
* helps users discover that the feature exists
* helps users notice when the feature is in use (people look at the spec on existing resources to see input that affects an API's behavior; it's easy to miss input that's provided elsewhere, such as an annotation)
* makes it clear how the field will evolve in the future. If it get deprecated, replaced, or changes in other ways, having it in the Spec gives us a clear process for managing that change.
* lets us use k8s validation features
* provides an opportunity to put corresponding info in the Status

internal/controller/controllers/preprovisioningimage_controller.go
carbonin: There's a default value that we use as a fallback in the next block.

Here the logic goes:
- If there's an override use it
- No override, try to get it from the hub release image
- Hub release failed and no override, use hardcoded/env driven default

We're going to remove the last one in a follow-up, but since it will require us to communicate to the user that we can't find an image in some way it will require a bit more effort.

mhrivnak: Is there a code path here where the user provides a value, an error happens in `getIronicAgentImageByRelease`, and then the code silently ignores the user-supplied value and uses the default instead?

If not, you can resolve this thread.

carbonin: If the user provides an override we don't call `getIronicImageByRelease` so no, that's not something that can happen.

api/v1beta1/infraenv_types.go
avishayt: We know have a CpuArchitecture property in the InfraEnv - could we use that to infer the correct image?

carbonin: We [currently](https://github.com/openshift/assisted-service/blob/8fbbce6515dc38ad028e5a47417c601a0a0c5495/internal/controller/controllers/preprovisioningimage_controller.go#L446-L448) use that to know when the hub image is incorrect for a particular infraenv, but if the hub is using a single-arch release image then there's no way for us to determine what the pull spec is for some other architecture ironic agent.

mhrivnak: That makes me think again about what we expect the user to do.

They could run the following. Is it what we'd recommend?

```
$ oc adm release info quay.io/openshift-release-dev/ocp-release:4.12.0-multi --filter-by-os=linux/arm64 --image-for=ironic-agent
W0117 09:51:02.049979   80777 manifest.go:449] Chose linux/arm64 manifest from the manifest list.
quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:69c5e28d0cbe374a6a663f85420e33c50f363c5084d3c36e82cfc9099f6a41b1
```

That leads to the question: can we just do that automatically if connected? And if disconnected, could we pick a reasonable release from those that are available? (choosing from the mirrored releases in the hive CRD whose name is escaping me at the moment)

In any case, it seems like we need to give the user fairly specific guidance on how to figure out what value to put here. If we can give them specific guidance on what to do, maybe instead we can just do it?

carbonin: Sure, but the idea was to make this easy to manage for everyone involved, us included.

For something like 95% of cases the spoke cluster arch is going to match the hub and this value won't be used. When the hub release image is multi arch, devs will probably be the only ones using it.

This way we'll get the exact image we need from the hub release image, our documentation will say "mirror the hub release image if you're disconnected" and everything will just work without us having to constantly maintain some list of images for every possible hub version.

carbonin:
Closing this as it's low priority and still needs us to decide on an API.

PR #4855, opened 2023-01-03
MGMT-12655: Propagate installation disk from Host to Agent

mkowalski:
This PR introduces a backwards synchronization of the installation disk ID from the Host (AI DB) to the Agent (K8s) in a scenario when no device hints are provided.

In the non-ZTP scenarios that use ACM it allows to query for the installation disk selected by the backend directly from the Agent CR. Until now Agent CR carried installation disk ID only in case there was an override, but with this change it should always contain the ID no matter whether it has been selected by the user explicitly or autocalculated by the backend.

Contributes-to: [MGMT-12655](https://issues.redhat.com//browse/MGMT-12655)

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -582,6 +582,7 @@ func main() {
 				APIReader:             ctrlMgr.GetAPIReader(),
 				Log:                   log,
 				Scheme:                ctrlMgr.GetScheme(),
+				Installer:             bm,
 				SpokeK8sClientFactory: spoke_k8s_client.NewSpokeK8sClientFactory(log),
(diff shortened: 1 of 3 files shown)


filanov:
I'm suggesting to add another field in the status that will show the selected installation disk, in this case when the user is not selecting anything we will show our selection in the status, when the user will select a disk in the spec we will show the same value in the status. 
Editing the spec with the controller is not a common pattern

mkowalski:
/close

Superseded by https://github.com/openshift/assisted-service/pull/4865

PR #4856, opened 2023-01-03
MGMT-12655: Set Agent's installation disk ID only if hints exist

mkowalski:
This PR changes the logic of setting installation disk ID in the Agent CR so that the backend modifies CR only if user has provided a custom root device hints. This prevents a scenario when empty hints cause setting the Agent's field to the empty value even if the value has been set explicitly by the user.

With this change the behaviour will be as follows

* empty ID in Agent and empty hints - nothing happens
* empty ID in Agent and non-empty hints - autocalculation
* non-empty ID in Agent and empty hints - ID is preserved
* non-empty ID in Agent and non-empty hints - autocalculation

Contributes-to: [MGMT-12655](https://issues.redhat.com//browse/MGMT-12655)


paul-maidment (approved):
/lgtm

PR #4867, opened 2023-01-05
MGMT-10869: Forbid multiple machine networks in single-stack clusters

mkowalski:
This PR makes the network config validation stricter so that single-stack clusters with multiple machine networks are not allowed.

As long as OCP does not allow for remote control plane, such a topology is not correct and AI should prevent it from being used.

Closes: [MGMT-10869](https://issues.redhat.com//browse/MGMT-10869)


mkowalski:
/hold

First https://github.com/openshift/assisted-service/pull/4856, then rebase

mkowalski:
/retest-required

quay flake

paul-maidment (approved):
These changes look good. I can see that as expected, we reject multiple machine networks for single stack clusters. /lgtm

PR #4870, opened 2023-01-08
MGMT-12547: Enhancement doc "api-for-skipping-validations"

paul-maidment:
Adding documentation to propose a new feature "API for skipping validations"

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,139 @@
...
+}
+```
+
+* Any entries found will be added to the existing list of validations to be skipped. This will appropriately de-deuplicated and no error will be returned if duplication is attempted.
+* If an attempt is made to reference a validation that does not exist then a warning will be logged but this will not be considered fatal.
+* After the call has completed, if this leads to a change in the collection of skipped validations then an event will be emitted.
+
+### Delete some skipped validations.
+
+DELETE api/v2/skippedValidations
```
ori-amizur, line 75: I am not sure about delete with payload.  Does not seem _restful_ 
Maybe this will be better ?
```
DELETE api/v2/skipped-cluster-validations/api-vips-defined
```

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,139 @@
...
+  "skipped-cluster-validations": [
+    "network-prefix-valid",
+    "api-vips-defined"
+  ]
+}
+```
+
+### Add some skipped validations.
+
+POST api/v2/skippedValidations
```
ori-amizur, line 55: Would it be simpler to add single validation to skip in the operation?

oourfali:
high level comment - not sure whether skipping specific validations(s) is considered hard implementation-wise, but I'd argue that in most (if not all) cases it would be enough to just skip all validations.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,139 @@
...
+# Summary
+
+We need new api that will allow us to skip cluster/host validations.
+This api should have it's own feature flag to allow us to disable it.
+
+It should be possible to use this API in an unsupported way, it is currently undecided whether or not this will be open or not.
+An event should be emitted whenever the API is used during an installation, this can be used to drive signatures etc.
+# Motivation
+
+Some customer and partners have very specific HW that doesn't pass our validations and we want to allow them to proceed with installation.
```
ori-amizur, line 20: Are there mandatory validations that cannot be skipped?

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,139 @@
...
+Sometimes we have bugs in our validations that block people from installing and we don't want our partners left in a position where they can't install.
+
+### Goals
+
+- Allow validations to be bypassed so that installation may proceed.
+
+### Non-Goals
+
+- We should not add uneccesary noise to triage tickets with this, it is therefore quite important to add some signatures to indicate that these validations have been bypassed.
```
osherdp, line 29: that's not part of "non-goals", as it requires more work to be done (adding signatures, feature-usage labels maybe?)
AFAIK non-goals are stuff that are not in the scope of the enhancement

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Proposal
+
+I think the API should be kept reasonably simple.
+
+### List existing validations.
+
+GET api/v2/skippedValidations
```
osherdp, line 37: just to stay with current naming conventions: 
```suggestion
GET api/v2/skipped-validations
```

docs/enhancements/api-for-skipping-validations.md
paul-maidment: Good question, from what I know of the motivations, it would seem that this feature is for the purpose of bypassing issues with specific validations. I question whether or not we should be skipping anything that is not known to be problematic by the user.

This would go against the idea of skipping all validations and instead favour skipping specific validations.

docs/enhancements/api-for-skipping-validations.md
paul-maidment: I personally think this would be fine if this request is "additive", i.e. if you call it again with a different validation id then this will additionally be skipped, how does this sound?

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Goals
+
+- Allow validations to be bypassed so that installation may proceed.
+
+### Non-Goals
+
+- We should not add uneccesary noise to triage tickets with this, it is therefore quite important to add some signatures to indicate that these validations have been bypassed.
+
+## Proposal
```
osherdp, line 31: I would add the "skipped" status to ``validations_info``. For example:
```
{
"id": "api-vips-defined",
"status": "skipped",
"message": "API virtual IPs validation has been suppressed by the user"
},
```

docs/enhancements/api-for-skipping-validations.md
paul-maidment: Agreed, let's keep this one simple.

paul-maidment:
> high level comment - not sure whether skipping specific validations(s) is considered hard implementation-wise, but I'd argue that in most (if not all) cases it would be enough to just skip all validations.

Personally I am not sure, there is another question pertaining to whether or not certain validations should be "hard mandatory" and impossible to skip. Whereas I don't think it's necessary to go this far, it does shift thought towards the intent of this feature.

It seems that the use case is to bypass specific validation problems, it would therefore seem logical to allow the skipping of specific validations while allowing others to run. This would allow the other validations to "sense check" the installation otherwise. 

Seems more rational to me but obviously a little less simple to implement.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,139 @@
...
+creation-date: 2023-01-08
+last-updated: 2023-01-08
+---
+
+# Allow validations to be skipped
+
+# Summary
+
+We need new api that will allow us to skip cluster/host validations.
+This api should have it's own feature flag to allow us to disable it.
```
osherdp, line 14: *its
in addition to a feature-flag, maybe we'll want to have the same ability as multi-arch images of specifying what domains can do this skipping action?
an advantage is having this thing scoped only for whatever company needs it. the disadvantage - showing an API that is blocked for certain users is problematic. wdyt?

docs/enhancements/api-for-skipping-validations.md
osherdp, line 55: I guess it should be under cluster resource URL?
In this case, would we want to only have this skipping mechanism under the cluster resource? The problem with host validations is that we would prefer to add a skip for all hosts in a cluster, instead of adding it for each host specifically

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,139 @@
...
+}
+```
+* Any entries found will be deleted from the existing list of validations to be skipped.
+* If an attempt is made to reference a validation that does not exist then a warning will be logged but this will not be considered fatal.
+* If a validation has not been skipped already and is marked for removal then this entry will be ignored and not emit an error.
+* After the call has completed, if this leads to a change in the collection of skipped validations then an event will be emitted.
+
+### Delete all skipped validations
+
+DELETE api/v2/skippedValidations/all
```
osherdp, line 95: DELETE ``.../all`` is not RESTful, maybe instead deleting the whole ``skippedValidations`` resource and treating it with ``PATCH/PUT`` requests instead of the ``POST`` requests

osherdp (changes requested):

docs/enhancements/api-for-skipping-validations.md
paul-maidment: That's fair actually, I'll take this out of non-goals.

docs/enhancements/api-for-skipping-validations.md
paul-maidment: One of the motivations for this feature is (as worded in the original ticket) "Sometimes we have bugs in our validations that block people from installing and we don't want our partners to stuck cause of us"

It strikes me that we are likely to advise a partner to skip a validation when we are aware of an issue and we are also aware that a workaround cannot be immediately implemented. I would imagine that this would require more flexibility. Locking this feature to specific users might cause issues with this.

Though, adding the _ability_ to lock this feature to a specific set of beta testers might not be a bad idea?

docs/enhancements/api-for-skipping-validations.md
paul-maidment: Agreed actually, but ideally we should be able to skip any validation, cluster or host, via this mechanism.
If you specify a host validation to skip, it should be skipped for all hosts.
Is this reasonable?

docs/enhancements/api-for-skipping-validations.md
paul-maidment: Accpted point about the delete operation.

Worth considering a bit of both.

If we reserve the  POST request for the _addition_ of another "skipped validation" and keep the patch for clearing or bulk adding validations, this might be best. 

The advantage of having the additional POST operation as described is that it is very easy to skip "one more of.." the validations without having to edit JSON etc.

docs/enhancements/api-for-skipping-validations.md
paul-maidment: I agree with this point

docs/enhancements/api-for-skipping-validations.md
paul-maidment: Agreed

docs/enhancements/api-for-skipping-validations.md
osherdp: Yes it's also what I would prefer
So the list includes both cluster and host validations, that are skipped for the scope of the whole cluster object. Correct?

paul-maidment: That's correct.

docs/enhancements/api-for-skipping-validations.md
osherdp: I'm divided on this, because as you say this mechanism removes some of the flexibility here (which is key in this whole feature)
I think all-in-all it's not worth locking to users. It requires more work to be implemented, complicates the system, and removes the flexibility. I'm withdrawing my idea :sweat_smile:

tsorya: Actually maybe we need to open it for certain partners and us, at least for now, like nvidia, marvel ?
We should allow it though in all onprem solutions

osherdp: It's possible. But keep in mind that approving a new partner in SaaS prod requires a MR approval from an outside team (meaning we don't have full control over it)

docs/enhancements/api-for-skipping-validations.md
tsorya: Not sure regarding this, i would not start allowing to do override in UI and if someone is doing it on purpose, it's sort of his problem. Lets set relevant message, flag in jira, event and skip those failures from our statistics

docs/enhancements/api-for-skipping-validations.md
tsorya: Yep, sounds good

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,144 @@
...
+
+As a Kaloom user, I would like to be able to bypass VIP validation in cases where VIPS could be incorrectly shown as busy. So that I may proceed with installation.
+
+"Kaloom has very specific setup where vips can be shown as busy though installation can proceed with them.
+Currently they need to override vips in install config to be able to install cluster.
+After adding the new api they can just run it and skip this specific validation."
+
+### Risks and Mitigations
+
+- We should not add uneccesary noise to triage tickets with this, it is therefore quite important to add some signatures to indicate that these validations have been bypassed.
```
tsorya, line 117: We should skip those tickets same way we do for jupiter installations, just no need to see them at all, i think

paul-maidment: That would be nice, do you have any information on how that skipping is set up?

osherdp: https://github.com/openshift-assisted/assisted-installer-deployment/blob/463dc6612c923c46d47e80762c72651758a690d3/tools/create_triage_tickets.py#L64

paul-maidment: Thank you

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,148 @@
...
+  - "@pmaidmen"
+creation-date: 2023-01-08
+last-updated: 2023-01-08
+---
+
+# Allow validations to be skipped
+
+# Summary
+
+We need new api that will allow us to skip cluster/host validations.
```
gamli75, line 13: do we want to skip it for a specific cluster or for all clusters? if it's for a specific cluster - we should add the cluster id to the POST API.
if it's for all clusters - we have something similar today - https://github.com/openshift/assisted-service/blob/master/openshift/template.yaml#L93 - let's mention it in this enhancement

docs/enhancements/api-for-skipping-validations.md
filanov: if we have set for specific orgs then it will be enabled automatically for on prem no?

gamli75:
Few remarks:
1. we have something similar today - https://github.com/openshift/assisted-service/blob/master/openshift/template.yaml#L93 - let's mention it and write that what we have today has a higher priority.
2. we have 2 skipped option - by the user(this API) or by the system(option 1)
3. please add a section about seeing this API usage in elastic - not sure if it should be in this enhancement doc or in the epic

docs/enhancements/api-for-skipping-validations.md
paul-maidment: OK, we will add "per customer feature enablement" as per https://github.com/openshift/assisted-service/blob/master/docs/dev/feature-per-customer.md

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,155 @@
...
+- We should not add uneccesary noise to triage tickets with this, it is therefore quite important to add some signatures to indicate that these validations have been bypassed.
+
+### Open Questions
+
+N/A
+
+### UI Impact
+
+As this change is API based, I am not 100% sure we would need to do too much in the UI at this stage. 
+Perhaps listing any blocked validations would be helpful?
```
osherdp, line 137: if "validations_info" has a new value "skipped", then UI might need to prepare for it (currently it expects either success or failure)

paul-maidment: I wonder, if it might be better to just log the validations that have been skipped instead of adding a field because

This feature is unsupported and generally we should steer away from UI indications.
Providing some indication to the end user that "some validations were skipped therefore we cannot support this installation" might be an idea, but I question how granular we have to be in this messaging

What are your thoughts?

osherdp: it's not adding a new field, it's adding a new possible value to an existing field
the problem with log lines, is that it makes it harder to understand the situation. think of a qe / dev / user that tries to understand the "history" of a validation. if it's marked as skipped it's more apparent than a log line that is not easily viewed (or not at all in some cases)

paul-maidment: OK, so seems like we need to add a UI task to support the new value. I'll take it into account.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,153 @@
...
+  "skipped-host-validations": [
+    "no-ip-collisions-in-network"
+  ],
+  "skipped-cluster-validations": [
+    "api-vips-defined"
+  ]
+}
+```
+
+* Any entries found will be added to the existing list of validations to be skipped. This will appropriately de-deuplicated and no error will be returned if duplication is attempted.
```
mhrivnak, line 67: It would be simpler if we can avoid a custom merging strategy in our API that requires reading documentation to understand.

What about using standard REST more deeply?

```
PUT api/v2/cluster/<id>/skipped-host-validations/no-ip-collisions-in-network

GET api/v2/cluster/<id>/skipped-host-validations/

[
    "no-ip-collisions-in-network"
]

DELETE api/v2/cluster/<id>/skipped-host-validations/no-ip-collisions-in-network
```

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,153 @@
...
+
+See the following link for more information.
+
+https://github.com/openshift/assisted-service/blob/master/docs/dev/feature-per-customer.md
+
+### User Stories
+
+#### Story 1
+
+As a Kaloom user, I would like to be able to bypass VIP validation in cases where VIPS could be incorrectly shown as busy. So that I may proceed with installation.
```
mhrivnak, line 118: It's probably best to keep specific company names out of user stories.

mhrivnak (changes requested):

mhrivnak:
> > high level comment - not sure whether skipping specific validations(s) is considered hard implementation-wise, but I'd argue that in most (if not all) cases it would be enough to just skip all validations.
> 
> Personally I am not sure, there is another question pertaining to whether or not certain validations should be "hard mandatory" and impossible to skip. Whereas I don't think it's necessary to go this far, it does shift thought towards the intent of this feature.
> 
> It seems that the use case is to bypass specific validation problems, it would therefore seem logical to allow the skipping of specific validations while allowing others to run. This would allow the other validations to "sense check" the installation otherwise.
> 
> Seems more rational to me but obviously a little less simple to implement.

I like this idea and think it's worth capturing at least as an alternative in the doc. It may even be worth implementing.

As I am interpreting Oved's idea, the experience would be:
* All validations run.
* One or more validations fail and describe a problem.
* The user evaluates the alleged problems and decides to try the installation anyway.

Advantages:
* The user will be faced with complete information about the validation failure before choosing to proceed.
* In cases where a validation can fail for multiple reasons, or perhaps with varying degrees of severity (for example the latency is barely out of spec vs WAY out of spec), there's less risk that a user will miss an important new validation failure because the entire test was skipped. This could be like disabling a smoke detector that's too sensitive when you cook, but then it doesn't alarm when the house is on fire.

Disadvantages:
* This would not help if there's just a bug in a validation, especially if that bug causes the whole install process to stall or fail.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,153 @@
...
+last-updated: 2023-01-08
+---
+
+# Allow validations to be skipped
+
+# Summary
+
+We need a new api that will allow us to skip cluster/host validations, this will be for unsupported use.
+This API should work on a per cluster basis and will have it's own feature flag to allow us to enable it for specific users.
+An event should be emitted whenever the API is used during an installation, this can be used to drive signatures etc.
```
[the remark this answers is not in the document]
paul-maidment: I meant that we can write a signature to look for these events and that this could be used on a Triage ticket to indicate that the user had overridden validations in this particular case.

I may remove this line completely from the document as there is feedback later that suggests that we should not even create triage tickets for clusters that have emitted such events.

docs/enhancements/api-for-skipping-validations.md
paul-maidment: I didn't mean to suggest that anything should be added to the UI to allow this. Simply that when a cluster has overridden validations that we may want to display something to reflect this.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Non-Goals
+
+- None identified at this stage
+
+## Proposal
+
+I think the API should be kept reasonably simple.
+
+### List existing validations.
```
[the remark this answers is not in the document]
paul-maidment: Agreed, it seems reasonable to discuss the idea of adding some CR's for the purpose of skipping validations.

docs/enhancements/api-for-skipping-validations.md
paul-maidment: I like this, a good suggestion I think

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,153 @@
...
+
+* Any entries found will be added to the existing list of validations to be skipped. This will appropriately de-deuplicated and no error will be returned if duplication is attempted.
+* If an attempt is made to reference a validation that does not exist then a warning will be logged but this will not be considered fatal.
+* After the call has completed, if this leads to a change in the collection of skipped validations then an event will be emitted.
+
+### Edit the validations
+
+PUT api/v2/cluster/<id>/skipped-validations
+
+Essentially a document will be submitted via POST in JSON
```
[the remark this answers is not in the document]
paul-maidment: Yes, I'll update

docs/enhancements/api-for-skipping-validations.md
paul-maidment: Agreed, will remove

docs/enhancements/api-for-skipping-validations.md
paul-maidment: Moreover, to clarify further, the only means by which a validation may be skipped is via this limited API. As agreed upon elsewhere, this will only occur for specific customers who will be aware of the consequences of using the feature. (namely that the cluster is unsupported.)

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Non-Goals
+
+- None identified at this stage
+
+## Proposal
+
+There are three main components as I see it
+
+### The REST API
```
avishayt, line 34: This seems very complex, both in the API and in the DB.  Why not just:
PUT api/v2/cluster/<id>/skip-host-validations (body has true/false)
GET api/v2/cluster/<id>/skip-host-validations (returns true/false)

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,220 @@
...
+
+* After the call has completed, if this leads to a change in the collection of skipped validations then an event will be emitted.
+
+#### Endpoint - Delete all skipped validations for cluster.
+DELETE api/v2/cluster/<id>/skipped-cluster-validations
+
+* After the call has completed, if this leads to a change in the collection of skipped validations then an event will be emitted.
+
+
+### The Kubernetes API
```
avishayt, line 87: Again, seems overly complex - we can just add an annotation to skip all validations

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,220 @@
...
+
+* If you want to delete all skipped validations, simply delete this CR and await reconciliation.
+* To add or remove a skipped validation, simply edit the CR and wait reconciliation.
+* Perhaps we would stop reconciling this CR once the installation starts as this would make sense.
+
+### The implementation of skipped-validations
+
+There are two ways that are currently proposed.
+
+#### Option 1: Run all validations, allow the user to review but proceed anyway. (Suggested by Michael Hrivnak as an intepretation of requirements)
```
avishayt, line 123: This is how I imagined it would work as well.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Risks and Mitigations
+
+- We should not add uneccesary noise to triage tickets with this, it is therefore quite important to add some signatures to indicate that these validations have been bypassed or to handle in a way that ensures they never appear as Triage tickets.
+
+### Open Questions
+
+N/A
+
+### UI Impact
```
avishayt, line 188: If we go with "Option 1: Run all validations, allow the user to review but proceed anyway." then there will be no "skipped" value and no UI impact, right?

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,220 @@
...
+It should be noted that it should not be made possible to skip validations by using the UI itself.
+
+### Test Plan
+
+A number of automated tests will be written as unit tests.
+Subsystem tests as neccessary.
+I think also a manual check of the feature similar to the following.
+
+1: Create an SNO cluster bring it to a ready state and deliberately set one of the hosts back to "localhost" to trigger that validation.
+2: Call the REST API to skip the validation and witness that the validation passes.
```
avishayt, line 210: I would just do this in a subsystem test.

docs/enhancements/api-for-skipping-validations.md
oourfali: same here.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,220 @@
...
+```
+{
+    "id": "api-vips-defined",
+    "status": "skipped",
+    "message": "API virtual IPs validation has been suppressed by the user"
+}
+```
+
+We should ensure that the UI accounts for this and notifies the user of any validations that have been skipped.
+It should be noted that it should not be made possible to skip validations by using the UI itself.
```
oourfali, line 201: When we discussed this RFE we mentioned that the ability to skip validations is an API-only ability. The UI shouldn't care about that at all.

avishayt: Right, the UI just shouldn't break.  That's what is being said here.

docs/enhancements/api-for-skipping-validations.md
filanov: It will make the api simple but it will be a bit harder for us to understand the reason behind the failure. 
We will be able to to probably look at the events and understand from it what happened. 
but on the other side if they disable all the validations so we will loos basic validations that can actually 100 chance to fail.

I would not expect this api commonly use and it should be used only in extreme cases. 
So we can probably start simple and see how it goes, we can always add parameters to the api to improve it.

docs/enhancements/api-for-skipping-validations.md
filanov: I don't think that we need something like that for kube-api in the first place. we have the unsupported way to disable validation on service level, i don't see a reason to have it in a cluster level. 
I don't think that ACM will have multiple different clients that will need to skip validations for a specific cluster.
At least this is my take on that.

docs/enhancements/api-for-skipping-validations.md
paul-maidment: That's very true

docs/enhancements/api-for-skipping-validations.md
paul-maidment: Precisely

docs/enhancements/api-for-skipping-validations.md
paul-maidment: So to clarify the proposal is to allow the skipping of host validations with a single setting and if this flag is set, we simply will not run any validations?

We can then put an item into a log to indicate we have done this, raise and event and then use the log entry / event to prevent the creation of triage tickets etc.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,220 @@
...
+{
+    "network-prefix-valid",
+    "api-vips-defined"
+}
+```
+
+#### Endpoint - Add a skipped validation.
+
+PUT api/v2/cluster/<id>/skipped-host-validations/no-ip-collisions-in-network
+PUT api/v2/cluster/<id>/skipped-cluster-validations/api-vips-defined
```
danielerez, line 61: How do we plan the expose the list of available validations to users? I mean maybe we need also an API for fetching the 'host-validation-id' [enum](https://github.com/openshift/assisted-service/blob/58b7b4b0fd8e5a43c4b27ab28086c896fa636f27/swagger.yaml#L5836).

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,89 @@
...
+
+See the following link for more information.
+
+https://github.com/openshift/assisted-service/blob/master/docs/dev/feature-per-customer.md
+
+### User Stories
+
+#### Story 1
+
+As a user with known validation issues that need to be bypassed, I would like to be able to bypass VIP validation in cases where VIPS could be incorrectly shown as busy. So that I may proceed with installation.
```
avishayt, line 63:
```suggestion
As a user with known validation issues that need to be bypassed, I would like to be able to bypass VIP validation in cases where VIPs are incorrectly shown as busy so that I may proceed with the installation.
```

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,89 @@
...
+The way we propose to deal with the validation is as follows
+
+#### Propoal - Run all validations, allow the user to review but proceed anyway. (Suggested by Michael Hrivnak as an intepretation of requirements)
+
+    * All validations run.
+    * One or more validations fail and describe a problem.
+    * The user evaluates the alleged problems and decides to try the installation anyway.
+    * When skipping validations, there will be some "absolutely mandatory" validations that cannot fail prior to proceeding.
+        - This list is TBC
+    * If the user skips validations (by making a change to a boolean in the cluster object), we should
```
avishayt, line 41: Instead of a boolean, let's have a string with a list of validations to skip, and maybe a special "all" value.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Drawbacks
+
+* We will have to be careful when examining bug reports and triage tickets to ensure we are aware of when this feature has been used.
+* Might allow some customers to bypass things they shouldn't bypass.
+
+
+
+
+
```
avishayt, line 89:
```suggestion
```

docs/enhancements/api-for-skipping-validations.md
paul-maidment: This seems at odds with this recommendation. 

"Again, seems overly complex - we can just add an annotation to skip all validations"

and 

"This seems very complex, both in the API and in the DB. Why not just:
PUT api/v2/cluster//skip-host-validations (body has true/false)
GET api/v2/cluster//skip-host-validations (returns true/false)"

Both of which seem to imply that we want a simple on/off switch for ALL validations, I think I may have misunderstood?

Can you clarify further  on what is actually needed here?

oourfali: I think that there are different opinions here also with regards to the requirements.
In my view, we should:
1. Allow to ignore all validations
2. API only - no impact whatsoever to the UI
3. We can have this gates by AMS capabilities - so it will be used only for whom we allow it (in my view we can also just open it to everyone, but if someone is concerned people will abuse it, we can allow it only to specific organizations)
4. We should put a flag on clusters that the user installed while ignoring validations.

avishayt: I was previously put off by the number of APIs to manage the list of validations to skip (there were operations to add, remove, list, ...).  However, boolean APIs are fragile because you need to break the API to add a third option.  So now I am suggesting a compromise, where we still only have one PUT and one GET API, but instead of having a boolean in the body, we have a list of strings.

I agree with Oved's requirements above, just in addition to allowing us to ignore all validations, we can enable ignoring some validations as well.

paul-maidment: OK, agreed that we will allow a string input with a list of strings for a single PUT and a single GET API, I'll update the document now.

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,220 @@
...
+apiVersion: agent-install.openshift.io/v1beta1
+kind: SkippedValidations
+metadata:
+  name: spoke-1-skipped-validations
+  namespace: spoke-1
+spec:
+  clusterRef:
+   name: spoke-1
+   namespace: spoke-1
+  skipped-host-validations:
```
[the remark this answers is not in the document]
paul-maidment: Probably overkill for a first draft of the feature but an interesting suggestion nontheless

docs/enhancements/api-for-skipping-validations.md
paul-maidment: Good question

docs/enhancements/api-for-skipping-validations.md
avishayt: That validation is only looking at the value in the DB - the actual measurement happens as part of the preparing-to-install phase.  Are you suggesting that we skip the actual measurement?  That's a very different behavior.  Have we seen other cases where the actual measurement was problematic?  I know that the connectivity check failed in a few cases.

danielerez: Yeah, I'm referring to the actual measurement that could take a while. I've seen an issue with that once where it timeouts after 10mins (using a VM though).

docs/enhancements/api-for-skipping-validations.md
```diff
@@ -0,0 +1,117 @@
...
+Will return a list of string 
+
+```
+{
+  "has-min-cpu-cores",
+  "has-min-memory"
+}
+```
+
+### Proposal - Run all validations, allow the user to review but proceed anyway. (Suggested by Michael Hrivnak as an intepretation of requirements)
```
avishayt, line 59: Please add "feature usage".

docs/enhancements/api-for-skipping-validations.md
paul-maidment: @osherdp The API addresses have changed to something more conventional in the implementation, could you mark this as resolved please?

osherdp: feel free to resolve any issue you're handling. I trust you :smile:

docs/enhancements/api-for-ignoring-validations.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Drawbacks
+
+* We will have to be careful when examining bug reports and triage tickets to ensure we are aware of when this feature has been used.
+* Might allow some customers to bypass things they shouldn't bypass.
+
+
+
+
+
```
osherdp, line 132: nit: remove redundant blank lines

docs/enhancements/api-for-ignoring-validations.md
```diff
@@ -0,0 +1,132 @@
+---
+title: api-for-skipping-validations
```
osherdp, line 2: really a nit, but choose either "ignoring" (in the file name and other places) "bypass" or "skipping" based on the more precise term. currently I see the 3 terms reoccurring in some places

maybe "ignore" is better because "skip" might mean something for not happening in the first place

docs/enhancements/api-for-ignoring-validations.md
```diff
@@ -0,0 +1,132 @@
...
+authors:
+  - "@pmaidmen"
+creation-date: 2023-01-08
+last-updated: 2023-01-08
```
osherdp, line 6: not sure why we have this "last-updated" but I guess it needs to be updated

docs/enhancements/api-for-ignoring-validations.md
```diff
@@ -0,0 +1,132 @@
...
+---
+
+# Allow validations to be ignored
+
+# Summary
+
+We need a new api that will allow us to ignore cluster/host validations, this will be for unsupported use.
+This API should work on a per cluster basis and will have it's own organisation level permissions to allow us to enable it for specific organisations.
```
osherdp, line 14:
```suggestion
This API should work on a per cluster basis and will have its own organisation level permissions to allow us to enable it for specific organisations.
```

docs/enhancements/api-for-ignoring-validations.md
```diff
@@ -0,0 +1,132 @@
...
+### The API
+This will consist of two endpoints.
+
+GET /v2/clusters/{cluster_id}/ignored-validations
+
+Will return the ignored host and cluster validations as a pair of JSON string arrays
+
+```
+{
+  "IgnoredHostValidations": "[\"has-memory-for-role\", \"has-cpu-cores-for-role\"]"
```
osherdp, line 42:
```suggestion
  "ignored_host_validations": "[\"has-memory-for-role\", \"has-cpu-cores-for-role\"]"
```
or at least that's what I seeing in the swagger as the convention for field names in the response

docs/enhancements/api-for-ignoring-validations.md
```diff
@@ -0,0 +1,132 @@
...
+    * All validations run.
+    * One or more validations fail and describe a problem.
+    * The user evaluates the alleged problems and decides to try the installation anyway.
+    * If the user ignores validations (by making a change to a boolean in the cluster object), we should
+        * Emit an event
+        * Record the validation skip in the cluster in some way
+        * Log the validations that were failing at the moment the validations were skipped.
+
+### Effect on Triage tickets.
+To ensure that there is less noise for Triage tickets, any cluster that has events consistent with this sort of override should be omitted from Triage ticketing, in a similar way to how this is done for jupiter installations.
```
osherdp, line 90: maybe remove customer reference from this sentence

osherdp (commented):
just a bunch of nits, the proposal itself as a whole looks ready to me

osherdp (approved):

PR #4880, opened 2023-01-09
MGMT-12483: Implement data sending for on-prem deployments

CrystalChun:
[MGMT-12483](https://issues.redhat.com/browse/MGMT-12483)
An event uploader controller is used to send event data to a remote data store (ingress/kafka) when assisted service is deployed on premises.

The controller handles all parts of sending the request which includes validating the user is opted-in, setting the correct authentication/authorization tokens, collecting the data to be sent, and sending the request.

The work being done in this pull request is related to the work being done for the SaaS deployment https://github.com/openshift/assisted-service/pull/4823

NOTE: due to potential changes to this PR, tests will be added once all reviewers have given feedback

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


CrystalChun:
/cc @filanov @carbonin 
Would appreciate any input/feedback from you all

internal/controller/controllers/events_uploader_handler.go
```diff
@@ -0,0 +1,319 @@
...
+		log:              log,
+		client:           client,
+		hasOCMPullSecret: hasOCMPullSecret,
+		enabled:          enabled,
+		endpoint:         endpoint,
+	}
+}
+
+func (u *uploader) IsEnabled(ctx context.Context) bool {
+	if u.hasOCMPullSecret && u.enabled {
```
carbonin, line 60: I wonder if `hasOCMPullSecret` is really useful.
We could just check all the time (assuming `client` is set) and only disable if we find it and the cloud auth is missing, right?

internal/events/api/event.go
```diff
@@ -72,3 +72,7 @@ type InfoEvent interface {
 	BaseEvent
 	GetInfo() string
 }
+
+type Uploader interface {
```
carbonin, line 76: Do we intend on creating a mock or alternate implementation for this?

Is there any other reason we have this interface?

cmd/main.go
```diff
@@ -158,6 +158,8 @@ var Options struct {
 
 	// Directory containing pre-generated TLS certs/keys for the ephemeral installer
 	ClusterTLSCertOverrideDir string `envconfig:"EPHEMERAL_INSTALLER_CLUSTER_TLS_CERTS_OVERRIDE_DIR" default:""`
+	EnableDataCollection      bool   `envconfig:"ENABLE_DATA_COLLECTION" default:"true"`
+	DataUploadEndpoint        string `envconfig:"DATA_UPLOAD_ENDPOINT" default:""`
```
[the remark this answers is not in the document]
CrystalChun: Agreed! Will modify to reflect this

cmd/main.go
```diff
@@ -360,6 +363,7 @@ func main() {
 		if Options.DeployTarget == deployment_type_ocp {
 			ocpClient, err = k8sclient.NewK8SClient("", log)
 			failOnError(err, "Failed to create client for OCP")
+			hasOCMPullSecret = true
```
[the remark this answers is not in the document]
CrystalChun: Gotcha, I didn't realize it was only set to `ocp` in tests. Thanks for the info!

I was hoping to use the deploy target `ocp` to determine that it's actually an OCP cluster, implying the OCM pull secret exists. Although, I think [your suggestion](https://github.com/openshift/assisted-service/pull/4880/files#r1066071945) to just check for the pull secret every time will be a way around it

internal/events/api/event.go
CrystalChun: Thought about creating a mock for unit tests, but maybe there's no need to. What do you think?

carbonin: If you need a mock for unit tests then it's fine, but I just mention this when I see an interface without multiple concrete implementations.

CrystalChun: I see, thanks! I think I'll add a mock for unit testing, but remove it if I realize a mock isn't needed.

cmd/main.go
```diff
@@ -368,12 +369,15 @@ func main() {
 
 	failOnError(autoMigrationWithLeader(autoMigrationLeader, db, log), "Failed auto migration process")
 
+	Options.EventsUploaderConfig.AssistedServiceVersion = versions.GetRevision()
+	uploadHandler := controllers.NewEventsUploaderWrapper(&Options.EventsUploaderConfig, db, log, ocpClient)
```
filanov, line 373: I don't think that we always need it no?

internal/cluster/cluster.go
```diff
@@ -123,6 +123,7 @@ type API interface {
 	DeregisterInactiveCluster(ctx context.Context, maxDeregisterPerInterval int, inactiveSince strfmt.DateTime) error
 	TransformClusterToDay2(ctx context.Context, cluster *common.Cluster, db *gorm.DB) error
 	RefreshSchedulableMastersForcedTrue(ctx context.Context, clusterID strfmt.UUID) error
+	UploadEvents(ctx context.Context, clusterID *strfmt.UUID)
```
filanov, line 126: why is that part of a cluster api?

internal/controller/controllers/events_uploader_handler.go
```diff
@@ -0,0 +1,304 @@
...
+	return &uploader{
+		db:                   db,
+		log:                  log,
+		client:               client,
+		EventsUploaderConfig: *cfg,
+	}
+}
+
+func (u *uploader) UploadEvents(ctx context.Context, clusterID *strfmt.UUID, eventsHandler eventsapi.Handler) error {
+	if u.EnableDataCollection {
```
filanov, line 63: How about having a dummy uploader in main that will be created if the feature is not enabled?

internal/controller/controllers/events_uploader_handler.go
```diff
@@ -0,0 +1,304 @@
...
+	Auths map[string]serializedAuth `json:"auths"`
+}
+type serializedAuth struct {
+	Auth string `json:"auth"`
+}
+
+type EventsUploaderConfig struct {
+	DataUploadEndpoint     string `envconfig:"DATA_UPLOAD_ENDPOINT" default:""`
+	AssistedServiceVersion string
+	EnableDataCollection   bool `envconfig:"ENABLE_DATA_COLLECTION" default:"true"`
```
filanov, line 39: why true by default?

internal/cluster/cluster.go
CrystalChun: I might be misunderstanding, but since we're uploading by cluster that it could exist as a cluster API. Is it better to leave it as a function that's not part of the API?

cmd/main.go
CrystalChun: That's true. It's not needed for SaaS. I'll modify to create a dummy as you mentioned in this comment https://github.com/openshift/assisted-service/pull/4880#discussion_r1067999044

internal/controller/controllers/events_uploader_handler.go
carbonin: I think this could be a good reason to use interfaces for this.
I imagine it would be similar to how we use the `NoneAuthenticator`.

internal/controller/controllers/events_uploader_handler.go
```diff
@@ -0,0 +1,304 @@
...
+	cluster, err := common.GetClusterFromDB(db, *clusterID, common.SkipEagerLoading)
+	if err != nil {
+		log.WithError(err).Errorf("failed to find cluster from db")
+		return nil
+	}
+	return []byte(cluster.PullSecret)
+
+}
+
+func ExtractToken(data []byte, key string) string {
```
carbonin, line 289: If this isn't used externally it probably doesn't need to be public.

internal/events/api/event.go
carbonin: Also I would put this with the implementation of the interface rather than in the events package unless there's some reason for it to be here that I'm not seeing.

internal/controller/controllers/events_uploader_handler.go
CrystalChun: It was decided that this would be enabled by default unless the user explicitly opts out to mimic the behavior of telemetry (decision #3 in [kick off meeting doc](https://docs.google.com/document/d/1AAkKfr4acWVuk6GyPL901dSbd1XKaFaFA7vSQQZkyn0/edit?usp=sharing))

Do you think it should actually be disabled by default instead?

internal/events/api/event.go
CrystalChun: Thanks for commenting on this. I agree that it doesn't fit with the events items. I moved this to its own folder with the interface as a [separate file](https://github.com/openshift/assisted-service/pull/4880/files#diff-97008778ca8858564f55e9743e204f1f2a4a6475a2fe58517fdbff858d4e8e5aR1). What do you think?

internal/controller/controllers/events_uploader_handler.go
CrystalChun: Modified the logic to reflect this! May I have your input on how it's implemented?

[`getAuthToken`](https://github.com/openshift/assisted-service/pull/4880/files#diff-be3c5c34d4670110a2ffa6cd391d344a1c14941e7813cce8eb05cd9e92faf686R43) is run every time and [`getOCMPullSecret`](https://github.com/openshift/assisted-service/pull/4880/files#diff-be3c5c34d4670110a2ffa6cd391d344a1c14941e7813cce8eb05cd9e92faf686R29) checks if the OCM pull secret exists (and returns it if it does)

Whenever the OCM pull secret doesn't exist, I'm assuming that it's not an OCP cluster and we can safely use the cluster's pull secret.

internal/controller/controllers/events_uploader_handler.go
CrystalChun: Good points! Created a dummy uploader [here](https://github.com/openshift/assisted-service/pull/4880/files#diff-ee4f3756eb0ee07d93469c4f3c7182fbaf5549c1e51f700ea4b4b32dddfdfbfdR1) and [`NewUploadHandler`](https://github.com/openshift/assisted-service/pull/4880/files#diff-97008778ca8858564f55e9743e204f1f2a4a6475a2fe58517fdbff858d4e8e5aR23) will create either the actual event uploader or the dummy uploader depending on if it's enabled.

internal/controller/controllers/events_uploader_handler.go
CrystalChun: Good point, thank you! Modified it to a private function

cmd/main.go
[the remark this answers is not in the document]
CrystalChun: Moved these to their own config [here](https://github.com/openshift/assisted-service/pull/4880/files#diff-97008778ca8858564f55e9743e204f1f2a4a6475a2fe58517fdbff858d4e8e5aR17-R21). How does it look?

internal/uploader/uploader.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+
+	"github.com/go-openapi/strfmt"
+	eventsapi "github.com/openshift/assisted-service/internal/events/api"
+	"github.com/openshift/assisted-service/pkg/k8sclient"
+	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+type UploadHandler interface {
```
carbonin, line 13: Maybe just `Client` since `uploader` is already the package name?

Then the callers would do something like:
```go
dataUploadClient := uploader.NewClient(...)
```

internal/uploader/events_uploader_handler.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/openshift/assisted-service/pkg/k8sclient"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm"
+)
+
+var _ UploadHandler = &EventsUploader{}
+
+type EventsUploader struct {
```
carbonin, line 29: Maybe the file name should match?

`events_uploader.go`?

internal/uploader/uploader.go
```diff
@@ -0,0 +1,28 @@
...
+	Upload(ctx context.Context, clusterID *strfmt.UUID, eventsHandler eventsapi.Handler) error
```
carbonin, line 14: Are we ever going to use this interface to upload anything other than events?
Maybe it's worth making this more specific like `UploadEvents` if we're eventually going to have `UploadLogs`, for example.

internal/uploader/events_uploader_handler.go
```diff
@@ -0,0 +1,247 @@
...
+	Config
+	db     *gorm.DB
+	log    log.FieldLogger
+	client k8sclient.K8SClient
+}
+
+func NewEventsUploader(cfg *Config, db *gorm.DB, log logrus.FieldLogger, client k8sclient.K8SClient) *EventsUploader {
```
carbonin, line 36: Is anyone using this directly? I figure all callers would be using the function that returns an interface.

internal/uploader/events_uploader_handler.go
carbonin: Also, does this need to be exported since we have an interface?

internal/uploader/dummy.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"context"
+
+	"github.com/go-openapi/strfmt"
+	eventsapi "github.com/openshift/assisted-service/internal/events/api"
+	"github.com/sirupsen/logrus"
+)
+
+type DummyUploader struct {
```
carbonin, line 11: Same here, not sure this type needs to be exported.

internal/uploader/dummy.go
```diff
@@ -0,0 +1,26 @@
...
+	log logrus.FieldLogger
+}
+
+var _ UploadHandler = &DummyUploader{}
+
+func NewDummyUploader(log logrus.FieldLogger) *DummyUploader {
```
carbonin, line 17: Probably don't need this either if the interface "constructor" is returning this.

carbonin (commented):
Looking good, just some questions about naming

internal/uploader/uploader.go
CrystalChun: That definitely sounds a lot better! Modified it to `Client`, how does it look?

internal/uploader/events_uploader_handler.go
CrystalChun: True, modified it to reflect this!

internal/uploader/uploader.go
CrystalChun: Gotcha, making it more specific is better. Modified it to `UploadEvents`!

internal/uploader/events_uploader_handler.go
CrystalChun: Good point, removed this function!

internal/uploader/dummy.go
CrystalChun: Good point, removed this function!

internal/uploader/dummy.go
CrystalChun: Gotcha, modified it to a private type!

CrystalChun:
/rename MGMT-12486: Implement data sending for on-prem deployments

CrystalChun:
/retitle [MGMT-12486](https://issues.redhat.com//browse/MGMT-12486): Implement data sending for on-prem deployments

CrystalChun:
/retitle MGMT-12486: Implement data sending for on-prem deployments

internal/uploader/auth_utils_test.go
```diff
@@ -0,0 +1,141 @@
...
+		Expect(token).To(BeEmpty())
+	})
+	It("fails to extract a token when the pull secret doesn't have the right entry", func() {
+		pullSecret := `{"auths":{"clouds.openshift.com":{"auth":"dXNlcjpwYXNzd29yZAo=","email":"r@r.com"}}}`
+		token, err := extractToken([]byte(pullSecret), "cloud.openshift.com")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(token).To(BeEmpty())
+	})
+	It("fails to extract a token when the pull secret doesn't have a token for the entry", func() {
+		pullSecret := `{"auths":{"clouds.openshift.com":{"email":"r@r.com"}}}`
```
carbonin, line 42: This looks like it should be `cloud.openshift.com` for this test to be valid. Otherwise it's just the same as the previous one, right?

internal/uploader/events_uploader_test.go
```diff
@@ -0,0 +1,299 @@
...
+				Expect(err).NotTo(HaveOccurred())
+				checkFileContents(tarReader, infraEnvJson)
+			case fmt.Sprintf("%s/events/cluster.json", clusterID):
+				cluster, err := common.GetClusterFromDBWithHosts(db, clusterID)
+				Expect(err).NotTo(HaveOccurred())
+				clusterJson, err := json.Marshal(cluster.Cluster)
+				Expect(err).NotTo(HaveOccurred())
+				checkFileContents(tarReader, clusterJson)
+			case fmt.Sprintf("%s/events.json", clusterID):
+				eventJson, err := json.MarshalIndent([]*common.Event{}, "", " ")
```
carbonin, line 245: Should we be testing with some events?

internal/uploader/events_uploader_test.go
```diff
@@ -0,0 +1,299 @@
...
+	"github.com/openshift/assisted-service/models"
+	"github.com/openshift/assisted-service/pkg/k8sclient"
+	"github.com/pkg/errors"
+	"gorm.io/gorm"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
```
carbonin, line 29: You can create a test for the request method using [httptest](https://pkg.go.dev/net/http/httptest@go1.19.5)

We do it in the image service in a few places if you need examples.

internal/uploader/auth_utils_test.go
CrystalChun: Oops, right! Thanks for catching this! Accidental copy-paste error 😅

internal/uploader/events_uploader_test.go
CrystalChun: Thanks for the pointer! I've tried to do it [here](https://github.com/openshift/assisted-service/pull/4880/commits/248f4ac043767feb8bbc6843cd7f1e09c143f509#diff-7eb6290c1b1e1c5de47c84bd0946399c43bb374390d36e89483bfb0a5af7e122R282-R302)

Is this alright? I'm not sure if I implemented it correctly or if I'm testing the right things 😅

internal/uploader/events_uploader_test.go
CrystalChun: Good idea! I've tried to do it [here](https://github.com/openshift/assisted-service/pull/4880/commits/248f4ac043767feb8bbc6843cd7f1e09c143f509#diff-7eb6290c1b1e1c5de47c84bd0946399c43bb374390d36e89483bfb0a5af7e122R220-R237) but I'm not sure it's a good test 😅 

Are there better tests to do or should I be testing in a different place?

internal/uploader/auth_utils.go
```diff
@@ -0,0 +1,96 @@
...
+}
+type serializedAuth struct {
+	Auth  string `json:"auth"`
+	Email string `json:"email,omitempty"`
+}
+
+func getPullSecret(db *gorm.DB, clusterID *strfmt.UUID, k8sclient k8sclient.K8SClient, key string) (*serializedAuth, error) {
+	data, exists := getOCMPullSecret(k8sclient)
+	if !exists {
+		data = getTokenFromCluster(db, clusterID)
```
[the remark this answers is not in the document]
CrystalChun: I thought the failure point could exist in [line 38](https://github.com/openshift/assisted-service/pull/4880/files/e42e509a56ae73ac17cda10a6a3412d8bdb9df6b#diff-be3c5c34d4670110a2ffa6cd391d344a1c14941e7813cce8eb05cd9e92faf686R38) where if `data` is `nil` after both of these calls, it would error out here. 

Do you think it's better to exit early before even trying to unmarshal?

internal/uploader/auth_utils.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		return secret.Data[corev1.DockerConfigJsonKey], true
+	}
+	return nil, false
+}
+
+func getTokenFromCluster(db *gorm.DB, clusterID *strfmt.UUID) []byte {
+	cluster, err := common.GetClusterFromDB(db, *clusterID, common.SkipEagerLoading)
+	if err != nil {
+		log.WithError(err).Errorf("failed to find cluster in db")
```
masayag, line 64: pls add cluster-ID to this message for debuggability

internal/uploader/events_uploader.go
```diff
@@ -0,0 +1,261 @@
...
+		return nil, errors.New("no cluster specified for hosts file")
+	}
+	var infraEnvID *strfmt.UUID
+	hosts := cluster.Cluster.Hosts
+	if hosts != nil {
+		hostJson, err := json.MarshalIndent(hosts, "", " ")
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed marshalling hosts for cluster %s into file", *cluster.ID)
+		}
+		for _, host := range hosts {
```
danielerez, line 233: funk.ForEach could look a bit nicer for this

internal/uploader/events_uploader.go
```diff
@@ -0,0 +1,261 @@
...
+
+func infraEnvFile(db *gorm.DB, tw *tar.Writer, infraEnvID *strfmt.UUID, clusterID *strfmt.UUID) error {
+	if infraEnvID != nil {
+		infraEnv, err := common.GetInfraEnvFromDB(db, *infraEnvID)
+		if err != nil {
+			return errors.Wrapf(err, "error getting infra-env %s from db", *infraEnvID)
+		}
+		iJson, err := json.Marshal(infraEnv.InfraEnv)
+		if err != nil {
+			return errors.Wrapf(err, "error marshalling infra-env")
```
danielerez, line 214: probably worth adding infraEnvID also to the error here.

internal/uploader/events_uploader.go
```diff
@@ -0,0 +1,261 @@
...
+			return nil, errors.Wrapf(err, "failed to get cluster %s from db", *clusterID)
+		}
+		// To distinguish who is uploading the data
+		if cluster.EmailDomain == "" {
+			cluster.EmailDomain = getEmailDomain(pullSecret.Email)
+		}
+		if cluster.UserName == "" {
+			if token, tokenErr := ensureTokenFormat(pullSecret.Auth); tokenErr == nil {
+				cluster.UserName = getUsernameFromToken(token)
+			}
```
danielerez, line 194: do we need some warning log in case of tokenErr?

internal/uploader/uploader.go
```diff
@@ -0,0 +1,34 @@
...
+func NewClient(cfg *Config, db *gorm.DB, log logrus.FieldLogger, client k8sclient.K8SClient) Client {
+	if cfg.EnableDataCollection {
+		return &eventsUploader{
+			db:     db,
+			log:    log,
+			client: client,
+			Config: *cfg,
+		}
+	}
+	return &dummyUploader{log: log}
```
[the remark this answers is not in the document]
CrystalChun: @filanov and @carbonin suggested to create a dummy ([convo](https://github.com/openshift/assisted-service/pull/4880#discussion_r1067999044))

From my understanding, having a dummy is a cleaner declaration in main.

Is there a benefit of doing it this way vs using a dummy?

internal/uploader/events_uploader.go
CrystalChun: Thanks for the suggestion! I'm not sure I fully understand, so how does this look?
```
		funk.ForEach(hosts, func(host *models.Host){
			if host.InfraEnvID.String() != "" {
				infraEnvID = &host.InfraEnvID
				return
			}
		})
```

internal/uploader/auth_utils.go
CrystalChun: Sorry this log was for my own testing; I've removed it since I thought the error returned here doesn't matter for finding the cluster's pull secret. 

Do you think it's better to have a debug log here, though?

internal/uploader/auth_utils.go
```diff
@@ -0,0 +1,96 @@
...
+func getOCMPullSecret(k8sclient k8sclient.K8SClient) ([]byte, bool) {
+	if k8sclient != nil {
+		secret, err := k8sclient.GetSecret("openshift-config", "pull-secret")
+		if err != nil {
+			err = client.IgnoreNotFound(err)
+			// A "not found" error (err == nil) should return false
+			// indicating that the secret does not exist
+			return nil, err != nil
+		}
+		return secret.Data[corev1.DockerConfigJsonKey], true
```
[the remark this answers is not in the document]
CrystalChun: Good idea, modified to check beforehand!

internal/uploader/dummy.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+type dummyUploader struct {
+	log logrus.FieldLogger
+}
+
+var _ Client = &dummyUploader{}
+
+func (d *dummyUploader) UploadEvents(ctx context.Context, clusterID *strfmt.UUID, eventsHandler eventsapi.Handler) error {
+	d.log.Infof("No uploader implemented")
```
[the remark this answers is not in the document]
CrystalChun: Good idea, thank you! Modified to add the cluster ID

internal/uploader/auth_utils.go
```diff
@@ -0,0 +1,96 @@
...
+func getTokenFromCluster(db *gorm.DB, clusterID *strfmt.UUID) []byte {
+	cluster, err := common.GetClusterFromDB(db, *clusterID, common.SkipEagerLoading)
+	if err != nil {
+		log.WithError(err).Errorf("failed to find cluster in db")
+		return nil
+	}
+	return []byte(cluster.PullSecret)
+}
+
+func ensureTokenFormat(token string) (string, error) {
```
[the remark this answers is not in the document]
CrystalChun: This is also used for the [OCM pull secret](https://docs.openshift.com/container-platform/4.10/openshift_images/managing_images/using-image-pull-secrets.html) (code [here](https://github.com/openshift/assisted-service/pull/4880/files#diff-be3c5c34d4670110a2ffa6cd391d344a1c14941e7813cce8eb05cd9e92faf686R44)) which is received from the Kubernetes secret in the cluster.

internal/uploader/auth_utils.go
```diff
@@ -0,0 +1,96 @@
...
+	formattedToken := strings.TrimSpace(token)
+	if strings.Contains(formattedToken, "\n") || strings.Contains(formattedToken, "\r") {
+		return "", fmt.Errorf("authorization token is not valid: contains newlines")
+	}
+	return formattedToken, nil
+}
+
+func getUsernameFromToken(token string) string {
```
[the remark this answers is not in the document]
CrystalChun: Thanks for the pointer to that! Didn't realize there was a whole library implemented already. I've modified it to use this instead, how does it look?

internal/uploader/auth_utils.go
```diff
@@ -0,0 +1,96 @@
...
+	if !exists {
+		data = getTokenFromCluster(db, clusterID)
+	}
+	var pullSecret serializedAuthMap
+	if err := json.Unmarshal(data, &pullSecret); err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal pull secret")
+	}
+	if auth, ok := pullSecret.Auths[key]; ok {
+		return &auth, nil
+	}
```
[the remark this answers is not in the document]
CrystalChun: Good idea, thank you for the pointer! Modified it to use this instead, how does it look?

internal/uploader/auth_utils.go
```diff
@@ -0,0 +1,96 @@
...
+			return string(res[0])
+		}
+	}
+	return ""
+}
+
+func getEmailDomain(email string) string {
+	var domain string
+	delimiterIdx := strings.LastIndex(email, "@")
+	if delimiterIdx >= 0 {
```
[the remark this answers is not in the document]
CrystalChun: Gotcha, thank you! Modified to reflect this, how does it look?

internal/uploader/events_uploader.go
```diff
@@ -0,0 +1,261 @@
...
+		failedFilesCount++
+	}
+
+	if err := eventsFile(ctx, clusterID, eventsHandler, tw); err != nil {
+		log.WithError(err).Warnf("failed finding events associated with cluster ID %s", clusterID)
+		failedFilesCount++
+	}
+
+	// There are no files to upload
+	if failedFilesCount >= 4 {
```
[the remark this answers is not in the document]
CrystalChun: Good idea, thank you! Modified it to reflect this

internal/uploader/events_uploader.go
CrystalChun: I actually ended up removing this since the `ParsePullSecret` function you mentioned gives us the username directly!

internal/uploader/auth_utils.go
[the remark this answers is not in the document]
CrystalChun: Also because I needed the email, is it alright if I modified the original `ParsePullSecret` function and `PullSecretCreds` to include an email field?

internal/uploader/auth_utils.go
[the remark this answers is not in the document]
masayag: I don't think it is needed to be logged here as well if it is already handled by the caller.

internal/uploader/events_uploader.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func prepareBody(buffer *bytes.Buffer) (*bytes.Buffer, string, error) {
+	if buffer == nil || buffer.Bytes() == nil {
+		return nil, "", errors.Errorf("no data passed to prepare body")
+	}
+	mimeHeader := make(textproto.MIMEHeader)
+	mimeHeader.Set("Content-Disposition", fmt.Sprintf(`form-data; name="%s"; filename="%s"`, "file", "events.tgz"))
+	mimeHeader.Set("Content-Type", "application/vnd.redhat.assisted-installer.events+tar")
+
```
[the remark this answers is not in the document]
CrystalChun: Oh that's a good idea, @rccrdpccl any thoughts? Do you think ingress will be able to pass the last-modified header to us when we consume the data from it?

internal/uploader/auth_utils.go
```diff
@@ -0,0 +1,94 @@
...
+		if pullSecret, ok := secret.Data[corev1.DockerConfigJsonKey]; ok {
+			return pullSecret, true
+		}
+		return nil, true
+	}
+	return nil, false
+}
+
+func getTokenFromCluster(db *gorm.DB, clusterID *strfmt.UUID) []byte {
+	if cluster, err := common.GetClusterFromDB(db, *clusterID, common.SkipEagerLoading); err == nil {
```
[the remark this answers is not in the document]
CrystalChun: Thank you for explaining this to me! I didn't realize it was in an open transaction. I went ahead and took both of your suggestions:

1. Passing in the full cluster object to `UploadEvents`
2. Uploading using a go-routine [here](https://github.com/openshift/assisted-service/pull/4880/commits/6514bf1b4cb0052b0c079e89d232a1c7837d25b6#diff-ecbdbfaa3e0752590ccb7b5c0387c308de661ba622120129edf662c02fa1ddadR1454) and [here](https://github.com/openshift/assisted-service/pull/4880/commits/6514bf1b4cb0052b0c079e89d232a1c7837d25b6#diff-ecbdbfaa3e0752590ccb7b5c0387c308de661ba622120129edf662c02fa1ddadR751) 

How does it look?

internal/uploader/events_uploader.go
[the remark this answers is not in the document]
rccrdpccl: I can't see in what way Last-Modified could be useful here? It is a response header useful for caching, but here we are uploading so I can't see any possible usage.
Note also that when consuming we'll be fetching directly from kafka (which is not http so wouldn't have these kind of headers) - and we will always fetch the tar from the S3-backed url, no matter of when the file was last-modified.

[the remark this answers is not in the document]
CrystalChun: I was thinking similarly, thank you for confirming and for the explanation about when we fetch!

internal/uploader/auth_utils.go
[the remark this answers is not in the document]
CrystalChun: Gotcha, thanks for confirming!

internal/uploader/auth_utils.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func getOCMPullSecret(k8sclient k8sclient.K8SClient) ([]byte, bool) {
+	if k8sclient != nil {
+		secret, err := k8sclient.GetSecret("openshift-config", "pull-secret")
+		if err != nil {
+			err = client.IgnoreNotFound(err)
+			// A "not found" error (err == nil) should return false
+			// indicating that the secret does not exist
+			return nil, err != nil
```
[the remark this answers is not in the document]
CrystalChun: I'm not sure about the importance of it; an error here indicates that there was something wrong with getting the OCM pull secret, but wouldn't that already indicate something is already wrong with the cluster?

internal/cluster/cluster.go
```diff
@@ -916,6 +920,14 @@ func (m *Manager) PrepareForInstallation(ctx context.Context, c *common.Cluster,
 	return err
 }
 
+func (m *Manager) UploadEvents(ctx context.Context, cluster *common.Cluster) {
+	log := logutil.FromContext(ctx, m.log)
+	log.Infof("Uploading events for %s", *cluster.ID)
```
paul-maidment, line 925: Can cluster be nil, should this be checked for somewhere?

internal/uploader/uploader.go
```diff
@@ -0,0 +1,34 @@
...
+
+//go:generate mockgen -source=uploader.go -package=uploader -destination=mock_uploader.go
+type Client interface {
+	UploadEvents(ctx context.Context, cluster *common.Cluster, eventsHandler eventsapi.Handler) error
+}
+
+type Config struct {
+	DataUploadEndpoint     string `envconfig:"DATA_UPLOAD_ENDPOINT" default:""`
+	AssistedServiceVersion string
+	EnableDataCollection   bool `envconfig:"ENABLE_DATA_COLLECTION" default:"true"`
```
danielerez, line 21: Do we need it also in the operator? If not, need to add this env var and set it to false in ASC [controller](https://github.com/openshift/assisted-service/blob/3c19e8e57e61279c145b3ea70e8fc63[…]nternal/controller/controllers/agentserviceconfig_controller.go](https://github.com/openshift/assisted-service/blob/3c19e8e57e61279c145b3ea70e8fc63e8f40d5bb/internal/controller/controllers/agentserviceconfig_controller.go#L1087)).
Also, need to disable it when running in a disconnected env(?) @masayag

slaviered: we do not need it in the operator but we do need it in the service for kube-api env. It should be propogated to it then.

internal/uploader/auth_utils.go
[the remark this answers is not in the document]
slaviered: Does this authorization scheme is common to all installation envs ?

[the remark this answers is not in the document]
CrystalChun: @danielerez oh you're right! Sorry about that, I misunderstood! I'll remove it since it is in the ParsePullSecret function.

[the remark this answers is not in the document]
CrystalChun: @slaviered all clusters getting installed will have their data sent using the `cloud.openshift.com` pull secret

internal/uploader/uploader.go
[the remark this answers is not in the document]
CrystalChun: Thanks for bringing it up, though! It's good to know there are several ways to implement this.

internal/uploader/events_uploader.go
CrystalChun: Wow that does look a lot cleaner, thank you! I'll modify this

internal/uploader/uploader.go
CrystalChun: I think you're both correct, it needs to be in the operator since this feature is exposed for ACM/ZTP and it needs to be in the service for the kube-api (which I think it's already propagated [here](https://github.com/openshift/assisted-service/blob/3849ee79dcab20bf77873edfee96ee084f1beca1/cmd/main.go#L138), but correct me if I need to add additional changes to propagate it!).

For disconnected, it'll auto-fail when attempting to send the data, but is it better to set it to false directly? If we're setting to false directly, how do we know it's a disconnected environment?

internal/uploader/events_uploader.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	buffer, err := prepareFiles(ctx, e.db, clusterID, eventsHandler, pullSecret)
+	if err != nil {
+		return errors.Wrapf(err, "failed to prepare files to upload for cluster %s", *clusterID)
+	}
+	formBuffer, contentType, err := prepareBody(buffer)
+	if err != nil {
+		return errors.Wrapf(err, "failed to prepare event data content to upload for cluster %s", clusterID)
+	}
+	req, err := http.NewRequest(http.MethodPost, e.DataUploadEndpoint, formBuffer)
```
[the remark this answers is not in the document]
CrystalChun: Thanks for the suggestion! This request actually goes to a 3rd party service - I'm not familiar with `NewRequestWithContext`, is there a benefit to using `NewRequestWithContext` even if our service is not on the receiving end of this request?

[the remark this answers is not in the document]
masayag: no.

TL;DR: 
See https://pkg.go.dev/net/http#NewRequestWithContext

> For an outgoing client request, the context controls the entire lifetime of a request and its response: obtaining a connection, sending the request, and reading the response headers and body

If the context is canceled, it will terminate the request, causing the entire flow to end.
Since now this one is running on its own go routine, it should probably be decoupled from the main flow and should not be executed with the main flow context.

internal/uploader/auth_utils.go
masayag: this one is no longer relevant since the cluster entity is passed to this function.

internal/uploader/events_uploader.go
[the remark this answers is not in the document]
CrystalChun: Thanks for the explanation! I'll leave this as it is then?

internal/uploader/uploader.go
slaviered: There are different build configuration depending on the environment (saas (staging and prod) and on prem). I would set it to false as default and set it intentionally to true on the selected environments. In this case you can deploy your feature gradually

danielerez: So do we need ENABLE_DATA_COLLECTION configurable also using the operator? If so, then it should be added to the [AgentServiceConfigSpec](https://github.com/openshift/assisted-service/blob/a3b5bacf62a593c3f1cb281654e01ffdd7d78a28/api/v1beta1/agentserviceconfig_types.go#L56) and propagate to the assisted-service [configmap](https://github.com/openshift/assisted-service/blob/3c19e8e57e61279c145b3ea70e8fc63e8f40d5bb/internal/controller/controllers/agentserviceconfig_controller.go#L1087) in agentserviceconfig_controller.
For deploying using podman, the configmap are [here](https://github.com/openshift/assisted-service/blob/84d7256be1c24f1ec4dc1dadf02bbc775d5ff3d3/deploy/podman/). But maybe having the default as `false` is indeed safer so we could explicitly enable it where we sure it's needed.

slaviered: I do not think it should be part of the AgentServiceConfigSpec. It belongs as other flags to the ai config map. The scrapping is done deep in the AI logic flow and has nothing to do with the operator

danielerez: Yeah, but the assisted-service configmap is getting overwritten on each asc reconcile loop. So the flag should be simply hardcoded in the cm if we don't want it in ASC.

CrystalChun: I think Daniel is correct, it needs to be configurable from the operator as well. I wasn't sure whether to 
1. Add the environment variable directly to the AgentServiceConfigSpec or
2. Just use the environment variable override annotation ([doc](https://github.com/openshift/assisted-service/blob/master/docs/operator.md#specifying-environmental-variables-via-configmap))

Which one do you all think is better?

CrystalChun: I did push the change to add these variables to the assisted-service [configmap](https://github.com/openshift/assisted-service/blob/3c19e8e57e61279c145b3ea70e8fc63e8f40d5bb/internal/controller/controllers/agentserviceconfig_controller.go#L1087) in agentserviceconfig_controller: https://github.com/openshift/assisted-service/blob/22b14d0e414ba9b529b14117f9b5609a2603c991/internal/controller/controllers/agentserviceconfig_controller.go#L1121-L1122

internal/uploader/events_uploader.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "failed preparing new https request for endpoint %s", e.DataUploadEndpoint)
```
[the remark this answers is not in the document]
CrystalChun: That's right, thank you! Modified it to be an `https` client and to use `tls` here: 
https://github.com/openshift/assisted-service/blob/eea830d3cd77dba7f8d8da31c4be8b740195c519/internal/uploader/events_uploader.go#L100-L105

Does it look alright?

internal/uploader/events_uploader.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	return buffer, nil
+}
+
+func eventsFile(ctx context.Context, clusterID *strfmt.UUID, eventsHandler eventsapi.Handler, tw *tar.Writer) error {
+	if eventsHandler == nil {
+		return errors.Errorf("failed to get events for cluster %s, events handler is nil", clusterID)
+	}
+	dbEvents, err := eventsHandler.V2GetEvents(ctx, clusterID, nil, nil, models.EventCategoryMetrics, models.EventCategoryUser)
+	if err != nil {
```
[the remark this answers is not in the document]
CrystalChun: > This query however does not bring the host events which I thing you want.

Is this considered a host event? 
```
 {
  "category": "user",
  "cluster_id": "20001bc6-2135-45c7-a475-a61e4023a750",
  "event_time": "2023-02-07T17:18:00.623Z",
  "host_id": "e2a4e446-90bb-4c8a-8be1-04459f6a1be5",
  "infra_env_id": "6d946cba-1238-482b-89a1-397b95b86c64",
  "message": "Host e2a4e446-90bb-4c8a-8be1-04459f6a1be5: Successfully registered",
  "name": "host_registration_succeeded",
  "request_id": "ac57d64f-b12a-4344-b0da-9c576a5ab827",
  "severity": "info"
 },
```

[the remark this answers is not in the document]
CrystalChun: > Since the events here are limited to a cluster I am not sure we will have tons of them but still it is worth having some number here. 

Sorry, I'm not sure what you mean by some number here?

internal/uploader/uploader.go
danielerez: > I think Daniel is correct, it needs to be configurable from the operator as well. I wasn't sure whether to
> 
> 1. Add the environment variable directly to the AgentServiceConfigSpec or
> 2. Just use the environment variable override annotation ([doc](https://github.com/openshift/assisted-service/blob/master/docs/operator.md#specifying-environmental-variables-via-configmap))
> 
> Which one do you all think is better?

Do we have any notion to how common this functionality could be for ACM/infrastructure-operator users?
It's just that using the override annotation is kind of a workaround, on the other hand, adding it to the ASC spec might be an overkill just for this. 
@filanov @avishayt  WDYT? Do we need `ENABLE_DATA_COLLECTION` configurable in the operator?

internal/uploader/events_uploader.go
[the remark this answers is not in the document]
CrystalChun: > Another problem is that you are keep sending old events from previous installation attempts. Do you have a mechanism to handle duplications down the road? Is it a concern at all?

@rccrdpccl and I have discussed data duplication a little bit, but I think it's a good topic to nail down. 

Initially, we discussed mitigating duplicate data from the receiving end. In other words, when the consumer of the ingress (kafka) server gets the data, it won't store any data that it's already seen before. 

To determine if it's already seen the data, we spoke about using the username and cluster id.

@rccrdpccl any thoughts? Would you like an additional field to be added to determine duplicate data?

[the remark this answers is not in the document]
slaviered: 1. Yes. The example you pasted here is a host level event
2. By numbers I mean how many events we have per cluster of 5 node, 20 nodes. We can extract this data by querying our prod db - I can show you how.
3. For identification you will also need the host_id and infraenv_id (for unbound events) and a timestamp. On multiple installations of the same cluster we have the same events (logically speaking) but their timestamp differ,

internal/uploader/uploader.go
slaviered: How is ENABLE_DATA_COLLECTION differ from LOG_LEVEL for example?

internal/cluster/cluster.go
```diff
@@ -745,6 +748,7 @@ func (m *Manager) CancelInstallation(ctx context.Context, c *common.Cluster, rea
 		if funk.Contains(installationStates, lastState.srcState) {
 			m.metricAPI.ClusterInstallationFinished(ctx, models.ClusterStatusCancelled, lastState.srcState, c.OpenshiftVersion, *c.ID, c.EmailDomain, c.InstallStartedAt)
 		}
+		go m.UploadEvents(ctx, c)
```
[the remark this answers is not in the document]
masayag: There is another good reason to avoid it (although I was the one to suggest it) - since the recent events aren't committed to the DB at this point (the transaction may be still open), they won't be collected.
Seems like a safer option is to push the call to `m.UploadEvents` to a later phase.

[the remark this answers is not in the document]
CrystalChun: Thanks both of you for catching this! 

Maybe an alternative option is to call it from here? https://github.com/openshift/assisted-service/blob/445e8646e95c70b74c9131c57945eff1289f73da/internal/bminventory/inventory.go#L3817
After the transaction completes? 


I'll look into what @slaviered suggested too - to make background goroutine and call it from a channel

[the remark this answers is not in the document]
filanov: Another option can be to have another thread as Sarah suggested just instead of using channels it can look at the DB for clusters that got into a final state, search events basically, and it can have a small table to track on which clusters it already uploaded. 
Moti, Sarah what do you think?

internal/uploader/uploader.go
filanov: If it's enable until fail i don't see a reason not to enable by default in the operator.

internal/cluster/cluster.go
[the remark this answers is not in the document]
masayag: Sounds better, as it decouples the execution of uploading the events from this context.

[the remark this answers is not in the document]
slaviered: I agree as well. The monitor thread is a good example on how to do it.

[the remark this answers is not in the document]
CrystalChun: Thank you all for the feedback, suggestions, and ideas! I've gone ahead and tried to implement the option @filanov suggested. 

Changes:
1. Started an uploader thread in main as @slaviered suggested https://github.com/openshift/assisted-service/pull/4880/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R390-R394
2. Modified the `UploadEvents` function that's called in the thread https://github.com/openshift/assisted-service/pull/4880/files#diff-ecbdbfaa3e0752590ccb7b5c0387c308de661ba622120129edf662c02fa1ddadR923 
    - This function checks all cluster statuses and only uploads for clusters that: 
        - haven't already been uploaded and 
        - its status is one of [`Installed`, `error`, `cancelled`] or ([`Adding Hosts`] AND `InstallTimestamp` is not zero)
4. Added a new DB table to track which clusters have been uploaded https://github.com/openshift/assisted-service/pull/4880/files#diff-01bdb7dd494494bebd086fc6d8172ab9a4da151e7fb194a40616bfba462a27e4R25-R28 (naming suggested appreciated) along with the following changes to support this table:
    - Creating an entry in the table during `registerCluster` https://github.com/openshift/assisted-service/pull/4880/files#diff-268adb06c81106332ee620da395932a03fdf3059b29acc0dbcd66d6cf484f8e9R74
    - Deleting the entry during `DeregisterCluster` https://github.com/openshift/assisted-service/pull/4880/files#diff-268adb06c81106332ee620da395932a03fdf3059b29acc0dbcd66d6cf484f8e9R148
    - Updating the entry if the data was uploaded https://github.com/openshift/assisted-service/pull/4880/files#diff-ecbdbfaa3e0752590ccb7b5c0387c308de661ba622120129edf662c02fa1ddadR970-R974

Please let me know what you think. Especially with regards the DB items!

internal/common/db.go
```diff
@@ -22,6 +22,11 @@ const (
 	ProgressWeightFinalizingStage               float64 = 0.15
 )
 
+type ClusterUpload struct {
+	ID       *strfmt.UUID `json:"id" gorm:"primaryKey"`
```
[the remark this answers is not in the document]
filanov: I suggest to use the existing cluster table, this way it will be easy to search in the monitor and you will not need to sync two tables.

internal/cluster/cluster.go
```diff
@@ -916,6 +920,61 @@ func (m *Manager) PrepareForInstallation(ctx context.Context, c *common.Cluster,
...
+			log.Info("not a leader, exiting upload")
+			return
+		}
+		if !clusterToUpload.Uploaded {
+			cluster, dberr := common.GetClusterFromDB(m.db, *clusterToUpload.ID, common.SkipEagerLoading)
+			if dberr != nil {
+				log.WithError(err).Warnf("couldn't get cluster %s from db", clusterToUpload.ID)
+				continue
+			}
+			if funk.ContainsString([]string{models.ClusterStatusInstalled, models.ClusterStatusCancelled, models.ClusterStatusError}, *cluster.Cluster.Status) ||
```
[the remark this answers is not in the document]
CrystalChun: Good idea, modified to `isEligibleForUpload(cluster)`!

internal/cluster/cluster.go
```diff
@@ -916,6 +920,61 @@ func (m *Manager) PrepareForInstallation(ctx context.Context, c *common.Cluster,
...
+				}
+				log.Infof("successfully uploaded events for cluster %s", *clusterToUpload.ID)
+				clustersUploaded = append(clustersUploaded, clusterToUpload.ID)
+
+				continue
+			}
+			log.Debugf("cluster %s is in status %s so will not upload", clusterToUpload.ID, *cluster.Cluster.Status)
+		}
+	}
+	for _, uploadedID := range clustersUploaded {
```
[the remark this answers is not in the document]
CrystalChun: That makes sense. Thank you for the suggestion and explanation! I've updated it to a singular for loop that will do as you suggested. How does it look?

internal/common/db.go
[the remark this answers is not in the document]
CrystalChun: Thank you both for your input! I've removed this table and added another column to the cluster table to indicate whether or not the data for a cluster was uploaded

internal/uploader/uploader.go
masayag: Shouldn't the default for `EnableDataCollection` here be set to `false` but in the config for the operator to `true`?

internal/cluster/cluster.go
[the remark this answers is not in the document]
masayag: Looks good :+1:

internal/cluster/cluster.go
```diff
@@ -917,6 +921,53 @@ func (m *Manager) PrepareForInstallation(ctx context.Context, c *common.Cluster,
...
+		log.WithError(err).Warnf("failed to get cluster upload table")
+		return
+	}
+	clustersUploaded := 0
+	for _, cluster := range clustersToUpload {
+		if !m.leaderElector.IsLeader() {
+			log.Info("not a leader, exiting event uploader")
+			return
+		}
+		if !cluster.Uploaded && isEligibleForUpload(cluster) {
```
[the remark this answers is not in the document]
CrystalChun: Yes, thank you for catching that! Moved it into the `isEligibleForUpload` function

internal/cluster/cluster.go
```diff
@@ -917,6 +921,53 @@ func (m *Manager) PrepareForInstallation(ctx context.Context, c *common.Cluster,
...
+	}
+	var (
+		requestID = requestid.NewID()
+		ctx       = requestid.ToContext(context.Background(), requestID)
+		log       = requestid.RequestIDLogger(m.log, requestID)
+		err       error
+	)
+	curUploadInvokedAt := time.Now()
+	defer func() {
+		m.prevUploadInvokedAt = curUploadInvokedAt
```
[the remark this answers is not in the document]
CrystalChun: Thanks for catching this 😅 I originally used it for debugging, but it's not really used

internal/cluster/cluster.go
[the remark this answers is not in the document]
CrystalChun: Sweet, thank you!

internal/uploader/uploader.go
CrystalChun: I think we want data collection to be enabled for podman too, which isn't deployed via the operator, right? 

Then again, the variable could be set to true in the podman deploy's config file. 

Is it better to have it default to `false`?

internal/uploader/events_uploader.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func sendRequest(req *http.Request) error {
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return errors.Wrapf(err, "failed to send request to %s", req.URL)
+	}
+	if res != nil {
+		log.Infof("Response from %s: %v", req.URL, res.Status)
+		if res.StatusCode > 199 && res.StatusCode < 300 {
```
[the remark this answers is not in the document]
CrystalChun: I added some comments, how does it look?

internal/cluster/cluster.go
```diff
@@ -917,6 +920,55 @@ func (m *Manager) PrepareForInstallation(ctx context.Context, c *common.Cluster,
 	return err
 }
 
+func isEligibleForUpload(cluster *common.Cluster) bool {
```
carbonin, line 923: Can you do this in the DB query instead of in code?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1118,6 +1118,8 @@ func newAssistedCM(ctx context.Context, log logrus.FieldLogger, asc ASC) (client
 			"JWKS_URL":                    "https://api.openshift.com/.well-known/jwks.json",
 			"PUBLIC_CONTAINER_REGISTRIES": unauthenticatedRegistries(ctx, asc),
 			"HW_VALIDATOR_REQUIREMENTS":   defaultControllerHardwareRequirements,
+			"ENABLE_DATA_COLLECTION":      "True",
```
carbonin, line 1121: Does this need to be configurable? Or does the user do that in some other way?

internal/common/db.go
```diff
@@ -66,6 +66,7 @@ type Cluster struct {
 
 	IgnoredClusterValidations string `gorm:"type:text"`
 	IgnoredHostValidations    string `gorm:"type:text"`
+	Uploaded                  bool   `json:"uploaded"`
```
carbonin, line 69: Maybe add a comment here about what this field is. "Uploaded" is fairly generic.

cmd/main.go
```diff
@@ -372,14 +375,22 @@ func main() {
...
 		instructionApi, &Options.HWValidatorConfig, metricsManager, &Options.HostConfig, lead, operatorsManager, providerRegistry, Options.EnableKubeAPI, objectHandler)
 	dnsApi := dns.NewDNSHandler(Options.BMConfig.BaseDNSDomains, log)
 	manifestsGenerator := network.NewManifestsGenerator(manifestsApi, Options.ManifestsGeneratorConfig)
 	clusterApi := cluster.NewManager(Options.ClusterConfig, log.WithField("pkg", "cluster-state"), db,
-		eventStreamWriter, eventsHandler, hostApi, metricsManager, manifestsGenerator, lead, operatorsManager, ocmClient, objectHandler, dnsApi, authHandler)
+		eventStreamWriter, eventsHandler, uploadClient, hostApi, metricsManager, manifestsGenerator, lead, operatorsManager, ocmClient, objectHandler, dnsApi, authHandler)
 	infraEnvApi := infraenv.NewManager(log.WithField("pkg", "host-state"), db, objectHandler)
 
+	clusterEventsUploader := thread.New(
+		log.WithField("pkg", "cluster-events-uploader"), "Cluster Events Uploader", Options.ClusterEventsUploaderInterval, clusterApi.UploadEvents)
```
carbonin, line 390: Why move this to a separate thread?

I don't see any huge problems with it, but I just generally don't like this pattern since we're not checking the health of this routine.

rccrdpccl: On this note, what is the advantage of having a dedicated thread for this (or any other dedicated thread in this application) vs using a goroutine? We'll get thread overhead, but I can't see the benefit as the operations we launch in threads are not very resource-demanding - what am I missing?

CrystalChun: This came from a suggestion to prevent uploading data in the middle of a transaction.

Here's the background conversation: https://github.com/openshift/assisted-service/pull/4880#discussion_r1100512234

I understand your concerns, is there a better way to do this?

CrystalChun: Would including it in the health check along with the other monitor threads be a good idea to watch its health? https://github.com/openshift/assisted-service/blob/9b1ee70507b913eb4c3b90e288e85d304585ed85/cmd/main.go#L521

carbonin: Talked with @CrystalChun about this and it sounds like the reasoning was to ensure we're not sending events to an external system before committing the transaction that would actually transition the cluster to the new state.

We looked at some alternatives (like moving this to a post-transition function on the state machine), but those also run inside the same transaction (at least for cancel).

This is fine, but I think eventually we probably want to think about moving these threads out into their own container at some point in the future.

>what is the advantage of having a dedicated thread for this (or any other dedicated thread in this application) vs using a goroutine?

@rccrdpccl these are actually just goroutines in a poorly named package. This thread package is actually something we have internally to allow for looping logic. You can check it out [here](https://github.com/openshift/assisted-service/blob/f41f4b24d8773e90a816827c426925e00520ec96/pkg/thread/thread.go#L43)

internal/common/db.go
CrystalChun: Added a comment! Is the explanation ok?

internal/controller/controllers/agentserviceconfig_controller.go
CrystalChun: Discussed with Nick offline. There's another way we recommend users to opt-out and that's using the [telemetry opt-out](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html/support/remote-health-monitoring-with-connected-clusters#opting-out-remote-health-reporting). Added a comment to clarify this.

internal/cluster/cluster.go
CrystalChun: Added a db query back in the original UploadEvents function! How does it look? https://github.com/openshift/assisted-service/pull/4880/files#diff-ecbdbfaa3e0752590ccb7b5c0387c308de661ba622120129edf662c02fa1ddadR944-R946

cmd/main.go
rccrdpccl: Thanks for the feedback @carbonin - it makes more sense now. My bad for not looking into its implementation :pray:

CrystalChun:
Hi all, thank you for your reviews and feedback thus far! May I get a re-review from each of you if you're free?
@danielerez @carbonin @masayag @slaviered @filanov

carbonin:
Looks good to me. I think you can definitely move it out of draft if you don't anticipate too many more changes and unit tests are passing.

internal/cluster/cluster.go
```diff
@@ -921,6 +924,50 @@ func (m *Manager) PrepareForInstallation(ctx context.Context, c *common.Cluster,
...
+		log       = requestid.RequestIDLogger(m.log, requestID)
+		err       error
+	)
+
+	clustersToUpload, err := common.GetClustersFromDBWhere(m.db, common.SkipEagerLoading, common.SkipDeletedRecords,
+		"uploaded = ? AND (status in (?) OR (status = ? AND install_completed_at > ?))",
+		false, []string{models.ClusterStatusInstalled, models.ClusterStatusCancelled, models.ClusterStatusError},
+		models.ClusterStatusAddingHosts, strfmt.NewDateTime())
+	if err != nil {
+		log.WithError(err).Warnf("failed to get cluster upload table")
```
masayag, line 949: nit: the error message can be rephrased (there is no longer cluster-upload table in this version)

CrystalChun: Good catch, thank you!

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1118,6 +1118,9 @@ func newAssistedCM(ctx context.Context, log logrus.FieldLogger, asc ASC) (client
 			"JWKS_URL":                    "https://api.openshift.com/.well-known/jwks.json",
 			"PUBLIC_CONTAINER_REGISTRIES": unauthenticatedRegistries(ctx, asc),
 			"HW_VALIDATOR_REQUIREMENTS":   defaultControllerHardwareRequirements,
+			// The user may opt-out by using the telemetry opt-out method (removing cloud.openshift.com from their OCM pull secret)
+			"ENABLE_DATA_COLLECTION": "True",
+			"DATA_UPLOAD_ENDPOINT":   "",
```
carbonin, line 1123: Where does this actually get set?

As it is in this PR the user would have to provide this value in a configmap for this to ever work, right?

My preference would be for us to put the "real" value in our application as the default value and remove this line. We can always still override using the configmap.

internal/uploader/events_uploader_test.go
carbonin: Looks good to me

internal/uploader/uploader.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+//go:generate mockgen -source=uploader.go -package=uploader -destination=mock_uploader.go
+type Client interface {
+	UploadEvents(ctx context.Context, cluster *common.Cluster, eventsHandler eventsapi.Handler) error
+	IsEnabled() bool
+}
+
+type Config struct {
+	DataUploadEndpoint     string `envconfig:"DATA_UPLOAD_ENDPOINT" default:""`
```
carbonin, line 20: I think we should put the real value in the default here.

carbonin (approved):

PR #4538, opened 2022-10-24
MGMT-11799: change all 'go get' commands

osherdp:
Since they have the tendency to change existing runtime dependencies, this changes all occurrences of 'go get' to 'go install' for wherever we want to just install a testing-related dependency.

## List all the issues related to this PR

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [x] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [x] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

If someone has an env with ``docker`` installed, I'd like to see if ``skipper make update-minimal-debug`` works for him/her!
Basically I want to make sure we use a default podman sock address if ``podman`` command doesn't exist. The resulting "default" value isn't important, as either way it's unused when using ``docker``, just that it should not crash.

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #4887, opened 2023-01-11
MGMT-11799: change all 'go get' commands

osherdp:
Since they have the tendency to change existing runtime dependencies, this changes all occurrences of 'go get' to 'go install' for wherever we want to just install a testing-related dependency.

Backports https://github.com/openshift/assisted-service/pull/4538


osherdp:
/override ci/prow/edge-subsystem-aws
Unrelated flake

eranco74 (commented):
/lgtm

PR #4893, opened 2023-01-15
MGMT-12996: enhancement for dynamic OCP fetching

osherdp:
Suggesting how assisted-service and assisted-installer UI can be adjusted to deliver all versions in the upgrade-channel API without the need of specified configuration.

Screenshot from 2023-01-15 11-43-30
Screenshot from 2023-01-15 11-43-41

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/dynamic-fetching-of-openshift-versions.md
+++ b/docs/enhancements/dynamic-fetching-of-openshift-versions.md
@@ -105,7 +105,7 @@ DYNAMIC_OPENSHIFT_RELEASES_CONFIG='{
 }'
 ```
 
-The assisted-service periodically polls versions from the ``upgrades_info`` API
+The assisted-service periodically polls versions from the upgrade-channels API
 in the form of ``https://api.openshift.com/api/upgrades_info/v1/graph?channel=<channel>&arch=<arch>``.
 For example:
 ```
@@ -116,10 +116,10 @@ $ curl -s 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4
 ...
 ```
 
-It then adds any new version found by this process in the database. When a
-consumer of the API requests the available versions via ``/v2/openshift-versions``
-(either it's the UI or a direct REST request) versions are read from the
-database.
+When an API consumer (either it's the UI or a direct HTTP consumer) requests
+for the available versions via ``/v2/openshift-versions``, versions will be read
+from the upgrades-channels API if it weren't queried for a configurable amount
+of time (for example, an hour).
 
 To help fetching only certain versions via auto-completion like U/X, we should
 provide a parameter ``version_pattern`` for the ``/v2/openshift-versions``
@@ -231,29 +231,24 @@ We should make sure we don't consume too much memory when fetching releases.
 It's especially critical when the user queries for the different versions, but
 might happen in other cases too.
 
-We need to think of the retention policy of versions. We don't expect to have
-versions removed from the upgrade-channels API, but just in case we should only
-add versions and not mirror exactly the given result of those API responses.
-As opposed to that, when a minor release gets EOL (end-of-life) and we remove it
-from the configuration, the service should remove all the stale versions from
-the database.
+With the suggested solution we don't have a way to retain versions that were
+removed from the upgrade-channels API. While it's not probable we should keep
+that in mind.
 
 ## Design Details
 
 ### Open Questions
 
-* Is the database usage a real necessity? The idea is having the periodic thread
-  updating the database, and then the HTTP server is able to filter versions from
-  the database at its leisure. If there's a good way to do it in-memory and it
-  doesn't consume too much of it, it might be better.
 * When implementing the binaries LRU cache, should we prioritize the latest
   versions of each y-release? We expect those to be the most common ones, but
   maybe it's better to let the LRU mechanism determine what are the common
   versions by itself?
-* How should we limit database entries? In case of a bug in the code that writes
-  way too much version entries, can we make sure to mitigate the issue on our
-  side? (without interrupting app-sre for their help on manual database
-  intervention)
+* How should we limit the amount of versions handled? In case of an enormous
+  amount of version entries returned from the API, can we make sure to mitigate
+  the issue on our side?
+* Should we find a way to retain versions if they get out of the
+  upgrade-channels API, and if so how. DB can be a solution but so some cache
+  implementations.
 
 ### UI Impact
 
--- a/docs/enhancements/dynamic-fetching-of-openshift-versions.md
+++ b/docs/enhancements/dynamic-fetching-of-openshift-versions.md
@@ -23,28 +23,27 @@ versions. It can be separated to the following cases:
   discovery images that are appropriate for the release.
 * Updating beta releases like EC/FC/RC by following candidate channels like
   ``candidate-4.12``.
-* Updating to the first official stable version when it's GA (Globally
-  Available). For example, updating to 4.12.0 once it's published and available
-  for customers to use.
+* Updating to the first official stable version when it's GA (General
+  Availability). For example, updating to 4.12.0 once it's published and available
+  for users to use.
 * Updating stable releases following stable channels like ``stable-4.12``.
 
 All those actions consist of either manual updates by developers or automatic
 tools like https://gitlab.cee.redhat.com/assisted-installer-ops/assisted-installer-cicd.
 Either manual or automatic, we usually override releases with the new ones (e.g.
-replacing 4.12.0 with 4.12.1) to make sure we don't grow too much our configuration
-files. As per special requests (usually made by customers), we append additional
-versions. For example, see:
-https://gitlab.cee.redhat.com/service/app-interface/-/blob/3fb1ac40a81c7de9bdf1ae6262798fee9336debf/data/services/assisted-installer/cicd/target/production/assisted-service.yaml
+replacing 4.12.0 with 4.12.1) to prevent our configuration files from becoming too
+large. We retain specific versions as per requests, usually made by
+customers/partners.
 
 The idea of this enhancement is configuring the assisted-service to the OpenShift
 channels it needs to follow (e.g. ``candidate-4.12``, ``stable-4.11``), and let
-it periodically fetch versions to provide them to customers.
+it periodically fetch versions to provide them to users.
 
 ## Motivation
 
 It becomes a pretty tedious process to handle each version separately. Some
 use-cases we tend to have:
-* A certain customer use a specific version x.y.z for certification, and we replace
+* A certain user uses a specific version x.y.z for certification, and we replace
   this version with a newer one x.y.z+w (w>=1). In this case we'll append again the
   x.y.z version to the configuration.
 * As described above, we do version updates either automatically or manually.
@@ -111,10 +110,9 @@ in the form of ``https://api.openshift.com/api/upgrades_info/v1/graph?channel=<c
 For example:
 ```
 $ curl -s 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.11&arch=amd64' \
-    | jq -r '.nodes[] | select(.version | startswith("4.11")) | .version'
-4.11.9
-4.11.4
-4.11.8
+    | jq -r '.nodes[] | select(.version | startswith("4.11")) | .version + " => " + .payload'
+4.11.9 => quay.io/openshift-release-dev/ocp-release@sha256:94b611f00f51c9acc44ca3f4634e46bd79d7d28b46047c7e3389d250698f0c99
+4.11.4 => quay.io/openshift-release-dev/ocp-release@sha256:e04ee7eea9e6a6764df218bcce3ee64e1816445158a0763954b599e0fc96580b
 ...
 ```
 
@@ -125,7 +123,7 @@ database.
 
 To help fetching only certain versions via auto-completion like U/X, we should
 provide a parameter ``version_pattern`` for the ``/v2/openshift-versions``
-endpoint. For example:
+endpoint, along with a ``only_latest`` parameter. For example:
 ```
 /v2/openshift-versions:
    get:
@@ -136,7 +134,26 @@ endpoint. For example:
           description: Version pattern for the returned releases. E.g. '4.12'.
           type: string
           required: false
+        - in: query
+          name: only_latest
+          description: It true, returns only the latest version for each minor.
+          type: boolean
+          allowEmptyValue: true
+          default: false
 ```
+
+In addition, we should expose a new API to show the current configuration
+of dynamic releases for debugging and transparency purposes.
+
+We will still need to handle each new minor version (x.y) by:
+* adding RHCOS images for the new version for each supported CPU architecture.
+* adding x.y to the new ``DYNAMIC_OPENSHIFT_RELEASES_CONFIG`` configuration.
+* setting the relevant feature-levels in feature-support-levels list.
+
+This way the assisted-service can support each x.y.z version by fetching a
+matching x.y RHCOS version from the assisted-image-service. The
+assisted-image-service stays up and ready to serve all supported RHCOS images
+and the amount of images is only the bare minimum.
 
 ### User Stories
 
@@ -197,11 +214,10 @@ DYNAMIC_OPENSHIFT_RELEASES_CONFIG='[
 ]'
 ```
 
-### Implementation Details/Notes/Constraints [optional]
-
-Mostly conveyed in ``Risks and Mitigations``.
-
 ### Risks and Mitigations
+
+We assume here that each x.y RHCOS version is backwards and forwards compatible
+with all x.y.z OpenShift versions.
 
 We have recently seen occasions where the disk space is being exhausted due to
 the assisted-service caching a lot of different openshift installer binaries.
@@ -222,11 +238,7 @@ As opposed to that, when a minor release gets EOL (end-of-life) and we remove it
 from the configuration, the service should remove all the stale versions from
 the database.
 
-## Design Details [optional]
-
-If an enhancement is complex enough, design details should be included. When not
-included, reviewers reserve the right to ask for this section to be filled in to
-enable more thoughtful discussion about the enhancement and it's impact.
+## Design Details
 
 ### Open Questions
 
@@ -236,8 +248,8 @@ enable more thoughtful discussion about the enhancement and it's impact.
   doesn't consume too much of it, it might be better.
 * When implementing the binaries LRU cache, should we prioritize the latest
   versions of each y-release? We expect those to be the most common ones, but
-  maybe it's better to let the LRU mechanism determine what's the common versions
-  by itself?
+  maybe it's better to let the LRU mechanism determine what are the common
+  versions by itself?
 * How should we limit database entries? In case of a bug in the code that writes
   way too much version entries, can we make sure to mitigate the issue on our
   side? (without interrupting app-sre for their help on manual database
@@ -266,8 +278,10 @@ versions is where they'll have to change the UI automation.
 
 * It means QE no longer verify that each OpenShift version is installable via
   assisted-installer.
-* We lose a bit of control over which versions we introduce. It means if
-  some existing version has a bug, we don't have a way to exclude it.
+* Current proposition doesn't include a mechanism for excluding specific
+  versions that are dynamically fetched, in case of critical bugs in them.
+  Since it's a low-risk (and low-impact) we can extend it in the future.
+
 
 ## Alternatives
 
--- a/docs/enhancements/dynamic-fetching-of-openshift-versions.md
+++ b/docs/enhancements/dynamic-fetching-of-openshift-versions.md
@@ -17,8 +17,8 @@ configure each of those versions.
 Currently, we have defined processes to support the growing list of OpenShift
 versions. It can be separated to the following cases:
 * Adding a newly released minor version of OpenShift when it's first introduced
-  as part of an EC (Engineering Candidate), FC (Feature Complete) or RC (Release
-  Complete) versions. For example, OpenShift 4.12 had its first appearance as
+  as part of an EC (Engineering Candidate), FC (Feature Candidate) or RC (Release
+  Candidate) versions. For example, OpenShift 4.12 had its first appearance as
   version ``4.12.0-ec.0``. With the new version we usually also add live-iso
   discovery images that are appropriate for the release.
 * Updating beta releases like EC/FC/RC by following candidate channels like
@@ -73,11 +73,10 @@ use-cases we tend to have:
 ## Proposal
 
 The assisted-service will make use of a new configuration variable called
-``DYNAMIC_OPENSHIFT_RELEASES_CONFIG``, which outlines the supported matrix. For
-example:
+``RELEASE_SOURCES``, which outlines the supported matrix. For example:
 
 ```
-DYNAMIC_OPENSHIFT_RELEASES_CONFIG='{
+RELEASE_SOURCES='{
    "4.11": {
       "upgrade_channels": [
          {
@@ -147,7 +146,7 @@ of dynamic releases for debugging and transparency purposes.
 
 We will still need to handle each new minor version (x.y) by:
 * adding RHCOS images for the new version for each supported CPU architecture.
-* adding x.y to the new ``DYNAMIC_OPENSHIFT_RELEASES_CONFIG`` configuration.
+* adding x.y to the new ``RELEASE_SOURCES`` configuration.
 * setting the relevant feature-levels in feature-support-levels list.
 
 This way the assisted-service can support each x.y.z version by fetching a
@@ -188,7 +187,7 @@ RELEASE_IMAGES='[
   }
 ]'
 
-DYNAMIC_OPENSHIFT_RELEASES_CONFIG='[
+RELEASE_SOURCES='[
   {
     "openshift_version": "4.11",
     "upgrade_channels": [
--- a/docs/enhancements/dynamic-fetching-of-openshift-versions.md
+++ b/docs/enhancements/dynamic-fetching-of-openshift-versions.md
@@ -1,0 +1,280 @@
+---
+title: dynamic-fetching-of-openshift-versions
+authors:
+  - "@osherdp"
+creation-date: 2023-01-12
+last-updated: 2023-01-12
+---
+
+# Dynamic Fetching of OpenShift Versions
+
+Letting users of Assisted Installer API and UI to choose an OpenShift version
+out of a variety of supported versions, all without the need to specifically
+configure each of those versions.
+
+## Summary
+
+Currently, we have defined processes to support the growing list of OpenShift
+versions. It can be separated to the following cases:
+* Adding a newly released minor version of OpenShift when it's first introduced
+  as part of an EC (Engineering Candidate), FC (Feature Complete) or RC (Release
+  Complete) versions. For example, OpenShift 4.12 had its first appearance as
+  version ``4.12.0-ec.0``. With the new version we usually also add live-iso
+  discovery images that are appropriate for the release.
+* Updating beta releases like EC/FC/RC by following candidate channels like
+  ``candidate-4.12``.
+* Updating to the first official stable version when it's GA (Globally
+  Available). For example, updating to 4.12.0 once it's published and available
+  for customers to use.
+* Updating stable releases following stable channels like ``stable-4.12``.
+
+All those actions consist of either manual updates by developers or automatic
+tools like https://gitlab.cee.redhat.com/assisted-installer-ops/assisted-installer-cicd.
+Either manual or automatic, we usually override releases with the new ones (e.g.
+replacing 4.12.0 with 4.12.1) to make sure we don't grow too much our configuration
+files. As per special requests (usually made by customers), we append additional
+versions. For example, see:
+https://gitlab.cee.redhat.com/service/app-interface/-/blob/3fb1ac40a81c7de9bdf1ae6262798fee9336debf/data/services/assisted-installer/cicd/target/production/assisted-service.yaml
+
+The idea of this enhancement is configuring the assisted-service to the OpenShift
+channels it needs to follow (e.g. ``candidate-4.12``, ``stable-4.11``), and let
+it periodically fetch versions to provide them to customers.
+
+## Motivation
+
+It becomes a pretty tedious process to handle each version separately. Some
+use-cases we tend to have:
+* A certain customer use a specific version x.y.z for certification, and we replace
+  this version with a newer one x.y.z+w (w>=1). In this case we'll append again the
+  x.y.z version to the configuration.
+* As described above, we do version updates either automatically or manually.
+* QE team wants to check if a newly added OpenShift version has a regression in its
+  behavior from a previous version.
+* QE team tests a flow which includes OpenShift upgrade process.
+
+### Goals
+
+* Assisted-service is able to provide all versions included in the upgrade-channels,
+  based on provided configuration.
+* We keep ourselves the option to specifically configure certain versions as we can
+  nowadays. This is to make sure we can handle edge-cases and be flexible if needed.
+* UI needs to be adjusted to handle this growing list of versions (at the time being,
+  a total of ~140 versions).
+
+### Non-Goals
+
+* Dynamic handling of y-versions, live-iso discovery images, and such. A new minor
+  OpenShift version is introduced every 4 months, requires additional code changes
+  (for example in feature-support-level list) and it isn't worth the costs and
+  implications of making assisted-image-service refresh itself and pull-in new
+  images.
+* Handling OpenShift versions in a kubernetes API interface. For this use-case
+  versions are externally configured in the ``ClusterImageSet`` custom resource.
+
+## Proposal
+
+The assisted-service will make use of a new configuration variable called
+``DYNAMIC_OPENSHIFT_RELEASES_CONFIG``, which outlines the supported matrix. For
+example:
+
+```
+DYNAMIC_OPENSHIFT_RELEASES_CONFIG='{
+   "4.11": {
+      "upgrade_channels": [
+         {
+            "cpu_architecture": "x86_64",
+            "channels": ["stable"]
+         }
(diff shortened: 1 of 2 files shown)


osherdp:
/hold for getting all the feedback
/cc @filanov @gamli75 @danielerez @avishayt  @carbonin 
(please suggest more reviewers as seems fit)

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+    ]
+  }
+]'
+```
+
+### Implementation Details/Notes/Constraints [optional]
+
+Mostly conveyed in ``Risks and Mitigations``.
+
+### Risks and Mitigations
```
gamli75, line 204: please add a case when we will get a new major release from the API, for example: 4.13 - we need to add it also to the image service repo

gamli75: please add a few words about the connection between assisted service versions and image service versions

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+* When implementing the binaries LRU cache, should we prioritize the latest
+  versions of each y-release? We expect those to be the most common ones, but
+  maybe it's better to let the LRU mechanism determine what's the common versions
+  by itself?
+* How should we limit database entries? In case of a bug in the code that writes
+  way too much version entries, can we make sure to mitigate the issue on our
+  side? (without interrupting app-sre for their help on manual database
+  intervention)
+
+### UI Impact
```
gamli75, line 246: I think that the default response for the openshift-versions API should include the default(latest) version per OCP release and we should add a filter to this API with option to send ALL versions otherwise it will take lots of time to load list of 180 versions + 
@nirfarkas @jkilzi WDYT?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+from the configuration, the service should remove all the stale versions from
+the database.
+
+## Design Details [optional]
+
+If an enhancement is complex enough, design details should be included. When not
+included, reviewers reserve the right to ask for this section to be filled in to
+enable more thoughtful discussion about the enhancement and it's impact.
+
+### Open Questions
```
romfreiman, line 231: inst our main limitation is that the image service downloads all the images upon restart? so every release we'll have to restart the service?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+            "channels": ["candidate", "stable"]
+         },
+         {
+            "cpu_architecture": "multi",
+            "channels": ["candidate", "stable"]
+         }
+      ]
+   }
+}'
+```
```
danielerez, line 107: Do we need some API to fetch this list? Maybe so the user could ensure it's configured properly.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+  versions are externally configured in the ``ClusterImageSet`` custom resource.
+
+## Proposal
+
+The assisted-service will make use of a new configuration variable called
+``DYNAMIC_OPENSHIFT_RELEASES_CONFIG``, which outlines the supported matrix. For
+example:
+
+```
+DYNAMIC_OPENSHIFT_RELEASES_CONFIG='{
```
danielerez, line 81: Do we need the "upgrade_channels" so the config could be extended later on? If not, maybe we can call the var something like 'OCP_RELEASES_UPGRADE_CHANNELS'. I.e. a map from 'version' to array.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+         {
+            "cpu_architecture": "multi",
+            "channels": ["candidate", "stable"]
+         }
+      ]
+   }
+}'
+```
+
+The assisted-service periodically polls versions from the ``upgrades_info`` API
```
danielerez, line 109: Is there some impl details for that? Worth discussing exactly how/where it should be polled.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+```
+$ curl -s 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.11&arch=amd64' \
+    | jq -r '.nodes[] | select(.version | startswith("4.11")) | .version'
+4.11.9
+4.11.4
+4.11.8
+...
+```
+
+It then adds any new version found by this process in the database. When a
```
danielerez, line 121: Not sure we actually need to store it in DB. If we fetch periodically anyway, maybe we can just cache the result for a while.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+To help fetching only certain versions via auto-completion like U/X, we should
+provide a parameter ``version_pattern`` for the ``/v2/openshift-versions``
+endpoint. For example:
+```
+/v2/openshift-versions:
+   get:
+      ...
+      parameters:
+        - in: query
+          name: version_pattern
```
danielerez, line 135: Should we support multiple 'version_pattern'? E.g. So the UI could fetch all vers in a single request.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Design Details [optional]
+
+If an enhancement is complex enough, design details should be included. When not
+included, reviewers reserve the right to ask for this section to be filled in to
+enable more thoughtful discussion about the enhancement and it's impact.
+
+### Open Questions
+
+* Is the database usage a real necessity? The idea is having the periodic thread
```
danielerez, line 233: Exactly. If we're gonna fetch periodically anyway, maybe we can just cache the result for a while.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+  updating the database, and then the HTTP server is able to filter versions from
+  the database at its leisure. If there's a good way to do it in-memory and it
+  doesn't consume too much of it, it might be better.
+* When implementing the binaries LRU cache, should we prioritize the latest
+  versions of each y-release? We expect those to be the most common ones, but
+  maybe it's better to let the LRU mechanism determine what's the common versions
+  by itself?
```
danielerez, line 240: Maybe we can use the [default](https://github.com/openshift/assisted-service/blob/58b7b4b0fd8e5a43c4b27ab28086c896fa636f27/swagger.yaml#L6382) flag for each y-stream. E.g. rely on it being defined in RELEASE_IMAGES.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+versions. It can be separated to the following cases:
+* Adding a newly released minor version of OpenShift when it's first introduced
+  as part of an EC (Engineering Candidate), FC (Feature Complete) or RC (Release
+  Complete) versions. For example, OpenShift 4.12 had its first appearance as
+  version ``4.12.0-ec.0``. With the new version we usually also add live-iso
+  discovery images that are appropriate for the release.
+* Updating beta releases like EC/FC/RC by following candidate channels like
+  ``candidate-4.12``.
+* Updating to the first official stable version when it's GA (Globally
+  Available). For example, updating to 4.12.0 once it's published and available
```
avishayt, line 27:
```suggestion
* Updating to the first official stable version when it's GA (General
  Availability). For example, updating to 4.12.0 once it's published and available
```

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+  for customers to use.
+* Updating stable releases following stable channels like ``stable-4.12``.
+
+All those actions consist of either manual updates by developers or automatic
+tools like https://gitlab.cee.redhat.com/assisted-installer-ops/assisted-installer-cicd.
+Either manual or automatic, we usually override releases with the new ones (e.g.
+replacing 4.12.0 with 4.12.1) to make sure we don't grow too much our configuration
+files. As per special requests (usually made by customers), we append additional
+versions. For example, see:
```
avishayt, line 36:
```suggestion
replacing 4.12.0 with 4.12.1) to prevent our configuration files from becoming too
large. We retain specific versions as per special requests (usually made by
customers/partners). For example, see:
```

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+            "cpu_architecture": "multi",
+            "channels": ["candidate", "stable"]
+         }
+      ]
+   }
+}'
+```
+
+The assisted-service periodically polls versions from the ``upgrades_info`` API
+in the form of ``https://api.openshift.com/api/upgrades_info/v1/graph?channel=<channel>&arch=<arch>``.
```
avishayt, line 110: We should make the prefix of the URL configurable also

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+  for customers to use.
+* Updating stable releases following stable channels like ``stable-4.12``.
+
+All those actions consist of either manual updates by developers or automatic
+tools like https://gitlab.cee.redhat.com/assisted-installer-ops/assisted-installer-cicd.
+Either manual or automatic, we usually override releases with the new ones (e.g.
+replacing 4.12.0 with 4.12.1) to make sure we don't grow too much our configuration
+files. As per special requests (usually made by customers), we append additional
+versions. For example, see:
+https://gitlab.cee.redhat.com/service/app-interface/-/blob/3fb1ac40a81c7de9bdf1ae6262798fee9336debf/data/services/assisted-installer/cicd/target/production/assisted-service.yaml
```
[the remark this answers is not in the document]
osherdp: but it's not an example for specifying x.y.z version for customers
in ``data/default_release_images.json`` we don't have an example for that

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+
+The idea of this enhancement is configuring the assisted-service to the OpenShift
```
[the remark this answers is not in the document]
osherdp: done!

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+  this version with a newer one x.y.z+w (w>=1). In this case we'll append again the
+  x.y.z version to the configuration.
+* As described above, we do version updates either automatically or manually.
+* QE team wants to check if a newly added OpenShift version has a regression in its
+  behavior from a previous version.
+* QE team tests a flow which includes OpenShift upgrade process.
+
+### Goals
+
+* Assisted-service is able to provide all versions included in the upgrade-channels,
```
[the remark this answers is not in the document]
osherdp: * You mean only adding the X latest versions for a channel?
* See below, currently the suggestion is to only include items in this API (and not exclude everything that misses from there). So by default it will always 'lock' a version in the list

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,280 @@
...
+  based on provided configuration.
+* We keep ourselves the option to specifically configure certain versions as we can
```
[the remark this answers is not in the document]
osherdp: Correct

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: yes it's for future extensions
I thought maybe we'll want to add more configuration in the future to control retention per-version

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: nice idea. added

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: it's debatable, but I thought doing that every hour with the existing threads mechanism we have (I think we're using it for the monitor)
wdyt?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: the part of ``https://api.openshift.com``?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: they can do it by not providing the ``version_pattern`` parameter (or having it empty)

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: added

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: in this suggestion the image service stays static with what images it provides (that are x.y images for each supported cpu-arch)

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: will be discussed in that other thread below

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: will it work well when being done from different threads?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: it is currently a global default (the one version out of all versions to be displayed by default)

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: Done!

docs/enhancements/dynamic-fetching-of-openshift-versions.md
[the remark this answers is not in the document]
danielerez: Yeah, meant we could have a configuration for the number of versions to include. But can be handled later if needed.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
danielerez: Sure can be discussed later, sounds like a good option to use our monitor infra. Though if we decide to just cache the list without using db, may worth looking into [gocron](https://github.com/go-co-op/gocron).

docs/enhancements/dynamic-fetching-of-openshift-versions.md
danielerez: I think [go-cache](https://github.com/patrickmn/go-cache) we usually use is thread safe, but we can also try just using [gocron](https://github.com/go-co-op/gocron) instead of our monitor.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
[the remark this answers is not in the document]
carbonin: Still we probably shouldn't have non-public information here.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
romfreiman: so we'll use outdated live iso?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,301 @@
...
+   }
+}'
+```
+
+The assisted-service periodically polls versions from the ``upgrades_info`` API
+in the form of ``https://api.openshift.com/api/upgrades_info/v1/graph?channel=<channel>&arch=<arch>``.
+For example:
+```
+$ curl -s 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.11&arch=amd64' \
+    | jq -r '.nodes[] | select(.version | startswith("4.11")) | .version'
```
carbonin, line 114: An example that includes the payload may be more useful here as that's what we're really interested in.
Something like:

```
$ curl -s 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.11&arch=amd64' | jq -r '.nodes[] | .version + " => " + .payload'
4.10.9 => quay.io/openshift-release-dev/ocp-release@sha256:39f360002b9b5c730d1167879ad6437352d51e72acc9fe80add3ec2a0d20400d
4.10.35 => quay.io/openshift-release-dev/ocp-release@sha256:766bc6b381163ecccfea00556dd336dd49067413f69cbc29337fea778295c7bb
4.10.23 => quay.io/openshift-release-dev/ocp-release@sha256:e40e49d722cb36a95fa1c03002942b967ccbd7d68de10e003f0baa69abad457b
4.11.4 => quay.io/openshift-release-dev/ocp-release@sha256:e04ee7eea9e6a6764df218bcce3ee64e1816445158a0763954b599e0fc96580b
4.11.8 => quay.io/openshift-release-dev/ocp-release@sha256:6499bc69a0707fcad481c3cb73226c364586761c45b867dc31b345c6e6204e28
```

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,301 @@
...
+* adding RHCOS images for the new version for each supported CPU architecture.
+* adding x.y to the new ``DYNAMIC_OPENSHIFT_RELEASES_CONFIG`` configuration.
+* setting the relevant feature-levels in feature-support-levels list.
+
+This way the assisted-service can support each x.y.z version by fetching a
+matching x.y RHCOS version from the assisted-image-service. The
+assisted-image-service stays up and ready to serve all supported RHCOS images
+and the amount of images is only the bare minimum.
+We assume here that each x.y RHCOS version is backwards and forwards compatible
+with all x.y.z OpenShift versions.
```
carbonin, line 160: I think it may be worth highlighting this somewhere more visible. Maybe in `Implementation Details/Notes/Constraints`?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,301 @@
...
+add versions and not mirror exactly the given result of those API responses.
+As opposed to that, when a minor release gets EOL (end-of-life) and we remove it
+from the configuration, the service should remove all the stale versions from
+the database.
+
+## Design Details [optional]
+
+If an enhancement is complex enough, design details should be included. When not
+included, reviewers reserve the right to ask for this section to be filled in to
+enable more thoughtful discussion about the enhancement and it's impact.
```
carbonin, line 250: You can remove the template text here :sweat_smile:

docs/enhancements/dynamic-fetching-of-openshift-versions.md
carbonin: I don't think we need polling here. We should be able to look up new versions when the user asks for one we don't know about.

carbonin: We already have a cache for versions in `internal/versions/versions.go` that should just continue to be the cache.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,301 @@
...
+
+* Is the database usage a real necessity? The idea is having the periodic thread
+  updating the database, and then the HTTP server is able to filter versions from
+  the database at its leisure. If there's a good way to do it in-memory and it
+  doesn't consume too much of it, it might be better.
+* When implementing the binaries LRU cache, should we prioritize the latest
+  versions of each y-release? We expect those to be the most common ones, but
+  maybe it's better to let the LRU mechanism determine what are the common
+  versions by itself?
+* How should we limit database entries? In case of a bug in the code that writes
```
carbonin, line 262: I we do go the route of putting this in the database, how would we delete entries we actually don't want to support anymore?

I think the real question about the DB is if we're going to be issuing too many requests to this versions service, but I think it's used directly by OCP clusters right? If that's the case we'd be a drop in the bucket asking a few times an hour.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,301 @@
...
+  way too much version entries, can we make sure to mitigate the issue on our
+  side? (without interrupting app-sre for their help on manual database
+  intervention)
+
+### UI Impact
+
+The UI should prioritize the latest versions when displaying the different
+options. Only in need to customize the version (meaning installing an older one),
+then it requires clicking on 'Show all available versions' and searching for the
```
carbonin, line 271: We should probably get UX involved earlier rather than later on this.

@nirfarkas

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,301 @@
...
+QE test-plan should be fairly simple on the API part - just trying to install
+some arbitrary "older" version. When UI gets adjusted to first show latest
+versions is where they'll have to change the UI automation.
+
+## Drawbacks
+
+* It means QE no longer verify that each OpenShift version is installable via
+  assisted-installer.
+* We lose a bit of control over which versions we introduce. It means if
+  some existing version has a bug, we don't have a way to exclude it.
```
carbonin, line 291: This seems like something we could handle though, right?
Maybe not necessary on the first path, but excluding some list of specific releases should be easy enough.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: no, we have an existing process to update rhcos images from mirror.openshift.com

docs/enhancements/dynamic-fetching-of-openshift-versions.md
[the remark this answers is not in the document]
osherdp: I'll remove the example so we can be left with no examples :smile:

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: > I don't think we need polling here. We should be able to look up new versions when the user asks for one we don't know about.

@carbonin the suggested U/X is to provide the user with available versions. It means deciding on that list on the assisted-service level
do you mean having the UI gather all the versions and let the client-side set a version for a cluster? (I just want to make sure I understand your suggestion)

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: It exists for staging and integration as well, so it does make sense to have the ``https://api.openshift.com`` configurable if we'll decide to match it with the environment
As for the rest of the URL, it makes sense pin the v1 API. I wouldn't expect it to change more than that

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: They already gave their input and this part describes it

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: done!

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: Done!

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: Right, we can easily add an exclude-list after finding out that a version has a bug
I'll update the doc

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: Actually it's ``OCM_BASE_URL``, so we can re-use it

docs/enhancements/dynamic-fetching-of-openshift-versions.md
danielerez: +1 on that. We can also just invoke the query on cache expire. I.e. if the value in cache is [stale](https://github.com/patrickmn/go-cache/blob/master/cache.go#L128), fetch the list from api.openshift

osherdp: I'm on the fence here, because it is much simpler but has a major disadvantage. If this logic happens in the same thread of the http server, some requests can stall with their response for too long

I need to benchmark it a bit to understand if it's really substantial on the nominal case
Either way remember it's a dependency on an external API that can also be down / have an outage / not perform well. We wouldn't want to wait too long for its responses to return

danielerez: We use a similar mechanism when calling AMS, so I think it should be safe enough. E.g. https://github.com/openshift/assisted-service/blob/master/pkg/auth/rhsso_authz_handler.go
There's also the fallback to the current release images list, right? So in case of an error/timeout we'd just return the fallback list.

carbonin: Indeed we should be able to configure a client-side timeout.
We will want to ensure we protect the in-memory release image list with a locking mechanism though.

I had considered this when doing the versions refactor, but we'll definitely need it now.

osherdp: Ack. I'll change to in-memory implementation and mention caveats of this solution in the mitigations & risks section

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: ack. going for in-memory no-additional-thread for now but I'll keep this in mind if we'll change our minds

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Open Questions
+
+* When implementing the binaries LRU cache, should we prioritize the latest
+  versions of each y-release? We expect those to be the most common ones, but
+  maybe it's better to let the LRU mechanism determine what are the common
+  versions by itself?
+* How should we limit the amount of versions handled? In case of an enormous
+  amount of version entries returned from the API, can we make sure to mitigate
+  the issue on our side?
```
carbonin, line 248: Not sure I understand this. Limit in what way? Isn't the idea behind this proposal that we don't limit what is available? What actual problem is this referencing?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,289 @@
...
+* Should we find a way to retain versions if they get out of the
+  upgrade-channels API, and if so how. DB can be a solution but so some cache
+  implementations.
```
carbonin, line 251: Is this not the reasoning for keeping the existing `RELEASE_IMAGES` variable?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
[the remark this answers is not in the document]
osherdp: it's being explained in the next sentence
just for maximal flexibility and for handling some edge-cases

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: limit the amount of versions, for example by taking only the latest X versions in a stream
everything that has the potential to grow too much might cause issues in some ways, but just as an example it might consume too much memory, cause performance issues, etc

We need to see how it goes in the future, some things are hard to foresee

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: It's only one edge-case that can be handled by it (but there's a lot more)
Another one is a release that didn't get into any channel yet (happened with 4.12 EC versions at first)

docs/enhancements/dynamic-fetching-of-openshift-versions.md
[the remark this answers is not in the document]
carbonin: Do we know what those edge cases are? That doesn't actually explain any real scenario we might encounter.

I'm imagining us in the future asking why this is still here and not being able to find a reason why.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
carbonin: Sure, but open questions are meant to be answered :sweat_smile: 

How many versions would we end up with if we adopted this approach today? Maybe try to configure all of them manually with the process we currently have and see if we run into any issues?

docs/enhancements/dynamic-fetching-of-openshift-versions.md
carbonin: Can you add these examples as potential reasons in by previous question about why we're keeping `RELEASE_IMAGES`. Then you can just remove this question since we have a solution for it.

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: Yeah, I did this calculation. See here: https://github.com/openshift/assisted-service/blob/4b4c27b5b99a3f1639f37cb81fb9ccd92a112f5f/docs/enhancements/dynamic-fetching-of-openshift-versions.md#L61

I can give the manual test a go, but my original concern is not about the *current* amount of images returned from the API but rather a potentially large response we're getting from it *in the future*. Do you understand? :)

docs/enhancements/dynamic-fetching-of-openshift-versions.md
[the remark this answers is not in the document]
osherdp: I'll mention some cases on "risks and mitigations"

docs/enhancements/dynamic-fetching-of-openshift-versions.md
carbonin: Sure, but is this something you want to address? Otherwise I think it should probably go into the risks/mitigations section.

osherdp: No, I think it's a minor one
I'll add it to the risks/mitigations section

docs/enhancements/dynamic-fetching-of-openshift-versions.md
osherdp: Done!

carbonin:
Looks good to me.

Any other feedback or are we good to merge?

osherdp:
@gamli75 @filanov @danielerez @avishayt anything else here?

danielerez:
> @gamli75 @filanov @danielerez @avishayt anything else here?

Looks good to me.

gamli75:
Looks good to me.

gamli75 (approved):

docs/enhancements/dynamic-fetching-of-openshift-versions.md
```diff
@@ -0,0 +1,290 @@
...
+          name: version_pattern
+          description: Version pattern for the returned releases. E.g. '4.12'.
+          type: string
+          required: false
+        - in: query
+          name: only_latest
+          description: It true, returns only the latest version for each minor.
+          type: boolean
+          allowEmptyValue: true
+          default: false
```
osherdp, line 142: @avishayt @danielerez do you think it should be ``true`` by default?
otherwise it returns a big list that might be a bit weird for some API consumers

danielerez: Yeah, changing the current behaviour of the API might be confusing for existing users.
What about introducing instead a property for getting all versions? Something like 'include_all_minor' boolean.
From the UI perspective, it seems that only two variants will be used: the old behaviour (only latest) and version_pattern. So getting all minor versions seems like a rare option that deserves a separate special param.

osherdp: ``include_all_minor`` is the same as having the proposed ``only_latest``, no?

danielerez: > `include_all_minor` is the same as having the proposed `only_latest`, no?

Yeah, just the opposite :) Not crucial, just sounds more straightforward than 'only_latest=false'

osherdp: hmm, by the way [minor is the y part](https://semver.org/#summary) so do you mean ``include_all_patch_versions``?

danielerez: > hmm, by the way [minor is the y part](https://semver.org/#summary) so do you mean `include_all_patch_versions`?

right, tried to shorten 'include all for minor', naming is hard:) maybe can be just 'all_versions' and explain in the description what is returned exactly.

osherdp: Sounds good
Thanks!

PR #4900, opened 2023-01-16
MGMT-12405: Enclose API endpoint with brackets if IPv6 address

mkowalski:
This PR fixes the issue with returning the ignition endpoint in the form of an IPv6 address by adding missing brackets around the IPv6 address. This is required in order to form a correct HTTP URI which requires brackets to make IPv6 address distinguishable from the port.

Fixes: [MGMT-12405](https://issues.redhat.com//browse/MGMT-12405)

/cc @omertuc


mkowalski:
/retest e2e-agent-compact

Not my error

PR #4906, opened 2023-01-17
MGMT-13313: bminventory: throw Bad Request error on arm64 Nutanix clusters

vrutkovs:
Block platform=nutanix when ARM64 CPU arch is set

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -445,6 +445,9 @@ func (b *bareMetalInventory) validateRegisterClusterInternalParams(params *insta
 		if err := validations.ValidateHighAvailabilityModeWithPlatform(params.NewClusterParams.HighAvailabilityMode, params.NewClusterParams.Platform); err != nil {
 			return common.NewApiError(http.StatusBadRequest, err)
 		}
+		if err := validations.ValidateArchitectureWithPlatform(&params.NewClusterParams.CPUArchitecture, params.NewClusterParams.Platform); err != nil {
+			return common.NewApiError(http.StatusBadRequest, err)
+		}
 	}
 
 	return nil
@@ -1915,6 +1918,10 @@ func (b *bareMetalInventory) v2UpdateClusterInternal(ctx context.Context, params
 	}
 
 	if err = validations.ValidateHighAvailabilityModeWithPlatform(cluster.HighAvailabilityMode, platform); err != nil {
+		return nil, common.NewApiError(http.StatusBadRequest, err)
+	}
+
+	if err = validations.ValidateArchitectureWithPlatform(&cluster.CPUArchitecture, platform); err != nil {
 		return nil, common.NewApiError(http.StatusBadRequest, err)
 	}
 
--- a/internal/cluster/validations/validations.go
+++ b/internal/cluster/validations/validations.go
@@ -931,3 +931,13 @@ func ValidateIgnitionImageSize(config string) error {
 
 	return nil
 }
+
+func ValidateArchitectureWithPlatform(architecture *string, platform *models.Platform) error {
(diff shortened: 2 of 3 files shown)


omertuc:
We should also handle the PATCH call mentioned in the ticket. IIUC this PR only handles the cluster registration

vrutkovs:
>We should also handle the PATCH call

Good idea, added this + unit test

internal/bminventory/inventory.go
```diff
@@ -795,6 +795,10 @@ func (b *bareMetalInventory) getNewClusterCPUArchitecture(newClusterParams *mode
 		return "", errors.Errorf("Non x86_64 CPU architectures for version %s are supported only with User Managed Networking", swag.StringValue(newClusterParams.OpenshiftVersion))
 	}
 
+	if getPlatformType(newClusterParams.Platform) == string(models.PlatformTypeNutanix) && newClusterParams.CPUArchitecture == common.ARM64CPUArchitecture {
```
omertuc, line 798: Is using `isArm64AllowedForPlatform` here possible?

vrutkovs: Rework using @slaviered's suggestion as `validations.ValidateArchitectureWithPlatform` in DRY spirit

internal/cluster/validations/validations.go
```diff
@@ -931,3 +931,13 @@ func ValidateIgnitionImageSize(config string) error {
 
 	return nil
 }
+
+func ValidateArchitectureWithPlatform(architecture *string, platform *models.Platform) error {
+	if swag.StringValue(architecture) == common.ARM64CPUArchitecture {
```
osherdp, line 936: would we want to do the same for ``s390x`` and ``ppc64le``? I guess they are out of question for Nutanix as well
or rather check positively that the architecture is x86

either approaches (positive & negative checks) have pros & cons, so I don't have a strong preference for this or that

osherdp (changes requested):

internal/cluster/validations/validations.go
vrutkovs: We don't provide these payloads in AI, do we?

I think it would be safer to limit this to x86_64, let me clarify that with Nutanix

osherdp: > We don't provide these payloads in AI, do we?

I think it's quite new that we do. Maybe 1 week ago it were all added, so I expect to see those alongside arm and x86

vrutkovs: Clarifying on the best way to approach this with @gamli75

vrutkovs: Reworked it to deny any non-x86_64 architecture, PTAL

internal/cluster/validations/validations.go
```diff
@@ -931,3 +931,13 @@ func ValidateIgnitionImageSize(config string) error {
 
 	return nil
 }
+
+func ValidateArchitectureWithPlatform(architecture *string, platform *models.Platform) error {
+  if platform != nil && platform.Type != nil && *platform.Type == models.PlatformTypeNutanix {
+	  if swag.StringValue(architecture) != common.X86CPUArchitecture {
+			return errors.New("non-x86_64 CPU architectures are not supported on Nutanix clusters")
```
osherdp, line 938: linter might yell here

osherdp (approved):

osherdp (approved):

vrutkovs:
Needs another LGTM - unit tests and error messaged needed update

gamli75 (approved):

PR #4907, opened 2023-01-17
MGMT-13204: Implement skipping of host and cluster validations

paul-maidment:
Two cluster level fields have been introduced "SkippedHostValidations" and "SkippedClusterValidations"

Each field will hold a JSON encoded list of strings with the names of the validations to be skipped or "all"

If the "all" option is chosen then all validations, including operator validations will not execute. If specific validations are entered then just these will be skipped.

- [x] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -77,6 +77,31 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+var NonIgnorableHostValidations []string = []string{"connected", "has-inventory", "machine-cidr-defined", "hostname-unique", "hostname-valid"}
+var NonIgnorableClusterValidations []string = []string{"api-vips-defined", "ingress-vips-defined", "all-hosts-are-ready-to-install", "sufficient-masters-count", "pull-secret-set", "cluster-preparation-succeeded"}
+
+var MayIgnoreValidation = func(validationID string, nonIgnorables []string) bool {
+	if validationID == "all" {
+		return true
+	}
+	return !swag.ContainsStrings(nonIgnorables, validationID)
+}
+
+var MayIgnoreValidations = func(validationIDs []string, nonIgnorables []string) (bool, string) {
+	result := true
+	cantBeIgnored := []string{}
+	for _, validation := range validationIDs {
+		if validation == "all" {
+			return true, ""
+		}
+		if !MayIgnoreValidation(validation, nonIgnorables) {
+			cantBeIgnored = append(cantBeIgnored, validation)
+			result = false
+		}
+	}
+	return result, strings.Join(cantBeIgnored, ",")
+}
+
 const DefaultUser = "kubeadmin"
 
 const WindowBetweenRequestsInSeconds = 10 * time.Second
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -77,31 +77,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
-var NonIgnorableHostValidations []string = []string{"connected", "has-inventory", "machine-cidr-defined", "hostname-unique", "hostname-valid"}
-var NonIgnorableClusterValidations []string = []string{"api-vips-defined", "ingress-vips-defined", "all-hosts-are-ready-to-install", "sufficient-masters-count", "pull-secret-set", "cluster-preparation-succeeded"}
-
-var MayIgnoreValidation = func(validationID string, nonIgnorables []string) bool {
-	if validationID == "all" {
-		return true
-	}
-	return !swag.ContainsStrings(nonIgnorables, validationID)
-}
-
-var MayIgnoreValidations = func(validationIDs []string, nonIgnorables []string) (bool, string) {
-	result := true
-	cantBeIgnored := []string{}
-	for _, validation := range validationIDs {
-		if validation == "all" {
-			return true, ""
-		}
-		if !MayIgnoreValidation(validation, nonIgnorables) {
-			cantBeIgnored = append(cantBeIgnored, validation)
-			result = false
-		}
-	}
-	return result, strings.Join(cantBeIgnored, ",")
-}
-
 const DefaultUser = "kubeadmin"
 
 const WindowBetweenRequestsInSeconds = 10 * time.Second
@@ -1245,10 +1220,17 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params 
 				autoAssigned = autoAssigned || updated
 			}
 		}
+		hasIgnoredValidations := common.IgnoredValidationsAreSet(cluster)
+		if hasIgnoredValidations {
+			eventgen.SendValidationsIgnoredEvent(ctx, b.eventsHandler, *cluster.ID)
+		}
 		//usage for auto role selection is measured only for day1 clusters with more than
 		//3 hosts (which would automatically be assigned as masters if the hw is sufficient)
 		if usages, u_err := usage.Unmarshal(cluster.Cluster.FeatureUsage); u_err == nil {
 			report := cluster.Cluster.TotalHostCount > common.MinMasterHostsNeededForInstallation && autoAssigned
+			if hasIgnoredValidations {
+				b.setUsage(true, usage.ValidationsIgnored, nil, usages)
+			}
 			b.setUsage(report, usage.AutoAssignRoleUsage, nil, usages)
 			b.usageApi.Save(tx, *cluster.ID, usages)
 		}
@@ -1476,7 +1458,6 @@ func (b *bareMetalInventory) V2InstallHost(ctx context.Context, params installer
 		log.Errorf("Failed to move host %s to installing", h.RequestedHostname)
 		return common.GenerateErrorResponder(err)
 	}
-
 	return installer.NewV2InstallHostAccepted().WithPayload(h)
 }
 
--- a/internal/cluster/refresh_status_preprocessor.go
+++ b/internal/cluster/refresh_status_preprocessor.go
@@ -2,14 +2,17 @@ package cluster
 
 import (
 	"context"
+	"fmt"
 	"sort"
 	"strings"
 
 	"github.com/go-openapi/swag"
+	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/host"
 	"github.com/openshift/assisted-service/internal/operators"
 	"github.com/openshift/assisted-service/internal/operators/api"
 	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
 )
@@ -53,6 +56,15 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
 	checkValidationsInStatuses := []string{
 		models.ClusterStatusInsufficient, models.ClusterStatusReady, models.ClusterStatusPendingForInput, models.ClusterStatusPreparingForInstallation,
 	}
+	var ignoredValidations []string
+	var err error
+	if c.cluster != nil {
+		ignoredValidations, err = common.DeserializeJSONList(c.cluster.IgnoredClusterValidations)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, fmt.Sprintf("Unable to deserialize ignored cluster validations for cluster %s", string(*c.cluster.ID)))
+		}
+	}
+
 	//if the cluster is not on discovery stages - skip the validations check
 	if !funk.ContainsString(checkValidationsInStatuses, swag.StringValue(c.cluster.Status)) {
 		return stateMachineInput, validationsOutput, nil
@@ -60,9 +72,10 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
 	for _, v := range r.validations {
 		st, message := v.condition(c)
 		stateMachineInput[v.id.String()] = st == ValidationSuccess
-		category, err := v.id.Category()
+		var category string
+		category, err = v.id.Category()
 		if err != nil {
-			logrus.WithError(err).Warn("id.category()")
+			r.log.WithError(err).Warn("id.category()")
 			return nil, nil, err
 		}
 		validationsOutput[category] = append(validationsOutput[category], ValidationResult{
@@ -79,9 +92,10 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
 	for _, result := range results {
 		stateMachineInput[result.ValidationId] = result.Status == api.Success
 		id := ValidationID(result.ValidationId)
+
 		category, err := id.Category()
 		if err != nil {
-			logrus.WithError(err).Warn("id.category()")
+			r.log.WithError(err).Warn("id.category()")
 			return nil, nil, err
 		}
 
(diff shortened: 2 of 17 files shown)


internal/cluster/transition.go
```diff
@@ -600,6 +628,22 @@ func (th *transitionHandler) PostRefreshCluster(reason string) stateswitch.PostT
...
 		}
+
+		// If the cluster just transitioned into the installing state then check whether or not we have ignored any validations.
+		if sCluster.srcState != swag.StringValue(sCluster.cluster.Status) && swag.StringValue(sCluster.cluster.Status) == models.ClusterStatusInstalling {
+			if HasIgnoredValidations(*sCluster.cluster) {
+				// Send an event to indicate that validations are being ignored.
+				eventgen.SendValidationsIgnoredEvent(params.ctx, th.eventsHandler, *sCluster.cluster.ID)
+				// Set a feature usage flag to indicate that validations are being ignored.
+				if usages, u_err := usage.Unmarshal(sCluster.cluster.FeatureUsage); u_err == nil {
+					th.setUsage(true, usage.ValidationsIgnored, nil, usages)
```
ori-amizur, line 639: Why is the setUsage called when moving to installing? Shouldn't it be set when moving to preparing ?

internal/host/refresh_status_preprocessor.go
```diff
@@ -59,11 +61,46 @@ func newRefreshPreprocessor(log logrus.FieldLogger, hwValidatorCfg *hardware.Val
...
+
 func (r *refreshPreprocessor) preprocess(ctx context.Context, c *validationContext) (map[string]bool, ValidationsStatus, error) {
 	conditions := make(map[string]bool)
 	validationsOutput := make(ValidationsStatus)
+	var ignoredValidations []string
+	var err error
+	if c.cluster != nil && c.cluster.IgnoredHostValidations != "" {
+		ignoredValidations, err = deserializeJSONList(c.cluster.IgnoredHostValidations)
+		if err != nil {
+			logrus.WithError(err).Warn("Unable to deserialize ignored host validations, no validations will be ignored.")
```
ori-amizur, line 93: Why does the flow continue ?

paul-maidment: The flow has a responsibility to perform validations, I felt that the correct behaviour is to continue as though the setting had not been applied. At the very least, you'll just get the regular failing validations.

Seems wrong to stop the validation flow due to this?

internal/cluster/transition.go
paul-maidment: Maybe... "preparing" is the first stage after the user starts the install right?

I'll look into that

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -47,24 +49,55 @@ func newRefreshPreprocessor(log logrus.FieldLogger, hostAPI host.API, operatorsA
...
-		stateMachineInput[v.id.String()] = st == ValidationSuccess
-		category, err := v.id.Category()
+		var category string
+		category, err = v.id.Category()
 		if err != nil {
 			logrus.WithError(err).Warn("id.category()")
 			return nil, nil, err
 		}
+
+		st, message := v.condition(c)
```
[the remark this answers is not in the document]
paul-maidment: I think you are right, it does not need to change (now)

I made a mistake earlier and felt that all conditions should be bypassed, so I had a section of code to deal with this, required a little reorganisation.

I'll try to undo this as it doesn't add value

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -47,24 +49,55 @@ func newRefreshPreprocessor(log logrus.FieldLogger, hostAPI host.API, operatorsA
 	}
 }
 
+func deserializeJSONList(jsonString string) ([]string, error) {
```
tsorya, line 52: i think it should move to utils, common. Some place that can be used by others too

internal/host/refresh_status_preprocessor.go
```diff
@@ -59,11 +61,46 @@ func newRefreshPreprocessor(log logrus.FieldLogger, hwValidatorCfg *hardware.Val
 
 const validationDisabledByConfiguration = "Validation disabled by configuration"
 
+func deserializeJSONList(jsonString string) ([]string, error) {
```
tsorya, line 64: Duplicated code, please move it to one place

internal/host/refresh_status_preprocessor.go
```diff
@@ -59,11 +61,46 @@ func newRefreshPreprocessor(log logrus.FieldLogger, hwValidatorCfg *hardware.Val
...
+	if jsonString != "" {
+		err := json.Unmarshal([]byte(jsonString), &list)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return list, nil
+}
+
+func shouldIgnoreValidation(ignoredValidations []string, validationId string) bool {
```
tsorya, line 75: same

internal/cluster/transition.go
paul-maidment: As discussed with you, I'm going to look at moving this to the "install" API call as this should greatly simplify things

internal/cluster/transition.go
```diff
@@ -43,6 +44,15 @@ type transitionHandler struct {
 	installationTimeout time.Duration
 	finalizingTimeout   time.Duration
 	eventsHandler       eventsapi.Handler
+	usageAPI            usage.API
+}
+
+func (th *transitionHandler) setUsage(enabled bool, name string, props *map[string]interface{}, usages map[string]models.Usage) {
```
tsorya, line 50: I think i don't fully understand the need for this function, it is used one time with enabled=true, so why we need it?
and not just using "th.usageAPI.Add(usages, name, props)"

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -47,24 +49,55 @@ func newRefreshPreprocessor(log logrus.FieldLogger, hostAPI host.API, operatorsA
...
 func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreprocessContext) (map[string]bool, map[string][]ValidationResult, error) {
 	stateMachineInput := make(map[string]bool)
 	validationsOutput := make(map[string][]ValidationResult)
 	checkValidationsInStatuses := []string{
 		models.ClusterStatusInsufficient, models.ClusterStatusReady, models.ClusterStatusPendingForInput, models.ClusterStatusPreparingForInstallation,
 	}
+	var ignoredValidations []string
+	var err error
+	if c.cluster != nil && c.cluster.IgnoredClusterValidations != "" {
+		ignoredValidations, err = deserializeJSONList(c.cluster.IgnoredClusterValidations)
```
tsorya, line 82: Maybe be better to move it closer to usage?

tsorya: It seems that in each place you ignore err and just log it can we have a function that gets cluster and returns ignored validations? and log in case of error?

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -99,6 +133,17 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
 	for _, validationResults := range validationsOutput {
 		sortByValidationResultID(validationResults)
 	}
+
+	// Now that all validations have been executed, skip the validations that we wish to ignore.
+	for _, currentResult := range validationsOutput {
+		for _, v := range currentResult {
+			if shouldIgnoreValidation(ignoredValidations, string(v.ID)) && !stateMachineInput[string(v.ID)] {
```
tsorya, line 140: Why there is a need for !stateMachineInput[string(v.ID)] ? Wouldn't it runs only once per validation?

internal/cluster/transition.go
paul-maidment: That's a fair point, though a suggestion from Ori negates the need to do this at all, so I will be changing this completely.

internal/bminventory/inventory.go
```diff
@@ -1220,10 +1220,16 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params
 				autoAssigned = autoAssigned || updated
 			}
 		}
+		if cluster.IgnoredClusterValidations != "" || cluster.IgnoredHostValidations != "" {
```
tsorya, line 1223: maybe we can add it as function?

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -33,6 +35,20 @@ type refreshPreprocessor struct {
...
 }
 
+func getIgnoredValidations(c *clusterPreprocessContext) ([]string, error) {
+	ignoredValidations := []string{}
+	if c.cluster != nil && c.cluster.IgnoredClusterValidations != "" {
+		var err error
+		ignoredValidations, err = common.DeserializeJSONList(c.cluster.IgnoredClusterValidations)
+		if err != nil {
+			message := "Unable to deserialize ignored cluster validations, no validations will be ignored."
+			logrus.WithError(err).Warn(message)
```
tsorya, line 45: I think better to log failure in callee or provide log to this one, in order to log with cluster id

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -99,6 +123,16 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
 	for _, validationResults := range validationsOutput {
 		sortByValidationResultID(validationResults)
 	}
+
+	// Now that all validations have been executed, skip the validations that we wish to ignore.
+	for _, currentResult := range validationsOutput {
```
tsorya, line 128: cant we do it on validationsOutput creation? While going through results?

internal/cluster/refresh_status_preprocessor_test.go
```diff
@@ -0,0 +1,156 @@
...
+			createCluster()
+			mockFailAllValidations()
+			validationContext = newClusterValidationContext(cluster, db)
+		})
+
+		AfterEach(func() {
+			deleteCluster()
+		})
+
+		It("Should not ignore any validations if IgnoredClusterValidations is empty or invalid", func() {
```
tsorya, line 113: You check here only invalid, can we check nil too?

internal/host/refresh_status_preprocessor.go
```diff
@@ -130,6 +156,15 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *validationConte
 		}
 	}
 
+	// Now that all validations have been executed, skip the validations that we wish to ignore.
+	for _, currentResult := range validationsOutput {
```
tsorya, line 160: same question as will cluster part

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -53,14 +69,21 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
...
+	}
+
 	//if the cluster is not on discovery stages - skip the validations check
 	if !funk.ContainsString(checkValidationsInStatuses, swag.StringValue(c.cluster.Status)) {
 		return stateMachineInput, validationsOutput, nil
 	}
 	for _, v := range r.validations {
 		st, message := v.condition(c)
 		stateMachineInput[v.id.String()] = st == ValidationSuccess
-		category, err := v.id.Category()
```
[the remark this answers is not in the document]
paul-maidment: to deal with the following error 

`             ignoredValidations, err =  common.GetIgnoredValidations(c.cluster.IgnoredClusterValidations, "cluster", string(*c.cluster.ID))
internal/cluster/refresh_status_preprocessor.go:89:13: shadow: declaration of "err" shadows declaration at line 73 (govet)`

which would require an even messier enclosure of brackets if I were to approach another way

internal/cluster/refresh_status_preprocessor.go
paul-maidment: This block is here for a reason and in attempting to resolve this comment I am quite sure that it should be here.

The issue is that we are not applying the ignore at the very last moment that anything could alter a condition and in testing, I have just demonstrated that this concern is justified by implementing your suggestion.

So, in the lack of any further suggestions, I opt to leave this code block here. Open to reviewing the block itself to simplify, but not to removing it.

internal/cluster/refresh_status_preprocessor_test.go
paul-maidment: No need to check for nil as a string cannot possibly have this value.

internal/host/refresh_status_preprocessor.go
paul-maidment: This block is here for a reason and in attempting to resolve this comment I am quite sure that it should be here.

The issue is that we are not applying the ignore at the very last moment that anything could alter a condition and in testing, I have just demonstrated that this concern is justified by implementing your suggestion.

So, in the lack of any further suggestions, I opt to leave this code block here. Open to reviewing the block itself to simplify, but not to removing it.

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -53,16 +54,27 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
 	checkValidationsInStatuses := []string{
 		models.ClusterStatusInsufficient, models.ClusterStatusReady, models.ClusterStatusPendingForInput, models.ClusterStatusPreparingForInstallation,
 	}
+	var ignoredValidations []string
+	var err error
+	if c.cluster != nil {
+		ignoredValidations, err = common.GetIgnoredValidations(c.cluster.IgnoredClusterValidations, "cluster", string(*c.cluster.ID))
```
tsorya, line 60: maybe just to provide clustet? and type?

internal/cluster/refresh_status_preprocessor.go
tsorya: I actually though about setting status to passed and then no need to set stateMachineInput[string(v.ID)] = true ?

internal/common/common.go
```diff
@@ -93,6 +94,34 @@ const NMDebugModeConf = `
 domains=ALL:DEBUG
 `
 
+func GetIgnoredValidations(validationsJSON string, validationType string, clusterID string) ([]string, error) {
```
tsorya, line 97: I think we should not provide validationType, you can handle error outside the function, no?

internal/cluster/refresh_status_preprocessor.go
tsorya: I think actually providing c.cluster.IgnoredClusterValidations should be enough, just print error here. We will get everything we need from this log line

internal/cluster/refresh_status_preprocessor.go
paul-maidment: Ah, this was discussed with Ori and I agree with his take on this that it is better to preserve the content and actual state of the validation. 

So we use the statemachine conditions. 

internal/common/common.go
paul-maidment: I guess it's just for the error message so I can refactor that

internal/common/common.go
```diff
@@ -93,6 +94,33 @@ const NMDebugModeConf = `
 domains=ALL:DEBUG
 `
 
+func GetIgnoredValidations(validationsJSON string, clusterID string) ([]string, bool) {
```
tsorya, line 97: lets return actual error it will help to see whats the problem

tsorya, line 97: Something i don't get here, it is actually the same as DeserializeJSONList, no? Maybe i somehow confused you

tsorya: clusterID is needed?

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -91,6 +108,10 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
 			Status:  status,
 			Message: strings.Join(result.Reasons, "\n"),
 		})
+		// if common.ShouldIgnoreValidation(ignoredValidations, string(id), common.NonIgnorableClusterValidations) {
```
tsorya, line 111: delete ?

PR #4925, opened 2023-01-23
MGMT-12535:  Service implementation for verify-vips

ori-amizur:
1. Create GetNextSteps implementation for verify vips
    This is executed when host is pending-for-input, insufficient or known.
    This step asks for verification for all VIPs.
    In addition free-addresses step is being run for IPv4 machine networks
    having CIDR prefix greater or equal to 22, and when kubeapi is not
    enabled. This is similar to the previous implemenation but the
    allowed network size have been reduced.

2. Add post step reply for verify-vips step
    This step modifies the vips verification value according to the results
    received in the step reply

3. Modify api-vips-valid and ingress-vips-valid validations to handle the results of verify-vips
    VIPs are now verified using hybrid mode.  If results from verify-vips
    exist for a vip it is used.  If not, then if there is a result from
    free-addresses, it is used.
    Otherwise, the validation is considered failing.
    This changes previous implementation that these validations were
    considered successful if there was no result from free-addresses.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


internal/host/hostcommands/verify_vips_cmd.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (f *verifyVipsCmd) prepareParam(host *models.Host) (string, error) {
+	var inventory models.Inventory
+	err := json.Unmarshal([]byte(host.Inventory), &inventory)
+	if err != nil {
+		f.log.WithError(err).Warn("Inventory parse")
+		return "", err
+	}
+	if host.ClusterID == nil || *host.ClusterID == "" {
+		return "", nil
```
danielerez, line 34: Probably worth a log also here.

internal/host/hostcommands/verify_vips_cmd.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	cluster, err := common.GetClusterFromDB(common.LoadTableFromDB(common.LoadTableFromDB(f.db, common.APIVIPsTable), common.IngressVIPsTable),
```
danielerez, line 36: The dual LoadTableFromDB looks weird:)
Maybe split it into two statements?

internal/host/hostcommands/verify_vips_cmd.go
```diff
@@ -0,0 +1,81 @@
...
+		*host.ClusterID, common.SkipEagerLoading)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to get cluster %s", host.ClusterID.String())
+	}
+	if len(cluster.APIVips) == 0 && len(cluster.IngressVips) == 0 {
+		return "", nil
```
danielerez, line 42: Shouldn't happen, but worth a log?

internal/host/hostcommands/verify_vips_cmd.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	request := models.VerifyVipsRequest{}
+	for _, vip := range cluster.APIVips {
```
danielerez, line 45: might look a little cleaner with funk.ForEach

internal/common/common.go
```diff
@@ -434,3 +434,7 @@ func ApplyYamlPatch(src []byte, ops []byte) ([]byte, error) {
 
 	return patched, nil
 }
+
+func VipVerificationAddress(v models.VipVerification) *models.VipVerification {
```
danielerez, line 438: For consistency, can we move it to internal/common/conversions.go, and call it VipVerificationPtr.
We already have some similar funcs there.

internal/cluster/cluster.go
```diff
@@ -1532,3 +1533,59 @@ func (m *Manager) updateSchedulableMastersForcedTrue(ctx context.Context, cluste
...
 }
+
+func (m *Manager) HandleVerifyVipsResponse(ctx context.Context, clusterID strfmt.UUID, stepReply string) error {
+	log := logutil.FromContext(ctx, m.log)
+	return m.db.Transaction(func(tx *gorm.DB) error {
+		var response models.VerifyVipsResponse
+		if err := json.Unmarshal([]byte(stepReply), &response); err != nil {
+			log.WithError(err).Error("HandleVerifyVipsResponse: unmarshal")
+		}
+		cluster, err := common.GetClusterFromDB(common.LoadTableFromDB(common.LoadTableFromDB(tx, common.APIVIPsTable), common.IngressVIPsTable),
```
danielerez, line 1544: The dual LoadTableFromDB is a bit confusing:) maybe extract to a another statement?

internal/cluster/cluster.go
```diff
@@ -1532,3 +1533,59 @@ func (m *Manager) updateSchedulableMastersForcedTrue(ctx context.Context, cluste
...
+			clusterID, common.SkipEagerLoading)
+		if err != nil {
+			log.WithError(err).Errorf("HandleVerifyVipsResponse: getting cluster %s", clusterID.String())
+			return err
+		}
```
danielerez, line 1549: This func is quite large:/ Maybe extract lines 1544-1549 to something like: getClusterWithVIPs

internal/cluster/cluster.go
```diff
@@ -1532,3 +1533,59 @@ func (m *Manager) updateSchedulableMastersForcedTrue(ctx context.Context, cluste
...
 	return nil
 }
+
+func (m *Manager) HandleVerifyVipsResponse(ctx context.Context, clusterID strfmt.UUID, stepReply string) error {
+	log := logutil.FromContext(ctx, m.log)
+	return m.db.Transaction(func(tx *gorm.DB) error {
+		var response models.VerifyVipsResponse
+		if err := json.Unmarshal([]byte(stepReply), &response); err != nil {
+			log.WithError(err).Error("HandleVerifyVipsResponse: unmarshal")
+		}
```
danielerez, line 1543: I see the response is used only in line 1551, so this can be move to there.

internal/cluster/cluster.go
```diff
@@ -1532,3 +1533,59 @@ func (m *Manager) updateSchedulableMastersForcedTrue(ctx context.Context, cluste
...
+		cluster, err := common.GetClusterFromDB(common.LoadTableFromDB(common.LoadTableFromDB(tx, common.APIVIPsTable), common.IngressVIPsTable),
+			clusterID, common.SkipEagerLoading)
+		if err != nil {
+			log.WithError(err).Errorf("HandleVerifyVipsResponse: getting cluster %s", clusterID.String())
+			return err
+		}
+		updated := false
+		for _, v := range response {
```
danielerez, line 1551: funk.ForEach could be a bit cleaner here.

internal/cluster/cluster.go
```diff
@@ -1532,3 +1533,59 @@ func (m *Manager) updateSchedulableMastersForcedTrue(ctx context.Context, cluste
...
+			vipResponse := v
```
danielerez, line 1552: why do we need a new var here? can't we just rename the original var?

internal/cluster/cluster.go
```diff
@@ -1532,3 +1533,59 @@ func (m *Manager) updateSchedulableMastersForcedTrue(ctx context.Context, cluste
...
+				if apiVip != nil {
+					if apiVip.Verification == nil || *apiVip.Verification != *vipResponse.Verification {
+						apiVip.Verification = vipResponse.Verification
+						if err = tx.Save(apiVip).Error; err != nil {
+							log.WithError(err).Errorf("saving verification for api vip %s of cluster %s", apiVip.IP, clusterID.String())
+							return err
+						}
+						updated = true
+					}
+				}
```
danielerez, line 1568: These lines are very similar to 1570-1580, is it difficult to create a generic function to cover both?
If so, at least worth extracting to two separate func, to make HandleVerifyVipsResponse a bit clearer:)

internal/network/machine_network_cidr.go
```diff
@@ -84,26 +84,41 @@ func ipInCidr(ipStr, cidrStr string) bool {
...
-func VerifyVip(hosts []*models.Host, machineNetworkCidr string, vip string, vipName string, mustExist bool, log logrus.FieldLogger) error {
-	if !mustExist && vip == "" {
-		return nil
+func VerifyVipFree(hosts []*models.Host, vip string, machineNetworkCidr string, verification *models.VipVerification, log logrus.FieldLogger) models.VipVerification {
+	if verification != nil {
+		switch *verification {
+		case models.VipVerificationSucceeded, models.VipVerificationFailed:
+			return *verification
+		default:
+			break
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/host/hostcommands/verify_vips_cmd.go
ori-amizur: OK

internal/network/machine_network_cidr.go
```diff
@@ -569,13 +584,16 @@ func MakeFreeAddressesSet(hosts []*models.Host, network string, prefix *string,
...
 	if len(freeSet) > 0 {
-		_, isFree = freeSet[strfmt.IPv4(vipIPStr)]
+		_, isFree := freeSet[strfmt.IPv4(vipIPStr)]
+		if isFree {
+			return models.VipVerificationSucceeded
+		}
+		return models.VipVerificationFailed
 	}
-	return isFree
+	return models.VipVerificationUnverified
```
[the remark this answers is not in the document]
ori-amizur: These results are used in 2 flows

1. From API.  Unverified is considered successful.
2. From validations.  Unverified is considered failure (with appropriate message).
Anyway, unverified does not mean not required.
Will add comment.

internal/host/hostcommands/verify_vips_cmd.go
ori-amizur: OK

internal/host/hostcommands/verify_vips_cmd.go
ori-amizur: May happen if VIPs were not defined yet.  Logging will cause noise in he log file.

internal/common/common.go
ori-amizur: OK

internal/cluster/cluster.go
ori-amizur: OK

internal/cluster/cluster.go
ori-amizur: Linter produces an error for loop variables that are passed to functions.

internal/cluster/cluster.go
ori-amizur: The types models.APIVip and models.IngressVip are different so go does allow it.
It will need swagger change which I prefer not to do here.

internal/cluster/cluster.go
ori-amizur: OK

internal/host/hostcommands/verify_vips_cmd.go
ori-amizur: I am not sure I agree. It needs to modify external variable request, so the append will be in the ForEach function parameter.
So the ForEach will look like
```
funk.ForEach(cluster.APIVips, func(vip *models.APIVip) {request = append(request, &models.VerifyVip{Vip:     vip.IP,  VipType: models.VipTypeAPI})})
```
Is it cleaner ?

danielerez: Apparently that's exactly the example in gofunk.[ForEachRight](https://github.com/thoas/go-funk#funkforeachright):) I think it looks a bit cleaner as we don't need the key inside the loop. But don't have a strong opinion on it.

internal/cluster/cluster.go
danielerez: > Linter produces an error for loop variables that are passed to functions.

Oh right. So I guess gofunk.ForEach worth a try also here.

internal/cluster/cluster.go
danielerez: Yeah makes sense:/ So maybe just extract to a couple of functions if worth the effort. I mean just for reusability maybe and to make this func a bit clearer. Not crucial..

internal/host/hostcommands/verify_vips_cmd.go
ori-amizur: To me seems to complicate.  Prefer to skip.

internal/cluster/cluster.go
ori-amizur: Seems quite a big function to put in ForEach.  Not sure it makes it better.

PR #4927, opened 2023-01-23
MGMT-13203: Create REST points for validation ignore feature.

paul-maidment:
[MGMT-13203](https://issues.redhat.com//browse/MGMT-13203): Create REST points for validation ignore feature.

This provides the API endpoint for this feature enhancement https://issues.redhat.com/browse/MGMT-12547

The aim of the feature is to allow the customer to ignore validations if they have org permissions to do so.
This PR implements the API enpoint to allow the customer to ignore the validations.

We have decided to keep this separate from the cluster API as we are not actively encouraging use of this feature.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


paul-maidment:
@cc @avishayt

paul-maidment:
cc: @avishayt

swagger.yaml
```diff
@@ -3498,7 +3498,104 @@ paths:
           schema:
             $ref: '#/definitions/error'
 
+  /v2/clusters/{cluster_id}/ignored-cluster-validations:
```
avishayt, line 3501:
```suggestion
  /v2/clusters/{cluster_id}/ignored-validations:
```

Let's save an API and just specify both cluster and host validations in the body - one JSON for each.
Also, need a GET API to return the current value.

swagger.yaml
```diff
@@ -3498,7 +3498,80 @@ paths:
...
+  /v2/clusters/{cluster_id}/ignored-validations:
+    get:
+      tags:
+        - installer
+      description: Fetch the validations which are to be ignored for this cluster.
+      operationId: v2GetIgnoredValidations
+      parameters:
+        - in: path
+          name: cluster_id
+          description: The cluster to which the list of validations to be ignored should be applied.
```
avishayt, line 3510:
```suggestion
          description: The cluster whose failing validations should be ignored according to this list.
```

swagger.yaml
```diff
@@ -3498,7 +3498,80 @@ paths:
...
+            $ref: '#/definitions/cluster' 
+    put:
+      tags:
+        - installer
+      description: Register the validations which are to be ignored for this cluster.
+      operationId: v2SetIgnoredValidations
+      parameters:
+        - in: path
+          name: cluster_id
+          description: The cluster to which the list of validations to be ignored should be applied.
```
avishayt, line 3527:
```suggestion
          description: The cluster whose failing validations should be ignored according to this list
```

swagger.yaml
```diff
@@ -3498,7 +3498,80 @@ paths:
...
+          type: string
+          format: uuid
+          required: true
+        - in: body
+          name: ignored_validations
+          description: The validations which are to be ignored.
```
avishayt, line 3533:
```suggestion
          description: The validations to be ignored.
```

swagger.yaml
```diff
@@ -3498,7 +3498,80 @@ paths:
...
+          name: cluster_id
+          description: The cluster to which the list of validations to be ignored should be applied.
+          type: string
+          format: uuid
+          required: true
+      responses:
+        "200":
+          description: Success.
+          schema:
+            $ref: '#/definitions/cluster'
```
avishayt, line 3518: There's no reason to have another API to return clusters - let's remove this GET API.

swagger.yaml
```diff
@@ -3498,7 +3498,80 @@ paths:
...
+        "500":
+          description: Error.
+          schema:
+            $ref: '#/definitions/error'
+
 definitions:
+  ignored-validations:
+    type: object
+    properties:
+      cluster-ignored-ids:
```
avishayt, line 3567:
```suggestion
      cluster-validation-ids:
```

swagger.yaml
```diff
@@ -3498,7 +3498,80 @@ paths:
...
+        type: string
+        format: string
+        description: A list of cluster validation id's to be ignored.
+      host-ignored-ids:
```
avishayt, line 3571:
```suggestion
      host-validation-ids:
```

swagger.yaml
```diff
@@ -4856,6 +4929,14 @@ definitions:
         type: string
         description: Json formatted string containing ip collisions detected in the cluster.
         x-go-custom-tag: gorm:"type:text"
+      ignored_host_validations:
+        type: string
+        description: Json formatted string containing a list of host validations to be ignored. May also contain a list with a single string "all" to ignore all host validations. Some validations cannot be ignored.
```
avishayt, line 4934:
```suggestion
        description: JSON-formatted list of host validation IDs that will be ignored.  It may also contain a list with a single string "all" to ignore all host validations.  Some validations cannot be ignored.
```

swagger.yaml
```diff
@@ -4856,6 +4929,14 @@ definitions:
...
+        x-go-custom-tag: gorm:"type:text"
+      ignored_cluster_validations:
+        type: string
+        description: Json formatted string containing a list of cluster validations to be ignored. May also contain a list with a single string "all" to ignore all cluster validations. Some validations cannot be ignored.
```
avishayt, line 4938:
```suggestion
        description: JSON-formatted list of cluster validation IDs that will be ignored.  It may also contain a list with a single string "all" to ignore all cluster validations.  Some validations cannot be ignored.
```

avishayt:
Looks good to me. Not sure about the split of code between inventory.go and inventory_v2_handlers.go. Let's have another pair of eyes on this.

swagger.yaml
```diff
@@ -3498,7 +3498,88 @@ paths:
...
+          description: Forbidden.
+          schema:
+            $ref: '#/definitions/infra_error'
+        "500":
+          description: Error.
+          schema:
+            $ref: '#/definitions/error'
+
 definitions:
+  ignored-validations:
```
omertuc, line 3572: I don't like that the host and cluster ignored validations are bundled up like this, clusters are one entity, hosts are another entity, each of them should have its own field.

Mostly because hosts are not always bound to clusters and can move between them

paul-maidment: This is a cluster level setting and should not be applied on a "per host" basis, that's not the intention here.

The intention is that for very specific customers, we would like to enable them to mark specific validations as skippable across the board. What this setting means is "during the installation of this cluster, whenever you encounter this host validation, ignore it..."

If we go the other way, registering this on a per host basis, it's possible but it would complicate the usage of this feature for the customer. 

Open to change my mind if there is broader support for changing in this way.

omertuc: Hosts have validations even when detached from clusters, how would users skip those then?

omertuc: At least, they do as far as I know. I may be wrong about this

paul-maidment: Checking this point, I didn't make it clear in the enhancement doc but unbound clusters are a non-goal for this feature. This is really for those customers who have a "more or less" viable cluster ready to install but are blocked by one validation that would allow the installation to proceed if they could just skip it.

avishayt: I'm sorry I missed this earlier, but @omertuc has a good point.  We may want to expose this functionality in our Kubernetes API, and in late binding, we do have some host validations that run before binding.  Worse, we block binding if a host has failing validations.

filanov: kube-api is one thing late binding is another thing, after all the controller will call what ever it needs. 
the downside of having it in the host level is that users will need to do it per host, on the other had having it in the infra-env level will mean that users that use late binding (we don't have many of those as far as i know) will affect all the hosts in the infra-env. 
So i assume that having the field for each host is more flexible. but it's sounds like an overkill for something that we don't want to officially support. 
for me for now it's enough to have it on the cluster level and if we see a demand we can decide on how to improve, it's not an api that we need to maintain and not break.

omertuc: > that users will need to do it per host

Not necessarily, we could have a hybrid solution where hosts inherit the validation-skips from their cluster, and also have their own list

omertuc: A hybrid solution can also be done in parts, i.e. first merge this PR as is, then add per-host skips in the future

omertuc: The downside of a hybrid solution is that it's much more complex in the backend - two tiers of validation skips that have to be merged

paul-maidment: The suggestion from @omertuc sounds reasonable as part of a strategy where we see whether or not there is demand to expand the functionality further.

paul-maidment: "we will need to think how to add it to kube-api without having that directly in the spec because we don't want to expose it as part of the official api" -> What about a custom CRD that accepts a cluster reference?

filanov: i was thinking about annotations, but for now i would not even add it to kube-api at this point.
reference is still something that you will have in the spec.

paul-maidment: So for now, this REST API should do the job?

paul-maidment: So for now, this REST API should do the job?

filanov: yes, lets start with that and see if it's used.

avishayt: > improve only on demand

It will be harder to change later once it's in the DB.  Why not fix it now?

omertuc: > > improve only on demand
> 
> It will be harder to change later once it's in the DB. Why not fix it now?

Assuming we're going for an eventual hybrid solution, you still need the clusters to specify both host and cluster validations, so clusters can affect all of their hosts in bulk, making it simpler for users

omertuc: > I think a hybrid solution would be even more complicated because we would need to deal with different values on the host vs on the cluster.

Yes, that's the point I was making [here](https://github.com/openshift/assisted-service/pull/4927#discussion_r1094429262)

Although the combination is a simple union(set(host_skips) | set(cluster_host_skips))

paul-maidment: @filanov looking for your further input here...

filanov: if it will be on the host level we will need to handle cases where host is reused, isn't that an overkill? 
Lets remember the goal, it's to skip validation in a very specific cases to enable starting instlalation, and it's not supported api. 
late binding can be a non goal, if the infra-env have a cluster reference host will not be unbound so the validations are not relevant there. 

> It will be harder to change later once it's in the DB. Why not fix it now?
what is the concern about the DB? cluster will have a list regardless which solution we have.

avishayt: > what is the concern about the DB? cluster will have a list regardless which solution we have.
Either the cluster has two lists (one for cluster and one for host validations), o the cluster resource has one list and the host resource has one list.

What about a host with some hardware that we're not validating properly, or someone wants to install it with a smaller disk?  The validation should be ignored regardless of the cluster that the host is bound to.

filanov: Maybe i'll ask it in a different way, with the existing solution besides late binding, do we have any case that will prevent the user to start installation if he is using the api that suggested here? and the second question is, what in this solution blocking us from making improvements later (Avishay said DB, not sure i understand why)?

avishayt: I think it will be relatively little effort to change now.  Later it will be more difficult.  We'll add it to the host, then people will ask why it's both on the host and on the cluster, then we'll remove it from the cluster.

paul-maidment: another option is to dispose of this whole host vs cluster flexibility, make it a single flag "ignore all validations (with the exception of the ones we can't ignore)"

So I could get rid of the JSON fields and replace with a boolean.

@ori-amizur made this suggestion, also sounds reasonable?

filanov: > I think it will be relatively little effort to change now. Later it will be more difficult. We'll add it to the host, then people will ask why it's both on the host and on the cluster, then we'll remove it from the cluster.

OK lets go with that then. 
two different apis, one for host and one for cluster. in hat case we have it per host so users will have a bit more work but ti's fine because the api is not recommended anyway

paul-maidment: @filanov before we go ahead, should we consider the suggestion from @ori-amizur ?

paul-maidment: Then let's just stick with what we have in this PR and not overcomplicate things

filanov: yes please

filanov: Resolving this conversation. 
The current api will provide a good solution for all the users, we are ignoring late binding for this api. 
as described it's a patch solution for blocked users and not a supported api.

paul-maidment:
Description updated

swagger.yaml
omertuc, line 3572: shouldn't the `cluster:` definition have `ignored-validations` under it?

swagger.yaml
```diff
@@ -3498,7 +3498,88 @@ paths:
...
+    type: object
+    properties:
+      cluster-validation-ids:
+        type: string
+        format: string
+        description: JSON-formatted list of cluster validation IDs that will be ignored.  It may also contain a list with a single string "all" to ignore all cluster validations.  Some validations cannot be ignored.
+      host-validation-ids:
+        type: string
+        format: string
+        description: JSON-formatted list of host validation IDs that will be ignored.  It may also contain a list with a single string "all" to ignore all host validations.  Some validations cannot be ignored.
```
omertuc, line 3582: I think this should explicitly mention "... that will be ignored for all hosts that belong to this cluster"

swagger.yaml
```diff
@@ -3498,7 +3498,88 @@ paths:
...
+            $ref: '#/definitions/error'
+
 definitions:
+  ignored-validations:
+    type: object
+    properties:
+      cluster-validation-ids:
+        type: string
+        format: string
+        description: JSON-formatted list of cluster validation IDs that will be ignored.  It may also contain a list with a single string "all" to ignore all cluster validations.  Some validations cannot be ignored.
```
omertuc, line 3578:
```suggestion
        description: JSON-formatted list of cluster validation IDs that will be ignored.  It may also contain a list with a single string "all" to ignore all cluster validations. Some validations cannot be ignored.
```

omertuc:
We should have a feature usage flag for this feature, so we can easily filter out clusters that have it

omertuc:
What is the dedicated GET endpoint for? Why not just return it along with all cluster fields when the user GETs a cluster as we do for all other cluster fields?

omertuc:
Also we already have a PATCH API for clusters to change cluster fields, we shouldn't have a dedicated endpoint for validations either

paul-maidment:
> We should have a feature usage flag for this feature, so we can easily filter out clusters that have it

We do, but it's part of another ticket in the epic, the feature usage is being recorded at the moment the cluster transitions to installing is this is the moment at which we can say this feature was used.

paul-maidment:
> What is the dedicated GET endpoint for? Why not just include those fields among all cluster fields when the user GETs an entire cluster as we do for all other cluster fields?

This was previously discussed with @avishayt and @filanov the main reason is to keep this isolated from existing API's as it's not a feature we are actively encouraging use of.

swagger.yaml
omertuc: We have a `cluster:` [object](https://github.com/openshift/assisted-service/blob/2688136ed6398480e15eb6575466c265f92983e2/swagger.yaml#L4688), the ignored host validations and ignored cluster validations should be a parameter in that object

paul-maidment:
> Also we already have a PATCH API for clusters to change cluster fields, we shouldn't have a dedicated endpoint for ignoring validations

See previous comment

omertuc:
> > What is the dedicated GET endpoint for? Why not just include those fields among all cluster fields when the user GETs an entire cluster as we do for all other cluster fields?
> 
> This was previously discussed with @avishayt and @filanov the main reason is to keep this isolated from existing API's as it's not a feature we are actively encouraging use of.

Ack

swagger.yaml
paul-maidment: Sorry, I disagree, we made a purposeful decision to exclude this from the cluster api

filanov: the suggestion from @avishayt was not to include it in the cluster object because it's not a supported api

omertuc: Alright. Next time please mention such non-obvious decisions in the commit message

paul-maidment: That's a fair point, I'll amend the commit message now.

internal/common/validations.go
```diff
@@ -3,7 +3,38 @@
...
+var NonIgnorableClusterValidations []string = []string{"api-vips-defined", "ingress-vips-defined", "all-hosts-are-ready-to-install", "sufficient-masters-count", "pull-secret-set", "cluster-preparation-succeeded"}
+
+var MayIgnoreValidation = func(validationID string, nonIgnorables []string) bool {
+	if validationID == "all" {
+		return true
+	}
+	return !swag.ContainsStrings(nonIgnorables, validationID)
+}
+
+var MayIgnoreValidations = func(validationIDs []string, nonIgnorables []string) (bool, string) {
```
omertuc, line 27: Why is this not a simple function? Why var?

paul-maidment: Good point, just fixed that.

internal/common/validations.go
```diff
@@ -3,7 +3,38 @@
...
+
 // IsAgentCompatible checks if the given agent image is compatible with what the service expects.
 func IsAgentCompatible(expectedImage, agentImage string) bool {
 	return agentImage == expectedImage
 }
+
+var NonIgnorableHostValidations []string = []string{"connected", "has-inventory", "machine-cidr-defined", "hostname-unique", "hostname-valid"}
+var NonIgnorableClusterValidations []string = []string{"api-vips-defined", "ingress-vips-defined", "all-hosts-are-ready-to-install", "sufficient-masters-count", "pull-secret-set", "cluster-preparation-succeeded"}
+
+var MayIgnoreValidation = func(validationID string, nonIgnorables []string) bool {
```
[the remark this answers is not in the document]
omertuc: And this one has no callers outside the package, it should be privat

internal/common/validations.go
```diff
@@ -3,7 +3,38 @@
...
+	for _, validation := range validationIDs {
+		if validation == "all" {
+			return true, ""
+		}
+		if !MayIgnoreValidation(validation, nonIgnorables) {
+			cantBeIgnored = append(cantBeIgnored, validation)
+			result = false
+		}
+	}
+	return result, strings.Join(cantBeIgnored, ",")
```
omertuc, line 39: `strings.Join` is for the sole purpose of displaying something to the user, which this function should not be responsible for. It's very confusing to see the join here as there's no obvious reason for it to happen in this context. The caller should perform the join instead, this function should return a simple array

omertuc:
[This](https://github.com/openshift/assisted-service/pull/4927#discussion_r1094817565) comment is in a resolved conversation but it was not resolved

internal/common/validations.go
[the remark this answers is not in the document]
paul-maidment: The private thing..it will have callers outside of the package when the next PR is applied, so if I make it private now, I have to change it later.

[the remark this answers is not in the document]
paul-maidment: > [This](https://github.com/openshift/assisted-service/pull/4927#discussion_r1094817565) comment is in a resolved conversation but it was not resolved
I have unresolved so that we can discuss, we can make this private, only to make it public in the next PR, seems fruitless to me.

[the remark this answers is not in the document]
omertuc: ok

internal/common/validations.go
paul-maidment: Done, join operation moved to the caller and changed return a string array

paul-maidment:
@omertuc I have addressed the review points made and would like to request further feedback or an LGTM

filanov (approved):

filanov (approved):

PR #4894, opened 2023-01-15
MGMT-13008: validation for ignition image size

danielerez:
Added a validation to v2UpdateHostIgnition API for ensuring that the ignition archive isn't larger than 256KiB See [ignition_img_size](https://github.com/coreos/coreos-assembler/blob/d2c968a1f3c75713a4e1449e3da657c5d5a5d7e7/src/cmd-buildextend-live#L113-L114) defined in coreos-assembler. This validation requires executing ignition [Archive](https://github.com/openshift/assisted-image-service/blob/568b59ec570d2b2ed15f5070bb29697fc01c9525/pkg/isoeditor/ignition.go#L15-L41) func on every override ignition to examine its size. 

Note that this should not affect kube-api as the size of annotations is already validated.
E.g. 'The Agent <id> is invalid: metadata.annotations: Too long: must have at most 262144 bytes'

## List all the issues related to this PR

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [x] Cloud
- [ ] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [x] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #4934, opened 2023-01-24
MGMT-13008: ignition size validation on register infraenv

danielerez:
As a follow-up to https://github.com/openshift/assisted-service/pull/4894, added the ignition size validation also to RegisterInfraEnv and UpdateInfraEnv APIs.
These APIs were missed in the aforementioned PR. Also, removed the validation in v2UpdateHostIgnition API which is redundant as we don't embed the host ignition in the ISO.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4380,6 +4380,12 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
 		kernelArguments = swag.String(string(b))
 	}
 
+	if params.InfraenvCreateParams.IgnitionConfigOverride != "" {
+		if err = validations.ValidateIgnitionImageSize(params.InfraenvCreateParams.IgnitionConfigOverride); err != nil {
+			return nil, common.NewApiError(http.StatusBadRequest, err)
+		}
+	}
+
 	infraEnv := common.InfraEnv{
 		Generated: false,
 		InfraEnv: models.InfraEnv{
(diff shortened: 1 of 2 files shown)


carbonin:
>which handled only the v2UpdateHostIgnition API

Sorry I must have missed this. This API is the only one that _doesn't_ need this validation as the host ignition isn't being embedded into the ISO. You can remove the validation from the host ignition APIs.

danielerez:
> > which handled only the v2UpdateHostIgnition API
> 
> Sorry I must have missed this. This API is the only one that _doesn't_ need this validation as the host ignition isn't being embedded into the ISO. You can remove the validation from the host ignition APIs.

Hmm, but won't it be confusing if we don't have it also in v2UpdateHostIgnition any way? Not sure we want to have such inconsistency, unless we can maybe treat it as a workaround(?):/

carbonin:
Why would we limit it for no reason? One needs to fit into a fixed size spot in the ISO, the other doesn't.
They have nothing to do with each other and won't likely have the same kind of content.

danielerez:
> Why would we limit it for no reason? One needs to fit into a fixed size spot in the ISO, the other doesn't. They have nothing to do with each other and won't likely have the same kind of content.

ok, removed.

carbonin (approved):

PR #4929, opened 2023-01-24
support lvm/lvms on 4.11/4.12 respectively

subhamkrai:
Signed-off-by: subhamkrai <srai@redhat.com>

## List all the issues related to this PR

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [x] No tests needed

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #4942, opened 2023-01-26
MGMT-13305: Fix invalid lvms version installed on 4.11

eliorerz:
Fix a bug on https://github.com/openshift/assisted-service/pull/4929 

/cc @eliorerz


internal/bminventory/inventory.go
```diff
@@ -2787,6 +2788,20 @@ func (b *bareMetalInventory) getOLMOperators(cluster *common.Cluster, newOperato
 		if err != nil {
 			return nil, common.NewApiError(http.StatusBadRequest, err)
 		}
+
+		// TODO - Need to find a better way for creating LVMO/LVMS operator on different openshift-version
+		if operator.Name == "lvm" {
+			lvmsMetMinOpenshiftVersion, err := common.VersionGreaterOrEqual(cluster.OpenshiftVersion, lvm.LvmsMinOpenshiftVersion)
+			if err != nil {
+				return nil, err
```
gamli75, line 2796: not sure that we want to return nil - maybe we should set lvms as the default and write error to log.
Another request - please make sure that we are not calling this method in day 2 because we don't have OCP version and it can be nil

gamli75 (approved):

gamli75 (approved):

osherdp:
I'll create it manually

PR #4943, opened 2023-01-26
NO-ISSUE: Wrap VIP update in the DB with transaction

mkowalski:
/cc @omertuc


omertuc:
I'd like to first understand the race and how a transaction will solve it before approving

mkowalski:
/test edge-unit-test

Maybe it's a race, maybe it is not

avishayt (approved):

PR #4945, opened 2023-01-26
Add automatic `ShellCheck` linting for `assisted-boot-reporter.sh`

omertuc:
Waiting for #4543 to merge first, because it includes the file which we aim to lint

For as long as `assisted-boot-reporter.sh` is implemented in Bash,
which is notoriously error-prone, we'll use the `ShellCheck` linter
to at-least avoid some common Bash mistakes

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [x] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
    - Ran the linter and saw that it works as expected
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/Dockerfile.assisted-service-build
+++ b/Dockerfile.assisted-service-build
@@ -33,6 +33,7 @@ COPY --from=golang /usr/local/go /usr/local/go
 COPY --from=quay.io/openshift/origin-cli:latest /usr/bin/oc /usr/bin
 COPY --from=quay.io/operator-framework/upstream-opm-builder:v1.16.1 /bin/opm /bin
 COPY --from=k8s.gcr.io/kustomize/kustomize:v4.3.0 /app/kustomize /usr/bin/
+COPY --from=quay.io/coreos/shellcheck-alpine:v0.5.0 /bin/shellcheck /usr/bin/shellcheck
 
 RUN dnf install -y 'dnf-command(config-manager)' && \
     dnf config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo && \
--- a/Makefile
+++ b/Makefile
@@ -187,6 +187,7 @@ endif
 
 lint:
 	golangci-lint run -v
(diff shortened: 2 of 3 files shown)


omertuc:
/cc @osherdp 
/cc @adriengentil 
/cc @rccrdpccl 

CI Dockerfiles are something you're often involved with so letting you know

Dockerfile.assisted-service-build
```diff
@@ -34,10 +34,10 @@ COPY --from=quay.io/openshift/origin-cli:latest /usr/bin/oc /usr/bin
 COPY --from=quay.io/operator-framework/upstream-opm-builder:v1.16.1 /bin/opm /bin
 COPY --from=k8s.gcr.io/kustomize/kustomize:v4.3.0 /app/kustomize /usr/bin/
 
-RUN dnf install -y 'dnf-command(config-manager)' && \
+RUN dnf install -y 'dnf-command(config-manager)' epel-release && \
```
osherdp, line 37: be aware we previously had some availability issues with epel
although I don't see other way to install ShellCheck

omertuc: Hmmm looks like `quay.io/coreos/shellcheck-alpine:v0.5.0` which I originally avoided because I thought it would expect musl is actually statically linked, so we can grab it from that quay image, and even get version pinning for free

omertuc: Sadly v0.5.0 is pretty old, you win some you lose some

rccrdpccl: Consider that each `COPY --from` would download the full image to the disk. At the moment a full build for the build image is taking 15GB of space in our local env machines (maybe only about 1 or 2GB are the actual output).

omertuc: You win some you lose even more? 

But seriously the alpine image is only 25MB so I guess it's not too bad

PR #4948, opened 2023-01-27
Allow removing the agent when BMH is deleted

carbonin:
In order to allow for full node life-cycle management for spoke clusters this commit will delete the agent resource associated with a BMH when the BMH is deleted and the new annotation
`bmac.agent-install.openshift.io/full-management` has also been set on the BMH.

Resolves https://issues.redhat.com/browse/MGMT-13504

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Test steps:
1. Annotate the BMH with `bmac.agent-install.openshift.io/full-management: "true"`
2. Check that the finalizer was added
3. Remove BMH
4. Validate agent was removed

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

Docs WIP

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -79,6 +79,8 @@ const (
 	BMH_INSPECT_ANNOTATION              = "inspect.metal3.io"
 	BMH_HARDWARE_DETAILS_ANNOTATION     = "inspect.metal3.io/hardwaredetails"
 	BMH_AGENT_IGNITION_CONFIG_OVERRIDES = "bmac.agent-install.openshift.io/ignition-config-overrides"
+	BMH_FINALIZER_NAME                  = "bmac.agent-install.openshift.io/deprovision"
+	BMH_DELETE_ANNOTATION               = "bmac.agent-install.openshift.io/remove-agent-and-node-on-delete"
 	MACHINE_ROLE                        = "machine.openshift.io/cluster-api-machine-role"
 	MACHINE_TYPE                        = "machine.openshift.io/cluster-api-machine-type"
 	MCS_CERT_NAME                       = "ca.crt"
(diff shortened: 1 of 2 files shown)


carbonin:
@avishayt @mhrivnak @filanov any suggestions for naming this annotation?

carbonin:
/hold

Depending on the direction we go with the larger epic this might not be needed.

Specifically we may not want to remove the agent with a finalizer on the BMH because we may want to actually delay removing the BMH so that we can use it to deprovision the host.

The idea I have is for the ZTP folks to add an annotation to the BMH that will tell BMAC to remove the spoke node, but once that's finished we'll use the BMH to clean and power off the host. For that to work we can't delete the BMH as the first step.

I'll leave this here for now, but may eventually close it if testing this new approach goes well.

carbonin:
/unhold

This is good to review now.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -214,6 +216,16 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
 
 	agent := r.findAgent(ctx, bmh)
 
+	// Finalizer is only needed if the user has indicated that they want full management of the BMH
+	// this means the spoke node will be removed when the BMH is deleted
```
mhrivnak, line 220: Does "full management" means *only* "the spoke node will be removed when the BMH is deleted"? If so, it would probably be more clear to give the annotation a name that more closely matches the behavior it induces.

mhrivnak: I like that. Though that brings up another good point; we're deleting the Agent, not the Node, right?

carbonin: In this PR, but it's just the first bit that I got done. I have a branch that adds node deletion, but that's still in progress.

They will use the same annotation unless we think there's value in doing one and not the other in some case.

mhrivnak: `remove-agent-and-node-on-delete` it is then. :smile: Unless you have a better idea.

mhrivnak (approved):

carbonin:
>Sorry, it looks like the commit/PR message is out of date. It mentions bmac.agent-install.openshift.io/full-management.

:man_facepalming: fixed

PR #4957, opened 2023-01-30
MGMT-13526: Fix wrong subscription name on pre-release versions

eliorerz:
When creating SNO cluster with LVM enabled alongside pre-release 4.12 OCP version there is wrong subscription-name (odf-lvm-operator instead of  lvms-operator)

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

/cc @eliorerz

### DIFF
--- a/internal/common/version.go
+++ b/internal/common/version.go
@@ -1,6 +1,10 @@
 package common
 
-import "github.com/hashicorp/go-version"
+import (
+	"strings"
+
+	"github.com/hashicorp/go-version"
+)
 
 func VersionGreaterOrEqual(version1, version2 string) (bool, error) {
 	v1, err := version.NewVersion(version1)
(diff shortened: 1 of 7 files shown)


osherdp (approved):

internal/common/version.go
```diff
@@ -13,3 +17,10 @@ func VersionGreaterOrEqual(version1, version2 string) (bool, error) {
 	}
 	return !v1.LessThan(v2), nil
 }
+
+func BaseVersionGreaterOrEqual(version1, version2 string) (bool, error) {
```
tsorya, line 21: Can you explain why we need it please?

eliorerz: Because it less intuitive to use 4.12.0-0-0, most of the time, we do not care if one pre-release is grater than other (pre-release) version, we care only about the comparison of the base version (most of the times).
Also, Not everyone familiar with that format and this leads to a lot of misses and inconsistencies.

osherdp (approved):

PR #4963, opened 2023-02-01
MGMT-13447: Fix LVM subscription name is empty on SNO CNV

eliorerz:
It seems that we are setting the subscription name before getting the operator dependencies, that causing the LVM subscription name to be empty when setting CNV alone via the API

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

/cc @osherdp 
/cc @gamli75 
/cc @tsorya

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -2789,27 +2789,34 @@ func (b *bareMetalInventory) getOLMOperators(cluster *common.Cluster, newOperato
 			return nil, common.NewApiError(http.StatusBadRequest, err)
 		}
 
-		// TODO - Need to find a better way for creating LVMO/LVMS operator on different openshift-version
-		if operator.Name == "lvm" {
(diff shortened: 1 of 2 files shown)


internal/bminventory/inventory.go
```diff
@@ -2789,27 +2789,34 @@ func (b *bareMetalInventory) getOLMOperators(cluster *common.Cluster, newOperato
 			return nil, common.NewApiError(http.StatusBadRequest, err)
 		}
 
+		operator.Properties = newOperator.Properties
+		monitoredOperators = append(monitoredOperators, operator)
+	}
+
+	operatorDependencies, err := b.operatorManagerApi.ResolveDependencies(cluster, monitoredOperators)
```
tsorya, line 2796: operatorDependencies has monitored operators that were asked for + all their dependencies? 
in case of cnv it will be cnv + lvm?

eliorerz: Yes, if CNV were selected it will store both CNV and LVM.
We cannot to do it at the beginning of that method, we need to create the monitored operators first.
In the future we need to do something with that function but this is different scope

PR #4972, opened 2023-02-08
MGMT-13549: Day1 implementation for K8S node labels

ori-amizur:
Contains 2 parts:
- Support setting node labels in BMH as annotations.  All of these annotations have the prefix "bmac.agent-install.openshift.io.node-label." in their key.  These are copied to a dedicated field "nodeLabels" in the spec of the agent CR.
- From the agent's "nodeLabels" field, copy to the DB in the associated host record to "node_labels" column.  This is done by the update host API call.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


internal/controller/controllers/agent_controller.go
```diff
@@ -1279,6 +1303,12 @@ func (r *AgentReconciler) updateIfNeeded(ctx context.Context, log logrus.FieldLo
 		InfraEnvID:       internalHost.InfraEnvID,
 		HostUpdateParams: &models.HostUpdateParams{},
 	}
+
+	hostUpdate, err = r.updateNodeLabels(log, internalHost, agent, params)
```
filanov, line 1307: suggesting to use a different parameter in case someone will add more code before this function he may forget that you are reassigning the value.

internal/controller/controllers/agent_controller.go
```diff
@@ -1250,6 +1250,30 @@ func (r *AgentReconciler) updateHostIgnition(ctx context.Context, log logrus.Fie
...
+	var err error
+	hostNodeLabels := make(map[string]string)
+	if host.NodeLabels != "" {
+		if err = json.Unmarshal([]byte(host.NodeLabels), &hostNodeLabels); err != nil {
+			log.WithError(err).Errorf("failed to unmarshal node labels for host %s infra-env %s", host.ID.String(), host.InfraEnvID.String())
+			return false, err
+		}
+	}
+	slicedNodeLabels := funk.Map(hostNodeLabels, func(key, value string) aiv1beta1.NodeLabel { return aiv1beta1.NodeLabel{Key: key, Value: value} })
+	if !funk.Subset(slicedNodeLabels, agent.Spec.NodeLabels) ||
```
tsorya, line 1263: Why subset? I think it should fully match it no?

internal/controller/controllers/agent_controller_test.go
```diff
@@ -485,6 +485,93 @@ var _ = Describe("agent reconcile", func() {
 		Expect(c.Get(ctx, key, agent)).To(BeNil())
 	})
 
+	Context("node labels", func() {
```
tsorya, line 488: add failure case please

avishayt:
Make sure that you validate the labels that you get as input
https://github.com/kubernetes/apimachinery/blob/fd8a60496be5e4ce19d586738adb48ac6fa66ef9/pkg/apis/meta/v1/validation/validation.go#L104

If this is missing for the REST API, then add there as well

internal/controller/controllers/agent_controller.go
ori-amizur: OK

api/v1beta1/agent_types.go
```diff
@@ -191,6 +191,15 @@ type AgentSpec struct {
 	IgnitionConfigOverrides string `json:"ignitionConfigOverrides,omitempty"`
 	// IgnitionEndpointTokenReference references a secret containing an Authorization Bearer token to fetch the ignition from ignition_endpoint_url.
 	IgnitionEndpointTokenReference *IgnitionEndpointTokenReference `json:"ignitionEndpointTokenReference,omitempty"`
+	// NodeLabels are the labels to be applied on the node associated with this agent
+	NodeLabels []NodeLabel `json:"nodeLabels,omitempty"`
```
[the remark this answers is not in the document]
ori-amizur: changed

internal/controller/controllers/agent_controller.go
ori-amizur: Changed to map, so irrelevant

internal/controller/controllers/agent_controller_test.go
ori-amizur: OK

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -82,6 +84,7 @@ const (
 	MCS_CERT_NAME                       = "ca.crt"
 	OPENSHIFT_MACHINE_API_NAMESPACE     = "openshift-machine-api"
 	ASSISTED_DEPLOY_METHOD              = "start_assisted_install"
+	NODE_LABEL_PREFIX                   = "node-label."
```
avishayt, line 87: We should prefix it with "bmac.agent-install.openshift.io/" like the other annotations we added.

avishayt: The downside is that it uses more characters from the label key, but I don't see any other choice.

avishayt: Right, good point about the slash.  On the one hand, it should be clear that this annotation "belongs to" BMAC/agent-install; on the other hand, we need to conform to the label key limitations.  @mhrivnak, what do you think?  How about "bmac.agent-install.nodel-label."?

ori-amizur: Changed the prefix to bmac.agent-install.openshift.io.node-label

internal/controller/controllers/agent_controller.go
```diff
@@ -1250,6 +1250,32 @@ func (r *AgentReconciler) updateHostIgnition(ctx context.Context, log logrus.Fie
 	return err
 }
 
+func (r *AgentReconciler) updateNodeLabels(log logrus.FieldLogger, host *common.Host, agent *aiv1beta1.Agent, params *installer.V2UpdateHostParams) (bool, error) {
```
tsorya, line 1253: I assume if we call agent -> host lets change it to host labels?

internal/controller/controllers/agent_controller.go
```diff
@@ -1279,6 +1305,14 @@ func (r *AgentReconciler) updateIfNeeded(ctx context.Context, log logrus.FieldLo
 		InfraEnvID:       internalHost.InfraEnvID,
 		HostUpdateParams: &models.HostUpdateParams{},
 	}
+
+	nodesUpdated, err := r.updateNodeLabels(log, internalHost, agent, params)
```
tsorya, line 1309: not enough just to use hostUpdate ? and remove it from 1302 ?

ori-amizur: Was requested by @filanov https://github.com/openshift/assisted-service/pull/4972/files#diff-ec0c8ee92a41e93de3f6906e7ddd408c175553ef5d57b247e7d0b56ec7a7be47

internal/controller/controllers/agent_controller.go
ori-amizur: To be exact, this function just modifies the update params. It doesn't modify the host here.

internal/controller/controllers/agent_controller.go
tsorya: seems like each "if" is reassigning
IMO i would go with 
```suggestion
	hostUpdate, err := r.updateNodeLabels(log, internalHost, agent, params)
```
and remove 	hostUpdate := false, then nobody can think that you reassign it

PR #4977, opened 2023-02-12
"MGMT-13133: MetalLB operator for assisted installer"

ydayagi:
Add another optional operator during installation
https://github.com/openshift/assisted-service/pull/4708
https://issues.redhat.com/browse/MGMT-13133

- [ ] Enhancement

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/operators/metallb/config.go
+++ b/internal/operators/metallb/config.go
@@ -1,0 +1,10 @@
+package metallb
+
+type Config struct {
+	MetalLBMinOpenshiftVersion string `envconfig:"METALLB_MIN_OPENSHIFT_VERSION" default:"4.11.0"`
+}
+
+type Properties struct {
+	ApiIP     string `json:"api_ip"`
+	IngressIP string `json:"ingress_ip"`
+}
--- a/internal/operators/metallb/manifest.go
+++ b/internal/operators/metallb/manifest.go
@@ -1,0 +1,201 @@
+package metallb
+
+import (
+	"bytes"
+	"text/template"
+
+	"github.com/openshift/assisted-service/internal/common"
+)
+
+// Manifests returns manifests needed to deploy MetalLB
+func Manifests(cluster *common.Cluster) (map[string][]byte, []byte, error) {
+	metalLBSubscription, err := getSubscription()
+	if err != nil {
+		return nil, nil, err
+	}
+	metalLBNamespace, err := getNamespace()
+	if err != nil {
+		return nil, nil, err
+	}
+	metalLBOperatorGroup, err := getOperatorGroup()
+	if err != nil {
+		return nil, nil, err
+	}
+	metalLBConfigs, err := getMetalLBOperand(cluster)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	openshiftManifests := make(map[string][]byte)
+
+	openshiftManifests["50_openshift-metallb_ns.yaml"] = metalLBNamespace
+	openshiftManifests["50_openshift-metallb_operator_group.yaml"] = metalLBOperatorGroup
+	openshiftManifests["50_openshift-metallb_subscription.yaml"] = metalLBSubscription
+
+	return openshiftManifests, metalLBConfigs, nil
+}
+
+func getSubscription() ([]byte, error) {
+	data := map[string]string{
+		"OPERATOR_NAMESPACE":         Operator.Namespace,
+		"OPERATOR_SUBSCRIPTION_NAME": Operator.SubscriptionName,
+	}
+	return executeTemplate(data, "MetalLBSubscription", MetalLBSubscription)
+}
+
+func getNamespace() ([]byte, error) {
+	data := map[string]string{
+		"OPERATOR_NAMESPACE": Operator.Namespace,
+	}
+	return executeTemplate(data, "MetalLBNamespace", MetalLBNamespace)
+}
+
+func getOperatorGroup() ([]byte, error) {
+	data := map[string]string{
+		"OPERATOR_NAMESPACE":  Operator.Namespace,
+		"OPERATOR_GROUP_NAME": Operator.SubscriptionName,
+	}
+	return executeTemplate(data, "MetalLBOperatorGroup", MetalLBOperatorGroup)
+}
+
+func getMetalLBOperand(cluster *common.Cluster) ([]byte, error) {
+	properties, err := parsePropertiesField(cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	data := map[string]string{
+		"OPERATOR_NAMESPACE": Operator.Namespace,
+		"API_IP":             properties.ApiIP,
+		"INGRESS_IP":         properties.IngressIP,
+	}
+	operand := MetalLBOperandPartA
+
+	if properties.ApiIP != "" {
+		operand += apiIpAddressPoolResource
+	}
+
+	if properties.IngressIP != "" {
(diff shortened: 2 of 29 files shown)


flaper87:
@ydayagi thanks for the PR. Could you mark it as draft until the enhancement goes in? :pray:

flaper87:
/retitle "MGMT-13133: MetalLB operator for assisted installer"

internal/operators/metallb/config.go
```diff
@@ -0,0 +1,12 @@
+package metallb
+
+type Config struct {
+	MetalLBCPUPerHost          int64  `envconfig:"METALLB_CPU_PER_HOST" default:"1"`
```
tsorya, line 4: I don't remember if we will add those values to hw node requirements. I think in case of metal lb we should not require more cpu or memory. Can you check that we don't add those values to hw requirements? or maybe you think we need to add them?

internal/operators/metallb/metallb_operator.go
```diff
@@ -0,0 +1,194 @@
...
+	if properties == "" {
+		return propertiesObj, fmt.Errorf("Properties field for operator %s can not be empty", Operator.Name)
+	}
+
+	err := json.Unmarshal([]byte(properties), &propertiesObj)
+	if err != nil {
+		return propertiesObj, err
+	}
+
+	ip := net.ParseIP(propertiesObj.ApiIP)
```
tsorya, line 183: From manifests we support only ipv4.  Should not we validate that ip is ipv4?

internal/operators/metallb/config.go
ydayagi: @flaper87 what should we do

internal/operators/metallb/metallb_operator.go
ydayagi: @flaper87?

internal/operators/metallb/manifest.go
```diff
@@ -0,0 +1,187 @@
...
+---
+apiVersion: metallb.io/v1beta1
+kind: IPAddressPool
+metadata:
+  name: metallb-ingress-ip
+  namespace: "{{.OPERATOR_NAMESPACE}}"
+spec:
+  autoAssign: false
+  addresses:
+    - {{.INGRESS_IP}}/32
```
flaper87, line 139: Let's make sure these 2 are applied *only* if the IPs are provided.

I think it's fine to apply the following:

- L2Advertisement
- Services
- Operand

The service won't be fully configured and the L2Advertisement won't have an `IPAddressPool` to advertise until they are actually created upon relocation. This should be fine, may be worth testing to see what happens, tho.

flaper87 (commented):
Overall, this looks good. Dropped a few comments to align the PR better with the enhancement. :muscle:

internal/operators/metallb/config.go
ydayagi: @flaper87 validations? at the moment when I enable metallb-operator it automatically requires additional CPU per node and about 1GB memory per node

internal/operators/metallb/manifest.go
ydayagi: in case IPs are not provided we do not reach this point

internal/operators/metallb/metallb_operator.go
```diff
@@ -0,0 +1,194 @@
...
+	if ip == nil {
+		return propertiesObj, fmt.Errorf("MetalLB API IP is not valid")
+	}
+
+	ip = net.ParseIP(propertiesObj.IngressIP)
+	if ip == nil {
+		return propertiesObj, fmt.Errorf("MetalLB Ingress IP is not valid")
+	}
```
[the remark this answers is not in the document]
ydayagi: At the moment it is enough for one of the IPs to be missing or not valid in order to fail the operation. Seems to me exactly what you are asking in your comment

internal/operators/metallb/config.go
ydayagi: in GetPreflightRequirements we return these values

tsorya: In that case i think we should set 0 , no reason to add one more core for it and more memory. Storage operators are cpu consuming that's why we add it but metalb should not have those problems i assume

internal/operators/metallb/metallb_operator.go
ydayagi: @flaper87 @tsorya I left the code is. I think that the code should not validate type of IP. If our manifests will support IPv6 in the future then no changes will be required here. If IPv6 is not supported then there should be a location, earlier in time, that validates it.

internal/operators/metallb/config.go
ydayagi: @flaper87 @tsorya I removed the hardware requirements. I think it is the right thing to do.

internal/operators/metallb/metallb_operator.go
tsorya: Just for me to understand. This ip can be any ip right or host should have this subnet configured? For example why adding ipv6 ip will fail earlier? Where do we verify it?

ydayagi: @tsorya  any ip that metallb operator can use is ok. so far I did not find any thing that prevents metallb for answering any ip on any interface (NIC/device).
You wrote that 'From manifests we support only ipv4'. Who limits that? if no one, then this limitation does not exist. If someone somewhere limits that then this is where the validation is performed. I think we should not add validations. It only makes the code more sensitive to coding mistakes and changes in the future. But your call

tsorya: The idea of assisted is actually add validation in order to make user life easier. We try to validate as maximum as we can actually. Currently you add ip as {ip}/32 that makes it valid only for ipv4. I think that we need or to change code to support both types of ips or to validate that ip is valid.

ydayagi: @tsorya @flaper87 done

tsorya: can you use https://github.com/openshift/assisted-service/blob/master/internal/network/family_utils.go#L8 ? just to make less code

ydayagi: @tsorya really? less code? what are we saving? a line? and instead of using a builtin, maintained, standard library use our implementation?

tsorya: Lets say we already have it and use it all over the code, why not to use it? 
 if i understand right it will be changed to:
`if !IsIPv4Addr(propertiesObj.ApiIP) {
   return propertiesObj, fmt.Errorf("MetalLB API IP is not a valid IPv4 address")
}
`
It is less enough code as for me and  more readable.

ydayagi: done

internal/operators/metallb/manifest.go
flaper87: If the IPs are not provided we still need to deploy metallb if requested. We won't configure the `IPAddressPool` but we will configure the rest. Let's make sure this happens, pls

flaper87:
I have one comment on the behavior in case IPs are not provided.

eliorerz:
I think we need e2e test before merging that PR (same as we have for other operators).

eliorerz:
Outdated - opened a new PR on https://github.com/openshift/assisted-service/pull/5274

PR #4987, opened 2023-02-19
MGMT-13308: Adding events pagination

danmanor:
Currently events are retrieved by the service as a list of events which match the query. This PR is implementing events pagination. As a result of events retrieved in portions, all of the filtering that is currently done by the UI has been also implemented.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/events/event.go
```diff
@@ -20,6 +21,8 @@ import (
 	"gorm.io/gorm"
 )
 
+const defaultPageSize int64 = 10
```
eranco74, line 24: In order to not break anything consider starting with a very big default (I think that currently we return 200 events)

internal/events/event.go
```diff
@@ -314,14 +317,30 @@ func (e Events) queryEvents(ctx context.Context, selectedCategories []string, cl
...
 	if result == nil { //non supported option
-		return make([]*common.Event, 0), nil
+		return make([]*common.Event, 0), swag.Int64(0), nil
 	}
 
 	var events []*common.Event
-	return events, WithIDs(result).Find(&events).Error
+
+	result = WithIDs(result)
+	totalRecords := result.Find(&events).RowsAffected
```
eranco74, line 326: you don't really want the events here just the count
```suggestion
    var totalRecords int64 
    result.Count(totalRecords)
```

internal/events/event.go
```diff
@@ -245,7 +248,7 @@ func (e *Events) V2AddMetricsEvent(ctx context.Context, clusterID *strfmt.UUID,
 	e.v2SaveEvent(ctx, clusterID, hostID, infraEnvID, name, models.EventCategoryMetrics, severity, msg, eventTime, requestID, props...)
 }
 
-func (e Events) queryEvents(ctx context.Context, selectedCategories []string, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID) ([]*common.Event, error) {
+func (e Events) queryEvents(ctx context.Context, selectedCategories []string, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID, limit, offset *int64) ([]*common.Event, *int64, error) {
```
eranco74, line 251: no reason for totalRecords to be a pointer
```suggestion
func (e Events) queryEvents(ctx context.Context, selectedCategories []string, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID, limit, offset *int64) ([]*common.Event, int64, error) {
```

internal/events/event.go
```diff
@@ -314,14 +317,30 @@ func (e Events) queryEvents(ctx context.Context, selectedCategories []string, cl
...
+	}
+
+	// if offset not specified, we return the first page. If offset is specified but not valid, we return an error.
+	if offset == nil {
+		offset = swag.Int64(0)
+	} else if *offset < 0 || *offset > totalRecords {
+		return nil, nil, NewInvalidOffsetError()
+	}
+
+	return events, swag.Int64(totalRecords), result.Offset(int(*offset)).Limit(int(*limit)).Find(&events).Error
```
eranco74, line 340: Consider
```suggestion
       err := result.Offset(int(*offset)).Limit(int(*limit)).Find(&events).Error
	return events, totalRecords, err 
```

eranco74 (commented):
Looks good, just nits

internal/events/event.go
rccrdpccl: I think currently is unlimited, so I would recommend setting -1 as default. Assisted-scraper truncates at 2000 (for some unknown reason)

internal/cluster/cluster.go
```diff
@@ -1023,7 +1023,7 @@ func (m *Manager) createClusterDataFiles(ctx context.Context, c *common.Cluster,
 	// we don't want to stop on error
 	_ = m.uploadDataAsFile(ctx, log, cluster, fileName, objectHandler)
 
-	events, err := m.eventsHandler.V2GetEvents(ctx, c.ID, nil, nil)
+	events, _, err := m.eventsHandler.V2GetEvents(ctx, c.ID, nil, nil, common.UnlimitedEvents, common.NoOffsetEvents)
```
rccrdpccl, line 1026: What do you think about definining option object and pass that instead? That would keep signature smaller and clearer even if we add additional options. Something like:
```
type QueryOptions struct {
   PageNumber int64
   ItemsPerPage int64
}
```
maybe with better naming :sweat_smile: 
Also it could be recycled for other APIs (like paginating clusters)

client/events/v2_list_events_parameters.go
```diff
@@ -92,6 +92,18 @@ type V2ListEventsParams struct {
...
+
+	   The maximum number of records to retrieve.
+	*/
+	Limit *int64
+
+	/* Offset.
+
+	   Number of records to skip before starting to return the records.
+	*/
+	Offset *int64
```
rccrdpccl, line 105: Wouldn't be easier to work with pages instead of `limit` and `offset`? It's very low level, I think for API we can use something like `itemsPerPage` and `pageNumber` - I find it more intuitive and easier to use from the client side (just define itemsPerPage and increase the page number, as opposed as calculating offset)

internal/events/event.go
```diff
@@ -314,14 +317,30 @@ func (e Events) queryEvents(ctx context.Context, selectedCategories []string, cl
...
+
+	// if limit not specified or not valid, we set it to default. -1 is a valid value for limit which means no limit.
+	if limit == nil || *limit < -1 || *limit == 0 {
+		limit = swag.Int64(defaultPageSize)
+	}
+
+	// if offset not specified, we return the first page. If offset is specified but not valid, we return an error.
+	if offset == nil {
+		offset = swag.Int64(0)
+	} else if *offset < 0 || *offset > totalRecords {
```
rccrdpccl, line 336: Why would we return error with offset larger than totalRecords? It would just return 0 results which I think it's acceptable (same behaviour as SQL). In any case this validation would be easier with the page-oriented approach - please consider validating at swagger level what we can.

swagger.yaml
```diff
@@ -3404,6 +3404,16 @@ paths:
           type: string
           format: uuid
           required: false
+        - in: query
```
rccrdpccl, line 3407: We can add validation in swagger, reducing boiler plate code https://swagger.io/docs/specification/data-models/data-types/#range

internal/events/event.go
danmanor: It doesn't work for me, not sure why. This is the only way I could find that gets the accurate count @eranco74

rccrdpccl, line 326: How is this working under the bonnet? Are we actually executing a full unlimited query, and then again another one with limit below?

danmanor: I thought it would be good for returning nil when an error occurs, WDYT?

danmanor: It doesn't work for me, not sure why. This is the only way I could find that gets the accurate count @eranco74

client/events/v2_list_events_parameters.go
danmanor: AFAIK, `limit` and `offset` works better for the UI, correct? @ammont82

ammont82: Yes, for us works better limit and offset for the UI

swagger.yaml
```diff
@@ -3404,6 +3404,16 @@ paths:
...
+          name: offset
+          description: Number of records to skip before starting to return the records.
+          type: integer
+          required: false
+        - in: query
+          name: limit
```
gamli75, line 3413: question - limit == number of records per page?

gamli75: another question - are we going to define a default number for the limit? I mean if the user will call the event API without setting the offset and the limit - will he get the same data as we have today with all the events(even if we have 1 million events)?

danmanor: > question - limit == number of records per page?
Maximum number of records per page yes @gamli75

danmanor: > another question - are we going to define a default number for the limit? I mean if the user will call the event API without setting the offset and the limit - will he get the same data as we have today with all the events(even if we have 1 million events)?

Yes, currently the logic that seems most reasonable for me and the UI as well AFAIK is:

- `limit` not set -> no limit (limit = -1)
- `limit` is set but is invalid value (limit < -1 or limit = 0) -> limit = < some default value >
- `offset` not set -> start from the first event (offset = 0)
- `offset` is set but invalid (offset < 0) -> start from the first event (offset = 0)
- `offset` is set but larger than the total number of matching records in the DB -> no results will be returned

WDYT? @gamli75

internal/events/api/event.go
```diff
@@ -37,7 +37,7 @@ type Sender interface {
 //go:generate mockgen -source=event.go -package=api -destination=mock_event.go
 type Handler interface {
 	Sender
-	V2GetEvents(ctx context.Context, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID, categories ...string) ([]*common.Event, error)
+	V2GetEvents(ctx context.Context, clusterID *strfmt.UUID, hostsID []strfmt.UUID, infraEnvID *strfmt.UUID, limit, offset *int64, severity []string, message *string, deletedHosts, clusterLevel *bool, categories ...string) ([]*common.Event, *common.EventSeverityCount, error)
```
rccrdpccl, line 40: I would recommend trying to group arguments to have less in the signature.
This would allow for easier code reading and easier testing, making it less brittle: by not having to change the signature adding/removing/changing a filter would break hundreds of tests that do use the method but they don't rely on the filters.

Also in the response we can group them and allow for extensibility, for example returning a response where we could get events from:
```
response, err := events.V2GetEvents(...)
events := response.GetEvents()
severityCount := response.GetStats().GetSeverityCount()
```

with the above change, a call like:
```
evs, _, err := theEvents.V2GetEvents(context.TODO(), clusterID, hostsID, infraEnvID, common.UnlimitedEvents, common.NoOffsetEvents, nil, nil, nil, nil)
```
could be turned into something like:
```
eventFilters := events.DefaultEventFiltersWithIDs(clusterID, hostID, infraEnvID)
response, err := theEvents.V2GetEvents(context.TODO(), eventFilters)
evs := response.GetEvents()
```
and even when adding some other filters would still be very readable (and we could also have utility methods to return some pre-initialized common filters)
```
eventFilters := events.DefaultEventFiltersWithIDs(clusterID, hostID, infraEnvID)
eventFilters.Limit = 100
eventFilters.Offset = 20
response, err := theEvents.V2GetEvents(context.TODO(), eventFilters)
evs := response.GetEvents()
```

internal/events/event.go
```diff
@@ -245,83 +249,221 @@ func (e *Events) V2AddMetricsEvent(ctx context.Context, clusterID *strfmt.UUID,
 	e.v2SaveEvent(ctx, clusterID, hostID, infraEnvID, name, models.EventCategoryMetrics, severity, msg, eventTime, requestID, props...)
 }
 
-func (e Events) queryEvents(ctx context.Context, selectedCategories []string, clusterID *strfmt.UUID, hostID *strfmt.UUID, infraEnvID *strfmt.UUID) ([]*common.Event, error) {
+func (e Events) queryEvents(ctx context.Context, selectedCategories []string, clusterID *strfmt.UUID, hostsID []strfmt.UUID, infraEnvID *strfmt.UUID, limit, offset *int64, severity []string, message *string, deletedHosts, clusterLevel *bool) ([]*common.Event, *common.EventSeverityCount, error) {
+
```
rccrdpccl, line 253: this method is 200+ lines long, it would be very difficult to read. I would recommend extracting some functionality for better readability

internal/events/event.go
```diff
@@ -24,6 +26,8 @@ var DefaultEventCategories = []string{
 	models.EventCategoryUser,
 }
 
+var defaultEventLimit int64 = 5000
```
rccrdpccl, line 29: Why default to 5000?

internal/events/event.go
```diff
@@ -233,23 +237,114 @@ func (e *Events) V2AddMetricsEvent(ctx context.Context, clusterID *strfmt.UUID,
...
+		}
+
+		// deleted hosts, cluster level and specific hosts events
+		if swag.BoolValue(clusterLevel) && swag.BoolValue(deletedHosts) {
+			db = db.Where("hosts.deleted_at IS NOT NULL").
+				Or("events.host_id IS NULL")
+		}
+
+		// technical condition for the next one in case none of the latter occurred
+		if !swag.BoolValue(clusterLevel) && !swag.BoolValue(deletedHosts) && hostsID != nil {
```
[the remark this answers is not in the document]
danmanor: This is a technical addition to the query that makes the code shorter, because before `Or` must be `Where`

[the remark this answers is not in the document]
rccrdpccl: But wouldn't `WHERE FALSE` return always no result?

internal/events/event.go
danmanor: There is no good reason, I coordinated with the UI that in this case we will set `Limit` to some default, I chose 5000 as it is usually large enough to retrieve all cluster's events, but not too large

rccrdpccl: Wouldn't UI use limit in the future? Why do we need to break compatibility with the past and limit requests that are not using limit parameters?
AFAIK only `assisted-events-scrape` uses this (apart from UI), and it takes only the first 2000 results reverse ordered by event_time, but I still wouldn't break compatibility.

internal/events/event.go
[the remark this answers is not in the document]
danmanor: It does, but it will always hit the condition after, therefore will be added with `Or`

internal/events/event.go
danmanor: If no `Limit` is specified (nil), it is set to -1 (unlimited). If `Limit` is set improperly (smaller than -1 or 0), I set it to default.

swagger.yaml
```diff
@@ -3447,17 +3447,52 @@ paths:
           format: uuid
           required: False
         - in: query
-          name: host_id
+          name: hosts_id
```
osherdp, line 3450: isn't it an API change?
regardless of UI being adjustable, there might be other consumers of this endpoint which might be affected

osherdp, line 3450: might be more English-like
```suggestion
          name: host_ids
```

osherdp: By only adding the parameter ``hosts_id`` to the existing parameters, documenting that it takes precedence over ``host_id``
``host_id`` can be documented to be deprecated, and also annotated ``deprecated: true`` (https://swagger.io/docs/specification/describing-parameters/)

internal/events/event.go
rccrdpccl: I see, sorry I didn't notice that, got tricked by the naming.

internal/events/event.go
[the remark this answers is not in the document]
rccrdpccl: Ok, can we improve the comment to make this clearer please, so other people are not as confused as me?

swagger.yaml
danmanor: I believe it can be annotated only in `OA 3.0 API` I can add (DEPRECATED) in the description

osherdp: no problem. just search the swagger for similar cases where we've changed parameters to their plural variant

pkg/app/middleware.go
```diff
@@ -70,6 +70,10 @@ func SetupCORSMiddleware(handler http.Handler, domains []string) http.Handler {
 		AllowedHeaders: []string{
 			"Authorization",
 			"Content-Type",
+			"Severity-Count-Info",
+			"Severity-Count-Warning",
+			"Severity-Count-Error",
+			"Severity-Count-Critical",
```
danmanor, line 76: This change should hopefully let cross origin requests access this resources

swagger.yaml
```diff
@@ -3445,19 +3445,60 @@ paths:
           description: The cluster to return events for.
           type: string
           format: uuid
-          required: False
+          required: false
         - in: query
           name: host_id
-          description: A host in the specified cluster to return events for.
+          description: A host in the specified cluster to return events for (DEPRECATED).
```
osherdp, line 3451: let's follow the existing conventions:
```suggestion
          description: (DEPRECATED) A host in the specified cluster to return events for.
```

osherdp, line 3451:
```suggestion
          description: A host in the specified cluster to return events for (DEPRECATED. Use `host_ids` instead).
```

osherdp (changes requested):

internal/events/event.go
```diff
@@ -261,64 +357,123 @@ func (e Events) queryEvents(ctx context.Context, selectedCategories []string, cl
...
 
-	if result == nil { //non supported option
-		return make([]*common.Event, 0), nil
+	//non supported option
+	return nil
+}
+
+func preparePaginationParams(limit, offset *int64) (*int64, *int64) {
+	if limit == nil {
+		// If limit is not valid (smaller than 1 or equals 0), we set it -1 (no limit).
```
osherdp, line 407:
```suggestion
		// If limit is not provided, we set it -1 (no limit).
```

internal/events/event.go
```diff
@@ -261,64 +357,123 @@ func (e Events) queryEvents(ctx context.Context, selectedCategories []string, cl
...
+		limit = swag.Int64(-1)
+	} else if *limit < -1 || *limit == 0 {
+		// If limit not valid (smaller than 1 or equals 0), we set it a default (currently 5000).
+		limit = swag.Int64(defaultEventLimit)
+	}
```
osherdp, line 412: it's a bit of a confusing behavior. I would do:
* if ``limit`` is null (not provided), let it use the default
* if ``limit`` is less than zero, be infinite 
* if someone wants ``limit=0``, that's ok. the code should return an empty list by itself

internal/events/event_test.go
```diff
@@ -503,15 +519,628 @@ var _ = Describe("Events library", func() {
...
+				time.Date(2023, 2, 21, 20, 0, 0, 0, time.UTC),
+			)
+			theEvents.V2AddEvent(
+				ctx,
+				&cluster1,
+				&host,
+				&infraEnv1,
+				eventgen.ClusterInstallationFailedEventName,
+				models.EventSeverityInfo,
+				"Installtion failed",
```
osherdp, line 557:
```suggestion
				"Installation failed",
```

osherdp (changes requested):
got over most of the code (anything except the newly added tests)
looks very good!

osherdp (changes requested):

internal/events/event.go
```diff
@@ -233,23 +237,109 @@ func (e *Events) V2AddMetricsEvent(ctx context.Context, clusterID *strfmt.UUID,
...
+		err = rows.Scan(&total, &severity)
+		if err != nil {
+			return nil, err
+		}
+		if severity == models.EventSeverityInfo ||
+			severity == models.EventSeverityWarning ||
+			severity == models.EventSeverityError ||
+			severity == models.EventSeverityCritical {
+			eventSeverityCount[severity] = int64(total)
+		}
```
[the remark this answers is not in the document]
danmanor: I figured it would be safer. WDYT?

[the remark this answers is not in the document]
osherdp: we fetch this info from the database, so if it contains an illegal value it's already game-over

osherdp (approved):
Looks very good!

PR #4989, opened 2023-02-20
MGMT-13659: Design a feature support mechanism that supports different feature-support criteria

eliorerz:
### Background:
Currently, we have an API for feature support that returns a map containing features-support-level per OCP version.
The API should support feature support per architecture (or maybe more cluster properties in the future).
For example:
Feature X on OCP 4.10 is only supported on x86 and supported on arm64 on 4.11.
In the current solution we will have to duplicate the entire [list](https://github.com/openshift/assisted-service/blob/master/internal/featuresupport/support_levels_list.go#L16) for each supported architecture.

### The current solution has the following issues:
- static - holds the entire data per version and returned to the user as is. Meaning that when adding a new version we have to add all the supported and unsupported features manually.
- The user have to get the entire feature-support list even if he wants only feature - version.
- Duplication - Adding more attributes to filter by is expensive, meaning that for each attribute (different from OpenshiftVersion) we will have to duplicate the entire features list for each attribute value
- Does not support adding hierarchy of another attribute (e.g. architecture) without breaking the api and duplicating the entire list n times.

### The solution
- Add two new API endpoints that will replace the old one (`/v2/feature-support-levels` - Deprecated). 
  The first one is `GET /v2/support-levels/features`, these endpoints will return objects containing features and their support level. 
  Another one is `GET /v2/support-levels/architectures`, these endpoints will return objects containing CPU architectures and their support level.
  Both of those endpoints can be filtered by `openshift_version` and/or `cpu_architecture`.
- Create a GO interface for feature-support level that will allow us to filter the support level of both features and architectures
  by multiple arguments.
- Create a feature-specific object (per feature ID) that will implement the described interface and will return the support level
  of that specific feature.
- The new API will not return a list of filtered architectures or features. The primary filter will be a given `openshift_version`
  (mandatory) and more optional arguments (`cpu_architecture`)


osherdp:
I feel like it should be an enhancement doc first, because we have lots of API considerations to make (especially in regards to breaking changes, but also to make sure we handle all probable cases in the future)

swagger.yaml
```diff
@@ -3715,36 +3794,51 @@ definitions:
...
+      - 'PLATFORM_SELECTION'
+      - 'SCHEDULABLE_MASTERS'
+      - 'AUTO_ASSIGN_ROLE'
+      - 'CUSTOM_MANIFEST'
+      - 'DISK_ENCRYPTION'
+      - 'CLUSTER_MANAGED_NETWORKING_WITH_VMS'
+      - 'X86_64_ARCHITECTURE'
+      - 'ARM64_ARCHITECTURE'
+      - 'PPC64LE_ARCHITECTURE'
+      - 'S390X_ARCHITECTURE'
```
avishayt, line 3832: These should be in a separate enum.  It looks like it makes the diff smaller, but is it the best way?

eliorerz: Moving those will not break the API? Also, they are still representing a features, just from a "different kind"

internal/bminventory/inventory.go
```diff
@@ -376,7 +376,7 @@ func getDefaultNetworkType(params installer.V2RegisterClusterParams) (*string, e
 		return params.NewClusterParams.NetworkType, nil
 	}
 
-	isOpenShiftVersionRecentEnough, err := common.BaseVersionGreaterOrEqual(swag.StringValue(params.NewClusterParams.OpenshiftVersion), minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes)
+	isOpenShiftVersionRecentEnough, err := common.BaseVersionGreaterOrEqual(minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes, swag.StringValue(params.NewClusterParams.OpenshiftVersion))
```
gamli75, line 379: why did you switch the order of the parameters?

internal/bminventory/inventory.go
```diff
@@ -798,9 +798,7 @@ func (b *bareMetalInventory) getNewClusterCPUArchitecture(newClusterParams *mode
 		return common.DefaultCPUArchitecture, nil
 	}
 
-	if !swag.BoolValue(newClusterParams.UserManagedNetworking) && !featuresupport.IsFeatureSupported(swag.StringValue(newClusterParams.OpenshiftVersion),
-		models.FeatureSupportLevelFeaturesItems0FeatureIDARM64ARCHITECTUREWITHCLUSTERMANAGEDNETWORKING) {
-
+	if !swag.BoolValue(newClusterParams.UserManagedNetworking) && !featuresupport.IsArchitectureSupported(models.ArchitectureSupportLevelIDARM64ARCHITECTURE, *newClusterParams.OpenshiftVersion) {
```
gamli75, line 801: not sure that you didn't change the behavior here - we should test it based on the old code 
4.10 - ARM - supported 
4.10 - ARM - with cluster manage network - unsupported 
4.11 - ARM - with cluster manage network - supported

internal/featuresupport/feature_support_level.go
```diff
@@ -0,0 +1,91 @@
...
+	featureSupportList := models.SupportLevels{}
+
+	for _, feature := range features {
+		featureID := feature.GetId()
+		featureSupportList[string(featureID)] = feature.GetSupportLevel(openshiftVersion)
+	}
+	return featureSupportList
+}
+
+func GetFeatureSupportList(filters SupportLevelFilters) models.SupportLevels {
```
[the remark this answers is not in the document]
eliorerz: My goal was to avoid a situation where we would have to update every feature signature if we wanted to add a parameter later on.
If you examine the features object closely, you'll see that the CPU architecture is often irrelevant, so its inclusion in the code can be aesthetically unappealing.

internal/bminventory/inventory.go
eliorerz: logically it was the opposite. I noticed that during that PR and updated it.
It should check if `params.NewClusterParams.OpenshiftVersion` >= `minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes`.

internal/bminventory/inventory.go
```diff
@@ -6197,11 +6212,15 @@ func (b *bareMetalInventory) GetKnownApprovedHosts(clusterId strfmt.UUID) ([]*co
...
+		OpenshiftVersion: version,
+		CPUArchitecture:  &cpuArchitecture,
+	}
+
 	return cpuArchitecture != common.MultiCPUArchitecture &&
 		cpuArchitecture != common.NormalizeCPUArchitecture(runtime.GOARCH) &&
 		common.PlatformTypeValue(platform) == models.PlatformTypeBaremetal &&
-		featuresupport.IsFeatureSupported(version,
-			models.FeatureSupportLevelFeaturesItems0FeatureIDARM64ARCHITECTUREWITHCLUSTERMANAGEDNETWORKING)
+		featuresupport.IsFeatureSupported(models.FeatureSupportLevelIDCLUSTERMANAGEDNETWORKING, featureSupportFilters)
```
eliorerz, line 6223: Not sure if the logic changed here

gamli75 (approved):

PR #5000, opened 2023-02-23
MGMT-13780: Write a feature-support-level enhancement doc

eliorerz:
As part of adding feature support per architecture alongside OCP version, an API change is needed.
This doc is describing what change is needed to be done and an proposal to solve this issue.

Note: An implementation proposal is available at https://github.com/openshift/assisted-service/pull/4989 

/cc @eliorerz

### DIFF
--- a/docs/enhancements/add-feature-support-by-generic-attributes.md
+++ b/docs/enhancements/add-feature-support-by-generic-attributes.md
@@ -1,0 +1,239 @@
+---
+title: Add supported-features endpoint
+authors:
+- "@eliorerz"
+  creation-date: 2023-02-23
+  last-updated: 2023-03-05
+---
+
+# Add supported-features endpoint
+
+## Summary
+
+In the current implementation of Assisted Installer, there is an API that allows the user to get a list of objects
+containing a list of features attached to an OpenShift version and its support-level.
+Whenever a user wants to know if a feature is supported in an OpenShift version, the user needs to call
+the feature-support-list API, get the entire list and filter the results to find the specific feature for that version.
+As part supporting multiple CPU architectures on Assisted Installer, we want the ability to add another dimension to the
+returned object - CPU architecture, so we can determine if a feature is supported or not by two parameters instead of one.
+
+
+## Motivation
+Adding another attribute to the feature-support-level API is indeed required but in its current implementation there are
+few flaws (from my perspective):
+1. The current implementation forces us to duplicate the entire features list for each cpu architecture (API change).\
+   For example if for `OpenshiftVersion = 4.12` we currently hold 10 features so adding just ARM64 as a CPU architecture to
+   that equation will require us to duplicate the 10 features (for that architecture) and set the support-level for both
+   x86 and arm64 (note that we also have `s390x` & `ppc64le` architectures).
+2. The list is statically defined in assisted-service code so on new OCP version we will have to look at the current release 
+   or duplicate the entire features from the previous version even if it supported or not supported for few past versions.
+3. It is hard to expand, beside of adding CPU architecture to that equation if in the future we will want to filter by a
+   third attribute we will have to duplicate everything again (API change).
+4. The user is have to get the entire feature-support list even if he needs to know just the support of a single feature.
+
+These four reasons led me to the realization that we must create a more flexible and easier-to-maintain API.
+
+### Goals
+
+- Add another dimension (of CPU architecture) to the feature-supported endpoint
+- Improve the maintainability of this code and less likely to require updates
+- Change the API to allow the user to get a list for a specific openshift_version without retrieving the entire list.
+- Design an easier way to add a new dimension to that API.
+
+## Proposal
+
+- Add two new API endpoints that will replace the old one (`/v2/feature-support-levels` - Deprecated). 
+  The first one is `GET /v2/support-levels/features`, these endpoints will return objects containing features and their support level. 
+  Another one is `GET /v2/support-levels/architectures`, these endpoints will return objects containing CPU architectures and their support level.
+  Both of those endpoints can be filtered by `openshift_version` and/or `cpu_architecture`.
+- Create a GO interface for feature-support level that will allow us to filter the support level of both features and architectures
+  by multiple arguments.
+- Create a feature-specific object (per feature ID) that will implement the described interface and will return the support level
+  of that specific feature.
+- The new API will not return a list of filtered architectures or features. The primary filter will be a given `openshift_version`
+  (mandatory) and more optional arguments (`cpu_architecture`) 
+
+
+Implementation Details
+
+### swagger.yaml
+```swagger codegen
+  /v2/support-levels/features:
+    get:
+      tags:
+        - installer
+      security:
+        - userAuth: [admin, read-only-admin, user]
+      operationId: GetSupportedFeatures
+      description: Retrieves the features support levels for each OpenShift version.
+      parameters:
+        - in: query
+          name: openshift_version
+          type: string
+          description: Version of the OpenShift cluster.
+          required: true
+        - in: query
+          name: cpu_architecture
+          description: The CPU architecture of the image (x86_64/arm64/etc).
+          type: string
+          # TODO: remove arm64 when AI moves to using aarch64
+          enum: [ 'x86_64', 'aarch64', 'arm64','ppc64le','s390x','multi' ]
+
+      responses:
+        "200":
+          description: Success.
+          schema:
+            type: object
+            properties:
+              features:
+                $ref: '#/definitions/support-levels'
+        "401":
+          description: Unauthorized.
+          schema:
+            $ref: '#/definitions/infra_error'
+        "403":
+          description: Forbidden.
+          schema:
+            $ref: '#/definitions/infra_error'
+        "503":
+          description: Unavailable.
+          schema:
+            $ref: '#/definitions/error'
+
+  /v2/support-levels/architectures:
+    get:
+      tags:
+        - installer
+      security:
+        - userAuth: [admin, read-only-admin, user]
+      operationId: GetSupportedArchitectures
+      description: Retrieves the architecture support-levels for each OpenShift version.
+      parameters:
+        - in: query
+          name: openshift_version
+          type: string
+          description: Version of the OpenShift cluster.
+          required: true
+      responses:
+        "200":
+          description: Success.
+          schema:
+            type: object
+            properties:
+              architectures:
+                $ref: '#/definitions/support-levels'
+        "401":
+          description: Unauthorized.
+          schema:
+            $ref: '#/definitions/infra_error'
+        "403":
+          description: Forbidden.
+          schema:
+            $ref: '#/definitions/infra_error'
+        "503":
+          description: Unavailable.
+          schema:
+            $ref: '#/definitions/error'
+
+  support-levels:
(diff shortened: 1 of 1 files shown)


eliorerz:
/cc  @gamli75 
/cc @avishayt 
/cc @osherdp 

FYI @celdrake @lalon4

eliorerz:
FYI @jkilzi

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,144 @@
...
+- Add another dimension (of CPU architecture) to the feature-supported endpoint  
+- Improve the maintainability of this code and less likely to require updates
+
+### Non-Goals
+
+- Change the API to allow the user to get the support level of a single feature without retrieving the entire list.
+- Design an easier way to add a new dimension to that API.
+- Get per cluster feature-support level
+
+## Proposal
```
avishayt, line 47: Here is an alternate proposal.  We can keep the current API but add a cpu_architecture query param that defaults to x86_64.  Then for x86_64, there is no change in the arguments or returned JSON.  We allow users to pass a different CPU architecture, returning a JSON with the same structure but different values.  In the future, we can add more query params without breaking the API.

celdrake: If the feature support levels can change based on CPU architecture, what would happen when there will be Day1 clusters with multiple CPU architectures? Or this could happen ATM with Day2 clusters which may have mixed CPU architectures.

It would be hard to express if a feature is supported or not if the values differ for the CPU architectures that the cluster might have. And it would imply that the UI / consuming application need to perform this logic which may not be simple.

For that reason, I find Elior's suggestion about being able to request per-cluster feature-support-levels.

avishayt: The enhancement doesn't describe how feature support for heterogeneous clusters - @eliorerz please address this.

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,144 @@
...
+### swagger.yaml
+```swagger codegen
+  /v2/support-level/{feature_id}:
+  parameters:
+    - in: path
+      name: feature_id
+      required: true
+    - in: query
+      name: cluster_id
+    - in: query
```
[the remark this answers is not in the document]
eliorerz: It will allow to get cluster specific feature-support-level without passing all the other arguments.
For example if the cluster OCP version is 4.12 and the CPU architecture is ARM than calling to `/v2/support-level/{feature_id}` with `cluster_id = x` will return the support-level for that feature.

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,144 @@
...
+x86 and arm64.
+2. The list is statically defined in assisted-service code so on new OCP version we will have to duplicate the entire 
+features from the previous version even if it supported or not supported for few past versions.
+3. It is hard to expand, beside of adding CPU architecture to that equation if in the future we will want to filter by a
+third attribute we will have to duplicate everything again (API change).
+4. The user is have to get the entire feature-support list even if he needs to know just the support of a single feature.
+
+These four reasons led me to the realization that we must create a more flexible and easier-to-maintain API. 
+
+### Goals
```
gamli75, line 36: one of the goals should be to have one place in the code that will return bad requests in case that feature combination is not supported and not the mess that we have today

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,144 @@
...
+- Change the current API to not return whole list but to return the support-level of a specific feature filtered by 
+some attributes (e.g. openshift_version, cpu_architecture, and more)
+- Get per cluster feature-support level - by given only cluster ID the service will determine the support level by 
+the cluster already exists attributes without the need to set multiple properties
+
+Implementation example:
+
+### swagger.yaml
+```swagger codegen
+  /v2/support-level/{feature_id}:
```
osherdp, line 62: if it's encapsulated in a cluster object (meaning the information is entirely included in a scope of a cluster) then I would expect something like:
```
GET /v2/clusters/{cluster_id}/support-levels
GET /v2/clusters/{cluster_id}/support-levels/{feature_id}
```

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Risks and Mitigations
+
+As with every large code and API change, this change contains a risk of breaking existing functionality.
+This can be mitigated by adding unit and subsystem tests to various scenarios.
+
+
+### UI Impact
+
+UI will to take impact from this change.\
```
osherdp, line 135: btw, what's up with those ``\`` at EOL?
saw it some places in this doc

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,144 @@
...
+Currently, UI get the feature-support level list and save it locally, and for every feature they query that "db" 
+for each feature.\
+This change will force the UI to change the way they handle the feature-support level (due to the API change),
```
osherdp, line 138: Not sure if it's mentioned anywhere, but if we have additional API endpoints we need to mark the old one deprecated and remove it when we move to v3 (just like we did on the move from v1->v2)

avishayt: If we break an API, our policy is to provide backward compatibility for 3 months.  We wouldn't move to v3 unless we dramatically changed models.

osherdp: Of course. It's just that deprecation process (in oppose to removal) can start as early as we'd like

docs/enhancements/add-feature-support-by-generic-attributes.md
osherdp: ok, I didn't know this was the arrangement 
Do UI follows this path? @jkilzi

docs/enhancements/add-feature-support-by-generic-attributes.md
eliorerz: New line on Markdown

docs/enhancements/add-feature-support-by-generic-attributes.md
eliorerz: I also didn't know that one  :) 
Anyway, For future release the user will always get automatically the return value that defined in `GetSupportLevel` function

docs/enhancements/add-feature-support-by-generic-attributes.md
eliorerz: This is the best approach except the following case: what we do when we don't have a cluster and need to decide if feature is supported of not (e.g. SNO)?

osherdp: Ah, you mean before creating a cluster UI depends on those endpoints
Fair point, although it means we still maintain two types of behaviors: cluster-scoped and globally-scoped. I wasn't aware of it (and maybe having it in one endpoint is a bit confusing)

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,166 @@
...
+
+### UI Impact
+
+UI will to take impact from this change.\
+Currently, UI get the multi openshift version feature-support-level list.\
+This change will force the UI to pass openshift_version alongside other arguments.
+
+### Test Plan
+
+- Create unittests and subsystem-tests for all existing feature support combinations
```
osherdp, line 166: not sure we need to test all combinations (it's quite a lot)

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,166 @@
...
+- Create an interface for feature-support level that will allow us to filter the feature-support level by
+  multiple arguments.
+- Create feature per feature ID that will implement this interface and will return the support level
+  of that specific feature.
+- Change the current API to not return whole list but to return a narrow support-level list filtered by a specific openshift-version
+  and other attributes (currently only cpu_architecture but more can be added in the future).
+- <del> Get per cluster feature-support level - by given only cluster ID the service will determine the support level by
+  the cluster already exists attributes without the need to set multiple properties </del>
+
+Implementation example:
```
osherdp, line 58: Let's put this in a ``Implementation Details`` section

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,166 @@
...
+returned object - CPU architecture, so we can determine if a feature is supported or not by two parameters instead of one.
+
+
+## Motivation
+Adding another attribute to the feature-support-level API is indeed required but in its current implementation there are
+few flaws (from my perspective):
+1. The current implementation forces us to duplicate the entire features list for each cpu architecture (API change).\
+   For example if for `OpenshiftVersion = 4.12` we currently hold 10 features so adding just ARM64 as a CPU architecture to
+   that equation will require us to duplicate the 10 features (for that architecture) and set the support-level for both
+   x86 and arm64.
```
osherdp, line 27: maybe even mentioning there are even more archs than that (``s390x`` & ``ppc64le``) so it's really is too much to just duplicate entries

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,166 @@
...
+2. The list is statically defined in assisted-service code so on new OCP version we will have to duplicate the entire
```
osherdp, line 28: I guess we can remove it as we now understand latest entry decides for later versions as well

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Non-Goals
+
+- Change the API to allow the user to get a list for a specific openshift_version without retrieving the entire list.
+- Design an easier way to add a new dimension to that API.
+- <del> Get per cluster feature-support level </del>
+
+## Proposal
+
+- Create an interface for feature-support level that will allow us to filter the feature-support level by
```
osherdp, line 49: maybe
```suggestion
- Enhance the interface for feature-support level to allow filtering based on
```

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,166 @@
...
+  multiple arguments.
+- Create feature per feature ID that will implement this interface and will return the support level
```
osherdp, line 51: not sure I understand ``Create feature per feature ID``. what do you mean?

docs/enhancements/add-feature-support-by-generic-attributes.md
```diff
@@ -0,0 +1,166 @@
...
+  of that specific feature.
+- Change the current API to not return whole list but to return a narrow support-level list filtered by a specific openshift-version
```
osherdp, line 53: but ``get a list for a specific openshift_version`` is part of the non-goals, no?

osherdp (changes requested):

avishayt (approved):

PR #5001, opened 2023-02-23
MGMT-13210: Revert "Increase timeout waiting for image-service"

osherdp:
Reverts openshift/assisted-service#4871

Now that we have https://github.com/openshift/assisted-test-infra/pull/1989 we don't need to wait longer for image-service.

/cc @adriengentil @gamli75 @eliorerz


adriengentil (approved):
:+1:

PR #5013, opened 2023-03-03
MGMT-13859: notify when creating cluster and resources

rccrdpccl:
This PR focuses on notifying when clusters and host objects are created, so that we have the first state to aggregate to events later down the pipeline.

We have reorganized the code so that it would be easier to follow and modify if need be.

One change focuses in adding an abstraction layer for notifying about resources (Hosts, Clusters, Events, etc).
This change is needed because we were repeating a lot the same code (just changing the type, but it was doing the same thing). This was bad for both maintainability (if we needed to change something, we had to touch several parts of the code) and also flexibility: if we had to add a new notification we'd have to copy and paste and slightly modify the code.
This has been kept separate from the StreamWriter, as they are dealing with different responsibilities: the stream writer is responsible for knowing how to write to kafka, the eventstream knows how to deal with the various resources in a centralized place.

Due to the above changes, tests had to be changed. We created specific tests for EventStream, and changed all other related tests that are using event stream. Due to the high complexity of these tests, it was too hard to actually assert anything about eventstream usage, so we injiected a dummy instead. Hopefully this can be improved in the future

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/bminventory/inventory.go
```diff
@@ -464,9 +464,13 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 	log := logutil.FromContext(ctx, b.log).WithField(ctxparams.ClusterId, id)
 	log.Infof("Register cluster: %s with id %s and params %+v", swag.StringValue(params.NewClusterParams.Name), id, params.NewClusterParams)
 	success := false
+	fullCluster := &common.Cluster{}
 	var err error
 	defer func() {
 		if success {
+			if b.stream != nil && err == nil {
+				_ = b.stream.Notify(ctx, fullCluster.Cluster)
```
osherdp, line 472: maybe only log the error as a warning so we keep track in case ``Notify`` errors

internal/stream/event_stream_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/google/uuid"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/assisted-service/internal/stream"
+	"github.com/openshift/assisted-service/models"
+	"github.com/sirupsen/logrus"
+)
+
+func GetDummyEventStream(ctrl *gomock.Controller) *stream.MockEventStreamInterface {
```
osherdp, line 19: seems exactly the same as in ``test/dummies.go``, no?

internal/stream/event_stream.go
```diff
@@ -0,0 +1,93 @@
...
+		s.writer.Close()
+	}
+}
+
+// Returns key, event type and fields to be logged
+func getNotificationDetails(resource Resource) (string, string, logrus.Fields, error) {
+	var key, eventType string
+	logFields := logrus.Fields{}
+
+	if res, ok := resource.(*models.InfraEnv); ok && res != nil {
```
osherdp, line 61: maybe type-switches can make it a bit cleaner: https://go.dev/tour/methods/16

osherdp (changes requested):

internal/stream/event_stream_test.go
rccrdpccl: Oops, forgot to remove it from here :facepalm:

internal/bminventory/inventory.go
rccrdpccl: In this specific implementation we are sending the notification asynchronously. Do you think it's worth adding the message anyway for future-proofing? My concern is that when reading the code we would consider this not failing because the message doesn't show, but really it could be failing with another (internal) message

osherdp: Even notify does *some action* synchronously (in addition to the asynchronous actions), so as long as it's returning an error we should show it in some way 
If the asynchronous actions have their respecting logs that's totally fine and expected. When reading the logs or the code it's expected we'll get into the thinking of asynchronous execution, if I understand you correctly

internal/stream/event_stream_test.go
```diff
@@ -0,0 +1,185 @@
...
+	BeforeEach(func() {
+		logger = logrus.New()
+		logger.Out = ioutil.Discard
+		ctrl = gomock.NewController(GinkgoT())
+		writer = stream.NewMockEventStreamWriter(ctrl)
+		clusterID = strfmt.UUID(uuid.New().String())
+
+	})
+
+	When("Closing event stream", func() {
```
carbonin, line 37: I'm assuming `When` is just `Context`? I'd just stick with `Context` since that's what is overwhelmingly used in the rest of the project.

For a single test context with no significant setup, just add the context string to the example (`It`).
In this case I would create a `Describe` block for each method you're testing then `It` blocks for each test like this:

```go
Describe("Close", func() {
  It("closes the underlying writer", func() {
  })

  It("succeeds when the underlying writer is nil", func() {
  })
})

Describe("Notify", func() {
  It("doesn't attempt to write to a nil writer", func() {
  })

  It("fails for a nil event", func() {
  })
  ...
})
```

carbonin (changes requested):
This is a bit confusing because it's mixing a refactoring with new stuff.

In the future I'd do the refactoring in an initial PR or at least a separate commit.

If all the refactoring is truly necessary I'd also like to see that talked about a bit more in the PR description. Explaining why all these changes are necessary would make this a lot easier to review.

internal/stream/event_stream_test.go
rccrdpccl: Yes, but it reads more naturally as plain English. However I agree, it's better to conform to the project's standard to cause least surprise when reading these tests.

internal/events/event.go
```diff
@@ -85,7 +85,12 @@ func (e *Events) v2SaveEvent(ctx context.Context, clusterID *strfmt.UUID, hostID
 			tx.Rollback()
 		} else {
 			tx.Commit()
-			e.notifyEventStream(ctx, &event.Event)
+			if e.stream != nil {
```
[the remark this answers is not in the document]
rccrdpccl: One reason for this change was exactly avoiding having nil pointers around the code that makes it very risky, so this object should actually never be nil. We're checking for nil as defensive strategy, but we could actually remove it if we wish.

internal/stream/event_stream.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+	"errors"
+	"reflect"
+
+	"github.com/openshift/assisted-service/models"
+	"github.com/sirupsen/logrus"
+)
+
+type EventStream struct {
+	writer EventStreamWriter
```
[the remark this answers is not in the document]
rccrdpccl: The reason is that I wanted to separate clearly the responsibility of the two objects: the event stream writer it's responsible to provide an easy way to write to kafka (it deals with lower level concepts as key, payload for example), the event stream knows how to notify relevant events about handled resources.

Another advantage of doing it this way is that we reduce cyclomatic complexity of tests but still having the same coverage, by testing one layer at the time.

rccrdpccl:
> This is a bit confusing because it's mixing a refactoring with new stuff.
> 
> In the future I'd do the refactoring in an initial PR or at least a separate commit.
> 
> If all the refactoring is truly necessary I'd also like to see that talked about a bit more in the PR description. Explaining why all these changes are necessary would make this a lot easier to review.

Sorry about that, I've added a better description but it was probably too late.

internal/stream/interfaces.go
```diff
@@ -0,0 +1,15 @@
...
+
+type EventStreamInterface interface {
+	Notify(ctx context.Context, resource Resource) error
+	Close(ctx context.Context)
+}
+
+type EventStreamWriter interface {
+	Write(ctx context.Context, eventName string, key []byte, payload interface{}) error
+	Close()
+}
```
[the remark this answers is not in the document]
rccrdpccl: As discussed in another comment, the reason for being different interfaces is the different responsibility they have: the writer takes care of low-level writing, the EventStream knows about the resources that we need to notify about.

internal/bminventory/inventory.go
```diff
@@ -464,9 +464,16 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 	log := logutil.FromContext(ctx, b.log).WithField(ctxparams.ClusterId, id)
 	log.Infof("Register cluster: %s with id %s and params %+v", swag.StringValue(params.NewClusterParams.Name), id, params.NewClusterParams)
 	success := false
+	fullCluster := &common.Cluster{}
```
carbonin, line 486: I think this can just be `cluster` or maybe change the name to `finalCluster` or `resultingCluster`.

"Full" makes me thing of HA mode.

internal/bminventory/inventory.go
```diff
@@ -464,9 +464,16 @@ func (b *bareMetalInventory) RegisterClusterInternal(
...
 	var err error
 	defer func() {
 		if success {
+			if b.stream != nil && err == nil && fullCluster != nil {
```
carbonin, line 490: If `stream` should never be nil I don't think we need these checks everywhere.

internal/common/testing/dummies.go
```diff
@@ -0,0 +1,12 @@
+package testing
+
+import (
+	"github.com/golang/mock/gomock"
+	"github.com/openshift/assisted-service/internal/stream"
+)
+
+func GetDummyEventStream(ctrl *gomock.Controller) *stream.MockEventNotifier {
+	dummyStream := stream.NewMockEventNotifier(ctrl)
+	dummyStream.EXPECT().Notify(gomock.Any(), gomock.Any()).AnyTimes().Return(nil)
```
carbonin, line 10: Is this always what we want or do we want to ensure we're not sending some events in some cases?

internal/events/event.go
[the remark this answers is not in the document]
carbonin: Mentioned this again in another thread, I think we're fine with removing them.

internal/host/transition.go
```diff
@@ -27,7 +26,6 @@ import (
 type transitionHandler struct {
 	db            *gorm.DB
 	log           logrus.FieldLogger
-	stream        stream.EventStreamWriter
```
carbonin, line 30: Was this just not being used?

internal/cluster/transition.go
```diff
@@ -38,7 +38,7 @@ var resetFields = append(append(resetProgressFields, resetLogsField...), "opensh
 type transitionHandler struct {
 	log                 logrus.FieldLogger
 	db                  *gorm.DB
-	stream              stream.EventStreamWriter
+	stream              stream.EventNotifier
```
carbonin, line 41: Is this not used? It's strange to change the type, but not have any other changes in the file.

internal/stream/event_stream.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const (
+	TypeInfraEnv = "InfraEnv"
+	TypeHost     = "HostState"
+	TypeCluster  = "ClusterState"
+	TypeEvent    = "Event"
+)
+
+type Resource interface{}
```
carbonin, line 24: Is there not an existing interface we can use? Like a gorm base model or some swagger thing?

internal/stream/event_stream.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (s *EventStream) Close() {
+	if s.writer != nil {
+		s.writer.Close()
+	}
+}
+
+// Returns key, event type and fields to be logged
+func getNotificationDetails(resource Resource) (string, string, logrus.Fields, error) {
```
carbonin, line 58: This looks like a really good place to actually use interfaces.

We should define an interface with this method on it that can be implemented by `common.<Thing>`.

pkg/kafka/event_stream_writer.go
```diff
@@ -16,7 +16,7 @@ const (
 	WriteTimeout time.Duration = 5 * time.Second
 )
 
-//go:generate mockgen -source=event_stream.go -package=stream -destination=mock_event_stream.go
+//go:generate mockgen -source=event_stream_writer.go -package=kafka -destination=mock_event_stream_writer.go
```
carbonin, line 19: I think if this is not specific to the `Event` model. We shouldn't call it an `event_stream`.

I think just `kafka.Writer` would be fine.

pkg/kafka/event_stream_writer.go
```diff
@@ -29,11 +29,6 @@ type EventEnvelope struct {
 	Metadata interface{} `json:"metadata"`
 }
 
-type EventStreamWriter interface {
-	Write(ctx context.Context, eventName string, key []byte, payload interface{}) error
-	Close()
-}
-
 type KafkaConfig struct {
```
carbonin, line 32: I'd also consider renaming these so that we don't have `kafka.Kafka...` everywhere.

internal/stream/event_stream.go
```diff
@@ -0,0 +1,94 @@
...
+const (
+	TypeInfraEnv = "InfraEnv"
+	TypeHost     = "HostState"
+	TypeCluster  = "ClusterState"
+	TypeEvent    = "Event"
+)
+
+type Resource interface{}
+
+func NewEventStream(writer EventStreamWriter, logger logrus.FieldLogger) *EventStream {
```
carbonin, line 26: This shouldn't be called `EventStream` if it's handling more than just `Events`

internal/stream/interfaces.go
```diff
@@ -0,0 +1,15 @@
...
+import "context"
+
+//go:generate mockgen -source=interfaces.go -package=stream -destination=mock_event_stream.go
+
+type EventNotifier interface {
+	Notify(ctx context.Context, resource Resource) error
+	Close()
+}
+
+type EventStreamWriter interface {
```
carbonin, line 12: I think this belongs in `internal/stream/writer_factory.go`

carbonin (commented):
Did an online review with @rccrdpccl

internal/common/testing/dummies.go
rccrdpccl: Ideally we would like to test some cases, but for now I think it's enough (due to high complexity of modifying current tests).
Possibly in a second phase we could implement some real tests about this.

internal/host/transition.go
rccrdpccl: :facepalm: I wonder why the linter didn't pick it up as not used?

internal/cluster/transition.go
rccrdpccl: After checking, I can confirm it is passed through as-is throughout the cluster transitionHandler. As receiver type also changed, no other change is needed.

rccrdpccl:
Will squash before merging - second and third commit message not meaningful.

internal/stream/event_stream.go
rccrdpccl: In the end this approach helped us extract most of the logic from kafka writer, leaving it as a basic wrapper for configuration and JSON Encoding. I would still leave it as separate, as it can be replaced for any other encoding we might want to use or separate storage for notifications

internal/common/db.go
```diff
@@ -89,6 +138,24 @@ type Host struct {
 	IgnitionEndpointToken string `json:"ignition_endpoint_token" gorm:"type:TEXT"`
 }
 
+func (h *Host) GetClusterID() *strfmt.UUID {
+	return h.ClusterID
+}
+func (h *Host) GetInfraEnvID() *strfmt.UUID {
+	return nil
```
carbonin, line 145: Why not return the host's infraenv id here?

internal/common/db.go
```diff
@@ -20,8 +20,21 @@ const (
...
+	NotificationTypeHost     = "HostState"
+	NotificationTypeInfraEnv = "InfraEnv"
 )
 
+type Notifiable interface {
+	GetClusterID() *strfmt.UUID
+	GetInfraEnvID() *strfmt.UUID
+	GetHostID() *strfmt.UUID
+	GetNotificationType() string
+	GetResource() any
```
carbonin, line 35: You can remove `Get` from all of these names.

internal/stream/notification_stream.go
```diff
@@ -0,0 +1,75 @@
...
+		log:      logger,
+	}
+
+}
+
+func (s *NotificationStream) Notify(ctx context.Context, notifiable common.Notifiable) error {
+	if s.writer == nil {
+		return nil
+	}
+	if notifiable == nil || reflect.ValueOf(notifiable).IsNil() {
```
carbonin, line 44: I don't think you need the reflect bit here.

internal/stream/notification_stream.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	key := ""
+	clusterID := notifiable.GetClusterID()
+	if clusterID != nil {
+		key = clusterID.String()
+	}
+
+	envelope := &Envelope{
+		Name:     notifiable.GetNotificationType(),
+		Payload:  notifiable.GetResource(),
```
carbonin, line 55: Is changing this to an empty interface really necessary? Does something not work if you just pass it as `common.Notifiable` type directly?

internal/cluster/common.go
```diff
@@ -190,8 +190,11 @@ func UpdateCluster(ctx context.Context, log logrus.FieldLogger, db *gorm.DB, eve
 	}
 
 	cluster, err := common.GetClusterFromDB(db, clusterId, common.UseEagerLoading)
-	if cluster != nil {
-		notifyEventStream(ctx, eventStream, &cluster.Cluster, log)
+	if cluster != nil && notificationStream != nil {
```
carbonin, line 193: I don't think you need the check for `notificationStream != nil` here, right?

carbonin (commented):
Looking good, just a few more comments.

internal/stream/notification_stream.go
rccrdpccl: I understand the concerns, however we need to pass the model here, not the wrapper, because:
* we need to leverage already-filtered fields
* we need to leverage well-defiined json conversion (named fields and such)
* but most importantly, we don't want to pass every column of the database, but rather the fields in the API. This would allow us to keep mindful retro compatibility (or break it, but still mindfully)

I would love to type check this, but I couldn't find a way to do it cleanly (that is not stretching it like adding a fake interface on some autogenerated methods, but I am not very comfortable with this as I am not sure what guarantees we have on such methods)

internal/stream/notification_stream.go
rccrdpccl: Apparently in Golang if you pass an empty interface the nil check is not enough, I've added a [test just to check the reflection bit](https://github.com/openshift/assisted-service/pull/5013/files/8b2c3565a7dcbe0b9ee26d0ca050da8433a79a86#diff-65c6c20c0876dfb3d2a2e3277f325d97eef102b1d4fe701ab59db0e76f15f2e1R92).

internal/cluster/common.go
rccrdpccl: I've left it because we receive it as argument in the function, would you say we should let it blow up if nil is passed?

internal/common/db.go
carbonin: That's incredibly unfortunate, wow. I guess that's the price we pay for being able to use functions as values :tired_face:

internal/cluster/common.go
rccrdpccl: Nevermind, just writing this I see there is no difference with the rest, why should we let it "fail" silently in this case? I'll update this line

PR #5024, opened 2023-03-09
MGMT-13903: Select IP inside machine CIDR for BMH

jhernand:
Currently when the service generates the bare metal hosts for the installed cluster it selects the first IP address of each NIC. But if there are multiple IP addresses in the same interface this may end up selecting an IP address that isn't inside the machine CIDR of the cluster. If that happens then there will be a missmatch between the IP addresses of machines and nodes. As a result machine will not be linked to nodes and machines will never reach the `Running` phase, and therefore the corresponding machine pool will never have the minium number of replicas, which makes the installation fail.

Related: https://issues.redhat.com/browse/MGMT-13903

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/ignition/ignition.go
+++ b/internal/ignition/ignition.go
@@ -949,9 +949,9 @@ func (g *installerGenerator) modifyBMHFile(file *config_latest_types.File, bmh *
 		}
 		switch {
 		case len(iface.IPV4Addresses) > 0:
-			hw.NIC[i].IP = g.getInterfaceIP(iface.IPV4Addresses[0])
+			hw.NIC[i].IP = g.selectInterfaceIPInsideMachineCIDR(iface.IPV4Addresses)
 		case len(iface.IPV6Addresses) > 0:
-			hw.NIC[i].IP = g.getInterfaceIP(iface.IPV6Addresses[0])
+			hw.NIC[i].IP = g.selectInterfaceIPInsideMachineCIDR(iface.IPV6Addresses)
 		}
 	}
 	for i, disk := range inventory.Disks {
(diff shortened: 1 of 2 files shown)


internal/ignition/ignition.go
```diff
@@ -1193,13 +1193,55 @@ func parseIgnitionFile(path string) (*config_latest_types.Config, error) {
 	return ParseToLatest(configBytes)
 }
 
-func (g *installerGenerator) getInterfaceIP(cidr string) string {
-	ip, _, err := net.ParseCIDR(cidr)
-	if err != nil {
-		g.log.Warnf("Failed to parse cidr %s for filling BMH CR", cidr)
-		return ""
+func (g *installerGenerator) selectInterfaceIPInsideMachineCIDR(interfaceCIDRs []string) string {
```
tsorya, line 1196: Try to use GetPrimaryMachineCIDRIP, it should find you ip that matched cidr

jhernand: I can't use that function directly because it returns a single IP for the machine, and this code is trying to find the IP for a particular network interface and IP version.

avishayt:
Thanks for the fix! Do we have a test for this?

jhernand:
> Thanks for the fix! Do we have a test for this?

@avishayt in the new version of the patch I added a couple of tests to verify that the IP addresses for the BMHs are generated correctly regardless of the order in the inventory. Please take a look.

avishayt (approved):

PR #5027, opened 2023-03-09
ACM-4127: Cache release images even if there is no matching OS image

carbonin:
This reduces the amount of work we have to do every time we encounter a cache miss.

Before this change a cache miss would re-evaluate all of the cluster image sets in the cluster and would require us to pull and inspect all of the images that we don't already have cache entries for. If we had a limited set of OS images this could mean a lot of additional processing time.

In some cases there were as many as 40 cluster image sets that didn't have OS images. Processing these caused requests to time out in some cases.

Additionally, these images were all being processed serially which also slowed everything down. This PR adds concurrency to the cluster image set processing to allow the first request to download an ISO to succeed even when it needs to process all (~80 in my testing) cluster image sets.

While just adding this concurrency would also have solved this issue I think the API (cluster create in this case) is the better place to validate the match between release image and OS version. Especially now that the release image handler and OS image handler are separated.

https://issues.redhat.com/browse/ACM-4127

- [ ] New Feature 
- [x] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Reproduced the issue using the information from the JIRA and the current latest image, deployed an image built from this PR and observed that the issue was resolved.

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -472,6 +472,32 @@ func (b *bareMetalInventory) getOLMMonitoredOperators(log *logrus.Entry, cluster
 	return nil, nil
 }
 
+func (b *bareMetalInventory) getNewClusterReleaseImage(ctx context.Context, params *models.ClusterCreateParams, arch string) (*models.ReleaseImage, error) {
+	releaseImage, err := b.versionsHandler.GetReleaseImage(ctx, swag.StringValue(params.OpenshiftVersion),
+		arch, swag.StringValue(params.PullSecret))
+	if err != nil {
+		return nil, errors.Wrapf(err, "Openshift version %s for CPU architecture %s is not supported",
+			swag.StringValue(params.OpenshiftVersion), arch)
+	}
+	if models.OpenshiftVersionSupportLevelMaintenance == releaseImage.SupportLevel {
+		return nil, errors.Errorf("Openshift version %s support level is: %s, and can't be used for creating a new cluster",
+			swag.StringValue(params.OpenshiftVersion), releaseImage.SupportLevel)
+	}
+	// Ensure a relevant OsImage exists. For multiarch we disabling the code below because we don't know yet
+	// what is going to be the architecture of InfraEnv and Agent.
+	if len(releaseImage.CPUArchitectures) == 1 {
+		releaseVersion := *releaseImage.OpenshiftVersion
+		releaseArch := releaseImage.CPUArchitectures[0]
+		var osImage *models.OsImage
+		osImage, err = b.osImages.GetOsImage(releaseVersion, releaseArch)
+		if err != nil || osImage.URL == nil {
+			return nil, errors.Errorf("No OS images are available for version %s and architecture %s", releaseVersion, releaseArch)
+		}
+	}
+
+	return releaseImage, nil
+}
+
 func (b *bareMetalInventory) RegisterClusterInternal(
 	ctx context.Context,
 	kubeKey *types.NamespacedName,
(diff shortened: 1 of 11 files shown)


internal/versions/versions.go
```diff
@@ -115,21 +115,27 @@ func (h *handler) GetReleaseImage(ctx context.Context, openshiftVersion, cpuArch
...
 	}
+	var wg sync.WaitGroup
 	for _, clusterImageSet := range clusterImageSets.Items {
-		existsInCache := false
-		for _, releaseImage := range h.releaseImages {
-			if releaseImage.URL != nil && *releaseImage.URL == clusterImageSet.Spec.ReleaseImage {
-				existsInCache = true
-				break
+		wg.Add(1)
+		go func(clusterImageSet hivev1.ClusterImageSet) {
```
carbonin, line 121: Right now this is spawning all the routines simultaneously, but I can also do this in batches if we think this is too much.

As is this would pull and inspect all 80 release images in parallel. While it works, I'm worried that it might not scale very well.

What do others think?

vrutkovs: +1 for batches, however we don't support a lot of old OpenShift releases, so its unlikely that current 80 release images will grow significantly. I wonder if we can test it on 200 release images to be sure

internal/bminventory/inventory.go
```diff
@@ -518,6 +518,18 @@ func (b *bareMetalInventory) RegisterClusterInternal(
...
 			swag.StringValue(params.NewClusterParams.OpenshiftVersion), releaseImage.SupportLevel))
 	}
+	// Ensure a relevant OsImage exists. For multiarch we disabling the code below because we don't know yet
+	// what is going to be the architecture of InfraEnv and Agent.
+	if len(releaseImage.CPUArchitectures) == 1 {
+		releaseVersion := *releaseImage.OpenshiftVersion
+		releaseArch := releaseImage.CPUArchitectures[0]
+		var osImage *models.OsImage
+		osImage, err = b.osImages.GetOsImage(releaseVersion, releaseArch)
+		if err != nil || osImage.URL == nil {
```
danielerez, line 528: URL property existence should actually be ensured on osimages [validate](https://github.com/openshift/assisted-service/blob/3c7a420f984202790ddbbc18e2b8aa04bbc392e4/internal/versions/osimages.go#L54). Did we encounter a null osImage.URL in some flow?

internal/versions/versions.go
```diff
@@ -115,21 +118,34 @@ func (h *handler) GetReleaseImage(ctx context.Context, openshiftVersion, cpuArch
...
+			existsInCache := false
+			for _, releaseImage := range h.releaseImages {
+				if releaseImage.URL != nil && *releaseImage.URL == clusterImageSet.Spec.ReleaseImage {
+					existsInCache = true
+					break
+				}
 			}
-		}
+			if !existsInCache {
+				_, err := h.addReleaseImage(clusterImageSet.Spec.ReleaseImage, pullSecret)
```
danielerez, line 141: Not sure we always validate that the pullSecret exists before calling GetReleaseImage.
So maybe worth an early return before starting the cache in case the pullSecret is missing?
I.e. so we won't try to cache all images and fail on missing pullSecret.

internal/versions/versions.go
```diff
@@ -115,21 +118,34 @@ func (h *handler) GetReleaseImage(ctx context.Context, openshiftVersion, cpuArch
 	if err := h.kubeClient.List(ctx, clusterImageSets); err != nil {
 		return nil, err
 	}
+	var wg sync.WaitGroup
```
danielerez, line 121: Just for clarity, would be nice to have a comment here that the rest of the func is essentially just for caching purposes.

internal/versions/versions.go
```diff
@@ -115,21 +118,34 @@ func (h *handler) GetReleaseImage(ctx context.Context, openshiftVersion, cpuArch
...
 	for _, clusterImageSet := range clusterImageSets.Items {
-		existsInCache := false
-		for _, releaseImage := range h.releaseImages {
-			if releaseImage.URL != nil && *releaseImage.URL == clusterImageSet.Spec.ReleaseImage {
-				existsInCache = true
-				break
-			}
+		if err := h.sem.Acquire(ctx, 1); err != nil {
```
danielerez, line 123: Maybe we can use something like [samber/lo/parallel](https://github.com/samber/lo/tree/master/parallel) to handle this for us. Haven't tried it, but seems it supports a parallel processing of ForEach: https://github.com/samber/lo/#foreach (looks like just a simple use of WaitGroup though https://github.com/samber/lo/blob/master/parallel/slice.go#L30)

carbonin: We have two separate concerns here.

The wait group is so that we can track when all the routines are finished.
The semaphore is to limit the number of routines that can run concurrently.

I think we need both and that library only handles one.

internal/versions/versions.go
carbonin: Again, this is just the same logic, but indented to be in a go routine. I'd prefer not to change it here unless running these in parallel somehow changes something about the pull secret.

internal/versions/versions.go
carbonin: Added

danielerez:
lint seems to fail on cyclomatic complexity:( 
```level=info msg="[runner] linters took 52.527154613s with stages: goanalysis_metalinter: 52.491963671s, structcheck: 29.451µs"
internal/bminventory/inventory.go:475:1: cyclomatic complexity 31 of func `(*bareMetalInventory).RegisterClusterInternal` is high (> 30) (gocyclo)
func (b *bareMetalInventory) RegisterClusterInternal(
```

carbonin:
/override e2e-agent-compact

This job is failing for all PRs

carbonin:
Does the bot retest command override my override?
Seems strange that that job was running again :worried:

PR #5031, opened 2023-03-14
MGMT-13925: Handle missing inspection URL

eranco74:
In case the converged flow is enabled and the assisted-service fails to get a valid ironic inspection URL it's using the Ironic URL as default, since the Ironic URL is likely the cluster API VIP this behavior leads to inspection failures in case the inspection service run on a node that isn't holding the cluster VIP. The inspection error results in Agents that never show up.
**As a temporary workaround for the issue,** the changes in this PR will return in case the inspection URL is empty, this error will cause the service to fail with a relevant error.

- Is this PR relying on CI for an e2e test run? Yes
- Should this PR be tested in a specific environment? Yes, with converged flow enabled

-->

https://issues.redhat.com/browse/MGMT-13925

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

- [ x] Title and description added to both, commit and PR.

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/bmo_utils.go
+++ b/internal/controller/controllers/bmo_utils.go
@@ -68,8 +68,18 @@ func (r *BMOUtils) GetIronicServiceURLS() (string, string, error) {
 		return "", "", err
 	}
 	ironicIP, inspectorIP, err := provisioning.GetIronicIP(r.kubeClient, provisioningInfo.Namespace, &provisioningInfo.Spec, r.osClient)
-	if err != nil || ironicIP == "" {
+	if err != nil {
 		r.log.WithError(err).Error("unable to determine Ironic's IP")
+		return "", "", err
+	}
+	if inspectorIP == "" {
+		err = errors.New("unable to determine inspector IP, check if metal3 pod is running")
+		r.log.WithError(err)
+		return "", "", err
+	}
+	if ironicIP == "" {
+		err = errors.New("unable to determine Ironic's IP")
+		r.log.WithError(err)
 		return "", "", err
 	}
 	ironicURL := getUrlFromIP(ironicIP)
--- a/internal/controller/controllers/bmo_utils_test.go
+++ b/internal/controller/controllers/bmo_utils_test.go
@@ -97,6 +97,30 @@ var _ = Describe("BMOUtils", func() {
 			Expect(serviceURL).Should(Equal("https://" + ironicIP))
 			Expect(inspectorURL).Should(Equal("https://" + ironicIP))
 		})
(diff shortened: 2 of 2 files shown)


filanov:
@eranco74 can you please elaborate why it's a temporary solution? 
In the original PR i didn't see why only ironic URL is tested, do you know the reason for that? 
In addition i didn't find a reference to where ` it's using the Ironic URL as default` can you please point to that?

eranco74:
> why it's a temporary solution?
Unlick the Ironic URL the inspection URL can change over time, we should get it as close as possible to the time we generate the image.
> In the original PR i didn't see why only ironic URL is tested, do you know the reason for that?
I think the inspection URL was added as part of a refactor that moved the code from the assisted-service repo to cluster-baremetal-operator repo.
Yet the tests for this flow are still lacking.
> reference to where  it's using the Ironic URL as default can you please point to that?
https://github.com/openshift/image-customization-controller/blob/94e8789385209b3a4044abc9f38d456b5d744f7b/pkg/ignition/builder.go#L43

filanov:
> Unlick the Ironic URL the inspection URL can change over time, we should get it as close as possible to the time we generate the image.

I guess you mean use the image, the generation is done once the infra env is created but it doesn't mean it will be used.

internal/controller/controllers/bmo_utils.go
```diff
@@ -68,10 +68,20 @@ func (r *BMOUtils) GetIronicServiceURLS() (string, string, error) {
 		return "", "", err
 	}
 	ironicIP, inspectorIP, err := provisioning.GetIronicIP(r.kubeClient, provisioningInfo.Namespace, &provisioningInfo.Spec, r.osClient)
-	if err != nil || ironicIP == "" {
+	if err != nil {
 		r.log.WithError(err).Error("unable to determine Ironic's IP")
 		return "", "", err
 	}
+	if inspectorIP == "" {
```
filanov, line 75: Is there a case where it's valid to have an empty `inspectorIP`? 
What can break because of this change?

osherdp:
/uncc
seems to have enough reviewers

internal/controller/controllers/bmo_utils.go
carbonin: I don't think so. And even if it was empty we only have a chance of things working out.

I'd rather see this fail.

PR #5039, opened 2023-03-15
MGMT-13958: MGMT-13862: Return bed request on wrong feature-feature or feature-architecture combination

eliorerz:
[MGMT-13862](https://issues.redhat.com//browse/MGMT-13862): Return bed request on wrong feature-feature or feature-architecture combination.
[MGMT-13958](https://issues.redhat.com//browse/MGMT-13958): Delete unused feature IDs from /v2/support-levels/features

Features that are being used by the UI:
image

Note: `CLUSTER_MANAGED_NETWORKING_WITH_VMS` has been removed due to that we are returning `Supported` statically regardless of cpu architecture or openshift version 

- [x] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

/cc @gamli75

### DIFF
--- a/internal/featuresupport/feature_support_level.go
+++ b/internal/featuresupport/feature_support_level.go
@@ -1,51 +1,28 @@
 package featuresupport
 
 import (
+	"fmt"
 	"reflect"
 
 	"github.com/go-openapi/swag"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/models"
+	"github.com/thoas/go-funk"
 )
 
 var featuresList = map[models.FeatureSupportLevelID]SupportLevelFeature{
-	models.FeatureSupportLevelIDADDITIONALNTPSOURCE:                &AdditionalNtpSourceFeature{},
-	models.FeatureSupportLevelIDREQUESTEDHOSTNAME:                  &RequestedHostnameFeature{},
-	models.FeatureSupportLevelIDPROXY:                              &ProxyFeature{},
-	models.FeatureSupportLevelIDSNO:                                &SnoFeature{},
-	models.FeatureSupportLevelIDDAY2HOSTS:                          &Day2HostsFeature{},
-	models.FeatureSupportLevelIDVIPAUTOALLOC:                       &VipAutoAllocFeature{},
-	models.FeatureSupportLevelIDDISKSELECTION:                      &DiscSelectionFeature{},
-	models.FeatureSupportLevelIDOVNNETWORKTYPE:                     &OvnNetworkTypeFeature{},
-	models.FeatureSupportLevelIDSDNNETWORKTYPE:                     &SdnNetworkTypeFeature{},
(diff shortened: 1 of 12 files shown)


internal/featuresupport/feature_support_level.go
```diff
@@ -122,3 +129,34 @@ func IsFeatureSupported(featureId models.FeatureSupportLevelID, openshiftVersion
...
+
+	for _, feature := range featuresList {
+		if feature.IsFeatureActivated(cluster, updateParams) {
+			activatedFeatures = append(activatedFeatures, feature)
+		}
+	}
+
+	for _, feature := range activatedFeatures {
+		if isCompatible, incompatibleFeature := IsFeatureCompatible(feature, activatedFeatures...); !isCompatible {
+			return fmt.Errorf("cannot activate %s feature alnogside %s feature", feature.GetId(), incompatibleFeature)
```
gamli75, line 157: if the feature is not supported for specific architecture - the message should be different and align with all error messages in the code 
"return errors.New("only x86-64 CPU architecture is supported on Nutanix clusters")"
so I think we should split this method and first check if it's compatible with the architecture - if not - return specific message 
if yes - return other message 
@nirfarkas WDYT?

nirfarkas:
Sounds reasonable.

internal/featuresupport/features.go
```diff
@@ -215,13 +253,33 @@ func (feature *SingleNodeExpansionFeature) GetSupportLevel(filters SupportLevelF
 	return models.SupportLevelSupported
 }
 
+func (feature *SingleNodeExpansionFeature) IsFeatureActivated(cluster common.Cluster, updateParams *models.V2ClusterUpdateParams) bool {
+	if swag.StringValue(cluster.HighAvailabilityMode) == models.ClusterHighAvailabilityModeFull || cluster.Hosts == nil {
```
gamli75, line 257: not sure about this condition

internal/featuresupport/features.go
```diff
@@ -238,19 +296,34 @@ func (feature *LvmFeature) GetSupportLevel(filters SupportLevelFilters) models.S
...
+		for _, operatorParams := range updateParams.OlmOperators {
+			if operatorParams.Name == "lvm" {
+				return true
+			}
+		}
+	}
 
-func (feature *DualStackNetworkingFeature) GetId() models.FeatureSupportLevelID {
-	return models.FeatureSupportLevelIDDUALSTACKNETWORKING
+	if cluster.MonitoredOperators != nil {
```
gamli75, line 308: why do we need to check also  cluster.MonitoredOperators? why updateParams.OlmOperators is not enough?

eliorerz: Because that there is few ways to determine if operator is activated or not.
If the user updates the cluster but not updating the (already selected) operators `updateParams` are not enough for that

gamli75 (approved):

PR #5041, opened 2023-03-15
Get the ironic URLs in reconcile rather than at startup

carbonin:
The ironic URL is proxied from each node to the pod, but this isn't currently done for the inspector URL. This means that if the pod IP changes between when the assisted-service starts and when we try to boot a host the host will fail to register.

This PR attempts to mitigate the issue by re-reading the URLs when the preprovisioning image is reconciled.

This requires us to compare the ignition override to the one currently stored for the infraenv as the infraenv controller has no knowledge of the ironic URLs so it won't update the infraenv when they change. Additionally the preprovisioning image controller now needs to evaluate every infraenv regardless if the ISO URL needs to be changed as it's possible that the inspector URL has been updated and the ignition override needs to be updated.

https://issues.redhat.com/browse/MGMT-13925

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Manually in dev-scripts.

I reproduced the bug before I added my first host by killing the metal3 pod and ensuring it got a new IP.
I was able to see the first agent fail to register.
I then updated the infraenv to force a preprovisioning image reconcile and saw the new IP get logged.
I was then able to add a second BMH and see that agent register.

To get the first agent to register I needed to remove and re-add the BMH due to https://issues.redhat.com/browse/OCPBUGS-5833

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -397,28 +395,40 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
 		log.Infof("Setting default ironic agent image (%s) for infraEnv %s", ironicAgentImage, infraEnv.Name)
 	}
 
-	conf, err := ignition.GenerateIronicConfig(r.IronicServiceURL, r.IronicInspectorURL, *infraEnvInternal, ironicAgentImage)
+	ironicServiceURL, inspectorURL, err := r.BMOUtils.GetIronicServiceURLS()
```
eranco74, line 398: Looks like `r.BMOUtils.GetIronicServiceURLS()` will not return an error in case it failed to get the inspectionURL.
I'm pretty sure I fixed that already, perhaps this PR needs rebasing

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -397,28 +395,40 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
...
 	if err != nil {
-		log.WithError(err).Error("failed to generate Ironic ignition config")
-		return ctrl.Result{}, err
+		log.WithError(err).Error("failed to get IronicServiceURLs")
```
eranco74, line 400: Consider setting error status in this case

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -397,28 +395,40 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
...
+		return false, err
 	}
 
-	_, err = r.Installer.UpdateInfraEnvInternal(ctx, installer.UpdateInfraEnvParams{InfraEnvID: *infraEnvInternal.ID, InfraEnvUpdateParams: &models.InfraEnvUpdateParams{}}, swag.String(string(conf)))
+	conf, err := ignition.GenerateIronicConfig(ironicServiceURL, inspectorURL, *infraEnvInternal, ironicAgentImage)
 	if err != nil {
-		return ctrl.Result{}, err
+		log.WithError(err).Error("failed to generate Ironic ignition config")
```
eranco74, line 406: Set error status here as well

eranco74: Actually, return an error you can use for status message and set the error in line 141

internal/controller/controllers/preprovisioningimage_controller.go
carbonin: Rebased

internal/controller/controllers/preprovisioningimage_controller.go
carbonin: Done

carbonin:
/hold

Waiting to see if metal team adds the inspector proxy

PR #5050, opened 2023-03-21
manifests: Add a shebang to NM dispatcher script

cgwalters:
This will reportedly fix a SELinux issue that happens with the RHEL 9 rebase for OCP 4.13, where the dispatcher policy is a bit tighter for scripts by default.

See also https://github.com/fedora-selinux/selinux-policy/issues/778 which is highly related to this - the policy is stricter for bash versus executables.

### DIFF
--- a/internal/network/manifests_generator.go
+++ b/internal/network/manifests_generator.go
@@ -88,7 +88,7 @@ const unmanagedResolvConf = `
 rc-manager=unmanaged
 `
 
-const forceDnsDispatcherScript = `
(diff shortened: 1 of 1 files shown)


dollierp:
To give more context.

Adding the shebang makes the script run as `NetworkManager_dispatcher_custom_t` domain which is unconfined (https://github.com/fedora-selinux/selinux-policy/pull/1288).
Without the shebang, the script run as `NetworkManager_dispatcher_t` which is not allowed to create files and fails.

PR #5066, opened 2023-03-23
MGMT-14074: Reload NM config after creation

jhernand:
Apparently in versions of RHCOS based on RHEL 9 when the `99-assisted-pre-static-network-config.sh` script is executed the NetworkManager service is already running and will not automatically reload the connection configuration files that we create. When using the minimal ISO that may prevent the dowload of the rootfs, as the network settings may not be correct. To avoid that this patch changes the script so that it explicitly runs `nmcli connection reload` after creating the files.

Related: https://issues.redhat.com/browse/MGMT-14074

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Tested running a 9.2 minimal ISO in a virtual machine. Interrupted the boot adding `rd.break=initqueue` to the kernel command line. Then modified the shell script manually, and exited the shell to continue the boot. With the change the boot continues and the rootfs is downloaded. Without change the boot fails to download the rootfs.

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/constants/scripts.go
+++ b/internal/constants/scripts.go
@@ -180,6 +180,7 @@ function copy_nmconnection_files_to_nm_config_dir() {
   done
 
   cp ${host_dir}/*.nmconnection ${ETC_NETWORK_MANAGER}/
+  nmcli connection reload
   echo "Info: Copied all nmconnection files from $host_dir to $ETC_NETWORK_MANAGER"
 }
 


cgwalters:
In RHEL9, NetworkManager runs as a persistent daemon in the initramfs, instead of "configure and quit" mode.

Instead of using the dracut initqueue stuff, I'd recommend using a systemd unit that is ordered `Before=NetworkManager.service`.

That said, this looks fine too - it's also supported to generate configuration after NM runs and AIUI this is indeed the way to enable the config.

But also probably in some cases you don't want the default "DHCP all the ethernets" behavior, and to fully control the network config then, the right fix is to run as a systemd unit.

cgwalters:
This all said, we also need to figure out a more sustainable/maintainable way for you guys to customize the ISO.

jhernand:
@tsorya @avishayt this issue is blocking the QE team. Do you have any objection to merge this fix while we work on a better solution that addresses the suggestions from @cgwalters?

tsorya:
Sounds good to me

avishayt:
Sounds good.  Please open a JIRA for the long-term fix so we can track it.

jhernand:
> Sounds good. Please open a JIRA for the long-term fix so we can track it.

Here it is: https://issues.redhat.com/browse/MGMT-14125

carbonin:
> ... the right fix is to run as a systemd unit.

We had to go this way to be able to run a single script for both the minimal and full ISO case.

Is being able to run a systemd unit in initramfs (before rootfs is downloaded) a feature of newer CoreOS versions?
We'd need this to work as far back as 4.8

carbonin:
> This all said, we also need to figure out a more sustainable/maintainable way for you guys to customize the ISO.

@cgwalters 
I opened https://issues.redhat.com/browse/MGMT-12340 a while back for this, but we would need all the features we support now (minimal iso, static network config, proxy, discovery kargs, etc) to be supported back through 4.8 before we can drop any of this home grown code.

Right now after a quick re-read it's blocked on https://github.com/coreos/coreos-installer/issues/794 at least, but this is perhaps not the best forum for discussing this.

carbonin:
In that same vein .. @jhernand are we sure this doesn't cause any issues with older RHCOS versions?

Is `nmcli` available in initramfs in all versions of RHCOS we support?
I recall having to use `ip` commands to setup networking directly in the emergency shell more than once.

cgwalters:
The dracut initqueue stuff has long been deprecated.  We've been using systemd units in the initramfs since RHEL8 came out AFAIK.

jhernand:
> In that same vein .. @jhernand are we sure this doesn't cause any issues with older RHCOS versions?
> 
> Is `nmcli` available in initramfs in all versions of RHCOS we support? I recall having to use `ip` commands to setup networking directly in the emergency shell more than once.

@carbonin I tested a cluster installation with 4.12 and this patch, but didn't explicitly check if `nmcli` is available in the RHEL8 initramfs. I will check and adjust the with a new patch if needed.

jhernand:
Thanks for confirming @cgwalters, this should fix that issue: https://github.com/openshift/assisted-service/pull/5087. Anyhow I believe it isn't a serious problem: if the `nmcli` command doesn't exist the shell will print an error message and continue running the script.

carbonin:
>When using the minimal ISO that may prevent the dowload of the rootfs, as the network settings may not be correct. To avoid that this patch changes the script so that it explicitly runs nmcli connection reload after creating the files.

Wasn't the entire purpose of the patch to reload things before trying to download the rootfs? Before we download the rootfs the file won't be available so what is the patch actually accomplishing?

jhernand:
The purpose was to do that in RHEL 9, where the `nmcli` is available in the intramfs.

carbonin:
Ah okay. I should have read @cgwalters 's comment more carefully I interpreted "It's not today" to mean that it wasn't available in the initramfs in any version.

cgwalters:
I edited the comment to clarify I was talking about rhel8.

PR #5077, opened 2023-03-27
MGMT-13918: Modify ignored validation should not be possible post-install

paul-maidment:
Once the user has started an installation, it appears that it is possible to modify ignored validation values. This may lead to issues when trying to diagnose the reasons for the failure of a cluster.

Do this PR disables the capability to modify the ignored validations in a post-install state

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory_v2_handlers.go
+++ b/internal/bminventory/inventory_v2_handlers.go
@@ -634,7 +634,10 @@ func (b *bareMetalInventory) V2SetIgnoredValidations(ctx context.Context, params
 		err = errors.Wrapf(err, "failed to fetch cluster %s to apply ignored validations", *cluster.ID)
 		return installer.NewV2SetIgnoredValidationsInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
 	}
-
+	err = b.clusterApi.VerifyClusterUpdatability(cluster)
+	if err != nil {
+		return b.setIgnoredValidationsBadRequest(err.Error())
+	}
 	problems := []string{}
 	cluster.IgnoredClusterValidations = params.IgnoredValidations.ClusterValidationIds
 	cluster.IgnoredHostValidations = params.IgnoredValidations.HostValidationIds
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -200,6 +200,27 @@ var (
 	serviceBaseURL      = "https://assisted.example.com:6008"
 )
 
+type clusterIdMatcher struct {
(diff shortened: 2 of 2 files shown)


internal/bminventory/inventory_v2_handlers.go
```diff
@@ -630,6 +630,13 @@ func (b *bareMetalInventory) V2SetIgnoredValidations(ctx context.Context, params
 		return b.setIgnoredValidationsBadRequest("the capability to ignore validations is not available")
 	}
 	cluster, err := b.getCluster(ctx, params.ClusterID.String())
+	preInstallStates := []string{
+		"insufficient", "ready", "pending-for-input",
+	}
+	if !swag.ContainsStrings(preInstallStates, *cluster.Status) {
+		return b.setIgnoredValidationsBadRequest(fmt.Sprintf("cluster %s is not in a pre install state, cannot modify ignored validations", *cluster.ID))
+	}
+
```
avishayt, line 639: Use cluster.VerifyClusterUpdatability()

avishayt (commented):
Commit message - Mofify -> Modify

internal/bminventory/inventory_v2_handlers.go
paul-maidment: This seems reasonable but it does raise some issues when it comes to testing.

The method VerifyClusterUpdatability accepts a cluster, the cluster in the test can be asynchronously updated during the lifespan of the test, meaning that mocks like

`mockClusterApi.EXPECT().VerifyClusterUpdatability(cluster).Return(errors.Errorf("Mock non updatable cluster")).AnyTimes()`

Do not function correcty.

While we are discussing this, I have implemented the mock as 
`mockClusterApi.EXPECT().VerifyClusterUpdatability(gomock.Any()).Return(errors.Errorf("Mock non updatable cluster")).AnyTimes()`

But I feel that this is insufficient. 

Rather, I think we should have a helper method alongside VerifyClusterUpdatability called  VerifyClusterUpdatabiltyByID, which would fetch the cluster from the database and then call VerifyClusterUpdatability, returning the result.

The benefit of this is that VerifyClusterUpdatabiltyByID would be mockable by simply providing a cluster ID and would not be subject to mutations that would prevent the mock from functioning due to async side effects.

avishayt (approved):

PR #5083, opened 2023-03-28
MGMT-13083: lru cache for release binaries

slaviered:
The extracted installer binaries are piling
up in the service's ephemeral storage and might
cause pod eviction.

This PR adds an LRU cache that stores the paths
to the installer binaries and on cache eviction
deletes the files themselves to make room for the
new installers.

Admittedly, LFU is a more suitable strategy, as it
avoids evicting frequently used entries on a burst of adding new entries. However, the LFU implementation does not support onEvict callback, so we will have to wait for future releases.

The code was tested with unit tests and manual e2e tests to ensure that the behavior is as expected. 

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/ignition/ignition.go
+++ b/internal/ignition/ignition.go
@@ -234,6 +234,7 @@ type installerGenerator struct {
 	providerRegistry              registry.ProviderRegistry
 	installerReleaseImageOverride string
 	clusterTLSCertOverrideDir     string
+	installerCache                *installercache.ReleaseCache
 }
 
 // IgnitionConfig contains the attributes required to build the discovery ignition file
@@ -280,7 +281,7 @@ func NewBuilder(log logrus.FieldLogger, staticNetworkConfig staticnetworkconfig.
 // NewGenerator returns a generator that can generate ignition files
 func NewGenerator(serviceBaseURL string, workDir string, installerDir string, cluster *common.Cluster, releaseImage string, releaseImageMirror string,
 	serviceCACert string, installInvoker string, s3Client s3wrapper.API, log logrus.FieldLogger, operatorsApi operators.API,
-	providerRegistry registry.ProviderRegistry, installerReleaseImageOverride, clusterTLSCertOverrideDir string) Generator {
+	providerRegistry registry.ProviderRegistry, installerReleaseImageOverride, clusterTLSCertOverrideDir string, cacheSize int) Generator {
 	return &installerGenerator{
 		cluster:                       cluster,
 		log:                           log,
@@ -297,6 +298,7 @@ func NewGenerator(serviceBaseURL string, workDir string, installerDir string, cl
 		providerRegistry:              providerRegistry,
 		installerReleaseImageOverride: installerReleaseImageOverride,
 		clusterTLSCertOverrideDir:     clusterTLSCertOverrideDir,
+		installerCache:                installercache.New(cacheSize, log),
 	}
 }
 
@@ -331,11 +333,6 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte,
 	}
 	defer removeIcspFile(icspFile)
 
-	installerPath, err := installercache.Get(g.installerReleaseImageOverride, g.releaseImageMirror, g.installerDir,
-		g.cluster.PullSecret, platformType, icspFile, log)
-	if err != nil {
-		return errors.Wrap(err, "failed to get installer path")
-	}
 	installConfigPath := filepath.Join(g.workDir, "install-config.yaml")
 
 	g.enableMetal3Provisioning, err = common.VersionGreaterOrEqual(g.cluster.Cluster.OpenshiftVersion, "4.7")
@@ -387,6 +384,14 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte,
 		log.WithError(err).Error("Failed to import cluster TLS certs")
 		return err
 	}
+
+	r, err := g.installerCache.Get(g.installerReleaseImageOverride, g.releaseImageMirror, g.installerDir,
+		g.cluster.PullSecret, platformType, icspFile)
+	defer r.Unlock()
+	if err != nil {
+		return errors.Wrap(err, "failed to get installer path")
+	}
+	installerPath := r.Path
 
 	err = g.runCreateCommand(ctx, installerPath, "manifests", envVars)
 	if err != nil {
--- a/internal/installercache/installercache.go
+++ b/internal/installercache/installercache.go
@@ -1,8 +1,10 @@
 package installercache
 
 import (
+	"os"
 	"sync"
 
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/openshift/assisted-service/internal/oc"
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/assisted-service/pkg/executer"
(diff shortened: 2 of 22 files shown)


pkg/generator/generator.go
```diff
@@ -33,6 +33,7 @@ type Config struct {
 	DummyIgnition      bool   `envconfig:"DUMMY_IGNITION"`
 	InstallInvoker     string `envconfig:"INSTALL_INVOKER" default:"assisted-installer"`
 	ServiceBaseURL     string `envconfig:"SERVICE_BASE_URL"`
+	InstallerCacheSize int    `envconfig:"INSTALLER_CACHE_SIZE" default:"5"`
```
osherdp, line 36: let's add it in ``openshift/template.yaml`` so we can control it for the different environments (in case we'll need to)

internal/installercache/installercache_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	AfterEach(func() {
+		_ = os.Remove(r1)
+		_ = os.Remove(r2)
+		_ = os.Remove(r3)
+	})
+
+	addFile := func(releaseID string) *release {
+		cache.releases.ContainsOrAdd(releaseID, &release{})
+		r, _ := cache.releases.Get(releaseID)
```
osherdp, line 31:
```suggestion
		r, err := cache.releases.Get(releaseID)
		Expect(err).ShouldNot(HaveOccurred())
```

to make sure we're not missing the more immediate error-paths in case something goes wrong there

internal/ignition/ignition.go
```diff
@@ -388,6 +390,12 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte,
 		return err
 	}
 
+	//lock the critical section where the files are writen to the installer
+	//path to avoid deletion of its contect mid air
+	r.Lock()
```
osherdp, line 395: is it possible to "lose" the object before this line (``r.Lock()``) gets the chance to get executed?

internal/installercache/installercache_test.go
```diff
@@ -0,0 +1,71 @@
...
+		r3 = addFile("4.10").Path
+		//verify that the oldest file was deleted
+		_, err := os.Stat(r1)
+		Expect(os.IsNotExist(err)).To(BeFalse())
+		_, err = os.Stat(r3)
+		Expect(os.IsNotExist(err)).To(BeFalse())
+	})
+})
+
+func TestSubsystem(t *testing.T) {
```
osherdp, line 68:
```suggestion
func TestInstallerCache(t *testing.T) {
```

internal/installercache/installercache_test.go
```diff
@@ -0,0 +1,71 @@
...
+		//mimic the operation of Get
+		r1 = addFile("4.8").Path
+		r2 = addFile("4.8").Path
+		r3 = addFile("4.10").Path
+		//verify that the oldest file was deleted
+		_, err := os.Stat(r1)
+		Expect(os.IsNotExist(err)).To(BeFalse())
+		_, err = os.Stat(r3)
+		Expect(os.IsNotExist(err)).To(BeFalse())
+	})
```
osherdp, line 65: I'd add a test for making sure locking mechanism doesn't allow evicting the release in mid-air

internal/installercache/installercache.go
```diff
@@ -1,61 +1,80 @@
...
 
-var cache installers = installers{
-	releases: make(map[string]*release),
+func New(capacity int, log logrus.FieldLogger) *ReleaseCache {
+	rc := ReleaseCache{
+		log: log,
+	}
+	c, err := lru.NewWithEvict(capacity, rc.onEvicted)
+	if err != nil {
+		log.WithError(err).Errorln("Failed to initialize the release binaries cache")
```
osherdp, line 35: Why not ``Fatalln``? If we couldn't construct the cache I don't see a reason to continue running

internal/installercache/installercache.go
```diff
@@ -1,61 +1,80 @@
...
+	defer r.Unlock()
 
-	r, present := i.releases[releaseID]
-	if !present {
-		r = &release{}
-		i.releases[releaseID] = r
+	i.log.Infof("Evicting entry %s %v", key, r.Path)
+	err := os.Remove(r.Path)
+	if err != nil {
+		i.log.Errorf("Failed to clear directory %s", r.Path)
```
osherdp, line 53: isn't ``os.Remove`` only suitable for either a file or an empty dir?

internal/installercache/installercache.go
```diff
@@ -1,61 +1,80 @@
...
+
+	r.Lock()
+	defer r.Unlock()
 
-	r, present := i.releases[releaseID]
-	if !present {
-		r = &release{}
-		i.releases[releaseID] = r
+	i.log.Infof("Evicting entry %s %v", key, r.Path)
+	err := os.Remove(r.Path)
```
osherdp, line 51: I wonder if just a one-time action of removing the release is enough here
I'm a bit concerned about leaking directories if we couldn't delete them on the ``onEvicted``, and the only "solution" I can think of is something that runs periodically and makes sure to GC those dirs

Anyway it's mainly a concern. We can see how much it's probable after letting it run for a while

osherdp (changes requested):
Looks great! Just a few notes

pkg/generator/generator.go
slaviered: Added

internal/installercache/installercache_test.go
slaviered: The other value is not error but ok (flag that indicates whether the value exists). In our case it guaranteed to exist.

internal/ignition/ignition.go
slaviered: Actually yes and it is a good catch. I changed the code so the lock is released only at the end of the flow.

internal/installercache/installercache_test.go
slaviered: Done

internal/installercache/installercache_test.go
slaviered: It is not something that can be done in unit tests because the actual code requires registries and other systems to be in place. This code in addition to the tests here need to be tested on load to see that there are also no races in addition to deadlocks.

internal/installercache/installercache.go
slaviered: ok

internal/installercache/installercache.go
slaviered: I agree we are leaking directories here but the main concern is the files and the space they take. In reality there are not that many versions running in parallel: 90% of the customers holds about 5 versions.

osherdp: yeah, I mean directories and files interchangeably. The concern here is leaking disk space because we only try to delete on a one-time basis

openshift/template.yaml
```diff
@@ -144,6 +144,9 @@ parameters:
 - name: ENABLE_UPGRADE_AGENT
   value: "false"
   required: false
+- name: INSTALLER_CACHE_SIZE
+  value: "5"
```
mhrivnak, line 148: This strikes me as small, and potentially leading to some cache churn. How many installers can we have before we risk hitting the disk space issue?

mhrivnak (commented):
What happens when the process restarts, due to being drained, upgraded, etc? I don't see anything that loads existing data from disk. Do we assume that the files are stored in the container's ephemeral storage, and thus in theory our process always starts with a clean slate? Depending on running this with ephemeral cache storage that's coupled to the lifecycle of the process would be nice to avoid.

It looks like the multiple layers of locking are required here because the underlying concurrency-safe structure is only aware of the in-memory data, and is not aware of or able to protect the on-disk data. I think we could end up with a simpler solution that's easier to understand if we could have one structure and one eviction routine that is aware of all the data.

What if we just use the filesystem itself as the source of truth for what is in the cache and how recently it was used? We can use mtime on the file and still do LRU, with a single lock.

`cache.Get`:
* acquires the lock
* updates the mtime on the installer if it's found on disk, otherwise obtains and extracts the installer
* releases the lock
* returns a reference to the installer on disk

`cache.garbageCollect` (runs periodically):
* acquire the lock
* get a list of installer files with their mtime values
* if there are too many, delete the ones with the oldest mtime as needed, except never delete a file if its mtime is within some grace period (5 minutes?)

The grace period helps ensure that an installer won't be deleted before the thing that got it uses it.

WDYT?

internal/ignition/ignition.go
```diff
@@ -388,6 +385,14 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte,
 		return err
 	}
 
+	r, err := g.installerCache.Get(g.installerReleaseImageOverride, g.releaseImageMirror, g.installerDir,
+		g.cluster.PullSecret, platformType, icspFile)
+	defer r.Unlock()
```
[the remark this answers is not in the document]
slaviered: I agree. I had no choice but to let the user unlock the lock, as the eviction routine might delete the file before it is used by the caller. But in this way it is at least explicit

openshift/template.yaml
slaviered: This number is based on the data we have about the ocp version deviation. It covers 90% of the cases. I attached the graph and a full explanation to the Jira ticket.

slaviered:
@mhrivnak  regarding your comment on cache implementation:

> cache.garbageCollect (runs periodically):
I am concerned that with fluctuation in usage and versions, we will encounter an overflow in the ephemeral storage, leading to an eviction of the pod. The LRU mechanism makes sure we always keep the limit no matter what. Also, the 3rd party package should provide multiple caching strategies, once the eviction callback is added. WDYT?

slaviered:
/hold

A different approach has been suggested. Will be implemented in a separate PR

slaviered:
replaced by https://github.com/openshift/assisted-service/pull/5120

PR #5091, opened 2023-03-29
MGMT-13668: Ensure that full ISO is always used for S390x and PowerCPU architectures

paul-maidment:
When generating the ISO URL, presently we just take the value straight from the infraenv. In the case of S390x and PowerCPU architectures, there is no support for minimal-iso. If we encounter these platforms, this PR will ensure that the full ISO link is always returned.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4898,6 +4898,13 @@ func (b *bareMetalInventory) UpdateInfraEnvInternal(ctx context.Context, params 
 		return nil, common.NewApiError(http.StatusNotFound, err)
 	}
 
+	if params.InfraEnvUpdateParams.ImageType == models.ImageTypeMinimalIso && infraEnv.CPUArchitecture == common.S390xCPUArchitecture {
+			msg := fmt.Sprintf("architecture %s does not support minimal ISO ", infraEnv.CPUArchitecture)
+			err := errors.New(msg)
+			log.Error(msg)
+			return nil, common.NewApiError(http.StatusBadRequest, err)
+	}
+
 	if params.InfraEnvUpdateParams.Proxy != nil {
 		if err = validateProxySettings(params.InfraEnvUpdateParams.Proxy.HTTPProxy,
 			params.InfraEnvUpdateParams.Proxy.HTTPSProxy,
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -7861,6 +7861,34 @@ var _ = Describe("infraEnvs", func() {
 				Expect(err).ToNot(HaveOccurred())
 				Expect(common.ImageTypeValue(i.Type)).To(Equal(models.ImageTypeFullIso))
 			})
+
+			It("Update Image type to minimal-iso on an unsupported architecture should not be permitted", func() {
+				mockOSImages.EXPECT().GetOsImageOrLatest(gomock.Any(), gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).AnyTimes()
+				mockOSImages.EXPECT().GetOsImage(gomock.Any(), gomock.Any()).Return(common.TestDefaultConfig.OsImage, nil).AnyTimes()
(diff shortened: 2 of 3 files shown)


internal/bminventory/inventory_v2_handlers.go
```diff
@@ -750,6 +750,11 @@ func (b *bareMetalInventory) GetInfraEnvDownloadURL(ctx context.Context, params
 }
 
 func (b *bareMetalInventory) generateImageDownloadURL(ctx context.Context, infraEnvID, imageType, version, arch, imageTokenKey string) (string, *strfmt.DateTime, error) {
+	// MGMT-13668 - TODO: Remove this when https://issues.redhat.com/browse/RFE-3647 is resolved
+	if imageType != "full-iso" && (arch == common.PowerCPUArchitecture || arch == common.S390xCPUArchitecture) {
+		b.log.Warn("Image type minimal-iso unsupported for architecture %s, forcing full-iso", arch)
+		imageType = "full-iso"
```
carbonin, line 756: Should we communicate this in some better way?

Ideally we could fail the infraenv update/create in the REST API.

Not so sure about kube API. Is there such a thing as a "warning condition/status"?

carbonin: Actually thinking about this more ... we definitely shouldn't allow this combination at all in the REST case. We know the inputs are invalid there and the user can simply choose the correct option.

paul-maidment: KubeAPI, I reckon we should reject the invalid combination if we are going down this path. No warning, just don't allow it.

filanov: Is there an ETA for it to be fixed?
If it will not happen anytime soon we can change block it in the api level. 
For kube-api it will be a sync error so it's the same

filanov: Actually kube-api will not work because there is no option to select the image type so it will be done in a controller level

avishayt: I agree.  We shouldn't return something other than what the user requested.
For REST, we should fail the create/update, as @carbonin said.  In the worst case of the user modifying the URL to change the image type, the download should return 400.

gamli75:
minimal ISO is supported for PowerCPU architecture

internal/bminventory/inventory.go
```diff
@@ -4898,6 +4898,13 @@ func (b *bareMetalInventory) UpdateInfraEnvInternal(ctx context.Context, params
 		return nil, common.NewApiError(http.StatusNotFound, err)
 	}
 
+	if params.InfraEnvUpdateParams.ImageType == models.ImageTypeMinimalIso && (infraEnv.CPUArchitecture == common.S390xCPUArchitecture || infraEnv.CPUArchitecture == common.PowerCPUArchitecture) {
```
gamli75, line 4901: minimal ISO is supported for PowerCPU architecture

paul-maidment: OK, So we should focus on only S390X

gamli75: yes

avishayt: Do we want to list the archs we support or the archs we don't support (will affect future archs)?

paul-maidment: Do we have a place where the supported archs by platform are enumerated at this time?

avishayt: Actually you should probably use the information from [this PR](https://github.com/openshift/assisted-service/pull/5101)

paul-maidment: It looks like the PR lists the unsupported arch's here

https://github.com/openshift/assisted-service/pull/5101/files#diff-9b84f395a9312d62831dda2f04163c4e99cb307046e2a0007b1b3bc005a7d3dcR570-R586

I don't see anything that lists supported arch's, I reckon "supported arch's" should also be covered by the PR if we intend to list them as it will keep things neat.

paul-maidment: What are you suggesting in this comment, 

It seems to just be around how we communicate an unsupported/supported architecture. I honestly think that what we already have is uncomplicated and sufficient.

WDYT?

PR #5107, opened 2023-04-05
MGMT-14125: Use systemd unit instead of dracut hook to configure network

jhernand:
Currently static network configurations are applied during discovery by a hook that dracut runs when udev has settled. Since OpenShift 4.13 and RHEL 9 the NetworkManager services stars earlier than that, and keeps running, which has forced us to modify that script to explicitly ask NetworkManager to reload the configuration that it generates. That works fine, but it has the inconvenient that NetworkManager will still try to use DHCP to activate the default connection. That adds approximately 90 seconds to the boot when there are no DHCP servers in the network. To avoid that this patch replaces the dracut hook with a systemd service that runs before NetworkManager has started.

Related: https://issues.redhat.com/browse/MGMT-14125
Related: https://issues.redhat.com/browse/MGMT-13970
Related: https://github.com/openshift/assisted-service/pull/5066

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


jhernand:
I already tested this with 4.13, but still need to test it with older versions.

jhernand:
@cgwalters In this [comment](https://github.com/openshift/assisted-service/pull/5066#issuecomment-1483244658) you suggested this approach using systemd. I would appreciate if you can take a look, specially to the [internal/constants/scripts.go](https://github.com/openshift/assisted-service/pull/5107/files#diff-a8e125b93e561e357af193995786e7b858898b2e1050a016b7a2c6340407a2f8) file containing the systemd service.

jhernand:
@paul-maidment If we eventually use this systemd approach it will no longer be a problem to `exit` from the script, in case you want to change that.

internal/isoeditor/rhcos.go
```diff
@@ -95,9 +106,48 @@ func formatRootfsServiceConfigFile(clusterProxyInfo *ClusterProxyInfo) (string,
...
+	if err := w.WriteHeader(hdr); err != nil {
+		return err
+	}
+	if _, err := w.Write([]byte(content)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func addSymlinkToArchive(w *cpio.Writer, path string, target string) error {
```
carbonin, line 127: I don't think this is necessary unless I'm missing some nuance.

Couldn't we pass the target as the content and change the mode using the existing function? Why add a new one?

internal/isoeditor/rhcos.go
```diff
@@ -95,9 +106,48 @@ func formatRootfsServiceConfigFile(clusterProxyInfo *ClusterProxyInfo) (string,
...
+		return err
+	}
+	return nil
+}
+
+func addDirToArchive(w *cpio.Writer, path string) error {
 	// add all the directories in path in the correct order, using dirsStack as FILO
 	dirsStack := []string{}
-	for dir := filepath.Dir(path); dir != "" && dir != "/"; dir = filepath.Dir(dir) {
+	for dir := filepath.Dir(path); dir != "" && dir != "/" && dir != "."; dir = filepath.Dir(dir) {
```
carbonin, line 150: Is this fixing a particular bug? If so, it would be good to call out what was going on in a separate commit.

internal/isoeditor/rhcos.go
```diff
@@ -43,12 +43,23 @@ func RamdiskImageArchive(netFiles []staticnetworkconfig.StaticNetworkConfigData,
...
 			return nil, err
 		}
+
+		servicePath := "/usr/lib/systemd/system/assisted-static-network-config.service"
+		serviceContent := constants.PreNetworkConfigService
+		if err := addFileToArchive(w, servicePath, serviceContent, 0o644); err != nil {
+			return nil, err
+		}
+
+		serviceLink := "/etc/systemd/system/initrd.target.wants/assisted-static-network-config.service"
```
carbonin, line 59: Is this only needed in the minimal ISO case?

We don't provide anything that runs during initrd in the full ISO so I'd be surprised if this worked there.

jhernand: I didn't check the full ISO, will do.

internal/isoeditor/rhcos.go
jhernand: It isn't necessary, I willl remove it.

internal/isoeditor/rhcos.go
jhernand: While trying to figure out how to write symlinks I added a `a -> b` symlink and this code got into an endless loop. This fixes that. I will move that to a separate patch.

internal/constants/scripts.go
```diff
@@ -196,3 +195,27 @@ if [ "$host_dir" ]
...
+
+const PreNetworkConfigService = `
+[Unit]
+Description=Assisted static network config
+DefaultDependencies=no
+
+# We need to run after the kernel and udev have detected all the network interface devices,
+# otherwise the mechanism that we use to find the MAC addresses will not work and the configuration
+# will not be applied.
+After=systemd-udev-settle.service
```
cgwalters, line 207: Note this service warns it's deprecated, because it's inherently racy.

internal/constants/scripts.go
```diff
@@ -196,3 +195,27 @@ if [ "$host_dir" ]
...
+
+# We need to run after this service because it runs the 'nm-initrd-generator' tool, which removes
+# all the connections and re-creates them from kernel command line arguments.
+After=coreos-enable-network.service
```
cgwalters, line 211: Hmm...I think this one is kind of an implementation detail that may change in the future, but probably not the near future.

internal/constants/scripts.go
```diff
@@ -196,3 +195,27 @@ if [ "$host_dir" ]
...
+
+# We need to run before this service so that when it starts all the connection files that we want
+# have already been created and they don't need to be re-loaded.
```
cgwalters, line 214: If we expect static configuration to be injected, perhaps what's actually simplest is to inject configuration that disables DHCP first?

https://docs.fedoraproject.org/en-US/fedora-coreos/sysconfig-network-configuration/#_disabling_automatic_configuration_of_ethernet_devices

Then we know NM will start up as usual, but *do nothing* until explicitly told to by this service.

And at that point, this service can wait for the devices it expects.

cgwalters: I don't fully understand everything that's going on in the flow here, but basically I think a flow of having a separate systemd unit run that's ordered *after* `NetworkManager.service` and injects config and then runs `nmcli reload` is actually better; the key is to just disable DHCP by default.

internal/isoeditor/rhcos.go
jhernand: The full ISO doesn't need this: the network isn't started early because there is no rootfs URL in the kernel command line arguments.

internal/constants/scripts.go
jhernand: I think that "no-auto-default" configuration is already there by default:

```
# cat usr/lib/NetworkManager/conf.d/initrd-no-auto-default.conf 
[.config]
enable=env:initrd

[main]
no-auto-default=*
```

But the "coreos-enable-network" service brings the "default_connection" back.

internal/constants/scripts.go
jhernand: Yes, I saw that explained in the service file, but didn't find a way around this. If I don't do it the that service re-creates the "default_connection", NetworkManager tries to activate it and then it introduces the 90 seconds wait till DHCP fails.

internal/constants/scripts.go
jhernand: The alternative to this would be to explicitly wait for the devices that we expect, as you suggest in another comment, but that would greatly complicate the current script.

internal/constants/scripts.go
cgwalters: > But the "coreos-enable-network" service brings the "default_connection" back.

Oooh.  That seems like a bug.  Or at least, something we need to enable suppressing.

dustymabe: is `usr/lib/NetworkManager/conf.d/initrd-no-auto-default.conf` in the initramfs?

jhernand: > is `usr/lib/NetworkManager/conf.d/initrd-no-auto-default.conf` in the initramfs?

Yes:

```
# mount -o loop,ro /files/images/rhcos-4.13.0-rc.2-x86_64-live.x86_64.iso /mnt
# lsinitrd /mnt/images/pxeboot/initrd.img | grep no-auto
-rw-r--r--   1 root     root           54 Jan  1  1970 usr/lib/NetworkManager/conf.d/initrd-no-auto-default.conf
```

dustymabe: I can't really find where that's defined in this repo: https://github.com/openshift/assisted-service/search?q=no-auto-default

dustymabe: oh actually I think it comes from NetworkManager itself: https://github.com/dracutdevs/dracut/blob/master/modules.d/35network-manager/initrd-no-auto-default.conf

ok, yes. This is by design. NM relies on `nm-initrd-generator` to generate the connections in the initramfs.

dustymabe:
ok. I think I understand the problem better now. What you are trying to do here is pretty much exactly what we are doing in [`coreos-copy-firstboot-network.service`](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service). My suggestion to you is to run your service `Before=coreos-copy-firstboot-network.service` and place your output files into `/etc/coreos-firstboot-network`. Those files will then get picked up by `coreos-copy-firstboot-network.service` and propagated appropriately.

IOW, hook back into our established flow for this rather than trying to figure out the ordering for yourself.

internal/constants/scripts.go
dustymabe: see my suggestion in https://github.com/openshift/assisted-service/pull/5107#issuecomment-1498005927

jhernand: > see my suggestion in [#5107 (comment)](https://github.com/openshift/assisted-service/pull/5107#issuecomment-1498005927)

I will try that, thanks @dustymabe.

jhernand: @dustymabe I quickly tried your suggestion, and unfortunately it doesn't work. The reason is that we read the `/sys/class/net/*` directories to find the MAC addresses of the network interfaces, and at the point that `coreos-copy-firstboot-network.service` runs those directories aren't populated yet. We need to know the MAC addresses of the devices because we generate a single ISO for all the hosts, containing the configuration files for all of them, and then we use the MAC addresses to determine what files should be applied in each particular host.

dustymabe: > and at the point that `coreos-copy-firstboot-network.service` runs those directories aren't populated yet

Could you add dependencies to your service such that those things exist but then also still ues `Before=coreos-copy-firstboot-network.service` or does that create an ordering loop?

jhernand: If I add `After=systemd-udev-settle.service` things seem to work correctly @dustymabe, so I assume there are no loops. I will update the pull request with that change.

dustymabe: Cool. Glad it's working for you!

dustymabe:
> ok. I think I understand the problem better now. What you are trying to do here is pretty much exactly what we are doing in [`coreos-copy-firstboot-network.service`](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.service). My suggestion to you is to run your service `Before=coreos-copy-firstboot-network.service` and place your output files into `/etc/coreos-firstboot-network`. Those files will then get picked up by `coreos-copy-firstboot-network.service` and propagated appropriately.
> 
> IOW, hook back into our established flow for this rather than trying to figure out the ordering for yourself and also competing with our flow at the same time.

This was done now. See discussion in https://github.com/openshift/assisted-service/pull/5107#discussion_r1158929240

internal/constants/scripts.go
```diff
@@ -29,12 +29,13 @@ package constants
 const PreNetworkConfigScript = `#!/bin/bash
 
 # The directory that contains nmconnection files of all nodes
-NMCONNECTIONS_DIR=/etc/assisted/network
+SRC_DIR=/etc/assisted/network
+DST_DIR=/etc/coreos-firstboot-network
```
carbonin, line 33: So is this now the destination regardless of if we're running during initrd or from ignition?

Please make sure you test both full and minimal ISO with this change.

internal/constants/scripts.go
```diff
@@ -186,13 +179,28 @@ find_host_directory_by_mac_address
...
-    rm -f ${ETC_NETWORK_MANAGER}/*
-    echo "Removing default connection files in '$ETC_NETWORK_MANAGER'"
     set_logical_nic_mac_mapping
     update_interface_names_by_mapping_file
     copy_nmconnection_files_to_nm_config_dir
 fi
 echo "PreNetworkConfig End"
 `
+
+const PreNetworkConfigService = `
```
carbonin, line 189: If this is only being used for minimal iso can we name it as such?

Maybe `MinimalISONetworkConfigService` or something?

carbonin: Otherwise it's easy to confuse with `templates/pre-network-manager-config.service`

jhernand: Done.

internal/constants/scripts.go
jhernand: No @carbonin , it isn't the same destination, unfortunately. I was trying to make this work only for the minimal ISO configuration. Now that it works I have adjusted it so that it works for both the minimal and full ISO. Will be in the next iteration of the patch.

internal/constants/scripts.go
```diff
@@ -196,3 +196,22 @@ if [ "$host_dir" ]
 fi
```
dustymabe, line 196: The `rm -f ${ETC_NETWORK_MANAGER}/*` is probably wrong now (reading the comments).

Though it shouldn't be needed in the flow where we are using `coreos-copy-firstboot-network.service` See [the code](https://github.com/coreos/fedora-coreos-config/blob/773528175403f8b6f78c70f74f0480bc6b00c038/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/coreos-copy-firstboot-network.sh#L16-L18).

internal/constants/scripts.go
```diff
@@ -52,7 +52,7 @@ declare -A logical_nic_mac_map
 
 # Find destination directory based on ISO mode
 if [[ -f /etc/initrd-release ]]; then
-  ETC_NETWORK_MANAGER="/run/NetworkManager/system-connections"
+  ETC_NETWORK_MANAGER="/etc/coreos-firstboot-network"
 else
   ETC_NETWORK_MANAGER="/etc/NetworkManager/system-connections"
 fi
```
dustymabe, line 58: I'm naive here as to how the assisted installer works. Seems like it would be nice if we could unify and just do it one way (i.e. always run in the initramfs). I'm sure there are great reasons for why it's done this way, but would like to learn the context if someone is willing to explain (happy to do a screenshare too).

carbonin: I can talk though this and all the history if you want to put something on my calendar

carbonin: The short answer to this though is that we're creating the placeholder for this initrd data when we create our minimal ISO (we're not currently using `coreos-installer` for any of this) so we don't have the ability to inject data into the initrd using the full ISO.

carbonin (approved):

internal/constants/scripts.go
jhernand: @dustymabe I am reluctant to change the behavior of the script because it is used also with the full ISO (where we don't use `coreos-copy-firstboot-network.service`) and with older versions of OpenShift (4.12, with RHEL 8, for example). But I can do something like this:

```diff
diff --git a/internal/constants/scripts.go b/internal/constants/scripts.go
index 7182fedf5..e2f7179fb 100644
--- a/internal/constants/scripts.go
+++ b/internal/constants/scripts.go
@@ -53,8 +53,10 @@ declare -A logical_nic_mac_map
 # Find destination directory based on ISO mode
 if [[ -f /etc/initrd-release ]]; then
   ETC_NETWORK_MANAGER="/etc/coreos-firstboot-network"
+  CLEAN_ETC_NETWORK_MANAGER="true"
 else
   ETC_NETWORK_MANAGER="/etc/NetworkManager/system-connections"
+  CLEAN_ETC_NETWORK_MANAGER="false"
 fi
 echo "Info: ETC_NETWORK_MANAGER was set to $ETC_NETWORK_MANAGER"
 
@@ -188,7 +190,10 @@ if [ "$host_dir" ]
   then
     # remove default connection file create by NM(nm-initrd-generator). This is a WA until
     # NM is back to supporting priority between nmconnections
-    rm -f ${ETC_NETWORK_MANAGER}/*
+    if [ "${CLEAN_ETC_NETWORK_MANAGER}" == "true" ]
+    then
+      rm -f ${ETC_NETWORK_MANAGER}/*
+    fi
     echo "Removing default connection files in '$ETC_NETWORK_MANAGER'"
     set_logical_nic_mac_mapping
     update_interface_names_by_mapping_file
```

Would that address your concern?

carbonin: If `coreos-copy-firstboot-network` is also removing these files, does it actually hurt to leave this here or is it just that it's not necessary?

dustymabe: > If `coreos-copy-firstboot-network` is also removing these files, does it actually hurt to leave this here or is it just that it's not necessary?

It doesn't hurt to leave it here, but can be quite confusing to others and your future self as well :).
You can leave the code the same, but I would highly recommend a comment to explain things a bit.

jhernand: Added some comments explaining that, please check again.

internal/constants/scripts.go
```diff
@@ -186,8 +188,12 @@ find_host_directory_by_mac_address
...
 if [ "$host_dir" ]
   then
-    # remove default connection file create by NM(nm-initrd-generator). This is a WA until
-    # NM is back to supporting priority between nmconnections
+    # Remove default connection file create by NM(nm-initrd-generator). This is a WA until
+    # NM is back to supporting priority between nmconnections.
+    #
+    # Note that when this is intended mostly for the full ISO scenario. In the minimal ISO scenario
+    # this script runs before the 'coreos-coreos-copy-firstboot-network' service, and that script
+    # already removes the connection files, but it doesn't hurt to do it again.
```
dustymabe, line 196: new comment LGTM

PR #5110, opened 2023-04-06
MGMT-13890: Drain before deprovisioning spoke nodes

carbonin:
This PR drains the spoke node before the BMH is configured for deprovisioning.

A few annotations are added to control and monitor this process:

- `bmac.agent-install.openshift.io/drain-started-at`: Set by BMAC when the drain process starts - this is used by BMAC to determine when the drain process has timed out.
- `bmac.agent-install.openshift.io/drain-timeout`: Set by the user in `time.Duration` format (eg `5m`). Used by BMAC to determine when the drain process has timed out.
- `bmac.agent-install.openshift.io/drain-status`: Set by BMAC to inform the user of the status of the drain process. Will be one of the following constant strings, or the error string returned from the drain function

```
drainStatusSuccess    = "drain succeeded"
drainStatusInProgress = "draining in progress"
drainStatusTimeout    = "drain timed out"
```

The `Drainer` interface is created to allow for easier unit testing, but it otherwise just a passthrough to the `kubectl` functions.

https://issues.redhat.com/browse/MGMT-13890

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Manually in dev-scripts.

- Created a 5 node spoke cluster
- Set `bmac.agent-install.openshift.io/remove-agent-and-node-on-delete` on the BMH
- Delete the BMH

Observe the node drain before deleting

Also tested with the timeouts by attempting to delete the other worker node which would result in drain timing out as the ingress pods had no way to be rescheduled.

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -1551,3 +1609,100 @@ func (r *BMACReconciler) configureBMHForCleaning(ctx context.Context, log logrus
...
+	node, err := client.GetNode(nodeName)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get node %s", nodeName)
+		return false, err
+	}
+
+	out := new(bytes.Buffer)
+	drainHelper := &drain.Helper{
+		Client:              clientset,
+		Ctx:                 ctx,
```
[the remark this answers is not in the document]
carbonin: The drain helper is short-lived. Only for the scope of the drain functions.

My guess is that they just didn't want a million function parameters and didn't want to define the functions on a struct (for some reason).

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -336,7 +348,53 @@ func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.Fiel
...
+
+	if !drainTimedOut && bmh.GetAnnotations()[BMH_NODE_DRAIN_STATUS_ANNOTATION] != drainStatusSuccess {
+		dirty := false
+		log.Info("Draining agent node...")
+		if _, present := bmh.GetAnnotations()[BMH_NODE_DRAIN_START_ANNOTATION]; !present {
+			setAnnotation(&bmh.ObjectMeta, BMH_NODE_DRAIN_START_ANNOTATION, time.Now().UTC().Truncate(time.Second).Format(time.RFC3339))
+			dirty = true
+
+		}
+		requeue, err := r.drainAgentNode(ctx, log, agent)
```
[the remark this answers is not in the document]
carbonin: A single call will take no longer than 20 seconds.
We set a timeout [here](https://github.com/openshift/assisted-service/pull/5110/files#diff-86415614c6ed40c56c5ca2b5f751a65746ef5c83b61e6f021a4f5d817e4425baR1682)

I don't want to use goroutines because it makes this significantly more difficult to track.
Each reconcile loop would have to know if there was a drain routine already running and if to start a new one for this.

I took this logic (including the timeout values and behavior) directly from the CAPI controller implementation so I'm hoping it will work fine for us if it works for them.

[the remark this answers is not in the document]
mhrivnak: Do we have the controller set for multiple concurrent reconciles at least? The default is `1`, so we should make sure we've configured it to have additional available, so other resources can be reconciled, especially in case there are multiple drains in a row.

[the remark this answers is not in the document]
mhrivnak: It gets set here: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/controller#Options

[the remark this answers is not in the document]
carbonin: Or one thread might be waiting on drain while another sees that the drain has timed out and deletes the BMH.

I think the worst that will happen is that we get an error about the object already being deleted or already being modified.

osherdp:
/uncc
de-assigning myself as a reviewer, because I'm not that familiar with the relevant code

internal/controller/controllers/bmh_agent_controller.go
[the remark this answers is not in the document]
filanov: Because bmh controller doesn't access the DB then there should not be any races with the DB. 
If this configuration is per controller then we have a lower risk enabling it for the bmh controller, after all the controller know how to handle object versioning, the issue can be that we will not handle it well in the code, meaning that you will not be able to submit something because the object have changed by another reconcile and you will just skip or log an error. 

BMH is handling very specific cases, if there will be delay and some host will not boot or will not update the url on time it's not a big deal. 

Because it can cause unexpected behavior and we are close to the feature freeze, i suggest to work with a the default `MaxConcurrentReconciles = 1` and open a different ticket to handle  it in 2.9.0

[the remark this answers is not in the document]
carbonin: I can also pretty easily add config for this with a default to 1 if we want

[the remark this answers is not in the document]
carbonin: Added the option for a user to set the number of concurrent reconciles. Currently set the default to 1 since this entire feature is opt-in.

carbonin:
@mhrivnak what do you think?

carbonin:
Looks like @mhrivnak is still away.

@avishayt @eranco74 can one of you take a look at this?

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -1551,3 +1616,100 @@ func (r *BMACReconciler) configureBMHForCleaning(ctx context.Context, log logrus
...
+		log.WithError(err).Errorf("failed to cordon node %s: output: %s", nodeName, out)
+		return false, errors.Wrapf(err, "failed to cordon node %s", nodeName)
+	}
+	if err := r.Drainer.RunNodeDrain(drainHelper, nodeName); err != nil {
+		log.WithError(err).Warnf("failed to drain node %s within timeout", nodeName)
+		log.Debugf("node %s drain output: %s", nodeName, out)
+		return true, nil
+	}
+
+	return false, nil
```
eranco74, line 1714: Return the error here and you can just eliminate the bool from the return, no?

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -336,7 +354,53 @@ func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.Fiel
...
+		}
+	}
+
+	dirty := false
+	if val := bmh.GetAnnotations()[BMH_NODE_DRAIN_STATUS_ANNOTATION]; val != drainStatusTimeout && val != drainStatusSuccess {
+		if drainTimedOut {
+			log.Warn("Timed out waiting to drain node, continuing with deprovisioning")
+			setAnnotation(&bmh.ObjectMeta, BMH_NODE_DRAIN_STATUS_ANNOTATION, drainStatusTimeout)
+			dirty = true
+		} else {
```
eranco74, line 397: Why do we set the status to success?
How do we know the drain completed successfully and not still in progress?

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -336,7 +354,53 @@ func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.Fiel
 		return reconcileComplete{stop: true, dirty: true}
 	}
 
-	if dirty := r.configureBMHForCleaning(ctx, log, bmh); dirty {
+	drainTimedOut, err := nodeDrainTimedout(bmh)
```
eranco74, line 357: Consider extracting all this to another function (`handleDrain`)

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -1551,3 +1616,100 @@ func (r *BMACReconciler) configureBMHForCleaning(ctx context.Context, log logrus
...
+	return false
+}
+
+// Mostly taken from https://github.com/kubernetes-sigs/cluster-api/blob/539760a/internal/controllers/machine/machine_controller.go#L586
+func (r *BMACReconciler) drainAgentNode(ctx context.Context, log logrus.FieldLogger, agent *aiv1beta1.Agent) (bool, error) {
+	log = log.WithFields(logrus.Fields{
+		"spoke_cluster_name":      agent.Spec.ClusterDeploymentName.Name,
+		"spoke_cluster_namespace": agent.Spec.ClusterDeploymentName.Namespace,
+	})
+	// get spoke connection
```
eranco74, line 1661: Add log saying "Draining agent node"

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -1551,3 +1616,100 @@ func (r *BMACReconciler) configureBMHForCleaning(ctx context.Context, log logrus
...
 	return
 }
+
+// nodeDrainTimedout returns true if the node drain should be skipped due to timout, false if draining should be run
+func nodeDrainTimedout(bmh *bmh_v1alpha1.BareMetalHost) (bool, error) {
+	annotations := bmh.GetAnnotations()
+	timeoutValue, timeoutExists := annotations[BMH_NODE_DRAIN_TIMEOUT_ANNOTATION]
+	startValue, startExists := annotations[BMH_NODE_DRAIN_START_ANNOTATION]
+
+	if !startExists || !timeoutExists {
```
eranco74, line 1626: If timeoutExists you can just return true, though I don't see where this annotation is being set.

eranco74 (commented):
Are we sure assisted is the one that should handle node draining?
Isn't this a task for some external service? 
The code looks good, just nits

internal/controller/controllers/bmh_agent_controller.go
carbonin: If the drain is still in progress `drainAgentNode` would have returned `requeue == true` and we would have returned from the function on line 387.

The only way we get here is if the drain returns successfully.
I can see if the logic can be improved here, but I remember this being a bit of a pain.

internal/controller/controllers/bmh_agent_controller.go
carbonin: I have a log in the caller, but I can move it here if you prefer.

internal/controller/controllers/bmh_agent_controller.go
carbonin: No, because we need to differentiate between three cases:
1. Some error happened that caused us not to attempt a drain (failed to connect, couldn't get node, etc)
    - Don't requeue the reconcile from the caller and return an error
    - `return false, error`
2. Drain ran but errored (timeout or otherwise)
    - Requeue the reconcile from the caller until we hit the timeout
    - `return true, nil`
3. Drain ran and succeeded
    - Continue the reconcile
    - `return false, nil`

carbonin:
> Are we sure assisted is the one that should handle node draining?

If we want to drain these nodes then yes, there's not another tool that can get in the middle of this processing.

> Isn't this a task for some external service?

In reality this is something MAPI or CAPI should handle, but since we're not using them in this case we have to do it.
At the start of this feature I looked into using MAPI on the spoke to handle this for us, but since the BMHs there don't have credentials (and we don't always have baremetal platform spokes) it wasn't an option.

internal/controller/controllers/bmh_agent_controller.go
carbonin: Done :+1:

internal/controller/controllers/bmh_agent_controller.go
carbonin: I refactored this in the new function. Hopefully it's a bit more clear now.

carbonin:
Retested this with the last round of changes and looks good.

@mhrivnak can you take a look as I believe @eranco74 and @avishayt are now out for the next few days.

internal/controller/controllers/drainer.go
```diff
@@ -0,0 +1,24 @@
...
+
+type KubectlDrainer struct{}
+
+var _ Drainer = &KubectlDrainer{}
+
+func (d *KubectlDrainer) RunCordonOrUncordon(helper *drain.Helper, node *corev1.Node, desired bool) error {
+	return drain.RunCordonOrUncordon(helper, node, desired)
+}
+
+func (d *KubectlDrainer) RunNodeDrain(helper *drain.Helper, nodeName string) error {
```
tsorya, line 22: What will happened on timeout?

internal/controller/controllers/drainer.go
```diff
@@ -0,0 +1,24 @@
...
+type Drainer interface {
+	RunCordonOrUncordon(helper *drain.Helper, node *corev1.Node, desired bool) error
+	RunNodeDrain(helper *drain.Helper, nodeName string) error
+}
+
+type KubectlDrainer struct{}
+
+var _ Drainer = &KubectlDrainer{}
+
+func (d *KubectlDrainer) RunCordonOrUncordon(helper *drain.Helper, node *corev1.Node, desired bool) error {
```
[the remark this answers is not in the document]
carbonin: No, we set it up to ignore them [here](https://github.com/openshift/assisted-service/pull/5110/files#diff-86415614c6ed40c56c5ca2b5f751a65746ef5c83b61e6f021a4f5d817e4425baR1695)

internal/controller/controllers/drainer.go
carbonin: Yes, [we configure a 20s timeout for each drain attempt](https://github.com/openshift/assisted-service/pull/5110/files#diff-86415614c6ed40c56c5ca2b5f751a65746ef5c83b61e6f021a4f5d817e4425baR1698) and the user can configure a global timeout after which we'll just move on to deprovisioning.

tsorya: Sorry i missed the bmh fille

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -1551,3 +1575,150 @@ func (r *BMACReconciler) configureBMHForCleaning(ctx context.Context, log logrus
...
+		// When the node is unreachable and some pods are not evicted for as long as this timeout, we ignore them.
+		drainHelper.SkipWaitForDeleteTimeoutSeconds = 60 * 5 // 5 minutes
+	}
+	if err := r.Drainer.RunCordonOrUncordon(drainHelper, node, true); err != nil {
+		log.WithError(err).Errorf("failed to cordon node %s: output: %s", nodeName, out)
+		return false, errors.Wrapf(err, "failed to cordon node %s", nodeName)
+	}
+	if err := r.Drainer.RunNodeDrain(drainHelper, nodeName); err != nil {
+		log.WithError(err).Warnf("failed to drain node %s within timeout", nodeName)
+		log.Debugf("node %s drain output: %s", nodeName, out)
```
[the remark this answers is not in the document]
carbonin: Yes, it's very large

internal/controller/controllers/bmh_agent_controller.go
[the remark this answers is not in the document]
mhrivnak: Is the expectation then that if a user wants to use this feature, they should also adjust the number of concurrent reconciles? How will they know to do that?

If we'll have some users running with concurrent reconciles, it would help to get as much testing of that configuration as we can. Making multiple the default would ensure it gets maximum testing, and then we have an easy way to go back to 1.

If we're not confident in being able to run concurrent reconciles with stability, I'm not sure how we can confidently deliver this feature as a good experience, unless we go with the other approach of doing drains in dedicated goroutines.

Thoughts?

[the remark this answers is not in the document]
carbonin: I'm not really concerned.

I don't think the delay in processing reconciles when a node is getting drained is going to impact user experience in a very noticeable way.
Also I think the people using this feature are a significant minority. Adding concurrency for the few people using this at the risk of causing difficult to debug issues for the majority of users isn't a good idea this late in the release cycle IMO.

I'd be okay enabling it by default after this PR gets backported.

PR #5120, opened 2023-04-16
MGMT-13083: limit the size of release binaries

slaviered:
with lru policies

The single source of truth is the file system, so no delete failures are retried.

instead of locking in memory, the caller receives a hard link to the file and deletes it after the processing is completed.

There is also a GC logic for links if they are left behind.

The hard link makes sure that the file is always available to the caller, even if eviction is taking place in parallel (in the rare event of a request storm)

A priority queue is used to keep track of the oldest file. Whenever a file is accessed its mtime is updated so it won't get evicted.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/ignition/ignition.go
+++ b/internal/ignition/ignition.go
@@ -42,6 +42,7 @@ import (
 	"github.com/openshift/assisted-service/internal/versions"
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/assisted-service/pkg/auth"
+	"github.com/openshift/assisted-service/pkg/executer"
 	logutil "github.com/openshift/assisted-service/pkg/log"
 	"github.com/openshift/assisted-service/pkg/mirrorregistries"
 	"github.com/openshift/assisted-service/pkg/s3wrapper"
@@ -234,6 +235,7 @@ type installerGenerator struct {
 	providerRegistry              registry.ProviderRegistry
 	installerReleaseImageOverride string
 	clusterTLSCertOverrideDir     string
+	installerCache                *installercache.Installers
 }
 
 // IgnitionConfig contains the attributes required to build the discovery ignition file
@@ -280,7 +282,7 @@ func NewBuilder(log logrus.FieldLogger, staticNetworkConfig staticnetworkconfig.
 // NewGenerator returns a generator that can generate ignition files
 func NewGenerator(serviceBaseURL string, workDir string, installerDir string, cluster *common.Cluster, releaseImage string, releaseImageMirror string,
 	serviceCACert string, installInvoker string, s3Client s3wrapper.API, log logrus.FieldLogger, operatorsApi operators.API,
-	providerRegistry registry.ProviderRegistry, installerReleaseImageOverride, clusterTLSCertOverrideDir string) Generator {
+	providerRegistry registry.ProviderRegistry, installerReleaseImageOverride, clusterTLSCertOverrideDir string, storageCapacityLimit int64) Generator {
 	return &installerGenerator{
 		cluster:                       cluster,
 		log:                           log,
@@ -297,6 +299,7 @@ func NewGenerator(serviceBaseURL string, workDir string, installerDir string, cl
 		providerRegistry:              providerRegistry,
 		installerReleaseImageOverride: installerReleaseImageOverride,
 		clusterTLSCertOverrideDir:     clusterTLSCertOverrideDir,
+		installerCache:                installercache.New(installerDir, storageCapacityLimit, log),
 	}
 }
 
@@ -331,11 +334,19 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte,
 	}
 	defer removeIcspFile(icspFile)
 
-	installerPath, err := installercache.Get(g.installerReleaseImageOverride, g.releaseImageMirror, g.installerDir,
-		g.cluster.PullSecret, platformType, icspFile, log)
+	mirrorRegistriesBuilder := mirrorregistries.New()
+	ocRelease := oc.NewRelease(&executer.CommonExecuter{}, oc.Config{
+		MaxTries: oc.DefaultTries, RetryDelay: oc.DefaltRetryDelay}, mirrorRegistriesBuilder)
+
+	release, err := g.installerCache.Get(g.installerReleaseImageOverride, g.releaseImageMirror,
+		g.cluster.PullSecret, ocRelease, platformType, icspFile)
 	if err != nil {
 		return errors.Wrap(err, "failed to get installer path")
 	}
+	//cleanup resources at the end
+	defer release.Release()
+
+	installerPath := release.Path
 	installConfigPath := filepath.Join(g.workDir, "install-config.yaml")
 
 	g.enableMetal3Provisioning, err = common.VersionGreaterOrEqual(g.cluster.Cluster.OpenshiftVersion, "4.7")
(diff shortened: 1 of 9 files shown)


openshift/template.yaml
```diff
@@ -181,6 +181,9 @@ parameters:
 - name: ENABLE_DATA_COLLECTION
   value: "false"
   required: false
+- name: INSTALLER_CAPACITY
```
osherdp, line 184: let's also add some kind of feature-flag. it can either be a separate parameter, or you can assign some value of ``INSTALLER_CAPACITY`` that implies "do no caching". I tend to the first option, but it's not critical

osherdp (changes requested):

internal/installercache/installercache.go
```diff
@@ -1,61 +1,160 @@
...
+		i.evict()
+
+		//extract the binary
+		_, err = ocRelease.Extract(i.log, releaseID, releaseIDMirror, i.cacheDir, pullSecret, platformType, icspFile)
 		if err != nil {
 			return "", err
 		}
-		r.path = path
+	} else {
+		//update the file mtime to signal it was recently used
```
osherdp, line 66: nit: shouldn't we only change ``atime``? because if we only accessed the file (and not modified it) the ``mtime`` isn't expected to change in linux

osherdp (changes requested):
partially reviewed. just a bunch of nits

internal/installercache/installercache.go
mhrivnak: `atime` is not updated for every access on most linux systems, so it's not a reliable or accurate depiction of when a file was used. You can imagine the performance impact if every file access included a write to disk, however small. See the `relatime` mount option for details.

internal/ignition/ignition.go
```diff
@@ -280,7 +282,7 @@ func NewBuilder(log logrus.FieldLogger, staticNetworkConfig staticnetworkconfig.
 // NewGenerator returns a generator that can generate ignition files
 func NewGenerator(serviceBaseURL string, workDir string, installerDir string, cluster *common.Cluster, releaseImage string, releaseImageMirror string,
 	serviceCACert string, installInvoker string, s3Client s3wrapper.API, log logrus.FieldLogger, operatorsApi operators.API,
-	providerRegistry registry.ProviderRegistry, installerReleaseImageOverride, clusterTLSCertOverrideDir string) Generator {
+	providerRegistry registry.ProviderRegistry, installerReleaseImageOverride, clusterTLSCertOverrideDir string, installerCap int) Generator {
```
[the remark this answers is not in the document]
mhrivnak: It would help to define this in the doc string.

internal/installercache/priorityqueue.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (pq *PriorityQueue[T]) Len() int {
+	return pq.items.Len()
+}
+
+func (pq *PriorityQueue[T]) Add(item T) {
+	heap.Push(&pq.items, item)
+}
+
+func (pq *PriorityQueue[T]) Pop() (T, bool) {
```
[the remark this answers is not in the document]
slaviered: Good catch. will do

internal/installercache/priorityqueue.go
```diff
@@ -0,0 +1,57 @@
...
+func (hi heapItems[E]) Len() int { return len(hi) }
+func (hi heapItems[E]) Less(i, j int) bool {
+	return hi[i].Compare(hi[j])
+}
+
+func (hi heapItems[E]) Swap(i, j int) {
+	hi[i], hi[j] = hi[j], hi[i]
+}
+
+func (hi *heapItems[E]) Push(item any) {
```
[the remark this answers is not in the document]
slaviered: Push is part of heap.Interface that does not support generics so I have to use any. E here is not any but has a constraint

internal/installercache/installercache.go
```diff
@@ -1,61 +1,160 @@
...
+		err = os.Chtimes(path, time.Now(), time.Now())
+		if err != nil {
+			return "", errors.Wrap(err, fmt.Sprintf("Failed to update release binary %s", path))
+		}
+	}
+	//return a new hard link to the file
```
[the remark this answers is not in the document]
slaviered: Completely agreed

internal/ignition/ignition.go
```diff
@@ -297,6 +299,7 @@ func NewGenerator(serviceBaseURL string, workDir string, installerDir string, cl
 		providerRegistry:              providerRegistry,
 		installerReleaseImageOverride: installerReleaseImageOverride,
 		clusterTLSCertOverrideDir:     clusterTLSCertOverrideDir,
+		installerManager:              installercache.New(installerDir, installerCap, log),
```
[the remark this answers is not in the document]
slaviered: Actually caching is just a part of what that code do. Maybe InstallerBinaryManager?

internal/installercache/installercache.go
osherdp: yes, I guessed atime is not updated on any change in linux
but my point is about mtime not being updated for any read action

openshift/template.yaml
mhrivnak: Maybe `INSTALLER_CACHE_LIMIT` would clarify that this is imposing a limit on cache size, and that leaving it unset means there's no limit.

internal/installercache/installercache.go
osherdp: any way it's not critical. but if a means "accessed" and m means "modified" it seems more reasonable to only update ``atime`` and leave ``mtime`` updated each time the file actually changes

mhrivnak (commented):
Other than a few small comments, this looks reasonable to me.

osherdp (commented):
looks good to me

slaviered:
test ci/prow/edge-unit-test

internal/installercache/installercache.go
```diff
@@ -1,61 +1,187 @@
...
+		return
+	}
+
+	//prune the hard links just in case the deletion in ignition.go
+	//did not succeeded as expected
+	for idx := 0; idx < len(links); idx++ {
+		finfo := links[idx]
+		grace := time.Now().Add(DeleteGracePeriod).Unix()
+		if finfo.info.ModTime().Unix() < grace {
+			os.Remove(finfo.path)
```
carbonin, line 155: Not sure I understand this.

Couldn't this delete something that is currently being used?
Or is that what the grace period is for? We're assuming that it won't be in use for more than 5 minutes?
If that's the case it's probably worth a comment next to the grace period definition.

internal/installercache/priorityqueue.go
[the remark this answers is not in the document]
carbonin: I still don't understand why the bool return value is here.
The comment doesn't mention it and it's not  being used in the one caller.

internal/installercache/installercache.go
slaviered: Exactly. we assume that it will not be used more than 5 minutes. I will add a comment

slaviered:
Tested E2E and verified that the eviction is working fine

slaviered:
test ci/prow/edge-unit-test

internal/installercache/priorityqueue.go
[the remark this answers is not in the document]
slaviered: Added an explanation

osherdp (approved):

PR #5143, opened 2023-04-21
WIP MGMT-13938: support external platform

adriengentil:
External platform is not supported by `openshift-installer`, but it is
quite simple to patch the manifests using `None` platform as a base
because external platform behaves in the same way.


adriengentil:
/cc @eliorerz 
This is the idea I was think when I talked to you last time

swagger.yaml
```diff
@@ -5317,6 +5317,21 @@ definitions:
...
+    x-go-custom-tag: gorm:"embedded;embeddedPrefix:platform_external_"
+    description: The configuration for the external platform type.
+    properties:
+      platform_name:
+        type: string
+        x-nullable: true
+        enum: ['Unknown', 'oci']
+        default: 'Unknown'
+        description: |-
+          Holds the infrastructure provider name.
```
adriengentil, line 5334: @gamli75 @eliorerz

Don't review the rest for the PR as it's WIP, I want your opinion on this API change for [platform external](https://github.com/openshift/enhancements/blob/master/enhancements/cloud-integration/infrastructure-external-platform-type.md).

This would be the API change we need to let the user customize the platform name in the installer config when choosing external platform type, this value would also be used to customize the assisted service behavior (if `platform_name` is `oci` then allow minimal ISO only).

I set the default to `Unknown` because it's the default in the [api](https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L237).

On create and update requests, we would check that `platform.external` field is set only when `platform.type` is equal to `external`.
On update request, if `platform.type` is different from `external` we will reset `platform.external` to `nil`.

About `platform_name` field, I wonder if shouldn't be better to call this field `profile` as this value conditions `platform_name` value in installer config and changes the behavior of the assisted service? (in that case we would have a `generic` and an `oci` profile)

@avishayt do you have an opinion on this change? Don't hesitate to contact me as I guess you don't have all the context.

eliorerz: I need to take a deeper look in it, but from a first glace, that structure of the API was changed after a lot of discussions. We used to have `vsphere` and `ovirt` under platform definition (the same as you are adding the external one) and we removed it.
Not sure what is the best approach here because (to my understanding) the `exteranl` platform type will not be represented in the `install-config.yaml`.
@avishayt @eranco74 

e.g. 
https://github.com/openshift/assisted-service/blob/cb48591dee6f39382362922d6354743574b7454b/swagger.yaml#L4666-L4678

avishayt: The question is if we want to consider "external" to be a platform and expose that (as in this PR), or expose "oci" as the platform (i.e., just adding "oci" to the enum of supported platforms).
The benefit of the API as in this PR is that it aligns more with openshift installer.  The downside is that those platforms aren't first-class citizens in AI.  For example, say one day we have 2 external platforms ext1 and ext2 (no relation to the file system).  ext1 has properties username and password, and ext2 has the property api_key.  How do we present that to the user in a way that's not confusing?  If they are first-class citizens, we could have each one's properties represented as in the swagger that @eliorerz pasted above for how we had vsphere and ovirt.

adriengentil: >  Not sure what is the best approach here because (to my understanding) the exteranl platform type will not be represented in the install-config.yaml.

FYI, the [implementation is on-going](https://issues.redhat.com/browse/OCPCLOUD-2036)

>  ext1 has properties username and password, and ext2 has the property api_key

in the install config, the external platform takes only 2 parameters:
- the platform name, for telemetry purposes
- whether we want to enable the external platform or not (I think that by default, we want to enable it, otherwise it behaves like none platform)

Then, we expect the user to upload the manifests that will configure the cloud integration (the CCM) at installation time. I think most of the details for the platform will be specified in this step.

I see your point at making the platform first-class citizen, I'll explore it and see with the UI team what it means for them as I think they will have to know if a platform is external or not in order to tune the journey for the user (e.g.: guide them to the upload manifest screen).

Otherwise, the idea I had was to define a profile for the external platform, and use one more indirection to pass additional options, which  I agree, makes things a bit more complex:
```
platform_type: external
external:
  profile: ext1
  ext1:
    username: ...
    password: ...
```

adriengentil: @jkilzi what would it mean on UI side if we define `oci` as a new platform type (under the hood the BE will use the external platform) ?

e.g.:
```
platform:
  type: oci
```

jkilzi: That could work well, could save us passing also `platform.name: 'oci'`. So IIUC `oci` will be an alias for `external`?

swagger.yaml
```diff
@@ -5321,6 +5321,11 @@ definitions:
     properties:
       type:
         $ref: '#/definitions/platform_type'
+      is_external:
+        type: boolean
+        default: false
+        x-nullable: true
+        description: Indicates if the underlying platform type is external (read-only).
```
adriengentil, line 5328: @avishayt @eliorerz @gamli75

As discussed last time, I created a flag `is_external` under the `platform` key, let me know that makes sense.

Also, since it's a read-only field, I plan to ignore any potential value set by the user on cluster create or update. The of `is_external` will be always refreshed against the platform type, is that an acceptable behavior?

eliorerz: Are we sure that we want this field to be read-only? 
If the user will set `oci` as platform type I thought that in this case he will must specify that this platform is external, else he might be "surprised" later when he will need to add custom manifest, also on the install-config the type will be `external` and not `oci` as expected.

avishayt: 1. @eliorerz: Why make the user set the property?  What set it to "false"?  The idea behind the property was that the service is indicating to the user (and the UI, and possibly future validations) that the platform requires adding manifests.
2. @eliorerz: Is the fact that the install-config has a different platform name concerning?  In my opinion, that file is an internal implementation detail.  Is this not always the case?
3. @adriengentil: If a property is read-only, the API shouldn't allow setting it.  We should have one object for create/update and one for "GET".

adriengentil: > Why make the user set the property? What set it to "false"? The idea behind the property was that the service is indicating to the user (and the UI, and possibly future validations) that the platform requires adding manifests.

Selecting OCI platform type implies the use of the external platform. Right now, I don't see a use case where it would be valid to have `platform.type = oci` and `platform.is_external = false`.

BTW, uploading manifests is not something we will enforce in the code yet, mainly because the first user will be Oracle and we want to let them experiment, also it's a bit complicated to check the content of the manifests is actually deploying a Cloud Controller Manager. We may revise this decision later.

> If a property is read-only, the API shouldn't allow setting it. We should have one object for create/update and one for "GET".

That makes a lot of sense, thank you!

avishayt: > Actually, swagger provides a [property to make a field read only](https://swagger.io/docs/specification/data-models/data-types/), that should be enough, no?

Interesting, I hadn't seen that.  We've never used it before in the project, but you're welcome to try it and see if our code generation works with it.

eliorerz: > edit: Actually, swagger provides a [property to make a field read only](https://swagger.io/docs/specification/data-models/data-types/), that should be enough, no?

I'm not sure that this is available on [swagger v2](https://swagger.io/specification/v2/)

eliorerz: > 1. Why make the user set the property

Mainly for 2 reasons:

- On creation the user acknowledge that this is an external platform 
- We theoretically can create vsphere (or other platforms) as external platform

However, I think that the second reason can be handled in the future if needed.

avishayt: > On creation the user acknowledge that this is an external platform
I would rather tell the user (and UI) then have the user tell us

> We theoretically can create vsphere (or other platforms) as external platform
I agree that this can be handled in the future if needed.

internal/provider/common.go
```diff
@@ -70,17 +86,33 @@ func checkPlatformWrongParamsInput(platform *models.Platform, userManagedNetwork
...
 }
 
+func createPlatformFromType(platformType models.PlatformType) *models.Platform {
+	platform := &models.Platform{
+		Type: &platformType,
+	}
+	return updatePlatform(platform)
+}
+
+func updatePlatform(platform *models.Platform) *models.Platform {
```
eliorerz, line 107: maybe give it a better name? I couldn't understand from the its name what it's doing

eliorerz (changes requested):
Note that [this doc](https://docs.google.com/document/d/1P09ArJSVeqMI3T4yu7rvemyJkYs-FmIF32Z8WDY--jM) might need an update

internal/bminventory/inventory.go
```diff
@@ -2535,6 +2535,7 @@ func (b *bareMetalInventory) updateNetworkTables(db *gorm.DB, cluster *common.Cl
 func (b *bareMetalInventory) updatePlatformParams(params installer.V2UpdateClusterParams, updates map[string]interface{}, usages map[string]models.Usage) error {
 	if params.ClusterUpdateParams.Platform != nil && common.PlatformTypeValue(params.ClusterUpdateParams.Platform.Type) != "" {
 		updates["platform_type"] = params.ClusterUpdateParams.Platform.Type
+		updates["platform_is_external"] = swag.BoolValue(params.ClusterUpdateParams.Platform.IsExternal)
```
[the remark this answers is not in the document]
adriengentil: I see, we update the DB directly from the user params struct instead of constructing a cluster struct in between to apply the business rules, your suggestion makes more sense in this way.

swagger.yaml
adriengentil: > Interesting, I hadn't seen that. We've never used it before in the project, but you're welcome to try it and see if our code generation works with it.

tested it, and it works well!

internal/bminventory/inventory.go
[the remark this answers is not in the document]
adriengentil: As discussed, it looks like the best way to update the flag is in `GetActualUpdateClusterPlatformParams`/`GetActualCreateClusterPlatformParams` as these functions contain the business logic to return a valid `Platform` struct.

internal/provider/common.go
adriengentil: renamed it `updatePlatformIsExternal`

internal/provider/common.go
```diff
@@ -42,96 +58,143 @@ func isUMNAllowedForCluster(cluster *common.Cluster) bool {
 	return cluster != nil && isUMNAllowedForPlatform(cluster.Platform)
 }
 
+func isUMNMandatoryForCluster(cluster *common.Cluster) bool {
+	return cluster != nil && isUMNMandatoryForPlatform(cluster.Platform)
+}
+
 func checkPlatformWrongParamsInput(platform *models.Platform, userManagedNetworking *bool, cluster *common.Cluster) error {
+	// check that user does not try to disable userManagedNetworking when cluster is SNO
```
gamli75, line 66: can we remove these lines?

adriengentil: done !

internal/bminventory/inventory.go
avishayt, line 2538: Should call common.isPlatformExternal() - user is not setting it

adriengentil: I update the flag in `GetActualUpdateClusterPlatformParams` / `GetActualCreateClusterPlatformParams`, those methods create  a valid `Platform` struct out of `ClusterUpdateParams`/`ClusterCreateParams`.

avishayt: Right, thanks

avishayt:
Looks good to me.  Does anyone else need to review?

adriengentil:
> Looks good to me. Does anyone else need to review?

@gamli75 I think we are good to go?

gamli75 (approved):

PR #5161, opened 2023-04-27
MGMT-14237: Remove all spoke resources when deleting the node

carbonin:
This commit deletes the spoke machine and BMH when removing the node if
the entire cluster is not being removed (indicated by the existence of the cluster deployment).

Removing the machine will allow the same host to be cleanly added back as a node in the future.

The machine and BMH are removed directly and separately if the machine is not part of a machine set.
If the machine is part of a set the machine is annotated to indicate it is the one which should be removed and the machineset is configured for autoscaling. Once this is done the BMH is removed. This triggers the machine set to scale down, removing the previously annotated machine.

Due to OCPBUGS-7581 the BMH will likely not be removed cleanly so we also remove the finalizers before attempting the delete.

In the case of a none-platform cluster we won't find the machine and will delete the node directly.

https://issues.redhat.com/browse/MGMT-14237

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

1.  Deployed a 5 node spoke
2. Removed the worker nodes one at a time by deleting the hub BMH
3. Observed that the machineset was scaled correctly and the machines and BMHs were removed.
4. Added the workers back as day-2 nodes
5. Removed a single worker
6. Observed that the machine, bmh and node were removed successfully
7. Deleted all remaining BMHs in the hub
8. Observed that hosts were all deprovisioned successfully and agents were removed from the hub

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -43,6 +43,7 @@ import (
 	"github.com/openshift/assisted-service/restapi/operations/installer"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	machinev1beta1 "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
@@ -124,6 +125,17 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 
 	res, err := r.handleAgentFinalizer(ctx, log, agent)
 	if res != nil {
+		if err != nil {
+			conditionsv1.SetStatusConditionNoHeartbeat(&agent.Status.Conditions, conditionsv1.Condition{
+				Type:    aiv1beta1.CleanupCondition,
+				Status:  corev1.ConditionFalse,
+				Reason:  aiv1beta1.CleanupFailedReason,
+				Message: err.Error(),
+			})
+			if updateErr := r.Status().Update(ctx, agent); updateErr != nil {
+				log.WithError(updateErr).Error("failed to update agent status")
+			}
+		}
 		return *res, err
 	}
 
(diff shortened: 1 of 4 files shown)


internal/controller/controllers/agent_controller.go
```diff
@@ -211,6 +212,160 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+		return errors.Wrap(err, "failed to parse BMH name")
+	}
+
+	bmh := &bmh_v1alpha1.BareMetalHost{}
+	bmhKey := client.ObjectKey{Namespace: bmhNamespace, Name: bmhName}
+	if err = spokeClient.Get(ctx, bmhKey, bmh); err != nil {
+		return errors.Wrapf(err, "failed to get BMH %s", bmhNSName)
+	}
+
+	// TODO: remove this once OCPBUGS-7581 is fixed
```
mhrivnak, line 240: With a fix on the way (https://github.com/metal3-io/baremetal-operator/pull/1259), is there any way we can avoid doing this at all? Or will the timing not work out for that?

Do we have a jira issue to come back and remove this?

internal/controller/controllers/agent_controller.go
```diff
@@ -211,6 +212,160 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+	}
+
+	if err = spokeClient.Delete(ctx, bmh); err != nil {
+		return errors.Wrapf(err, "failed to delete BMH %s", bmhNSName)
+	}
+
+	return nil
+}
+
+// removeSpokeResources removes all relevan resources from the agent's spoke cluster
```
mhrivnak, line 254: s/relevan/relevant/

mhrivnak (commented):
Is there any way to differentiate a spoke cluster that's temporarily not available, due to network outage or similar, vs. a spoke cluster that's expected to be gone? As we get into more and further edge use cases, constant connectivity can't be assumed.

internal/controller/controllers/agent_controller.go
carbonin: I'm not sure we'd be able to get that backported this late in the 4.13 cycle.

Even if that PR were merged today it would probably be a week or so before it gets pulled into `openshift/baremetal-operator` then we'd need to do the bug dance to get it into 4.13

I can open a ticket to track removing this.

carbonin:
>Is there any way to differentiate a spoke cluster that's temporarily not available, due to network outage or similar, vs. a spoke cluster that's expected to be gone? As we get into more and further edge use cases, constant connectivity can't be assumed.

Not that I can think of.

Doing something like checking the other agents in the cluster deployment (to see if they're also being deleted) is a bit racy. Even then we'd have to make the assumption that the hub knows about all the nodes that exist in the spoke.

Maybe the underlying http response codes would be different, but probably not in all cases.

I'd rather not delay the finalizer handling (or implement more complex timeout logic) if we don't have a handle on what these kinds of cases look like. Also I think I'd rather implement this in a follow up either way since without this ZTP users can't re-add hosts to the cluster after they were removed. Even if it doesn't work out sometimes having this here is better than nothing.

One thing I'd like to handle though is how to indicate to the user that we failed to clean up. The problem is that the agent resource is being deleted so reporting status there doesn't really work unless we just don't allow the deletion to continue. That may still be better than trying to figure out a way to coordinate getting something into the clusterdeployment status (assuming that also isn't being deleted).

mhrivnak:
> > Is there any way to differentiate a spoke cluster that's temporarily not available, due to network outage or similar, vs. a spoke cluster that's expected to be gone? As we get into more and further edge use cases, constant connectivity can't be assumed.
> 
> Not that I can think of.
> 
> Doing something like checking the other agents in the cluster deployment (to see if they're also being deleted) is a bit racy. Even then we'd have to make the assumption that the hub knows about all the nodes that exist in the spoke.
> 
> Maybe the underlying http response codes would be different, but probably not in all cases.
> 
> I'd rather not delay the finalizer handling (or implement more complex timeout logic) if we don't have a handle on what these kinds of cases look like. Also I think I'd rather implement this in a follow up either way since without this ZTP users can't re-add hosts to the cluster after they were removed. Even if it doesn't work out sometimes having this here is better than nothing.
> 
> One thing I'd like to handle though is how to indicate to the user that we failed to clean up. The problem is that the agent resource is being deleted so reporting status there doesn't really work unless we just don't allow the deletion to continue. That may still be better than trying to figure out a way to coordinate getting something into the clusterdeployment status (assuming that also isn't being deleted).

Can we examine the AgentClusterInstall, and as long as it exists and isn't being deleted, then we'll re-try cleanup?

carbonin:
> Can we examine the AgentClusterInstall, and as long as it exists and isn't being deleted, then we'll re-try cleanup?

I'm okay with this. This just assumes that someone isn't deleting the masters without removing the cluster resources which would be a very strange thing to do.

Then at that point their way out would be to remove the ACI anyway so I don't think we need another timeout mechanism.

internal/controller/controllers/agent_controller.go
carbonin: Also just thought of this ...

Should we be doing this update after we delete to ensure BMO doesn't add it back before we can delete it and have it GCd?

carbonin: It might be harder to unit test, but probably worth it to avoid this bug.

carbonin:
Also if we go the route of ensuring the cleanup is done I guess I should remove all the contingencies.

For example, currently if we fail to find a machineset that a machine is referencing I delete the machine. We'd probably want to remove that now.

carbonin:
> got a "not found" error from the spoke API. This is confirmation of state in that cluster and can be acted upon.

Okay, so this sounds like we want to do all the corner case handling that I already have, but only if we get a not found error vs what I'm doing now which is the same, but for every error (and then not retrying).

/hold to handle :point_up:

internal/controller/controllers/agent_controller.go
```diff
@@ -211,6 +212,160 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 	return r.updateStatus(ctx, log, agent, origAgent, &h.Host, h.ClusterID, nil, false)
 }
 
+func splitNamespacedName(nsName string) (string, string, error) {
```
paul-maidment, line 215: The splitting of namespaced names seems like functionality that could be used more broadly than just in the context of BMH, This function seems more like it should reside somewhere "common"

carbonin: I'm not sure how frequently we're consuming a combined namespace and name string. Also this is usable across the controllers package as it is and I doubt we'd have many places outside of this package that need this kind of logic.

internal/controller/controllers/agent_controller.go
carbonin: Created https://issues.redhat.com/browse/MGMT-14495 to track removing the finalizer removal once BMO fixes their bug.

carbonin: I've tested this a lot and haven't seen this. Resolving this conversation for now.

internal/controller/controllers/agent_controller.go
```diff
@@ -234,16 +395,16 @@ func (r *AgentReconciler) handleAgentFinalizer(ctx context.Context, log logrus.F
...
-					nodeName := getAgentHostname(agent)
-					if err := spokeClient.DeleteNode(ctx, nodeName); err != nil {
-						log.WithError(err).Errorf("failed to delete spoke node %s", nodeName)
-					} else {
-						log.Infof("spoke node %s deleted", nodeName)
+					log.WithError(err).Errorf("failed to check cluster deployment presence")
+					return &ctrl.Result{}, err
+				}
+				if clusterExists {
+					if err := r.removeSpokeResources(ctx, log, agent); err != nil {
```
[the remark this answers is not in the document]
carbonin: Added this in the caller

carbonin:
/unhold

@mhrivnak @avishayt can you take a look when you get a chance?

internal/controller/controllers/agent_controller.go
```diff
@@ -211,6 +223,166 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
...
+	}
+
+	machineNSName, haveMachine := node.GetAnnotations()["machine.openshift.io/machine"]
+	if !haveMachine {
+		return deleteNode()
+	}
+
+	machineNamespace, machineName, err := splitNamespacedName(machineNSName)
+	if err != nil {
+		log.WithError(err).Error("failed to parse machine name from node annotation")
```
mhrivnak, line 305: It would help to log the value of the annotation here. In case this happens in the real world, it may be hard for our future selves to figure out what went wrong if we immediately deleted the input without preserving it somehow.

mhrivnak (commented):
Just one tweak. Otherwise it looks good. If I'm following the logic correctly, if the ClusterDeployment exists, if the `spokeClient` encounters an error other than "NotFound", the finalizer is left in place. Right?

carbonin:
>If I'm following the logic correctly, if the ClusterDeployment exists, if the spokeClient encounters an error other than "NotFound", the finalizer is left in place. Right?

Correct, mostly.

We're not checking generally for not found since some calls could be hub cluster calls (like in the case that we can't get the spoke client).
Also we error and leave the finalizer if we hit any error when editing an object that we were able to previously query since we'd just hit the not found case on the get on the next reconcile attempt if that error suddenly was not found.

internal/controller/controllers/agent_controller.go
carbonin: Will do

mhrivnak (approved):
Good work!

PR #5180, opened 2023-05-03
MGMT-14509 return 404 if cluster not found when host try to register

filanov:
When host try to register if the clsuter is not exist the service should return 404, if it return 500 the host will continue to try and register again and again, it will csae the api rate to spike.

The root cause is not clear yet, meaning that we don't understand why infra-env still exist when cluster was deleted, it can probably happen if there is a bug in delete logic.

Eventailly the infra-env sohuld be deleted by GC bug we would like to avoid api skipes

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] Unit tests

## Checklist

## Reviewers Checklist


carbonin:
I think this is probably fine, but there's some strange logic in `getBoundClusterForUpdate`.

It might be able to be improved, but this is a fine fix for now.

carbonin (approved):

PR #5185, opened 2023-05-04
OCPBUGS-13081: Support by-path root device hints

zaneb:
In many cases, the /dev/disk/by-path symlink is the only way to stably identify a disk without having prior knowledge of the hardware from some external source (e.g. a spreadsheet of disk serial numbers). It should be possible to specify this path in the root device hints.

Metal³ is also changing to allow this, and both uses of the Metal³ root device hints (i.e. in the baremetal-operator and in assisted-service) should behave consistently.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] Ephemeral installer
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/hostutil/host_utils.go
+++ b/internal/host/hostutil/host_utils.go
@@ -190,7 +190,7 @@ func GetAcceptableDisksWithHints(disks []*models.Disk, hints *bmh_v1alpha1.RootD
 		}
 
 		if hints != nil {
-			if hints.DeviceName != "" && hints.DeviceName != disk.Path {
+			if hints.DeviceName != "" && hints.DeviceName != disk.Path && hints.DeviceName != disk.ByPath {
 				continue
 			}
 
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -623,9 +623,13 @@ func (r *BMACReconciler) reconcileAgentInventory(log logrus.FieldLogger, bmh *bm
 
 	// Add storage
 	for _, d := range inventory.Disks {
+		device := d.Path
+		if d.ByPath != "" {
+			device = d.ByPath
+		}
 		// missing WWNVendorExtension, WWNWithExtension
 		disk := bmh_v1alpha1.Storage{
-			Name:         d.Path,
+			Name:         device,
 			HCTL:         d.Hctl,
 			Model:        d.Model,
 			SizeBytes:    bmh_v1alpha1.Capacity(d.SizeBytes),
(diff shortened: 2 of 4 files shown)


internal/host/hostutil/host_utils.go
```diff
@@ -190,7 +190,7 @@ func GetAcceptableDisksWithHints(disks []*models.Disk, hints *bmh_v1alpha1.RootD
 		}
 
 		if hints != nil {
-			if hints.DeviceName != "" && hints.DeviceName != disk.Path {
+			if hints.DeviceName != "" && hints.DeviceName != disk.Path && hints.DeviceName != disk.ByPath {
```
avishayt, line 193: Any objection to adding ByID as well?

zaneb: Yes, 3:
- The ByID path isn't reported in the hardwareDetails of the BMH, so users have no way of knowing what the right path to put is anyway.
- The same support is not available in Metal³, so it will require users to understand the implementation details of how the BMH CR is being used to know what values are acceptable.
- The existing WWN, Model, and Serial Number fields already provide a more flexible version of the same functionality - they can be tested independently, and partial matches are allowed for the model.

The first two could be resolved by implementing this in ironic as well. The third is why nobody is likely to bother.

avishayt: 1 and 2: This code doesn't depend on ironic.  The assisted agent reports by-id.
3: I think it's confusing to users that they can use the device name and by-path but not by-id.

filanov: agree with Avishay

zaneb: The problem is we're not talking about two similar fields in different APIs here. It's literally the _same_ field.
In ZTP, assisted reads the rootDeviceHint out of the BareMetalHost object. It shouldn't interpret the field differently from the way Metal³ would, because then you have to explain to users "you can pass by-id to _this_ BareMetalHost but not _that_ BareMetalHost because they're actually interpreted by different backends" :exploding_head: 
Similarly in the agent installer we'll take the rootDeviceHints from the baremetal platform section so that the same install-config can be used with IPI and Agent installer, but adding by-id to only one of those will mean that the field has different meanings depending on which method you choose.

zaneb: In fact the converged ZTP means hosts are provisionable using _either_ the assisted agent (e.g. agent CAPI provider) or directly through CAPBM, since it runs both agents. So the same field on the _same_ host could end up meaning different things depending on which CAPI provider provisions the host.

avishayt: OK, I'm convinced.  Thanks.

PR #5186, opened 2023-05-04
MGMT-14425: Changing event counts behavior

danmanor:
At present, we collect 4 headers for event counts of each severity level, which indicate the number of filtered events based on their severity. However, we aim to modify these counts to exclude severity filtering. To accomplish this, we require an additional count header that represents the total number of retrieved events after all filtering.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/events/event.go
+++ b/internal/events/event.go
@@ -270,11 +270,6 @@ func filterEvents(tx *gorm.DB, clusterID *strfmt.UUID, hostIds []strfmt.UUID, in
 
 		tx = tx.Where("events.cluster_id = ?", clusterID.String())
 
-		// filter by event severity
-		if severities != nil {
-			tx = tx.Where("events.severity IN (?)", severities)
-		}
-
 		// filter by event message
 		if message != nil {
 			tx = tx.Where("events.message LIKE ?", fmt.Sprintf("%%%s%%", *message))
@@ -423,7 +418,7 @@ func isDescending(order *string) (*bool, error) {
 	return nil, errors.New("incompatible order parameter")
 }
 
-func (e Events) queryEvents(ctx context.Context, params *common.V2GetEventsParams) ([]*common.Event, *common.EventSeverityCount, error) {
+func (e Events) queryEvents(ctx context.Context, params *common.V2GetEventsParams) ([]*common.Event, *common.EventSeverityCount, *int64, error) {
 
 	cleanQuery := e.db.Session(&gorm.Session{})
 	tx := e.db.Where("category IN (?)", params.Categories)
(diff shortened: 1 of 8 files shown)


internal/events/event.go
```diff
@@ -459,29 +465,30 @@ func (e Events) queryEvents(ctx context.Context, params *common.V2GetEventsParam
 
 	params.Limit, params.Offset = preparePaginationParams(params.Limit, params.Offset)
 	if *params.Limit == 0 {
-		return make([]*common.Event, 0), eventSeverityCount, nil
+		return make([]*common.Event, 0), eventSeverityCount, swag.Int64(0), nil
```
paul-maidment, line 468: I don't think we should be returning zero for eventCount here. I think we should return eventCount.

It looks like params.Limit is being used to paginate through results and it seems from the existing code that the significance of `*params.Limit == 0` is to return an empty resultset in the paged data. This does not however change the reality that the `eventCount` has a value that is potentially nonzero.

Better to return `eventCount` here instead of `swag.Int64(0)`

danmanor: Thanks for that, I fixed it

internal/events/event.go
```diff
@@ -459,29 +465,30 @@ func (e Events) queryEvents(ctx context.Context, params *common.V2GetEventsParam
 
 	params.Limit, params.Offset = preparePaginationParams(params.Limit, params.Offset)
 	if *params.Limit == 0 {
-		return make([]*common.Event, 0), eventSeverityCount, nil
+		return make([]*common.Event, 0), eventSeverityCount, &eventCount, nil
```
paul-maidment, line 468: I think `&eventCount` should be changed to `eventCount` here. `&eventCount` will give you the address of the pointer, this is not what you want.

Proof:
https://go.dev/play/p/PGWEbbUj7r-

internal/events/event.go
```diff
@@ -459,29 +465,30 @@ func (e Events) queryEvents(ctx context.Context, params *common.V2GetEventsParam
...
 	}
 
 	err = tx.Offset(int(*params.Offset)).Limit(int(*params.Limit)).Find(&events).Error
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
-	return events, eventSeverityCount, nil
+	return events, eventSeverityCount, &eventCount, nil
```
paul-maidment, line 476: I think &eventCount should be changed to eventCount here. &eventCount will give you the address of the pointer, this is not what you want.

Proof:
https://go.dev/play/p/PGWEbbUj7r-

internal/events/event.go
paul-maidment: There also needs to be a test to assert that you are getting the right stuff from this function after your changes.

paul-maidment: As discussed, the change makes sense

internal/events/event.go
paul-maidment: No longer relevant

paul-maidment (approved):
lgtm

PR #5194, opened 2023-05-09
MGMT-14530: Changing event message filtering to escape wildcards and not be case sensetive

danmanor:
To obtain accurate results in `SQL`, we need to escape the letters `%` and `_ `as they are considered wildcards. Also, we aim to retrieve messages that correspond to events regardless of the letter case.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/events/event.go
+++ b/internal/events/event.go
@@ -272,7 +272,7 @@ func filterEvents(tx *gorm.DB, clusterID *strfmt.UUID, hostIds []strfmt.UUID, in
 
 		// filter by event message
 		if message != nil {
-			tx = tx.Where("events.message LIKE ?", fmt.Sprintf("%%%s%%", *message))
+			tx = tx.Where("LOWER(events.message) LIKE ?", fmt.Sprintf("%%%s%%", strings.ToLower(escapePlaceHolders(*message))))
 		}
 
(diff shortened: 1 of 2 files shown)


internal/events/event.go
```diff
@@ -272,7 +272,7 @@ func filterEvents(tx *gorm.DB, clusterID *strfmt.UUID, hostIds []strfmt.UUID, in
 
 		// filter by event message
 		if message != nil {
-			tx = tx.Where("events.message LIKE ?", fmt.Sprintf("%%%s%%", *message))
+			tx = tx.Where("LOWER(events.message) LIKE ?", processEventMessage(*message))
```
rccrdpccl, line 275:
```suggestion
			tx = tx.Where("LOWER(events.message) LIKE %?%", quoteQueryPlaceHolders(*message))
```
I think this would make the intent clearer, WDYT?

danmanor: I think GORM doesn't take `%?%` well

PR #5217, opened 2023-05-16
NO-ISSUE: Fix install customization doc

empovit:
This is a documentation fix:
* The request params for manifests now include `folder` and `file_name`
* Resource name in the example must comply with K8s rules for resource names
* HTTPS has to be used where TLS is enabled, e.g. with the SaaS API

### DIFF
--- a/docs/user-guide/install-customization.md
+++ b/docs/user-guide/install-customization.md
@@ -8,6 +8,8 @@ These will only take effect after the machine config operator is up and running 
 
 ### Create a cluster manifest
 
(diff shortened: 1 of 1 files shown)


omertuc:
Shouldn't we delete those docs now that we have official docs? To encourage people to maintain the new ones

empovit:
> Shouldn't we delete those docs now that we have official docs? To encourage people to maintain the new ones

@omertuc Good point. I didn't know there were official docs and where to find them :(

gamli75 (approved):

PR #5222, opened 2023-05-17
MGMT-14356: Update 4.13 release image to use GA version

eliorerz:
/cc @eliorerz


gamli75:
you need to update your PR and only make it as a default, it's redundant to https://github.com/openshift/assisted-service/pull/5225

PR #5226, opened 2023-05-18
MGMT-13643: added MaxLength to additional_trust_bundle

danielerez:
As the max length of a string in postgres is 65,535 bytes, additional_trust_bundle property length should be limited on: infra-env-create-params and infra-env-update-params.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/vendor/github.com/openshift/assisted-service/models/infra_env_create_params.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/infra_env_create_params.go
@@ -28,6 +28,7 @@ type InfraEnvCreateParams struct {
 	// infra-env will trust the certificates in this bundle. Clusters formed
 	// from the hosts discovered by this infra-env will also trust the
 	// certificates in this bundle.
+	// Max Length: 65535
 	AdditionalTrustBundle string `json:"additional_trust_bundle,omitempty"`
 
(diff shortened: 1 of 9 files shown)


omertuc:
RSA-signed-RSA certs can reach 2000 bytes, this means we're limiting users to only around 30-40 certs. Maybe it's enough, but we should maybe consider just making the database field bigger?

Of course doesn't have to happen in this PR, since this PR is an obvious net positive

danielerez:
> PEM DER RSA-signed-RSA x.509 certs can range around 1000-2800 bytes, this means we're limiting users to only around 24-50 certs. Maybe it's enough, but we should maybe consider just making the database field bigger?
> 
> Of course doesn't have to happen in this PR, since this PR is an obvious net positive

Yeah, good point. Not sure how relevant this use-case for users, but we can indeed consider it in the context of another ticket.

filanov (approved):
the test is not really required because you are checking a generated code

PR #5207, opened 2023-05-11
OCPBUGS-13310 support setting CPUPartitioningMode with install config…

filanov:
… overrides

A feature that is being intoduced in https://github.com/openshift/installer/commit/206c7b88af92fe19b48fa33372dd89ed4a9cf02d Required for ZTP installation

## List all the issues related to this PR

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [X] Cloud
- [X] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [x] No tests needed

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #5227, opened 2023-05-18
Fix cpu partitioning struct tag/field

carbonin:
The struct tag for cpu partitioning was using but in order for the
install config to be rendered correctly in all cases it need the yaml
tag like the other fields.

Without the yaml tag cpu partitioning was being included in the install
config even if it was empty with an invlid name.

This meant that all 4.10 installs always failed to prepare for
installation with an event like:

```
Failed to prepare the installation due to an unexpected error: failed
generating install config for cluster
1583dc9b-e059-48fc-96b9-5e711d1de6b1: error running openshift-install
manifests, level=fatal msg=failed to fetch Master Machines: failed to
load asset "Install Config": failed to unmarshal install-config.yaml:
failed to parse first occurence of unknown field: error unmarshaling
JSON: while decoding JSON: json: unknown field "cpupartitioning" : exit
status 1. Please retry later
```

To correct this while also maintaining the current overrides behavior
the field name is updated to match the API name (cpuPartitioningMode)
and the json struct tag is changed to yaml.

This allows for overrides to work in a case-insensitive way (as is
supported currently for all other fields) as well as for the install
config yaml to render properly.

NOTE:

This field name is intentionally different than the one in the openshift
installer repo. This allows for us to take advantage of the
case-insensitive matching allowed by json unmarshaling while also using
the correct name.

Originally committed in https://github.com/openshift/assisted-service/pull/5207

https://issues.redhat.com/browse/OCPBUGS-13310

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/installcfg/builder/builder_test.go
+++ b/internal/installcfg/builder/builder_test.go
@@ -610,7 +610,7 @@ aEA8gNEmV+rb7h1v0r3EwDQYJKoZIhvcNAQELBQAwYTELMAkGA1UEBhMCaXMxCzAJBgNVBAgMAmRk
 		err = yaml.Unmarshal(data, &result)
 		Expect(err).ShouldNot(HaveOccurred())
 		// test that overrides worked
-		Expect(string(result.CPUPartitioning)).Should(Equal("AllNodes"))
+		Expect(string(result.CPUPartitioningMode)).Should(Equal("AllNodes"))
 	})
 
 	Context("networking", func() {
--- a/internal/installcfg/installcfg.go
+++ b/internal/installcfg/installcfg.go
@@ -195,7 +195,7 @@ type InstallerConfigBaremetal struct {
 	Platform              Platform             `yaml:"platform"`
 	BootstrapInPlace      BootstrapInPlace     `yaml:"bootstrapInPlace,omitempty"`
 	FIPS                  bool                 `yaml:"fips"`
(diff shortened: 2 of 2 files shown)


omertuc:
I don't understand what this have to do with JSON vs YAML and why do we still keep both the JSON and YAML tags after this change?

carbonin:
The failure was caused because we serialized a struct with an empty cpu partitioning field into yaml, but without the yaml tag.

This left us with a field like `cpupartitioning: ""` in the yaml which isn't recognized by the openshift installer.

All the other fields don't have json tags but this field is the only one where the serialized name doesn't match the field name so it also needs to be specified to ensure it's consistent. We use json deserialization to do the install config overrides.

That said, after doing a brief test with this I noticed that all the json field names are capitalized because we are not currently providing the json struct tags. I'll change this one to agree with the others for consistency.

carbonin:
/hold

Going to ensure consistency for overrides

carbonin:
This should be ready for final review whenever you get the chance @omertuc

PR #5233, opened 2023-05-22
MGMT-13997: Issues when using multiple bonds with CIM Assisted Installer

ori-amizur:
When using NMSTATE if a user defines several NMSTATE files for a single host, the data from only single NMSTATE file will be appied on the host. The reason for that is that the script that places the configuration files in the Network Manager configuration directory (system-connections), stops searching when a configuration files from single NMSTATE file is found.
This commit changes that.  It gathers all the confuguration files generated from NMSTATE files that the host that runs the script.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @carbonin


internal/constants/scripts.go
```diff
@@ -66,140 +70,149 @@ function map_host_macs_to_interfaces() {
...
 
-    if [[ -z "$(ls -A $d/*.nmconnection)" ]]
+    if [[ -z "$(ls -A $host_src_dir/*.nmconnection)" ]]
     then
       echo "Warning: Host directory does not contain any nmconnection files, skipping"
       continue
     fi
+    host_work_dir=$(mktemp -d)
 
     # check if mapping file contains mac-address that exists on the current host
```
carbonin, line 160: Is this still accurate?

internal/constants/scripts.go
```diff
@@ -66,140 +70,149 @@ function map_host_macs_to_interfaces() {
...
+          extension="${filename#*.}"
+          [[ -n "$prefix" ]] && [[ "$prefix" == "$logical_nic" ]] &&  cp $nmfile $host_work_dir/${host_nic}.${extension} && rm -f $nmfile
+        done
+      fi
+    fi
+  done
+}
 
-    mapping_file="${d}/${MAC_NIC_MAPPING_FILE}"
+# If any nmstate files of the current host directory were copied, copy the remaining nmstate files
```
carbonin, line 98: Why do we do this? Can you add some more explanation here and to the commit message?

internal/constants/scripts.go
```diff
@@ -66,140 +70,149 @@ function map_host_macs_to_interfaces() {
...
+function copy_other_files() {
+  host_src_dir=$1
+  host_work_dir=$2
+  if [ "$(ls -A $host_work_dir )" ] && [[ -n "$(ls -A ${host_src_dir}/*.nmconnection)" ]] ; then
+    for nmfile in ${host_src_dir}/*.nmconnection ; do
+      filename=$(basename $nmfile)
+      target_file=$host_work_dir/$filename
+
+      # If the same filename was not used already, then use it. Otherwise generate another name.
```
carbonin, line 107: Is the file name important? Does it still need to match the device name at this point?

internal/constants/scripts.go
```diff
@@ -66,140 +70,149 @@ function map_host_macs_to_interfaces() {
...
+  for mac_address in $(cat $mapping_file | cut -d= -f1 | tr '[:lower:]' '[:upper:]');  do
+    host_nic="${host_macs_to_hw_iface[${mac_address}]:-}"
+    if [[  -n "$host_nic" ]]; then
+      logical_nic="${logical_mac_nic_map[${mac_address}]:-}"
+      if [[ -n "$(ls -A $host_src_dir/*.nmconnection )" ]] ; then
+        for nmfile in $host_src_dir/*.nmconnection ; do
+          filename=$(basename $nmfile)
+          prefix="${filename%%.*}"
+          extension="${filename#*.}"
+          [[ -n "$prefix" ]] && [[ "$prefix" == "$logical_nic" ]] &&  cp $nmfile $host_work_dir/${host_nic}.${extension} && rm -f $nmfile
```
carbonin, line 91: Would it be easier for this to be a move rather than copy and delete?

internal/constants/scripts.go
```diff
@@ -66,140 +70,149 @@ function map_host_macs_to_interfaces() {
...
+function copy_other_files() {
+  host_src_dir=$1
+  host_work_dir=$2
+  if [ "$(ls -A $host_work_dir )" ] && [[ -n "$(ls -A ${host_src_dir}/*.nmconnection)" ]] ; then
+    for nmfile in ${host_src_dir}/*.nmconnection ; do
+      filename=$(basename $nmfile)
+      target_file=$host_work_dir/$filename
+
+      # If the same filename was not used already, then use it. Otherwise generate another name.
+      if ! [[ ${allocated_nics[$filename]} ]] ; then
```
carbonin, line 108: Only these "other" files make it into `allocated_nics`, right?
That means that we could overwrite a "real" file with a duplicated one here I think.

Do you think it would be safer/easier to fail this copy instead of trying to recover with a generated name?

internal/constants/scripts.go
```diff
@@ -66,140 +70,149 @@ function map_host_macs_to_interfaces() {
...
+        target_file=$host_work_dir/${name}.${extension}
+        sed  -e "s/=$prefix\$/=$name/g" -e "s/=$prefix\./=$name\./g"  $nmfile > $target_file && rm -f $nmfile
+      fi
+      allocated_nics[$filename]=true
+    done
+  fi
+}
+
+# Update references to host nics in the nmstate files
+function update_physical_nics_references() {
```
carbonin, line 124: Why do this later vs editing the file when we select it to go into the working directory in `copy_physical_nics_files`?
I don't think we'll have hardware names for virtual interfaces (bonds/vlans).

internal/constants/scripts.go
```diff
@@ -66,140 +70,149 @@ function map_host_macs_to_interfaces() {
...
+# Make sure we do not run any of the following functions if there was no matching config.
+if [ "$(ls -A $work_dir )" ];  then
     # Remove default connection file create by NM(nm-initrd-generator). This is a WA until
     # NM is back to supporting priority between nmconnections.
     #
     # Note that when this is intended mostly for the full ISO scenario. In the minimal ISO scenario
     # this script runs before the 'coreos-coreos-copy-firstboot-network' service, and that script
     # already removes the connection files, but it doesn't hurt to do it again.
-    rm -f ${ETC_NETWORK_MANAGER}/*
+    mkdir -p ${ETC_NETWORK_MANAGER}
```
carbonin, line 209: Why did this need to be added?

internal/constants/scripts.go
```diff
@@ -66,140 +70,149 @@ function map_host_macs_to_interfaces() {
...
     echo "Removing default connection files in '$ETC_NETWORK_MANAGER'"
-    set_logical_nic_mac_mapping
-    update_interface_names_by_mapping_file
-    copy_nmconnection_files_to_nm_config_dir
+    rm -f ${ETC_NETWORK_MANAGER}/*
+    echo "Copying files from working directory to '$ETC_NETWORK_MANAGER'"
+    cp $work_dir/* ${ETC_NETWORK_MANAGER}
```
carbonin, line 213: Maybe add `-v` here?

carbonin:
Also ccing @paul-maidment since you worked on this script most recently I thnk

internal/constants/scripts.go
ori-amizur: It was in different location

internal/constants/scripts.go
ori-amizur: This is how vlan connection file looks like:
```
[connection]
autoconnect=true
autoconnect-slaves=-1
id=eth1.101
interface-name=eth1.101
type=vlan
uuid=6d0f1528-fd9c-52a9-9aa5-b825aa883bc3

[ipv4]
method=disabled

[ipv6]
method=disabled

[vlan]
id=101
parent=eth1

```
As you can see, the parent hw interface is referenced here

internal/constants/scripts.go
ori-amizur: The files that represent hw nics are renamed to their corresponding nics.  Assuming we have 2 `bond0` files in 2 nmstates.  The first one will use the name `bond0` and the second one will use an extension

internal/constants/scripts.go
ori-amizur: OK

internal/constants/scripts.go
ori-amizur: Not for files that don't match device names (example: bond)

internal/constants/scripts.go
ori-amizur: 1. That what was done before
2. Some nmstate generate several files.  Bond generate a file for the bond, and a file for every slave.  The slave files are copied in the first function.  The bond file is copied here.

internal/constants/scripts.go
ori-amizur: Yes

internal/constants/scripts.go
```diff
@@ -5,29 +5,30 @@ package constants
...
+//		+-- host2
+//		      +--- *.nmconnection
+//		      +--- mac_interface.ini
+//	 1. *.nmconnections - files generated by nmstate based on yaml files provided by the user
+//	 2. mac_interface.ini - the file contains mapping of mac-address to logical interface name.
+//	    There are two usages for the file:
+//	 1. Map logical interface name to MAC Address of the host. The logical interface name is a
+//	    name provided by the user for the interface. It will be replaced by the script with the
+//	    actual network interface name.
+//	 2. Identify the host directory which belongs to the current host by matching a MAC Address
```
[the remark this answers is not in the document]
ori-amizur: That what it is called (the directory names), I didn't change that.
The change is that there might be many directories matching the current host, not just the first one.
Do you think it should be called differently ?

internal/constants/scripts.go
```diff
@@ -5,29 +5,30 @@ package constants
...
+//	    from the mapping file with host network interfaces.
```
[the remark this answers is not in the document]
ori-amizur: Goland did it for me by default.  Will change back.

internal/constants/scripts.go
ori-amizur: OK

internal/constants/scripts.go
```diff
@@ -66,140 +70,156 @@ function map_host_macs_to_interfaces() {
...
-  echo "Info: Checking '$mapping_file' for logical nic to mac mappings"
-  # initialize logical_nic_mac_map with mapping file entries
-  readarray -t lines < "${mapping_file}"
-  for line in "${lines[@]}"
-  do
-    mac=${line%%=*}
-    nic=${line#*=}
-    logical_nic_mac_map[$nic]=${mac^^}
+      # Copy the rest of the files from the host directory if a file was copied from the host directory
+      copy_other_files $host_src_dir $host_work_dir
```
paul-maidment, line 170: I notice that a lot of the echo statements that were in the previous implementation have been removed in this PR. 

The original intent of these echo statements is to provide some useful output in the cases where there is failure. This was added deliberately.

I think we should make an effort to continue to give verbose feedback on what we are processing.

internal/constants/scripts.go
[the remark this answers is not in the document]
carbonin:
```suggestion
//	 2. Identify if a host directory belongs to the current host by matching a MAC Address
```

internal/constants/scripts.go
carbonin: Oh so you're replacing both the usages in the file corresponding to that interface as well as the usage in the bond/vlans. Okay makes sense.

internal/constants/scripts.go
carbonin: I still think this should be a failure case.

I don't think we should be creating a new name when the user already had the chance to pick a correct one.

internal/constants/scripts.go
carbonin: Ah okay, it probably makes more sense to be right before the copy, right?

internal/constants/scripts.go
ori-amizur: What should we do for failure case? skip it?

carbonin: How visible of an error can we create? Anything we print should be in the journal, right?

internal/constants/scripts.go
ori-amizur: Yes

internal/constants/scripts.go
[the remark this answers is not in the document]
ori-amizur: OK

internal/constants/scripts.go
ori-amizur: I am not sure.
With this change we rename files that start with physical nic names.  This causes files having same local nic names to be renamed to different names which means they will coexist together.
Example: `eth0.nmconnection` in 2 different directories might be renamed to `ens2.nmconnection` and `ens3.nmconnection`.
If the interface is virtual (for example bond), then if we have file `bond0.nmconnection` in 2 different directories, this will fail while it is possible that it will succeed.  Why should we fail it if we allow similar duplication for physical nics ?

internal/constants/scripts_test.go
```diff
@@ -0,0 +1,170 @@
...
+			if len(e.replaces) > 0 {
+				Expect(source).ToNot(Equal(destination))
+			}
+			for _, r := range e.replaces {
+				source = strings.ReplaceAll(source, r.from, r.to)
+			}
+			Expect(source).To(Equal(destination))
+		}
+	}
+	It("No host directory", func() {
```
carbonin, line 82: Can you name tests with the context and expected result?

Something like:

```suggestion
	It("script runs successfully when there are no host directories", func() {
```

This is also something that should never happen as we only add the script if we have static networking, but still reasonable to test.

internal/constants/scripts_test.go
```diff
@@ -0,0 +1,170 @@
...
+		copySys()
+		_, err := exec.Command("bash", "-c", fmt.Sprintf("PATH_PREFIX=%s %s", root, cmd)).CombinedOutput()
+		Expect(err).ToNot(HaveOccurred())
+		_, err = os.Stat(systemConnections)
+		Expect(os.IsNotExist(err)).To(BeTrue())
+	})
+	It("With matching host directories", func() {
```
carbonin, line 89:
```suggestion
	It("copies all the network config from matching host directories", func() {
```

internal/constants/scripts.go
```diff
@@ -43,163 +45,183 @@ fi
...
-
-# The mapping file of the current host
-mapping_file=""
+# A mac-to-nic map created from the mapping file associated with the host
+declare -A logical_mac_nic_map
 
-# A nic-to-mac map created from the mapping file associated with the host
-declare -A logical_nic_mac_map
+# Working directory used for storage of processed files before installation in the target directory
+work_dir=$(mktemp -d)
```
carbonin, line 52: When running tests does this get cleaned up?

carbonin: Same for the other `mktemp` calls in this file.

ori-amizur: Here it does https://github.com/openshift/assisted-service/blob/10d0a94f3d6164167056f1f95d2c5cce97705336/internal/constants/scripts.go#L224

ori-amizur: Others are cleaned too

carbonin (approved):

carbonin:
Thanks for adding those tests @ori-amizur !

PR #5249, opened 2023-05-25
MGMT-14582: Set OCI platform behind a capability

adriengentil:
Ensure OCI platform cannot be set if user's organization doesn't
have `bare_metal_installer_platform_oci` capability.

### DIFF
--- a/internal/cluster/validations/validations.go
+++ b/internal/cluster/validations/validations.go
@@ -2,6 +2,7 @@ package validations
 
 import (
 	"bytes"
+	"context"
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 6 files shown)


internal/cluster/validations/validations.go
```diff
@@ -976,3 +978,30 @@ func ValidateArchitectureWithPlatform(architecture *string, platform *models.Pla
...
+		return nil
+	}
+
+	available, err := authzHandler.HasOrgBasedCapability(ctx, *capabilityName)
+	if err == nil && available {
+		return nil
+	}
+
+	if err != nil {
+		log.WithError(err).Errorf("error getting user %s capability", *capabilityName)
```
eliorerz, line 1003: Why is that log not inside `HasOrgBasedCapability`? I saw that it's duplicated 3 times now

eliorerz (changes requested):

internal/cluster/validations/validations.go
adriengentil: Done!

gamli75:
@adriengentil there is a fix for it https://github.com/openshift/assisted-service/pull/5259
rebase the PR

PR #5250, opened 2023-05-27
OCPBUGS-7076: Copy Day2 BMH if Agent is installing

CrystalChun:
[OCPBUGS-7076](https://issues.redhat.com/browse/OCPBUGS-7076)
Previously the BMAC would copy the BMH and 
Machine CRs of a Day2 worker to the spoke cluster before
the Agent began installing. If the Agent doesn't
start installing within the two hours these CRs
were copied over, the CSRs for these nodes will
not be approved by the machine approver pod.

To prevent this from happening, only copy
the BMH and Machine CRs to the spoke cluster
when the Agent has started installing.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @eranco74

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -976,6 +976,12 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		return reconcileComplete{}
 	}
 
+	if !funk.ContainsString([]string{models.HostStatusInstalling, models.HostStatusInstallingInProgress}, agent.Status.DebugInfo.State) {
+		// If agent hasn't started installing, do not create the spoke BMH yet
+		log.Debugf("Agent has not started installing, not creating the spoke BMH for %s on cluster %s", bmh.Name, agent.Spec.ClusterDeploymentName.Name)
+		return reconcileComplete{stop: false}
+	}
+
 	secret, err := getSecret(ctx, r.Client, r.APIReader, key)
 	if err != nil {
 		return reconcileError{err: err}
--- a/internal/controller/controllers/bmh_agent_controller_test.go
+++ b/internal/controller/controllers/bmh_agent_controller_test.go
@@ -1068,7 +1068,7 @@ var _ = Describe("bmac reconcile", func() {
 		})
 
 		Context("when agent role worker and cluster deployment is set", func() {
-			It("should set spoke BMH", func() {
+			It("should set spoke BMH when agent is not installing", func() {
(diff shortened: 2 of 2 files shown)


internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -976,6 +976,12 @@ func (r *BMACReconciler) reconcileSpokeBMH(ctx context.Context, log logrus.Field
 		return reconcileComplete{}
 	}
 
+	if !funk.ContainsString([]string{models.HostStatusInstalling, models.HostStatusInstallingInProgress}, agent.Status.DebugInfo.State) {
```
avishayt, line 979: HostStatusInstalled should also be included, in case the controller somehow missed the installation (e.g., it was down or stuck)

avishayt:
Why not wait until the node is fully installed?  What happens if the installation itself takes more than 2-3 hours (for example, a long time to pull images, or the host booted into the wrong disk and it took time to address)?

CrystalChun:
> Why not wait until the node is fully installed? What happens if the installation itself takes more than 2-3 hours (for example, a long time to pull images, or the host booted into the wrong disk and it took time to address)?

@avishayt Thank you for the feedback! 

I've tested this out by only checking for `HostStatusInstalled`, as suggested, and observed that the Agent gets stuck in `Rebooting` with the state `installing-pending-user-action`. The install never completes.

The spoke cluster's `machine approver` pod errors out with
```
E0602 00:52:57.290085       1 csr_check.go:263] csr-xbfck: failed to find machine for node extraworker-1, cannot approve
I0602 00:52:57.290119       1 controller.go:233] csr-xbfck: CSR not authorized
```
where `extraworker-1` is my day 2 worker node that never gets installed.

Please correct me if I'm wrong, but it seems like copying the BMH & Machine of the day 2 worker node must be done while the Agent is in the installing phase. Otherwise, the CSR will never be approved and the installation will never complete.

avishayt (approved):
Your PR looks good - sorry for the noise.  It slipped my mind that we won't set the status to installed until the node is Ready, and that won't happen until the CSR is approved.

CrystalChun:
> Your PR looks good - sorry for the noise. It slipped my mind that we won't set the status to installed until the node is Ready, and that won't happen until the CSR is approved.

No problem! Thank you again for the review!

PR #5278, opened 2023-06-05
MGMT-14704: Provide info on custom/vs non custom manifest in manifest endpoint.

paul-maidment:
It has been reported that users have no way to determine whether or not a manifest is a custom manifest or one generated by the assisted-service. This PR introduces a change that uses the file system to store additional metadata about a manifest to indicate whether or not the manifest is a custom one.

For example, if the original file is stored in

```
"e09df13f-6f31-42c2-8361-2b5605f80e77/manifests/openshift/99-openshift-machineconfig-master-kargs.yaml"
```

Then, if an only if the manifest is custom -  a corresponding metadata file will be created in the following path.

```
"e09df13f-6f31-42c2-8361-2b5605f80e77/manifest-attributes/openshift/99-openshift-machineconfig-master-kargs.yaml/user-defined"
```

Any internally generated manifests are considered to be "non custom" and are created as such. All other manifests are considered to be custom and will follow the scheme above.

When the user retrieves the manifest list, an additional parameter will be supplied for each manifest to indicate the custom/non custom status of the manifest.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
Building a custom version of the service, creating a cluster and uploading manifests
Fetching the manifest list
Observing that behaviour is as expected.
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -1415,7 +1415,7 @@ func (m *Manager) deleteClusterFiles(ctx context.Context, c *common.Cluster, obj
 	var failedToDelete []string
 	for _, file := range files {
 		//skip log and manifests deletion when deleting cluster files
-		if folder == "" && (strings.Contains(file, "logs") || strings.Contains(file, "manifests")) {
+		if folder == "" && (strings.Contains(file, "logs") || strings.Contains(file, constants.ManifestFolder) || strings.Contains(file, constants.ManifestMetadataFolder)) {
 			continue
 		}
 		log.Debugf("Deleting cluster %s S3 file: %s", c.ID.String(), file)
@@ -1435,7 +1435,11 @@ func (m *Manager) deleteClusterFiles(ctx context.Context, c *common.Cluster, obj
 }
 
 func (m *Manager) deleteClusterManifests(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error {
-	return m.deleteClusterFiles(ctx, c, objectHandler, "manifests")
+	return m.deleteClusterFiles(ctx, c, objectHandler, constants.ManifestFolder)
+}
+
+func (m *Manager) deleteClusterManifestAttributes(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error {
+	return m.deleteClusterFiles(ctx, c, objectHandler, constants.ManifestMetadataFolder)
 }
 
 func (m *Manager) DeleteClusterLogs(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error {
@@ -1486,6 +1490,10 @@ func (m Manager) PermanentClustersDeletion(ctx context.Context, olderThan strfmt
 		if err := m.deleteClusterManifests(ctx, c, objectHandler); err != nil {
 			deleteFromDB = false
 			m.log.WithError(err).Warnf("Failed deleting s3 manifests of cluster %s", c.ID.String())
+		}
+		if err := m.deleteClusterManifestAttributes(ctx, c, objectHandler); err != nil {
+			deleteFromDB = false
+			m.log.WithError(err).Warnf("Failed deleting s3 manifest-attributes of cluster %s", c.ID.String())
 		}
 		if _, err := objectHandler.DeleteObject(ctx, c.ID.String()); err != nil {
 			deleteFromDB = false
--- a/internal/controller/controllers/clusterdeployments_controller.go
+++ b/internal/controller/controllers/clusterdeployments_controller.go
@@ -1149,6 +1149,7 @@ func (r *ClusterDeploymentsReconciler) syncManifests(ctx context.Context, log lo
 
 	// create/update all manifests provided by configmap data
 	for filename, manifest := range manifestsFromConfigMap {
+		isUserManifest := true
 		log.Infof("Creating cluster deployment %s manifest %s", cluster.KubeKeyName, filename)
 		_, err := r.Manifests.CreateClusterManifestInternal(ctx, operations.V2CreateClusterManifestParams{
 			ClusterID: *cluster.ID,
@@ -1156,7 +1157,8 @@ func (r *ClusterDeploymentsReconciler) syncManifests(ctx context.Context, log lo
 				Content:  swag.String(base64.StdEncoding.EncodeToString([]byte(manifest))),
 				FileName: swag.String(filename),
 				Folder:   swag.String(models.ManifestFolderOpenshift),
-			}})
+			}},
+			isUserManifest)
 		if err != nil {
 			log.WithError(err).Errorf("Failed to create cluster deployment %s manifest %s", cluster.KubeKeyName, filename)
 			return err
--- a/internal/manifests/manifests.go
+++ b/internal/manifests/manifests.go
@@ -16,6 +16,7 @@ import (
 	"github.com/go-openapi/swag"
 	"github.com/openshift/assisted-service/internal/cluster"
 	"github.com/openshift/assisted-service/internal/common"
+	"github.com/openshift/assisted-service/internal/constants"
 	manifestsapi "github.com/openshift/assisted-service/internal/manifests/api"
 	"github.com/openshift/assisted-service/internal/usage"
 	"github.com/openshift/assisted-service/models"
@@ -31,9 +32,6 @@ import (
 )
 
 var _ manifestsapi.ManifestsAPI = &Manifests{}
-
-// ManifestFolder represents the manifests folder on s3 per cluster
-const ManifestFolder = "manifests"
 
 // NewManifestsAPI returns manifests API
 func NewManifestsAPI(db *gorm.DB, log logrus.FieldLogger, objectHandler s3wrapper.API, usageAPI usage.API) *Manifests {
@@ -53,7 +51,7 @@ type Manifests struct {
 	usageAPI      usage.API
 }
 
-func (m *Manifests) CreateClusterManifestInternal(ctx context.Context, params operations.V2CreateClusterManifestParams) (*models.Manifest, error) {
+func (m *Manifests) CreateClusterManifestInternal(ctx context.Context, params operations.V2CreateClusterManifestParams, isCustomManifest bool) (*models.Manifest, error) {
 	log := logutil.FromContext(ctx, m.log)
 	log.Infof("Creating manifest in cluster %s", params.ClusterID.String())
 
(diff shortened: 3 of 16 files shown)


swagger.yaml
```diff
@@ -6655,6 +6655,11 @@ definitions:
       file_name:
         type: string
         description: The file name prefaced by the folder that contains it.
+      is_custom_manifest:
+        description: Is this a custom (user uploaded) manifest?
```
omertuc, line 6659: I think we should explain to users what it means when this is `false`, as they don't know the service itself might be also generating manifests for their cluster

```suggestion
        description: Indicates whether this is a manifest that was manually uploaded by the user, as opposed to being internally generated by the Assisted Service
```

omertuc:
Was this tested with our S3 filesystem driver? When running as our online SaaS we store files in S3, not a filesystem, and I think this calls for some testing before we push.

Alternatively, you can wait for it to go through integration/staging and then test it and hope it works

internal/cluster/cluster.go
```diff
@@ -1415,7 +1415,7 @@ func (m *Manager) deleteClusterFiles(ctx context.Context, c *common.Cluster, obj
 	var failedToDelete []string
 	for _, file := range files {
 		//skip log and manifests deletion when deleting cluster files
-		if folder == "" && (strings.Contains(file, "logs") || strings.Contains(file, "manifests")) {
+		if folder == "" && (strings.Contains(file, "logs") || strings.Contains(file, "manifests") || strings.Contains(file, "manifest-attributes")) {
```
omertuc, line 1418: Now that we can tell apart user manifests from our own service manifests, I think it's best that this function deletes the service manifests.

The original motivation for this skip was that we didn't want to destroy use uploaded manifests when the user resets their installation, but preserving the serving generated ones is useless, as they would anyway get regenerated when the user hits install again

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1156,7 +1156,8 @@ func (r *ClusterDeploymentsReconciler) syncManifests(ctx context.Context, log lo
 				Content:  swag.String(base64.StdEncoding.EncodeToString([]byte(manifest))),
 				FileName: swag.String(filename),
 				Folder:   swag.String(models.ManifestFolderOpenshift),
-			}})
+			}},
+			false)
```
omertuc, line 1160: Manifests originating from the configmap are custom user manifests, why don't we consider them custom user manifets?

internal/manifests/manifests.go
```diff
@@ -118,12 +128,20 @@ func (m *Manifests) ListClusterManifestsInternal(ctx context.Context, params ope
 	for _, file := range files {
 		parts := strings.Split(strings.Trim(file, string(filepath.Separator)), string(filepath.Separator))
 		if len(parts) > 2 {
-			manifests = append(manifests, &models.Manifest{FileName: filepath.Join(parts[3:]...), Folder: parts[2]})
+			// Is this a custom manifest?
+			fileName := filepath.Join(parts[3:]...)
+			folder := parts[2]
+			manifestMetadataPrefix := filepath.Join(params.ClusterID.String(), ManifestMetadataFolder)
+			is_custom_manifest, err := m.objectHandler.DoesObjectExist(ctx, filepath.Join(manifestMetadataPrefix, folder, fileName, "user-defined"))
```
omertuc, line 135: I dislike that we only have `user-defined` for custom manifests, and "nothing" for service created manifests.

In reality we need to make a distinction between 3 different types of manifests:
- Manifests created before this PR
- Manifests created after this PR through user upload
- Manifests created after this PR, created by the service

In the current state of this PR, it's impossible to tell apart the first and last bullets. If we also introduce a `service-defined` manifest metadata "magic" file, then we can tell apart all three bullets (both for purposes of user indication, and for the purpose of deleting service manifests during reset as suggested in one of my previous comments)

I suggest that we make the response field a non-`omitEmpty` field, this way its absence would indicate the first bullet, it's presence with `false` would indicate the second bullet, and it's presence with `true` would indicate the third bullet

paul-maidment:
> Was this also tested with the service running with our S3 filesystem driver? When running as our online SaaS we store files in S3, not a filesystem, and I think this PR calls for some testing of that before we merge it.
> 
> Alternatively, you can wait for it to go through integration/staging and then test it and hope it works before it reaches prod

I'll test this in integration and take action if needed. I have not used any new API's so am not expecting problems in this area.

internal/manifests/manifests.go
```diff
@@ -202,16 +282,30 @@ func (m *Manifests) UpdateClusterManifestInternal(ctx context.Context, params op
 		return nil, err
 	}
 
+	isCustomManifest := m.isCustomManifest(ctx, params.ClusterID.String(), srcFolder, srcFileName)
+	if isCustomManifest != nil {
+		err = m.createMetadataForManifest(ctx, params.ClusterID, destPath, *isCustomManifest)
+		if err != nil {
+			return nil, err
+		}
+	}
```
carbonin, line 291: I think we only need this if src and dest path are different.

internal/manifests/manifests.go
```diff
@@ -202,16 +282,30 @@ func (m *Manifests) UpdateClusterManifestInternal(ctx context.Context, params op
 		return nil, err
 	}
 
+	isCustomManifest := m.isCustomManifest(ctx, params.ClusterID.String(), srcFolder, srcFileName)
```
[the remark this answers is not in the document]
carbonin: And even if a user updates a generated manifest, does that not make it a user manifest?

carbonin (commented):
Do you think we could create a subsystem test to cover cases with generated manifests or even manifests that don't have the metadata file?

internal/cluster/cluster.go
omertuc: Please link the ticket once you created it

swagger.yaml
```diff
@@ -6655,6 +6655,10 @@ definitions:
       file_name:
         type: string
         description: The file name prefaced by the folder that contains it.
+      is_custom_manifest:
+        description: Indicates whether this is a manifest that was manually uploaded by the user, as opposed to being internally generated by the Assisted Service
+        type: boolean
```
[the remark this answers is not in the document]
avishayt: Why should we return manifests that the user didn't generate at all?  As a user, I expect to see only the manifests I created.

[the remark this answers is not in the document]
omertuc: I wondered the same - the reason is explained in [the ticket ](https://issues.redhat.com/browse/MGMT-14704) - see the screenshot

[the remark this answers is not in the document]
carbonin: I think @avishayt asked the correct question in the ticket.

The distinction is if we want to communicate to the user where all these manifests are coming from or if we only care about showing them the ones they created. If it's the former I think a string enum is the choice, but if we only need to know if we're showing them or not then using a bool internally (not in the API) makes more sense to me.

internal/cluster/cluster.go
paul-maidment: https://issues.redhat.com/browse/MGMT-15073

internal/manifests/manifests.go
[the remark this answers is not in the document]
paul-maidment: I don't think there is really an opportunity to do this as manifest generation is one of the last actions before install starts, as part of the preparation phase (I think.)
This would present no opportunity for the user to override this.

internal/manifests/manifests.go
```diff
@@ -90,11 +88,50 @@ func (m *Manifests) CreateClusterManifestInternal(ctx context.Context, params op
...
+func (m *Manifests) isUserManifest(ctx context.Context, clusterID string, folder string, fileName string) bool {
+	manifestMetadataPrefix := filepath.Join(clusterID, constants.ManifestMetadataFolder)
+	isCustomManifest, err := m.objectHandler.DoesObjectExist(ctx, filepath.Join(manifestMetadataPrefix, folder, fileName, constants.ManifestSourceUserSupplied))
+	if err != nil {
+		m.log.Warnf("Could not determine if manifest %s/%s is a %s manifest due to error %s, leaving this as nil", folder, fileName, constants.ManifestSourceUserSupplied, err.Error())
+	}
+	return isCustomManifest
+}
+
+func (m *Manifests) deleteMetadataForManifest(ctx context.Context, clusterID strfmt.UUID, path string) error {
```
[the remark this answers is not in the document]
avishayt: Because in the method name you call it "metadata", but in reality you're only dealing with "user supplied".

internal/manifests/manifests.go
```diff
@@ -90,11 +88,50 @@ func (m *Manifests) CreateClusterManifestInternal(ctx context.Context, params op
...
 	log.Infof("Done creating manifest %s for cluster %s", path, params.ClusterID.String())
 	manifest := models.Manifest{FileName: fileName, Folder: folder}
 	return &manifest, nil
 }
 
+func (m *Manifests) isUserManifest(ctx context.Context, clusterID string, folder string, fileName string) bool {
+	manifestMetadataPrefix := filepath.Join(clusterID, constants.ManifestMetadataFolder)
+	isCustomManifest, err := m.objectHandler.DoesObjectExist(ctx, filepath.Join(manifestMetadataPrefix, folder, fileName, constants.ManifestSourceUserSupplied))
+	if err != nil {
+		m.log.Warnf("Could not determine if manifest %s/%s is a %s manifest due to error %s, leaving this as nil", folder, fileName, constants.ManifestSourceUserSupplied, err.Error())
```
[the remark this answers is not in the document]
paul-maidment: As requested, I will propagate this error and make it visible by making it fatal.
This might lead to more actual failures when fetching manifests but will at least make those errors visible.

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1149,14 +1149,16 @@ func (r *ClusterDeploymentsReconciler) syncManifests(ctx context.Context, log lo
 
 	// create/update all manifests provided by configmap data
 	for filename, manifest := range manifestsFromConfigMap {
+		is_user_manifest := true
```
avishayt:
```suggestion
		isUserManifest := true
```

internal/manifests/manifests.go
```diff
@@ -118,12 +152,19 @@ func (m *Manifests) ListClusterManifestsInternal(ctx context.Context, params ope
 	for _, file := range files {
 		parts := strings.Split(strings.Trim(file, string(filepath.Separator)), string(filepath.Separator))
 		if len(parts) > 2 {
-			manifests = append(manifests, &models.Manifest{FileName: filepath.Join(parts[3:]...), Folder: parts[2]})
+			fileName := filepath.Join(parts[3:]...)
+			folder := parts[2]
+			err, is_user_manifest := m.isUserManifest(ctx, params.ClusterID, folder, fileName)
```
avishayt:
```suggestion
			err, isUserManifest := m.isUserManifest(ctx, params.ClusterID, folder, fileName)
```

avishayt (approved):

PR #5279, opened 2023-06-05
MGMT-14830: Enable TechPreviewNoUpgrade when platform is external

adriengentil:
External platform will only be available behind the `TechPreviewNoUpgrade`
feature set. This change enables it automatically when an external
platform is selected (e.g.: oci).

### DIFF
--- a/internal/installcfg/installcfg.go
+++ b/internal/installcfg/installcfg.go
@@ -2,6 +2,7 @@ package installcfg
 
 import (
 	"github.com/go-openapi/strfmt"
+	configv1 "github.com/openshift/api/config/v1"
 	cluster_validations "github.com/openshift/assisted-service/internal/cluster/validations"
 )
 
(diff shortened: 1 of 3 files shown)


gamli75 (approved):

internal/installcfg/installcfg.go
```diff
@@ -162,6 +163,8 @@ type Capabilities struct {
 	AdditionalEnabledCapabilities []ClusterVersionCapability  `yaml:"additionalEnabledCapabilities,omitempty"`
 }
 
+type FeatureSet string
```
adriengentil, line 166: oops, forgot to cleanup this one

gamli75 (approved):

internal/provider/external/installConfig.go
```diff
@@ -15,6 +16,9 @@ func (p externalProvider) AddPlatformToInstallConfig(cfg *installcfg.InstallerCo
 		None: &installcfg.PlatformNone{},
 	}
 
+	// Enable feature set "TechPreviewNoUpgrade" in order to access External platform feature
```
gamli75, line 19: @filanov This code does not exist in the master code, see: https://github.com/openshift/assisted-service/blob/master/internal/provider/external/installConfig.go

gamli75 (commented):
@adriengentil can you please review this PR?

PR #5284, opened 2023-06-07
MGMT-13284: Check for BMH CRD before creating controller

CrystalChun:
[MGMT-13284](https://issues.redhat.com/browse/MGMT-13284)
Explicitly check for the existence of the BMH CRD before allowing the controller to register with the manager. This prevents an unclear fatal error if the controller was created and the BMH CRD is not defined.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @carbonin

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -89,6 +89,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
@@ -534,6 +535,7 @@ func main() {
 
 	go func() {
 		if Options.EnableKubeAPI {
+			failOnError(doesBMHCRDExist(ctrlMgr), "BareMetalHost CRD does not exist in cluster")
 			clientConfig := ctrl.GetConfigOrDie()
 			osClient := osclientset.NewForConfigOrDie(clientConfig)
 			kubeClient := kubernetes.NewForConfigOrDie(clientConfig)
@@ -868,4 +870,15 @@ func getNotificationStream(log *logrus.Logger) *stream.NotificationStream {
 		log.WithError(err).Fatal("kafka writer failed to initialize")
 	}
 	return stream.NewNotificationStream(writer, log, metadata)
+}
+
+func doesBMHCRDExist(mgr manager.Manager) error {
+	gvk, err := apiutil.GVKForObject(&metal3_v1alpha1.BareMetalHost{}, mgr.GetScheme())
+	if err != nil {
+		return err
+	}
+	if _, err = mgr.GetRESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version); err != nil {
+		return err
+	}
+	return nil
 }


filanov:
Please add an example of the error message to the description/commit message

carbonin:
I'm all for a more clear error message, but is this the solution we want or would we rather just not start BMAC when the BMH CRD isn't present?

carbonin:
I suppose that would be a bigger change though.

CrystalChun:
> I'm all for a more clear error message, but is this the solution we want or would we rather just not start BMAC when the BMH CRD isn't present?

Good point! I've modified the PR to reflect this. How does it look? I'm unsure about where I placed the checking function `DoesBMHCRDExist`. 

Another option is for this checking function to just live in `main.go` since it's only really called once at the beginning of the program. WDYT?

internal/controller/controllers/crd_utils.go
```diff
@@ -164,6 +167,18 @@ func (u *CRDUtils) CreateAgentCR(ctx context.Context, log logrus.FieldLogger, ho
 	return nil
 }
 
+// DoesBMHCRDExist returns true if the BareMetalHost CRD is defined in the cluster
+func (u *CRDUtils) DoesBMHCRDExist(mgr manager.Manager) bool {
```
CrystalChun, line 171: Thinking this function doesn't belong here... perhaps just in main since it's called once at the beginning of the program?

carbonin:
>I'm all for a more clear error message, but is this the solution we want or would we rather just not start BMAC when the BMH CRD isn't present?

Sorry for the confusion, I wasn't expecting you to change anything just yet. I was more asking for opinions.

What you you think @eranco74 @avishayt @filanov ?

CrystalChun:
> Sorry for the confusion, I wasn't expecting you to change anything just yet. I was more asking for opinions.
> 
> What you you think @eranco74 @avishayt @filanov ?

Np! For the others, the previous change was to fail the assisted service completely with a more clear error message if the BMH CRD doesn't exist:

```
func CheckBMHCRD() error {
	gvk, err := apiutil.GVKForObject(&bmh_v1alpha1.BareMetalHost{}, mgr.GetScheme())
	if err != nil {
		return err
	}
	if _, err = mgr.GetRESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version); err != nil {
		return err
	}
}
...
        failOnError(CheckBMHCRD())
        failOnError((&controllers.BMACReconciler{
...
```

The new error message would look like:
```
time="2023-06-07T02:52:34Z" level=fatal msg="unable to create controller BMH" func=main.main.func1 file="/assisted-service/cmd/main.go:221" error="no matches for kind \"BareMetalHost\" in version \"metal3.io/v1alpha1\""
```

avishayt:
Can you please clarify the options?  We definitely shouldn't fail everything if BMH isn't defined, and we want a clear error message.  What's the effect if BMAC starts?  Does it spam the log continuously, or make some noise and go away?

carbonin:
Today we fail entirely with an unclear fatal error message if the BMH CRD doesn't exist:

`time="2023-01-12T14:26:03Z" level=fatal msg="failed to run manager" func=main.main.func1 file="/remote-source/assisted-service/app/cmd/main.go:204" error="failed to wait for baremetal-agent-controller caches to sync: timed out waiting for cache to be synced"`

The options are:
1. Improve the error message, but still fail to start the service
2. Don't start BMAC if there is no BMH CRD

I'm leaning toward 2, but that would mean that even if the BMH CRD is created after the service is started the user would need to restart our service to get BMAC running.

We need to decide if that's okay.

internal/controller/controllers/crd_utils.go
filanov: +1

filanov:
What will happen if the user apply BMH? if we need to watch it and then reboot or start bmac then for now i prefer 1.

carbonin:
> even if the BMH CRD is created after the service is started the user would need to restart our service to get BMAC running.

@filanov does that answer the question?

But @avishayt also said "We definitely shouldn't fail everything if BMH isn't defined".
I'm unsure how to do both.

carbonin:
I suppose we could have our operator watch CRDs and set some env based on the existence of the BMH CRD.

filanov:
> > even if the BMH CRD is created after the service is started the user would need to restart our service to get BMAC running.
> 
> @filanov does that answer the question?
> 
> But @avishayt also said "We definitely shouldn't fail everything if BMH isn't defined". I'm unsure how to do both.

It does.
I just suspect that users will not know how to do it or would expect this process to happen automatically. 
I can imagine tickets like `i deployed BMH but nothing is happening and the agent doesn't boot but the bmh is provisioned` or something like that.

CrystalChun:
After speaking with Nick and Michael, the consensus is to have the service fail to start if the BMH CRD is missing with a more clear error message. 

A follow-up issue will be created to discuss a better solution.

cmd/main.go
```diff
@@ -534,6 +535,7 @@ func main() {
 
 	go func() {
 		if Options.EnableKubeAPI {
+			failOnError(doesBMHCRDExist(ctrlMgr), "BMH CRD does not exist in cluster")
```
carbonin, line 538: Maybe put the whole name here (`BareMetalHost`). Otherwise looks good

CrystalChun: Good idea, thank you! Modified to reflect full name of CRD.

eranco74 (commented):
/lgtm
This is an improvement compared to the current state (the service fail to start and we need to guess why).

CrystalChun:
Thanks all for the reviews!

Follow-up issue to discuss a better solution
https://issues.redhat.com/browse/MGMT-15111

PR #5295, opened 2023-06-15
OCPBUGS-13738 enforce additional ntp sources added into chrony

filanov:
Chrony configuration applied into machine config manifest based on the reply from the agent, meaning that when user add additional ntp source first the service will ask the agent to add it then return the reply to the service it will be stored in the DB and used to create the machine config chrony configuration

Because NTP is a soft validation, in ZTP flow there are cases that the installation will start before the agent return a reply to the service, it will mean that those sources will not be applied, to avoid that when building the manifests in addition to taking the agent reply the additional ntp sources will be added directly from the DB

The downside of this solution is that if the source is incorrect it will appear in /etc/chrony.conf but not in `chronyc sources` and it will not be resolved as the regular reply from the agent

It will not harm the cluster installation or affect the cluster after the installation if the source is incorrect

Example i added `harta,barta` in additional ntp sources
They didn't return from the agent so without this code they will not be added to the installation event if there is no race 
After installation 
```
cat /etc/chrony.conf 
...
server harta iburst
server barta iburst

chronyc sources
MS Name/IP address         Stratum Poll Reach LastRx Last sample               
===============================================================================
^? ntp.wdc2.us.leaseweb.net      0   9     0     -     +0ns[   +0ns] +/-    0ns
^? 152.70.159.102                0   9     0     -     +0ns[   +0ns] +/-    0ns
^? time.cloudflare.com           0   9     0     -     +0ns[   +0ns] +/-    0ns
^? haka.ruselabs.com             0   9     0     -     +0ns[   +0ns] +/-    0ns
^? ip229.ip-51-81-226.us         0   9     0     -     +0ns[   +0ns] +/-    0ns
^? 108.175.15.67                 0   9     0     -     +0ns[   +0ns] +/-    0ns
^? connected.by.freedominte>     0   9     0     -     +0ns[   +0ns] +/-    0ns
^? 168.61.215.74                 0   9     0     -     +0ns[   +0ns] +/-    0ns
^? tick.srs1.ntfo.org            0   9     0     -     +0ns[   +0ns] +/-    0ns
^? pacific.latt.net              0   9     0     -     +0ns[   +0ns] +/-    0ns
^? shaka.ruselabs.com            0   9     0     -     +0ns[   +0ns] +/-    0ns
^? 198.30.92.2                   0   9     0     -     +0ns[   +0ns] +/-    0ns
^? pi3.rellim.com                0   9     0     -     +0ns[   +0ns] +/-    0ns
^? srcf-ntp.stanford.edu         0   9     0     -     +0ns[   +0ns] +/-    0ns
^? time.cloudflare.com           0   9     0     -     +0ns[   +0ns] +/-    0ns
^* ntp1.ntp-001.prod.iad2.d>     2   8   377   121   +162us[ +215us] +/-   27ms
```

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -382,7 +382,7 @@ func main() {
 	hostApi := host.NewManager(log.WithField("pkg", "host-state"), db, notificationStream, eventsHandler, hwValidator,
 		instructionApi, &Options.HWValidatorConfig, metricsManager, &Options.HostConfig, lead, operatorsManager, providerRegistry, Options.EnableKubeAPI, objectHandler, versionHandler)
 	dnsApi := dns.NewDNSHandler(Options.BMConfig.BaseDNSDomains, log)
-	manifestsGenerator := network.NewManifestsGenerator(manifestsApi, Options.ManifestsGeneratorConfig)
+	manifestsGenerator := network.NewManifestsGenerator(manifestsApi, Options.ManifestsGeneratorConfig, db)
 	clusterApi := cluster.NewManager(Options.ClusterConfig, log.WithField("pkg", "cluster-state"), db,
 		notificationStream, eventsHandler, uploadClient, hostApi, metricsManager, manifestsGenerator, lead, operatorsManager, ocmClient, objectHandler, dnsApi, authHandler)
 	infraEnvApi := infraenv.NewManager(log.WithField("pkg", "host-state"), db, objectHandler)
--- a/internal/network/manifests_generator.go
+++ b/internal/network/manifests_generator.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"strings"
 	"text/template"
 
 	"github.com/go-openapi/swag"
@@ -19,6 +20,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
+	"gorm.io/gorm"
 )
 
 //go:generate mockgen -source=manifests_generator.go -package=network -destination=mock_manifests_generator.go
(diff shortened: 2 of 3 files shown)


filanov:
Not 100% sure about this solution because the user will be able to add invalid ntp sources, @carbonin @avishayt what do you think?

carbonin:
> Not 100% sure about this solution because the user will be able to add invalid ntp sources

If there was no race and we did whatever we do to add the sources to the host during discovery, would the user have seen an error?

Or put another way, are we doing anything to attempt to check the validity of these sources currently?

filanov:
> > Not 100% sure about this solution because the user will be able to add invalid ntp sources
> 
> If there was no race and we did whatever we do to add the sources to the host during discovery, would the user have seen an error?
> 
> Or put another way, are we doing anything to attempt to check the validity of these sources currently?

so what we are doing right now is what i described in the commit message 

> Chrony configuration applied into machine config manifest based on the reply from the agent, meaning that when user add additional ntp source first the service will ask the agent to add it then return the reply to the service it will be stored in the DB and used to create the machine config chrony configuration

Meaning that invalid sources will not be added, but on the other had the user requested something so it will make sense to add them even if right now they are not available.

in addition we had a lot of confusion from  the users related to the resolution that the agent is doing, we are being asked `i added X` and i don't see it in the chrony and with this change they will.

carbonin:
I'm fine with this.

I think it's less confusing to add the ntp sources the user asks for even if they may be incorrect as long as it doesn't negatively impact the cluster.

They can always remove or change the sources using the machineconfig once the cluster is installed.

avishayt:
If the user asked for an NTP source, we should add it.  It might currently be offline or unreachable, but later on, it will be (for example, if the cluster is relocated).

It looks like this PR doesn't fix the actual race though - is that in a different PR?

filanov:
> If the user asked for an NTP source, we should add it. It might currently be offline or unreachable, but later on, it will be (for example, if the cluster is relocated).
> 
> It looks like this PR doesn't fix the actual race though - is that in a different PR?

The way to resolve the race is to add additional validation that will block the installation, but it will conflict with another validation that is not required so it will create a strange UX for the user, in ACM we don't have defaults so with another validation we will require the user to add additional NTP sources.

With the proposed solution we will copy exactly what the user requested.

PR #5296, opened 2023-06-15
MGMT-14620 Only emit succesful host creation message for pre install hosts

paul-maidment:
When a host reboots into the discovery image  it will attempt to re-register with assisted-service.

If the host has already been registered, this may lead to the confusing event "Host registration succeeded" The general consensus is that it's only desirable to emit this event in cluster statuses that are prior to the start of an installation.

This PR ensures that the event is only emitted in states where installation has not yet started (or in the adding hosts state)

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
- [] No tests needed

## Checklist

## Reviewers Checklist


nmagnezi (commented):
A question: what would happen in a case where a host that booted from discovery reboots for whatever reason, and goes back to boot from discovery? will we still emit this event?

internal/bminventory/inventory.go
```diff
@@ -5359,8 +5359,17 @@ func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installe
 
 	b.customizeHost(c, host)
 
-	eventgen.SendHostRegistrationSucceededEvent(ctx, b.eventsHandler, *params.NewHostParams.HostID,
-		params.InfraEnvID, host.ClusterID, hostutil.GetHostnameForMsg(host))
+	// Only send SendHostRegistrationSucceededEvent if cluster in a pre-installation state.
+	postInstallStages := []string{
```
filanov, line 5363: did you consider to move this event into https://github.com/openshift/assisted-service/blob/f218f6c0bcb729f5bca66608f357cd7401b2363a/internal/host/transition.go#L85 
then you will be able to avoid looking at specific states

internal/bminventory/inventory.go
```diff
@@ -5359,8 +5359,17 @@ func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installe
...
+		models.ClusterStatusPreparingForInstallation,
+		models.ClusterStatusInstallingPendingUserAction,
+		models.ClusterStatusInstalling,
+		models.ClusterStatusInstalled,
+	}
+	if cluster != nil && !swag.ContainsStrings(postInstallStages, *cluster.Status) {
```
filanov, line 5369: if cluster is nil you still want to get this event.

filanov:
> A question: what would happen in a case where a host that booted from discovery reboots for whatever reason, and goes back to boot from discovery? will we still emit this event?

The goal is not to send the event during installation, for all the other cases this event should be sent

internal/bminventory/inventory.go
paul-maidment: I considered moving it but I think a lot of our tests are structured in a way that would require a larger rewrite in order to prove the event emission. I felt it was more concise to leave things as they are.

filanov (approved):

PR #5312, opened 2023-06-21
MGMT-15034: Fix patch of infrastructure CR with external platform

adriengentil:
Patch directly the Infrastructure CR from PostCreateManifestsHook
function which makes it easier to test.

This part of the code is a workaround until the openshift installer
supports external platform.

### DIFF
--- a/internal/provider/external/ignition.go
+++ b/internal/provider/external/ignition.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 
 	"github.com/openshift/assisted-service/internal/common"
+	"github.com/pkg/errors"
 )
 
 // Workaround as openshift installer does not support the external platform.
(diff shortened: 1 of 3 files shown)


internal/provider/external/ignition.go
```diff
@@ -6,48 +6,61 @@ import (
...
-	clusterInfraPatchPath := filepath.Join(workDir, "manifests", "cluster-infrastructure-02-config.yml.patch_01_configure_external_platform")
-	err := os.WriteFile(clusterInfraPatchPath, []byte(clusterInfraPatch), 0600)
+	p.Log.Infof("Patching Infrastructure CR with external platform: platformName=%s", platformName)
+
+	infraManifest := filepath.Join(workDir, "manifests", "cluster-infrastructure-02-config.yml")
+	data, err := os.ReadFile(infraManifest)
+	if err != nil {
+		return errors.Wrapf(err, "failed to read Infrastructure Manifest \"%s\"", infraManifest)
+	}
+	p.Log.Debugf("read the infrastructure manifest at %s", infraManifest)
```
gamli75, line 51: if this log is important and we will remove it anyway after we will have the OCP version - let's change it to info

internal/provider/external/ignition.go
```diff
@@ -6,48 +6,61 @@ import (
...
+
+	data, err = common.ApplyYamlPatch(data, []byte(clusterInfraPatch))
+	if err != nil {
+		return errors.Wrapf(err, "failed to patch Infrastructure Manifest \"%s\"", infraManifest)
+	}
+	p.Log.Debugf("applied the yaml patch to the infrastructure manifest at %s: \n %s", infraManifest, string(data[:]))
```
gamli75, line 57: same here

gamli75:
few remarks:
1. please add a task in your epic to remove this code
2. do we want this patch only in staging or also in production?

adriengentil:
> few remarks:
> 
>     1. please add a task in your epic to remove this code

Created this one: https://issues.redhat.com/browse/MGMT-15047

>     2. do we want this patch only in staging or also in production?

We will need this patch in production once an OCP version with external platform is available

internal/provider/external/ignition.go
adriengentil: done!

internal/provider/external/ignition.go
adriengentil: done!

gamli75 (approved):

PR #5319, opened 2023-06-26
MGMT-14852: Allow AutomatedCleaningMode to be set by user

CrystalChun:
https://issues.redhat.com/browse/MGMT-14852
Epic: https://issues.redhat.com/browse/MGMT-13017
Removes setting automated cleaning mode to disabled by default, which lets the user set the field.
Part of the larger epic to allow wiping disks before installing by using Ironic's automated cleaning feature.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -868,16 +868,11 @@ func (r *BMACReconciler) reconcileBMH(ctx context.Context, log logrus.FieldLogge
 	}
 
 	dirty := false
-	// in case of converged flow set the custom deploy and cleaning mode instead of the annotations
+	// in case of converged flow set the custom deploy instead of the annotations
 	if r.ConvergedFlowEnabled {
 		if bmh.Spec.CustomDeploy == nil || bmh.Spec.CustomDeploy.Method != ASSISTED_DEPLOY_METHOD {
 			log.Infof("Updating BMH CustomDeploy to %s", ASSISTED_DEPLOY_METHOD)
 			bmh.Spec.CustomDeploy = &bmh_v1alpha1.CustomDeploy{Method: ASSISTED_DEPLOY_METHOD}
-			dirty = true
-		}
-		if bmh.Spec.AutomatedCleaningMode != bmh_v1alpha1.CleaningModeDisabled {
-			log.Infof("Updating BMH AutomatedCleaningMode to %s", bmh_v1alpha1.CleaningModeDisabled)
-			bmh.Spec.AutomatedCleaningMode = bmh_v1alpha1.CleaningModeDisabled
 			dirty = true
 		}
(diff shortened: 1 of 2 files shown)


eranco74 (commented):
1. Note that in handleBMHFinalizer BMAC sets AutomatedCleaningMode to metadata.
2. This PR changes the default behavior in case the user didn't specify `AutomatedCleaningMode` in the BMH CR (the default is `metadata`) hence it should come with docs update.

osherdp:
/uncc 
as I'm not that familiar with this part of the system

CrystalChun:
> 1. Note that in handleBMHFinalizer BMAC sets AutomatedCleaningMode to metadata.

Thanks for pointing that out! I think that's intended behavior for deprovisioning when the converged flow is enabled, right @carbonin? 

Maybe I should've said the scope of this is to allow it to be set by the user on provisioning.

> 2. This PR changes the default behavior in case the user didn't specify `AutomatedCleaningMode` in the BMH CR (the default is `metadata`) hence it should come with docs update.

Agreed! I have a separate [issue](https://issues.redhat.com/browse/MGMT-14851) for that, but do you think the docs update should be part of this PR?

carbonin:
>I think that's intended behavior for deprovisioning when the converged flow is enabled, right @carbonin?

Yes, but since that's opt-in I imagine it would be fine

avishayt:
> 1. Note that in handleBMHFinalizer BMAC sets AutomatedCleaningMode to metadata.
> 2. This PR changes the default behavior in case the user didn't specify `AutomatedCleaningMode` in the BMH CR (the default is `metadata`) hence it should come with docs update.

Should we leave the default as is?

carbonin:
>Should we leave the default as is?

We actually can't.

IIRC the field is set by some kind of webhook so we can't tell the difference between when a user explicitly selects `metadata` vs `metadata` as the default.

Given this behavior the only way to preserve the default as `disabled` would be to create another API (in an annotation, probably) for the _actual_ value the user wants and I'd rather change the default to what it's always been for BMO than add a duplicate field.

CrystalChun:
Thanks @carbonin for explaining! 

We've spoken and aligned with both the ACM team and the Far Edge team (@imiller0's team) and came to the consensus that Ian's team will:
1. Keep the default set to `disabled` in `SiteConfig` if the user doesn't explicitly set anything
2. Expose this field (`automatedCleaningMode`) in `SiteConfig` so that the user may take advantage of us allowing them to set this field

Tracker for their progress: https://issues.redhat.com/browse/CNF-8620

carbonin (approved):

PR #5324, opened 2023-06-28
MGMT-15114: garbage collect orphaned events

rccrdpccl:
Due to performance testing, in stage we are filling up DB disk with "failed to registered cluster" events. As those events have not an associated cluster, garbage collector will have no effect: as it currently works we garbage collect clusters, host and infraenv and cascade from there.

This PR will garbage collect orphaned events (with associated cluster_id).

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/openshift/template.yaml
+++ b/openshift/template.yaml
@@ -198,6 +198,9 @@ parameters:
 - name: INSTALLER_CACHE_CAPACITY
   value: "6442450944"
   required: false
+- name: ORPHANED_EVENTS_MAX_AGE
+  required: false
+  value: "72h"
 apiVersion: v1
 kind: Template
 metadata:
@@ -455,6 +458,8 @@ objects:
                 value: ${ENABLE_DATA_COLLECTION}
               - name: INSTALLER_CACHE_CAPACITY
                 value: ${INSTALLER_CACHE_CAPACITY}
+              - name: ORPHANED_EVENTS_MAX_AGE
+                value: ${ORPHANED_EVENTS_MAX_AGE}
             volumeMounts:
               - name: route53-creds
                 mountPath: "/etc/.aws"
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -417,7 +417,7 @@ func main() {
 
 	if Options.EnableDeregisterInactiveGC || Options.EnableDeletedUnregisteredGC {
 		gc := garbagecollector.NewGarbageCollectors(Options.GCConfig, db, log.WithField("pkg", "garbage_collector"),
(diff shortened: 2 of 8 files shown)


openshift/template.yaml
```diff
@@ -198,6 +198,9 @@ parameters:
 - name: INSTALLER_CACHE_CAPACITY
   value: "6442450944"
   required: false
+- name: ORPHANED_EVENTS_MAX_AGE
+  required: false
+  value: "72h"
```
osherdp, line 203: did you mean 720h?

rccrdpccl: I actually meant to override it to 72h, but now that you mention it it would be good to have consistency at least with the starting value.
I believe this is pretty safe to delete after 72h as only records that did not cascade delete would be here, and l would set it by default in the code as well. WDYT?

osherdp: sgtm

rccrdpccl: @gamli75 WDYT? given what we found out, is 3 days enough?

avishayt: Do these events have any use?

rccrdpccl: They are pretty much unusable from the application perspective, as they will never be able to be retrieved via API until a cluster with the same cluster ID exists.
They are also collected by event collection system AKA "kibana"
AFAIU their only use would be diagnostic

avishayt: The question is if we should just not send the event.  WDYT?

rccrdpccl: I think the event is giving good insight for people investigating the cluster, and would give good feedback to the user. Example messages:
`Failed to register cluster. Error: OpenShiftSDN network type is not allowed in single node mode`

`Failed to register cluster. Error: Invalid format for Tags: tag-3. Tags should be a comma-separated list (e.g. tag1,tag2,tag3). Each tag can consist of the following characters: Alphanumeric (aA-zZ, 0-9), underscore (_) and white-spaces.`

`Failed to register cluster. Error: Can't set oci platform on s390x architecture`

If anything, I would try to show it to the user. For now this is not possible because the cluster is not registered, and we can't retrieve events from non-registered clusters (due to security measure to avoid cross-tenant access) - chicken and egg problem.

rccrdpccl: Fair enough, I agree we can probably remove the event generation and clean up in a different way.

PR #5331, opened 2023-06-29
RHTAPBUGS-318: Fetch tag in RHTAP build

gbenhaim:
The build process requires tags to be presented (it runs git describe). Configure the pipeline to fetch the recent 100 commits and all the tags.


gbenhaim:
@rccrdpccl can you please add "/ok-to-test" so RHTAP will trigger the pipeline?

PR #5333, opened 2023-07-03
RHTAPBUGS-318: Fetch tag in RHTAP build

gbenhaim:
The build process requires tags to be presented (it runs git describe). Configure the pipeline to fetch the recent 100 commits and all the tags.

### DIFF
--- a/.tekton/assisted-service-q2vh-pull-request.yaml
+++ b/.tekton/assisted-service-q2vh-pull-request.yaml
@@ -143,6 +143,8 @@ spec:
         value: $(params.git-url)
       - name: revision
         value: $(params.revision)
+      - name: depth
+        value: "100"
       runAfter:
       - init
       taskRef:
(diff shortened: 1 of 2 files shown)


gbenhaim:
@rccrdpccl there is a bug in Tekton which blocking the pipeline from starting even if the author of the PR (or a user that issue the ok-to-test comment) is a public member of the Github org or in the [OWNERS_ALIASES](https://github.com/openshift/assisted-service/blob/master/OWNERS_ALIASES) file. I think that the only way right now is to have someone which is a collaborator to add the "ok-to-test" comment.

rccrdpccl:
@adriengentil could you please try to remove the label and add it again? (ok-to-test)

gbenhaim:
@adriengentil can you please add "ok-to-test" comment?

gbenhaim:
thanks @osherdp, but because of the bug I mentioned above we need someone which is an owner or a collaborator to add the "ok-to-test" comment.

osherdp:
> thanks @osherdp, but because of the bug I mentioned above we need someone which is an owner or a collaborator to add the "ok-to-test" comment.

why should it care who adds the label? can you provide a link for this issue?

gbenhaim:
thanks @rccrdpccl

PR #5338, opened 2023-07-05
MGMT-15073:Cluster reset - only system generated manifests to be deleted

paul-maidment:
Presently when a cluster is reset, we perform all sorts of awkward workarounds to exclude files from deletion, when in reality all we should do is delete only the system generated manifests, leaving all other files and logs intact.

Moreover, when performing a permenant deletion, ALL files should be deleted.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -25,6 +25,7 @@ import (
 	eventsapi "github.com/openshift/assisted-service/internal/events/api"
 	"github.com/openshift/assisted-service/internal/host"
 	"github.com/openshift/assisted-service/internal/host/hostutil"
+	manifestsapi "github.com/openshift/assisted-service/internal/manifests/api"
 	"github.com/openshift/assisted-service/internal/metrics"
 	"github.com/openshift/assisted-service/internal/network"
 	"github.com/openshift/assisted-service/internal/operators"
@@ -38,6 +39,7 @@ import (
 	"github.com/openshift/assisted-service/pkg/ocm"
 	"github.com/openshift/assisted-service/pkg/requestid"
 	"github.com/openshift/assisted-service/pkg/s3wrapper"
+	"github.com/openshift/assisted-service/restapi/operations/manifests"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
@@ -117,7 +119,7 @@ type API interface {
 	SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID, db *gorm.DB) error
 	DetectAndStoreCollidingIPsForCluster(clusterID strfmt.UUID, db *gorm.DB) error
 	DeleteClusterLogs(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error
-	DeleteClusterFiles(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error
+	ResetClusterFiles(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error
 	UpdateLogsProgress(ctx context.Context, c *common.Cluster, progress string) error
 	GetClusterByKubeKey(key types.NamespacedName) (*common.Cluster, error)
 	UpdateAmsSubscriptionID(ctx context.Context, clusterID, amsSubscriptionID strfmt.UUID) *common.ApiErrorResponse
@@ -167,12 +169,13 @@ type Manager struct {
 	monitorQueryGenerator *common.MonitorClusterQueryGenerator
 	authHandler           auth.Authenticator
 	uploadClient          uploader.Client
+	manifestApi           manifestsapi.ManifestsAPI
 }
 
 func NewManager(cfg Config, log logrus.FieldLogger, db *gorm.DB, stream stream.Notifier, eventsHandler eventsapi.Handler,
 	uploadClient uploader.Client, hostAPI host.API, metricApi metrics.API, manifestsGeneratorAPI network.ManifestsGeneratorAPI,
 	leaderElector leader.Leader, operatorsApi operators.API, ocmClient *ocm.Client, objectHandler s3wrapper.API,
-	dnsApi dns.DNSApi, authHandler auth.Authenticator) *Manager {
+	dnsApi dns.DNSApi, authHandler auth.Authenticator, manifestApi manifestsapi.ManifestsAPI) *Manager {
 	th := &transitionHandler{
 		log:                 log,
 		db:                  db,
@@ -202,6 +205,7 @@ func NewManager(cfg Config, log logrus.FieldLogger, db *gorm.DB, stream stream.N
 		dnsApi:                dnsApi,
 		authHandler:           authHandler,
 		uploadClient:          uploadClient,
+		manifestApi:           manifestApi,
 	}
 }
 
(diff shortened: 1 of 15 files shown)


internal/cluster/cluster.go
```diff
@@ -1399,32 +1403,106 @@ func (m *Manager) SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID,
...
-		return common.NewApiError(
+		return nil, common.NewApiError(
 			http.StatusInternalServerError,
 			errors.Errorf(msg))
 	}
+	return files, err
+}
 
-	var failedToDelete []string
+func (m *Manager) ResetClusterFiles(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error {
```
carbonin, line 1430: This function is only dealing with manifests, right? Maybe we should name it that way.

internal/cluster/cluster.go
```diff
@@ -1399,32 +1403,106 @@ func (m *Manager) SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID,
...
+			isSystemGeneratedManifest := true
+			for _, manifest := range userManifests {
+				if manifest.Folder == currentFolder && manifest.FileName == currentFile {
+					isSystemGeneratedManifest = false
+					break
+				}
+			}
+			if isSystemGeneratedManifest {
+				log.Debugf("Deleting cluster %s S3 file: %s", c.ID.String(), file)
+				_, err := objectHandler.DeleteObject(ctx, file)
```
carbonin, line 1454: Where do we handle deleting the metadata? Or is that only created for the user manifests? I forget which direction we ended up going on that one.

internal/cluster/cluster.go
```diff
@@ -1399,32 +1403,106 @@ func (m *Manager) SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID,
...
 
-	var failedToDelete []string
+func (m *Manager) ResetClusterFiles(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error {
+	log := logutil.FromContext(ctx, m.log)
+	files, err := m.getClusterFilesForFolder(ctx, c, objectHandler, "")
+	if err != nil {
+		return err
+	}
+	// ListClusterManifestsInternal only returns user generated manifests so we can use this to determine whether or not a manifest should be deleted on reset.
+	userManifests, err := m.manifestApi.ListClusterManifestsInternal(ctx, manifests.V2ListClusterManifestsParams{ClusterID: *c.Cluster.ID})
```
carbonin, line 1437: It seems like it would be easier to create a manifestsApi function to return the system manifests then just delete them directly.

This strategy of getting the user manifests then iterating through all the files eliminating those is strange. Is there any other reason you're doing it this way?

internal/cluster/cluster.go
```diff
@@ -1399,32 +1403,106 @@ func (m *Manager) SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID,
 	return m.setConnectivityMajorityGroupsForClusterInternal(cluster, db)
 }
 
-func (m *Manager) deleteClusterFiles(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API, folder string) error {
-	log := logutil.FromContext(ctx, m.log)
+func (m *Manager) extractManifestFolderAndFilename(ctx context.Context, path string) (string, string, bool) {
```
carbonin, line 1406: I think making this a function is actually a bit more confusing than just embedding this in the only place it's used.

internal/cluster/cluster.go
```diff
@@ -1399,32 +1403,106 @@ func (m *Manager) SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID,
...
+	if err != nil {
+		return err
+	}
 	for _, file := range files {
-		//skip log and manifests deletion when deleting cluster files
-		if folder == "" && (strings.Contains(file, "logs") || strings.Contains(file, constants.ManifestFolder) || strings.Contains(file, constants.ManifestMetadataFolder)) {
-			continue
+		// Path is in the form <cluster-id>/manifests/<folder-name>/<file-name>
+		currentFolder, currentFile, isManifest := m.extractManifestFolderAndFilename(ctx, file)
+		if isManifest {
```
carbonin, line 1444: Was there previously anything else in the `""` folder other than manifests? It seems like we would not be deleting that now, but would have been previously.

internal/cluster/cluster.go
```diff
@@ -1399,32 +1403,106 @@ func (m *Manager) SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID,
...
+	}
+	return nil
+}
+
+func (m *Manager) DeleteClusterLogs(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error {
+	return m.deleteClusterFilesInFolders(ctx, c, objectHandler, []string{"logs"})
+}
+
+func (m *Manager) deleteAllClusterFiles(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error {
+	folders := []string{"", "logs", constants.ManifestFolder, constants.ManifestMetadataFolder}
```
carbonin, line 1469: Does `""` not contain all the other entries? If we know we're going to delete everything do we really need all the other specific folders?

internal/cluster/cluster_test.go
```diff
@@ -3172,6 +3179,9 @@ var _ = Describe("Permanently delete clusters", func() {
 		Expect(db.First(&common.Cluster{}, "id = ?", c2.ID).RowsAffected).Should(Equal(int64(0)))
 		Expect(db.Unscoped().First(&common.Cluster{}, "id = ?", c2.ID).RowsAffected).Should(Equal(int64(1)))
 
+		// ListClusterManifestsInternal(ctx, manifests.V2ListClusterManifestsParams{ClusterID: *c.Cluster.ID})
```
carbonin, line 3182: You can remove this comment.

carbonin (changes requested):
This implementation seems more complicated than the previous one to me.
We have a few instances of nested loops that I think we can remove and the logic to determine if a manifest should be deleted on reset seems inverted to me.

internal/cluster/cluster.go
paul-maidment: Metadata is only created for user manifests

internal/cluster/cluster.go
paul-maidment: What about an API function to delete system manifests directly via API, maybe that would work better?

internal/cluster/cluster.go
paul-maidment: Old behaviour did this too, did not quite understand why, I am just replicating here but was curious about this too.

internal/cluster/cluster_test.go
```diff
@@ -3172,6 +3179,9 @@ var _ = Describe("Permanently delete clusters", func() {
...
+		manifestsAPI.EXPECT().ListClusterManifestsInternal(gomock.Any(), manifests.V2ListClusterManifestsParams{ClusterID: *c1.Cluster.ID}).AnyTimes()
+		manifestsAPI.EXPECT().ListClusterManifestsInternal(gomock.Any(), manifests.V2ListClusterManifestsParams{ClusterID: *c2.Cluster.ID}).AnyTimes()
 		mockS3Api.EXPECT().DeleteObject(gomock.Any(), c1.ID.String()).Return(false, nil).Times(1)
```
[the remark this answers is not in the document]
paul-maidment: Because these functions are listing user manifests since the change.
Each file from the directory is then checked against the list of known user manifests as follows

If the file is a manifest and is not a user manifest then it is a system manifest.
In the case of legacy clusters (ones from before the "manifest metadata" change) the files are always considered to be "user manifests"

internal/manifests/manifests_test.go
```diff
@@ -791,6 +792,30 @@ spec:
 			Expect(responsePayload.Payload.Folder).To(Equal(defaultFolder))
 		})
 	})
+
+	Context("Manifest path parsing functions", func() {
```
carbonin, line 796: Use `Describe`  with the function you're testing and `It` for individual behaviors you're testing. So something like

```go
Describe("ParsePath", func() {
  It("parses a path into folder and filename", func() {
    ...
  })
  It("fails when provided a non-manifest path", func() {
    ..
  })
})
Describe("IsManifest", func() {
  ...
})
```

This way a test failure will show exactly which function is failing and what about it isn't working.

internal/manifests/manifests.go
```diff
@@ -327,7 +338,8 @@ func GetManifestObjectName(clusterID strfmt.UUID, fileName string) string {
 
 // GetManifestObjectName returns the manifest object name as stored in S3
 func GetManifestMetadataObjectName(clusterID strfmt.UUID, fileName string, manifestSource string) string {
-	return filepath.Join(string(clusterID), constants.ManifestMetadataFolder, fileName, manifestSource)
+	objectName := filepath.Join(string(clusterID), constants.ManifestMetadataFolder, fileName, manifestSource)
+	return objectName
```
carbonin, line 342: Why was this changed?

paul-maidment: Mainly debugging, will revert it when I commit.

internal/cluster/cluster.go
```diff
@@ -1399,32 +1403,47 @@ func (m *Manager) SetConnectivityMajorityGroupsForCluster(clusterID strfmt.UUID,
...
 		m.log.WithError(err).Errorf(msg)
-		return common.NewApiError(
+		return nil, common.NewApiError(
 			http.StatusInternalServerError,
 			errors.Errorf(msg))
 	}
+	return files, err
+}
 
+func (m *Manager) ResetClusterFiles(ctx context.Context, c *common.Cluster, objectHandler s3wrapper.API) error {
```
carbonin, line 1419: I think we should add unit tests for this.

internal/cluster/cluster_test.go
```diff
@@ -3985,3 +3993,73 @@ var _ = Describe("UploadEvents", func() {
...
+		mockObjectHandler = s3wrapper.NewMockAPI(ctrl)
+		capi = NewManager(getDefaultConfig(), common.GetTestLog(), db, commontesting.GetDummyNotificationStream(ctrl), nil, nil, nil, nil, nil, dummy, mockOperators, nil, mockObjectHandler, nil, nil, mockManifestsApi)
+	})
+
+	AfterEach(func() {
+		common.DeleteTestDB(db, dbName)
+		ctrl.Finish()
+	})
+
+	It("Should clear only user generated manifests when performing a reset", func() {
```
carbonin, line 4042: Isn't this test name incorrect?
It should clear only system manifests when resetting, right?

paul-maidment: Yes, will change this

carbonin (approved):

PR #5334, opened 2023-07-03
MGMT-15150: Use same installer binary for all platform types

zaneb:
There's nothing special about platform:none that requires it to use a different installer binary. This was originally done to avoid a problem with the openshift-baremetal-install binary not being available for non-x86 targets, but this was resolved by 1d025b8108 ([MGMT-9206](https://issues.redhat.com//browse/MGMT-9206)). We no longer need all this complexity.

This partially reverts commit f9b2f3d3246f91f7e80eab0b85eb9c34742eaa46.

## List all the issues related to this PR

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #5347, opened 2023-07-09
[release-4.12] MGMT-15150: Use same installer binary for all platform types

zaneb:
There's nothing special about platform:none that requires it to use a different installer binary. This was originally done (in f9b2f3d324) to avoid a problem with the openshift-baremetal-install binary not being available for non-x86 targets, but this was resolved by 1d025b8108 ([MGMT-9206](https://issues.redhat.com//browse/MGMT-9206)).

(Backported from #5346)

### DIFF
--- a/internal/oc/release.go
+++ b/internal/oc/release.go
@@ -336,15 +336,7 @@ func (r *release) Extract(log logrus.FieldLogger, releaseImage string, releaseIm
 // extractFromRelease returns the path to an openshift-baremetal-install binary extracted from
 // the referenced release image.
 func (r *release) extractFromRelease(log logrus.FieldLogger, releaseImage, cacheDir, pullSecret string, insecure bool, platformType models.PlatformType, icspFile string) (string, error) {
-	// Using platform type as an indication for which openshift install binary to use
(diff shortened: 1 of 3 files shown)


internal/oc/release.go
```diff
@@ -336,15 +336,7 @@ func (r *release) Extract(log logrus.FieldLogger, releaseImage string, releaseIm
 // extractFromRelease returns the path to an openshift-baremetal-install binary extracted from
 // the referenced release image.
 func (r *release) extractFromRelease(log logrus.FieldLogger, releaseImage, cacheDir, pullSecret string, insecure bool, platformType models.PlatformType, icspFile string) (string, error) {
```
rccrdpccl, line 338: Looks like `platformType` is not used any longer in this method

zaneb: That's correct, but I don't see any point backporting a [huge](https://github.com/openshift/assisted-service/pull/5334/commits/e73b9c5a68037f217110160d8912d54e583b631b) change to get rid of it all the way to 4.12.

rccrdpccl (approved):

PR #5354, opened 2023-07-12
MGMT-14656: Sort disks by HCTL after the other options

avishayt:
Before this change, when choosing a default installation disk, we sorted the disks by drive type and size and then selected the first one.  Disks of the same type and size were implicitly ordered by name (e.g., "sda") because that is the order in the host inventory.

In some cases, users expected us to select the "first" disk, and this is in fact what we did.  However, in more recent versions of RHEL, the disk names are less "stable" and we ended up selecting a disk that wasn't configured to boot (mainly tests using libvirt).

The solution implemented here is to sort by HCTL after disk type and size instead of relying on the implicit alphabetic sort.  HCTL represents the physical address of the disk in a server.

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Before the fix:
We have a cluster definition with five hosts.  The Assisted Installer recommends /dev/sda as the installation disk for all five hosts.  In four hosts, /dev/sda has HCTL 0:0:0:0, but in one host, /dev/sda has HCTL 0:0:0:3.  After beginning the installation, this host's status becomes "pending user action" because it tried to boot from the wrong disk.

After the fix:
A new cluster definition with five different hosts.  This time, in four hosts, in hosts where /dev/sda has HCTL 0:0:0:0, the  Assisted Installer still chooses /dev/sda as the installation disk.  However, in the one where /dev/sda has a different HCTL (for example, 0:0:0:3 or 0:0:0:4), Assisted Installer chose another disk (e.g., /dev/sdb or /dev/sdc) that has HCTL 0:0:0:0.  All hosts rebooted successfully.

## Checklist

## Reviewers Checklist


internal/hardware/validator.go
```diff
@@ -175,19 +175,27 @@ func (v *validator) ListEligibleDisks(inventory *models.Inventory) []*models.Dis
...
+		if v := strings.Compare(string(eligibleDisks[i].DriveType), string(eligibleDisks[j].DriveType)); v != 0 {
 			return v < 0
 		}
+
+		// 3. Sort according to disk size
+		if eligibleDisks[i].SizeBytes != eligibleDisks[j].SizeBytes {
+			return eligibleDisks[i].SizeBytes < eligibleDisks[j].SizeBytes
+		}
+
+		// 4. Sort according to HCTL (physical order)
```
omertuc, line 197: It would be nice if this comment went into detail about why we're doing this, in case we want make changes to this in the future

avishayt: Done, and added explanations to the other steps as well.

PR #5357, opened 2023-07-14
MGMT-11456: kube-api should allow user to specify infraenv additional trust bundle

paul-maidment:
Users that install OCP on servers that communicate through transparent proxies must trust the proxy's CA for the communication to work.

The only way users can currently do that is by using both infraenv ignition overrides and install-config overrides. These are generic messy APIs that are very error prone. We should give users a more formal, simpler API to achieve both at the same time.

The epic surrounding this PR [MGMT-11520](https://issues.redhat.com//browse/MGMT-11520) aims to resolve this by allowing the user to specify an "Additional Trust Bundle", in essence the CA for the proxy.

This work has already been completed for the REST API in https://issues.redhat.com/browse/MGMT-11455

The aim of this PR is to ensure that the same changes are effective in KubeAPI CRD's

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [] Manual (Elaborate on how it was tested)
- [] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4940,6 +4940,15 @@ func (b *bareMetalInventory) UpdateInfraEnvInternal(ctx context.Context, params 
 		params.InfraEnvUpdateParams.PullSecret = pullSecretBackup
 	}
 
+	// The OpenShift installer validation code for the additional trust bundle
+	// is buggy and doesn't react well to additional newlines at the end of the
+	// certs. We need to strip them out to not bother assisted users with this
+	// quirk.
+	if params.InfraEnvUpdateParams.AdditionalTrustBundle != nil {
+		additionalTrustBundle := strings.TrimSpace(*params.InfraEnvUpdateParams.AdditionalTrustBundle)
+		params.InfraEnvUpdateParams.AdditionalTrustBundle = &additionalTrustBundle
+	}
+
 	if params, err = b.validateAndUpdateInfraEnvParams(ctx, &params); err != nil {
 		return nil, common.NewApiError(http.StatusBadRequest, err)
 	}
--- a/internal/controller/controllers/infraenv_controller.go
+++ b/internal/controller/controllers/infraenv_controller.go
@@ -167,6 +167,9 @@ func (r *InfraEnvReconciler) updateInfraEnv(ctx context.Context, log logrus.Fiel
 	if infraEnv.Spec.SSHAuthorizedKey != internalInfraEnv.SSHAuthorizedKey {
 		updateParams.InfraEnvUpdateParams.SSHAuthorizedKey = &infraEnv.Spec.SSHAuthorizedKey
 	}
+	if strings.TrimSpace(infraEnv.Spec.AdditionalTrustBundle) != internalInfraEnv.AdditionalTrustBundle {
+		updateParams.InfraEnvUpdateParams.AdditionalTrustBundle = &infraEnv.Spec.AdditionalTrustBundle
+	}
 
 	pullSecret, err := r.PullSecretHandler.GetValidPullSecret(ctx, getPullSecretKey(infraEnv.Namespace, infraEnv.Spec.PullSecretRef))
 	if err != nil {
@@ -373,16 +376,14 @@ func (r *InfraEnvReconciler) ensureISO(ctx context.Context, log logrus.FieldLogg
 				clusterID = cluster.ID
 				openshiftVersion = cluster.OpenshiftVersion
 			}
-			infraEnvInternal, err = r.createInfraEnv(ctx, log, &key, infraEnv, clusterID, openshiftVersion)
+			infraEnvInternal, err = r.createInfraEnv(ctx, log, &key, infraEnv, clusterID, openshiftVersion, infraEnv.Spec.AdditionalTrustBundle)
 			if err != nil {
 				log.Errorf("fail to create InfraEnv: %s, ", infraEnv.Name)
 				return r.handleEnsureISOErrors(ctx, log, infraEnv, err, nil)
-			} else {
-				return r.updateInfraEnvStatus(ctx, log, infraEnv, infraEnvInternal)
 			}
-		} else {
-			return r.handleEnsureISOErrors(ctx, log, infraEnv, err, infraEnvInternal)
+			return r.updateInfraEnvStatus(ctx, log, infraEnv, infraEnvInternal)
 		}
+		return r.handleEnsureISOErrors(ctx, log, infraEnv, err, infraEnvInternal)
 	}
 
 	// Check for updates from user and update the infraenv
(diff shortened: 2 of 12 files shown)


internal/controller/controllers/infraenv_controller.go
```diff
@@ -406,6 +418,7 @@ func CreateInfraEnvParams(infraEnv *aiv1beta1.InfraEnv, imageType models.ImageTy
 			CPUArchitecture:        infraEnv.Spec.CpuArchitecture,
 			ClusterID:              clusterID,
 			OpenshiftVersion:       openshiftVersion,
+			AdditionalTrustBundle:  additionalTrustBundle,
```
carbonin, line 421: Why is this `additionalTrustBundle` not coming from `infraEnv.Spec.AdditionalTrustBundle`?

internal/controller/controllers/infraenv_controller.go
```diff
@@ -366,23 +374,27 @@ func (r *InfraEnvReconciler) ensureISO(ctx context.Context, log logrus.FieldLogg
...
 			if cluster != nil {
 				clusterID = cluster.ID
 				openshiftVersion = cluster.OpenshiftVersion
 			}
-			infraEnvInternal, err = r.createInfraEnv(ctx, log, &key, infraEnv, clusterID, openshiftVersion)
+			// The OpenShift installer validation code for the additional trust bundle
+			// is buggy and doesn't react well to additional newlines at the end of the
+			// certs. We need to strip them out to not bother assisted users with this
+			// quirk.
+			additionalTrustBundle := strings.TrimSpace(infraEnv.Spec.AdditionalTrustBundle)
```
carbonin, line 389: I don't think this should be done here. Can we do this when we get the internal request to create the infraenv instead of in the controller?

carbonin (changes requested):
API looks okay to me, but it seems like you're passing around this string unnecessarily.

Also since you're touching ABI probably worth it for @bfournie to have a look.

internal/bminventory/inventory.go
```diff
@@ -4564,6 +4564,12 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
 
 	params.InfraenvCreateParams.CPUArchitecture = common.NormalizeCPUArchitecture(params.InfraenvCreateParams.CPUArchitecture)
 
+	// The OpenShift installer validation code for the additional trust bundle
+	// is buggy and doesn't react well to additional newlines at the end of the
+	// certs. We need to strip them out to not bother assisted users with this
+	// quirk.
+	params.InfraenvCreateParams.AdditionalTrustBundle = strings.TrimSpace(params.InfraenvCreateParams.AdditionalTrustBundle)
+
```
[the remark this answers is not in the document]
carbonin: Triplicate?

It's already in this function here https://github.com/openshift/assisted-service/pull/5357/files#diff-747a429a0e12307ef65d754c83b6898ce3858b621005e1e7af6e420da1f1657cR4623-R4627

Surely it doesn't also need to be in the controller and here again?

internal/bminventory/inventory.go
```diff
@@ -4614,12 +4614,6 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
 		}
 	}
 
-	// The OpenShift installer validation code for the additional trust bundle
-	// is buggy and doesn't react well to additional newlines at the end of the
-	// certs. We need to strip them out to not bother assisted users with this
-	// quirk.
-	params.InfraenvCreateParams.AdditionalTrustBundle = strings.TrimSpace(params.InfraenvCreateParams.AdditionalTrustBundle)
```
carbonin, line 4621: Don't we also need this for the REST API case?

internal/controller/controllers/infraenv_controller.go
```diff
@@ -167,6 +167,9 @@ func (r *InfraEnvReconciler) updateInfraEnv(ctx context.Context, log logrus.Fiel
 	if infraEnv.Spec.SSHAuthorizedKey != internalInfraEnv.SSHAuthorizedKey {
 		updateParams.InfraEnvUpdateParams.SSHAuthorizedKey = &infraEnv.Spec.SSHAuthorizedKey
 	}
+	if infraEnv.Spec.AdditionalTrustBundle != internalInfraEnv.AdditionalTrustBundle {
```
carbonin, line 170: Thinking about this a bit more ... if we trim this for the user on the way in won't this always be true if the added it with a space?

Is that a problem? Wondering if this could cause a hot loop.

internal/bminventory/inventory.go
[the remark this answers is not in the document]
carbonin: Why move this? Just leave it where it was previously.

internal/controller/controllers/infraenv_controller.go
paul-maidment: If the only way that this value can be set is through the registerInfraEnvInteral and updateInfraEnvInternal methods and these both perform the trim operation then I don't see the problem?

Or am I mistaken?

carbonin: The value in the infraenv spec will continue to have the space and the stored value will be updated.

internal/controller/controllers/infraenv_controller.go
```diff
@@ -366,23 +369,22 @@ func (r *InfraEnvReconciler) ensureISO(ctx context.Context, log logrus.FieldLogg
 	}
 	infraEnvInternal, err := r.Installer.GetInfraEnvByKubeKey(key)
 	if err != nil {
+		// If the infraenv does not exist then create it.
```
carbonin, line 372: This isn't needed I think

carbonin (approved):

internal/bminventory/inventory.go
```diff
@@ -4940,6 +4940,15 @@ func (b *bareMetalInventory) UpdateInfraEnvInternal(ctx context.Context, params
...
 
+	// The OpenShift installer validation code for the additional trust bundle
+	// is buggy and doesn't react well to additional newlines at the end of the
+	// certs. We need to strip them out to not bother assisted users with this
+	// quirk.
+	if params.InfraEnvUpdateParams.AdditionalTrustBundle != nil {
+		additionalTrustBundle := strings.TrimSpace(*params.InfraEnvUpdateParams.AdditionalTrustBundle)
+		params.InfraEnvUpdateParams.AdditionalTrustBundle = &additionalTrustBundle
+	}
+
```
avishayt, line 4951: Why is this code copy-pasted here?

avishayt: The following line is b.validateAndUpdateInfraEnvParams() which contains the exact same code.  What am I missing here?

carbonin: You're not missing anything. I missed that.

@paul-maidment can you open a followup to further deduplicate this logic?

paul-maidment: Where exactly is the duplication, I'm not seeing it?

carbonin: Addressed in https://github.com/openshift/assisted-service/pull/5372

PR #5358, opened 2023-07-14
MGMT-15215: Allow setting of UI specific data for a cluster

paul-maidment:
Presently, when creating a cluster in the UI, there is a checkbox that the user can set to indicate that they wan t to use custom manifests.

Presently this will cause the upload of an empty manifest, the presence of which is later used to determine whether the checkbox is checked or not (and whether the custom manifest tab should be shown in the UI).

This is a clunky approach that confuses the user and leads to validation issues.

To rememdy this, we would like to give the UI team a facility to store
raw JSON data containing freeform UI specific settings for a cluster.

This PR enables that.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

- [] This change does not require a documentation update (docstring, `docs`, README, etc)

## Reviewers Checklist


swagger.yaml
```diff
@@ -7223,3 +7289,13 @@ definitions:
         type: string
         format: date-time
         description: Expiration time for the URL token.
+  
+  installer-ui-settings:
```
carbonin, line 7293: If this is already an unstructured json blob, I'm not sure there's any reason to create a separate object with a single field. Can we not just take in and return the string directly?

@avishayt any thoughts on this?

swagger.yaml
```diff
@@ -3786,6 +3786,72 @@ paths:
           schema:
             $ref: '#/definitions/error'
 
+  /v2/clusters/{cluster_id}/installer-ui-settings:
```
carbonin, line 3789: Why not just `ui-settings`? Why add `installer`? The whole application is an installer.

swagger.yaml
paul-maidment: Regarding the overly complex request object for the JSON blob, I agree, however I think we have ran into this issue before where anything submitted as part of an in: body clause in swagger needs to be defined as a schema

paul-maidment: and adding this to query params seems unwise

carbonin: Looks like you can just define it as a string.
We do it for install config overrides here:
https://github.com/openshift/assisted-service/blob/7d2554116546ab0ce0d26c8d9de8be0f77e3d6d7/swagger.yaml#L2288-L2293

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -671,6 +671,29 @@ func (b *bareMetalInventory) V2SetIgnoredValidations(ctx context.Context, params
...
+}
+
+func (b *bareMetalInventory) V2UpdateClusterInstallerUISettings(ctx context.Context, params installer.V2UpdateClusterInstallerUISettingsParams) middleware.Responder {
+	cluster, err := b.getCluster(ctx, params.ClusterID.String())
+	if err != nil {
+		err = errors.Wrapf(err, "failed to fetch cluster %s to get UI settings", *cluster.ID)
+		return installer.NewV2UpdateClusterInstallerUISettingsInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+	}
+	cluster.UISettings = params.UISettings
+	if err = b.db.Save(cluster).Error; err != nil {
```
carbonin, line 690: Is this really the best way to do this? Theoretically this could be single statement in sql. Do we have a way to do that?

swagger.yaml
```diff
@@ -7223,3 +7289,4 @@ definitions:
         type: string
         format: date-time
         description: Expiration time for the URL token.
+
```
carbonin, line 7292: No need for bonus new line I think

internal/bminventory/inventory_v2_handlers.go
carbonin: Alternatively you could start a transaction and do a get `FOR UPDATE`.

The install config overrides endpoint does this but I imagine that's just because it also needs to do validation.

swagger.yaml
```diff
@@ -3786,6 +3786,72 @@ paths:
...
+          required: true
+      responses:
+        "200":
+          description: Success.
+          schema:
+            type: string
+        "400":
+          description: Error.
+          schema:
+            $ref: '#/definitions/error'
```
avishayt, line 3810: In what case would 400 be returned?

swagger.yaml
```diff
@@ -3786,6 +3786,72 @@ paths:
...
+            type: string
+      responses:
+        "200":
+          description: Success.
+          schema:
+            type: string
+        "400":
+          description: Error.
+          schema:
+            $ref: '#/definitions/error'
```
avishayt, line 3845: In what case would 400 be returned?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -671,6 +671,23 @@ func (b *bareMetalInventory) V2SetIgnoredValidations(ctx context.Context, params
...
+		err = errors.Wrapf(err, "failed to fetch cluster %s to get UI settings", *cluster.ID)
+		return installer.NewV2GetClusterInstallerUISettingsInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+	}
+	return installer.NewV2GetClusterInstallerUISettingsOK().WithPayload(cluster.UISettings)
+}
+
+func (b *bareMetalInventory) V2UpdateClusterInstallerUISettings(ctx context.Context, params installer.V2UpdateClusterInstallerUISettingsParams) middleware.Responder {
+	if err := b.db.Model(&common.Cluster{}).Where("id = ?", params.ClusterID.String()).Updates(map[string]interface{}{"ui_settings": params.UISettings}).Error; err != nil {
+		b.log.WithError(err).Errorf("failed to apply UI settings to cluster %s", params.ClusterID.String())
+		return installer.NewV2UpdateClusterInstallerUISettingsInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
```
avishayt, line 686: Do we want to check for 404, or always return 500?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -671,6 +671,29 @@ func (b *bareMetalInventory) V2SetIgnoredValidations(ctx context.Context, params
...
 }
 
+func (b *bareMetalInventory) V2GetClusterInstallerUISettings(ctx context.Context, params installer.V2GetClusterInstallerUISettingsParams) middleware.Responder {
+	cluster, err := b.getCluster(ctx, params.ClusterID.String())
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return installer.NewV2GetClusterInstallerUISettingsNotFound().WithPayload(common.GenerateError(http.StatusNotFound, err))
+		}
+		err = errors.Wrapf(err, "failed to fetch cluster %s to get UI settings", *cluster.ID)
+		return installer.NewV2GetClusterInstallerUISettingsInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
```
carbonin, line 681: Please use [`common.GenerateErrorResponder`](https://github.com/openshift/assisted-service/blob/3756e7f3c06bdc66446f29bb9f600af0d8bd6fb2/internal/common/error_utils.go#L118-L133) this handles returning 404 for gorm not found errors as well as some other cases with wrapped errors.

internal/bminventory/inventory_test.go
```diff
@@ -5665,6 +5665,32 @@ var _ = Describe("cluster", func() {
 			})
 		})
 
+		It("Should store and retrieve the UI specific settings for a cluster when requested to do so", func() {
```
carbonin, line 5668: Can you add an error case for 404 at least since we're specifically addressing that?

swagger.yaml
```diff
@@ -3786,6 +3786,64 @@ paths:
...
+      description: Fetch cluster specific installer UI settings.
+      operationId: v2GetClusterInstallerUISettings
+      parameters:
+        - in: path
+          name: cluster_id
+          description: The cluster for which UI settings should be retrieved.
+          type: string
+          format: uuid
+          required: true
+      responses:
```
carbonin, line 3802: Shouldn't we also be including 401 at least, maybe also 403?

swagger.yaml
```diff
@@ -3786,6 +3786,64 @@ paths:
           schema:
             $ref: '#/definitions/error'
 
+  /v2/clusters/{cluster_id}/ui-settings:
+    get:
+      tags:
+        - installer
+      description: Fetch cluster specific installer UI settings.
+      operationId: v2GetClusterInstallerUISettings
```
carbonin, line 3794: I think you can remove `Installer` from this as well.

carbonin (changes requested):

swagger.yaml
paul-maidment: I disagree with this, removing this tag sees the code generation placing the request objects under the operations package, somehow when it does this, you lose the ability to return a payload. 

So I will bve sticking with this tag for now.

paul-maidment: Do you have any suggestion on how this can be made to work or are we OK to keep the installer tag?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -671,6 +671,29 @@ func (b *bareMetalInventory) V2SetIgnoredValidations(ctx context.Context, params
 	return installer.NewV2SetIgnoredValidationsCreated().WithPayload(&ignoredValidations)
 }
 
+func (b *bareMetalInventory) V2GetClusterInstallerUISettings(ctx context.Context, params installer.V2GetClusterInstallerUISettingsParams) middleware.Responder {
```
carbonin, line 674:
```suggestion
func (b *bareMetalInventory) V2GetClusterUISettings(ctx context.Context, params installer.V2GetClusterUISettingsParams) middleware.Responder {
```

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -671,6 +671,29 @@ func (b *bareMetalInventory) V2SetIgnoredValidations(ctx context.Context, params
...
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return installer.NewV2GetClusterInstallerUISettingsNotFound().WithPayload(common.GenerateError(http.StatusNotFound, err))
+		}
+		err = errors.Wrapf(err, "failed to fetch cluster %s to get UI settings", *cluster.ID)
+		return installer.NewV2GetClusterInstallerUISettingsInternalServerError().WithPayload(common.GenerateError(http.StatusInternalServerError, err))
+	}
+	return installer.NewV2GetClusterInstallerUISettingsOK().WithPayload(cluster.UISettings)
+}
+
+func (b *bareMetalInventory) V2UpdateClusterInstallerUISettings(ctx context.Context, params installer.V2UpdateClusterInstallerUISettingsParams) middleware.Responder {
```
carbonin, line 686:
```suggestion
func (b *bareMetalInventory) V2UpdateClusterUISettings(ctx context.Context, params installer.V2UpdateClusterUISettingsParams) middleware.Responder {
```

swagger.yaml
carbonin:
```suggestion
      operationId: V2GetClusterUISettings
```

carbonin (changes requested):

internal/bminventory/inventory_test.go
paul-maidment: Added

swagger.yaml
paul-maidment: Done

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -671,6 +671,26 @@ func (b *bareMetalInventory) V2SetIgnoredValidations(ctx context.Context, params
...
+	return installer.NewV2GetClusterUISettingsOK().WithPayload(cluster.UISettings)
+}
+
+func (b *bareMetalInventory) V2UpdateClusterUISettings(ctx context.Context, params installer.V2UpdateClusterUISettingsParams) middleware.Responder {
+	cluster, err := b.getCluster(ctx, params.ClusterID.String())
+	if err != nil {
+		return common.GenerateErrorResponder(err)
+	}
+	cluster.UISettings = params.UISettings
+	if err = b.db.Save(cluster).Error; err != nil {
```
carbonin, line 688: Generally I'd advise against this and say it's outright incorrect without a transaction.

This will re-save every field in `cluster` which will override any other requests that came in between when we got the cluster and when we saved it.

If the one line with `Updates` doesn't work then I'm fine with one call checking for existence then the second call being the update to only the one column.

swagger.yaml
carbonin: I think this is still a valid question. If you're not validating the input I'm not sure when we'd return 400.

internal/bminventory/inventory_test.go
```diff
@@ -5665,6 +5665,48 @@ var _ = Describe("cluster", func() {
 			})
 		})
 
+		It("Should handle not found correctly on fetch of UI specific settings for cluster", func() {
```
carbonin, line 5668: nit, but a better name for this would explain the condition when this situation happens rather than "handles not found correctly"

So something like:
```
It("returns not found when the cluster doesn't exist", func() {
```

Also I'd group all these function that are testing `V2GetClusterUISettings` in a `Describe` block.

internal/bminventory/inventory_test.go
```diff
@@ -5665,6 +5665,48 @@ var _ = Describe("cluster", func() {
...
+				Expect(reply).Should(BeAssignableToTypeOf(installer.NewV2GetClusterUISettingsOK()))
+				Expect(reply.(*installer.V2GetClusterUISettingsOK).Payload).To(Equal(""))
+			})
+			By("Should be able to store UI specific settings for the cluster", func() {
+				settingsJSON := "{\"foo\":\"bar\"}"
+				reply := bm.V2UpdateClusterUISettings(ctx, installer.V2UpdateClusterUISettingsParams{
+					ClusterID:  clusterID,
+					UISettings: settingsJSON,
+				})
+				Expect(reply).Should(BeAssignableToTypeOf(installer.NewV2UpdateClusterUISettingsOK()))
```
carbonin, line 5699: What are you actually testing here?

This should verify the correct value was set in the database.

Also this should be separate from the `Get` tests. I would expect those to set some state in the DB and then test that the API returns the value correctly.

eliorerz:
I think that there is something that I might missing here. Why do we need that? Only for custom manifests? If so so why not just add a boolean to the cluster that indicates that the user selected the custom manifests or there are custom manifests.
We can use this indicator in different places to know if there are custom manifests of not without calling to `/v2/clusters/{cluster_id}/manifests` (e.g validations, feature-usage and more)
Further more this can cause incompatibility issues when the user uses API + UI (some flows can be done only when using the API)

swagger.yaml
```diff
@@ -3786,6 +3786,88 @@ paths:
           schema:
             $ref: '#/definitions/error'
 
+  /v2/clusters/{cluster_id}/ui-settings:
```
eliorerz, line 3789: I moved this comment to here so the entire conversation will be in single thread.
I think that there is something that I might missing here. Why do we need that? Only for custom manifests? If so so why not just add a boolean to the cluster that indicates that the user selected the custom manifests or there are custom manifests.
We can use this indicator in different places to know if there are custom manifests of not without calling to /v2/clusters/{cluster_id}/manifests (e.g validations, feature-usage and more)
Further more this can cause incompatibility issues when the user uses API + UI (some flows can be done only when using the API)

eliorerz: @avishayt @gamli75 @paul-maidment @jkilzi

jkilzi: It is not going to be only for custom manifests. The API is for persisting UI-only state.

paul-maidment: Also, the intention of this boolean is not as you describe it

This setting is **UI Specific**, meaning that it is only to be used to determine the state of a UI component (in this case the "tab" that determines whether or not you will be able to add custom manifests to your installation.)

A cluster with this UI preference set is not guaranteed to have custom manifests uploaded to it. A connection to S3 would still be required to determine whether or not there are manifests available.

I understand the confusion but the intent and scope of this field does not go beyond the UI.

In discussion with @filanov and @jkilzi it was agreed that the UI team would in fact like a field to hold generic UI values, owned by the UI team for which they would take responsibility.

filanov: Why does it need to be a new api? why not just make it as part of cluster create and update?

paul-maidment: The intent of this field is to make sure that it is for UI specific uses. 

Would it be acceptable to export this field (which has a UI specific purpose) when giving API responses via the UI. It is my understanding that this would be necessary if we take the approach suggested by @filanov 

@avishayt do you have any thoughts on this?

jkilzi: I think it is more of a separation of concerns. Why does the Cluster as a stand-alone entity need an attribute called ui_settings?

paul-maidment: `I think it is more of a separation of concerns. Why does the Cluster as a stand-alone entity need an attribute called ui_settings?` @jkilzi - I agree

avishayt: On the one hand, I agree that separate is nicer in this case.  However, will the UI need to "atomically" update both the regular cluster properties and the UI settings?  If so, a cluster property would help.

paul-maidment: @jkilzi is this the case, does the UI need to atomically update cluster state and UI settings?

filanov: Elior just to make it clear this field is used by the ui to store specific status for the wizard, for example if user is planning to add custom manifests or static network or any other field in the future, instead of having a new boolean each time we prefer to have a field that the ui will manage and store the data there

eliorerz: > Elior just to make it clear this field is used by the ui to store specific status for the wizard,

I know, but while working on https://github.com/openshift/assisted-service/pull/5364 I needed this exact property. I think that 

> is not relevant we are not support it

Currently, but why to create this separation now? I think that the objects that we currently have (e.g. cluster, infraEnv.. ) supposed to reflect the state. To my opinion when we add this kind of API we are allowing to hold another state on the same objects, that for the long run I believe will cause us more headaches.
For example what about feature usages? If the user mark the `Custom Manifests` checkbox, how can we tell that the user did that? 

> instead of having a new boolean each time we prefer to have a field that the ui will manage and store the data there

I don't think that we will need to add a boolean each time, but in this case I think we should (from the reasons I mentioned above). If I encountered this need only few days ago I'm sure that others will need i n the future.

paul-maidment: "I know, but while working on https://github.com/openshift/assisted-service/pull/5364 I needed this exact property. I think that"
It appears that the purpose of that PR is to block installation if there are no manifests uploaded, the subtle distinction here is that this flag does not guarantee that user manifests are available as it exists for the sole purpose of storing a user selected "checkbox" to indicate that they wish to add custom manifests.

It is still entirely possible that the user will make this selection and then fail to add manifests. So this boolean does not serve the purpose you describe. (not without very strict validation/enforcement prior to starting installation at least.)

The purpose of this checkbox is completely for UI use, it has nothing to do with the cluster domain.


"Currently, but why to create this separation now? " -
The way I view this is that we place the UISettings "JSON blob" object inside the cluster database table only for convenience and to save the expense of creating a completely independent datastore, for this task it's overkill.

I see this as an implementation detail rather than something that belongs to the cluster. The API call encapsulates this neatly and allows for the furture possibility of improving this isolatin by moving the underlying data to a more appropriate structure. 

To tie this data structure to the cluster by exposing the field in the API is tight coupling, generally discouraged as good design practice.

Moreover it exists (at the request of the UI team) for the purpose of providing a simple way in which they can store "UI settings" for a specific "workflow" in a cluster.

I personally think that this PR strikes the right balance of separation of concerns and capability to keep most stakeholders happy.

I do hear the concern about having a field that would negate the need to connect to S3 to simply validate that the user has uploaded custom manifests. My thoughts are that this should be addressed elsewhere rather than in this PR, for which the focus is strictly UI.

On that topic, I have my own apprehensions about why we even use S3 to store manifests instead of a database, cases like the one identified by @eliorerz would be greatly simplified, but realise that a migration would be costly and require a lot of work.

avishayt: The UI should only use this field for UI state that doesn't affect the cluster/host/infraenv resources.  In the example of custom manifests, the UI isn't saving the state of "the user added custom manifests" but rather "the user clicked that they want to add custom manifests so we should show that step in the wizard".  Another example might be "the user didn't choose platform integration in the first screen, so we'll allow them to choose UMN/CMN in the network screen".  This is more straight-forward to save in a special UI field than adding random fields to the cluster resource.

filanov: Agree with what Avishay said, in addition i don't think that https://github.com/openshift/assisted-service/pull/5364 is needed if we have this api, the ui will not create an empty manifest and we don't care for the logic in the wizard for this case.

avishayt: We still need #5364 - it validates that a custom manifest was provided for the OCI platform

jkilzi: For the record, the UI doesn't need to update this endpoint and the cluster endpoint atomically.

PR #5364, opened 2023-07-17
MGMT-15264: Block installation if no custom manifests in OCI cluster

eliorerz:
/cc @eliorerz 

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

### DIFF
--- a/internal/cluster/validator.go
+++ b/internal/cluster/validator.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/host"
 	"github.com/openshift/assisted-service/internal/network"
+	"github.com/openshift/assisted-service/internal/usage"
 	"github.com/openshift/assisted-service/models"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -398,6 +399,27 @@ func (v *clusterValidator) allHostsAreReadyToInstall(c *clusterPreprocessContext
 		return ValidationSuccess, "All hosts in the cluster are ready to install."
 	}
 	return ValidationFailure, "The cluster has hosts that are not ready to install."
+}
+
+func (v *clusterValidator) platformRequirementsSatisfied(c *clusterPreprocessContext) (ValidationStatus, string) {
+	// If cluster platform type is not OCI ignore that validation
+	if c.cluster.Platform != nil && common.PlatformTypeValue(c.cluster.Platform.Type) != models.PlatformTypeOci {
+		return ValidationSuccess, "Platform requirements satisfied"
+	}
+
+	usages, err := usage.Unmarshal(c.cluster.Cluster.FeatureUsage)
+	if err != nil {
+		v.log.Errorf("platform validation failure, failed to parse feature usages, %s", err.Error())
+		return ValidationFailure, "Failed to parse feature usages"
+	}
+
+	for _, usg := range usages {
+		if usg.Name == usage.CustomManifest {
+			return ValidationSuccess, "Platform requirements satisfied"
+		}
+	}
+
+	return ValidationFailure, "The custom manifest required for Oracle Cloud Infrastructure platform integration has not been added. Add a custom manifest to continue."
 }
 
 func (v *clusterValidator) isDNSDomainDefined(c *clusterPreprocessContext) (ValidationStatus, string) {
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -6614,6 +6614,7 @@ definitions:
       - 'lvm-requirements-satisfied'
       - 'mce-requirements-satisfied'
       - 'network-type-valid'
+      - 'platform-requirements-satisfied'
 
   logs_type:
     type: string
--- a/api/vendor/github.com/openshift/assisted-service/models/cluster_validation_id.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/cluster_validation_id.go
@@ -98,6 +98,9 @@ const (
 
 	// ClusterValidationIDNetworkTypeValid captures enum value "network-type-valid"
 	ClusterValidationIDNetworkTypeValid ClusterValidationID = "network-type-valid"
+
+	// ClusterValidationIDPlatformRequirementsSatisfied captures enum value "platform-requirements-satisfied"
+	ClusterValidationIDPlatformRequirementsSatisfied ClusterValidationID = "platform-requirements-satisfied"
 )
 
 // for schema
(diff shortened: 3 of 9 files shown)


swagger.yaml
```diff
@@ -6614,6 +6614,7 @@ definitions:
       - 'lvm-requirements-satisfied'
       - 'mce-requirements-satisfied'
       - 'network-type-valid'
+      - 'oci-manifests-satisfied'
```
gamli75, line 6617: I think we should add something more generic for other platforms that we will have in the future, something like: invalid_platform_configuration and the description will be more specific about OCI
@avishayt WDYT?

avishayt: I agree since we'll have at most one platform selected at a time.  "platform-requirements-satisfied" seems to fit with the others.

avishayt: "configuration" seems appropriate

gamli75: > "configuration" seems appropriate

please talk with UI about this new category before implementing it - they need to decide where to display this validation and maybe they are using the category
@jkilzi @nirfarkas FYI

avishayt: The category already exists.  It's currently only used for "pull secret set".

internal/cluster/validator.go
```diff
@@ -400,6 +401,26 @@ func (v *clusterValidator) allHostsAreReadyToInstall(c *clusterPreprocessContext
...
+		return ValidationFailure, "Failed to parse feature usages"
+	}
+
+	for _, usg := range usages {
+		if usg.Name == usage.CustomManifest {
+			return ValidationSuccess, "Found custom manifests"
+		}
+	}
+
+	return ValidationFailure, "Can't find custom manifests"
```
nirfarkas, line 421: We need to add here what needs to be done.
kaedward@redhat.com FYI

internal/cluster/validator.go
```diff
@@ -400,6 +401,26 @@ func (v *clusterValidator) allHostsAreReadyToInstall(c *clusterPreprocessContext
 	return ValidationFailure, "The cluster has hosts that are not ready to install."
 }
 
+func (v *clusterValidator) OciManifestsSatisfied(c *clusterPreprocessContext) (ValidationStatus, string) {
+	// If cluster platform type is not OCI ignore that validation
+	if !(c.cluster.Platform != nil && common.PlatformTypeValue(c.cluster.Platform.Type) == models.PlatformTypeOci) {
+		return ValidationSuccess, "OCI platform wasn't set"
```
avishayt, line 407: Once you make the validation name generic:
```suggestion
		return ValidationSuccess, "Platform requirements satisfied"
```

internal/cluster/validator.go
```diff
@@ -400,6 +401,26 @@ func (v *clusterValidator) allHostsAreReadyToInstall(c *clusterPreprocessContext
...
+	}
+
+	usages, err := usage.Unmarshal(c.cluster.Cluster.FeatureUsage)
+	if err != nil {
+		return ValidationFailure, "Failed to parse feature usages"
+	}
+
+	for _, usg := range usages {
+		if usg.Name == usage.CustomManifest {
+			return ValidationSuccess, "Found custom manifests"
```
avishayt, line 417:
```suggestion
			return ValidationSuccess, "Platform requirements satisfied"
```

internal/cluster/validator.go
avishayt: How about "The custom manifests required for OCI platform integration have not been added"

internal/cluster/validator.go
```diff
@@ -400,6 +401,27 @@ func (v *clusterValidator) allHostsAreReadyToInstall(c *clusterPreprocessContext
 	return ValidationFailure, "The cluster has hosts that are not ready to install."
 }
 
+func (v *clusterValidator) platformRequirementsSatisfied(c *clusterPreprocessContext) (ValidationStatus, string) {
+	// If cluster platform type is not OCI ignore that validation
+	if !(c.cluster.Platform != nil && common.PlatformTypeValue(c.cluster.Platform.Type) == models.PlatformTypeOci) {
```
gamli75, line 406:
```suggestion
	if (c.cluster.Platform != nil && common.PlatformTypeValue(c.cluster.Platform.Type) != models.PlatformTypeOci) {
```
more readable

internal/cluster/validator.go
```diff
@@ -400,6 +401,27 @@ func (v *clusterValidator) allHostsAreReadyToInstall(c *clusterPreprocessContext
...
+		return ValidationSuccess, "Platform requirements satisfied"
+	}
+
+	usages, err := usage.Unmarshal(c.cluster.Cluster.FeatureUsage)
```
gamli75, line 410: why we are using the feature usage for checking it and not data from the cluster? what will happen if there will be a bug in the feature usage?

internal/cluster/validator.go
```diff
@@ -400,6 +401,27 @@ func (v *clusterValidator) allHostsAreReadyToInstall(c *clusterPreprocessContext
...
+	if err != nil {
+		v.log.Errorf("platform validation failure, failed to parse feature usages, %s", err.Error())
+		return ValidationFailure, "Failed to parse feature usages"
+	}
+
+	for _, usg := range usages {
+		if usg.Name == usage.CustomManifest {
```
gamli75, line 417: did you verify that in case the user added a custom manifest and then removed it - there is no usage.CustomManifest in the list?

gamli75 (commented):
I don't remember if we have a specific code but we should not run this validation during the installation

internal/cluster/validator.go
avishayt: I actually just fixed that :)
https://github.com/openshift/assisted-service/pull/5363

eliorerz: With https://github.com/openshift/assisted-service/pull/5363 it should, I'll re verify it

internal/cluster/validator.go
KaelanKhiatani: Is "OCI platform integration" something that the user will understand? If it is then go ahead and keep it, otherwise I think it can be taken out to avoid confusion.
**Cluster installation failed**
The custom manifest required for installation has not been added. Add a custom manifest to BLANK and try again"

nirfarkas: Some context and clarifications:

- This message should be shown only in the CLI (because in the UI you can't proceed in the wizard if you don't provide manifests
- This message is not about a failure of a cluster installation, the cluster installation was not triggered yet
- I don't think we should call it "OCI" as we call the integration "Oracle" in the UI drop-down. If we want to stick with "OCI", we should also adjust the UI.

avishayt: "Oracle" is too generic - it should be "Oracle Cloud Infrastructure"

avishayt: Regarding the message here, how about:
The custom manifests required for OCI platform integration have not been added. Add the custom manifests and try again.

nirfarkas: Then it should be the same in the UI.

eliorerz: So what is the decision?

nirfarkas: "The custom manifest required for Oracle Cloud Infrastructure platform integration has not been added. Add a custom manifest to continue."

gamli75 (approved):

PR #5375, opened 2023-07-21
NO-ISSUE: Run network config before NetworkManager

jhernand:
Currently the systemd service that we use to apply static network configuration from the non-minimal discovery environment wants to run after `dracut-cmdline.service` and before `dracut-initqueue.service`. But that isn't possible because this service is created by an ignition configuration, and services created that way only run after the root has pivoted, and at that point the `cmdline` and `initqueue` steps of Dracut have already been completed.

The net result is that this ordering dependencies are just ignored, and the service will run as part of `multi-user.target` without any special ordering restriction. It may even run before NetworkManager. If that happens then the configuration files that it generates will be ignored. That is unlikely, but it have been observed when there are many network interfaces with static configuration. To avoid that the service should be explicitly configured to run before NetworkManager.

Note that there is another diferent systemd service that is used to generate the network config when using the minimal ISO. That is added directly to the initrd, not via ignition. It runs before the root pivots because the network is needed to fetch the root file system. This patch doesn't change that service.

Related: https://issues.redhat.com//browse/OCPBUGS-16219
Related: https://github.com/openshift/installer/pull/7355

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


zaneb:
typo:

> It may even run before NetworkManager. If that happens then the configuration files that it generates will be ignored.

s/before/after/

Also, I think this is worth opening a bug for, because this has been wrong for a while and might need to be backported.

/lgtm

jhernand:
> typo:
> 
> > It may even run before NetworkManager. If that happens then the configuration files that it generates will be ignored.
> 
> s/before/after/

Done.

> Also, I think this is worth opening a bug for, because this has been wrong for a while and might need to be backported.

Here is the bug: https://issues.redhat.com/browse/MGMT-15339

romfreiman:
@tsorya @eranco74 ptal

jhernand:
To verify this I introduced an artificial delay in the script that generates the NetworkManager configuration, something like this:

```diff
diff --git a/internal/constants/scripts.go b/internal/constants/scripts.go
index c118f254c..89be73284 100644
--- a/internal/constants/scripts.go
+++ b/internal/constants/scripts.go
@@ -30,6 +30,8 @@ package constants
 //    '/etc/coreos-firstboot-network' when the script is part of the minimal ISO.
 const PreNetworkConfigScript = `#!/bin/bash
 
+sleep 60
+
 PATH_PREFIX=${PATH_PREFIX:=''}
 
 # The directory that contains nmconnection files of all nodes
```

Then I tried to install a cluster without this patch and using static network configuration. I observed that the static network configuration was ignored and instead the network interfaces got the IP addresses from the DHCP server.

I repeated the same test with this patch and then the network interfaces got the IP addresses specified in the configuration.

PR #5397, opened 2023-07-28
MGMT-12397: Enhancement document for pre installation error improvements.

paul-maidment:
Background:
When a cluster installation is triggered, either via Assisted Installer KubeAPI or via the UI, there are preparation steps that take place before the installation is launched. Presently, not all of these failures are appropriately reported back to users and there is a lack of consistency. The aim of this enhancement is to come up with a set of changes that improve this situation greatly. Please see [MGMT-12397](https://issues.redhat.com//browse/MGMT-12397) for more background on this.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/improved-handling-for-pre-installation-errors.md
+++ b/docs/enhancements/improved-handling-for-pre-installation-errors.md
@@ -1,0 +1,128 @@
+---
+title: Improved handling for pre installation errors
+authors:
+  - "@pmaidmen"
+creation-date: 2023-07-29
+last-updated: 2023-07-29
+---
+
+# Summary
+
+When a cluster installation is triggered, either via Assisted Installer KubeAPI or via the UI, there are preparation steps that take place before the installation is launched. Presently, not all of these failures are appropriately reported back to users and there is a lack of consistency.
+The aim of this enhancement is to come up with a set of changes that improve this situation greatly.
+Please see MGMT-12397 for more background on this.
+
+# Issues identified
+
+There are a number of issues that contribute to a lack of clarity for the end user.
+
+* There is no guarantee that every preinstallation failure will be reported back to the user, either via KubeAPI or via the UI.
+* Some parts of the preinstallation depend on other conditions, at least with these the cause of the error can be somewhat determined.
+* The UI does not show enough detail in the error summary for the user to have a clue why an error occurred.
+* Inconsistent error handling here https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory.go#L1293-L1360
+    - Sure, every error is handled but not enough is done to ensure this is comminicated in a useful way to the user
+    - Also inconsistent terminology is used here, making it unclear if it's important that we are failing a Preparation or a Preinstallation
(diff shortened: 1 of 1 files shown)


docs/enhancements/improved-handling-for-pre-installation-errors.md
```diff
@@ -0,0 +1,128 @@
...
+- In cluster validation, add a validation to show the status of cluster preinstallation and any errors.
+- Ensure that the UI is updated appropriately to display failure information.
+
+### Non-Goals
+
+- TBC
+
+### Implementation proposal 
+
+    - Add a cluster "pre installation status" validation
```
avishayt, line 46: How does the failing validation succeed again so that the user can retry?

docs/enhancements/improved-handling-for-pre-installation-errors.md
```diff
@@ -0,0 +1,128 @@
...
+title: Improved handling for pre installation errors
+authors:
+  - "@pmaidmen"
+creation-date: 2023-07-29
+last-updated: 2023-07-29
+---
+
+# Summary
+
+When a cluster installation is triggered, either via Assisted Installer KubeAPI or via the UI, there are preparation steps that take place before the installation is launched. Presently, not all of these failures are appropriately reported back to users and there is a lack of consistency.
```
gamli75, line 11: few remarks/ questions:
1. I think you should add more details about what are we running today in this step. 
2. Do we run it in parallel? 
3. Is there something that the user can do to fix it?
4. Do we need the ability to collect data per each failure in this stage? 
5. Are we going to change the cluster state machine due to this change?

paul-maidment: @gamli75 

I think that we are talking generally about failures during this step, I would say that we do not have a specific enough "list" of unique errors in this area and that any approach to error reporting will certainly be catching "general failures" in this area, this could (non exhaustive list) be due to.

Because we are dealing with an agent, I am sure that a number of activities do run in parallel on the agent. But not all steps in the installation process appear to be performed in parallel.

**Preparation failures:**
A poorly defined manifest file
Bad networking (might manifest as an inability to pull images)
A failure in an external service

I think that the best approach is to catch any general failure to prepare and then to report on this to the end user. Certainly any error that prevents the installation from progressing will require user attention. Presenting the user with the error message in any exception we encounter should enable the user to take this action.

As I have alluded to, there are a number of possible ways in which preparation can fail and I guess we could try to enumerate some of them, however, to do this would increase the scope of this ticket to a refactor of the preparation code itself and I am not sure this is in scope here.

For API use there will be no change to the cluster state and the user will still be able to restart the installation in the same way as they did before. 

The "Installation attempts counter" will make  a change to cluster state as we will need to track whether or not the installation limit has been reached to prevent the installation becoming stuck in an infinite and fruitless "retry" loop.

docs/enhancements/improved-handling-for-pre-installation-errors.md
avishayt: Did we get any input from UX on this?  How would this kind of failure be displayed ideally?

@nirfarkas

PR #5413, opened 2023-08-04
MGMT-14933: Validate hosts running in OCI

adriengentil:
Check the manufacturer to check if hosts are running in OCI when
platform OCI is selected.


adriengentil:
@eliorerz is there any other places where I need to add validations ?

avishayt (approved):

PR #5416, opened 2023-08-06
MGMT-15491: Validate vSphere disk.EnableUUID ignoring cluster platform

eliorerz:
If disk.EnableUUID flag is not enabled on vSphere VMs in any platform, including baremetal, then no symlinks are generated in /dev/disk/by-id for attached disks.

Installing ODF via LSO or something on such a platform results in somewhat fragile installation because disks themselves could be renamed on reboot and since no permanent ids exists for disks, the PVs could become invalid.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

/cc @eliorerz

### DIFF
--- a/internal/host/host_test.go
+++ b/internal/host/host_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/openshift/assisted-service/internal/metrics"
 	"github.com/openshift/assisted-service/internal/operators"
 	"github.com/openshift/assisted-service/internal/operators/api"
+	"github.com/openshift/assisted-service/internal/provider/registry"
 	"github.com/openshift/assisted-service/internal/versions"
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/assisted-service/pkg/conversions"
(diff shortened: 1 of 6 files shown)


romfreiman:
Lets understand whether this is indeed the right behavior. And in case it's a blocking validation...

internal/common/common_unitest_db.go
```diff
@@ -407,7 +407,7 @@ func openTestDB(dbName string) (*gorm.DB, error) {
 		})
 	}
 
-	for attempts := 0; attempts < 5; attempts++ {
+	for attempts := 0; attempts < 30; attempts++ {
```
eliorerz, line 410: Not related, will be removed on the next push

gamli75 (approved):

PR #5418, opened 2023-08-07
MGMT-15425: Get subscription version by a given openshift version

eliorerz:
/cc @eliorerz 

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

### DIFF
--- a/internal/operators/mce/manifest.go
+++ b/internal/operators/mce/manifest.go
@@ -2,11 +2,15 @@ package mce
 
 import (
 	"bytes"
+	"fmt"
+	"strconv"
 	"text/template"
+
+	"github.com/openshift/assisted-service/internal/common"
 )
 
 // Manifests returns manifests needed to deploy MCE.
-func Manifests() (openshiftManifests map[string][]byte, customManifests []byte, err error) {
+func Manifests(openshiftVersion string) (openshiftManifests map[string][]byte, customManifests []byte, err error) {
 	// Generate the OpenShift manifests:
 	namespaceManifest, err := getNamespace()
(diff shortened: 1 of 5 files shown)


internal/operators/mce/manifest.go
```diff
@@ -34,11 +38,20 @@ func Manifests() (openshiftManifests map[string][]byte, customManifests []byte,
 	return openshiftManifests, mceManifest, nil
 }
 
-func getSubscription() ([]byte, error) {
+func getSubscriptionChannel(openshiftVersion string) string {
+	// OCP 4.(10 + x) -> MCE 2.x
+	parsedOpenshiftVersion := common.GetParsedVersion(openshiftVersion)
+	openshiftMinorVersion, _ := strconv.Atoi(parsedOpenshiftVersion[1])
+	mceChannelMinor := openshiftMinorVersion - 10
```
gamli75, line 45: this code is problematic because now if the user installs 4.14.ec.X we need to install stable 2.3.
maybe it will be better if we will have a configuration value that we will be able to change without changing the code

eliorerz: I think that we need to understand if we do need to install stable 2.3.
After the attempt to do so for LSO I'm not sure that this is the right approach

gamli75:
closing this PR - was replaced by https://github.com/openshift/assisted-service/pull/5716

PR #5433, opened 2023-08-17
MGMT-15362: Import local cluster into ZTP

paul-maidment:
When managing a cluster via ZTP, either through MCE or any other method that results in assisted installer being conmfigured in ZTP mode, we want to import the local cluster into ZTP or which the assisted-operator is running.

The intent is to allow the user to perform day2 operations on the local cluster, to allow the addition of workers and so on.

Presently this is only possible via manual efforts and is not very customer friendly.

This PR aims to resolve this by adding functionality to import the local cluster as described above.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


Makefile
```diff
@@ -278,6 +278,9 @@ endef
 _verify_cluster:
 	python3 ./tools/wait_for_cluster_info.py --namespace "$(NAMESPACE)"
 
+# TODO: Write a makefile target to inject the right KubeAPI entities (for local cluster) into the DB so that we can test the "auto import" in ACM.
```
filanov, line 281: no need, it should be tested in an e2e test

cmd/main.go
```diff
@@ -641,6 +642,17 @@ func main() {
...
 
+	if Options.EnableKubeAPI {
+		// Splitting into API reader and writer interfaces as the reader bypasses the cache settings that are on the regular client
+		// and we need to be able to read secrets that we cannot match in cache settings
+		// The cachedApiClient can be used for writes as these are unaffected by cache.
+		apiReader := ctrlMgr.GetAPIReader()
+		cachedApiClient := ctrlMgr.GetClient()
+		localClusterImportOperations := localclusterimport.NewLocalClusterImportOperations(apiReader, cachedApiClient, log)
+		localClusterImport := localclusterimport.NewImportLocalCluster(&localClusterImportOperations, log, "local-cluster")
+		localClusterImport.ImportLocalCluster()
```
filanov, line 653: you can check and log the error here

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,262 @@
...
+	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/openshift/hive/apis/hive/v1/agent"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type Local_cluster_import struct {
```
filanov, line 17: please fix to camel case

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,262 @@
...
+	clusterImportOperations ClusterImportOperations
+	localClusterNamespaceName    string
+	log                          *logrus.Logger
+}
+
+func NewImportLocalCluster(localClusterImportOperations ClusterImportOperations, log *logrus.Logger, localClusterNamespaceName string) Local_cluster_import {
+	return Local_cluster_import{clusterImportOperations: localClusterImportOperations, log: log, localClusterNamespaceName: localClusterNamespaceName}
+}
+
+func (i *Local_cluster_import) shouldImportLocalCluster() (bool, error) {
```
filanov, line 27: what will happen after upgrade? will this cluster be usable? or do you need to re-create it?

cmd/main.go
```diff
@@ -641,6 +642,17 @@ func main() {
...
 	}()
 
+	if Options.EnableKubeAPI {
+		// Splitting into API reader and writer interfaces as the reader bypasses the cache settings that are on the regular client
+		// and we need to be able to read secrets that we cannot match in cache settings
+		// The cachedApiClient can be used for writes as these are unaffected by cache.
+		apiReader := ctrlMgr.GetAPIReader()
+		cachedApiClient := ctrlMgr.GetClient()
+		localClusterImportOperations := localclusterimport.NewLocalClusterImportOperations(apiReader, cachedApiClient, log)
+		localClusterImport := localclusterimport.NewImportLocalCluster(&localClusterImportOperations, log, "local-cluster")
```
filanov, line 652: if `local-cluster` is hard coded why do you need to add it a parameter?

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,262 @@
...
+		return errors.New(message)
+	}
+	namespace, err := i.clusterImportOperations.GetNamespace(i.localClusterNamespaceName)
+	if err != nil {
+		message := fmt.Sprintf("hubClusterDay2Importer: Unable to find the %s namesapce due to error: %s", i.localClusterNamespaceName, err.Error())
+		i.log.Error(message)
+		return err
+	}
+	if namespace != nil {
+		i.log.Infof("Found namespace %s", namespace.Name)
```
filanov, line 58: is that for debug? not sure how this message can help

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,262 @@
...
+		i.log.Info(message)
+		return errors.New(message)
+	}
+	namespace, err := i.clusterImportOperations.GetNamespace(i.localClusterNamespaceName)
+	if err != nil {
+		message := fmt.Sprintf("hubClusterDay2Importer: Unable to find the %s namesapce due to error: %s", i.localClusterNamespaceName, err.Error())
+		i.log.Error(message)
+		return err
+	}
+	if namespace != nil {
```
filanov, line 57: in general for get operations is there a case where the error was nil and the object is nil as well?

pkg/localclusterimport/local_cluster_import_operations.go
```diff
@@ -0,0 +1,179 @@
...
+	GetClusterVersion(name string) (*configv1.ClusterVersion, error)
+	GetClusterImageSet(name string) (*hivev1.ClusterImageSet, error)
+	CreateClusterImageSet(clusterImageSet *hivev1.ClusterImageSet) (*hivev1.ClusterImageSet, error)
+	CreateClusterDeployment(clusterDeployment *hivev1.ClusterDeployment) (*hivev1.ClusterDeployment, error)
+	GetNodes() (*v1.NodeList, error)
+	GetNumberOfControlPlaneNodes() (int, error)
+	GetDNS(name string) (*configv1.DNS, error)
+}
+
+type Local_cluster_import_operations struct {
```
filanov, line 32: please don't use snake case

pkg/localclusterimport/local_cluster_import_operations.go
```diff
@@ -0,0 +1,179 @@
...
+	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+//go:generate mockgen -source=local_cluster_import_operations.go -package=localclusterimport -destination=local_cluster_import_operations_mocks.go
+type ClusterImportOperations interface {
```
filanov, line 17: Why do you need a client wrapper? aren't those regular k8s functions that can be part of `Local_cluster_import`?

cmd/main.go
paul-maidment: I guess we could make it a constant

pkg/localclusterimport/import_local_cluster.go
paul-maidment: That's a good question, how do we know an upgrade has happened?

pkg/localclusterimport/import_local_cluster.go
paul-maidment: good point, I will remove the unnecessary checks

pkg/localclusterimport/local_cluster_import_operations.go
paul-maidment: It's not a very good separation of concerns to be honest. The business action of performing an import vs making operations available in the client are two distinct things.

Moreover it easier for creating a focused set of mocks also.

They could be merged together but I honestly don't think they should be.

pkg/localclusterimport/import_local_cluster.go
filanov: probably a different cluster version.

pkg/localclusterimport/local_cluster_import_operations.go
filanov: It's just calling a single function without any logic, this is why i asked.

pkg/localclusterimport/local_cluster_import_operations.go
```diff
@@ -0,0 +1,164 @@
...
+	return agentClusterInstall, nil
+}
+
+func (o *Local_cluster_import_operations) CreateAgentClusterInstall(agentClusterInstall *hiveext.AgentClusterInstall) error {
+	return o.cachedApiClient.Create(o.context, agentClusterInstall)
+}
+
+func (o *Local_cluster_import_operations) GetNamespace(name string) (*v1.Namespace, error) {
+	nsList := &v1.NamespaceList{}
+	err := o.apiReader.List(o.context, nsList)
```
filanov, line 61: isn't there a get to a namespace? why do you need to do a list?

paul-maidment: I found that a call to get the namespace did not return the namespace

pkg/localclusterimport/local_cluster_import_operations.go
paul-maidment: without any logic -> I don't think this is true

At the very least, management of how the responses are marshaled into various objects is being handled here as well as decisions around how the object should be fetched. 

IMO this is a question of where the functions live, if we were to merge the code, we would just be moving the functions from one file to the other. 

I think we would lose the neatness of mocks and it would be harder to maintain.

pkg/localclusterimport/local_cluster_import_operations.go
paul-maidment: Good point, will fix

cmd/main.go
```diff
@@ -641,6 +642,22 @@ func main() {
 		}
 	}()
 
+	if Options.EnableKubeAPI {
+		// Splitting into API reader and writer interfaces as the reader bypasses the cache settings that are on the regular client
+		// and we need to be able to read secrets that we cannot read with the client due to how the cache is configured.
+		// The cachedApiClient can be used for writes as these are unaffected by cache.
+		// See: https://redhat-internal.slack.com/archives/C3VS0LV41/p1641919541042900?thread_ts=1641573529.018300&cid=C3VS0LV41
```
filanov, line 649: i don't think it's a good idea to reference to internal slack in the code

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,302 @@
...
+			},
+		},
+	}
+	clusterDeployment.Name = LocalClusterNamespace + "-cluster-deployment"
+	clusterDeployment.Namespace = LocalClusterNamespace
+	clusterDeployment.Spec.ClusterName = LocalClusterNamespace + "-cluster-deployment"
+	clusterDeployment.Spec.BaseDomain = dns.Spec.BaseDomain
+	//
+	// Adding this ownership reference to ensure we can submit clusterDeployment without ManagedClusterSet/join permission
+	// See: https://redhat-internal.slack.com/archives/C01FT9E4Q10/p1645603511153759
```
filanov, line 286: suggesting not to reference slack in commits.

PR #4996, opened 2023-02-21
MGMT-13374: support Nutanix in kube-api

danielerez:
Added 'Nutanix' to PlatformType enum which is used for platform selection in the AgentClusterInstall CR (via `spec.platformType`). Note: validation is already handled on [RegisterCluster](https://github.com/openshift/assisted-service/blob/6808af2917cad45dcb370e0ae3c220b4f0f95325/internal/bminventory/inventory.go#L420).

I.e.
```
kind: AgentClusterInstall
spec:
   platformType: Nutanix
```

## List all the issues related to this PR

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [x] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #5438, opened 2023-08-22
AGENT-694: Support external platform type for agent installer

pawanpinjarkar:
Add k8s API changes to support a new platform type `external` for agent based installer.
- Min openshift version 4.14 is required
- UserManagedNetworking must be true
- When platform `type` is set `external`, `external.platformName` must be set to `oci`
e.g. 
```
type: external
external:
  platformName: oci 
```

Ref: https://github.com/openshift/installer/pull/7442

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
Tested locally by incorporating the changes in the local installer and generating agent ISO by providing install-config.yaml with platform external. Cluster deployment is yet to be tested on OCI.
- [ ] No tests needed

## Checklist

## Reviewers Checklist


api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -387,6 +387,9 @@ const (
 
 	// NutanixPlatformType represents Nutanix infrastructure.
 	NutanixPlatformType PlatformType = "Nutanix"
+
+	// ExternalPlatformType represents external cloud provider infrastructure.
+	ExternalPlatformType PlatformType = "External"
```
adriengentil, line 392: line 375 should be updated as well, no ?

carbonin: What are the problems with adding it here?

carbonin: Also do we need a [provider implementation](https://github.com/openshift/assisted-service/tree/ba04805cce13bf9b32647539c33ee63f85831586/internal/provider) for this?

mhrivnak: @adriengentil did we add external platform support to assisted-service without adding it to the k8s API? Maybe I'm not understanding the purpose of this PR.

pawanpinjarkar: @mhrivnak This PR is basically to help with the supported platform types for agent installer Ref https://github.com/openshift/installer/pull/7442/files#diff-fbe7c3bdbc1d6bab9c543d30777d3955cd42872e4eb9c4f592f6137a6da5472dR40

pawanpinjarkar: > What are the problems with adding it here?
None 🙂 For some reason, I checked https://github.com/openshift/hive/tree/master/apis/hive/v1 and questioned about it initially.

mhrivnak: @pawanpinjarkar I get that, but I'm not clear on why this specific change is what's needed in order to "help with the supported platform types for agent installer".

Making some assumptions, I suppose that agent-installer is using the k8s API structures for input, and assisted-service may have neglected to add external platform support to its k8s API? If that's correct, then presumably more code is required in assisted-service to enable this in the k8s API than just adding another enum value.

Or did assisted-service enable it differently in the k8s API?

adriengentil: I did nothing related to the k8s API, I discussed with @pawanpinjarkar and from what I understand we'll need to implement a new platform type "external" in AI (through a new provider, as @carbonin said), then we'll be able to plumb it in the k8s API.

I'll try to hack something today, it shouldn't be very complex as most of the pieces are already in place.

filanov: I'm not sure that this change is enough, looking at how nutanix support was added https://github.com/openshift/assisted-service/pull/4996 
you will need to generate the CRs post api changes. 
Because it's in the api it will mean that k8s api need to support that and lets not forget that we are GA so if someone will use that they will expect it to work. 
In the service is there a support for external platform? i suspect that not, based on https://github.com/openshift/assisted-service/blob/8534d1ece9dbd935a40436b69c1dfd5322ff9a09/swagger.yaml#L6221-L6228

pawanpinjarkar: > I did nothing related to the k8s API, I discussed with @pawanpinjarkar and from what I understand we'll need to implement a new platform type "external" in AI (through a new provider, as @carbonin said), then we'll be able to plumb it in the k8s API.
> 
> I'll try to hack something today, it shouldn't be very complex as most of the pieces are already in place.

Regarding the external platform, as assisted service expects the platformName to be `oci`, we in the agent installer, for 4.14, will add validation at the install-config level to only set `platformName` to `oci` if platform is `external` that way, the generic implementation of the platformName can be deferred until 4.15

pawanpinjarkar: e.g. https://github.com/openshift/installer/pull/7442/files#diff-0becd9dba8c3961ae565ce49accf81a33fdd6e513de100c0cc9ac7f55650ddc9R3

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -702,6 +702,10 @@ func getPlatform(openshiftPlatformType hiveext.PlatformType) *models.Platform {
 		return &models.Platform{
 			Type: common.PlatformTypePtr(models.PlatformTypeBaremetal),
 		}
+	case hiveext.ExternalPlatformType:
+		return &models.Platform{
+			Type: common.PlatformTypePtr(models.PlatformTypeOci),
```
zaneb, line 707: In the long term I think we'll probably want to add something like an `externalPlatformName` field so that we can conditionally return a different platform here. That will require us to add a vanilla 'external' platform type to core assisted. (The difference between `oci` and `external` would be what validations are run.)

The question for me is what to do in the short term - there are several options:

1. Add "External" to the k8s API as proposed here. A downside of this is that on this line we default to PlatformTypeOci, but once there is an `externalPlatformName` field then oci will not be the default. So we are creating a future backward-compat problem.
2. Add "OCI" to the k8s API instead. A downside of this is that ZTP does not and probably never will deploy to OCI, so it's a little weird for this to appear as a first-class option in the k8s API.
3. Go ahead and add the vanilla 'external' platform and the `externalPlatformName` field now. A downside of this is that it will take time that we don't really have.

Is there another option I have missed?

subsystem/kubeapi_test.go
```diff
@@ -1229,6 +1255,86 @@ var _ = Describe("[kube-api]cluster installation", func() {
...
+				agent.Spec.Approved = true
+				return kubeClient.Update(ctx, agent)
+			}, "30s", "10s").Should(BeNil())
+		}
+		installkey := types.NamespacedName{
+			Namespace: Options.Namespace,
+			Name:      clusterDeploymentSpec.ClusterInstallRef.Name,
+		}
+		By("verify external platform type status and spec")
+		checkPlatformStatus(ctx, installkey, hiveext.ExternalPlatformType, hiveext.ExternalPlatformType, swag.Bool(false))
```
danielerez, line 1319: Seems like the subsystem test fails here. Is UserManagedNetworking enabled for the cluster? if so, it should be 'true'

internal/controller/controllers/clusterdeployments_controller.go
pawanpinjarkar: @adriengentil said he is going to explore the 3rd option, which is best technically, provided it doesn't take a lot of time.

api/hiveextension/v1beta1/agentclusterinstall_types.go
mhrivnak: We're going to have a huddle on the assisted-service side to make sure we're providing the right "generic" external platform API for you to use. I know it's time-sensitive, but due to PTOs, it'll be Monday before we can have a full conversation. But I wanted to make sure you're in the loop on that intent. @adriengentil is working on it in the meantime as well.

internal/controller/controllers/clusterdeployments_controller.go
adriengentil: Playing a bit here: https://github.com/openshift/assisted-service/pull/5443

api/hiveextension/v1beta1/agentclusterinstall_types.go
pawanpinjarkar: @zaneb Once the generic external platform is available, should the agent installer then just accept `external` or we can also accept `oci` along with `external`?

pawanpinjarkar: > We're going to have a huddle on the assisted-service side to make sure we're providing the right "generic" external platform API for you to use. I know it's time-sensitive, but due to PTOs, it'll be Monday before we can have a full conversation. But I wanted to make sure you're in the loop on that intent. @adriengentil is working on it in the meantime as well.

@mhrivnak Please keep us posted. As of today, I believe we have everything up for review from the agent installer side. With a new PR in assisted-image-service, we are now able to generate the minimal ISO as well.

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -317,7 +317,7 @@ type Networking struct {
 	NetworkType string `json:"networkType,omitempty"`
 
 	// UserManagedNetworking indicates if the networking is managed by the user.
-	// For single-node installations, set to true or leave empty.
+	// For single-node installations for none or external platfrom, set to true or leave empty.
```
andfasano, line 320:
```suggestion
	// For single-node installations (none or external platform), set to true or leave empty.
```

api/hiveextension/v1beta1/agentclusterinstall_types.go
pawanpinjarkar: Got the necessary info from slack thread discussions.

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -317,7 +321,7 @@ type Networking struct {
 	NetworkType string `json:"networkType,omitempty"`
 
 	// UserManagedNetworking indicates if the networking is managed by the user.
-	// For single-node installations, set to true or leave empty.
+	// For single-node installations (none or external platfrom), set to true or leave empty.
```
zaneb, line 324: typo: platfrom

I believe that userManagedNetworking actually needs to be true if _any_ of {SNO, platform none, platform external} are true.
There's also some code to handle the case where it's left empty that will eventually need to be updated to include the external platform type: https://github.com/openshift/assisted-service/blob/master/pkg/webhooks/hiveextension/v1beta1/agentclusterinstall_mutating_hook.go#L158-L190 (for now it's not used in the agent installer though).

zaneb (commented):
This looks ready to me, if everyone is happy with the names.

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -192,6 +192,10 @@ type AgentClusterInstallSpec struct {
 	// +optional
 	PlatformType PlatformType `json:"platformType,omitempty"`
 
+	// ExternalSpec is the spec for the external platform
+	// +optional
+	ExternalSpec ExternalSpec `json:"external,omitempty"`
```
adriengentil, line 197:
```suggestion
	ExternalSpec* ExternalSpec `json:"external,omitempty"`
```
I think it needs to be a pointer to work along with "omitempty"

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -372,9 +376,14 @@ const (
 )
 
 // PlatformType is a specific supported infrastructure provider.
-// +kubebuilder:validation:Enum="";BareMetal;None;VSphere;Nutanix
+// +kubebuilder:validation:Enum="";BareMetal;None;VSphere;Nutanix;External
 type PlatformType string
 
+// ExternalSpec contains a specific supported platform name.
+type ExternalSpec struct {
```
adriengentil, line 383:
```suggestion
// ExternalPlatformSpec stores configuration related to external platform.
type ExternalPlatformSpec struct {
```
I think it could be a bit more clear

adriengentil (commented):
Look good overall! Just 2 comments

api/hiveextension/v1beta1/agentclusterinstall_types.go
avishayt:
```suggestion
	ExternalSpec *ExternalSpec `json:"external,omitempty"`
```

avishayt: No need to guess - please just copy-paste
https://github.com/openshift/api/blob/edd317298780d562af7d5d979b237d14fc299785/config/v1/types_infrastructure.go#L311-L314

api/hiveextension/v1beta1/agentclusterinstall_types.go
avishayt: Copy-paste from here
https://github.com/openshift/api/blob/edd317298780d562af7d5d979b237d14fc299785/config/v1/types_infrastructure.go#L232-L241

pawanpinjarkar:
/uncc @zaneb  @andfasano  @mhrivnak  @carbonin @CrystalChun @danmanor @filanov @danielerez @eliorerz

avishayt (approved):

PR #5443, opened 2023-08-24
Introduce vanilla external platform

adriengentil:
- Introduce a new platform type for external, and its implementation as a provider
  - I kept the current behavior, `platformName` is not customizable by the user, and will be set to `external` in the instal-config 
- Introduce a new feature support level for it, behaves like `None` platform, the restrictions are:
  - Platform Managed Networking required (seems the new name for UMN?)
  - OCP 4.14 or greater is required
- We may want to put this new feature behind a capability for some time


swagger.yaml
```diff
@@ -3584,7 +3584,7 @@ paths:
           name: platform_type
           description: The provider platform type.
           type: string
-          enum: [ 'baremetal', 'none', 'nutanix', 'vsphere', 'oci' ]
+          enum: [ 'baremetal', 'none', 'nutanix', 'vsphere', 'oci', 'external' ]
```
avishayt, line 3587: Let's remove oci

swagger.yaml
```diff
@@ -4086,6 +4086,7 @@ definitions:
       - 'EXTERNAL_PLATFORM_OCI'
       - 'DUAL_STACK'
       - 'PLATFORM_MANAGED_NETWORKING'
+      - 'EXTERNAL_PLATFORM'
```
avishayt, line 4089: Remove EXTERNAL_PLATFORM_OCI above

swagger.yaml
```diff
@@ -6226,6 +6227,7 @@ definitions:
       - vsphere
       - none
       - oci
+      - external
```
avishayt, line 6230: Remove oci above

swagger.yaml
adriengentil: don't we want to remove it once the UI has migrated to external+platform name?

avishayt: Yes

adriengentil:
/close
I'll re-open once we start to work on https://issues.redhat.com/browse/MGMT-15654

PR #5445, opened 2023-08-25
MGMT-15559: Change detached annotation condition in non-converged flow

CrystalChun:
[MGMT-15559](https://issues.redhat.com/browse/MGMT-15559)
Day 2 workers create BMH and Machine CRs on the spoke cluster when the host starts installing. The non-converged flow initially added the detached annotation for the BMH when the host starts installing too. This causes the BMH to stop being reconciled so the BMH and Machine CRs aren't created in the spoke cluster.

This change adds the detached annotation when the host reaches rebooting, joined, or failed instead of installing so that it doesn't conflict with adding the BMH/Machine to the spoke cluster.

[MGMT-15639](https://issues.redhat.com/browse/MGMT-15639)

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @ori-amizur

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -306,7 +306,7 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
 	} else {
 		// After the agent has started installation, Ironic should not manage the host.
 		// Adding the detached annotation to the BMH stops Ironic from managing it.
-		result = r.addBMHDetachedAnnotationIfAgentHasStartedInstallation(log, bmh, agent)
+		result = r.addBMHDetachedAnnotationIfHostIsRebooting(log, bmh, agent)
 	}
 	if res := r.handleReconcileResult(ctx, log, result, bmh); res != nil {
 		return res.Result()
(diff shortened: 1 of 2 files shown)


internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -553,27 +554,6 @@ func (r *BMACReconciler) ensureBMHDetached(log logrus.FieldLogger, bmh *bmh_v1al
 	return reconcileComplete{dirty: true, stop: true}
 }
 
-// The detached annotation is added if the installation of the agent associated with
```
ori-amizur, line 556: Actually I think it is better to keep they were before this change and just rename and change the function **addBMHDetachedAnnotationIfAgentHasStartedInstallation**.

ori-amizur (commented):
Is it possible to add tests to verify that the problem is fixed?

internal/controller/controllers/bmh_agent_controller.go
CrystalChun: Ok sounds good! I reverted it back but I'm not sure what the name of the function should be 😅 any ideas?

CrystalChun:
> Is it possible to add tests to verify that the problem is fixed?

Yes! Probably an e2e test that ensures that the day2 workers are installed and their BMHs exist in the release repo https://github.com/openshift/release/blob/master/ci-operator/step-registry/assisted/baremetal/operator/ztp/assisted-baremetal-operator-ztp-workflow.yaml#L9
Wdyt?

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: addBMHDetachedAnnotationIfHostIsRebooting ?

PR #5459, opened 2023-09-05
MGMT-15704  Assisted service should create Day2 import CR for hub cluster.

paul-maidment:
When using assisted-service in KubeAPI mode, we want to import the local cluster automatically, allowing it to be managed in a similar fashion to the spoke cluster.

The intent is to allow the user to perform day2 operations on the local cluster, to allow the addition of workers and so on.

Presently this is only possible via manual efforts and is not very customer friendly.

This PR aims to resolve this by adding functionality to import the local cluster as described above.

The behaviour during upgrade will be dealt with in another ticket.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (i *LocalClusterImport) shouldImportLocalCluster() (bool, error) {
+	// The presence of agentClusterInstall for the local cluster will determine whether or not we need to perform an import
+	agentClusterInstall, err := i.clusterImportOperations.GetAgentClusterInstall(LocalClusterNamespace, LocalClusterNamespace+"-cluster-install")
+	if err != nil && !k8serrors.IsNotFound(err) {
+		i.log.Errorf("hubClusterDay2Importer: There was an error fetching the AgentClusterInstall while checking for need to register local cluster.")
+		return false, err
+	}
+	if agentClusterInstall != nil {
```
filanov, line 36: please add a todo here with the relevant ticket to handle upgrade

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	shouldImportLocalCluster, err := i.shouldImportLocalCluster()
+	if err != nil {
+		return err
+	}
+	if !shouldImportLocalCluster {
+		message := "hubClusterDay2Importer: no need to import local cluster as registration already detected"
+		i.log.Info(message)
+		return errors.New(message)
```
filanov, line 57: no need to return an error in this case, nothing have failed

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	openshift_version := ""
+	release_image := ""
+	if clusterVersion != nil && clusterVersion.Status.History[0].State == configv1.CompletedUpdate {
```
filanov, line 61: is it always the first?

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,240 @@
...
+		openshift_version = clusterVersion.Status.History[0].Version
+		release_image = clusterVersion.Status.History[0].Image
+		i.log.Infof("hubClusterDay2Importer: Found openshift_version %s", openshift_version)
+	} else {
```
filanov, line 65: suggesting to change the order of the if
```go
if clusterVersion == nil || clusterVersion.Status.History[0].State != configv1.CompletedUpdate {
		message := "hubClusterDay2Importer: Cluster version info is empty, cannot proceed"
		i.log.Error(message)
		return errors.New(message)
}
```
then you will not need the else and the code is clearer that way.

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,240 @@
...
+		i.log.Error(message)
+		return errors.New(message)
+	}
+
+	kubeConfigSecret, err := i.clusterImportOperations.GetSecret("openshift-kube-apiserver", "node-kubeconfigs")
+	if err != nil {
+		i.log.Errorf("hubClusterDay2Importer: Unable to fetch local cluster kubeconfigs due to error %s", err.Error())
+		return err
+	}
+	if kubeConfigSecret != nil {
```
filanov, line 76: is there a case where you have no error and the kubeconfig is nil?

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	if pullSecret != nil {
+		i.log.Infof("hubClusterDay2Importer: Found secret %s", pullSecret.Name)
+	} else {
+		message := "hubClusterDay2Importer: Could not find pull secret for local cluster"
+		i.log.Error(message)
+		return errors.New(message)
+	}
+
+	dns, err := i.clusterImportOperations.GetDNS("cluster")
```
filanov, line 97: it's a bit strange to have function that gets a const

filanov (commented):
Please add in the commit message why did you had to add permissions to specific resources

General comment about if else 
usually it's cleaner to handle the error first, then you will not need the else in most cases

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,240 @@
...
+			ClusterInstallRef: &hivev1.ClusterInstallLocalReference{
+				Name:    LocalClusterNamespace + "-cluster-install",
+				Group:   "extensions.hive.openshift.io",
+				Kind:    "AgentClusterInstall",
+				Version: "v1beta1",
+			},
+			Platform: hivev1.Platform{
+				AgentBareMetal: &agent.BareMetalPlatform{
+					AgentSelector: metav1.LabelSelector{
+						MatchLabels: map[string]string{"infraenv": fmt.Sprintf("%s-day2-infraenv", LocalClusterNamespace)},
```
filanov, line 208: you can remove this part, you don't need to decide for the user what will be the infra env. 
you can write something like local cluster or something generic

pkg/localclusterimport/import_local_cluster.go
paul-maidment: Seems that way, yes, the State is the crucial thing, it will be `configv1.PartialUpdate` until the installation has completed.
But yes, I tested this assumption using `oc adm upgrade` -- the first history item is always the latest

pkg/localclusterimport/import_local_cluster.go
paul-maidment: This is easier to mock, the underlying operation can get any DNS entry. I'd rather keep this as is

cmd/main.go
```diff
@@ -205,6 +206,19 @@ func maxDuration(dur time.Duration, durations ...time.Duration) time.Duration {
...
+func importLocalCluster(ctrlMgr manager.Manager, log *logrus.Logger) {
+	// Splitting into API reader and writer interfaces as the reader bypasses the cache settings that are on the regular client
+	// and we need to be able to read secrets that we cannot read with the client due to how the cache is configured.
+	// The cachedApiClient can be used for writes as these are unaffected by cache.
+	localClusterImportOperations := localclusterimport.NewLocalClusterImportOperations(ctrlMgr.GetAPIReader(), ctrlMgr.GetClient(), log)
+	localClusterImport := localclusterimport.NewLocalClusterImport(&localClusterImportOperations, log)
+	err := localClusterImport.ImportLocalCluster()
+	if err != nil {
+		// Failure to import the local cluster is not fatal but we should warn in the log.
+		log.Warnf("Could not import local cluster into ACM due to error %s", err.Error())
```
nmagnezi, line 218:
```suggestion
		log.Warnf("Could not import local cluster due to error %s", err.Error())
```

nmagnezi: I don't it should mention ACM specifically:
1. It might not be relevant to some deployment types we support.
2. ACM is a downstream name.

cmd/main.go
```diff
@@ -641,6 +655,10 @@ func main() {
 		}
 	}()
 
+	if Options.EnableKubeAPI {
+		importLocalCluster(ctrlMgr, log)
```
nmagnezi, line 659: Is there no feature flag for this import, specifically?

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (i *LocalClusterImport) shouldImportLocalCluster() (bool, error) {
+	// The presence of agentClusterInstall for the local cluster will determine whether or not we need to perform an import
+	agentClusterInstall, err := i.clusterImportOperations.GetAgentClusterInstall(LocalClusterNamespace, LocalClusterNamespace+"-cluster-install")
+	if err != nil && !k8serrors.IsNotFound(err) {
+		i.log.Errorf("hubClusterDay2Importer: There was an error fetching the AgentClusterInstall while checking for need to register local cluster.")
+		return false, err
+	}
+	if agentClusterInstall != nil {
+		i.log.Infof("hubClusterDay2Importer: Found AgentClusterInstall for hub cluster, assuming that hub has been correctly registered for ZTP day 2 operations: %s", agentClusterInstall.Name)
```
nmagnezi, line 39: Are there any checks we can do here to ensure that cluster is actually healthy?
Also, why is ZTP mentioned here?

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,226 @@
...
+
+func NewLocalClusterImport(localClusterImportOperations ClusterImportOperations, log *logrus.Logger) LocalClusterImport {
+	return LocalClusterImport{clusterImportOperations: localClusterImportOperations, log: log}
+}
+
+func (i *LocalClusterImport) shouldImportLocalCluster() (bool, error) {
+	// The presence of agentClusterInstall for the local cluster will determine whether or not we need to perform an import
+	agentClusterInstall, err := i.clusterImportOperations.GetAgentClusterInstall(LocalClusterNamespace, LocalClusterNamespace+"-cluster-install")
+	if err != nil && !k8serrors.IsNotFound(err) {
+		i.log.Errorf("hubClusterDay2Importer: There was an error fetching the AgentClusterInstall while checking for need to register local cluster.")
```
nmagnezi, line 33: Why do all the errors have a `hubClusterDay2Importer` prefix?

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	shouldImportLocalCluster, err := i.shouldImportLocalCluster()
+	if err != nil {
+		return err
+	}
+	if !shouldImportLocalCluster {
+		message := "hubClusterDay2Importer: no need to import local cluster as registration already detected"
+		i.log.Info(message)
+		return errors.New(message)
```
nmagnezi: +1  , I was about to comment the same thing.
In line 37 you treated this as an Info log.

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	release_image = clusterVersion.Status.History[0].Image
+
+	kubeConfigSecret, err := i.clusterImportOperations.GetSecret("openshift-kube-apiserver", "node-kubeconfigs")
+	if err != nil {
+		i.log.Errorf("hubClusterDay2Importer: Unable to fetch local cluster kubeconfigs due to error %s", err.Error())
+		return err
+	}
+	i.log.Infof("hubClusterDay2Importer: Found secret %s", kubeConfigSecret.Name)
```
nmagnezi, line 75: 1. This should probably be a debug.
2. Please specify the details of the related cluster and/or also the secret namespace - to make it more specific.

pkg/localclusterimport/import_local_cluster.go
paul-maidment: Done

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,228 @@
...
+	// TODO: We need to determine what to do if there has been a recent upgrade
+	// See MGMT-15705 for more information.
+	if agentClusterInstall != nil {
+		i.log.Infof("hubClusterDay2Importer: Found AgentClusterInstall for hub cluster, assuming that hub has been correctly registered for ZTP day 2 operations: %s", agentClusterInstall.Name)
+		return false, nil
+	}
+	return true, nil
+}
+
+func (i *LocalClusterImport) ImportLocalCluster() error {
```
nmagnezi, line 45: Few things about this function:
1. It does to many things, which should be split to sub functions for both readability and tests.
2. It can potentially fail in too many places (which is related to the first point, in part). I also think that for many things here it is worth to use multiErr to aggregate a much of the problems and reflect them to the user as early as possible.

nmagnezi (changes requested):

cmd/main.go
paul-maidment: What is the correct terminology for what we are doing here?

pkg/localclusterimport/import_local_cluster.go
filanov: can you install older ocp and upgrade to make sure it will remain the first one?

pkg/localclusterimport/import_local_cluster.go
filanov: maybe make it GetClusterDNS it will be clearer and you will not need to pass this parameter

cmd/main.go
filanov: +1 we probably need to have something that can disable this feature if needed

pkg/localclusterimport/import_local_cluster.go
paul-maidment: I'm not sure that this is possible at this stage.

The one thing we do perform is a check of the cluster-version history, which will show up as completed if the last update was applied.

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,226 @@
...
+	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/openshift/hive/apis/hive/v1/agent"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const LocalClusterNamespace = "local-cluster"
```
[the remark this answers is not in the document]
paul-maidment: I'm not 100% sure that this is in line with the goals of the feature. The main goal I see is that we want to simplify the process of management of the hub cluster by creating some of the CRs the user needs. 

Maybe we could accept an env var optionally and default to "local-cluster"

Also note that "local-cluster" is the name of the cluster that is created in multi cluster hub.

cmd/main.go
paul-maidment: After discussion with @filanov we have settled on "import local cluster" as the terminology to be used.

pkg/localclusterimport/import_local_cluster.go
paul-maidment: I did this to make things more "greppable" but if it's non standard or weird, I can remove it?
I think that if you are trying to debug a specific piece of functionality, having some text to grep helps with filtering
that's the main reason I did this

pkg/localclusterimport/import_local_cluster.go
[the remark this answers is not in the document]
filanov: the only option to set env vars at the moment is with the unsupported configmap.
i don't think that it will be wise to add it as an official api at the moment, lets get a feedback on that in the next release and if needed we  will add it as an option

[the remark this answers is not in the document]
filanov: but env var with a default value can be a good option even if it is set via unsupported configmap

cmd/main.go
filanov: +1 to what Nir suggested, we have both ACM and MCE, so his suggestion cover both

pkg/localclusterimport/import_local_cluster.go
paul-maidment: Removing the prefix

pkg/localclusterimport/import_local_cluster.go
paul-maidment: already fixed

pkg/localclusterimport/import_local_cluster.go
paul-maidment: Adding a feature flag to options

cmd/main.go
paul-maidment: adding a flag to allow feature to be disabled.

paul-maidment:
> Using ZTP in the commit message is incorrect, here you are referring to kube-api or the operator, have nothing to do with ZTP

Fixed now

> Using ZTP in the commit message is incorrect, here you are referring to kube-api or the operator, have nothing to do with ZTP

Fixed

pkg/localclusterimport/import_local_cluster.go
paul-maidment: I will try this now to verify

pkg/localclusterimport/import_local_cluster.go
paul-maidment: 1: Done
2: Done

pkg/localclusterimport/import_local_cluster.go
filanov: i'm not sure which tests can be done in this case, domain resolution and valid kubeconfig, but if you are taking them from the cluster and even if they are invalid, what can be done?

cmd/main.go
```diff
@@ -205,6 +208,23 @@ func maxDuration(dur time.Duration, durations ...time.Duration) time.Duration {
 	return ret
 }
 
+func importLocalCluster(ctrlMgr manager.Manager, log *logrus.Logger) {
+	if !Options.EnableLocalClusterImport {
+		log.Debug("EnableLocalClusterImport disabled in options, skipping...")
```
filanov, line 213: can be info, it's not something periodic

cmd/main.go
```diff
@@ -162,6 +163,8 @@ var Options struct {
 	AllowConvergedFlow                   bool          `envconfig:"ALLOW_CONVERGED_FLOW" default:"true"`
 	PreprovisioningImageControllerConfig controllers.PreprovisioningImageControllerConfig
 	BMACConfig                           controllers.BMACConfig
+	EnableLocalClusterImport             bool   `envconfig:"ENABLE_LOCAL_CLUSTER_IMPORT" default:"true"`
+	LocalClusterImportNamespace          string `envconfig:"LOCAL_CLUSTER_IMPORT_NAMESPACE" defualt:"local-cluster"`
```
filanov, line 167: please make sure with ACM/MCE  that they are not using the same namespace

pkg/localclusterimport/local_cluster_import_operations.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	err := o.apiReader.Get(o.context, namespacedName, clusterImageSet)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Unable to fetch cluster image set %s", name)
+	}
+	return clusterImageSet, nil
+}
+
+func (o *LocalClusterImportOperations) CreateClusterImageSet(clusterImageSet *hivev1.ClusterImageSet) (*hivev1.ClusterImageSet, error) {
+	err := o.cachedApiClient.Create(o.context, clusterImageSet)
```
filanov, line 134: what will happen is it's already exists? 
same question to all the resources.

pkg/localclusterimport/import_local_cluster.go
paul-maidment: I can confirm that the latest version is always the [0]th entry

`        "history": [
            {
                "acceptedRisks": "Target release version=\"\" image=\"quay.io/openshift-release-dev/ocp-release:4.13.11-x86_64\" cannot be verified, but continuing anyway because the update was forced: release images that are not accessed via digest cannot be verified\nPrecondition \"ClusterVersionRecommendedUpdate\" failed because of \"UnknownUpdate\": RetrievedUpdates=True (), so the update from 4.13.9 to 4.13.11 is probably neither recommended nor supported.",
                "completionTime": null,
                "image": "quay.io/openshift-release-dev/ocp-release:4.13.11-x86_64",
                "startedTime": "2023-09-06T08:52:43Z",
                "state": "Partial",
                "verified": false,
                "version": "4.13.11"
            },
            {
                "completionTime": "2023-09-06T08:52:09Z",
                "image": "quay.io/openshift-release-dev/ocp-release@sha256:a266d3d65c433b460cdef7ab5d6531580f5391adbe85d9c475208a56452e4c0b",
                "startedTime": "2023-09-06T08:19:17Z",
                "state": "Completed",
                "verified": false,
                "version": "4.13.9"
            }
        ],
`

pkg/localclusterimport/local_cluster_import_operations.go
paul-maidment: It would fail to write the cluster image set and the old one would remain in place. 
Not sure what the best thing to do is, on one hand, we could delete anything we find when performing this write, but seems to me that this would not be the smart choice until we have looked into the impact of the update ticket.

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,288 @@
...
+	if agentClusterInstall != nil {
+		i.log.Infof("found AgentClusterInstall for hub cluster, assuming that hub has been correctly registered for assisted day 2 operations: %s", agentClusterInstall.Name)
+		return false, nil
+	}
+	return true, nil
+}
+
+func (i *LocalClusterImport) ImportLocalCluster() error {
+
+	shouldImportLocalCluster, err := i.shouldImportLocalCluster()
```
filanov, line 195: As we talked, this function can be removed and you will try to create all the objects and if they are already exist you will ignore them, using `k8serrors.IsAlreadyExists(err)` to check the error

pkg/localclusterimport/local_cluster_import_operations.go
filanov: until we handle upgrade it should be good enough

pkg/localclusterimport/import_local_cluster.go
paul-maidment: Confirmed, latest is always the [0]th item in the history...

```
[
  {
    "acceptedRisks": "Target release version=\"\" image=\"quay.io/openshift-release-dev/ocp-release:4.14.0-ec.4-x86_64\" cannot be verified, but continuing anyway because the update was forced: release images that are not accessed via digest cannot be verified\nForced through blocking failures: Multiple precondition checks failed:\n* Precondition \"ClusterVersionUpgrade
able\" failed because of \"AdminAckRequired\": Kubernetes 1.27 and therefore OpenShift 4.14 remove several APIs which require admin consideration. Please see the knowledge article https://access.redhat.com/articles/6958395 for details and instructions.\n* Precondition \"EtcdRecentBackup\" failed because of \"UpgradeBackupInProgress\": RecentBackup: Backup pod phase: \"P
ending\"\n* Precondition \"ClusterVersionRecommendedUpdate\" failed because of \"UnknownUpdate\": RetrievedUpdates=False (VersionNotFound), so the recommended status of updating from 4.13.11 to 4.14.0-ec.4 is unknown.",
    "completionTime": null,
    "image": "quay.io/openshift-release-dev/ocp-release:4.14.0-ec.4-x86_64",
    "startedTime": "2023-09-06T12:37:19Z",
    "state": "Partial",
    "verified": false,
    "version": "4.14.0-ec.4"
  },
  {
    "acceptedRisks": "Target release version=\"\" image=\"quay.io/openshift-release-dev/ocp-release:4.13.11-x86_64\" cannot be verified, but continuing anyway because the update was forced: release images that are not accessed via digest cannot be verified\nPrecondition \"ClusterVersionRecommendedUpdate\" failed because of \"UnknownUpdate\": RetrievedUpdates=True (), so
 the update from 4.13.9 to 4.13.11 is probably neither recommended nor supported.",
    "completionTime": "2023-09-06T10:06:05Z",
    "image": "quay.io/openshift-release-dev/ocp-release:4.13.11-x86_64",
    "startedTime": "2023-09-06T08:52:43Z",
    "state": "Completed",
    "verified": false,
    "version": "4.13.11"
  },
  {
    "completionTime": "2023-09-06T08:52:09Z",
    "image": "quay.io/openshift-release-dev/ocp-release@sha256:a266d3d65c433b460cdef7ab5d6531580f5391adbe85d9c475208a56452e4c0b",
    "startedTime": "2023-09-06T08:19:17Z",
    "state": "Completed",
    "verified": false,
    "version": "4.13.9"
  }
]

```

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,276 @@
...
+	agentClusterInstall.Namespace = i.localClusterNamespace
+	agentClusterInstall.Name = i.localClusterNamespace + "-cluster-install"
+	err := i.clusterImportOperations.CreateAgentClusterInstall(agentClusterInstall)
+	if err != nil {
+		i.log.Errorf("could not create AgentClusterInstall due to error %s", err.Error())
+		return nil, err
+	}
+	// Fetch the recently stored AgentClusterInstall so that we can obtain the UID
+	aci, err := i.clusterImportOperations.GetAgentClusterInstall(i.localClusterNamespace, i.localClusterNamespace+"-cluster-install")
+	if err != nil && !k8serrors.IsAlreadyExists(err) {
```
filanov, line 115: get cannot return already exist

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,276 @@
...
+
+	release_image = clusterVersion.Status.History[0].Image
+
+	err = i.createClusterImageSet(release_image)
+	if err != nil {
+		errorList = multierror.Append(errorList, err)
+	}
+
+	err = i.createNamespace(i.localClusterNamespace)
+	if err != nil && !k8serrors.IsAlreadyExists(err) {
```
filanov, line 241: minor comment - you already handling already exist error inside `createNamespace`

pkg/localclusterimport/import_local_cluster.go
```diff
@@ -0,0 +1,276 @@
...
+		errorList = multierror.Append(errorList, err)
+	}
+
+	if kubeConfigSecret != nil {
+		err = i.createAdminKubeConfig(kubeConfigSecret)
+		if err != nil && !k8serrors.IsAlreadyExists(err) {
```
filanov, line 247: minor comment - you already handling already exist error inside `createAdminKubeConfig`

pkg/localclusterimport/import_local_cluster.go
paul-maidment: I actually changed this, I need to be able to mock this so I moved the handling of this stuff into import_local_cluster

pkg/localclusterimport/import_local_cluster.go
paul-maidment: I actually changed this, I need to be able to mock this so I moved the handling of this stuff into import_local_cluster

filanov (approved):

PR #5471, opened 2023-09-08
[cloud_hotfix_releases] MGMT-15715: Allow handling multi-document YAML custom manifests

adriengentil:
Multi document yamls are not supported during installation.
    
 This change looks into the custom manifests files uploaded by the user,
 and splits multi document yaml files into several ones.

### DIFF
--- a/internal/ignition/ignition.go
+++ b/internal/ignition/ignition.go
@@ -26,6 +26,7 @@ import (
 	"github.com/coreos/vcontext/report"
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
+	"github.com/google/uuid"
 	bmh_v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	clusterPkg "github.com/openshift/assisted-service/internal/cluster"
 	"github.com/openshift/assisted-service/internal/common"
@@ -409,6 +410,12 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte,
 		}
 	}
 
+	err = g.expandUserMultiDocYamls(ctx)
+	if err != nil {
+		log.WithError(err).Errorf("failed expand multi-document yaml for cluster '%s'", g.cluster.ID)
+		return err
+	}
+
 	err = g.applyManifestPatches(ctx)
 	if err != nil {
 		log.WithError(err).Errorf("failed to apply manifests' patches for cluster '%s'", g.cluster.ID)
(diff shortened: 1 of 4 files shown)


adriengentil:
/hold
I would like to test locally first

/cc @gamli75 @osherdp

internal/manifests/manifests.go
```diff
@@ -450,6 +477,24 @@ func (m *Manifests) validateUserSuppliedManifest(ctx context.Context, clusterID
 	return nil
 }
 
+// isValidYaml checks if one or more (in case of multi-doc yaml) yaml documents are valid
```
paul-maidment, line 480: Might be better to say that this checks if all yaml docs present in manifestContent are valid, that this may be more than one document in the case of multi-doc yaml

```
// isValidYaml checks if all yaml documents are valid, in the case of multi-doc yaml this may be more than one document.
```

internal/manifests/manifests_test.go
```diff
@@ -391,7 +391,7 @@ spec:
 			It("fails for manifest with invalid yaml format", func() {
 				clusterID := registerCluster().ID
 				fileName := "99-test.yml"
-				invalidYAMLContent := encodeToBase64("not a valid YAML content")
+				invalidYAMLContent := encodeToBase64("not a valid YAML: {{ content }}")
```
paul-maidment, line 394: I don't really understand why this was changed, in the original code the intention was to have the phrase "not a valid YAML content", I don't think `content` is to be treated as a variable in this case.

Note the phrase "not a valid YAML content" doesn't flow well in English grammar and I think it should say "invalid YAML content" instead, but that's a separate issue.

internal/manifests/manifests_test.go
```diff
@@ -473,6 +473,51 @@ spec:
...
+				Expect(responsePayload.Payload.Folder).To(Equal(defaultFolder))
+			})
+
+			It("Upload fails when one of the yaml in muti-doc yaml file is invalid", func() {
+				clusterID := registerCluster().ID
+				fileName := "99-test.yml"
+				content := encodeToBase64(`---
+first: one
+---
+not a valid YAML: {{ content }}
```
paul-maidment, line 507: I think the {{ content }} here just adds confusion the simple phrase "invalid YAML content" would suffice as such content does not even form a key/value pair with is the minimum requirement for any YAML content (as demonstrated by `first: one`

internal/manifests/manifests_test.go
```diff
@@ -949,6 +994,21 @@ spec:
...
 		})
 	})
+
+	Context("GetUserManifestSuffixes", func() {
+		It("Should be able to list only user manifests if user and non user manifests are present", func() {
+			clusterId := registerCluster().ID
+			s3Metadata := []string{
+				filepath.Join(clusterId.String(), constants.ManifestMetadataFolder, "manifests", "first.yaml", constants.ManifestSourceUserSupplied),
+				filepath.Join(clusterId.String(), constants.ManifestMetadataFolder, "openshift", "second.yaml", constants.ManifestSourceUserSupplied),
+				filepath.Join(clusterId.String(), constants.ManifestMetadataFolder, "manifests", "third.yaml", "other-metadata"),
```
paul-maidment, line 1004: I would not label this as "other-metadata", I would consider labeling "other-manifest-source" to make it clear that this is not just some arbitrary metadata field.

internal/manifests/manifests_test.go
```diff
@@ -473,6 +473,51 @@ spec:
 				Expect(err.Error()).To(ContainSubstring("Manifest content of file 99-openshift-machineconfig-master-kargs.json for cluster ID " + clusterID.String() + " exceeds the maximum file size of 1MiB"))
 			})
 
+			It("accept muti-doc yaml file", func() {
```
paul-maidment, line 476: "upload succeeds when each yaml in multi-doc yaml file is valid"

internal/ignition/ignition.go
osherdp, line 1: Just making sure here,
we'll have this kind of change for master branch as well?

internal/manifests/manifests.go
```diff
@@ -450,6 +477,24 @@ func (m *Manifests) validateUserSuppliedManifest(ctx context.Context, clusterID
...
+	dec := yaml.NewDecoder(bytes.NewReader(manifestContent))
+
+	for {
+		var doc interface{}
+		err := dec.Decode(&doc)
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if err != nil {
+			return false
```
paul-maidment, line 491: are we not interested in what the error was, should we return it, should we log it?

internal/manifests/manifests.go
```diff
@@ -355,6 +361,29 @@ func GetClusterManifests(ctx context.Context, clusterID *strfmt.UUID, s3Client s
 	return manifestFiles, nil
 }
 
+// Return the list of the manifests provided by the user into a list of "{folder}/{filename}"
+func GetUserManifestSuffixes(ctx context.Context, clusterID *strfmt.UUID, s3Client s3wrapper.API) ([]string, error) {
```
paul-maidment, line 365: I am not sure I would call this `GetUserManifestSuffixes`, 

as far as I can see you are expecting an entry like 

```
filepath.Join(clusterId.String(), constants.ManifestMetadataFolder, "manifests", "first.yaml", constants.ManifestSourceUserSupplied)
``` 

to resolve to 

```
"manifests"/"first.yaml"
```

Seems that this is more about resolving the manifest names independent of any cluster id

internal/manifests/manifests_test.go
adriengentil:
```
"not a valid YAML content"
```
parses successfully into a string

> I don't think content is to be treated as a variable in this case.

the issue is with the curly braces, which makes an invalid way of defining a dictionary. Maybe I can update the string to `invalid YAML content: {` to make the "mistake" obvious.

internal/manifests/manifests_test.go
adriengentil: see my previous comment,  "invalid YAML content" is a valid yaml string

internal/manifests/manifests.go
adriengentil: We were not handling the error previously, I did not change that.

I don't think we should log anything here, it's expected that sometimes users will upload invalid yamls.

wdyt?

internal/manifests/manifests.go
adriengentil: thanks, I think I will split this into 2 functions:
- GetManifestMetadata =>  returns full path of metadata
- ResolveManifestNamesFromMetadata => which takes full path of metadata and just extract folder and manifest names

internal/manifests/manifests.go
eliorerz: I think @paul-maidment is right, we should return the error and add it when logging the error

adriengentil: I now return the error to the user

adriengentil:
@paul-maidment can you have another look? Thanks!

paul-maidment (approved):
/lgtm

adriengentil:
@osherdp @paul-maidment here is the PR on master: https://github.com/openshift/assisted-service/pull/5480

adriengentil:
/override ci/prow/e2e-agent-compact-ipv4
it was not supposed to run :thinking:

PR #5475, opened 2023-09-11
[release-ocm-2.9] MGMT-15501: ClusterServiceVersion for release 2.9 branch references `latest` tag

eliorerz:
/cc @adriengentil @danielerez @osherdp


osherdp (approved):

osherdp:
@eliorerz @adriengentil I think ``e2e-agent-compact-ipv4`` is only relevant to ABI (i.e. openshift versions). So we should remove it from this branch

adriengentil:
> @eliorerz @adriengentil I think e2e-agent-compact-ipv4 is only relevant to ABI (i.e. openshift versions). So we should remove it from this branch

it's not defined in the config, it was triggered because this PR was initially opened against master and then was changed to target release-ocm-2.9.

PR #5505, opened 2023-09-20
MGMT-15771: Allow to keep vSphere with UMN installation

eliorerz:
Change UMN and CMN feature support features to ignore `cluster.user_managed_networking` as a preparation for removing `user_managed_networking`

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

/cc @eliorerz


avishayt:
As we previously discussed, I recommend adding a new platform for vsphere-umn to make it explicit for the user.

eliorerz:
> As we previously discussed, I recommend adding a new platform for vsphere-umn to make it explicit for the user.

@avishayt We eventually decided against implementing it by adding an `assisted` unique platform. 
Personally, I believe we shouldn't add a "new platform type" to `assisted` just to mark enabled UMN for vSphere. In the future, this approach might result in duplicated platforms for each platform supporting both CMN and UMN. If clarity for the user is the goal, we can either make the cluster UMN property read-only or integrate it into the platform structure, similar to how we marked external platforms.

avishayt:
> > As we previously discussed, I recommend adding a new platform for vsphere-umn to make it explicit for the user.
> 
> @avishayt We eventually decided against implementing it by adding an `assisted` unique platform. Personally, I believe we shouldn't add a "new platform type" to `assisted` just to mark enabled UMN for vSphere. In the future, this approach might result in duplicated platforms for each platform supporting both CMN and UMN. If clarity for the user is the goal, we can either make the cluster UMN property read-only or integrate it into the platform structure, similar to how we marked external platforms.

So how does a user say that they want vsphere UMN vs CMN?

eliorerz:
> > > As we previously discussed, I recommend adding a new platform for vsphere-umn to make it explicit for the user.
> > 
> > 
> > @avishayt We eventually decided against implementing it by adding an `assisted` unique platform. Personally, I believe we shouldn't add a "new platform type" to `assisted` just to mark enabled UMN for vSphere. In the future, this approach might result in duplicated platforms for each platform supporting both CMN and UMN. If clarity for the user is the goal, we can either make the cluster UMN property read-only or integrate it into the platform structure, similar to how we marked external platforms.
> 
> So how does a user say that they want vsphere UMN vs CMN?

Set or unset the VIPs

PR #5521, opened 2023-10-01
MGMT-15878: Ensure that hosts emit event showing why preparation failed.

paul-maidment:
When preparation fails for a host, we do not implement any kind of a timeout for the host to indicate that this has occurred.

This means that it is sometimes not possible for the user to determine the cause of a cluster timeout (which will inevitably be caused by the timeout of a host during preparation.)

Presently, there are two ways in which a host may time out (no result received within the cluster timeout)

- An inconclusive result from the pulling of cluster images
- An inconclusive result from the disk speed check

This PR introduces a timeout to detect these scenarios and report on them in a host timeout event so that the user may have a clue as to what has happened. 

This PR is in addition to [MGMT-15814](https://issues.redhat.com//browse/MGMT-15814) which introduces a cluster condition to track a cluster timeout when there is a failure to configure the preparation of a cluster within a given time frame (for example if the assisted pod crashes)

Together these PR's should improve the overall quality of error reporting.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/statemachine.go
+++ b/internal/host/statemachine.go
@@ -554,6 +554,20 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th TransitionHandler) stat
 		SourceStates: []stateswitch.State{
 			stateswitch.State(models.HostStatusPreparingForInstallation),
 		},
+		Condition:        stateswitch.And(If(IsConnected), If(IsMediaConnected), th.IsPreparingTimedOut, stateswitch.Or(installationDiskSpeedUnknown, imagesAvailabilityUnknown), allConditionsSuccessfulOrUnknown),
+		DestinationState: stateswitch.State(models.HostStatusPreparingFailed),
+		PostTransition:   th.PostHostPreparationTimeout(),
+		Documentation: stateswitch.TransitionRuleDoc{
+			Name:        "Preparing timed out host move to known",
+			Description: "TODO: Document this transition rule",
+		},
+	})
+
+	sm.AddTransitionRule(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeRefresh,
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusPreparingForInstallation),
+		},
 		Condition:        stateswitch.And(If(IsConnected), If(IsMediaConnected), stateswitch.Not(If(SucessfullOrUnknownContainerImagesAvailability))),
 		DestinationState: stateswitch.State(models.HostStatusPreparingFailed),
 		PostTransition:   th.PostRefreshHost(statusInfoHostPreparationFailure),
--- a/internal/host/transition.go
+++ b/internal/host/transition.go
@@ -60,6 +60,8 @@ type TransitionHandler interface {
 	PostRegisterHost(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error
 	PostResettingPendingUserAction(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error
 	PostUnbindHost(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error
+	IsPreparingTimedOut(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) (bool, error)
+	PostHostPreparationTimeout() stateswitch.PostTransition
 }
 
 var resetLogsField = []interface{}{"logs_info", "", "logs_started_at", strfmt.DateTime(time.Time{}), "logs_collected_at", strfmt.DateTime(time.Time{})}
@@ -665,6 +667,37 @@ func (th *transitionHandler) HasInstallationInProgressTimedOut(sw stateswitch.St
 	return time.Since(time.Time(sHost.host.Progress.StageUpdatedAt)) > maxDuration, nil
 }
 
+func (th *transitionHandler) PostHostPreparationTimeout() stateswitch.PostTransition {
+	ret := func(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error {
+		sHost, ok := sw.(*stateHost)
+		if !ok {
+			return errors.New("PostHostPreparationTimeout incompatible type of StateSwitch")
+		}
+		params, ok := args.(*TransitionArgsRefreshHost)
+		if !ok {
+			return errors.New("PostRefreshHost invalid argument")
+		}
+		var (
+			err error
+		)
+		failingConditons := []string{}
+		if !params.conditions["installation-disk-speed-check-successful"] {
+			failingConditons = append(failingConditons, statusInfoPreparationTimeoutDiskSpeed)
+		}
+		if !params.conditions["successful-container-image-availability"] {
+			failingConditons = append(failingConditons, statusInfoPreparationTimeoutImageAvailability)
+		}
+		statusInfo := strings.Replace(statusInfoPreparationTimeout, "$FAILING_CONDITIONS", strings.Join(failingConditons, "\n"), 1)
+		if sHost.srcState != swag.StringValue(sHost.host.Status) || swag.StringValue(sHost.host.StatusInfo) != statusInfo {
+			_, err = hostutil.UpdateHostStatus(params.ctx, logutil.FromContext(params.ctx, th.log), params.db,
+				th.eventsHandler, sHost.host.InfraEnvID, *sHost.host.ID,
+				sHost.srcState, swag.StringValue(sHost.host.Status), statusInfo)
+		}
+		return err
+	}
+	return ret
+}
+
 // Return a post transition function with a constant reason
 func (th *transitionHandler) PostRefreshHost(reason string) stateswitch.PostTransition {
 	ret := func(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) error {
@@ -790,4 +823,16 @@ func (th *transitionHandler) PostRefreshHostRefreshStageUpdateTime(
 		*sHost.host.ID,
 		sHost.srcState)
 	return err
+}
+
+// check if prepare for installation reach to timeout
+func (th *transitionHandler) IsPreparingTimedOut(sw stateswitch.StateSwitch, args stateswitch.TransitionArgs) (bool, error) {
+	sHost, ok := sw.(*stateHost)
+	if !ok {
+		return false, errors.New("IsPreparingTimedOut incompatible type of StateSwitch")
+	}
+	if time.Since(time.Time(sHost.host.StatusUpdatedAt)) > th.config.PrepareConfig.PrepareForInstallationTimeout {
+		return true, nil
+	}
+	return false, nil
 }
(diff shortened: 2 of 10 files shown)


internal/cluster/transition.go
```diff
@@ -143,7 +144,7 @@ func (th *transitionHandler) PostPrepareForInstallation(sw stateswitch.StateSwit
 	if !ok {
 		return errors.New("PostPrepareForInstallation invalid argument")
 	}
-	extra := append(append(make([]interface{}, 0), "install_started_at", strfmt.DateTime(time.Now()), "installation_preparation_completion_status", ""), resetLogsField...)
+	extra := append(append(make([]interface{}, 0), "install_started_at", strfmt.DateTime(time.Now())), resetLogsField...)
```
avishayt, line 147: ?

paul-maidment: This is to prevent that field being wiped on restart of installation.
If this remains here, we cannot use this field as the basis of any condition that will persist across installation attempts.

internal/bminventory/inventory.go
```diff
@@ -1292,10 +1292,12 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params
 	// prepare cluster and hosts for installation
 	err = b.db.Transaction(func(tx *gorm.DB) error {
 		if err = b.clusterApi.PrepareForInstallation(ctx, cluster, tx); err != nil {
+			b.clusterApi.HandlePreInstallError(ctx, cluster, fmt.Errorf("failed to transition to installation preparation due to error:%w", err))
```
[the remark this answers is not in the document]
filanov: +1 i don't think that we have an issue in the sync part

paul-maidment:
cc @ori-amizur

internal/cluster/transition.go
```diff
@@ -524,6 +525,56 @@ func (th *transitionHandler) PostPreparingTimedOut(sw stateswitch.StateSwitch, a
...
+				for _, v := range vRes {
+					if v.Status == ValidationFailure {
+						issues = append(issues, fmt.Sprintf("validation with ID %s is failing for host %s", models.ClusterValidationID(v.ID), clusterHost.ID))
+					}
+				}
+			}
+		}
+
+		if clusterHost.Status != nil && *clusterHost.Status == models.HostStatusInsufficient {
+			// Indicates that the host has failed the disk speed check
```
avishayt, line 556: We already have this information from the sufficient-installation-disk-speed validation.

internal/cluster/transition.go
```diff
@@ -524,6 +525,56 @@ func (th *transitionHandler) PostPreparingTimedOut(sw stateswitch.StateSwitch, a
...
+			issues = append(issues, fmt.Sprintf("host %s has failed the installation disk speed check", clusterHost.ID))
+		}
+
+		if clusterHost.Status != nil && *clusterHost.Status == models.HostStatusPreparingFailed {
+			// Indicates that the host has failed the disk speed check
```
avishayt, line 561: Comment doesn't match code

internal/cluster/transition.go
```diff
@@ -524,6 +525,56 @@ func (th *transitionHandler) PostPreparingTimedOut(sw stateswitch.StateSwitch, a
...
+			issues = append(issues, fmt.Sprintf("host %s has failed to fetch images", clusterHost.ID))
```
avishayt, line 562: We already have this information from the container-images-available validation.

avishayt (changes requested):

internal/host/statemachine.go
```diff
@@ -549,6 +549,22 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th TransitionHandler) stat
...
 	})
 
+	sm.AddTransitionRule(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeRefresh,
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusPreparingForInstallation),
+			stateswitch.State(models.HostStatusPreparingFailed),
+			stateswitch.State(models.HostStatusKnown),
+		},
+		Condition:        stateswitch.And(If(IsConnected), If(IsMediaConnected), th.IsPreparingTimedOut, stateswitch.And(stateswitch.Or(installationDiskSpeedUnknown, imagesAvailabilityUnknown), allConditionsSuccessfulOrUnknown)),
```
ori-amizur, line 559: NIT: No need to put `And` in `And`

internal/host/statemachine.go
```diff
@@ -549,6 +549,22 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th TransitionHandler) stat
 		},
 	})
 
+	sm.AddTransitionRule(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeRefresh,
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusPreparingForInstallation),
+			stateswitch.State(models.HostStatusPreparingFailed),
```
ori-amizur, line 556: Why not only `HostStatusPreparingForInstallation` ?

internal/host/statemachine.go
```diff
@@ -549,6 +549,22 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th TransitionHandler) stat
...
+			stateswitch.State(models.HostStatusKnown),
+		},
+		Condition:        stateswitch.And(If(IsConnected), If(IsMediaConnected), th.IsPreparingTimedOut, stateswitch.And(stateswitch.Or(installationDiskSpeedUnknown, imagesAvailabilityUnknown), allConditionsSuccessfulOrUnknown)),
+		DestinationState: stateswitch.State(models.HostStatusInsufficient),
```
ori-amizur, line 560: If destination state is `insufficient`, will the next refresh move it to `known`?

internal/host/transition.go
```diff
@@ -685,6 +686,17 @@ func (th *transitionHandler) PostRefreshHost(reason string) stateswitch.PostTran
 			reason == statusInfoInstallationInProgressTimedOut {
 			template = statusInfoInstallationInProgressWritingImageToDiskTimedOut
 		}
+		if reason == statusInfoPreparationTimeout {
```
ori-amizur, line 689: I don't think we should rely on status info to make logical decisions.

internal/host/statemachine.go
paul-maidment: The aim here is to raise an event, the further transition of the host should follow the same pattern as for failure so I don't see a problem here?

internal/host/statemachine.go
paul-maidment: Good point, will change this.

internal/host/statemachine.go
paul-maidment: The error cases for disk speed check and image availability eventually end up "insufficient" anyway, so the focus of this transition is mainly to make sure that an event is raised as appropriate. Whatever the host transitions to afterwards is not a massive concern.

Especially as the timeout mirrors the same state changes we use for the error state.

internal/host/transition.go
paul-maidment: I will change this

internal/host/statemachine.go
paul-maidment: The short answer to `If destination state is insufficient, will the next refresh move it to known` is "yes"

ori-amizur: Seems strange behavior, that first refresh will move to one state, and the second one moves to another.  Do we have such behavior in different place?

internal/host/transition.go
```diff
@@ -665,6 +667,38 @@ func (th *transitionHandler) HasInstallationInProgressTimedOut(sw stateswitch.St
...
+		)
+		template := statusInfoPreparationTimeout
+		failingConditons := []string{}
+		if !params.conditions["installation-disk-speed-check-successful"] {
+			failingConditons = append(failingConditons, statusInfoPreparationTimeoutDiskSpeed)
+		}
+		if !params.conditions["successful-container-image-availability"] {
+			failingConditons = append(failingConditons, statusInfoPreparationTimeoutImageAvailability)
+		}
+		template = strings.Replace(template, "$FAILING_CONDITIONS", strings.Join(failingConditons, "\n"), 1)
```
ori-amizur, line 691:
```suggestion
		statusInfo := strings.Replace(statusInfoPreparationTimeout, "$FAILING_CONDITIONS", strings.Join(failingConditons, "\n"), 1)
```

internal/host/transition.go
```diff
@@ -665,6 +667,38 @@ func (th *transitionHandler) HasInstallationInProgressTimedOut(sw stateswitch.St
...
+			return errors.New("PostHostPreparationTimeout incompatible type of StateSwitch")
+		}
+		params, ok := args.(*TransitionArgsRefreshHost)
+		if !ok {
+			return errors.New("PostRefreshHost invalid argument")
+		}
+		var (
+			err error
+		)
+		template := statusInfoPreparationTimeout
```
ori-amizur, line 683: Not needed

internal/host/transition.go
paul-maidment: Here for example:

https://github.com/openshift/assisted-service/blob/master/internal/host/statemachine.go#L536-L550

This is a transition on error to "insufficient" which works in the same way as far as I can tell?

paul-maidment: I believe that this is the transition that then moves to known

https://github.com/openshift/assisted-service/blob/master/internal/host/statemachine.go#L918-L938

I was under the impression that this was already well understood?

ori-amizur: They are not the same:
The first has condition `stateswitch.Not(If(SufficientOrUnknownInstallationDiskSpeed)))`
And the second has the opposite condition:
`If(SufficientOrUnknownInstallationDiskSpeed)`  (in `isSufficientForInstall`)

internal/host/statemachine.go
paul-maidment: Agreed in conversation that we will have this transition straight to "Known"

internal/host/statemachine.go
```diff
@@ -549,6 +549,20 @@ func NewHostStateMachine(sm stateswitch.StateMachine, th TransitionHandler) stat
...
+	sm.AddTransitionRule(stateswitch.TransitionRule{
+		TransitionType: TransitionTypeRefresh,
+		SourceStates: []stateswitch.State{
+			stateswitch.State(models.HostStatusPreparingForInstallation),
+		},
+		Condition:        stateswitch.And(If(IsConnected), If(IsMediaConnected), th.IsPreparingTimedOut, stateswitch.Or(installationDiskSpeedUnknown, imagesAvailabilityUnknown), allConditionsSuccessfulOrUnknown),
+		DestinationState: stateswitch.State(models.HostStatusKnown),
+		PostTransition:   th.PostHostPreparationTimeout(),
+		Documentation: stateswitch.TransitionRuleDoc{
+			Name:        "Preparing timed out host move to insufficient",
```
ori-amizur, line 561: It moves to lnown

paul-maidment:
/hold 
Issue spotted with messaging around timeout

paul-maidment:
/unhold
Tested manually and working well

PR #5523, opened 2023-10-02
MGMT-14409: generate short image URL

adriengentil:
Generate short URL in the infraenv to download discovery ISO, see the [enhancement](https://github.com/openshift/assisted-service/blob/master/docs/enhancements/short-url-enhancement.md) for more details.

Test locally if there are no issues updating an infraenv from the former URL to the short URL. Tested also the opposite if we do a rollback.

### DIFF
--- a/internal/bminventory/inventory_v2_handlers.go
+++ b/internal/bminventory/inventory_v2_handlers.go
@@ -760,7 +760,7 @@ func (b *bareMetalInventory) GetInfraEnvDownloadURL(ctx context.Context, params 
 		return common.GenerateErrorResponder(errors.Errorf("OS image entry '%+v' missing OpenshiftVersion field", osImage))
 	}
 
-	newURL, expiresAt, err := b.generateImageDownloadURL(ctx, infraEnv.ID.String(), string(*infraEnv.Type), *osImage.OpenshiftVersion, infraEnv.CPUArchitecture, infraEnv.ImageTokenKey)
+	newURL, expiresAt, err := b.generateShortImageDownloadURL(infraEnv.ID.String(), string(*infraEnv.Type), *osImage.OpenshiftVersion, infraEnv.CPUArchitecture, infraEnv.ImageTokenKey)
 	if err != nil {
 		return common.GenerateErrorResponder(err)
 	}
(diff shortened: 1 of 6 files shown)


internal/imageservice/url_test.go
```diff
@@ -115,6 +126,43 @@ var _ = Describe("URL building", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		parsedISOType, parsedArch, parsedVersion, err := ParseDownloadURL(u)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(parsedISOType).To(Equal(isoType))
+		Expect(parsedArch).To(Equal(arch))
+		Expect(parsedVersion).To(Equal(version))
+	})
+
+	It("successfully parse malformed URL", func() {
```
osherdp, line 160: it should successfully parse a malformed URL?

internal/bminventory/inventory_v2_handlers.go
```diff
@@ -778,20 +778,51 @@ func (b *bareMetalInventory) GetInfraEnvDownloadURL(ctx context.Context, params
...
 		return "", nil, err
 	}
-	return urlString, expiresAt, nil
+
+	shortURL, err := imageservice.ShortImageURL(b.ImageServiceBaseURL, imageservice.ByTokenPath, token, version, arch, imageType)
+	return shortURL, exp, err
+}
+
+func (b *bareMetalInventory) generateShortImageDownloadURLByID(infraEnvID, imageType, version, arch string) (string, *strfmt.DateTime, error) {
+	var expiresAt strfmt.DateTime
```
osherdp, line 822: we don't need to set expiration date to some value?

internal/imageservice/url_test.go
adriengentil: maybe I should call it `successfully parse a non-infraenv URL`? as the function is best-effort

osherdp: yeah, just making it clearer

internal/bminventory/inventory_v2_handlers.go
adriengentil: `nil` should be returned in case of `byid` and `byapikey`, will fix it

adriengentil: actually no, before we were calling `ParseExpirationFromURL`: https://github.com/openshift/assisted-service/pull/5523/files#diff-9f6fae8a28b132e15c426658b6b5d41adf8a20006ddeb271a22fea270c573610R85-R96

which returns the default value of `strfmt.DateTime` when no `image _token` is passed.

adriengentil: We have tests here:
https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory_test.go#L18547
https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory_test.go#L18579
https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory_test.go#L18601

osherdp: so for all cases we're always returning an expiration date of `0001-01-01T00:00:00.000Z`?

adriengentil: no, in case of `image_token` we don't expect `0001-01-01T00:00:00.000Z`

osherdp: ack. thanks for the explanation!

osherdp (approved):

PR #5529, opened 2023-10-04
MGMT-15762 - fix odf validation failing where is one small disk

eifrach:
This PR fixing issue where if one of the disk is small it will fail the validation even if there is sufficient disks 

* add new test cases for this scenario 
* return test disk to random UUID 

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/operators/odf/validations.go
+++ b/internal/operators/odf/validations.go
@@ -167,20 +167,20 @@ func (o *operator) setStatusInsufficientResources(odfClusterResources *odfCluste
 
 // count all disks of drive type ssd or hdd
 func (o *operator) getValidDiskCount(disks []*models.Disk, installationDiskID string, mode odfDeploymentMode) (int64, error) {
-	var countDisks int64
-	var smallDisks int64
+	var countEligibleDisks int64
+	var countRegularDisks int64
 
 	for _, disk := range disks {
 		if (disk.DriveType == models.DriveTypeSSD || disk.DriveType == models.DriveTypeHDD) && installationDiskID != disk.ID && disk.SizeBytes != 0 {
 			if disk.SizeBytes >= conversions.GbToBytes(o.config.ODFMinDiskSizeGB) {
-				countDisks++
+				countEligibleDisks++
 			} else {
-				smallDisks++
+				countRegularDisks++
 			}
 		}
 	}
-	if countDisks == 0 && smallDisks != 0 {
-		return countDisks, fmt.Errorf("Insufficient resources to deploy ODF in %s mode. ODF requires a minimum of 3 hosts. Each host must have at least 1 additional disk of %d GB minimum and an installation disk.", strings.ToLower(string(mode)), o.config.ODFMinDiskSizeGB)
+	if countEligibleDisks == 0 && countRegularDisks > 0 {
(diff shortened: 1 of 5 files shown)


internal/operators/odf/validations.go
```diff
@@ -168,16 +168,19 @@ func (o *operator) setStatusInsufficientResources(odfClusterResources *odfCluste
 // count all disks of drive type ssd or hdd
 func (o *operator) getValidDiskCount(disks []*models.Disk, installationDiskID string, mode odfDeploymentMode) (int64, error) {
 	var countDisks int64
```
osherdp, line 170: Let's rename to ``countRegularDisks`` or ``countEligibleDisks`` so we'll know it's not the total count of disks

internal/operators/odf/validations.go
```diff
@@ -168,16 +168,19 @@ func (o *operator) setStatusInsufficientResources(odfClusterResources *odfCluste
...
-			} else {
+			if disk.SizeBytes >= conversions.GbToBytes(o.config.ODFMinDiskSizeGB) {
 				countDisks++
+			} else {
+				smallDisks++
 			}
 		}
 	}
-	return countDisks, err
+	if countDisks == 0 && smallDisks != 0 {
```
osherdp, line 182: nit
```suggestion
	if countDisks == 0 && smallDisks > 0 {
```

internal/operators/odf/validations.go
```diff
@@ -168,16 +168,19 @@ func (o *operator) setStatusInsufficientResources(odfClusterResources *odfCluste
...
+		return countDisks, fmt.Errorf("Insufficient resources to deploy ODF in %s mode. ODF requires a minimum of 3 hosts. Each host must have at least 1 additional disk of %d GB minimum and an installation disk.", strings.ToLower(string(mode)), o.config.ODFMinDiskSizeGB)
```
osherdp, line 183: isn't this message already duplicated [here](https://github.com/openshift/assisted-service/blob/890e9c7f8f1d178bf08c6f8c817c3f7ce776b39e/internal/operators/odf/validations.go#L162)?
in that case maybe this function can just count disks without validating the amount and raising an error

osherdp (changes requested):

internal/operators/odf/validations.go
eifrach: so, I want to create a follow up tasks to change this function. For now I created a dirty fix just for the bug.

so this function should return only the amount of eligible disks without an error and move all the error handling to the main / called by function.

it will be more readable and we can have more complex logic that needed to be add to the validation

internal/operators/odf/validations.go
eifrach: it is, this becuse there is some test's  / validation  will be broken if we dont add it. 
or we will need to re-factor a lot of test's.

as I metion on top, after changing this function logic it will be esier to handle and create better validation flow.

osherdp (approved):

PR #5533, opened 2023-10-04
MGMT-8115: Soft install timeout

jhernand:
This patch adds an enhacement proposal about replacing hard timeouts with soft timeouts, so that users can manually fix issues and installation can resume.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,122 @@
...
+This is important because users may be able to fix he issues that delay the
+installation instead of having to start it over.
+
+### Goals
+
+### Non-Goals
+
+It is not a goal to change the global installation timeout configured via the
+`INSTALLATION_TIMEOUT` environment variable. That is set by default to `24h`
+and at that point the assisted service stops monitoring the cluster.
```
mhrivnak, line 29: What do we gain by stopping monitoring of a cluster whose installation hasn't completed? Is it much of a resource burden for us?

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Proposal
+
+### User Stories
+
+#### Allow manually fixing a SasS installation
```
mhrivnak, line 35: s/SasS/SaaS/

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,122 @@
...
+### UI Impact
+
+The UI will need to explicitly show to the user that the cluster installation
+is taking longer than expected, and give suggestions on how to proceed. For
+example, we could present a warning message with this text:
+
+> Cluster installation is taking too long
+>
+> It should have completed in approximately 45 minutes, but it took 23 hours
+> and 52 minutes already. Check the logs to find out why or reset the cluster
```
mhrivnak, line 90: Can we provide more specific guidance than just "check the logs"? For example if a cluster-operator is in a bad state, can we just tell the user to look into that specific CO?

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,122 @@
...
+  service continues with the installation.
+
+These tests may require introducing a mechanism to artificially delay the
+installation in the agent or the installer.
+
+## Drawbacks
+
+Most of our cluster installation failures are due to these timeouts. If we just
+disable them then we will have a large amount of installations that have failed
+but will not be accounted as such.
```
mhrivnak, line 118: Why is that a drawback? How is the accounting of failed installations used?

For internal metrics, we could switch from "% of installs that succeeded" to "% of installs that succeeded in X time".

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,122 @@
...
+## Summary
+
+Currently we have an installation timeout of one hour. When that expires the
+cluster is automatically moved to the error state. We want to change that to a
+warning, so that users are informed that the installation is taking longer than
+expected and the installation will continue.
+
+## Motivation
+
+This is important because users may be able to fix he issues that delay the
```
mhrivnak, line 20: s/he/the/

docs/enhancements/soft-install-timeout.md
jelkosz: Also, what will be available to the users to troubleshoot? I guess in certain cases they will have a kubeadmin already, in certain they wont. Is it the case? If yes, can we explain the user why? Anything else we might provide to the user? Maybe a direct link to "something (tm)" on the cluster to check or hints to troubleshoot (e.g we noticed there might be networking issues, this is how you verify) or something like that?

avishayt: This is limited by (1) the information that the service has and (2) creating some expert system based on that information.  Until we can start working on that, at the minimum, we should identify some common error scenarios and instruct the user on how to determine the causes and fix them.

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,122 @@
...
+finalizing phases:
+
+- Prepare a cluster that exceeds the timeout and verify that fixing the issue
+  manually allows the service to continue with the installation.
+
+- Verify that the UI explicitly shows the information about the expired
+  timeout, and that it recovers when the issue is eventually fixed and the
+  service continues with the installation.
+
+These tests may require introducing a mechanism to artificially delay the
```
avishayt, line 111: How would we test without such a delay mechanism?
```suggestion
These tests may require introducing a mechanism to artificially delay the
```

docs/enhancements/soft-install-timeout.md
avishayt: I agree, it's not a drawback but a consideration.  It will probably also lead to more cancellations ("I don't know what's going on, I'll cancel and try again").

docs/enhancements/soft-install-timeout.md
jhernand: Done.

docs/enhancements/soft-install-timeout.md
jhernand: Done.

docs/enhancements/soft-install-timeout.md
jhernand: We could try to find a way to introduce delays without changing the code, like traffic shaping to make the network look artificially slow, for example. But I'd rather change the code; I think it will be simpler.

docs/enhancements/soft-install-timeout.md
mhrivnak: > After 24 hours the installation will fail anyway due to expired certificates, so there isn't much value in continuing.

Unless the cluster is up and running, but one cluster operator is degraded?

docs/enhancements/soft-install-timeout.md
ammont82: @avishayt the numbers in the UI message are always the same? **45** minutes and **23** hours and **52** minutes? Or do we need to calculate it?

docs/enhancements/soft-install-timeout.md
avishayt: True.

docs/enhancements/soft-install-timeout.md
avishayt: I suggest we get the exact message by UX and then decide.  In this case, the 23 hours and 52 minutes can be calculated by the timestamp on the cluster.  The 45 minutes would be hardcoded.  But again, let's get the exact message from UX first.

jhernand:
This conversation continues in #5687. Thanks @ori-amizur!

PR #5543, opened 2023-10-10
MGMT-15947: Add assisted images service short URL routes

adriengentil:
Add short URL routes into the ingress object when deploying through the
scripts.


gamli75 (approved):

adriengentil:
/retitle MGMT-15947: Add assisted images service short URL routes

gamli75 (approved):

PR #5548, opened 2023-10-12
MGMT-15796: set CloudControllerManager to External for OCI

adriengentil:
The installer indroduces a breaking change in the definition of the external
platform through https://github.com/openshift/installer/pull/7581. Previously,
the CCM was enabled without a way to disable it, now it will be disabled by
default with a way to enable it.

This change ensure the CCM is enabled when platform is oci in order to keep the
current behaviour.

### DIFF
--- a/internal/installcfg/installcfg.go
+++ b/internal/installcfg/installcfg.go
@@ -128,8 +128,23 @@ type NutanixPrismElement struct {
 	Name                           string          `yaml:"name"`
 }
 
+// CloudControllerManager describes the type of cloud controller manager to be enabled.
+type CloudControllerManager string
+
+const (
+	// CloudControllerManagerTypeExternal specifies that an external cloud provider is to be configured.
+	CloudControllerManagerTypeExternal = "External"
+
+	// CloudControllerManagerTypeNone specifies that no cloud provider is to be configured.
+	CloudControllerManagerTypeNone = ""
+)
+
 type ExternalInstallConfigPlatform struct {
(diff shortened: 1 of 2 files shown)


adriengentil:
/cc @pawanpinjarkar @zaneb 
I will un-draft once the backport will reach the nightly

adriengentil:
/test edge-e2e-oci-assisted
I guess it will fail since the key `cloudControllerManager` doesn't exist yet in the install config

gamli75 (approved):

adriengentil:
> I guess it will fail since the key `cloudControllerManager` doesn't exist yet in the install config

The install config was generated properly: 
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-service/5548/pull-ci-openshift-assisted-service-master-edge-e2e-oci-assisted/1712407212791435264/artifacts/e2e-oci-assisted/assisted-common-gather/artifacts/2023-10-12_10-54-16_7425552e-27db-4c5a-8880-c24bfa5839a4/cluster_files/install-config.yaml

... and it did not fail, so I think we can go ahead with this change without waiting for the back-port.

adriengentil:
/override ci/prow/edge-e2e-nutanix-assisted ci/prow/edge-e2e-vsphere-assisted
these jobs are broken for sometime and this PR does not impact nutanix/vsphere

PR #5565, opened 2023-10-17
OCPBUGS-19823: Ignore hostPrefix validation for plugins other than OVN/SDN

bfournie:
For plugins other than OVN/SDN (e.g. Calico), the clusterNetwork hostPrefix setting is not used so remove the validation. Since the networkType for plugins other than OVN/SDN can be passed in installconfig-overrides, that must also be checked.

Note that this is similar to the installer change to ignore the hostPrefix for these networkTypes - https://github.com/openshift/installer/pull/3888.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/validator.go
+++ b/internal/cluster/validator.go
@@ -536,14 +536,16 @@ func (v *clusterValidator) networkPrefixValid(c *clusterPreprocessContext) (Vali
 		return ValidationPending, "The Cluster Network CIDR is undefined."
 	}
 
-	if v.skipNetworkHostPrefixCheck(c) {
-		return ValidationSuccess, "The Cluster Network prefix is valid."
-	}
+	skipHostPrefix := v.skipNetworkHostPrefixCheck(c)
+	hostPrefixForCidrValidation := 25 // default to 128 addresses
 
 	validClusterNetworks := funk.Filter(c.cluster.ClusterNetworks, func(clusterNetwork *models.ClusterNetwork) bool {
+		if !skipHostPrefix {
+			hostPrefixForCidrValidation = int(clusterNetwork.HostPrefix)
+		}
 		return clusterNetwork != nil &&
-			network.VerifyNetworkHostPrefix(clusterNetwork.HostPrefix) == nil &&
-			network.VerifyClusterCidrSize(int(clusterNetwork.HostPrefix), string(clusterNetwork.Cidr), len(c.cluster.Hosts)) == nil
+			(skipHostPrefix || network.VerifyNetworkHostPrefix(clusterNetwork.HostPrefix) == nil) &&
+			network.VerifyClusterCidrSize(hostPrefixForCidrValidation, string(clusterNetwork.Cidr), len(c.cluster.Hosts)) == nil
 	}).([]*models.ClusterNetwork)
 
 	if len(validClusterNetworks) == len(c.cluster.ClusterNetworks) {
--- a/internal/cluster/validator.go
+++ b/internal/cluster/validator.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
@@ -11,6 +12,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/host"
+	"github.com/openshift/assisted-service/internal/installcfg"
 	"github.com/openshift/assisted-service/internal/network"
 	"github.com/openshift/assisted-service/internal/usage"
 	"github.com/openshift/assisted-service/models"
@@ -534,9 +536,9 @@ func (v *clusterValidator) networkPrefixValid(c *clusterPreprocessContext) (Vali
 		return ValidationPending, "The Cluster Network CIDR is undefined."
 	}
 	validClusterNetworks := funk.Filter(c.cluster.ClusterNetworks, func(clusterNetwork *models.ClusterNetwork) bool {
-		return clusterNetwork != nil &&
+		return clusterNetwork != nil && (v.skipNetworkHostPrefixCheck(c, clusterNetwork.HostPrefix) ||
 			network.VerifyNetworkHostPrefix(clusterNetwork.HostPrefix) == nil &&
-			network.VerifyClusterCidrSize(int(clusterNetwork.HostPrefix), string(clusterNetwork.Cidr), len(c.cluster.Hosts)) == nil
+				network.VerifyClusterCidrSize(int(clusterNetwork.HostPrefix), string(clusterNetwork.Cidr), len(c.cluster.Hosts)) == nil)
 	}).([]*models.ClusterNetwork)
 
 	if len(validClusterNetworks) == len(c.cluster.ClusterNetworks) {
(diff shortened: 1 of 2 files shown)


gamli75:
@ori-amizur @tsorya please review

internal/cluster/validator.go
```diff
@@ -534,9 +536,9 @@ func (v *clusterValidator) networkPrefixValid(c *clusterPreprocessContext) (Vali
 		return ValidationPending, "The Cluster Network CIDR is undefined."
 	}
 	validClusterNetworks := funk.Filter(c.cluster.ClusterNetworks, func(clusterNetwork *models.ClusterNetwork) bool {
-		return clusterNetwork != nil &&
+		return clusterNetwork != nil && (v.skipNetworkHostPrefixCheck(c, clusterNetwork.HostPrefix) ||
```
tsorya, line 539: maybe better to take v.skipNetworkHostPrefixCheck(c, clusterNetwork.HostPrefix) outside this line? it is not readable actually.
Maybe we can just set if before and return success in case we should skip it?

bfournie: Thanks, good point. Updated to check the skip first.

internal/cluster/validator.go
```diff
@@ -566,3 +573,36 @@ func (v *clusterValidator) isNtpServerConfigured(c *clusterPreprocessContext) (V
...
+// Ignore hostPrefix set to 0 for non-OVN/SDN plugins.
+func (v *clusterValidator) skipNetworkHostPrefixCheck(c *clusterPreprocessContext, prefix int64) bool {
+	// list of known plugins that require hostPrefix to be set
+	var pluginsUsingHostPrefix = []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
+
+	if prefix != 0 {
+		return false
+	}
+
+	if !funk.ContainsString(pluginsUsingHostPrefix, swag.StringValue(c.cluster.NetworkType)) {
```
tsorya, line 586: currently this condition will we always false as we allow only 2 network types, so maybe there is no point to validate it?

internal/cluster/validator.go
```diff
@@ -566,3 +573,36 @@ func (v *clusterValidator) isNtpServerConfigured(c *clusterPreprocessContext) (V
 	return ValidationFailure, fmt.Sprintf("Hosts' clocks are not synchronized (there's more than a %d minutes gap between clocks), "+
 		"please configure an NTP server via DHCP or set clocks manually.", common.MaximumAllowedTimeDiffMinutes)
 }
+
+// Ignore hostPrefix set to 0 for non-OVN/SDN plugins.
+func (v *clusterValidator) skipNetworkHostPrefixCheck(c *clusterPreprocessContext, prefix int64) bool {
+	// list of known plugins that require hostPrefix to be set
+	var pluginsUsingHostPrefix = []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
+
+	if prefix != 0 {
```
tsorya, line 582: what this means means? Didn't understand to say the truth

internal/cluster/validator.go
```diff
@@ -534,8 +536,13 @@ func (v *clusterValidator) networkPrefixValid(c *clusterPreprocessContext) (Vali
 		return ValidationPending, "The Cluster Network CIDR is undefined."
 	}
 	validClusterNetworks := funk.Filter(c.cluster.ClusterNetworks, func(clusterNetwork *models.ClusterNetwork) bool {
-		return clusterNetwork != nil &&
-			network.VerifyNetworkHostPrefix(clusterNetwork.HostPrefix) == nil &&
+		if clusterNetwork == nil {
+			return false
+		}
+		if v.skipNetworkHostPrefixCheck(c, clusterNetwork.HostPrefix) {
```
tsorya, line 542: I actually thought to set it before  validClusterNetworks , i mean there is no point to validate anything if you know you should skip it no?

bfournie: The thought was to skip the validation *only* if the prefix was not set and the networkTypes was not OVN of SDN.
If we take the approach that the validation should be skipped regardless of the prefix then we can move this check before.

internal/cluster/validator.go
bfournie: The thought was to skip the validation *only* if the prefix was not set and the networkTypes was not OVN of SDN.
If we take the approach that the validation should be skipped regardless of the prefix then this check can be removed (and as above the check can be moved).

internal/cluster/validator.go
bfournie: I was mainly added this in response to the comment about adding Calico or a customer network to the network list https://issues.redhat.com/browse/OCPBUGS-19823?focusedId=23278074&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23278074
If that was done in the future this check would not have to change.

internal/cluster/validator.go
bfournie: Updated to skip validations only based on networkType and ignore the prefix.

internal/cluster/validator.go
bfournie: Updated to skip validations only based on networkType and ignore the prefix.

internal/cluster/validator.go
```diff
@@ -533,6 +535,11 @@ func (v *clusterValidator) networkPrefixValid(c *clusterPreprocessContext) (Vali
...
 		return ValidationPending, "The Cluster Network CIDR is undefined."
 	}
+
+	if v.skipNetworkHostPrefixCheck(c) {
+		return ValidationSuccess, "The Cluster Network prefix is valid."
+	}
+
 	validClusterNetworks := funk.Filter(c.cluster.ClusterNetworks, func(clusterNetwork *models.ClusterNetwork) bool {
 		return clusterNetwork != nil &&
 			network.VerifyNetworkHostPrefix(clusterNetwork.HostPrefix) == nil &&
```
ori-amizur, line 545: `VerifyNetworkHostPrefix` verifies that the requested number of hosts matches the network parameters.  Even if we assume that each host has 1 address, can we verify that the cluster network is enough for the requested number of hosts?

internal/cluster/validator.go
```diff
@@ -566,3 +573,32 @@ func (v *clusterValidator) isNtpServerConfigured(c *clusterPreprocessContext) (V
...
+	// list of known plugins that require hostPrefix to be set
+	var pluginsUsingHostPrefix = []string{models.ClusterNetworkTypeOVNKubernetes, models.ClusterNetworkTypeOpenShiftSDN}
+
+	if !funk.ContainsString(pluginsUsingHostPrefix, swag.StringValue(c.cluster.NetworkType)) {
+		v.log.Infof("skipping network prefix check for %s", swag.StringValue(c.cluster.NetworkType))
+		return true
+	}
+
+	// Also check networkType in install-config overrides
+	if c.cluster.InstallConfigOverrides != "" {
```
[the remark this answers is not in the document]
bfournie: I can move the check around but I'm not sure I understand, the first time its checking c.cluster.NetworkType and then the 2nd time its checking cfg.Networking.NetworkType, so its not checking it twice.

internal/cluster/validator.go
bfournie: Do you mean VerifyClusterCidrSize() where that verification is done?  
I could move skipNetworkHostPrefixCheck() back into the loop here https://github.com/openshift/assisted-service/blob/master/internal/cluster/validator.go#L538 and check clusterNetworkPrefix for each cluster against the number of hosts, returning false to not skip, to let VerifyClusterCidrSize() return the error. Is that OK?

bfournie: @ori-amizur - I added back in check to ensure that cidr validation is done, let me know if that looks reasonable. Thanks.

internal/cluster/validator.go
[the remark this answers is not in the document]
ori-amizur: If there is `c.cluster.InstallConfigOverrides` is not empty and it contains NetworkType, isn't it the only relevant value ?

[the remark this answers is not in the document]
bfournie: Yes, I see what you mean. Updated to move the checks around. Thanks.

internal/cluster/validator.go
```diff
@@ -566,3 +577,33 @@ func (v *clusterValidator) isNtpServerConfigured(c *clusterPreprocessContext) (V
...
+		overrideDecoder := json.NewDecoder(strings.NewReader(c.cluster.InstallConfigOverrides))
+		overrideDecoder.DisallowUnknownFields()
+
+		cfg := &installcfg.InstallerConfigBaremetal{}
+		if overrideDecoder.Decode(cfg) != nil {
+			v.log.Infof("could not decode install-config overrides %s", c.cluster.InstallConfigOverrides)
+			return false
+		}
+
+		if cfg.Networking.NetworkType != "" && !funk.ContainsString(pluginsUsingHostPrefix, cfg.Networking.NetworkType) {
```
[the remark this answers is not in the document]
bfournie: Yes, good catch. Fixed.

PR #5635, opened 2023-10-29
MGMT-15683: Ensure that manifest filename has valid name part

paul-maidment:
Presently when uploading a manifest, it's possible to upload a filename with just an extension, for example a filename '.yaml' is considered to be valid.

Clearly this is incorrect and this PR seeks to rectify this issue.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
- [] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/manifests/manifests.go
+++ b/internal/manifests/manifests.go
@@ -451,6 +451,15 @@ func (m *Manifests) fetchManifestContent(ctx context.Context, clusterID strfmt.U
 
 func (m *Manifests) validateManifestFileNames(ctx context.Context, clusterID strfmt.UUID, fileNames []string) error {
 	for _, fileName := range fileNames {
+		fileNameWithoutExtension := strings.TrimSuffix(fileName, filepath.Ext(fileName))
+		if len(strings.TrimSpace(fileNameWithoutExtension)) == 0 {
+			return m.prepareAndLogError(
+				ctx,
+				http.StatusUnprocessableEntity,
+				errors.Errorf("Cluster manifest %s for cluster %s has an invalid filename.",
+					fileName,
+					clusterID))
+		}
 		if strings.Contains(fileName, " ") {
 			return m.prepareAndLogError(
 				ctx,
--- a/internal/manifests/manifests_test.go
+++ b/internal/manifests/manifests_test.go
@@ -258,6 +258,23 @@ var _ = Describe("ClusterManifestTests", func() {
 				return fmt.Sprintf("{\"data\":\"%s\"}", largeElementBuilder.String())
 			}
 
+			It("Does not accept a filename that does not contain a name before the extension", func() {
+				clusterID := registerCluster().ID
+				content := "{}"
+				fileName := ".yaml"
+				response := manifestsAPI.V2CreateClusterManifest(ctx, operations.V2CreateClusterManifestParams{
+					ClusterID: *clusterID,
+					CreateManifestParams: &models.CreateManifestParams{
+						Content:  &content,
+						FileName: &fileName,
+					},
+				})
+				err := response.(*common.ApiErrorResponse)
+				expectedErrorMessage := fmt.Sprintf("Cluster manifest %s for cluster %s has an invalid filename.", fileName, clusterID)
+				Expect(err.StatusCode()).To(Equal(int32(http.StatusUnprocessableEntity)))
+				Expect(err.Error()).To(Equal(expectedErrorMessage))
+			})
+
 			It("Does not accept a filename that contains spaces", func() {
 				clusterID := registerCluster().ID
 				content := "{}"


internal/manifests/manifests.go
```diff
@@ -451,6 +451,15 @@ func (m *Manifests) fetchManifestContent(ctx context.Context, clusterID strfmt.U
 
 func (m *Manifests) validateManifestFileNames(ctx context.Context, clusterID strfmt.UUID, fileNames []string) error {
 	for _, fileName := range fileNames {
+		fileNameWithoutExtension := strings.TrimSuffix(fileName, filepath.Ext(fileName))
+		if len(strings.TrimSpace(fileNameWithoutExtension)) == 0 {
+			return m.prepareAndLogError(
+				ctx,
+				http.StatusUnprocessableEntity,
```
rccrdpccl, line 458: Why `StatusUnprocessableEntity`? Seems that is not very much used throughout the codebase, any particular reason why we picked this one as opposed to the more used `http.StatusBadRequest`? I don't disagree with the choice, but I'd prefer consistency with other validation if there isn't a compelling reason to not be consistent

paul-maidment: This is basically in line with bugfixes I was asked to make in https://issues.redhat.com/browse/MGMT-15684 where filename non compliance is requested to be treated as StatusUnprocessableEntity as opposed to BadRequest

Happy to open this up for broader debate but I felt that this was the best way to remain consistent?

@rccrdpccl

avishayt: I agree with @rccrdpccl and apologize for not looking more deeply into MGMT-15684.  It looks like this is another example of surprising code generation.  We have a pattern defined in the swagger for the file name, and I assume that returns 422 based on the JIRA.  Our code generally just returns 400 for invalid input.  The better fix would be to switch these cases to 400 and see how to fix the code generation so that it also returns 400.

paul-maidment: @avishayt Though there are no 422 codes declared in the swagger.yaml?

paul-maidment: So alter the code generation for the API in such a way that it returns a 400 error code for all cases where it presently returns a 422, can you clarify that this is what you are suggesting @avishayt ?

avishayt: Yes, ideally the generated code would return 400 instead of 422 for all cases

rccrdpccl: my bad, I thought that the generated code returned 400 on name validation, as I didn't see any mention of 422 in the manifest file.
If it's already returning 422 for name validation, I think (for the scope of this PR) we should just be consistent with what the generated code does (if we already return 422 as it seems on name validation, IMHO we should keep returning 422 with other naming validation even if done by internal code instead of generated one.
If then we don't like the 422 code, I'd tackle that in a different PR, but personally I have no issues with it as long as it's a consistent behaviour.
Sorry for rising all this dust on this trivial change :sweat:

paul-maidment: I'm actually not sure I agree that this should be changed in the way described.

https://www.webfx.com/web-development/glossary/http-status-codes/what-is-a-422-status-code/

"A 422 status code is typically caused when **a request is well-formed** but the server is unable to process it because it **contains semantic errors or does not meet certain conditions**. Some common causes of a 422 status code include:"

In all of the cases we are discussing, the syntax is correct but the errors are semantic (bad filename)
I would argue that we should be using 422...

Additionally, it looks like our codegen is based on `goswagger` and any change to the default behavior would require us to either define a template of our own or to deviate from the "factory" behavior of swagger. Or perhaps upgrade swagger to a new version.

I heard some discussion that some clients don't support a 422 error code, though I have not seen any concrete documentation of this problem.

@avishayt What are your thoughts?

avishayt: I'm OK with changing the generated code to 400 in a different PR as well.
We own the stratoscale template, so no worries there.  422 might be closer in terms of meaning, but realistically, clients won't differentiate between 400 and 422.  Browsers will likely see 422 and fall back to 400 (user error).  So I'd rather go with consistency and simplicity over possibly being more "correct" in this case.

paul-maidment: > I'm OK with changing the generated code to 400 in a different PR as well. We own the stratoscale template, so no worries there. 422 might be closer in terms of meaning, but realistically, clients won't differentiate between 400 and 422. Browsers will likely see 422 and fall back to 400 (user error). So I'd rather go with consistency and simplicity over possibly being more "correct" in this case.

What do we mean when we say the browser will "fall back" to 400 and what consequences does this have?

paul-maidment: If we change the goswagger template (actually not sure if this goes as deep as OpenAPI) then this will have consequences for any user who uses that template, I'm not sure we should be promoting incorrect behavior in the name of consistency.

PR #5687, opened 2023-11-14
MGMT-8115: Soft install timeouts enhancement document

ori-amizur:
This is continuation of enhancement document started by @jhernand with suggested implementation.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @filanov 
/cc @jhernand


docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+| Rebooting                  | host|                                  stage |    Assisted service |                 HOST_STAGE_REBOOTING_TIMEOUT |                       40m |                                      Move pending user action |
+| Wait for nodes             | installation (cluster)| controller | Assisted controller | hard coded |                       10h |                                       Abort waiting for nodes |
+| Wait for finalizing        | installation (cluster)| controller | Assisted controller | hard coded |                       10h | Don't perform post install + don't send complete installation |
+| Wait for cluster operators | installation (cluster)| controller | Assisted controller | hard coded |                       10h |  Don't perform rest of controller operations | Only CVO and console |
+| Add router CA              | installation (cluster)| controller | Assisted controller | hard coded |                       70m | Don't perform rest of controller operations |
+| Wait for OLM operators     | installation (cluster)| controller | Assisted controller || calculated from operators | | Don't perform rest of controller operations |
```
avishayt, line 97:
```suggestion
| Wait for OLM operators     | installation (cluster)| controller | Assisted controller || calculated from operators | Don't perform rest of controller operations |
```

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+In the flow managed by the controller all steps must be completed (in the table above).  The last step is a 
+a notification to complete installation.
+
+### Suggested changes
+- Only assisted service should notify installation completion
+- All steps by controller will be known as stages in the assisted service.  This will enable the service to manage 
+them in a similar way that host stages are managed.
+- Controller will not stop activity due to timeout.
+- Assisted service should be able to terminate the controller.
+- All existing timeouts except installation timeout and rebooting timeout should be treated as soft timeouts (i.e notification only)
```
avishayt, line 120: What kind of notification?  Cluster event?

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+  manually allows the service to continue with the installation.
+
+- Verify that the UI explicitly shows the information about the expired
+  timeout, and that it recovers when the issue is eventually fixed and the
+  service continues with the installation.
+
+These tests may require introducing a mechanism to artificially delay the
+installation in the agent or the installer.
+
+## Drawbacks
```
avishayt, line 164: Other drawbacks:
1. We will have more cancelations in SaaS and may miss bugs because we won't triage them.
2. Automated installations (e.g., ZTP or REST scripts) would hang for a long time, instead of failing and retrying.

docs/enhancements/soft-install-timeout.md
ori-amizur: Depends what is the subject of the timeout.  Either host or cluster.

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+- kubeconfig must be uploaded
+- cluster operators are successful
+- Monitored operators are either successful or failed
+
+In the flow managed by the controller all steps must be completed (in the table above).  The last step is a 
+a notification to complete installation.
+
+### Suggested changes
+- Only assisted service should notify installation completion
+- All steps by controller will be known as stages in the assisted service.  This will enable the service to manage
```
filanov, line 116: can you elaborate, which steps are we talking about here?

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+- All steps by controller will be known as stages in the assisted service.  This will enable the service to manage 
+them in a similar way that host stages are managed.
+- Controller will not stop activity due to timeout.
+- Assisted service should be able to terminate the controller.
+- All existing timeouts except installation timeout and rebooting timeout should be treated as soft timeouts (i.e notification only)
+- In case of some optional operation (i.e managed operators), the user will be able to decide to skip it and set the cluster as completed.
+
+### Open Questions
+
+- Do we also soften the 24h `INSTALLATION_TIMEOUT` timeout?
```
filanov, line 125: what will happen with the certificates? do we renew them after 24 hours? i suggest to make it hard validation

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+
+- Should we collect must gather logs when the installation timeout expires, even
```
filanov, line 127: when triage a cluster you should get the latest state no? 
Or you suspect that the log size will increase?

docs/enhancements/soft-install-timeout.md
filanov: 1. agree 
2. @berenss what do you think?

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+  if we don't mark it as failed?
+- Should any of these changes be optional ?
+
+### UI Impact
+
+The UI will need to explicitly show to the user that the cluster installation
+is taking longer than expected, and give suggestions on how to proceed. For
+example, we could present a warning message with this text:
+
+> Cluster installation is taking too long
```
mhrivnak, line 137: s/too long/longer than usual/

"too long" implies that it is unacceptably long, as if the amount of time itself is inherently bad. And it begs the question: "too long" ... for what?

We're proposing here that the length of time is just a crude indicator that something might be going awry, but is not itself a fundamental problem.

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+>
+> It should have completed in approximately 45 minutes, but it took 23 hours
```
mhrivnak, line 139: s/It should have completed/Most installations complete/

We don't have a good way to judge how long an installation "should" take on any particular combination of network, equipment, and environment.

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+  timeout, and that it recovers when the issue is eventually fixed and the
+  service continues with the installation.
+
+These tests may require introducing a mechanism to artificially delay the
+installation in the agent or the installer.
+
+## Drawbacks
+
+Most of our cluster installation failures are due to these timeouts. If we just
+disable them then we will have a large amount of installations that have failed
```
mhrivnak, line 167: Do we know for sure that the ones we mark today as failed are unrecoverably failed?

After implementing this proposal, if a user does nothing with the API or UI, it'll eventually be marked as failed anyway, right?

After implementing this proposal, if a user resets their installation, it seems that should still count as some sort of failure. If people start an installation and then have to abandon it and reset, that seems like some sort of failure worth tracking.

mhrivnak (changes requested):

docs/enhancements/soft-install-timeout.md
mhrivnak: Comparing today where a timeout causes a failure and then the user tries the install again, vs. tomorrow where a timeout gives a warning and then the user tries the install again, what information are we capturing today that we wouldn't be capturing tomorrow in order to identify and triage bugs?

docs/enhancements/soft-install-timeout.md
mhrivnak: If the new control plane is running, then the 24 hour certs no longer exist, right?

That said, 24h is a long time, and it seems reasonable to give up at that point.

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+## Summary
+
+Currently we have an installation timeout of one hour. When that expires the
+cluster is automatically moved to the error state. We want to change that to a
+warning, so that users are informed that the installation is taking longer than
+expected and the installation will continue.
+
+## Motivation
+
+This is important because users may be able to fix the issues that delay the
```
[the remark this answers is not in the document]
mhrivnak: Telcos and edge generally are situations where a slow network might cause the installation to take longer than usual. So this is still a useful change, even though I agree that for ZTP nobody wants to be hands-on fixing an installation. Until we can tell some AI bot to fix all the installations at scale for us. ;)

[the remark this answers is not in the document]
berenss: It is very unlikely, unless in some smaller POC phase of the ZTP deployment (ie, break/fix/testing) that hands-on fixing will be pursued.  Even then, it would probably break the ZTP methodology to introduce a human hand into the flow.
it should be expected that some low level percentage (ie, < 1%) of the ZTP high scale deployment will fail.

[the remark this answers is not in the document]
avishayt: In my opinion, not many people will try to fix a failed installation.  I would retry the installation because seeing the green success icon gives me confidence that my cluster is healthy.  If I spend time debugging, I would worry that there is some issue with the cluster that I didn't notice or fix.  For deployments like telco which are clearly cattle, they definitely won't debug.

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+installation instead of having to start it over.
```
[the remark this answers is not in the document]
avishayt: Are our timeouts affected by slow networks?  I think that the part of installation where we download from the network has a timeout that keeps resetting as long as we make progress.  @ori-amizur is this true or do I not remember correctly?

docs/enhancements/soft-install-timeout.md
avishayt: For example, there could be a bug where clusterversion never gets to healthy.  With today's code, the installation would time out, a ticket would be opened, and somebody would look at it and see that something is broken.  With this suggested change, the user would eventually give up and cancel, and we wouldn't triage it.  Of course, we could mitigate this by mimicking what we do today - in case the user cancels after X time, still open a ticket with all the logs and triage it as if it failed.

docs/enhancements/soft-install-timeout.md
avishayt: We already have a fair amount of cancelations today that we can't explain.  This would add more "real" failures into the mix of other reasons why somebody would cancel.

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+- "@jhernand"
+creation-date: 2023-10-04
+last-updated: 2022-10-04
+---
+
+# Soft install timeout
+
+## Summary
+
+Currently we have an installation timeout of one hour. When that expires the
```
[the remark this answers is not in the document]
ori-amizur: For each stage

docs/enhancements/soft-install-timeout.md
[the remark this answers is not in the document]
ori-amizur: OK.  So for ZTP installation by default existing behavior will be kept

docs/enhancements/soft-install-timeout.md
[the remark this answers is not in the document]
ori-amizur: Timeouts are counted from status/stage change timestamp.  Internal progress (in status/stage) is not relevant for timeout calculation

docs/enhancements/soft-install-timeout.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Non-Goals
+
+It is not a goal to change the global installation timeout configured via the
+`INSTALLATION_TIMEOUT` environment variable. That is set by default to `24h`
+and at that point the assisted service stops monitoring the cluster.
+
+## Proposal
+
+### User Stories
```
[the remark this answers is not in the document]
ori-amizur: OK

docs/enhancements/soft-install-timeout.md
ori-amizur: These are the controller rows in the table above

docs/enhancements/soft-install-timeout.md
ori-amizur: Seems to me 24h is a hard timeout

docs/enhancements/soft-install-timeout.md
ori-amizur: I think we should gather the logs, even if the cluster is not marked as failed

docs/enhancements/soft-install-timeout.md
ori-amizur: OK.  It should be phrased "longer than usual".  But again, what is the usual case?

docs/enhancements/soft-install-timeout.md
ori-amizur: When we say most, is it 51% or 99%?
What should the user do with such phrase?  Seems confusing

docs/enhancements/soft-install-timeout.md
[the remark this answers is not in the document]
filanov: please make it clear in the summary

docs/enhancements/soft-install-timeout.md
filanov: what about the logs size? 
if possible lets check how the size will be affected, currently we have docs that estimate the size and require the users to create a PV matching that size, i know that the behavior will be disabled for ACM but in case we enable it it's good to know what are the limits

PR #5710, opened 2023-11-16
WIP: MGMT-12996: Allow using custom OCP release

danmanor:
## In General
This PR is an Implementation of this enhancement proposal - https://github.com/openshift/assisted-service/pull/4893/files. There are two main parts to this implementation that are in the same PR because they are dependent on one another:

1. A utility that periodically and fetches `OCP` releases from an API, and stored them in the `DB`.
2. Using the release from part (1) to allow installation any supported `OCP` version

### part 1
This part is about creating a periodic utility that updates current `OCP` releases in `postgresql` DB, for later use of assisted service.

The main flow is:

1. main fetches new configuration called `RELEASE_SOURCES` which specifies which `OCP` releases we want to keep in the DB and support (new table called `openshift_versions`) as well as from which channel, what CPU architectures, etc. Then it starts a new go routine that runs periodically.
3. This go routine fetches releases from the API according to the configuration, and `OCP` versions support levels from another API.
4. It labels the default release (latest stable release), supporting levels.
5. Replaces this data with the old one in the DB.

### part 2
This part is about retrieving the new list of supported openshift versions in endpoint `/api/assisted-install/v2/openshift-versions`, and allowing their installation.

The main changes are:

1. In openshift-versions endpoint - In addition to the releases that we retrieved before this PR, which are the releases we configured using `RELEASE_IMAGES`, we retrieve the ones from the DB (from part (1)) as well, such that the configuration has precedence, meaning if a version appears in both DB and config, the config i the one being retrieved. We also filter by 2 new query parameters (`only_latest`, `version_pattern`) that will be used by the `UI` to present short list of the versions. We also filter by a new configuration parameter `IGNORED_RELEASE_IMAGES`, which contains a list of versions we don't want to retrieve (as one might have a blocker bug).
2. In order to allow the actual installation of these new releases, we Adding lookup in the `DB` releases for `version handler` functions if they are not found in the configuration.

### Remarks

1.  As proposed in: https://github.com/openshift/assisted-service/blob/e604bc46caf1fac53eab35817c739add1c8911ee/docs/enhancements/dynamic-fetching-of-openshift-versions.md?plain=1#L145 I exposed new end point in - `/api/assisted-install/v2/release-sources`

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [x] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md,


internal/versions/versions.go
```diff
@@ -338,6 +358,83 @@ func (h *handler) validateVersions() error {
...
+}
+
+func syncReleaseImages(db *gorm.DB, releaseSources models.ReleaseSources) error {
+	for _, releaseSource := range releaseSources {
+		openshiftVersion := *releaseSource.OpenshiftVersion
+		for _, upgradeChannel := range releaseSource.UpgradeChannels {
+			cpuArchitecture := *upgradeChannel.CPUArchitecture
+			for _, channel := range upgradeChannel.Channels {
+				// OCP releases API needs amd64, arm64 instead of x86_64, aarch64 respectively
+				if cpuArchitecture == "x86_64" {
```
rccrdpccl, line 377: using a map instead of if-else will improve readability and improve this translating back-and-forth

```suggestion
cpuArchDict := map[string]string{
  "x86_64": "amd64",
  "aarch64": "arm64",
}
cpuArch := defaultArch
if  arch, ok :=  cpuArchDict[cpuArchitecture]; ok {
   cpuArch := arch
}
```

internal/versions/versions.go
```diff
@@ -338,6 +358,83 @@ func (h *handler) validateVersions() error {
...
+							OpenshiftVersion: node.Version,
+							Channel:          channel,
+							CpuArchitecture:  cpuArchitecture,
+							Url:              releaseUrl,
+						}
+
+						if existAlready(db, openShiftReleaseImage) {
+							continue
+						}
+						db.Create(&openShiftReleaseImage)
```
rccrdpccl, line 416: Could some image cease to exist? Thinking early access here, but there might be other cases. How are we going to handle those?

internal/versions/versions.go
```diff
@@ -338,6 +358,83 @@ func (h *handler) validateVersions() error {
 	return nil
 }
 
+func existAlready(db *gorm.DB, openShiftReleaseImage common.OpenShiftReleaseImage) bool {
```
rccrdpccl, line 361: Why is this (both functions) part of this package/file? It seems independent to me.
Think about extracting this logic to a separate layer: this would improve testability (both of the extracted method and the current one) and maintanability.
If it will be necessary in versions package/method receiver we can always inject said indipendent layer

internal/versions/versions_test.go
```diff
@@ -650,9 +653,556 @@ var _ = Describe("GetReleaseImageByURL", func() {
...
+		It("Should be True when exist", func() {
+			ocpReleaseImage := common.OpenShiftReleaseImage{
+				OpenshiftVersion: "4.14.1",
+				Channel:          "stable",
+				CpuArchitecture:  "x86_64",
+				Url:              fmt.Sprintf(OcpImageUrlTemplate, "stable", "4.14.1"),
+			}
+
+			db.Create(&ocpReleaseImage)
+			result := existAlready(db, ocpReleaseImage)
```
rccrdpccl, line 726: It's good practice to test only public methods, so that we don't have tests that are too fragile. We still most likely want to cover cases where the image already exists, testing it implicitly.

internal/versions/versions.go
rccrdpccl, line 361: nit: I think the right spelling is `existsAlready`

internal/versions/versions_test.go
```diff
@@ -650,9 +653,556 @@ var _ = Describe("GetReleaseImageByURL", func() {
 	})
 })
 
+var _ = Describe("SyncReleaseImages", func() {
+	var (
+		db          *gorm.DB
+		dbName      string
+		ctrl        *gomock.Controller
+		err         error
+		handler     *handler
```
rccrdpccl, line 662: Why adding this dependency? Is it really so much better than stdlib [httptest](https://go.dev/src/net/http/httptest/example_test.go) ?

rccrdpccl (changes requested):

cmd/main.go
```diff
@@ -261,6 +263,14 @@ func main() {
...
 	}
 
+	var releaseSourcesArray models.ReleaseSources
+	if Options.ReleaseSources == "" {
+		releaseSourcesArray = models.ReleaseSources{}
+	} else {
+		failOnError(json.Unmarshal([]byte(Options.ReleaseSources), &releaseSourcesArray),
+			"Failed to parse RELEASE_SOURCES json %s", Options.ReleaseSources)
+	}
+
```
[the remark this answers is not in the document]
danmanor: Agree. Just followed the way the last option was done

internal/versions/versions.go
danmanor: Nice

internal/versions/versions.go
danmanor: That is a good point. I am not sure it can happen, but either way flushing the table each run is a good idea. WDYT?

internal/versions/versions_test.go
danmanor: I was now aware of httptest. I will try it out thanks

internal/versions/versions.go
danmanor: I was not sure where is the best place to inject this code. I thought starting a new directory for this was a bit redundant, so I Just put it where I thought it is most fitting. Do you think I should create `internal/ocpversions` ?

danmanor: Thanks

cmd/main.go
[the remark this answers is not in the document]
rccrdpccl: I know there are other such example in the current code, feel free to change those ones too if you wish :)

internal/versions/versions.go
rccrdpccl: I think maybe flushing, given we can safely lock the table for a small amount of time... but we might risk to be without data if things go wrong.
Another option would be having an internal field transaction ID or something, update all records then delete the ones that don't have such transaction ID.

internal/versions/versions.go
rccrdpccl: I think having small "classes" or equivalent in go is always good: it keeps the complexity low. If I have to pick, I'd rather have too many files than files with too many lines/responsibilities.
I think `internal/ocpversions` could be a good option, however the model is already named release_streams, WDYT about keeping it consistent with that?

internal/versions/versions.go
danmanor: Not sure I understand, we can just update then if it is already exists no?

internal/versions/versions.go
danmanor: You mean release_images ?

danmanor: or release_source ?

internal/versions/versions.go
rccrdpccl: for the second option, yes, we can update if it already exists, and then delete all the entries with different transaction/version ID, i.e.:
```
version | arch | versionID |
--------+------+-----------|
foobar  | x86  | 0af...    |
foobar  | arm  | 0af...    |
```

next update we only get `foobar, arm`:
```
version | arch | versionID |
--------+------+-----------|
foobar  | x86  | 0af...    |
foobar  | arm | 123...     |
```
then delete all records with versionID != 123

another way could be what you suggested about flushing, but my concern is the risk of locking or leaving empty. However the versionID solution risks of reporting stale data.

internal/versions/versions.go
rccrdpccl: I meant release_sources , sorry

internal/versions/versions_test.go
danmanor: I think `httpmock` is better for this case, as it doesn't actually make an `http` request, unlike `httptest`. If I want to use `httptest`, I will need to change the endpoint I make the request to configurable, but that is a bit odd as the endpoint is computed on the fly during execution (the query params). WDYT ?

rccrdpccl: We are already looping through params to make up endpoint every time we want to make a request.
Not sure I see the issue here, in my mind it would be something like that:
```
var responseMatcher map[string]RequestResponseParameter

for _, parameters := range requestsResponsesParameters {
    endpoint := fmt.Sprintf(VersionsAPITemplate, parameters.Channel, parameters.Version, parameters.CPUArchitecture)
    responseMatcher[endpoint] = parameters
}

server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    if params, ok := responseMatcher[strings.TrimSpace(r.URL.Path)] {
         w.WriteHeader(http.StatusOK)
         w.Write(params.Response)
         return
    }
    w.WriteHeader(http.StatusNotFound)
}))
```

Maybe slightly more verbose (can be easily factored it out to something like `mockHTTPRequests(requestsResponsesParameters)`, but very simple and not requiring external dependency.
It would be also easy to mock errors as you'd just need to add an error element to the list (although I guess the same can be done with httpmock if we change approach slightly)
Does it make sense? WDYT?

danmanor: Got you now. Nice thank you

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,413 @@
...
+		if *majorMinorVersion == openshiftMajorMinorVersion && release.Channel == "candidate" {
+			releasesToDelete = append(releasesToDelete, release)
+		}
+	}
+	// Nothing to delete
+	if len(releasesToDelete) == 0 {
+		return nil
+	}
+
+	return db.Where("TRUE").Delete(&releasesToDelete).Error
```
rccrdpccl, line 247: Why where TRUE?

danmanor: I think `Delete` must have Where as a prerequisite, not sure though I will check

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,413 @@
...
+type Data struct {
+	Versions []Version `json:"versions"`
+}
+
+type Version struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
+}
+
+const openshiftReleasesAPIBaseUrl string = "https://api.openshift.com/api/upgrades_info/v1/graph"
```
rccrdpccl, line 45: shouldnt this be parametrized? i.e Changing environment, versions, etc

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,413 @@
...
+	err = setDefaultRelease(db)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h *ReleaseSourcesHandler) SyncReleaseImages() {
+	defer func() {
```
rccrdpccl, line 374: should we return early if we have no release sources?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,413 @@
...
+		}
+
+		if lessThan {
+			latestReleaseVersion = release.DisplayName
+			latestStableRelease = release
+		}
+	}
+
+	// Changing to new default
+	err = db.Model(&common.OpenshiftVersion{}).
```
rccrdpccl, line 312: Not critical, but this should run in a transaction together with setting a new default, and possibly more.

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,413 @@
...
+			if convertion, ok := mapAPISupportLevelToOurSupportLevel[version.Type]; ok {
+				version.Type = convertion
+			}
+
+			err = deletePreReleaseVersions(db, version.Name)
+			if err != nil {
+				return err
+			}
+
+			err = setSupportLevel(db, version)
```
rccrdpccl, line 348: Can we try to unify the insert to contain this data too, without adding complexity in the code?
This will add N inserts + one update on few rows later. Not a big deal but not necessary

internal/releasesources/release_sources_test.go
```diff
@@ -0,0 +1,1359 @@
...
+						DisplayName:  swag.String("4.15.0-ec.2"),
+						SupportLevel: swag.String("beta"),
+						Default:      false,
+					},
+					CpuArchitectures: []string{"x86_64"},
+					Channel:          "candidate",
+				},
+			}
+
+			err = syncReleaseImages(db, handler.ReleaseSources, OpenShiftReleasesAPIClient{BaseUrl: releasesServer.URL})
```
rccrdpccl, line 1186: It's best to test public interface and not private methods. Testing private methods make tests fragile

internal/releasesources/release_sources_test.go
```diff
@@ -0,0 +1,1359 @@
...
+			Response        string
+		}
+
+		type QueryParameters struct {
+			Channel string
+			Arch    string
+		}
+
+		// Should match defaultReleaseSources
+		requestsResponsesParameters := []RequestResponseParameters{
```
rccrdpccl, line 121: If we hid this behind a utility method the test setup would be more readable.

rccrdpccl: (same with other long-texted vars)

internal/releasesources/release_sources_test.go
```diff
@@ -0,0 +1,1359 @@
...
+)
+
+func TestHandler(t *testing.T) {
+	RegisterFailHandler(Fail)
+	common.InitializeDBTest()
+	defer common.TerminateDBTest()
+	RunSpecs(t, "releasesources")
+}
+
+var _ = Describe("SyncReleaseImages", func() {
```
rccrdpccl, line 27: Maybe we can add a test around candidate versions deletion?

internal/releasesources/release_sources.go
danmanor: Do you mean I should select the table before everything and then use it everywhere? I would need to select it again every time I am ending a transaction with `DELETE/UPDATE/...` no?

internal/versions/api.go
```diff
@@ -64,10 +76,243 @@ func GetModelVersions(v Versions) models.Versions {
...
+			if newMultiRelease != nil && hasMultiArchAuthorization {
+				openshiftVersions[*newMultiRelease.DisplayName] = newMultiRelease.OpenshiftVersion
+			}
+		}
+	}
+
+	return openshiftVersions
+}
+
+func (h *apiHandler) getOpesnhiftReleasesForOnlyLatestParamFromDB(onlyLatest *bool) ([]*common.OpenshiftVersion, error) {
```
carbonin, line 175: Typo should be `Openshift`, not `Opesnhift`

internal/versions/api.go
```diff
@@ -64,10 +76,243 @@ func GetModelVersions(v Versions) models.Versions {
...
+	dbOpenshiftVersions, err = common.GetOpenshiftVersionsFromDBWhere(h.versionsHandler.db)
+	if err != nil {
+		h.log.WithError(err).Debug("error occurred while trying to get openshift versions from the DB")
+		return nil, err
+	}
+
+	return dbOpenshiftVersions, nil
+}
+
+func (h *apiHandler) getOpesnhiftReleasesForVersionPatternParamFromDB(versionPattern *string) ([]*common.OpenshiftVersion, error) {
```
carbonin, line 214: Also typo here

cmd/main.go
```diff
@@ -406,22 +428,28 @@ func main() {
...
+	clusterStateMonitor.Start(false)
 	defer clusterStateMonitor.Stop()
 
 	hostStateMonitor := thread.New(
 		log.WithField("pkg", "host-monitor"), "Host State Monitor", Options.HostStateMonitorInterval, hostApi.HostMonitoring)
-	hostStateMonitor.Start()
+	hostStateMonitor.Start(false)
 	defer hostStateMonitor.Stop()
 
+	OpenShiftReleaseSyncer := thread.New(
```
carbonin, line 448: This shouldn't run in kubeAPI since these are managed by ClusterImageSet

internal/versions/versions.go
```diff
@@ -18,32 +17,50 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
 	"golang.org/x/sync/semaphore"
+	"gorm.io/gorm"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const releaseImageURLPrefix string = "quay.io/openshift-release-dev/ocp-release:"
```
carbonin, line 24: Not sure what this is being used for but it would be worth considering that on-prem we allow users to specify their own release image through ClusterImageSet

internal/versions/versions.go
```diff
@@ -18,32 +17,50 @@ import (
...
+
+var supportedMultiArchitectures []string = []string{common.X86CPUArchitecture, common.ARM64CPUArchitecture, common.S390xCPUArchitecture, common.PowerCPUArchitecture}
```
carbonin, line 26: Is this defined elsewhere?

internal/versions/versions.go
```diff
@@ -92,37 +121,238 @@ func (h *handler) GetMustGatherImages(openshiftVersion, cpuArchitecture, pullSec
...
 	}
 
-	// The image doesn't exist in the cache.
-	// Fetch all the cluster image sets, cache them, then search the cache again
+	arch, ok := mapToURLCpuArchitecture[cpuArchitecture]
+	if !ok {
+		arch = cpuArchitecture
+	}
 
+	return swag.String(fmt.Sprintf(releaseImageURLPrefix+"%s-%s", *release.DisplayName, arch))
```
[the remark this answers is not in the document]
carbonin: I'd say we should store the url in the database when we create the release there/

That will make things more portable if we end up storing releases in the DB for on-prem as well.

internal/versions/versions.go
```diff
@@ -92,37 +121,238 @@ func (h *handler) GetMustGatherImages(openshiftVersion, cpuArchitecture, pullSec
...
+}
+
+func (h *handler) getReleaseSupportLevel(release models.ReleaseImage) (*string, error) {
```
carbonin, line 187: I feel like a lot of this belongs with the db-specific code rather than in the versions handler.

internal/versions/versions.go
```diff
@@ -92,37 +121,238 @@ func (h *handler) GetMustGatherImages(openshiftVersion, cpuArchitecture, pullSec
...
+
+	targetVersion, err = h.getLatestDBRelease(dbOpenshiftVersions)
+	if err != nil {
+		return nil, err
+	}
+
+	return h.createReleaseImageFromDBOpenshiftVersion(targetVersion, cpuArchitecture)
+}
+
+func (h *handler) getLatestDBRelease(releases []*common.OpenshiftVersion) (*common.OpenshiftVersion, error) {
```
carbonin, line 305: I don't think we should have two different types.

Ideally we would use ReleaseImage for the DB.

internal/versions/versions.go
```diff
@@ -136,14 +366,14 @@ func (h *handler) GetReleaseImage(ctx context.Context, openshiftVersion, cpuArch
 				wg.Done()
 				h.sem.Release(1)
 			}()
-			existsInCache := false
+			existsInConfiguration := false
```
carbonin, line 369: I think don't change these names here because the config is acting as a cache in this case.

internal/versions/versions.go
```diff
@@ -153,9 +383,140 @@ func (h *handler) GetReleaseImage(ctx context.Context, openshiftVersion, cpuArch
...
+
+	if releaseImage != nil {
+		return releaseImage, nil
+	}
+
+	// Then if not found, from configuration
+	image, err = h.getReleaseImageFromConfig(*majorMinorVersion, cpuArchitecture, 2, multiTargetRelease)
+	if err != nil {
+		return nil, err
+	}
```
carbonin, line 480: We should have a clear priority here regardless if we are looking for X.Y.Z or just X.Y.

Right now we're prioritizing the config when we're looking for and exact version but DB comes first when we're looking for latest.

internal/versions/versions.go
```diff
@@ -92,37 +121,238 @@ func (h *handler) GetMustGatherImages(openshiftVersion, cpuArchitecture, pullSec
...
+	releaseModel.SupportLevel = *newSuppotLevel
+
+	return &releaseModel, nil
+}
+
+func (h *handler) getReleaseImageFromDB(
+	openshiftVersion,
+	cpuArchitecture string,
+	versionSegmentsLength int,
```
carbonin, line 260: Maybe create named types for the options. Looking at the caller it's really hard to tell what this param is.

internal/versions/versions.go
```diff
@@ -166,57 +527,108 @@ func (h *handler) GetReleaseImageByURL(ctx context.Context, url, pullSecret stri
...
+		}
+	}
+
+	return latestRelease, nil
+}
+
+func (h *handler) getReleaseImageFromConfig(
+	openshiftVersion,
+	cpuArchitecture string,
+	versionSegmentsLength int,
```
carbonin, line 590: Same here

carbonin (commented):
I think this would be simplified greatly if we use the database for all use cases rather than continuing to use an in-memory cache for kube-api use cases while that same map is used for additional release images (not provided by the sources) in the SaaS. That way each source of release images can be more targeted.

I think we should have the thread that syncs release images from the openshift release API to the database table, but only for SaaS. The envvar should be renamed to something that indicates it's more about temporary or extra release images (that are not returned from the release API), this also wouldn't be used on-prem. These extra release images should by synced to the db table on startup.

For on-prem I think we should create a controller that watches clusterImageSets and syncs the release image to the same database table.

With all of this, we shouldn't have to look anywhere but the database at runtime for any deployment mode.

PR #5716, opened 2023-11-19
MGMT-15425: Change chosen MCE release to match OCP version

danmanor:
Change chosen `MCE` release to match `OCP` version

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/operators/mce/config.go
```diff
@@ -1,8 +1,66 @@
...
+		OpenshiftVersion string   `json:"openshift_version"`
+		MceVersions      []string `json:"mce_versions"`
+	}
+
+	baseVersion, err := common.GetBaseVersion(openshiftVersion)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get base version of %s", openshiftVersion)
+	}
+
+	content, err := jsonFile.ReadFile("ocp_mce_release_matrix.json")
```
eifrach, line 27: where this file is from ?

maybe we should put it in a const for  future name chagnes ?

internal/common/version.go
```diff
@@ -57,3 +58,15 @@ func BaseVersionEqual(version1, versionMayEqual string) (bool, error) {
 
 	return v1[0] == v2[0] && v1[1] == v2[1], nil
 }
+
+func GetBaseVersion(version string) (*string, error) {
```
eifrach, line 62: maybe we should add different test for different version formats?

internal/operators/mce/ocp_mce_release_matrix.json
```diff
@@ -0,0 +1,8 @@
+[
```
eifrach, line 1: I Think this should not be a file, will also removes the code from reading / unmarshal etc..

danmanor: Can you explain "will also removes the code from reading / unmarshal etc..". I think having its own configuration json file will be an advantage as it will need to be updated by some automation.

internal/common/version.go
danmanor: Will do thanks

internal/operators/mce/config.go
danmanor: Will do thanks

internal/operators/mce/ocp_mce_release_matrix.json
adriengentil: If we want it to make it configurable, I think it should be set as a environment variable like we do with the OS_IMAGES and others.

internal/operators/mce/config.go
```diff
@@ -1,8 +1,61 @@
...
+
+	latestVersion := mceVersions[0]
+	for _, version := range mceVersions {
+		isGreaterOrEqaul, err := common.VersionGreaterOrEqual(version, latestVersion)
+		if err != nil {
+			return nil, fmt.Errorf("failed to compare version %s with version %s", latestVersion, version)
+		}
+
+		if isGreaterOrEqaul {
+			latestVersion = version
```
adriengentil, line 50: why keep a list of all the MCE versions for each OCP release if we just take the highest one?

internal/operators/mce/config.go
```diff
@@ -1,8 +1,61 @@
...
+			mceVersions = record.MceVersions
+		}
+	}
+	if len(mceVersions) == 0 {
+		return nil, errors.New("failed to find mce versions for the given openshift version")
+	}
+
+	latestVersion := mceVersions[0]
+	for _, version := range mceVersions {
+		isGreaterOrEqaul, err := common.VersionGreaterOrEqual(version, latestVersion)
```
adriengentil, line 44:
```suggestion
		isGreaterOrEqual, err := common.VersionGreaterOrEqual(version, latestVersion)
```

internal/operators/mce/config.go
danmanor: I thought it might be handy in case you wouldn't want the latest one in the future for some reason

adriengentil: I think we should keep it simple, we can easily improve it later once we'll need it

internal/common/version.go
```diff
@@ -36,24 +38,35 @@ func BaseVersionGreaterOrEqual(version, versionMayGreaterThan string) (bool, err
 }
 
 func BaseVersionLessThan(version, versionMayLessThan string) (bool, error) {
-	isGraterOrEqual, err := BaseVersionGreaterOrEqual(version, versionMayLessThan)
+	isGreaterOrEqual, err := BaseVersionGreaterOrEqual(version, versionMayLessThan)
 	if err != nil {
 		return false, err
 	}
-	return !isGraterOrEqual, nil
+	return !isGreaterOrEqual, nil
```
eifrach, line 45: just nit, make it more readable - doesn't make much differents 

```suggestion
	isGreaterOrEqual, err := BaseVersionGreaterOrEqual(version, versionMayLessThan)
	if err != nil {
		return false, err
	}
	if isGreaterOrEqual { 
	         return false, nil
	}
	return true, nil
```

internal/operators/mce/manifest.go
```diff
@@ -2,11 +2,14 @@ package mce
...
 	"bytes"
+	"fmt"
 	"text/template"
+
+	"github.com/go-openapi/swag"
 )
 
 // Manifests returns manifests needed to deploy MCE.
-func Manifests() (openshiftManifests map[string][]byte, customManifests []byte, err error) {
+func Manifests(openshiftVersion string) (openshiftManifests map[string][]byte, customManifests []byte, err error) {
```
eifrach, line 12: not your doing but, this function has an  return param `err` but it never return's an error

internal/operators/mce/manifest.go
```diff
@@ -34,11 +37,16 @@ func Manifests() (openshiftManifests map[string][]byte, customManifests []byte,
 	return openshiftManifests, mceManifest, nil
 }
 
-func getSubscription() ([]byte, error) {
+func getSubscription(openshiftVersion string) ([]byte, error) {
+	mceVersion, err := getMCEVersion(openshiftVersion)
+	if err != nil {
+		return nil, err
+	}
+	mceChannel := fmt.Sprintf("stable-%s", swag.StringValue(mceVersion))
```
eifrach, line 45: in MCE this may apply, but the channel format is not standart this may break in the future 

here some examples
```
NAME                                          DISPLAY NAME                                            	DEFAULT CHANNEL
3scale-operator								Red Hat Integration - 3scale                             	threescale-2.13
advanced-cluster-management					Advanced Cluster Management for Kubernetes               	release-2.8
aci-containers-operator						Cisco ACI Containers Operator                            	stable
odf-operator								OpenShift Data Foundation  							   		stable-4.14
amq-broker-rhel8							Red Hat Integration - AMQ Broker for RHEL 8 (Multiarch)  	7.11.x
```

internal/operators/mce/manifest.go
danmanor: This is an interface function

internal/operators/mce/manifest.go
danmanor: But this is not generic though this is mce code, so its ok no ?

internal/operators/mce/ocp_mce_release_matrix.json
```diff
@@ -0,0 +1,8 @@
...
+    {"openshift_version": "4.10", "mce_versions": ["2.0", "2.1", "2.2", "2.3"]},
+    {"openshift_version": "4.11", "mce_versions": ["2.0", "2.1", "2.2", "2.3"]},
+    {"openshift_version": "4.12", "mce_versions": ["2.1", "2.2", "2.3", "2.4"]},
+    {"openshift_version": "4.13", "mce_versions": ["2.2", "2.3", "2.4"]},
+    {"openshift_version": "4.14", "mce_versions": ["2.3", "2.4"]},
+    {"openshift_version": "4.15", "mce_versions": ["2.4"]}
```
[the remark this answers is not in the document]
danmanor: Can you elaborate ?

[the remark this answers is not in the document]
danmanor: What does it mean to inject custom channels in the ci ?

internal/operators/mce/manifest.go
eifrach: it's OK - just an FYI

internal/operators/mce/manifest.go
eifrach: so lets remove the return err

```suggestion
func Manifests(openshiftVersion string) (openshiftManifests map[string][]byte, customManifests []byte) {
```

danmanor:
```
	namespaceManifest, err := getNamespace()
	if err != nil {
		return
	}
	operatorGroupManifest, err := getOperatorGroup()
	if err != nil {
		return
	}
	operatorSubscriptionManifest, err := getSubscription(openshiftVersion)
	if err != nil {
		return
	}
```

internal/operators/mce/config.go
danmanor: Ack.

internal/operators/mce/ocp_mce_release_matrix.json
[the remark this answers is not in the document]
danmanor: Ack. Can you explain to me how is this going to work please ? 
(the injection)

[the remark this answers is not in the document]
adriengentil: https://github.com/openshift/assisted-service/pull/5716#discussion_r1399462714 if we do that. So, I would work the same way as we do with the OCP version and RHCOS versions.

internal/operators/mce/ocp_mce_release_matrix.json
danmanor: I thought about it, but because it is not in the same package it will require some changes to give access to this variable, which I thought is not worth it as we already do similar things in bump_ocp...

internal/operators/mce/ocp_mce_release_matrix.json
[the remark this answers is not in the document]
danmanor: Oh I see what you mean now

internal/operators/mce/ocp_mce_release_matrix.json
danmanor: @adriengentil WDYT? Does it worth passing around this var for it ?

adriengentil: I think it's worth because adding/updating a new channel will just require a config change (in app-sre fo example) and not a complete rollout of the code.

danmanor: Thank you

internal/operators/mce/mce_operator.go
```diff
@@ -29,13 +29,12 @@ var Operator = models.MonitoredOperator{
...
-func NewMceOperator(log logrus.FieldLogger) *operator {
-	cfg := Config{}
-	err := envconfig.Process(common.EnvConfigPrefix, &cfg)
+func NewMceOperator(log logrus.FieldLogger, config Config) *operator {
+	err := envconfig.Process(common.EnvConfigPrefix, &config)
 	if err != nil {
 		log.Fatal(err.Error())
 	}
-	return newMceOperatorWithConfig(log, &cfg)
+	return newMceOperatorWithConfig(log, &config)
```
adriengentil, line 37: I think we should unmarshall the config here, so we can fail the startup of the service quickly if the config is wrong. Also, it's worth to unmarshall it only once, not each time we call `getMCEVersion()`

adriengentil: I guess `NewMceOperator` is called when the assisted-service starts, right?

adriengentil: > I guess NewMceOperator is called when the assisted-service starts, right?

yes, through `NewManager` https://github.com/danmanor/assisted-service/blob/master/cmd/main.go#L339C1-L339C119

danmanor: Ack.

eifrach (approved):

internal/operators/mce/config.go
```diff
@@ -13,6 +36,34 @@ const (
...
 
-type Config struct {
-	MceMinOpenshiftVersion string `envconfig:"MCE_MIN_OPENSHIFT_VERSION" default:"4.10.0"`
+func ParseMCEConfig(config Config) (*ParsedConfig, error) {
+	var parsedMatrix []OCPMCEReleaseMatrix
+	if err := json.Unmarshal([]byte(config.OcpMceReleaseMatrix), &parsedMatrix); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal ocp mce release matrix: %w", err)
+	}
+
+	return &ParsedConfig{
```
adriengentil, line 45: Why two structs? Just return the `OCPMCEReleaseMatrix`, no?

internal/operators/mce/config.go
```diff
@@ -13,6 +36,34 @@ const (
 	SNOMinimumCpu    int64 = 8
 )
 
-type Config struct {
-	MceMinOpenshiftVersion string `envconfig:"MCE_MIN_OPENSHIFT_VERSION" default:"4.10.0"`
+func ParseMCEConfig(config Config) (*ParsedConfig, error) {
```
adriengentil, line 39:
```suggestion
func parseMCEConfig(config Config) (*ParsedConfig, error) {
```
no need to expose it outside of the package

internal/operators/mce/config.go
```diff
@@ -1,8 +1,31 @@
...
+
+	"github.com/openshift/assisted-service/internal/common"
+)
+
+type Config struct {
+	MceMinOpenshiftVersion string `envconfig:"MCE_MIN_OPENSHIFT_VERSION" default:"4.10.0"`
+	OcpMceReleaseMatrix    string `envconfig:"OCP_MCE_RELEASE_MATRIX" default:"[{\"openshift_version\": \"4.10\", \"mce_channel\": \"stable-2.3\"}, {\"openshift_version\": \"4.11\", \"mce_channel\": \"stable-2.3\"}, {\"openshift_version\": \"4.12\", \"mce_channel\": \"stable-2.4\"}, {\"openshift_version\": \"4.13\", \"mce_channel\": \"stable-2.4\"}, {\"openshift_version\": \"4.14\", \"mce_channel\": \"stable-2.4\"}, {\"openshift_version\": \"4.15\", \"mce_channel\": \"stable-2.4\"}]"`
+}
+
+type OCPMCEReleaseMatrix struct {
```
adriengentil, line 17:
```suggestion
type OCPMCEVersionMap struct {
```

it doesn't represent a matrix

internal/operators/mce/config.go
```diff
@@ -13,6 +36,34 @@ const (
...
-type Config struct {
-	MceMinOpenshiftVersion string `envconfig:"MCE_MIN_OPENSHIFT_VERSION" default:"4.10.0"`
+func ParseMCEConfig(config Config) (*ParsedConfig, error) {
+	var parsedMatrix []OCPMCEReleaseMatrix
+	if err := json.Unmarshal([]byte(config.OcpMceReleaseMatrix), &parsedMatrix); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal ocp mce release matrix: %w", err)
+	}
+
+	return &ParsedConfig{
+		MceMinOpenshiftVersion: config.MceMinOpenshiftVersion,
```
adriengentil, line 46: what is the usage of this variable?

internal/operators/mce/config.go
```diff
@@ -1,8 +1,31 @@
...
+	MceMinOpenshiftVersion string `envconfig:"MCE_MIN_OPENSHIFT_VERSION" default:"4.10.0"`
+	OcpMceReleaseMatrix    string `envconfig:"OCP_MCE_RELEASE_MATRIX" default:"[{\"openshift_version\": \"4.10\", \"mce_channel\": \"stable-2.3\"}, {\"openshift_version\": \"4.11\", \"mce_channel\": \"stable-2.3\"}, {\"openshift_version\": \"4.12\", \"mce_channel\": \"stable-2.4\"}, {\"openshift_version\": \"4.13\", \"mce_channel\": \"stable-2.4\"}, {\"openshift_version\": \"4.14\", \"mce_channel\": \"stable-2.4\"}, {\"openshift_version\": \"4.15\", \"mce_channel\": \"stable-2.4\"}]"`
+}
+
+type OCPMCEReleaseMatrix struct {
+	OpenshiftVersion string `json:"openshift_version"`
+	MceChannel       string `json:"mce_channel"`
+}
+
+type ParsedConfig struct {
```
adriengentil, line 22:
```suggestion
type MCEConfig struct {
```

openshift/template.yaml
```diff
@@ -461,6 +463,8 @@ objects:
                 value: ${INSTALLER_CACHE_CAPACITY}
               - name: ENABLE_OKD_SUPPORT
                 value: ${ENABLE_OKD_SUPPORT}
+              - name: OCP_MCE_RELEASE_MATRIX
```
adriengentil, line 466: I think there are also thing that should be updated here: https://github.com/openshift/assisted-service/blob/master/deploy/assisted-service-configmap.yaml

internal/operators/mce/config.go
adriengentil: asked a question about MceMinOpenshiftVersion

internal/operators/mce/config.go
adriengentil: if it's used, it's fine to leave it as a `const`, not need to centralize everything in one place

danmanor: I see this approach in other operators as well, makes sense as it will make it easier to configure it differently in case it will be changed no ?

internal/operators/mce/config.go
danmanor: Because if we are keeping `MceMinOpenshiftVersion` configurable it is necessary for marshaling

internal/operators/mce/config.go
danmanor: But both `Config` and `MCEConfig` are `MCE` config eventually, isn't it more meaningful name ?

internal/operators/mce/mce_operator.go
```diff
@@ -79,12 +84,13 @@ func (o *operator) ValidateCluster(_ context.Context, cluster *common.Cluster) (
 	if err != nil {
 		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{err.Error()}}, nil
 	}
-	minOpenshiftVersionForMce, err = version.NewVersion(o.config.MceMinOpenshiftVersion)
+	MceMinOpenshiftVersion, err := getMinMceOpenshiftVersion(o.config.OcpMceVersionMap)
```
adriengentil, line 87:
```suggestion
	mceMinOpenshiftVersion, err := getMinMceOpenshiftVersion(o.config.OcpMceVersionMap)
```

internal/operators/mce/config.go
```diff
@@ -1,9 +1,29 @@
...
+	OCPMCEVersionMap string `envconfig:"OCP_MCE_VERSION_MAP" default:"[{\"openshift_version\": \"4.10\", \"mce_channel\": \"stable-2.3\"}, {\"openshift_version\": \"4.11\", \"mce_channel\": \"stable-2.3\"}, {\"openshift_version\": \"4.12\", \"mce_channel\": \"stable-2.4\"}, {\"openshift_version\": \"4.13\", \"mce_channel\": \"stable-2.4\"}, {\"openshift_version\": \"4.14\", \"mce_channel\": \"stable-2.4\"}, {\"openshift_version\": \"4.15\", \"mce_channel\": \"stable-2.4\"}]"`
+}
+
+type OcpMceVersionMap struct {
+	OpenshiftVersion string `json:"openshift_version"`
+	MceChannel       string `json:"mce_channel"`
+}
+
+type Config struct {
+	MceMinOpenshiftVersion string
```
adriengentil, line 22: can be removed, no?

internal/operators/mce/mce_operator_test.go
```diff
@@ -76,16 +78,19 @@ var _ = Describe("MCE Operator", func() {
 		)
 	})
 	Context("ValidateCluster", func() {
+		MceMinOpenshiftVersion, err := getMinMceOpenshiftVersion(operator.config.OcpMceVersionMap)
```
adriengentil, line 81:
```suggestion
		mceMinOpenshiftVersion, err := getMinMceOpenshiftVersion(operator.config.OcpMceVersionMap)
```

Makefile
```diff
@@ -323,7 +325,7 @@ deploy-service-requirements: | deploy-namespace deploy-inventory-service-file
 		--disk-encryption-support $(DISK_ENCRYPTION_SUPPORT) --hw-requirements '$(subst ",\",$(HW_REQUIREMENTS))' \
 		--disabled-host-validations "$(DISABLED_HOST_VALIDATIONS)" --disabled-steps "$(DISABLED_STEPS)" \
 		--enable-org-tenancy $(ENABLE_ORG_TENANCY) --enable-org-based-feature-gate $(ENABLE_ORG_BASED_FEATURE_GATES) \
-		$(ALLOW_CONVERGED_FLOW_CMD) $(DISABLE_TLS_CMD)
+		--ocp-mce-version-map $(OCP_MCE_VERSION_MAP) $(ALLOW_CONVERGED_FLOW_CMD) $(DISABLE_TLS_CMD)
```
adriengentil, line 328:
```suggestion
		--ocp-mce-version-map '$(subst ",\",$(OCP_MCE_VERSION_MAP))' $(ALLOW_CONVERGED_FLOW_CMD) $(DISABLE_TLS_CMD)
```

internal/operators/mce/config.go
```diff
@@ -13,6 +32,48 @@ const (
...
+			lowestVersion = version.OpenshiftVersion
+		}
+	}
+
+	return &lowestVersion, nil
+}
+
+func parseMCEConfig(config EnvironmentalConfig) (*Config, error) {
+	var parsedMatrix []OcpMceVersionMap
+	if err := json.Unmarshal([]byte(config.OCPMCEVersionMap), &parsedMatrix); err != nil {
```
adriengentil, line 52:
```suggestion
	if err := json.Unmarshal(config, &parsedConfig); err != nil {
```
should works now, no?

internal/operators/mce/config.go
```diff
@@ -13,6 +32,48 @@ const (
...
+		return nil, fmt.Errorf("failed to unmarshal ocp mce release matrix: %w", err)
```
adriengentil, line 53:
```suggestion
		return nil, fmt.Errorf("failed to unmarshal ocp mce release config: %w", err)
```
remove all reference to "matrix" to avoid any confusions since we replaced it with "map" or "config"

internal/operators/mce/config.go
```diff
@@ -13,6 +32,48 @@ const (
...
+
+	var mceChannel string
+	for _, record := range ocpMceVersionMap {
+		if record.OpenshiftVersion == *baseVersion {
+			mceChannel = record.MceChannel
+		}
+	}
+
+	if mceChannel == "" {
+		return nil, errors.New("failed to find mce channel for the given openshift version")
```
adriengentil, line 75: log the openshift version

internal/operators/mce/config.go
danmanor: No AFAIK it must be of type []bytes and need to match exactly the same

adriengentil:
There is a wierd error in the subscription object:
```  - message: 'constraints not satisfiable: no operators found in package multicluster-engine
      in the catalog referenced by subscription multicluster-engine, subscription
      multicluster-engine exists'
    reason: ConstraintsNotSatisfiable
    status: "True"
    type: ResolutionFailed
```

danmanor:
@adriengentil where do you see it ? thank you

adriengentil:
In the must-gather: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_assisted-service/5716/pull-ci-openshift-assisted-service-master-edge-e2e-metal-assisted-mce-4-15/1728789578434744320/artifacts/e2e-metal-assisted-mce-4-15/assisted-common-gather/artifacts/2023-11-26_15-17-17_265a27c9-f8da-45f5-8068-35994d882a65/

danmanor:
@adriengentil apparently mce-2.4 is no longer available - https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.7/html/clusters/cluster_mce_overview#mce-release-notes

danmanor:
I will change the config accordingly and retest

eifrach (approved):

PR #5728, opened 2023-11-22
MGMT-16273: Allow installing on iSCSI disks on OCI

avishayt:
Allow booting from iSCSI for x86_64 OpenShift versions at least 4.15.0 on the OCI platform.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/hardware/validator.go
+++ b/internal/hardware/validator.go
@@ -16,6 +16,7 @@ import (
 	"github.com/openshift/assisted-service/internal/feature"
 	"github.com/openshift/assisted-service/internal/host/hostutil"
 	"github.com/openshift/assisted-service/internal/operators"
+	"github.com/openshift/assisted-service/internal/provider/registry"
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/assisted-service/pkg/conversions"
 	"github.com/sirupsen/logrus"
@@ -37,7 +38,7 @@ type Validator interface {
 	GetInfraEnvHostRequirements(ctx context.Context, infraEnv *common.InfraEnv) (*models.ClusterHostRequirements, error)
 	DiskIsEligible(ctx context.Context, disk *models.Disk, infraEnv *common.InfraEnv, cluster *common.Cluster, host *models.Host, hostArchitecture string, allDisks []*models.Disk) ([]string, error)
 	ListEligibleDisks(inventory *models.Inventory) []*models.Disk
-	IsValidStorageDeviceType(disk *models.Disk, hostArchitecture string) bool
+	IsValidStorageDeviceType(disk *models.Disk, hostArchitecture string, openshiftVersion string, ociPlatformType bool) bool
 	GetInstallationDiskSpeedThresholdMs(ctx context.Context, cluster *common.Cluster, host *models.Host) (int64, error)
 	// GetPreflightHardwareRequirements provides hardware (host) requirements that can be calculated only using cluster information.
 	// Returned information describe requirements coming from OCP and OLM operators.
@@ -45,7 +46,7 @@ type Validator interface {
 	GetPreflightInfraEnvHardwareRequirements(ctx context.Context, infraEnv *common.InfraEnv) (*models.PreflightHardwareRequirements, error)
 }
 
-func NewValidator(log logrus.FieldLogger, cfg ValidatorCfg, operatorsAPI operators.API) Validator {
+func NewValidator(log logrus.FieldLogger, cfg ValidatorCfg, operatorsAPI operators.API, reg registry.ProviderRegistry) Validator {
 	diskEligibilityMatchers := []*regexp.Regexp{
 		compileDiskReasonTemplate(tooSmallDiskTemplate, ".*", ".*"),
 		compileDiskReasonTemplate(wrongDriveTypeTemplate, ".*", ".*"),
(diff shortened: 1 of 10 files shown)


internal/host/hostcommands/install_cmd.go
```diff
@@ -303,8 +303,12 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
 	}
 
 	for _, disk := range inventory.Disks {
-		if disk.DriveType == models.DriveTypeMultipath && disk.ID == host.InstallationDiskID {
-			installerArgs = append(installerArgs, "--append-karg", "root=/dev/disk/by-label/dm-mpath-root", "--append-karg", "rw", "--append-karg", "rd.multipath=default")
+		if disk.ID == host.InstallationDiskID {
+			if disk.DriveType == models.DriveTypeMultipath {
+				installerArgs = append(installerArgs, "--append-karg", "root=/dev/disk/by-label/dm-mpath-root", "--append-karg", "rw", "--append-karg", "rd.multipath=default")
+			} else if disk.DriveType == models.DriveTypeISCSI {
+				installerArgs = append(installerArgs, "--append-karg", "rd.iscsi.firmware=1", "--append-karg", "ip=ibft")
```
jlebon, line 310: `ip=ibft` is only needed specifically for Oracle Cloud to work around NICs taking too long to get DHCP. I wouldn't use it by default across all bare metal installs that use an iSCSI boot disk.

avishayt: Updated, thank you

adriengentil (approved):

internal/host/hostcommands/install_cmd.go
```diff
@@ -303,8 +303,16 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
 	}
 
 	for _, disk := range inventory.Disks {
-		if disk.DriveType == models.DriveTypeMultipath && disk.ID == host.InstallationDiskID {
-			installerArgs = append(installerArgs, "--append-karg", "root=/dev/disk/by-label/dm-mpath-root", "--append-karg", "rw", "--append-karg", "rd.multipath=default")
+		if disk.ID == host.InstallationDiskID {
+			if disk.DriveType == models.DriveTypeMultipath {
+				installerArgs = append(installerArgs, "--append-karg", "root=/dev/disk/by-label/dm-mpath-root", "--append-karg", "rw", "--append-karg", "rd.multipath=default")
+			} else if disk.DriveType == models.DriveTypeISCSI {
+				installerArgs = append(installerArgs, "--append-karg", "rd.iscsi.firmware=1")
```
jlebon, line 310: Since this PR looks like it's adding generic support for iSCSI on bare metal, I will mention here that auto-configuration via iBFT (used by `rd.iscsi.firmware`) is not always a given. The other way iSCSI configuration is provided is via `netroot=iscsi:...` (see [dracut.cmdline(7)](https://www.man7.org/linux/man-pages/man7/dracut.cmdline.7.html)).

If we want to provide a nice UX for it, we could provide separate fields for e.g. server name, target name, initiator name, etc... and assemble the `netroot` karg ourselves. This is similar to what blivet (Anaconda storage backend) does: https://github.com/storaged-project/blivet/blob/57f1382f19a1390dd2176e4f4772bab89380cfdf/blivet/devices/disk.py#L379-L394.

Alternatively, if the UX allows users to provide custom kargs, then we could skip this for now and users can always directly provide the `netroot` karg themselves as an escape hatch (and having `rd.iscsi.firmware=1` still hardcoded _shouldn't_ hurt, though this code could also auto-detect if one of the custom kargs is `netroot` and skip it I guess).

adriengentil:
/test edge-e2e-oci-assisted
The issue should be fixed by now

internal/host/hostcommands/install_cmd.go
adriengentil: Users can provides custom kargs to the installer though the API (https://github.com/openshift/assisted-service/blob/master/docs/user-guide/install-customization.md#set-the-installer-params). Maybe we can add a word about customizing iSCSI settings in our doc in the future.

adriengentil (approved):

internal/host/hostcommands/install_cmd.go
avishayt: That is a good point about generic support.  I'll make this PR focused on OCI.  We can explore generic support in the future if there is demand for it.

internal/hardware/validator.go
```diff
@@ -147,8 +157,8 @@ func (v *validator) DiskIsEligible(ctx context.Context, disk *models.Disk, infra
 	return notEligibleReasons, nil
 }
 
-func (v *validator) IsValidStorageDeviceType(disk *models.Disk, hostArchitecture string) bool {
-	return funk.ContainsString(v.getValidDeviceStorageTypes(hostArchitecture), string(disk.DriveType))
+func (v *validator) IsValidStorageDeviceType(disk *models.Disk, hostArchitecture string, openshiftVersion string, ociPlatformType bool) bool {
```
adriengentil, line 160:
```suggestion
func (v *validator) IsValidStorageDeviceType(disk *models.Disk, hostArchitecture string, openshiftVersion string, isISCSIEnabled bool) bool {
```
nit: Even though this change is specific to oci, I would prefer to make the intention of this parameter clearer for this function and in `getValidDeviceStorageTypes`. WDYT?

adriengentil (commented):
Just one nit, otherwise it looks good

internal/hardware/validator.go
avishayt: I think it's OK for now.  The next big change to this code will probably be when iSCSI support will be generic, and this parameter will be removed.

PR #5732, opened 2023-11-23
MGMT-15690: add support for external platform

adriengentil:
- Define external platform type with its configration options in swagger
- Check usage of external platform against a capability
- Change the registry logic to take the platform struct instead of the platform type, this is required for "supported" external platform like "oci" where platform type will be "external" and platform name is "oci"
  - platform.type = oci and platform.type = external + platform.external.platform_name = oci behave in the same way
- The support of plaform.type = "oci" will be removed in a follow-up PR, once UI and ABI will move to this external API


internal/common/common_unitest_db.go
```diff
@@ -308,6 +308,8 @@ func getDBContext() DBContext {
 		return gDbCtx
 	}
 
+	return &NoDBContext{}
```
adriengentil, line 311: To be removed

internal/common/testing/mock_matchers.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (e eqPlatformTypeMatcher) Matches(x interface{}) bool {
+	platform, ok := x.(*models.Platform)
+	if !ok {
+		return false
+	}
+
+	toto := (e.platformType == *platform.Type)
+	return toto
```
adriengentil, line 21:
```suggestion
	return e.platformType == *platform.Type
```

adriengentil:
/close
will re-open with a smaller scope

PR #5738, opened 2023-11-27
MGMT-15690: Add support for external platform

adriengentil:
- Add a new platform type `external` and new struct `External` under `Platform` in order to configure it
- Add validations to the `External` struct and ability to create/update it:
  - `platformName`: is required and must not be empty
  - `cloudControllerManager`: set to an empty string (CCM disabled) if not set explicitly
- Add feature support, OCP 4.14 and user managed networking is required to install a cluster with external platform
- This new platform is behind an OCM capability, disabled by default

I will plug external platform/platformName=oci with the existing `oci` integration in a follow-up PR.


adriengentil:
Created a SNO cluster with:
```
aicli -U http://10.1.155.26:8080/ create cluster ext-sno -Ppull_secret=~/pull-secret -Popenshift_version=4.14 -Puser_managed_networking=True -Pplatform='{"type":"external", "external": {"platform_name": "agdcorp" , "cloud_controller_manager": ""}}' -Psno=True
```

Install config is updated accordingly:
```
...
  "platform": {
    "external": {
      "PlatformName": "agdcorp",
      "CloudControllerManager": ""
    }
  },
...
```

And the infrastructure object of the cluster using the external integration without CCM as expected:
```
apiVersion: config.openshift.io/v1
kind: Infrastructure
metadata:
  creationTimestamp: "2023-11-27T15:39:13Z"
  generation: 1
  name: cluster
  resourceVersion: "527"
  uid: 391f1b27-77ea-4696-a299-e99e71a694f2
spec:
  cloudConfig:
    name: ""
  platformSpec:
    external:
      platformName: agdcorp
    type: External
status:
  apiServerInternalURI: https://api-int.ext-sno.karmalabs.corp:6443
  apiServerURL: https://api.ext-sno.karmalabs.corp:6443
  controlPlaneTopology: SingleReplica
  cpuPartitioning: None
  etcdDiscoveryDomain: ""
  infrastructureName: ext-sno-gprfr
  infrastructureTopology: SingleReplica
  platform: External
  platformStatus:
    external:
      cloudControllerManager:
        state: None
    type: External

```

adriengentil:
/retest
looks like flakes in prow

internal/bminventory/inventory.go
```diff
@@ -2543,6 +2543,17 @@ func (b *bareMetalInventory) updatePlatformParams(params installer.V2UpdateClust
...
+			if params.ClusterUpdateParams.Platform.External != nil && params.ClusterUpdateParams.Platform.External.PlatformName != nil {
+				updates["platform_external_platform_name"] = params.ClusterUpdateParams.Platform.External.PlatformName
+			}
+			if params.ClusterUpdateParams.Platform.External != nil && params.ClusterUpdateParams.Platform.External.CloudControllerManager != nil {
+				updates["platform_external_cloud_controller_manager"] = params.ClusterUpdateParams.Platform.External.CloudControllerManager
+			}
+		} else {
+			updates["platform_external_platform_name"] = nil
+			updates["platform_external_cloud_controller_manager"] = nil
+		}
```
rccrdpccl, line 2556: else condition is never necessary

internal/provider/common.go
```diff
@@ -140,10 +266,14 @@ func getUpdateParamsForPlatformBM(platform *models.Platform, userManagedNetworki
...
 	if (userManagedNetworking == nil || *userManagedNetworking) && isUMNMandatoryForPlatform(platform) {
 		// userManagedNetworking is already set to true, nothing to do
-		if platform == nil || *cluster.Platform.Type == *platform.Type {
+		if platform == nil || platform.Type == nil {
+			return nil, nil, nil
+		} else if *cluster.Platform.Type == *platform.Type && *cluster.Platform.Type == models.PlatformTypeExternal {
+			return checkExternalPlatformUpdate(*platform, cluster), nil, nil
+		} else if *cluster.Platform.Type == *platform.Type {
 			return nil, nil, nil
 		} else {
```
rccrdpccl, line 275: those conditions are returning early, wouldn't it be clearer if we remove the `else if/else` part?

internal/provider/common.go
```diff
@@ -62,6 +59,109 @@ func isUMNMandatoryForCluster(cluster *common.Cluster) bool {
...
+			return validateExternalSettingsForUpdate(*platform)
+		} else if areExternalSettingsSet(*platform) {
+			// external setting shouldn't be set if platform type is not external
+			return common.NewApiError(http.StatusBadRequest, errors.Errorf("External settings can only be set with external platform type"))
+		}
+	}
+
+	return nil
+}
+
```
rccrdpccl, line 164: Why are we placing this in "common" file? Wouldn't be better to split it and create a `validations.go` and maybe `utils.go` (or maybe leaving as common) so each file has clear responsibilities?

internal/bminventory/inventory.go
adriengentil: why? I need to clean-up the external settings when the platform type isn't `external`

internal/provider/common.go
adriengentil: How do you think we can remove it?

internal/provider/common.go
adriengentil: right, I can move the validations to another file

internal/provider/common.go
adriengentil: I will move it to a more generic function `checkPlatformUpdate`, it's required to no set the usages each time we update the platform type. I really don't know if we need to keep that behavior as it makes things complicated, but I'll let it like that for the moment.

rccrdpccl:
```suggestion
	if platform == nil || platform.Type == nil {
		return nil, nil, nil
	}
        if *cluster.Platform.Type == *platform.Type && *cluster.Platform.Type == models.PlatformTypeExternal {
		return checkExternalPlatformUpdate(*platform, cluster), nil, nil
	}
	if *cluster.Platform.Type == *platform.Type {
		return nil, nil, nil
	}
	return updatePlatformIsExternal(platform), nil, nil

```

internal/bminventory/inventory.go
rccrdpccl: AFAIU something like this would be equivalent, but easier to follow IMO

```suggestion
func (b *bareMetalInventory) updatePlatformParams(params installer.V2UpdateClusterParams, updates map[string]interface{}, usages map[string]models.Usage) error {
	if params.ClusterUpdateParams.Platform == nil || common.PlatformTypeValue(params.ClusterUpdateParams.Platform.Type) == "" {
		return nil
	}

	updates = getUpdatesForPlatformParams(params, updates)
	err := b.providerRegistry.SetPlatformUsages(
		common.PlatformTypeValue(params.ClusterUpdateParams.Platform.Type), usages, b.usageApi)
	if err != nil {
		return fmt.Errorf("failed setting platform usages, error is: %w", err)
	}
	return nil
}

func getUpdatesForPlatformParams(params installer.V2UpdateClusterParams, updates map[string]interface{}) map[string]interface{} {
	updates["platform_type"] = params.ClusterUpdateParams.Platform.Type
	updates["platform_is_external"] = swag.BoolValue(params.ClusterUpdateParams.Platform.IsExternal)
        if *params.ClusterUpdateParams.Platform.Type != models.PlatformTypeExternal {
		updates["platform_external_platform_name"] = nil
		updates["platform_external_cloud_controller_manager"] = nil
		return updates
        }
	if params.ClusterUpdateParams.Platform.External != nil && params.ClusterUpdateParams.Platform.External.PlatformName != nil {
		updates["platform_external_platform_name"] = params.ClusterUpdateParams.Platform.External.PlatformName
	}
	if params.ClusterUpdateParams.Platform.External != nil && params.ClusterUpdateParams.Platform.External.CloudControllerManager != nil {
		updates["platform_external_cloud_controller_manager"] = params.ClusterUpdateParams.Platform.External.CloudControllerManager
	}
	return updates
}
```

internal/bminventory/inventory.go
```diff
@@ -2543,6 +2543,17 @@ func (b *bareMetalInventory) updatePlatformParams(params installer.V2UpdateClust
 	if params.ClusterUpdateParams.Platform != nil && common.PlatformTypeValue(params.ClusterUpdateParams.Platform.Type) != "" {
 		updates["platform_type"] = params.ClusterUpdateParams.Platform.Type
 		updates["platform_is_external"] = swag.BoolValue(params.ClusterUpdateParams.Platform.IsExternal)
+		if *params.ClusterUpdateParams.Platform.Type == models.PlatformTypeExternal {
```
gamli75, line 2546: sometime we use common.PlatformTypeValue(params.ClusterUpdateParams.Platform.Type) and sometimes we use  *params.ClusterUpdateParams.Platform.Type - why?

internal/bminventory/inventory.go
```diff
@@ -2543,6 +2543,17 @@ func (b *bareMetalInventory) updatePlatformParams(params installer.V2UpdateClust
...
+			if params.ClusterUpdateParams.Platform.External != nil && params.ClusterUpdateParams.Platform.External.PlatformName != nil {
+				updates["platform_external_platform_name"] = params.ClusterUpdateParams.Platform.External.PlatformName
```
gamli75, line 2548: do we have consts for platform_external_platform_name, platform_external_cloud_controller_manager?

swagger.yaml
```diff
@@ -6221,6 +6225,28 @@ definitions:
       - vsphere
       - none
       - oci
+      - external
```
gamli75, line 6228: I thought that we are going to remove oci, @avishayt ?

adriengentil: I'll come in a follow-up PR, we need to plug the oci code to the external/platformName=oci, then I will be able to the remove oci platform type

internal/bminventory/inventory.go
adriengentil: no, those are columns names in DB, looks like we never defined any consts for them in the past :thinking:

internal/bminventory/inventory.go
adriengentil: `common.PlatformTypeValue()` returns an string empty string when `Platform.Type` is `nil`.

internal/provider/common.go
```diff
@@ -62,6 +59,109 @@ func isUMNMandatoryForCluster(cluster *common.Cluster) bool {
...
+func areExternalSettingsSet(platform models.Platform) bool {
+	return platform.External != nil && (platform.External.PlatformName != nil || platform.External.CloudControllerManager != nil)
+}
+
+func validateExternalPlatformName(platformName *string) bool {
+	return platformName != nil && len(*platformName) > 0
+}
+
+func validateExternalCloudControllerManager(cloudControllerManager string) bool {
```
avishayt, line 102: This is already defined in the swagger, do we need this here too?

adriengentil: true, for some reasons it was not working during early testings, that's why I added the check. But I saw the swagger validation work yesterday.

internal/provider/common.go
adriengentil: updated !

pawanpinjarkar:
@adriengentil I successfully tested SNO IPV4 cluster installation for the external platform with the below config in the install-config via dev-scripts. 
```
"platform": {
        "external": {
           "platformName": "nonoci"
            }
    },
```

Ref installer PR https://github.com/openshift/installer/pull/7585

adriengentil:
/retest
looks like flakes

adriengentil:
/retest
flakes :(

adriengentil:
@avishayt @gamli75 @rccrdpccl are we good to go with this PR?

adriengentil:
/override ci/prow/edge-e2e-nutanix-assisted ci/prow/edge-e2e-nutanix-assisted-4-14 ci/prow/edge-e2e-vsphere-assisted
nutanix jobs is broken
vsphere job is flaky

rccrdpccl:
/lgtm
/hold

Unhold whenever we're ready :)

PR #5740, opened 2023-11-27
MGMT-16289: Add configuration variable and cluster attribute defining if soft timeouts feature is enabled

ori-amizur:

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @filanov

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -165,6 +165,9 @@ var Options struct {
 
 	// Directory containing pre-generated TLS certs/keys for the ephemeral installer
 	ClusterTLSCertOverrideDir string `envconfig:"EPHEMERAL_INSTALLER_CLUSTER_TLS_CERTS_OVERRIDE_DIR" default:""`
+
+	// EnableSoftTimeouts is a boolean flag to enable Soft timeouts by assisted installer
+	EnableSoftTimeouts bool `envconfig:"ENABLE_SOFT_TIMEOUTS" default:"false"`
 }
 
 func InitLogs(logLevel, logFormat string) *logrus.Logger {


swagger.yaml
```diff
@@ -5003,6 +5003,10 @@ definitions:
         type: string
         description: A comma-separated list of tags that are associated to the cluster.
         x-nullable: true
+      enable_soft_timeouts:
```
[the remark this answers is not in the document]
ori-amizur: OK

cmd/main.go
```diff
@@ -137,7 +138,6 @@ var Options struct {
 	ValidationsConfig                    validations.Config
 	ManifestsGeneratorConfig             network.Config
 	UploaderConfig                       uploader.Config
-	EnableKubeAPI                        bool `envconfig:"ENABLE_KUBE_API" default:"false"`
```
filanov, line 140: why did you made this change as part of this pr?

ori-amizur: I moved it to common.Config, so it is together with the soft timeouts configuration.
This gives the ability to call the function `Config.IsSoftTimeoutsEnabled` which needs both of these variables.

filanov: why not make `ENABLE_SOFT_TIMEOUTS` with a default false and then enable it only in SAAS?

ori-amizur: I wanted to distinguish between cases that it is not set, it is set to false, or set to true.
If it will be default false, and enable it in SAAS, there will be no way to disable in SAAS.

filanov: why not? If there is a bug that we want to avoid until this feature is stable we should be able to disable no?

ori-amizur: But this is your suggestion

ori-amizur: If the default is "false", there will not be a way to distinguish between when the variable value is received as default value and it is explicitly set to "false".  Therefore for SAAS, since we don't know to distinguish between these 2 cases, we should always set the value to true, which means that for SAAS it can never be disabled.

internal/common/config.go
```diff
@@ -0,0 +1,17 @@
+package common
+
+import "github.com/go-openapi/swag"
+
+type Config struct {
+	EnableKubeAPI bool `envconfig:"ENABLE_KUBE_API" default:"false"`
+
+	// EnableSoftTimeouts is a boolean flag to enable Soft timeouts by assisted installer
+	EnableSoftTimeouts *bool `envconfig:"ENABLE_SOFT_TIMEOUTS"`
```
filanov, line 9: prefer to use  a single variable that will be disabled by default, then we will not need to update the CI because no one will try to resolve installation issues in CI

filanov (approved):

PR #5751, opened 2023-11-29
MGMT-16312 fix update host ignition for unbound host

filanov:
Becuase of a recover function in V2UpdateHostIgnition if a panic occurs
then a controller will get nil as a reply and assume that the call
didn't fail, it casued a hot loop that tried to update the ignition
again and again, because and event was called before the panic and
before the ignition was commited we ended up with 95 million events
before we ran out of space on the disk

This commit will remove the recover in order to return a proper error
and fail the service if tehre is a painc

The pacin was caused by calling setIgnitionConfigOverrideUsage and
assuming that cluster ID exists, added a validation for the cluster id
before using it.

added unit tests to check functionality to unbound host

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -5813,60 +5813,56 @@ func (b *bareMetalInventory) V2UpdateHostIgnition(ctx context.Context, params in
 func (b *bareMetalInventory) V2UpdateHostIgnitionInternal(ctx context.Context, params installer.V2UpdateHostIgnitionParams) (*models.Host, error) {
 	log := logutil.FromContext(ctx, b.log)
 
-	txSuccess := false
-	tx := b.db.Begin()
-	defer func() {
-		if !txSuccess {
-			log.Error("UpdateHostIgnition failed")
-			tx.Rollback()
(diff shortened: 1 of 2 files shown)


ori-amizur (commented):
Just an idea - Maybe we should move to gorm.Transaction function (also in other locations).  It could eliminate some of the problems - especially the `recover` code .

filanov:
> Just an idea - Maybe we should move to gorm.Transaction function (also in other locations). It could eliminate some of the problems - especially the `recover` code .

will start with this function, removing the recover is handled by MGMT-16157 so it will be done in a different PR.

filanov:
/retest 
infra issues

filanov:
/hold 
found the issue that cause the panic, will update this PR

PR #5757, opened 2023-12-01
OCPBUGS-24319: fix: allow LVMS to be installed for MultiNode / HA Configurations in OpenShift >=4.15.0

jakobmoellerdev:
With https://issues.redhat.com/browse/OCPEDGE-64 LVMS will officially support HA configurations out of the box. Currently the assisted-service logic still blocks these installations. I added a flag for a minimum HA version support for 4.15.0 and a test case so that we can make sure this is supported in the future with assisted-service as well and not blocked.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


jakobmoellerdev:
@eifrach anything else necessary here?

internal/operators/lvm/lvm_operator.go
```diff
@@ -95,6 +91,16 @@ func (o *operator) ValidateCluster(_ context.Context, cluster *common.Cluster) (
...
+	minOpenshiftVersionForMultiNodeSupport, err = version.NewVersion(o.config.LvmMinMultiNodeSupportVersion)
+	if err != nil {
+		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{err.Error()}}, nil
+	}
+	if ocpVersion.LessThan(minOpenshiftVersionForMultiNodeSupport) && !common.IsSingleNodeCluster(cluster) {
+		message := fmt.Sprintf("Logical Volume Manager is only supported for highly available openshift with version %s or above",
+			minOpenshiftVersionForMultiNodeSupport.String())
+		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetClusterValidationID(), Reasons: []string{message}}, nil
+	}
+
```
eifrach, line 103: small nit, 

maybe we should wrap it with if, and check only what is necessary for the case
no need to check the version twice 
```suggestion
	if common.IsSingleNodeCluster(cluster) {
		minOpenshiftVersionForLvm, err = version.NewVersion(o.config.LvmMinOpenshiftVersion)
		if err != nil {
			return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{err.Error()}}, nil
		}
		if ocpVersion.LessThan(minOpenshiftVersionForLvm) {
			message := fmt.Sprintf("Logical Volume Manager is only supported for openshift versions %s and above", o.config.LvmMinOpenshiftVersion)
			return api.ValidationResult{Status: api.Failure, ValidationId: o.GetClusterValidationID(), Reasons: []string{message}}, nil
		}
	} else {
		minOpenshiftVersionForMultiNodeSupport, err = version.NewVersion(o.config.LvmMinMultiNodeSupportVersion)
		if err != nil {
			return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{err.Error()}}, nil
		}
		message := fmt.Sprintf("Logical Volume Manager is only supported for highly available openshift with version %s or above",
			minOpenshiftVersionForMultiNodeSupport.String())
		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetClusterValidationID(), Reasons: []string{message}}, nil
	}
```

eifrach:
> @eifrach anything else necessary here?
sry, I was looking if there is any other tests we can use here .

added a small nit, but looks good to me

internal/operators/lvm/lvm_operator.go
jakobmoellerdev: I believe this logic works because minOpenshiftVersionForMultiNodeSupport is always greater than minOpenshiftVersionForLvm, I wrote the logic without this assumption.

Good catch, will correct.

PR #5769, opened 2023-12-04
MGMT-16199: block unbind host while cluster is progressing

eifrach:

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
 	yamlpatch "github.com/krishicks/yaml-patch"
+	"github.com/openshift/assisted-service/internal/constants"
 	"github.com/openshift/assisted-service/models"
 	"github.com/thoas/go-funk"
 	"gorm.io/gorm"
@@ -100,8 +101,10 @@ const NMDebugModeConf = `
 domains=ALL:DEBUG
 `
 
-const ValidationTypeHost = "host"
-const ValidationTypeCluster = "cluster"
+const (
+	ValidationTypeHost    = "host"
+	ValidationTypeCluster = "cluster"
+)
 
 func GetIgnoredValidations(validationsJSON string, clusterID string) ([]string, bool) {
 	ignoredValidations := []string{}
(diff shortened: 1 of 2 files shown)


avishayt:
According to SourceStates, these transitions don't apply.  The bug isn't about unbind, it's about unregister.  If you look at UnRegisterHost() in internal/host/host.go, it just deletes the host from the DB without any checks.  We should add the check here (return 409).

eifrach:
@avishayt  changed, should this solve it?

internal/host/host.go
```diff
@@ -1441,14 +1441,28 @@ func (m *Manager) GetHostByKubeKey(key types.NamespacedName) (*common.Host, erro
...
+	}
+	var cluster common.Cluster
+	err := m.db.Select("status").Take(&cluster, "id = ?", h.ClusterID.String()).Error
+	if err != nil {
 		return err
 	}
-	if h.ClusterID != nil {
-		if err := m.db.Model(&common.Cluster{}).Where("id = ?", h.ClusterID).Update("trigger_monitor_timestamp", time.Now()).Error; err != nil {
-			return err
+	clusterProgressingStates := []string{models.ClusterStatusInstalling, models.ClusterStatusFinalizing}
```
gamli75, line 1452: I think we should block it also in case of preparing-for-installation, @avishayt WDYT?

avishayt: Yes, and also installing-pending-user-action.  I suggest putting this list of states into internal/cluster/cluster.go, similar to CanDownloadFiles() or CanDownloadKubeconfig().

Another thing, while looking at that file, I see three places where we check for ClusterStatusInstalling but ignore ClusterStatusInstallingPendingUserAction:
1. CancelInstallation() doesn't allow canceling while in installing-pending-user-action
2. IsOperatorAvailable() will return false while in installing-pending-user-action
3. SetVipsData() won't catch VIPs changing during installation

eifrach: @avishayt so we can't move it to cluster package due to import loop issues
so I've create it in common

let me know if it's ok i will start working on the test's

internal/common/common.go
```diff
@@ -432,6 +435,21 @@ func GetInventoryInterfaces(inventory string) (string, error) {
 	return inventory[interfacesLocation : interfacesLocation+endLocation], nil
 }
 
+// blocking host removal while cluster is progressing
+func CanUnbindhost(c *Cluster) (err error) {
+	clusterStatus := swag.StringValue(c.Status)
+	NotAllowedStatuses := []string{
+		models.ClusterStatusFinalizing,
```
avishayt, line 442: Also preparing-for-installation

avishayt (commented):
Looks OK.  There are a bunch of unrelated formatting changes here.  I think it's better to not commit them for the sake of clarity.  If there is some formatting convention you feel strongly about, we can add consider adding it to the linter.

eifrach:
> Looks OK. There are a bunch of unrelated formatting changes here. I think it's better to not commit them for the sake of clarity. If there is some formatting convention you feel strongly about, we can add consider adding it to the linter.

So the reason is that I'm running `skipper make lint` before pushing 

the linter is also fixing where it can, not sure why I am the only one doing it :smiley: 

I'll recreate the PR

eifrach:
recreated #5809

/close

PR #5790, opened 2023-12-08
NO-ISSUE: fix golangci-lint failing to be pulled

adriengentil:
app-sre made their repo mirroring golangci-lint private for legal reasons, and
now our jobs are failing to install it. Installing it now from the cli.

We could also have pulled golangci-lint from dockerhub, but rate
limitation there is quite aggressive.


gamli75 (approved):

adriengentil:
/retitle MGMT-16370: fix golangci-lint failing to be pulled

PR #5795, opened 2023-12-10
MGMT-16157 Remove recover usage from DB transactions

filanov:
The downside of recover is that it may hide bugs like accessing nil pointer and can create a situation where a function panic and instead of returning an error it will return nil, in some cases it cause a hot loop in our controllers, or just move a bug to an upper later.

To avoid those situations we decided to remove the recover function usage in the transaction code and move to a new transaction style that will make scope of the transaction and handling of errors clearer.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -11,7 +11,6 @@ import (
 	"net"
 	"net/http"
 	"runtime"
-	"runtime/debug"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


filanov:
I didn't removed all the old transaction style code because this is not the purpose of this PR, i will do it in another ticket.

filanov:
An example of a code that can create strange issues

```golang
package main

import (
	"fmt"
)

func Foo() (*int, error) {
	i := 5
	var b *int

	defer func() {
		if r := recover(); r != nil {
			fmt.Println("foo recover")
		}
	}()

	fmt.Println("create a panic", *b)
	return &i, nil
}

func main() {
	i, err := Foo()
	if err != nil {
		panic("had an error")
	}

	fmt.Println(*i)
}
```

PR #5809, opened 2023-12-13
MGMT-16199: block unbind host while cluster is progressing

eifrach:
bugfix
-  remove the ability to unbind Host while the cluster is progressing 
- allow removing host's which are not bound to a cluster , for ACM 

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -440,7 +440,7 @@ func CanUnbindhost(c *Cluster) (err error) {
 	clusterStatus := swag.StringValue(c.Status)
 	NotAllowedStatuses := []string{
 		models.ClusterStatusFinalizing,
-		models.ClusterStatusInstalled,
+		models.ClusterStatusInstalling,
 		models.ClusterStatusInstallingPendingUserAction,
 		models.ClusterStatusPreparingForInstallation,
 	}
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
 	yamlpatch "github.com/krishicks/yaml-patch"
+	"github.com/openshift/assisted-service/internal/constants"
 	"github.com/openshift/assisted-service/models"
 	"github.com/thoas/go-funk"
 	"gorm.io/gorm"
@@ -100,8 +101,10 @@ const NMDebugModeConf = `
 domains=ALL:DEBUG
 `
 
-const ValidationTypeHost = "host"
-const ValidationTypeCluster = "cluster"
+const (
+	ValidationTypeHost    = "host"
+	ValidationTypeCluster = "cluster"
+)
 
 func GetIgnoredValidations(validationsJSON string, clusterID string) ([]string, bool) {
 	ignoredValidations := []string{}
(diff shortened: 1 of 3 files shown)


internal/common/common_test.go
```diff
@@ -466,6 +466,55 @@ var _ = Describe("JSON serialization checks", func() {
...
+				models.ClusterStatusFinalizing,
+				models.ClusterStatusInstalling,
+				models.ClusterStatusInstallingPendingUserAction,
+				models.ClusterStatusPreparingForInstallation,
+			}
+			for _, failState := range notAllowedStatus {
+				cluster.Status = swag.String(failState)
+				err := CanUnbindhost(&cluster)
+
+				Expect(err).ToNot(BeNil())
```
paul-maidment, line 496: Whereas this checks that there is error handling generally it doesn't verify that the error here https://github.com/openshift/assisted-service/pull/5809/files#diff-e0a57b903e60761f673a4485b4a747f500f4155828e182cc19786ee5648cd6e6R448 was the one emitted. 

I'd personally update the test to verify that the expected error is returned.

internal/common/common.go
```diff
@@ -432,6 +435,22 @@ func GetInventoryInterfaces(inventory string) (string, error) {
 	return inventory[interfacesLocation : interfacesLocation+endLocation], nil
 }
 
+// blocking host removal while cluster is progressing
+func CanUnbindhost(c *Cluster) (err error) {
+	clusterStatus := swag.StringValue(c.Status)
+	NotAllowedStatuses := []string{
+		models.ClusterStatusFinalizing,
+		models.ClusterStatusInstalling,
+		models.ClusterStatusInstallingPendingUserAction,
```
tsorya, line 444: I can think about scenario were i would actually like to remove node in ClusterStatusInstallingPendingUserAction.
Lets say worker is stuck in it and i have more workers wouldn't i want to continue installation and not handle stuck worker? @avishayt ?

paul-maidment: I agree, this seems like a likely use case for wanting to remove the node.

internal/host/host.go
```diff
@@ -1441,15 +1447,23 @@ func (m *Manager) GetHostByKubeKey(key types.NamespacedName) (*common.Host, erro
 }
 
 func (m *Manager) UnRegisterHost(ctx context.Context, h *models.Host) error {
-	if err := common.DeleteHostFromDB(m.db, h.ID.String(), h.InfraEnvID.String()); err != nil {
-		return err
-	}
 	if h.ClusterID != nil {
```
tsorya, line 1450:
```suggestion
	if h.ClusterID == nil {
	     return err
	 }
	
```

internal/common/common.go
avishayt: It leave some room for people to shoot themselves in the foot, but I'd be OK with that compromise.

internal/common/common_test.go
eifrach: correct, it does verify the exact error, just check if we go err back.
I'll added also the error string check 
thanks

internal/common/common.go
eifrach: I think we should advise the user to cancel the installation and then remove nodes.

so when, in `ClusterStatusInstallingPendingUserAction` we are still progressing the installation thus we don't want user to remove nodes.

internal/host/host.go
eifrach: I do return an error at the end.  do you think it better this way ?

internal/host/host.go
```diff
@@ -1441,14 +1447,29 @@ func (m *Manager) GetHostByKubeKey(key types.NamespacedName) (*common.Host, erro
...
+			m.log.Infof("Host %v is not Bound to Cluster.", h.ID)
+		}
 	}
 	if h.ClusterID != nil {
-		if err := m.db.Model(&common.Cluster{}).Where("id = ?", h.ClusterID).Update("trigger_monitor_timestamp", time.Now()).Error; err != nil {
+		var cluster *common.Cluster
+		if err := m.db.Select("status").Take(&cluster, "id = ?", h.ClusterID.String()).Error; err != nil {
+			return fmt.Errorf("failed to get cluster from DB: %+v", err)
+		}
+		if err := common.CanUnbindhost(cluster); err != nil {
```
carbonin, line 1462: This function `UnRegisterHost` is for deleting hosts, not unbinding them.

If you want this to be run during unbind you should call it from `UnbindHost` if it's checking if a host can be deleted it should have a different name.

carbonin (changes requested):

carbonin:
Also it would be much easier to review if you stick to only making changes that are related to your bug fix.

All the formatting changes are unnecessary in this PR.

PR #5811, opened 2023-12-13
MGMT-16241: Ensure that LastInstallationPreparationStatus is reset on cluster install.

paul-maidment:
A recent change to how cluster installation is presented introduced a bug that caused problems when encountering an
error during installation. On fixing the error, the user would have to start the installation twice in order to move
to the `installing` state.

This issue addresses that by resetting the LastInstallationPreparationStatus when an installation is started.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -1222,6 +1222,17 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params 
 		return nil, common.NewApiError(http.StatusNotFound, err)
 	}
 
+	err = b.db.Model(&models.Cluster{}).Where("id = ?", cluster.ID.String()).Updates(&models.Cluster{
+		LastInstallationPreparation: models.LastInstallationPreparation{
+			Status: models.LastInstallationPreparationStatusNotStarted,
+			Reason: "Preparation never performed",
+		},
+	}).Error
+	if err != nil {
+		b.log.WithError(err).Errorf("Failed to reset last installation preparation state for cluster %s", cluster.ID.String())
+		return nil, common.NewApiError(http.StatusInternalServerError, err)
+	}
+
 	var autoAssigned bool
 
 	// auto select hosts roles if not selected yet.
--- a/api/vendor/github.com/openshift/assisted-service/models/last_installation_preparation.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/last_installation_preparation.go
@@ -24,7 +24,7 @@ type LastInstallationPreparation struct {
 	Reason string `json:"reason,omitempty"`
 
 	// The last installation preparation status
-	// Enum: [preparation_never_performed failed success]
+	// Enum: [not_started failed success]
 	Status string `json:"status,omitempty"`
 }
 
@@ -46,7 +46,7 @@ var lastInstallationPreparationTypeStatusPropEnum []interface{}
 
 func init() {
 	var res []string
-	if err := json.Unmarshal([]byte(`["preparation_never_performed","failed","success"]`), &res); err != nil {
+	if err := json.Unmarshal([]byte(`["not_started","failed","success"]`), &res); err != nil {
 		panic(err)
 	}
 	for _, v := range res {
@@ -56,8 +56,8 @@ func init() {
 
 const (
 
-	// LastInstallationPreparationStatusPreparationNeverPerformed captures enum value "preparation_never_performed"
-	LastInstallationPreparationStatusPreparationNeverPerformed string = "preparation_never_performed"
+	// LastInstallationPreparationStatusNotStarted captures enum value "not_started"
+	LastInstallationPreparationStatusNotStarted string = "not_started"
 
 	// LastInstallationPreparationStatusFailed captures enum value "failed"
 	LastInstallationPreparationStatusFailed string = "failed"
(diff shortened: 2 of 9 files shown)


avishayt:
Hi, can you please explain more about the problem and how you're trying to fix it?  I'm not 100% clear on the flow.

Would the problem be solved if we reset the existing field when starting a new installation?  If that would solve it, what are the drawbacks?

paul-maidment:
> Hi, can you please explain more about the problem and how you're trying to fix it? I'm not 100% clear on the flow.
> 
> Would the problem be solved if we reset the existing field when starting a new installation? If that would solve it, what are the drawbacks?

The problem would be solved by resetting the existing field when starting a new installation. In fact this PR does that.
However in order to not lose the functionality that was intended, an extra field is needed.

When you are in "KubeAPI" mode and you have an installation failure, the cluster will return to "Ready", eventually installation will automatically start again. It is not very clear to the user what is going on, so in a previous PR one of the actions I took was to add a condition to the cluster that would show you the state of the "Last Installation Preparation", this was based on an existing field and added more context by including the reason for the preparation failure.

Of course, when we start a new installation (automatically) this information is wiped...

It is desirable to keep this information until there is another outcome for installation preparation. The intent of this field is to show the last outcome.

I hope this makes sense?

filanov:
> > Hi, can you please explain more about the problem and how you're trying to fix it? I'm not 100% clear on the flow.
> > Would the problem be solved if we reset the existing field when starting a new installation? If that would solve it, what are the drawbacks?
> 
> The problem would be solved by resetting the existing field when starting a new installation. In fact this PR does that. However in order to not lose the functionality that was intended, an extra field is needed.
> 
> When you are in "KubeAPI" mode and you have an installation failure, the cluster will return to "Ready", eventually installation will automatically start again. It is not very clear to the user what is going on, so in a previous PR one of the actions I took was to add a condition to the cluster that would show you the state of the "Last Installation Preparation", this was based on an existing field and added more context by including the reason for the preparation failure.
> 
> Of course, when we start a new installation (automatically) this information is wiped...
> 
> It is desirable to keep this information until there is another outcome for installation preparation. The intent of this field is to show the last outcome.
> 
> I hope this makes sense?

I don't think that we need to focus on the k8s side with that, let the controller to handle this part, for example if we stop the installation attempt after several tries or delay the installation of the next attempt the user will get all the data. 

So in general one field of the last failure should be enough.

paul-maidment:
> > > Hi, can you please explain more about the problem and how you're trying to fix it? I'm not 100% clear on the flow.
> > > Would the problem be solved if we reset the existing field when starting a new installation? If that would solve it, what are the drawbacks?
> > 
> > 
> > The problem would be solved by resetting the existing field when starting a new installation. In fact this PR does that. However in order to not lose the functionality that was intended, an extra field is needed.
> > When you are in "KubeAPI" mode and you have an installation failure, the cluster will return to "Ready", eventually installation will automatically start again. It is not very clear to the user what is going on, so in a previous PR one of the actions I took was to add a condition to the cluster that would show you the state of the "Last Installation Preparation", this was based on an existing field and added more context by including the reason for the preparation failure.
> > Of course, when we start a new installation (automatically) this information is wiped...
> > It is desirable to keep this information until there is another outcome for installation preparation. The intent of this field is to show the last outcome.
> > I hope this makes sense?
> 
> I don't think that we need to focus on the k8s side with that, let the controller to handle this part, for example if we stop the installation attempt after several tries or delay the installation of the next attempt the user will get all the data.
> 
> So in general one field of the last failure should be enough.

I don't think you fully understand.

I repurposed a field that was not being used for any real communication with the user, this introduced the bug, but this field is necessary.

The Other field is the one for the user.

You need both.

.tekton/assisted-service-q2vh-pull-request.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       - name: IMAGE_URL
         value: $(tasks.build-container.results.IMAGE_URL)
       taskRef:
-        bundle: quay.io/redhat-appstudio-tekton-catalog/task-show-sbom:0.1@sha256:4b8a47b3cd3f80d0a9ec46ec43fd2ebe0487979f351a8eeebe5aa04854d3861c
+        bundle: quay.io/redhat-appstudio-tekton-catalog/task-show-sbom:0.1@sha256:202d3c3385120ea847d8f0a82bd8d9d5e873d67f981d6f8a51fb1706caaf6bef
```
filanov, line 40: What is  the tektone change?

internal/bminventory/inventory.go
```diff
@@ -1222,6 +1222,16 @@ func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params
 		return nil, common.NewApiError(http.StatusNotFound, err)
 	}
 
+	err = b.db.Model(&models.Cluster{}).Where("id = ?", cluster.ID.String()).Updates(&models.Cluster{
```
filanov, line 1225: what will be the side effect if it will fail?

.tekton/assisted-service-q2vh-pull-request.yaml
paul-maidment: I'm guessing this is not meant to be here, will fix

internal/bminventory/inventory.go
paul-maidment: The effect would be that the user finds themselves having to restart the installation twice. Maybe we should hard fail if this field cannot be updated?

filanov: yes it should be a hard failure

paul-maidment: OK, hard failure added for this case.

filanov (approved):

PR #5825, opened 2023-12-18
MGMT-16373: KubeAPI - Ensure that ISO type is full-iso when cpuarchitecture is S390x

paul-maidment:
Presently when an infraenv is created via KubeAPI for the S390x architecture it is created with the default "minimum-iso" imagetype. This is incompatible with S390x.

This PR makes sure that the imagetype is updated to "full-iso" to maintain compatibility.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/infraenv_controller.go
+++ b/internal/controller/controllers/infraenv_controller.go
@@ -140,6 +140,15 @@ func (r *InfraEnvReconciler) Reconcile(origCtx context.Context, req ctrl.Request
 	return r.ensureISO(ctx, log, infraEnv)
 }
 
+func (r *InfraEnvReconciler) getImageType(ctx context.Context, infraEnv *aiv1beta1.InfraEnv) models.ImageType {
+	// S390X platform does not support minimal ISO, ensure that full ISO is requested instead.
+	imageType := r.Config.ImageType
+	if infraEnv.Spec.CpuArchitecture == models.ClusterCPUArchitectureS390x {
+		imageType = models.ImageTypeFullIso
+	}
+	return imageType
+}
+
 func (r *InfraEnvReconciler) updateInfraEnv(ctx context.Context, log logrus.FieldLogger, infraEnv *aiv1beta1.InfraEnv, internalInfraEnv *common.InfraEnv) (*common.InfraEnv, error) {
 	updateParams := installer.UpdateInfraEnvParams{
 		InfraEnvID:           *internalInfraEnv.ID,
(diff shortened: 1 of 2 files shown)


carbonin (changes requested):
This looks like it's missing handling the create case [here](https://github.com/openshift/assisted-service/blob/e56197390fd182c3e43682dd3c9d5c74e7b206bc/internal/controller/controllers/infraenv_controller.go#L479)

Can we maybe test this in subsystem kube API tests?

paul-maidment:
> This looks like it's missing handling the create case [here](https://github.com/openshift/assisted-service/blob/e56197390fd182c3e43682dd3c9d5c74e7b206bc/internal/controller/controllers/infraenv_controller.go#L479)
> 
> Can we maybe test this in subsystem kube API tests?

This is the only location from which a call to the create method is made

https://github.com/openshift/assisted-service/blob/e56197390fd182c3e43682dd3c9d5c74e7b206bc/internal/controller/controllers/infraenv_controller.go#L384

Immediately, and unconditionally, a call is made to `updateInfraEnv` where the change of ISO type is applied.

As far as I can see this leaves no uncovered edge cases.

I can certainly add a subsystem test though...

carbonin:
>As far as I can see this leaves no uncovered edge cases.
>I initially considered making the same change inside the createInfraEnv method but it would make no semantic difference. As this method is private also, I feel it's best to leave it as is.

I disagree. I think it's best not to rely on this pattern. We're setting the image type in create to something incorrect and we shouldn't do that. I could very easily imagine someone making those calls to create and update mutually exclusive in a refactoring and I think they would be correct to do so. 

It should be easy enough to put the image type logic into a helper function and use it in both places.

Either way we should have a test to ensure the image is set correctly for both create and update cases.

paul-maidment:
> > As far as I can see this leaves no uncovered edge cases.
> > I initially considered making the same change inside the createInfraEnv method but it would make no semantic difference. As this method is private also, I feel it's best to leave it as is.
> 
> I disagree. I think it's best not to rely on this pattern. We're setting the image type in create to something incorrect and we shouldn't do that. I could very easily imagine someone making those calls to create and update mutually exclusive in a refactoring and I think they would be correct to do so.
> 
> It should be easy enough to put the image type logic into a helper function and use it in both places.
> 
> Either way we should have a test to ensure the image is set correctly for both create and update cases.

Agreed.

I have made sure we apply the change to image type on create as well as update and added a subsystem test to check this.

carbonin (approved):

PR #5884, opened 2024-01-15
MGMT-16508: Add API to AgentServiceConfig CRD to allow pass of CA certificates for image pull.

paul-maidment:
[MGMT-16508](https://issues.redhat.com//browse/MGMT-16508): Add API to AgentServiceConfig CRD to allow pass of CA certificates for image pull.

This PR introduces a new field to the AgentServiceConfig CRD `imagePullCAConfigMap` which is a LocalObjectReference to a config map containing CA certificates
These certificates are to be used by the image service for the purpose of verifying the CA of HTTPS connections used for pulling images.

`imagePullConfigMap` is expected to reference a ConfigMap containing a single file `extra-ca.pem` which contains the CA certificate.
If multiple CA certificates are to be used, the user is expected to append these into the same PEM.

This code also passes the path `ADDITIONAL_CA_FILE` to the image service if the additional CA file is present.

This code sets up the Volume for this and maps to either an empty directory or maps the content of the ConfigMap.

This will ensure that the CA's will be available for use by the image service.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

- [] This change does not require a documentation update (docstring, `docs`, README, etc)
Docstrings have been updated to describe the purpose of this change.
There is also a ticket in the Epic to support more general documentation.

## Reviewers Checklist


gamli75:
@avishayt  @carbonin please review

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -109,6 +109,12 @@ type AgentServiceConfigSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="List of container registries without authentication"
 	// +optional
 	UnauthenticatedRegistries []string `json:"unauthenticatedRegistries,omitempty"`
+	// ImagePullCAConfigMapRef is a reference to a config map containing a certificate authority certificate
+	// this is an optional certificate to allow a customer to add a certificate authority for a HTTPS source of images
+	// this certificate will be used by the assisted-image-service when pulling OS images.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Image pull CA config map reference"
+	// +optional
+	ImagePullCAConfigMapRef *corev1.LocalObjectReference `json:"imagePullCAConfigMapRef,omitempty"`
```
carbonin, line 117: I think this should reference `OSImages` in some way to make sure it's not confused with pulling container images.

Maybe `OSImageCACertRef`?

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -109,6 +109,12 @@ type AgentServiceConfigSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="List of container registries without authentication"
 	// +optional
 	UnauthenticatedRegistries []string `json:"unauthenticatedRegistries,omitempty"`
+	// ImagePullCAConfigMapRef is a reference to a config map containing a certificate authority certificate
```
carbonin, line 112: Can you just add a blank line between all the fields and the next field comment?
I see it's inconsistent in this struct, but we can fix that now.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -109,6 +109,12 @@ type AgentServiceConfigSpec struct {
...
+	// this is an optional certificate to allow a customer to add a certificate authority for a HTTPS source of images
```
carbonin, line 113: s/customer/user/

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1261,6 +1261,48 @@ func newImageServiceStatefulSet(ctx context.Context, log logrus.FieldLogger, asc
...
+			_, ok := additionalCABundleCM.Data["extra-ca.pem"]
+			if !ok {
+				err = multierror.Append(err, fmt.Errorf("expected to find single file `extra-ca.pem` in the ConfigMap %s but it was not found", asc.spec.ImagePullCAConfigMapRef.Name))
+			}
+			if len(additionalCABundleCM.Data) != 1 {
+				err = multierror.Append(err, fmt.Errorf("found multiple files in ConfigMap %s, a single file was expected but %d files were found", asc.spec.ImagePullCAConfigMapRef.Name, len(additionalCABundleCM.Data)))
+			}
+			if err != nil {
+				return err
+			}
```
carbonin, line 1282: I feel like all this validation doesn't belong in the StatefulSet mutate function.

I'd say this should be in the `validate` function and we should set a condition if any of the references are invalid.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1261,6 +1261,48 @@ func newImageServiceStatefulSet(ctx context.Context, log logrus.FieldLogger, asc
...
+				Name: "additional-ca-bundle",
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: volumeSource,
+				},
+			})
+		} else {
+			volumes = ensureVolume(volumes, corev1.Volume{
+				Name: "additional-ca-bundle",
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
```
carbonin, line 1300: We're not putting anything here in this case so the volume and volume mount should be added only if we have certs to add.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1261,6 +1261,48 @@ func newImageServiceStatefulSet(ctx context.Context, log logrus.FieldLogger, asc
...
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: asc.spec.ImagePullCAConfigMapRef.Name,
+				},
+			}
+			volumes = ensureVolume(volumes, corev1.Volume{
+				Name: "additional-ca-bundle",
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: volumeSource,
+				},
+			})
```
carbonin, line 1295: I don't think you need the `volumeSource` intermediate variable here I'd just put it all in one struct.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1177,6 +1178,7 @@ func newImageServiceStatefulSet(ctx context.Context, log logrus.FieldLogger, asc
 			{Name: "tls-certs", MountPath: "/etc/image-service/certs"},
 			{Name: "service-cabundle", MountPath: "/etc/image-service/ca-bundle"},
 			{Name: "image-service-data", MountPath: "/data"},
+			{Name: "additional-ca-bundle", MountPath: "/additional-ca-bundle"},
```
carbonin, line 1181: Maybe put this in `/etc/image-service` like the others. How about `/etc/image-service/os-images-ca-bundle`?

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -109,6 +109,12 @@ type AgentServiceConfigSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="List of container registries without authentication"
 	// +optional
 	UnauthenticatedRegistries []string `json:"unauthenticatedRegistries,omitempty"`
+	// OSImageCACertConfigMapRef is a reference to a config map containing a certificate authority certificate
+	// this is an optional certificate to allow a user to add a certificate authority for a HTTPS source of images
+	// this certificate will be used by the assisted-image-service when pulling OS images.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OS Image CA Cert ConfigMap reference"
+	// +optional
+	OSImageCACertConfigMapRef *corev1.LocalObjectReference `json:"osImageCACertConfigMap,omitempty"`
```
carbonin, line 117: I don't think we need to say `ConfigMap` here.
We don't in `MirrorRegistryRef` and that's also a reference to a `ConfigMap`

carbonin: Also the json serialization name should match the field name whenever possible just to reduce confusion.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1261,6 +1260,23 @@ func newImageServiceStatefulSet(ctx context.Context, log logrus.FieldLogger, asc
 			},
 		})
 
+		if asc.spec.OSImageCACertConfigMapRef != nil {
+			container.Env = append(container.Env, corev1.EnvVar{Name: "OS_IMAGE_DOWNLOAD_TRUSTED_CA_FILE", Value: "/etc/image-service/os-images-ca-bundle/extra-ca.pem"})
```
carbonin, line 1264: Can you make this file name a constant since it's used here and in the validation?

internal/controller/controllers/agentserviceconfig_controller_storage_validation_test.go
```diff
@@ -102,6 +102,7 @@ var _ = Describe("Agent service config controller storage validation", func() {
 	})
 
 	When("Objects haven't been created", func() {
+
```
carbonin, line 105: This doesn't need to have changed.

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -2357,6 +2373,44 @@ func getImageService(ctx context.Context, log logrus.FieldLogger, asc ASC) strin
 }
 
 func validate(ctx context.Context, log logrus.FieldLogger, asc ASC) (bool, error) {
+
+	if asc.spec.OSImageCACertConfigMapRef != nil && asc.spec.OSImageCACertConfigMapRef.Name != "" {
+		var multiErr *multierror.Error
```
carbonin, line 2378: I don't think we need multierror here.

If we fail to get the CM we won't check for the next two conditions.
And for the other two cases (wrong key/multiple keys) we can create one error message saying something like:
```
ConfigMap referenced by OSImgaeCACertRef is expected to contain a single key named "extra-ca.pem"
```

That should simplify this a bit I think

internal/controller/controllers/agentserviceconfig_controller_test.go
```diff
@@ -85,6 +87,212 @@ func AssertReconcileFailure(ctx context.Context, log logrus.FieldLogger, ascc AS
...
+			})
+		err = client.Update(ctx, osImageCACertConfigMap)
+		Expect(err).ToNot(HaveOccurred())
+		awaitOSImageCACertValidationSuccess()
+
+		// Grab the statefulset and verify that a volume has been mapped
+		assertOSImageCACertVolumePresent()
+	})
+
+	It("Should create statful set with no OSImageCACert volume if OSImageCACert ConfigMap has not been supplied", func() {
```
carbonin, line 286: Do we have a test for the volume being created successfully when it is provided?

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: Also not sure why it needs to be called `extra` since it's the only cert the user is really aware of.
Maybe just `tls.crt` as that's the name used for our other serving certs.

internal/controller/controllers/agentserviceconfig_controller_test.go
carbonin: Everything looks good except for this missing test case.

paul-maidment: I did actually add this, it's part of `assertOSImageCACertVolumePresent()`
I verify that there is a both a VolumeMount and a Volume mapped to that mount.

Is this what you meant?

carbonin: Ah sorry I missed the test case above this one. (`"Should add volume and volumemap for OSImageCACert when OSImageCACertRef is provided"`)

carbonin (approved):

carbonin:
/hold

Just thought of this. Are we doing anything to ensure we update the image service when the contents of this config map change?
Are we doing something like that currently for the other reference?

paul-maidment:
> /hold
> 
> Just thought of this. Are we doing anything to ensure we update the image service when the contents of this config map change? Are we doing something like that currently for the other reference?

When you make a change to the content of the ConfigMap these are kept in sync automatically, I actually believe this is core functionality in Kubernetes.

There doesn't appear to be a redeployment but the volume is simply kept in sync, if you shell into the image service then you can see the files being updated within 30 seconds of any change to the ConfigMap

As we can see in this excerpt of the statefulset, there is a reference to the ConfigMap in the volume, I think Kubernetes must automatically check these periodically.

```
...
      volumes:
      - name: tls-certs
        secret:
          defaultMode: 420
          secretName: assisted-image-service
      - configMap:
          defaultMode: 420
          name: assisted-image-service
        name: service-cabundle
      - configMap:
          defaultMode: 420
          name: image-service-additional-ca
        name: additional-ca-bundle
...
```

I would assume this is similar for EnvFrom.
So in conclusion, nothing to worry about...

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1712,6 +1743,7 @@ func newAssistedServiceDeployment(ctx context.Context, log logrus.FieldLogger, a
 		setAnnotation(meta, assistedConfigHashAnnotation, assistedConfigHash)
 		setAnnotation(meta, mirrorConfigHashAnnotation, mirrorConfigHash)
 		setAnnotation(meta, userConfigHashAnnotation, userConfigHash)
+		setAnnotation(meta, osImagesCAConfigHashAnnotation, osImagesCAConfigHash)
```
carbonin, line 1746: This should be being set on the image service, not assisted.

carbonin (changes requested):

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -168,6 +171,7 @@ func initASC(r *AgentServiceConfigReconciler, instance *aiv1beta1.AgentServiceCo
 	return asc
 }
 
+type NewStatefulSetFn func(context.Context, logrus.FieldLogger, ASC) (*appsv1.StatefulSet, controllerutil.MutateFn)
```
carbonin, line 174: I don't think this should be defined here if it's only used in tests.

carbonin: You are also only using `AssertReconcileSuccessStatefulSetFn` in one place. Maybe just don't define that function and then you wouldn't need this type.

carbonin (approved):

PR #5893, opened 2024-01-18
MGMT-16603: Allow setting agent cluster reference via BHM

ori-amizur:
In order to allow declarative association of host with a cluster in late binding flow, annotation support will be added to bmc. The annotation is "bmac.agent-install.openshift.io/cluster-reference". The annotation value is a json encoded string containing dictionary with keys [name,namespace] that correspond to the name and namespace of the cluster deployment. Reconcile of the BMH will cause the value of this annotation to be reflected to the cluster deployment in the corresponding agent spec.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @carbonin

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -88,6 +88,7 @@ const (
 	BMH_AGENT_IGNITION_CONFIG_OVERRIDES = "bmac.agent-install.openshift.io/ignition-config-overrides"
 	BMH_FINALIZER_NAME                  = "bmac.agent-install.openshift.io/deprovision"
 	BMH_DELETE_ANNOTATION               = "bmac.agent-install.openshift.io/remove-agent-and-node-on-delete"
+	BMH_CLUSTER_REFERENCE               = "bmac.agent-install.openshift.io/cluster-reference"
 	MACHINE_ROLE                        = "machine.openshift.io/cluster-api-machine-role"
 	MACHINE_TYPE                        = "machine.openshift.io/cluster-api-machine-type"
 	MCS_CERT_NAME                       = "ca.crt"
@@ -253,6 +254,10 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
 			"agent_namespace": agent.Namespace,
 		})
 	}
+	infraEnv, err := r.findInfraEnvForBMH(ctx, log, bmh)
+	if err != nil {
+		return reconcileError{err: err}.Result()
+	}
 
 	result := r.handleBMHFinalizer(ctx, log, bmh, agent)
 	if res := r.handleReconcileResult(ctx, log, result, bmh); res != nil {
@@ -266,7 +271,7 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
 		}
 	}
 
-	result = r.reconcileBMH(ctx, log, bmh, agent)
+	result = r.reconcileBMH(ctx, log, bmh, agent, infraEnv)
 	if res := r.handleReconcileResult(ctx, log, result, bmh); res != nil {
 		return res.Result()
 	}
@@ -282,7 +287,7 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
 	// with the BMH being reconciled. We will call both, reconcileAgentSpec and
 	// reconcileAgentInventory, every time. The logic to decide whether there's
 	// any action to take is implemented in each function respectively.
-	result = r.reconcileAgentSpec(log, bmh, agent)
+	result = r.reconcileAgentSpec(log, bmh, agent, infraEnv)
 	if res := r.handleReconcileResult(ctx, log, result, agent); res != nil {
 		return res.Result()
 	}
@@ -403,7 +408,7 @@ func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.Fiel
 // Unless there are errors, the agent should be `Approved` at the end of this
 // reconcile and a label should be set on it referencing the BMH. No changes to
 // the BMH should happen in this reconcile step.
-func (r *BMACReconciler) reconcileAgentSpec(log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent) reconcileResult {
+func (r *BMACReconciler) reconcileAgentSpec(log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent, infraEnv *aiv1beta1.InfraEnv) reconcileResult {
 
 	log.Debugf("Setting agent spec according to BMH")
 
@@ -472,6 +477,15 @@ func (r *BMACReconciler) reconcileAgentSpec(log logrus.FieldLogger, bmh *bmh_v1a
 		dirty = true
 	}
 
+	setDirty, err := r.reconcileClusterReference(bmh, agent, infraEnv)
+	if err != nil {
+		log.WithError(err).Errorf("failed to reconcile bmh cluster reference %s/%s", bmh.Namespace, bmh.Name)
+		return reconcileError{err: err}
+	}
+	if setDirty {
+		dirty = true
+	}
+
 	if r.reconcileNodeLabels(bmh, agent) {
 		dirty = true
 	}
@@ -497,6 +511,38 @@ func (r *BMACReconciler) reconcileNodeLabels(bmh *bmh_v1alpha1.BareMetalHost, ag
 		return true
 	}
 	return false
+}
+
+func (r *BMACReconciler) reconcileClusterReference(bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent, infraEnv *aiv1beta1.InfraEnv) (bool, error) {
+	clusterReferenceStr, annotationExists := bmh.Annotations[BMH_CLUSTER_REFERENCE]
+
+	switch {
+	case clusterReferenceStr != "":
+		if infraEnv != nil && infraEnv.Spec.ClusterRef != nil {
+			return false, errors.Errorf("failed to reconcile cluster reference.  InfraEnv already has cluster reference %s/%s",
+				infraEnv.Spec.ClusterRef.Namespace, infraEnv.Spec.ClusterRef.Name)
+		}
+		splits := strings.Split(clusterReferenceStr, "/")
+		if len(splits) != 2 {
+			return false, errors.Errorf("failed to reconcile cluster reference.  Parse of cluster reference %s failed",
+				clusterReferenceStr)
+		}
+		clusterReference := aiv1beta1.ClusterReference{
+			Namespace: splits[0],
+			Name:      splits[1],
+		}
+
+		if !reflect.DeepEqual(agent.Spec.ClusterDeploymentName, &clusterReference) {
+			agent.Spec.ClusterDeploymentName = &clusterReference
+			return true, nil
+		}
+
+	case annotationExists && clusterReferenceStr == "" && agent.Spec.ClusterDeploymentName != nil &&
+		(infraEnv == nil || infraEnv.Spec.ClusterRef == nil):
+		agent.Spec.ClusterDeploymentName = nil
+		return true, nil
+	}
+	return false, nil
 }
 
 // The detached annotation is added if the BMH provisioning state is provisioned
@@ -824,14 +870,10 @@ func (r *BMACReconciler) findInfraEnvForBMH(ctx context.Context, log logrus.Fiel
 // The above changes will be done only if the ISODownloadURL value has already
 // been set in the `InfraEnv` resource and the Image.URL value has not been
 // set in the `BareMetalHost`
-func (r *BMACReconciler) reconcileBMH(ctx context.Context, log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent) reconcileResult {
+func (r *BMACReconciler) reconcileBMH(ctx context.Context, log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent, infraEnv *aiv1beta1.InfraEnv) reconcileResult {
 	log.Debugf("Started BMH reconcile")
 	log.Debugf("BMH value %v", bmh)
 
-	infraEnv, err := r.findInfraEnvForBMH(ctx, log, bmh)
-	if err != nil {
-		return reconcileError{err: err}
-	}
 	// Stop `Reconcile` if BMH does not have an InfraEnv.
 	if infraEnv == nil {
 		return reconcileComplete{stop: true}
--- a/internal/controller/controllers/bmh_agent_controller_test.go
+++ b/internal/controller/controllers/bmh_agent_controller_test.go
@@ -312,6 +312,7 @@ var _ = Describe("bmac reconcile", func() {
 					CreatedTime:    &isoTimestamp,
 				}
 				Expect(c.Create(ctx, infraEnv)).To(BeNil())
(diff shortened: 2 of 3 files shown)


internal/controller/controllers/bmh_agent_controller_test.go
```diff
@@ -992,6 +1097,7 @@ var _ = Describe("bmac reconcile", func() {
 			host = newBMH(bmhName, &bmh_v1alpha1.BareMetalHostSpec{Image: image, BootMACAddress: macStr, BMC: bmh_v1alpha1.BMCDetails{CredentialsName: fmt.Sprintf(adminKubeConfigStringTemplate, clusterName)}})
 			annotations := make(map[string]string)
 			annotations[BMH_AGENT_IGNITION_CONFIG_OVERRIDES] = `{"ignition":{"version":"3.1.0", "security": {"tls":{"certificateAuthorities":[{"source":"data:text/plain;charset=utf-8;base64,c29tZSBjZXJ0aWZpY2F0ZQ=="}]}}}}`
+			annotations[BMH_CLUSTER_REFERENCE] = fmt.Sprintf(`{"name":"%s","namespace":"%s"}`, clusterName, testNamespace)
```
carbonin, line 1100: Why set this in these tests? Wouldn't we want this to continue testing the behavior as it was before?

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -499,6 +509,38 @@ func (r *BMACReconciler) reconcileNodeLabels(bmh *bmh_v1alpha1.BareMetalHost, ag
...
+		var clusterReference aiv1beta1.ClusterReference
+		if err = json.Unmarshal([]byte(clusterReferenceStr), &clusterReference); err != nil {
+			return false, errors.Wrapf(err, "failed to reconcile cluster reference.  Unmarshal of cluster reference %s failed",
+				clusterReferenceStr)
+		}
+		if !reflect.DeepEqual(agent.Spec.ClusterDeploymentName, &clusterReference) {
+			agent.Spec.ClusterDeploymentName = &clusterReference
+			return true, nil
+		}
+	} else if agent.Spec.ClusterDeploymentName != nil {
```
carbonin, line 531: Doesn't this effectively prevent users from setting the cluster reference directly on the agent if their agents were discovered using a BMH? This is a pretty large change in behavior given that we already support late-binding with BMH-discovered hosts.

I think we want to keep the existing behavior as well as add this new feature. One option would be to have the annotation present, but with no value, indicate that the agent should not be bound to a cluster. No annotation would then mean that the value on the agent should be respected.

@avishayt what do you think?

carbonin (commented):
/hold

Holding until we sort out the backwards compatibility issue.

internal/controller/controllers/bmh_agent_controller.go
ori-amizur: Changed the removal to depend on annotation on agent which is created when the cluster-reference is initially set.  It means that if the cluster-reference was not set initially by the bmh, it will not be removed if the cluster reference is missing from bmh.

internal/controller/controllers/bmh_agent_controller_test.go
ori-amizur: Removed

subsystem/kubeapi_test.go
```diff
@@ -5244,6 +5244,33 @@ var _ = Describe("bmac reconcile flow", func() {
...
+			}
+
+			Expect(kubeClient.Update(ctx, bmh)).ToNot(HaveOccurred())
+
+			Eventually(func() bool {
+				// expect bmh cluster reference annotation to be set
+				bmh = getBmhCRD(ctx, kubeClient, bmhNsName)
+				if bmh.ObjectMeta.Annotations == nil || bmh.ObjectMeta.Annotations[controllers.BMH_CLUSTER_REFERENCE] == "" {
+					return false
+				}
```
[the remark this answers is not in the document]
ori-amizur: OK

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -499,6 +511,44 @@ func (r *BMACReconciler) reconcileNodeLabels(bmh *bmh_v1alpha1.BareMetalHost, ag
...
+		if err = json.Unmarshal([]byte(clusterReferenceStr), &clusterReference); err != nil {
+			return false, errors.Wrapf(err, "failed to reconcile cluster reference.  Unmarshal of cluster reference %s failed",
+				clusterReferenceStr)
+		}
+		if !reflect.DeepEqual(agent.Spec.ClusterDeploymentName, &clusterReference) {
+			agent.Spec.ClusterDeploymentName = &clusterReference
+			if agent.ObjectMeta.Annotations == nil {
+				agent.ObjectMeta.Annotations = make(map[string]string)
+			}
+			agent.ObjectMeta.Annotations[AGENT_CLUSTER_REFERENCE_MANAGED_BY_BMH] = "true"
```
carbonin, line 534: We have a helper for this [here](https://github.com/openshift/assisted-service/blob/2c149a3ddc43362414021dfc38598d3b9a8a467f/internal/controller/controllers/common.go#L385-L390)

carbonin: Also do we ever remove this annotation? I feel like managing two is harder than the suggestion I made in my previous comment.

ori-amizur: I don't see a reason to remove the annotation. It is an indication that the cluster-reference is managed by BMH, unless the user wants to use also the API against the agent - which doesn't make sense once it was already set by the BMH.

Assuming when using BMH as the CR to update, the cluster-reference is removed in the reverse of the method it was added.  Leaving the annotation empty seems like a strange way that is passed to the user.

ori-amizur: > We have a helper for this [here](https://github.com/openshift/assisted-service/blob/2c149a3ddc43362414021dfc38598d3b9a8a467f/internal/controller/controllers/common.go#L385-L390)

OK

carbonin: I just think you can accomplish the same thing with a single annotation rather than two. 

You only have three cases to deal with, right?

1. If a user wants the cluster ref managed by the BMH and want it bound they set the value of the annotation.
2. If they want it managed by the BMH and want it unbound they set the annotation to empty.
3. If they don't want the cluster ref managed by the BMH they remove the annotation.

This is also how the other annotations on the BMH work that currently control the agent spec, the only complication here should be that we're also ensuring there isn't already a cluster ref set on the infraenv.

Hostname, for example, doesn't have some annotation on the agent to indicate that it's managed by the BMH. And if you change the hostname annotation value to empty it unsets hostname, if you remove the annotation then hostname is no longer managed by BMH. I don't see why this annotation should work differently.

ori-amizur: Don't you think it is strange that a user added cluster-ref by adding an annotation that didn't exist before, and to remove it instead of returning it the way it was in the beginning, change it to a third form?  Seems strange to me.

Hostname is different.  There is reason to remove the hostname if it doesn't exist.  It doesn't make sense.  So it is one way update.   Foe cluster-ref it makes sense to remove the reference.

If you will look at node-labels you will be able to see that they are always synced.

Other than node-labels, there is no clearing operation on any other properties such as the cluster-reference.

carbonin: I think the way node labels works is wrong. If nothing is set on the BMH the user should be able to use the API on the agent the same as all the other annotation based options.

avishayt: For hostname, node labels, ignition config overrides, installation disk, etc., we override the value in the Agent as long as the value is set in the BMH.  If the user unsets the value in the BMH, we stop syncing it, but the value on the Agent remains.  There's no way to unset any of these properties on the Agent today except for unsetting it on both the BMH and the Agent.  This was OK until now because those properties are scoped to installation, so we didn't come across the case where a user set something on a BMH and then regretted it.  Even less so because we assumed people setting annotations on BMHs are using gitops and are more experienced.

So we have a gap here, but so far it hasn't bothered us.

The cluster ref property isn't scoped to installation - we assume that users will set and unset throughout the lifetime of the host.

Ideally we would unify the behavior for all annotations, but this changes the current behavior, so let's think about it past the scope of this PR.  I propose the behavior should be:
1. If we see MANAGED_BY_BMH= true, set or unset all of the Agent properties according to the BMH annotations.
2. If we see MANAGED_BY_BMH= false, ignore the BMH annotations.
3. If we see MANAGED_BY_BMH is not present on the Agent, and at least one of the annotations is set on the BMH, then set MANAGED_BY_BMH=true.

So if any of the annotations is set or unset on the BMH, all of the properties should be mirrored to the Agent (anything that was set or will be set on the Agent will be overwritten).  If the user wants to stop this behavior, they need to set MANAGED_BY_BMH=false.

As I said, for now, let's do it only for the cluster ref property.
@ori-amizur , this is close to what you implemented, but you're missing some of the behavior (e.g., #2 in the description above).

@mhrivnak Any thoughts?

carbonin: I'm okay with this approach as long as it's consistent and well communicated to the user.

If we only implement this for the cluster ref in this PR I'm worried no one will ever go and update the rest of the annotations to subscribe to the new method and we'd end up in this state forever which would be very unfortunate.

avishayt: My guess is that changing the behavior for all properties won't break anyone because users usually either make changes via BMH or Agent, and not both, but it's a risk.  WDYT?

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -499,6 +511,44 @@ func (r *BMACReconciler) reconcileNodeLabels(bmh *bmh_v1alpha1.BareMetalHost, ag
...
 
+func (r *BMACReconciler) reconcileClusterReference(ctx context.Context, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent, log logrus.FieldLogger) (bool, error) {
+	if clusterReferenceStr, ok := bmh.ObjectMeta.Annotations[BMH_CLUSTER_REFERENCE]; ok {
+		infraEnv, err := r.findInfraEnvForBMH(ctx, log, bmh)
+		if err != nil {
+			return false, errors.Wrapf(err, "failed to find infraenv for bmh %s/%s", bmh.Namespace, bmh.Name)
+		}
+		if infraEnv != nil && infraEnv.Spec.ClusterRef != nil {
+			return false, errors.Errorf("failed to reconcile cluster reference.  InfraEnv already has cluster reference %s/%s",
+				infraEnv.Spec.ClusterRef.Namespace, infraEnv.Spec.ClusterRef.Name)
```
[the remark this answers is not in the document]
ori-amizur: It is returned to here: https://github.com/openshift/assisted-service/blob/98d4771b597efe20ba4302d873579e2560b4ef3b/internal/controller/controllers/bmh_agent_controller.go#L481

And later here https://github.com/openshift/assisted-service/blob/98d4771b597efe20ba4302d873579e2560b4ef3b/internal/controller/controllers/bmh_agent_controller.go#L288

Here it is analyzed before returning https://github.com/openshift/assisted-service/blob/98d4771b597efe20ba4302d873579e2560b4ef3b/internal/controller/controllers/bmh_agent_controller.go#L208

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -499,6 +511,44 @@ func (r *BMACReconciler) reconcileNodeLabels(bmh *bmh_v1alpha1.BareMetalHost, ag
 	return false
 }
 
+func (r *BMACReconciler) reconcileClusterReference(ctx context.Context, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent, log logrus.FieldLogger) (bool, error) {
+	if clusterReferenceStr, ok := bmh.ObjectMeta.Annotations[BMH_CLUSTER_REFERENCE]; ok {
+		infraEnv, err := r.findInfraEnvForBMH(ctx, log, bmh)
+		if err != nil {
+			return false, errors.Wrapf(err, "failed to find infraenv for bmh %s/%s", bmh.Namespace, bmh.Name)
+		}
```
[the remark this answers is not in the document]
ori-amizur: OK

mhrivnak:
It's common in k8s to see a namespaced name expressed as `<namespace>/<name>`. Can we just use that as the value of the annotation instead of json? It's much more usable, searchable, sortable, typeable, readable, ...

internal/controller/controllers/bmh_agent_controller.go
mhrivnak: If I'm following the question correctly, I think the agent's cluster reference should be set to the value of the BMH annotation when that annotation exists, and not changed when the annotation doesn't exist.

Want to declaratively use the BMH to ensure that the agent is not bound to a cluster? You should make sure the annotation has an empty value; that empty value will be propagated faithfully onto the agent's spec.

We should not be tracking who set what in the past or using that as context for behavioral decisions.

internal/controller/controllers/bmh_agent_controller.go
mhrivnak: We'll have a better API and UX if the API user doesn't have to manage the behavior of controllers in addition to declaring desired state, and this `MANAGED_BY_BMH` value is really just changing the behavior of a controller. 

Do we break anyone if we just do this?
* if one of the annotations exists on the BMH, copy its value (even the empty value) to the agent.
* for annotations that don't exist on the BMH, don't do anything to corresponding values on the agent.

@ori-amizur I understand your concern that if I 1) add an annotation that has some effect, then 2) remove that annotation, it should un-do the effect. Unfortunately that's not how declarative APIs work. Whatever desired state is specified *right now*, the job of controllers is to ensure that the real state of the world matches that desired state. What happened or was desired in the past is irrelevant.

This discussion highlights the complexity of managing default values. If an API user unsets a field, should it be automatically set to a well-documented default? Or should the real-world state associated with that field be left alone? It depends on the specific field, but either way, fields that can be unset should have clear documentation about what will happen.

Then what should happen if someone deletes an annotation? Not only did they unset the field; they deleted the field itself! Obviously we can't set the field to a sensible default if the field no longer exists.

In this specific case, it's even more complex, since we're using annotations on API X as a pass-through to set desired state on API Y. If any defaulting is going to happen, it really should happen on API X, but we can't set defaults for annotations. I think it would be a mistake to add a setting on API Y that changes its behavior in an attempt to compensate for the uncertainties that come with using annotations on API X.

TL;DR let's please try to find a way to do this that doesn't care what happened in the past nor add an annotation that really just changes what assumptions a controller makes when it deals with unset values.

ori-amizur:
> It's common in k8s to see a namespaced name expressed as `<namespace>/<name>`. Can we just use that as the value of the annotation instead of json? It's much more usable, searchable, sortable, typeable, readable, ...

OK

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -499,6 +513,35 @@ func (r *BMACReconciler) reconcileNodeLabels(bmh *bmh_v1alpha1.BareMetalHost, ag
 	return false
 }
 
+func (r *BMACReconciler) reconcileClusterReference(bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent, infraEnv *aiv1beta1.InfraEnv) (bool, error) {
+	if clusterReferenceStr, ok := bmh.Annotations[BMH_CLUSTER_REFERENCE]; ok {
```
avishayt, line 517: It would have less indentation and easier readability if you reversed the if
```suggestion
	if clusterReferenceStr, ok := bmh.Annotations[BMH_CLUSTER_REFERENCE]; !ok {
		return false, nil
	}
```

ori-amizur: OK

avishayt (approved):

PR #5898, opened 2024-01-21
MGMT-16332: fix validation for LVM multinode

eifrach:
Fixing requirements for LVM multinode deployment

1. full cluster should require disk on worker nodes only
2. full cluster shouldn't allow auto assign 
3. fixing requirment calculation 
4. adding tests for requirements  resources 

- [ ] New Feature 
- [X] Enhancement 
- [ ] Bug fix
- [X] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/operators/cnv/cnv_operator_test.go
+++ b/internal/operators/cnv/cnv_operator_test.go
@@ -51,7 +51,7 @@ var _ = Describe("CNV operator", func() {
 		It("request for lso, ocp version older than 4.11 will not get lvmo", func() {
 			haMode := models.ClusterHighAvailabilityModeNone
 			cluster := common.Cluster{
-				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: "4.11.0-0.0"},
+				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: "4.10.0-0.0"},
 			}
 
 			requirements, err := operator.GetDependencies(&cluster)
--- a/internal/operators/cnv/cnv_operator_test.go
+++ b/internal/operators/cnv/cnv_operator_test.go
@@ -39,7 +39,7 @@ var _ = Describe("CNV operator", func() {
 		It("request for lvmo", func() {
 			haMode := models.ClusterHighAvailabilityModeNone
 			cluster := common.Cluster{
-				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: lvm.LvmsMinOpenshiftVersion},
+				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: lvm.LvmsMinOpenshiftVersion4_12},
 			}
 
 			requirements, err := operator.GetDependencies(&cluster)
@@ -51,7 +51,7 @@ var _ = Describe("CNV operator", func() {
 		It("request for lso, ocp version older than 4.11 will not get lvmo", func() {
 			haMode := models.ClusterHighAvailabilityModeNone
 			cluster := common.Cluster{
-				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: "4.10.0-0.0"},
+				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: "4.11.0-0.0"},
 			}
 
 			requirements, err := operator.GetDependencies(&cluster)
@@ -68,7 +68,7 @@ var _ = Describe("CNV operator", func() {
 		BeforeEach(func() {
 			mode := models.ClusterHighAvailabilityModeFull
 			cluster = common.Cluster{
-				Cluster: models.Cluster{HighAvailabilityMode: &mode, OpenshiftVersion: lvm.LvmsMinOpenshiftVersion},
+				Cluster: models.Cluster{HighAvailabilityMode: &mode, OpenshiftVersion: lvm.LvmsMinOpenshiftVersion4_12},
 			}
 		})
 
@@ -234,7 +234,7 @@ var _ = Describe("CNV operator", func() {
 			host := models.Host{Role: models.HostRoleMaster}
 			haMode := models.ClusterHighAvailabilityModeNone
 			cluster = common.Cluster{
-				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: lvm.LvmsMinOpenshiftVersion},
+				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: lvm.LvmsMinOpenshiftVersion4_12},
 			}
 
 			requirements, err := operator.GetHostRequirements(context.TODO(), &cluster, &host)
--- a/internal/operators/lvm/config.go
+++ b/internal/operators/lvm/config.go
@@ -7,11 +7,11 @@ import (
 const (
 	// LvmMinOpenshiftVersion is the minimum OCP version in which lvmo is supported
 	// Any changes here should be updated at line 16 too.
-	LvmsMinOpenshiftVersion                            string = "4.11.0"
+	LvmoMinOpenshiftVersion                            string = "4.11.0"
 	LvmsMinOpenshiftVersion4_12                        string = "4.12.0"
 	LvmsMinOpenshiftVersion_ForNewResourceRequirements string = "4.13.0"
 
-	LvmoSubscriptionName string = "lvmo-operator"
+	LvmoSubscriptionName string = "odf-lvm-operator"
 	LvmsSubscriptionName string = "lvms-operator"
 
 	// LvmsMemoryRequirement           int64 = 400
--- a/internal/operators/lvm/config.go
+++ b/internal/operators/lvm/config.go
@@ -11,7 +11,7 @@ const (
 	LvmsMinOpenshiftVersion4_12                        string = "4.12.0"
 	LvmsMinOpenshiftVersion_ForNewResourceRequirements string = "4.13.0"
 
-	LvmoSubscriptionName string = "odf-lvm-operator"
+	LvmoSubscriptionName string = "lvmo-operator"
 	LvmsSubscriptionName string = "lvms-operator"
 
 	// LvmsMemoryRequirement           int64 = 400
--- a/internal/operators/lvm/config.go
+++ b/internal/operators/lvm/config.go
@@ -7,20 +7,22 @@ import (
 const (
 	// LvmMinOpenshiftVersion is the minimum OCP version in which lvmo is supported
 	// Any changes here should be updated at line 16 too.
-	LvmoMinOpenshiftVersion                           string = "4.11.0"
-	LvmsMinOpenshiftVersion                           string = "4.12.0"
-	LvmsMinOpenshiftVersionForNewResourceRequirements string = "4.13.0"
+	LvmsMinOpenshiftVersion                            string = "4.11.0"
+	LvmsMinOpenshiftVersion4_12                        string = "4.12.0"
+	LvmsMinOpenshiftVersion_ForNewResourceRequirements string = "4.13.0"
 
 	LvmoSubscriptionName string = "odf-lvm-operator"
 	LvmsSubscriptionName string = "lvms-operator"
 
-	LvmsMemoryRequirement           int64 = 400
-	LvmsMemoryRequirementBefore4_13 int64 = 1200
+	// LvmsMemoryRequirement           int64 = 400
+	// LvmsMemoryRequirementBefore4_13 int64 = 1200
+
 )
 
 type Config struct {
 	LvmCPUPerHost                 int64  `envconfig:"LVM_CPU_PER_HOST" default:"1"`
 	LvmMemoryPerHostMiB           int64  `envconfig:"LVM_MEMORY_PER_HOST_MIB" default:"400"`
+	LvmMemoryPerHostMiBBefore4_13 int64  `envconfig:"LVM_MEMORY_PER_HOST_MIB" default:"1200"`
 	LvmMinOpenshiftVersion        string `envconfig:"LVM_MIN_OPENSHIFT_VERSION" default:"4.11.0"`
 	LvmMinMultiNodeSupportVersion string `envconfig:"LVM_MIN_MULTI_NODE_SUPPORT_VERSION" default:"4.15.0"`
 }
--- a/internal/operators/lvm/lvm_operator.go
+++ b/internal/operators/lvm/lvm_operator.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/hashicorp/go-version"
+	"github.com/go-openapi/swag"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/oc"
@@ -17,7 +17,7 @@ import (
 // operator is an ODF LVM OLM operator plugin; it implements api.Operator
 type operator struct {
 	log       logrus.FieldLogger
-	config    *Config
+	Config    *Config
 	extracter oc.Extracter
 }
 
@@ -43,7 +43,7 @@ func NewLvmOperator(log logrus.FieldLogger, extracter oc.Extracter) *operator {
 func newLvmOperatorWithConfig(log logrus.FieldLogger, config *Config, extracter oc.Extracter) *operator {
 	return &operator{
 		log:       log,
-		config:    config,
+		Config:    config,
 		extracter: extracter,
 	}
 }
@@ -74,34 +74,17 @@ func (o *operator) GetHostValidationID() string {
 
 // ValidateCluster always return "valid" result
 func (o *operator) ValidateCluster(_ context.Context, cluster *common.Cluster) (api.ValidationResult, error) {
-	ocpVersion, err := version.NewVersion(cluster.OpenshiftVersion)
-	if err != nil {
-		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{err.Error()}}, nil
+	if ok, _ := common.BaseVersionLessThan(o.Config.LvmMinOpenshiftVersion, cluster.OpenshiftVersion); ok {
+		message := fmt.Sprintf("Logical Volume Manager is only supported for openshift versions %s and above", o.Config.LvmMinOpenshiftVersion)
+		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetClusterValidationID(), Reasons: []string{message}}, nil
 	}
 
-	if common.IsSingleNodeCluster(cluster) {
-		minOpenshiftVersionForLvm, err := version.NewVersion(o.config.LvmMinOpenshiftVersion)
-		if err != nil {
-			return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{err.Error()}}, nil
+	if swag.StringValue(cluster.HighAvailabilityMode) == models.ClusterHighAvailabilityModeFull {
+		if ok, _ := common.BaseVersionLessThan(o.Config.LvmMinMultiNodeSupportVersion, cluster.OpenshiftVersion); ok {
+			message := fmt.Sprintf("Logical Volume Manager is only supported for highly available openshift with version %s or above", o.Config.LvmMinMultiNodeSupportVersion)
+			return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{message}}, nil
 		}
-		if ok, _ := common.BaseVersionLessThan(minOpenshiftVersionForLvm.String(), ocpVersion.String()); ok {
-			message := fmt.Sprintf("Logical Volume Manager is only supported for openshift versions %s and above", o.config.LvmMinOpenshiftVersion)
-			return api.ValidationResult{Status: api.Failure, ValidationId: o.GetClusterValidationID(), Reasons: []string{message}}, nil
-		}
-	} else {
-		// HA support was introduced after LVM support in general, so we need to check for a different version
-		minOpenshiftVersionForMultiNodeSupport, err := version.NewVersion(o.config.LvmMinMultiNodeSupportVersion)
-		if err != nil {
-			return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{err.Error()}}, nil
-		}
-		if ok, _ := common.BaseVersionLessThan(minOpenshiftVersionForMultiNodeSupport.String(), ocpVersion.String()); ok {
-			message := fmt.Sprintf("Logical Volume Manager is only supported for highly available openshift with version %s or above",
-				minOpenshiftVersionForMultiNodeSupport.String())
-			return api.ValidationResult{Status: api.Failure, ValidationId: o.GetClusterValidationID(), Reasons: []string{message}}, nil
-		}
-
 	}
-
 	return api.ValidationResult{Status: api.Success, ValidationId: o.GetClusterValidationID()}, nil
 }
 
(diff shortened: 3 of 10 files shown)


internal/operators/lvm/lvm_operator.go
```diff
@@ -163,40 +197,81 @@ func (o *operator) GetPreflightRequirements(context context.Context, cluster *co
...
-	log := logutil.FromContext(ctx, o.log)
-	greaterOrEqual, err := common.BaseVersionGreaterOrEqual(LvmsMinOpenshiftVersionForNewResourceRequirements, cluster.OpenshiftVersion)
-	if err != nil {
-		log.Warnf("Error parsing cluster.OpenshiftVersion: %s, setting lvms memory requirement to %d", err.Error(), o.config.LvmMemoryPerHostMiB)
-		return o.config.LvmMemoryPerHostMiB
+	baseHardwareRequirement, _ := getBaseRequiemnts(get_page)
+
+	var memoryOverhead int64
+
+	if ok, _ := common.BaseVersionLessThan("4.13.0", cluster.OpenshiftVersion); ok {
```
gamli75, line 237: please use consts for 4.13

internal/operators/lvm/lvm_operator.go
```diff
@@ -118,14 +122,31 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
...
+			return api.ValidationResult{Status: api.Success, ValidationId: o.GetHostValidationID()}, nil
+		}
 	}
-	if diskCount == 0 {
-		message := "Logical Volume Manager requires at least one non-installation HDD/SSD disk on the host"
-		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{message}}, nil
+
+	if mode == common.ClusterDeploymentModeStandard {
+		if host.Role == models.HostRoleAutoAssign {
+			status := "For LVM Standard Mode, host role must be assigned to master or worker."
```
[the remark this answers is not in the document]
nirfarkas: When do we show this message and why?

internal/operators/lvm/common.go
```diff
@@ -0,0 +1,62 @@
...
+
+func downloadFromUrl(pageDownloadFunc pageDownloader) (*http.Response, error) {
+	return pageDownloadFunc(urlRequirement)
+}
+
+func get_page(url string) (*http.Response, error) {
+	return http.Get(url)
+}
+
+func getBaseRequiemnts(pageDownloadFunc pageDownloader) (hardwareRequirement, error) {
```
avishayt, line 40: Why are we downloading the requirements from a URL?

internal/operators/lvm/lvm_operator.go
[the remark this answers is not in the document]
avishayt: Why can't it work with auto-assign?  It should use common.GetEffectiveRole(host)

internal/operators/lvm/lvm_operator.go
```diff
@@ -118,14 +122,31 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
...
+			return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{status}}, nil
+		}
+	}
+
+	// check worker hosts or  unbind hosts to cluster
+	if host.Role != models.HostRoleMaster {
```
avishayt, line 146: Do we know for a fact that this is a requirement?  LVMS won't manage if one worker doesn't have disks?

internal/common/common.go
```diff
@@ -166,6 +171,20 @@ func IsSingleNodeCluster(cluster *Cluster) bool {
 	return swag.StringValue(cluster.HighAvailabilityMode) == models.ClusterHighAvailabilityModeNone
 }
 
+func GetClusterDeploymentMode(cluster *Cluster) string {
```
avishayt, line 174: We already have these properties for these, not according to the number of hosts.
SNO would be cluster.highAvailabilityMode == "None"
You actually don't care about compact, you want to know if cluster.SchedulableMasters is true.

internal/operators/lvm/lvm_operator.go
eifrach: The storage must be consistent across clusters so the requirement should be at least one disk per valid node.
 Assisted is enforcing one disk on each valid node.
The operator can be deployed without disks on all valid nodes.

internal/operators/lvm/common.go
eifrach: This is for local deployment and/or testing.

The service is realying on env var only. we can remove it if you think we can ignore it

avishayt: It looks weird.  Why not just use env var?

internal/operators/lvm/lvm_operator.go
avishayt: OK, if that's in fact the requirement, we should check it

internal/common/common.go
eifrach: the Idea here is for  what type of cluster,
this `cluster.SchedulableMasters== true` applies to SNO and compact but I still need to check how many nodes or do  `cluster.SchedulableMasters== true && cluster.highAvailabilityMode == "None"` 

this simplify the check which used in LVM and ODF (  may be in other places )

internal/operators/lvm/lvm_operator.go
```diff
@@ -118,14 +124,33 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
...
+			if diskCount == 0 {
+				return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{message}}, nil
+			}
+			return api.ValidationResult{Status: api.Success, ValidationId: o.GetHostValidationID()}, nil
+		}
+	}
+
+	if mode == common.ClusterDeploymentModeStandard {
+		if role == models.HostRoleAutoAssign {
+			status := "For LVM Standard Mode, host role must be assigned to master or worker."
```
avishayt, line 144: You can make this code much more straightforward:

```
validationStatus := api.Success
if common.GetEffectiveRole(host) == models.HostRoleWorker || common.AreMastersSchedulable(cluster) {
    if diskCount == 0 {
        validationStatus = api.Failure
    }
}
```

No need for all the consts of cluster types, the validations about autoassign, etc.

internal/operators/lvm/common.go
```diff
@@ -0,0 +1,67 @@
...
+	if baseRequirement != "" {
+		err := json.Unmarshal([]byte(baseRequirement), &baseHardwareRequirement)
+		if err == nil {
+			return baseHardwareRequirement[0], nil
+		}
+	}
+	resp, err := downloadFromUrl(pageDownloadFunc)
+	if err != nil {
+		return baseHardwareRequirement[0], fmt.Errorf("Couldn't download requirement file")
+	}
```
adriengentil, line 58: I don't think we should download files from github at runtime or for the tests, download can fail depending on github status, and because of rate limiting policies. Also, it can break for unrelated reasons if someone moves/rename the file.

I think we should enforce `envRequirementsKey` to be set when we start the service.

internal/operators/lvm/common.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+const (
+	envRequirementsKey string = "HW_VALIDATOR_REQUIREMENTS"
+	urlRequirement     string = "https://raw.githubusercontent.com/openshift/assisted-service/master/data/default_hw_requirements.json"
+
+	openshiftVersion4_10 string = "4.10.0"
+	openshiftVersion4_11 string = "4.11.0"
+	openshiftVersion4_13 string = "4.13.0"
+	openshiftVersion4_15 string = "4.15.0"
```
adriengentil, line 30: Maybe we can we move them to some common package?

internal/operators/lvm/common.go
```diff
@@ -0,0 +1,67 @@
...
+
+func downloadFromUrl(pageDownloadFunc pageDownloader) (*http.Response, error) {
+	return pageDownloadFunc(urlRequirement)
+}
+
+func GetPage(url string) (*http.Response, error) {
+	return http.Get(url)
+}
+
+func GetBaseRequiemnts(pageDownloadFunc pageDownloader) (hardwareRequirement, error) {
```
adriengentil, line 45:
```suggestion
func GetBaseRequirements(pageDownloadFunc pageDownloader) (hardwareRequirement, error) {
```

internal/operators/lvm/common_test.go
```diff
@@ -0,0 +1,95 @@
...
+		"ram_mib": 2,
+		"disk_size_gb": 3,
+		"installation_disk_speed_threshold_ms": 10,
+		"network_latency_threshold_ms": 1000,
+		"packet_loss_percentage": 10
+	  }
+	}
+]`
+
+func mock_get_page(url string) (*http.Response, error) {
```
adriengentil, line 46:
```suggestion
func mockGetPage(url string) (*http.Response, error) {
```

let's be consistent for the naming of the functions

internal/operators/lvm/common_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"net/http"
+	"os"
+	"strings"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var hwRequirement = `
+[
```
adriengentil, line 16: why it starts with a list? Looks like there's always 1 element?

internal/operators/lvm/lvm_operator.go
```diff
@@ -147,56 +171,109 @@ func (o *operator) GetMonitoredOperator() *models.MonitoredOperator {
...
+		return &models.OperatorHardwareRequirements{}, fmt.Errorf("LVM is not supported")
+	}
+
 	dependecies, err := o.GetDependencies(cluster)
 	if err != nil {
 		return &models.OperatorHardwareRequirements{}, err
 	}
-	return &models.OperatorHardwareRequirements{
+
+	baseHardwareRequirement, _ := GetBaseRequirements(GetPage)
```
avishayt, line 215: Fetching from the web?  How will this work when the hosts don't have internet access?  Or GitHub is down?

internal/operators/lvm/lvm_operator.go
```diff
@@ -118,14 +120,36 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
 		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{message}}, err
 	}
 
-	// GetValidDiskCount counts the total number of valid disks in each host and return an error if we don't have the disk of required size
 	diskCount := o.getValidDiskCount(inventory.Disks, host.InstallationDiskID)
+
+	role := common.GetEffectiveRole(host)
+	message := "Logical Volume Manager requires at least one non-installation HDD/SSD disk on the host"
+	if (role == models.HostRoleWorker && !*cluster.SchedulableMasters) || *cluster.SchedulableMasters {
```
avishayt, line 127: If the host is a worker, you want to validate it regardless of schedulable masters
```suggestion
	if (role == models.HostRoleWorker || *cluster.SchedulableMasters {
```

internal/operators/lvm/lvm_operator.go
```diff
@@ -147,56 +171,109 @@ func (o *operator) GetMonitoredOperator() *models.MonitoredOperator {
...
 	preflightRequirements, err := o.GetPreflightRequirements(ctx, cluster)
 	if err != nil {
-		log.WithError(err).Errorf("Cannot retrieve preflight requirements for cluster %s", cluster.ID)
+		log.WithError(err).Errorf("Cannot retrieve preflight requirements for host %s", host.ID)
 		return nil, err
 	}
+
+	if len(cluster.Hosts) <= 3 {
+		return preflightRequirements.Requirements.Master.Quantitative, nil
+	}
```
avishayt, line 184: What is this trying to validate?  LVM operator no longer works for SNO?

internal/operators/lvm/lvm_operator.go
```diff
@@ -147,56 +171,109 @@ func (o *operator) GetMonitoredOperator() *models.MonitoredOperator {
...
+
+	role := common.GetEffectiveRole(host)
+
+	if role == models.HostRoleWorker && *cluster.HighAvailabilityMode == models.ClusterHighAvailabilityModeFull {
```
avishayt, line 188: What is this trying to validate?  What about masters with schedulable masters?  What about SNO+workers?

internal/operators/lvm/common.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	data, _ := io.ReadAll(resp.Body)
+	err = json.Unmarshal(data, &baseHardwareRequirement)
+	if err != nil {
+		return baseHardwareRequirement[0], fmt.Errorf("Couldn't marshal JSON  from URL")
+
+	}
+
+	return baseHardwareRequirement[0], nil
+}
```
avishayt, line 67: I still don't understand why we ever want to download requirements from GitHub.

avishayt (changes requested):

internal/operators/lvm/lvm_operator.go
```diff
@@ -147,56 +171,109 @@ func (o *operator) GetMonitoredOperator() *models.MonitoredOperator {
...
+	}
+
+	// default to Master
 	return preflightRequirements.Requirements.Master.Quantitative, nil
 }
 
 // GetPreflightRequirements returns operator hardware requirements that can be determined with cluster data only
 func (o *operator) GetPreflightRequirements(context context.Context, cluster *common.Cluster) (*models.OperatorHardwareRequirements, error) {
+	// unsupportted before 4.11
+	if ok, _ := common.BaseVersionLessThan(openshiftVersion4_11, cluster.OpenshiftVersion); ok {
```
rccrdpccl, line 206: In this case I am not sure having the constant `openshiftVersion4_11` helps: wouldn't it be more readable as `BaseVersionLessThan(minOpenshiftVersionLVMOperator, clusterOpenshiftVersion)`? Or maybe `BaseVersionLessThan("4.11", clusterOpenshiftVersion)`. Same for all other similar constants

internal/operators/lvm/common.go
```diff
@@ -0,0 +1,67 @@
...
+	"net/http"
+	"os"
+)
+
+type baseHardware struct {
+	CPUCores int64 `json:"cpu_cores"`
+	RAMMib   int64 `json:"ram_mib"`
+}
+
+type hardwareRequirement struct {
```
rccrdpccl, line 16: any reason why we shouldn't use [this](https://github.com/openshift/assisted-service/blob/master/internal/hardware/versioned-requirements.go)?

internal/operators/lvm/config.go
```diff
@@ -7,9 +7,9 @@ import (
 const (
 	// LvmMinOpenshiftVersion is the minimum OCP version in which lvmo is supported
 	// Any changes here should be updated at line 16 too.
-	LvmoMinOpenshiftVersion                           string = "4.11.0"
-	LvmsMinOpenshiftVersion                           string = "4.12.0"
-	LvmsMinOpenshiftVersionForNewResourceRequirements string = "4.13.0"
+	LvmsMinOpenshiftVersion                            string = "4.11.0"
```
rccrdpccl, line 10: Why do we have 3 different `minVersions`?

internal/operators/lvm/lvm_operator.go
```diff
@@ -147,56 +171,113 @@ func (o *operator) GetMonitoredOperator() *models.MonitoredOperator {
...
+	}
 
-func (o *operator) getLvmMemoryPerHostMib(ctx context.Context, cluster *common.Cluster) int64 {
-	log := logutil.FromContext(ctx, o.log)
-	greaterOrEqual, err := common.BaseVersionGreaterOrEqual(LvmsMinOpenshiftVersionForNewResourceRequirements, cluster.OpenshiftVersion)
-	if err != nil {
-		log.Warnf("Error parsing cluster.OpenshiftVersion: %s, setting lvms memory requirement to %d", err.Error(), o.config.LvmMemoryPerHostMiB)
-		return o.config.LvmMemoryPerHostMiB
+	// unsupportted multi node before 4.15
+	if ok, _ := common.BaseVersionLessThan(openshiftVersion4_15, cluster.OpenshiftVersion); ok {
```
rccrdpccl, line 249: maybe this constant can be `lvmOnMultinodeMaxVersion` or something along the lines?

internal/operators/cnv/cnv_operator_test.go
```diff
@@ -51,7 +51,7 @@ var _ = Describe("CNV operator", func() {
 		It("request for lso, ocp version older than 4.11 will not get lvmo", func() {
 			haMode := models.ClusterHighAvailabilityModeNone
 			cluster := common.Cluster{
-				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: "4.11.0-0.0"},
+				Cluster: models.Cluster{HighAvailabilityMode: &haMode, OpenshiftVersion: "4.10.0-0.0"},
```
adriengentil, line 54: it this change part of this PR?

internal/operators/lvm/lvm_operator.go
eifrach: dont like the term `maxVersion`  we can use `lvmOnMultinodeMinVersion` which is 4.15
but I think the version here is more readable than something like min/max

internal/operators/lvm/lvm_operator.go
avishayt: Right, so make it clearer with my code suggestion - either worker or schedulablemasters.

avishayt: Also can cluster.SchedulableMasters be nil?

internal/operators/lvm/lvm_operator.go
avishayt: The number of hosts doesn't mean anything.  10 hosts can be 3 masters and 7 workers, or SNO plus 9 workers.  Check schedulableMasters instead.  It should be the same code as above:

```
role := common.GetEffectiveRole(host)
if (role == models.HostRoleWorker || *cluster.SchedulableMasters {
```

internal/operators/lvm/lvm_operator.go
```diff
@@ -147,56 +171,123 @@ func (o *operator) GetMonitoredOperator() *models.MonitoredOperator {
...
+		return &models.OperatorHardwareRequirements{}, fmt.Errorf("LVM is not supported")
+	}
+
 	dependecies, err := o.GetDependencies(cluster)
 	if err != nil {
 		return &models.OperatorHardwareRequirements{}, err
 	}
-	return &models.OperatorHardwareRequirements{
+
+	// baseHardwareRequirement, _ := GetBaseRequirements(GetPage)
```
osherdp, line 215: I think you've left some commented-out code

internal/operators/lvm/lvm_operator.go
```diff
@@ -118,14 +120,36 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
...
+	if role == models.HostRoleAutoAssign && !*cluster.SchedulableMasters {
+		status := "For LVM Standard Mode, host role must be assigned to master or worker."
+		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{status}}, nil
+	}
+
+	requirements, err := o.GetHostRequirements(ctx, cluster, host)
 	if err != nil {
-		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{err.Error()}}, nil
+		message := fmt.Sprintf("Failed to get host requirements for host %s", host.ID)
+		o.log.Error(message)
```
osherdp, line 141: we're kind of losing `err` here in our logs, no?
in case it has any valuable information about the specifics of the error

internal/operators/lvm/lvm_operator.go
```diff
@@ -118,14 +120,36 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
...
+		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{message, err.Error()}}, err
+
+	}
+	if inventory.CPU.Count < requirements.CPUCores {
+		failedStatus := fmt.Sprintf("Insufficient CPU to deploy LVM Required CPU count is %d but found %d", requirements.CPUCores, inventory.CPU.Count)
```
osherdp, line 146:
```suggestion
		failedStatus := fmt.Sprintf("Insufficient CPU to deploy LVM. Required CPU count is %d but found %d", requirements.CPUCores, inventory.CPU.Count)
```

internal/operators/lvm/lvm_operator.go
```diff
@@ -147,56 +171,123 @@ func (o *operator) GetMonitoredOperator() *models.MonitoredOperator {
...
+
+	// default to Master
 	return preflightRequirements.Requirements.Master.Quantitative, nil
 }
 
 // GetPreflightRequirements returns operator hardware requirements that can be determined with cluster data only
 func (o *operator) GetPreflightRequirements(context context.Context, cluster *common.Cluster) (*models.OperatorHardwareRequirements, error) {
+	// unsupportted before 4.11
+	if ok, _ := common.BaseVersionLessThan(o.Config.LvmMinOpenshiftVersion, cluster.OpenshiftVersion); ok {
+		return &models.OperatorHardwareRequirements{}, fmt.Errorf("LVM is not supported")
```
osherdp, line 207:
```suggestion
		return &models.OperatorHardwareRequirements{}, fmt.Errorf("LVM is not supported for OpenShift %s", cluster.OpenshiftVersion)
```

internal/operators/lvm/lvm_operator.go
```diff
@@ -118,14 +120,36 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
...
+	if inventory.CPU.Count < requirements.CPUCores {
+		failedStatus := fmt.Sprintf("Insufficient CPU to deploy LVM Required CPU count is %d but found %d", requirements.CPUCores, inventory.CPU.Count)
+		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{failedStatus}}, nil
 	}
-	if diskCount == 0 {
-		message := "Logical Volume Manager requires at least one non-installation HDD/SSD disk on the host"
-		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{message}}, nil
+	memoryBytes := conversions.MibToBytes(requirements.RAMMib)
+	if inventory.Memory.UsableBytes < memoryBytes {
+		failedStatus := fmt.Sprintf("Insufficient Memory to deploy LVM Required Memory is %d but found %d", memoryBytes, inventory.Memory.UsableBytes)
```
osherdp, line 151:
```suggestion
		failedStatus := fmt.Sprintf("Insufficient Memory to deploy LVM. Required Memory is %d but found %d", memoryBytes, inventory.Memory.UsableBytes)
```

osherdp (commented):
Mainly syntactic comments

internal/operators/lvm/config.go
eifrach: so there is no difference between `lvms` and `lvmo` they are both called `lvm`.

the only place that has a differtents is in the OLM where they are two different bundles 
so they only place you will see it is in the [manifest](https://github.com/openshift/assisted-service/blob/master/internal/operators/lvm/manifest.go#L38-L55)

internal/operators/lvm/lvm_operator.go
eifrach: the code is changed, there is no longer an error here

internal/operators/lvm/lvm_operator.go
eifrach: that code was removed

internal/operators/lvm/lvm_operator.go
eifrach: that code was removed

internal/operators/lvm/lvm_operator.go
eifrach: that code was removed

internal/operators/lvm/lvm_operator.go
eifrach: was removed

internal/operators/lvm/lvm_operator.go
eifrach: changed to the suggested and moved to line 109

internal/operators/lvm/config.go
rccrdpccl: Not suggesting to remove, just to name accordingly and clearly.

Why do we have a constant if we have a config set from env var with default values?
What's the difference between constant `LvmMinOpenshiftVersion` and env var `LVM_MIN_OPENSHIFT_VERSION`?
I think it's important that we make all these "min versions" as less confusing as possible.

eifrach: there is only one min version

rccrdpccl:
```
        LvmsMinOpenshiftVersion                            string = "4.11.0"
	LvmsMinOpenshiftVersion4_12                        string = "4.12.0"
	LvmsMinOpenshiftVersion_ForNewResourceRequirements string = "4.13.0"
```

plus all the config values, which it's not clear what's the relation between constant "min version" and env var "min version"

eifrach: we talked offline, we need to release the version 
I will create a new PR for naming and better usage

PR #5916, opened 2024-01-24
MGMT-16262: Part1 of implementing usage of custom OCP releases

danmanor:
## In General
This PR is part 1 of the Implementation of this enhancement proposal - https://github.com/openshift/assisted-service/blob/master/docs/enhancements/dynamic-fetching-of-openshift-versions.md, which is enabling usage of dynamic `openshift` releases.

### Main flow
This part is about creating a periodic utility that updates current `OCP` releases in `postgresql` DB, for later use of assisted service -

The main changes are:

1. main fetches new configuration called `RELEASE_SOURCES` which specifies which `OCP` releases we want to keep in the DB and support (new table called `release_images`) as well as from which channel, what CPU architectures, etc. Then it starts a new go routine that runs periodically.
3. This go routine fetches releases from the API according to the configuration, and `OCP` versions support levels from another API.
4. It labels the default release (latest stable release), supporting levels.
5. Adds release images from the configuration (with precedence in case of conflicts).
6. Replaces this data with the old one in the DB.

### Plan forward (part 2)
This part is about retrieving the new list of supported openshift versions in endpoint `/api/assisted-install/v2/openshift-versions`, and enabling their installation.

The main changes are:

1. In openshift-versions endpoint - We retrieve the release images from the DB (from part (1), which contains release images from configuration and API), and in addition we filter by 2 new query parameters (`only_latest`, `version_pattern`) that will be used by the `UI` to present short list of the versions. We also filter by a new configuration parameter `IGNORED_RELEASE_IMAGES`, which contains a list of versions we don't want to retrieve (as one might have a blocker bug).
2. We change the release handler struct to use the new source of truth (DB) instead of the configuration, to enable their installation.

### Remarks

1.  As proposed in: https://github.com/openshift/assisted-service/blob/e604bc46caf1fac53eab35817c739add1c8911ee/docs/enhancements/dynamic-fetching-of-openshift-versions.md?plain=1#L145 I exposed new end point in - `/api/assisted-install/v2/release-sources`
2. This feature will will be active **only** in connected REST API deployments, using a new configuration called `ENABLE_OPENSHIFT_RELEASE_SYNCER`. This is because this new feature does not have much meaning in kubeAPI deployments as then openshift release images are managed through a resource called `ClusterIMageSet`, also specified in the enhancement - https://github.com/openshift/assisted-service/blob/ee0a3501ebe84f3ee59f5f9318b55bfd076fceed/docs/enhancements/dynamic-fetching-of-openshift-versions.md?plain=1#L71 In order to prevent this feature from crashing the service in case of isolated environments (such as in agent installations)
3. I deployed a test instance with extra endpoint available in - http://rdu-infra-edge-06.infra-edge.lab.eng.rdu2.redhat.com:6008/api/assisted-install/v2/release-images
4. I chose to add the configuration releases to the DB for simplicity sake.
5. I chose not to keep the `cpu_architectures` for each release image in the DB for simplicity sake, as it can be derived from `cpu_architecture`, and it requires using SQL arrays
6. The current release sources configuration is a suggestion, we can remove some architectures/channels/versions to reduce the releases list size.
7. As specified in the enhancement: https://github.com/openshift/assisted-service/blob/f5fc19587c15870b3dd03ad8060a6f1b70634c4f/docs/enhancements/dynamic-fetching-of-openshift-versions.md?plain=1#L237 I kept `RELEASE_IMAGES` in order to enable unsupported release images. It is merged into the API release images during fetching process in order to have one source of truth. 

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [x] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md,


cmd/main.go
```diff
@@ -892,3 +908,12 @@ func startPPROF(log *logrus.Logger) {
 		}
 	}
 }
+
+func runOpenshiftReleaseSyncerIfNeeded(kubeAPIMode, enableOpenshiftReleaseSyncer bool, ReleaseSourcesHandler releasesources.ReleaseSourcesHandler, log logrus.FieldLogger) {
+	if !kubeAPIMode && enableOpenshiftReleaseSyncer {
+		OpenShiftReleaseSyncer := thread.New(
+			log.WithField("pkg", "opensift-release-syncer"), "Openshift Releases Syncer", Options.OpenShiftReleaseSyncerInterval, ReleaseSourcesHandler.SyncReleaseImages)
+		OpenShiftReleaseSyncer.Start(true)
```
omertuc, line 916: Can you add a comment explaining why this needs to be `true`?

cmd/main.go
```diff
@@ -265,6 +270,12 @@ func main() {
 			"Failed to parse RELEASE_IMAGES json %s", Options.ReleaseImages)
 	}
 
+	var releaseSourcesArray = models.ReleaseSources{}
+	if Options.ReleaseSources != "" {
```
omertuc, line 274: If RELEASE_IMAGES and RELEASE_SOURCES are both given, is that valid?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,468 @@
...
+
+	if err = tx.Create(&supportLevels).Error; err != nil {
+		h.log.WithError(err).Debug("error occurred while trying to create suppurt levels table")
+		return nil, err
+	}
+
+	return supportLevels, nil
+}
+
+// All candidate/pre-release releases are considered support level 'beta'.
```
omertuc, line 309: function doc comments should follow standard go function doc format (e.g. start with the name of the function)

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,468 @@
...
+	if err != nil {
+		return err
+	}
+	configurationReleaseImages := h.getConfigurationReleaseImages()
+	err = h.setSupportLevels(configurationReleaseImages, supportLevels)
+	if err != nil {
+		return err
+	}
+	apiReleaseImages = h.removeReleaseImageDuplicates(apiReleaseImages)
+	releaseImages := h.mergeReleaseImages(configurationReleaseImages, apiReleaseImages)
```
omertuc, line 248: Supporting this seems like opening a can of worms, what's the use case of merging API release images with user configuration release images?

cmd/main.go
danmanor: `RELEASE_IMAGES` is our current configuration for release images that could be installed, and `RELEASE_SOURCES` is a new configuration that specifies which release images the go routine should collect, basically they are undependable on one  another, but the fact if one of them missing (or both) there won't be nothing to merge later on

cmd/main.go
danmanor: I added the ability to trigger this function when the service start running because I didn't want to create flakiness later on for part 2 in the CI or even every time the service is deployed. Otherwise there will be a period of time when we will have only configuration release images, which is not the desired result and also hard to test

internal/releasesources/release_sources.go
omertuc: Can't we just get all the release images from the `RELEASE_SOURCES` API? Why do we still need to maintain a config if we have API access?

I understand the use case for `RELEASE_SOURCES` in isolation, I understand the user case for `RELEASE_IMAGES` in isolation, but I'm not sure why would anyone would want to have both of them together and why would we support that

danmanor: Because using `RELEASE_SOURCES` we can fetch releases according to `version`, `channel` and `cpu_architecture` - so for example if a client have a special approval to use an old release which we don't support anymore, lets say 4.10.<something> (x86_64), then in order to provide this release using only `RELEASE_SOURCES` we would have to fetch all (stable 4.9 x86_64) from the API, and keep them in the DB only for one release. The other alternative to use only `RELEASE_IMAGE` requires maintaining log configuration lists as we do now but much longer

cmd/main.go
rccrdpccl: Interesting. Can you expand on what test what flaky, or what test would run this code within a thread? I belive production code should be easily teasteable but not made appositely for tests

danmanor: If we don't implement this modification, there will be a duration - from the deployment of the service until the completion of the current interval (which is 30 minutes) - during which the database will not contain any releases. This is because the function has yet to run. As a result, certain tests, specifically those that anticipate retrieving a specific release image from `versionHandler.GetReleaseImage`, will yield different outcomes during this time period compared to after it. This mismatch could affect tests like some subsystem tests.

rccrdpccl: In what tests is this happening? Wouldn't that happen also in real world clusters if that's the case?

I think it's good to trigger it at the startup, but shouldn't we change it in a way that this is what happens *always*, and not only when started with `true`?

danmanor: Hypothetically I agree, but I am not aware of the factors that made the current threads to run first only after 1 interval, in order to stay on the safe side I kept them as they are now

danmanor: I think we can change them also in a separate PR

cmd/main.go
```diff
@@ -407,22 +419,25 @@ func main() {
...
+	clusterStateMonitor.Start(false)
 	defer clusterStateMonitor.Stop()
 
 	hostStateMonitor := thread.New(
 		log.WithField("pkg", "host-monitor"), "Host State Monitor", Options.HostStateMonitorInterval, hostApi.HostMonitoring)
-	hostStateMonitor.Start()
+	hostStateMonitor.Start(false)
 	defer hostStateMonitor.Stop()
 
+	runOpenshiftReleaseSyncerIfNeeded(Options.EnableKubeAPI, Options.EnableOpenshiftReleaseSyncer, releaseSourcesHandler, log)
```
rccrdpccl, line 439: What is the intention here? Run once or run now and each interval?
Asking because AFAIU this is running always at most once

cmd/main.go
rccrdpccl: If this change would break too many tests, I think even running manually once at startup and still starting the "thread" would be better than changing this signature, wouldn't you agree?

cmd/main.go
danmanor: The intention is run now and each interval, I just extracted it to a function because main was getting "cyclomatic complexity" rate too high

cmd/main.go
danmanor: Yes I agree thank you

cmd/main.go
rccrdpccl: I think it's good to extract it to a funciton, the thing is that AFAIU this function closes the thread on completion, so it will run at most once. We should probably defer close from main instead?

danmanor: Oh right sorry, Thank you

cmd/main.go
```diff
@@ -892,3 +909,21 @@ func startPPROF(log *logrus.Logger) {
 		}
 	}
 }
+
+func runOpenshiftReleaseSyncerIfNeeded(kubeAPIMode, enableOpenshiftReleaseSyncer bool, ReleaseSourcesHandler releasesources.ReleaseSourcesHandler, log logrus.FieldLogger) *thread.Thread {
+	if !kubeAPIMode && enableOpenshiftReleaseSyncer {
+		go ReleaseSourcesHandler.SyncReleaseImages()
```
rccrdpccl, line 915: Would it make sense to run synchronously on startup? Not sure if this data is necessary on startup or not

internal/common/common.go
```diff
@@ -54,13 +54,22 @@ const (
 	MultiCPUArchitecture   = "multi"
 
 	ExternalPlatformNameOci = "oci"
+
+	OpenshiftReleaseChannelStable    = "stable"
+	OpenshiftReleaseChannelCandidate = "candidate"
```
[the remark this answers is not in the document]
rccrdpccl: same question with all other the changes in common

internal/common/version.go
```diff
@@ -70,3 +71,18 @@ func GetMajorMinorVersion(version string) (*string, error) {
...
+	str = strings.TrimSuffix(str, "-multi")
+	semVersion, err := version.NewVersion(str)
+	if err != nil {
+		return nil, err
+	}
+
+	return swag.Bool(semVersion.Prerelease() != ""), nil
+}
+
+func CheckIfValidVersion(v string) error {
```
rccrdpccl, line 85: Why do we need to wrap this?

cmd/main.go
danmanor: I think a run of this code can take some time, I don't know if it is better to keep the service waiting. Is there a serious con in running it in the background ?

internal/common/version.go
danmanor: For clarity reasons mainly, as to why we create a version and not use it

danmanor:
Can we merge it? @osherdp  @omertuc  @rccrdpccl  @eranco74 @gamli75 @carbonin 
This is a blocker for me for part 2

Makefile
```diff
@@ -101,6 +101,11 @@ DISABLE_TLS := $(or ${DISABLE_TLS},false)
 ENABLE_ORG_TENANCY := $(or ${ENABLE_ORG_TENANCY},False)
 ALLOW_CONVERGED_FLOW := $(or ${ALLOW_CONVERGED_FLOW}, false)
 ENABLE_ORG_BASED_FEATURE_GATES := $(or ${ENABLE_ORG_BASED_FEATURE_GATES},False)
+OPENSHIFT_MAJOR_VERSION := $(or ${OPENSHIFT_MAJOR_VERSION}, 4)
+OPENSHIFT_RELEASE_API_BASE_URL := $(or ${OPENSHIFT_RELEASE_API_BASE_URL}, https://api.openshift.com/api/upgrades_info/v1/graph)
+OPENSHIFT_SUPPORT_LEVEL_API_BASE_URL := $(or ${OPENSHIFT_SUPPORT_LEVEL_API_BASE_URL}, https://access.redhat.com/product-life-cycles/api/v1/products)
```
osherdp, line 106: We already have the default value in `internal/releasesources/config.go`. Any reason to have it here as well?

osherdp (changes requested):
partial review

Makefile
```diff
@@ -101,6 +101,11 @@ DISABLE_TLS := $(or ${DISABLE_TLS},false)
 ENABLE_ORG_TENANCY := $(or ${ENABLE_ORG_TENANCY},False)
 ALLOW_CONVERGED_FLOW := $(or ${ALLOW_CONVERGED_FLOW}, false)
 ENABLE_ORG_BASED_FEATURE_GATES := $(or ${ENABLE_ORG_BASED_FEATURE_GATES},False)
+OPENSHIFT_MAJOR_VERSION := $(or ${OPENSHIFT_MAJOR_VERSION}, 4)
+OPENSHIFT_RELEASE_API_BASE_URL := $(or ${OPENSHIFT_RELEASE_API_BASE_URL}, https://api.openshift.com/api/upgrades_info/v1/graph)
```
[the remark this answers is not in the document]
danmanor: Thank you, I was not aware of that

Makefile
```diff
@@ -101,6 +101,11 @@ DISABLE_TLS := $(or ${DISABLE_TLS},false)
...
+OPENSHIFT_SUPPORT_LEVEL_API_BASE_URL := $(or ${OPENSHIFT_SUPPORT_LEVEL_API_BASE_URL}, https://access.redhat.com/product-life-cycles/api/v1/products)
+OPENSHIFT_RELEASE_SYNCER_INTERVAL := $(or ${OPENSHIFT_RELEASE_SYNCER_INTERVAL}, 30m)
+ENABLE_OPENSHIFT_RELEASE_SYNCER := $(or ${ENABLE_OPENSHIFT_RELEASE_SYNCER},False)
```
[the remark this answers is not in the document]
danmanor: Yes that is a better idea

internal/releasesources/config.go
```diff
@@ -0,0 +1,7 @@
+package releasesources
+
+type Config struct {
+	OpenshiftMajorVersion           string `envconfig:"OPENSHIFT_MAJOR_VERSION" default:"4"`
```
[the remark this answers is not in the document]
danmanor: This means that I will need to make an API request for every release to get its support level, do we prefer that instead of a configuration option are just hard coding it ?

Makefile
danmanor: We plug these variables as options for `./tools/deploy_assisted_installer_configmap.py` script, which is triggered by test-infra with `skipper make ... deploy-all ...` in the process of deployment of `assisted-service`, so need to give defaults in case it is nut given for this env vars, not sure how `internal/releasesources/config.go` defaults can be used

danmanor:
@osherdp BTW, Do you think I should add the new configuration options in more deployment types?  for example `OKD`, etc.

Makefile
danmanor: No reason I guess, I was following the pattern, I will remove it thanks

internal/releasesources/config.go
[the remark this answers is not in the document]
danmanor: Or at least one for each major, which is basically 1, so it is better I will remove it

Makefile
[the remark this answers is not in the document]
danmanor: The question is what to do in the `.../release-sources` endpoint in this case? return empty string ?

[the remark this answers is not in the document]
danmanor: I think it would be easier with `OPENSHIFT_RELEASE_SYNCER_INTERVAL`, so I will use it

internal/releasesources/config.go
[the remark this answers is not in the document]
osherdp: For each major, yes
So it's still 1 API request

carbonin (changes requested):

cmd/main.go
```diff
@@ -892,3 +909,21 @@ func startPPROF(log *logrus.Logger) {
 		}
 	}
 }
+
+func runOpenshiftReleaseSyncerIfNeeded(kubeAPIMode, enableOpenshiftReleaseSyncer bool, ReleaseSourcesHandler releasesources.ReleaseSourcesHandler, log logrus.FieldLogger) *thread.Thread {
+	if !kubeAPIMode && enableOpenshiftReleaseSyncer {
```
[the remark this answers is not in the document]
danmanor: I had `cyclomatic complexity` issue, I had to extract something out of main.

internal/common/db.go
```diff
@@ -208,6 +208,20 @@ func (i *InfraEnv) Payload() any {
...
 
+type ReleaseImage struct {
+	CreatedAt time.Time
+	models.ReleaseImage
+
+	// The release channel from which the release was fetched from
+	Channel *string
+}
+
+type OpenshiftVersionSupportLevel struct {
```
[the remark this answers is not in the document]
danmanor: I will remove it. before we had one source of truth, I decided to have a table of support levels in the DB as well in order to get the support levels for the static release images in `versions` code, is it is missing (which makes sense as it is dynamic). Thank you

internal/common/db.go
```diff
@@ -256,6 +272,47 @@ func AutoMigrate(db *gorm.DB) error {
...
+	if err := db.Model(&ReleaseImage{}).Find(&dbReleaseImages, where...).Error; err != nil {
+		return nil, err
+	}
+
+	for _, releaseImage := range dbReleaseImages {
+		cpuArchitectures := []string{*releaseImage.CPUArchitecture}
+		if *releaseImage.CPUArchitecture == MultiCPUArchitecture {
+			cpuArchitectures = SupportedMultiArchitectures
+		}
+		releaseImage.CPUArchitectures = cpuArchitectures
```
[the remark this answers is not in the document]
danmanor: I agree it is not optimal, but unfortunately I didn't find any way to obtain this information dynamically. I looked into the builds and saw that since `4.11` we are supporting this 4 architectures for every multi release, do you have an idea how to go around it?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,469 @@
...
+}
+
+func (h *ReleaseSourcesHandler) deleteAllReleases(tx *gorm.DB) error {
+	var count int64
+	err := tx.Model(&common.ReleaseImage{}).Count(&count).Error
+	if err != nil {
+		return err
+	}
+	h.log.Debugf("Deleting all openshift_versions table records. %d records", count)
+	return tx.Where("TRUE").Delete(&common.ReleaseImage{}).Error
```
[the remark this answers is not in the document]
danmanor: I was not aware of `TRUNCATE` thank you

Makefile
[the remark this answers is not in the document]
osherdp: > The question is what to do in the .../release-sources endpoint in this case

return an empty list. it sounds mathematically right in this case :)

cmd/main.go
rccrdpccl: Potentially, the con of running in background is that if we need those versions for the application to work, it won't work until the task is done

osherdp: @rccrdpccl we need those versions for the application to run? I'm not sure about it
we might miss the "dynamic" versions for some time, but it shouldn't make the application crash or something like that

osherdp:
> @osherdp BTW, Do you think I should add the new configuration options in more deployment types? for example `OKD`, etc.

no no. keep it scoped and only if a request comes then we can expand that

Makefile
danmanor: OK, thinking about this again, I need these defaults because of the way we currently deploy the `configmap`, We specify the options, so a value must be given, therefore I need to give it the default

internal/releasesources/clients.go
```diff
@@ -0,0 +1,146 @@
...
+//go:generate mockgen -source=clients.go -destination=mock_clients.go -package=releasesources
+type OpenShiftReleasesAPIClientInterface interface {
+	getEndpoint(channel, openshiftVersion, cpuArchitecture string) string
+	parseResponse(response *http.Response) (*releaseGraph, error)
+	GetReleases(channel, openshiftVersion, cpuArchitecture string) (*releaseGraph, error)
+}
+
+//go:generate mockgen -source=clients.go -destination=mock_clients.go -package=releasesources
+type OpenShiftSupportLevelAPIClientInterface interface {
+	getEndpoint(openshiftMajorVersion string) string
```
tsorya, line 28: Why we need private methods in interface?

internal/releasesources/clients.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (o OpenShiftSupportLevelAPIClient) getEndpoint(openshiftMajorVersion string) string {
+	return fmt.Sprintf(supportLevelAPIQueryTemplate, o.BaseUrl, openshiftMajorVersion)
+}
+
+func (o OpenShiftSupportLevelAPIClient) parseResponse(response *http.Response) (*supportLevelGraph, error) {
+	var graph supportLevelGraph
+	err := json.NewDecoder(response.Body).Decode(&graph)
+	if err != nil {
+		return nil, err
```
tsorya, line 107: Lets add more description to err

internal/releasesources/clients.go
```diff
@@ -0,0 +1,146 @@
+package releasesources
```
tsorya, line 1: I assume this sort of ocm client?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,478 @@
...
+				}
+
+				if shouldSwitch {
+					if arch, ok := cpuArchMapFromAPIArch[cpuArchitecture]; ok {
+						cpuArchitecture = arch
+					}
+				}
+
+				for _, node := range graph.Nodes {
+					if strings.HasPrefix(node.Version, openshiftVersion) {
```
tsorya, line 170: will not 4.11 match 4.1 in this case?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,478 @@
...
+						// Got a match for OpenShift ReleaseImage
+						newReleaseImage, err := h.createReleaseImage(channel, node.Version, cpuArchitecture)
```
tsorya, line 172: I believe it will not be running too much? Maybe lets log each new release image?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,478 @@
...
+					cpuArchitecture = apiCpuArchitecture
+				}
+
+				graph, err := h.releasesClient.GetReleases(channel, openshiftVersion, cpuArchitecture)
+				if err != nil {
+					h.log.WithError(err).Warnf("failed to get releases from api with: %s-%s, %s", channel, openshiftVersion, cpuArchitecture)
+					continue
+				}
+
+				if shouldSwitch {
```
tsorya, line 163: Why this shouldSwitch is needed? can't we just get it from map? I mean it is not enough to get cpu arch from mapping dict only once 
https://github.com/openshift/assisted-service/pull/5916/files#diff-22441f7a4340c9547ef0e7b14e79b43caca92e10721eb318513bfb480653aa45R157 ?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,478 @@
...
+	h.log.Debugf("Got %d releases", len(releases))
+	return releases, nil
+}
+
+func (h *ReleaseSourcesHandler) getStaticReleaseImages() []*common.ReleaseImage {
+	dbReleaseImages := []*common.ReleaseImage{}
+
+	for _, releaseImage := range h.staticReleaseImages {
+		dbReleaseImage := common.ReleaseImage{ReleaseImage: *releaseImage}
+		dbReleaseImage.CPUArchitectures = nil
```
tsorya, line 192: Why nil is needed? Can you comment ?

internal/releasesources/clients.go
danmanor: I think the confusion comes from the fact I wrote this interface as public instead of private, it is intended only for this package

internal/releasesources/clients.go
danmanor: Isn't the description for the error in `GetReleases` sufficient ?

internal/releasesources/release_sources.go
danmanor: Yes basically the results are only for the specified version or adjacent ones e.g. "4.13", "4.14", "4.15" but I agree I should change it to avoid this matter, thank you

internal/releasesources/release_sources.go
danmanor: Won't it flood the logs? there could be hundreds of releases

internal/releasesources/release_sources.go
danmanor: The API requires some terminology that might be different from ours, for example `x86_64` should be changed to `amd64`. I wanted the initial specification to remain, meaning if one day we will decide to name it `amd64` it will not brake this code

internal/releasesources/clients.go
tsorya: There is no OCM client provided by ocm?

danmanor: I did not find any. There is an OCM client for golang but it doesn't provide a way to reach `OpenShift Cincinnati`

swagger.yaml
```diff
@@ -3131,6 +3131,32 @@ paths:
...
+      description: Retrieves openshift release sources configuration.
+      responses:
+        "200":
+          description: Success.
+          schema:
+            $ref: '#/definitions/release-sources'
+        "400":
+          description: Bad Request
+          schema:
+            $ref: '#/definitions/error'
```
osherdp, line 3150: Can we really return 4xx in any case? There's no body/parameters for this endpoint so I don't see how that's an option

swagger.yaml
```diff
@@ -6937,7 +6963,7 @@ definitions:
         description: Name of the version to be presented to the user.
       support_level:
         type: string
-        enum: [beta, production, maintenance]
+        enum: [beta, production, maintenance, end of life]
```
osherdp, line 6966:
```suggestion
        enum: [beta, production, maintenance, end-of-life]
```

to match the convention we're using for support-level

swagger.yaml
```diff
@@ -7002,9 +7028,11 @@ definitions:
         items:
           type: string
         description: List of CPU architectures provided by the image.
+        x-go-custom-tag: gorm:"type:text[]"
```
osherdp, line 7031: does it require any DB migration? what this means?
question holds for any `gorm` option being added/changed

swagger.yaml
```diff
@@ -3131,6 +3131,32 @@ paths:
...
+          schema:
+            $ref: '#/definitions/error'
+        "500":
+          description: Error.
+          schema:
+            $ref: '#/definitions/error'
+        "503":
+          description: Unavailable.
+          schema:
+            $ref: '#/definitions/error'
```
osherdp, line 3158: let's delete. it applies to all endpoints and doesn't add any added information

swagger.yaml
```diff
@@ -7013,9 +7041,46 @@ definitions:
...
+      cpu_architecture:
+        type: string
+        enum: ['x86_64', 'aarch64', 'arm64','ppc64le','s390x','multi']
+        description: The CPU architecture of the image.
+        x-go-custom-tag: gorm:"default:'x86_64'"
+      channels:
+        type: array
+        items:
+          type: string
+          enum: [stable, candidate]
```
osherdp, line 7062: let's add `fast` and `eus` (in general, all possible channels: https://docs.openshift.com/container-platform/4.14/updating/understanding_updates/understanding-update-channels-release.html )

osherdp (changes requested):

swagger.yaml
danmanor: I was contemplating whether to use `eus` and `fast` as well. I decided to not incluse them as IMHO they don't give us more functionality, but they do add complexity to  `releasesources` logic, so I decided to omit them. all releases in `fast` channel will soon be in `sable` as well (probably, or they shouldn't), and `eus` doesn't really say anything extra for us because every release in `eus` is also in `stable` .
WDYT?

osherdp: What kind of complexity does it add?

danmanor: Thinking about this again I think after the recent changes we don't need to store `channel` field in the DB, so it reduces the complexity of adding these channels.

swagger.yaml
danmanor: What about 500? we should not hit that also

swagger.yaml
danmanor: This one is a left over from when I considered saving the CPU architectures in the DB, but it is not relevant anymore, thanks

danmanor: As for the `primaryKey`, I am not sure what kind of migration do you mean? this is a new table so it will apply from the get-go

internal/releasesources/release_sources_test.go
```diff
@@ -0,0 +1,1348 @@
...
+					},
+				}
+
+				err := handler.syncReleaseImagesWithErr()
+				Expect(err).To(HaveOccurred())
+			})
+		})
+	})
+
+	Context("Functionality", func() {
```
carbonin, line 405: This context is not needed.

internal/versions/api.go
```diff
@@ -29,17 +29,28 @@ type apiHandler struct {
 	log             logrus.FieldLogger
 	versionsHandler *handler
 	osImages        OSImages
+	releaseSources  models.ReleaseSources
+}
+
+type NewAPIHandlerParams struct {
```
carbonin, line 35: At this point why even have a `NewAPIHandler` function, just export apiHandler and have the caller set the fields directly.

swagger.yaml
```diff
@@ -7013,9 +7029,46 @@ definitions:
...
+        type: array
+        items:
+          $ref: '#/definitions/upgrade-channel'
+
+  release-sources:
+    type: array
+    items:
+      $ref: '#/definitions/release-source'
+
   release-images:
```
carbonin, line 7072: Super minor but maybe leave this next to `release-image` instead of putting the sources stuff in between the singular and array types.

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,499 @@
...
+	var count int64
+	err := tx.Model(&models.ReleaseImage{}).Count(&count).Error
+	if err != nil {
+		return err
+	}
+	h.log.Debugf("Truncating all release_images table records. %d records", count)
+	return tx.Exec("TRUNCATE TABLE release_images").Error
+}
+
+func (h *releaseSourcesHandler) syncReleaseImagesWithErr() error {
```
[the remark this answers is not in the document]
danmanor: I can't return an error with `SyncReleaseImages` because of the way `thread` package is currently functioning, so I  created a wrapper function that can return an error, which is much easier to test with

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,499 @@
...
+			h.log.Debugf("Failed to create a version struct from %s", *releaseImage.Version)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func getReleaseImageReference(version, cpuArchitecture string) string {
+	if cpuArchitecture == common.ARM64CPUArchitecture {
```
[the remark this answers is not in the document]
danmanor: It cannot be `amd64` currently, only `x86_64`. About [common.NormalizeCPUArchitecture](https://github.com/openshift/assisted-service/blob/759e3e01336c3a7dc5da43f265179ecc470cf6f6/internal/common/common.go#L135-L144), it doesn't achieve what I am trying to do

internal/versions/api.go
danmanor: Reverted params struct

danmanor:
@carbonin I fixed your comments. I tested on the new inspection on my beaker machine, seems to not consume too much resources

cmd/main.go
```diff
@@ -895,3 +914,31 @@ func startPPROF(log *logrus.Logger) {
...
+	db *gorm.DB,
+	log logrus.FieldLogger,
+	releaseHandler oc.Release,
+) *thread.Thread {
+	if !Options.EnableKubeAPI && Options.ReleaseSourcesConfig.ReleaseSources != "" {
+		releaseSourcesHandler, err := releasesources.NewReleaseSourcesHandler(releaseSources, releaseImages, log, db, Options.ReleaseSourcesConfig, releaseHandler)
+		if err != nil {
+			log.WithError(err).Debug("error occurred while creating release sources handler")
+		}
+		go releaseSourcesHandler.SyncReleaseImages()
```
tsorya, line 930: Can you describe why you are starting go routine and then running thread? Why running thread is not enough ?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,510 @@
...
+		if dynamicRelease.Default && defaultStaticReleaseExists {
+			dynamicRelease.Default = false
+		}
+		mergedReleaseImages = append(mergedReleaseImages, dynamicRelease)
+	}
+
+	return mergedReleaseImages
+}
+
+func (h *releaseSourcesHandler) syncReleaseImagesWithErr() error {
```
tsorya, line 260: I think it should be blocked with leader elector no? What will happened if 2 instances will run that code at the same time?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,510 @@
...
+		openshiftVersion := *releaseSource.OpenshiftVersion
+		for _, upgradeChannel := range releaseSource.UpgradeChannels {
+			cpuArchitecture := *upgradeChannel.CPUArchitecture
+			for _, channel := range upgradeChannel.Channels {
+				apiCpuArchitecture, shouldSwitch := cpuArchMapToAPIArch[cpuArchitecture]
+				if shouldSwitch {
+					cpuArchitecture = apiCpuArchitecture
+				}
+
+				graph, err := h.releasesClient.getReleases(channel, openshiftVersion, cpuArchitecture)
```
tsorya, line 181: As far as i understand this is your client no? In this case i think better to make 
apiCpuArchitecture, shouldSwitch := cpuArchMapToAPIArch[cpuArchitecture] inside of it.
In anycase i don't understand why we need to change it back as here 
`if arch, ok := cpuArchMapFromAPIArch[cpuArchitecture]; ok {
						cpuArchitecture = arch
					}`
why not to simply use apiCpuArchitecture as param to h.releasesClient.getReleases(channel, openshiftVersion, cpuArchitecture ?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,510 @@
...
+
+// removeReleaseImageDuplicates removes release images such that only one release image
+// of each version, CPU architecture will remain
+func (h *releaseSourcesHandler) removeReleaseImageDuplicates(releases []*common.ReleaseImage) ([]*common.ReleaseImage, error) {
+	releaseImagesMap := map[string]*common.ReleaseImage{}
+
+	for _, releaseImage := range releases {
+		duplicateDBReleaseImage, exists := releaseImagesMap[*releaseImage.URL]
+		if exists {
+			// If the release image that already exists or its diplicate are labeled as defualt,
```
tsorya, line 127: nit: defualt -> default

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,510 @@
...
+	h.log.Debugf("Inserting %d records to openshift_versions table", len(dbReleaseImages))
+
+	if len(dbReleaseImages) == 0 {
+		return nil
+	}
+
+	return tx.Create(dbReleaseImages).Error
+}
+
+// removeReleaseImageDuplicates removes release images such that only one release image
```
tsorya, line 119: Why we will have dumplicates?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,510 @@
...
+	h.log.Debugf("Found %d dynamic release Images", len(dynamicDBReleaseImages))
+	supportLevels, err := h.getSupportLevels()
+	if err != nil {
+		return err
+	}
+	err = h.setSupportLevels(dynamicDBReleaseImages, supportLevels)
+	if err != nil {
+		return err
+	}
+	staticDBReleaseImages := h.getStaticReleaseImages()
```
tsorya, line 283: static images can change without service restart? maybe we can list them once?

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,510 @@
...
+
+	dynamicDBReleaseImages, err = h.removeReleaseImageDuplicates(dynamicDBReleaseImages)
+	if err != nil {
+		return err
+	}
+
+	dbReleaseImages := h.mergeDBReleaseImages(staticDBReleaseImages, dynamicDBReleaseImages)
+
+	h.log.Debug("Starting SQL transaction")
+	tx := h.db.Begin()
```
tsorya, line 302: why not to use same template we use at all other places 
`txSuccess := false
	tx := h.db.Begin()
	defer func() {
		if !txSuccess {
			log.Error("update monitored operator failed")
			tx.Rollback()
		}
		if r := recover(); r != nil {
			log.Error("update monitored operator failed")
			tx.Rollback()
		}
	}()`
no idea though why we are not making it a function to be reusable

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,510 @@
...
+		}
+		majorVersions[*majorVersion] = true
+	}
+
+	return majorVersions, nil
+}
+
+// getSupportLevels retrieves a mapping from OCP major.minor versions
+// to their corresponding support levels for both static and dynamic release images.
+func (h *releaseSourcesHandler) getSupportLevels() (ocpVersionSupportLevels, error) {
```
tsorya, line 348: interesting how many api calls we will produce each run? Afraid of ddos of ocm a little bit

cmd/main.go
danmanor: The way `thread` package currently behaving is running the first time after 1 interval. Therefore I ran it once before because I wanted to have the release images in the DB when assisted-service starts

internal/releasesources/release_sources.go
danmanor: Every run all the previous releases are replaced with new ones inside a transaction, so we should be good in that context. Will be more efficient though to have only one instance running it  instead of all

internal/releasesources/release_sources.go
danmanor: First, most releases exist in more than one channel, so if we collect for example 4.14 releases from `stable` and `candidate`, we will have the same release twice. Second I thought it might be possible for the API to mistake and give me duplicate records, which will cause us an error as the `url` (whic his composed of `version` and `arch`) is the primary key.

internal/releasesources/release_sources.go
danmanor: AFAIK in `RESTAPI` mode they shouldn't be changed without a restart from the service. Not sure I understand the part about listing them, do you mean logging them for debugging ?

internal/releasesources/release_sources.go
danmanor: I think the main issue with this is the `recover`, I had a talk about it with @filanov and @rccrdpccl, we came to a conclusion that hides errors, therefore we should not recover.

danmanor: We can create a function without recovering though

internal/releasesources/release_sources.go
danmanor: for - `OCM`, for example if we have this `ReleaseSources`:

```
var releaseSources = models.ReleaseSources{
	{
		OpenshiftVersion: swag.String("4.10"),
		UpgradeChannels: []*models.UpgradeChannel{
			{
				CPUArchitecture: swag.String(common.X86CPUArchitecture),
				Channels:        []string{OpenshiftReleaseChannelStable, OpenshiftReleaseChannelCandidate},
			},
                        {
				CPUArchitecture: swag.String(common.ARM64CPUArchitecture),
				Channels:        []string{OpenshiftReleaseChannelStable, OpenshiftReleaseChannelCandidate},
			},
		},
	},
        .
        .
        .
}
```

we will have an API call for each combination so -
```
4.10, x86_64, stable
4.10, x86_64, candidate
4.10, ARM64CPUArchitecture, stable,
4.10, ARM64CPUArchitecture, candidate,
``` 

So assuming we collect `(x86_64, arm64, aarch64, s390x, ppcle64, multi)`, from 2 channels `(candidate, stable)`, for about 4 versions each time (2 production, 2 maintenance), we should sum up to about
6 * 2 * 4 = 48 API calls every run, (every 30 min) I think we should be good.

internal/releasesources/release_sources.go
filanov: all the recovers should already be removed, we replaced the code with using a new format that will fail the transaction and handle the rollback https://github.com/openshift/assisted-service/blob/a8dca1f6d37e0e17b37c93e1c24426fc9f80ff1c/internal/bminventory/inventory.go#L1288 
otherwise this recover will return a `nil, nil` as a replay and the caller will think that everything is ok

cmd/main.go
carbonin: If you want it there when the service starts then don't do a go routine.

Do we know how long this takes?

internal/releasesources/release_sources.go
carbonin: Also I think every connected OCP cluster in the world uses these endpoints. Our few deployments of our service will be a drop in the bucket.

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,522 @@
...
+	var count int64
+	err := tx.Model(&models.ReleaseImage{}).Count(&count).Error
+	if err != nil {
+		return err
+	}
+	h.log.Debugf("Truncating all release_images table records. %d records", count)
+	return tx.Exec("TRUNCATE TABLE release_images").Error
+}
+
+func (h *releaseSourcesHandler) SyncReleaseImages() {
```
carbonin, line 458: I'd name this function specifically for how it's used because it's more the exception and we really shouldn't call this anywhere except as the thread function.

So maybe something like `SyncReleaseImagesThreadFunc`?

internal/releasesources/release_sources.go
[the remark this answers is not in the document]
carbonin: :shrug: sure

internal/releasesources/release_sources.go
```diff
@@ -0,0 +1,528 @@
...
+
+// setSupportLevels sets support level for given release images as follows:
+// candidate/pre-release releases are considered support level 'beta'.
+// The rest are stable releases, and their support level is set according to the mapping
+// retrieved from the API.
+func (h *releaseSourcesHandler) setSupportLevels(releases []*enrichedReleaseImage, supportLevels ocpVersionSupportLevels) error {
+	for _, release := range releases {
+		if release.SupportLevel != "" {
+			continue
+		}
```
carbonin, line 392: This will need to be removed if we persist enrichedReleaseImages for static.

Currently this will always we empty for static release images, but if we save them it will always have the value from the previous loop. To deal with this we can just always set the level to the new one for this loop.

swagger.yaml
```diff
@@ -7013,8 +7035,55 @@ definitions:
...
+    properties:
+      openshift_version:
+        type: string
+        description: Version of the OpenShift cluster.
+        example: "4.14"
+      multi_cpu_architectures:
+        type: array
+        items:
+          type: string
+          enum: ['x86_64', 'aarch64', 'arm64','ppc64le','s390x','multi']
```
carbonin, line 7056: I think t his should only include one of `aarch64` or `arm64` and probably shouldn't include `multi`, right?

swagger.yaml
```diff
@@ -7013,8 +7035,55 @@ definitions:
...
+
+  upgrade-channel:
+    type: object
+    required:
+      - cpu_architecture
+      - channels
+    properties:
+      cpu_architecture:
+        type: string
+        enum: ['x86_64', 'aarch64', 'arm64','ppc64le','s390x','multi']
```
carbonin, line 7071: Same here I don't think we should have both `arm64` and `aarch64` also is `multi` actually valid for a channel?

swagger.yaml
danmanor: About `aarch64/arm64`, currently we have these archs available for release images - https://github.com/openshift/assisted-service/blob/59f99694cfb82f33f8e7461d02eb46323e5abbc4/swagger.yaml#L6996
Not sure about whether we should have both of them or only one

swagger.yaml
danmanor: Same as for `arm64/aarch64` as last comment, but here multi is a valid release to pick from the API

PR #5940, opened 2024-02-01
MGMT-16721: handle hosts with no luks

paul-maidment:
In line 41 of internal/host/hostcommands/tang_connectivity_check_cmd.go
Both `luks` and `luks.Clevis` are pointers which can be nil, there was no checking in place to ensure that we avoid a nil pointer dereference

This PR fixes that by making appropriate checks to ensure that neither of these pointers are nil.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/hostcommands/tang_connectivity_check_cmd.go
+++ b/internal/host/hostcommands/tang_connectivity_check_cmd.go
@@ -38,6 +38,14 @@ func (c *tangConnectivityCheckCmd) getTangServersFromHostIgnition(host *models.H
 		if err != nil {
 			return nil, err
 		}
+		if luks == nil {
+			return nil, nil
+		}
+		if luks.Clevis == nil {
+			// This cluster does not use tang, so we shouldn't perform a tang connectivity check
+			c.log.Warn("luks clevis configuration is missing or incomplete in the host ignition, disk encryption will be assumed to be disabled. (MGMT-16721)")
+			return nil, nil
+		}
 		if len(luks.Clevis.Tang) != 0 {
 			if tangServers, err = json.Marshal(luks.Clevis.Tang); err != nil {
 				return nil, err
--- a/internal/host/hostcommands/tang_connectivity_check_cmd_test.go
+++ b/internal/host/hostcommands/tang_connectivity_check_cmd_test.go
@@ -39,36 +39,88 @@ var _ = Describe("tangConnectivitycheckcmd", func() {
 
 	})
 
-	It("get_step: Tang imported cluster should use host ignition", func() {
+	const hostIgnition = `{
+		"ignition": {
+		  "config": {},
+		  "version": "3.2.0"
+		},
+		"storage": {
+			"luks": [
+				{
+				  "clevis": {
+					"tang": [
+					  {
+						"thumbprint": "nWW89qAs1hDPKiIcae-ey2cQmUk",
+						"url": "http://foo.bar"
+					  }
+					]
+				  },
+				  "device": "/dev/disk/by-partlabel/root",
+				  "name": "root",
+				  "options": [
+					"--cipher",
+					"aes-cbc-essiv:sha256"
+				  ],
+				  "wipeVolume": true
+				}
+			],
+		  "files": []
+		}
+	  }`
 
-		const hostIgnition = `{
-			"ignition": {
-			  "config": {},
-			  "version": "3.2.0"
-			},
-			"storage": {
-				"luks": [
-					{
-					  "clevis": {
-						"tang": [
-						  {
-							"thumbprint": "nWW89qAs1hDPKiIcae-ey2cQmUk",
-							"url": "http://foo.bar"
-						  }
-						]
-					  },
-					  "device": "/dev/disk/by-partlabel/root",
-					  "name": "root",
-					  "options": [
-						"--cipher",
-						"aes-cbc-essiv:sha256"
-					  ],
-					  "wipeVolume": true
-					}
-				],
-			  "files": []
-			}
-		  }`
+	const hostIgnitionWithoutLuks = `{
+		"ignition": {
+		  "config": {},
+		  "version": "3.2.0"
+		},
+		"storage": {
+		  "files": []
+		}
+	  }`
+
+	const hostIgnitionWithoutClevis = `{
+		"ignition": {
+		  "config": {},
+		  "version": "3.2.0"
+		},
+		"storage": {
+			"luks": [
+			],
+		  "files": []
+		}
+	  }`
+
+	It("Should not panic if Luks is undefined", func() {
+		imported := true
+		cluster.Imported = &imported
+		Expect(db.Create(&cluster).Error).ShouldNot(HaveOccurred())
+		apiVipConnectivity, err := json.Marshal(models.APIVipConnectivityResponse{
+			IsSuccess: true,
+			Ignition:  hostIgnitionWithoutLuks,
+		})
+		Expect(err).ToNot(HaveOccurred())
+		host.APIVipConnectivity = string(apiVipConnectivity)
+		Expect(db.Save(&host).Error).ShouldNot(HaveOccurred())
+		stepReply, stepErr = tangConnectivityCheckCmd.GetSteps(ctx, &host)
+		Expect(stepErr).ShouldNot(HaveOccurred())
+	})
+
+	It("Should not panic if Clevis is undefined", func() {
+		imported := true
+		cluster.Imported = &imported
+		Expect(db.Create(&cluster).Error).ShouldNot(HaveOccurred())
+		apiVipConnectivity, err := json.Marshal(models.APIVipConnectivityResponse{
+			IsSuccess: true,
+			Ignition:  hostIgnitionWithoutClevis,
+		})
+		Expect(err).ToNot(HaveOccurred())
+		host.APIVipConnectivity = string(apiVipConnectivity)
+		Expect(db.Save(&host).Error).ShouldNot(HaveOccurred())
+		stepReply, stepErr = tangConnectivityCheckCmd.GetSteps(ctx, &host)
+		Expect(stepErr).ShouldNot(HaveOccurred())
+	})
+
+	It("get_step: Tang imported cluster should use host ignition", func() {
 
 		By("Set cluster to imported and inject host ignition for test", func() {
 			imported := true
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -420,7 +420,7 @@ func (v *validator) diskEncryptionRequirementsSatisfied(c *validationContext) (V
 			return ValidationPending, "Missing ignition information"
 		}
 		if luks == nil || luks.Clevis == nil {
-			// Disk encryption is disabled for workers on day1 cluster
+			// No tang servers to validate for the target cluster
 			return ValidationSuccessSuppressOutput, ""
 		}
 		c.cluster.DiskEncryption = &models.DiskEncryption{}


gamli75:
@rccrdpccl please review

omertuc:
What does it mean for Luks to be present in an ignition file but for Clevis to be nil? Is that a valid situation? What is the expected behavior from the service in that situation? To ignore it? To report an error? 

Please elaborate

rccrdpccl:
> What does it mean for Luks to be present in an ignition file but for Clevis to be nil? Is that a valid situation? What is the expected behavior from the service in that situation? To ignore it? To report an error?
> 
> Please elaborate

Could you please also change description of the issue and commit according to the answer for the question above :)?

paul-maidment:
> What does it mean for Luks to be present in an ignition file but for Clevis to be nil? Is that a valid situation? What is the expected behavior from the service in that situation? To ignore it? To report an error?
> 
> Please elaborate

I just observed that both luks and clevis are pointers and can have a value of nil.
Simply based on the fact that pointers _can_ be nil, (whether or not this is valid) is reasonable grounds to check for nil I believe.

omertuc:
Even something like `// TODO: not sure how this could happen or whether its possibly valid, for now pretend encryption is disabled just so we avoid a nil pointer dereference` is also completely valid, as it makes things better than they were before

paul-maidment:
> Even something like `// TODO: not sure how this could happen or whether its possibly valid, for now pretend encryption is disabled just so we avoid a nil pointer dereference` is also completely valid, as it's obviously improving the previous situation by avoiding a crash, but at least we can recognize that this needs to be thought out more thoroughly and needs looking into

I'll go with this suggestion and will also add test cases for luks == nil and luks.clevis == nil, just to ensure we have the best coverage.

Seems to me that just because a cluster is imported does not necessarily imply that there needs to be disk encryption, maybe we are simply covering the case where an imported cluster does not have disk encryption?

Certainly the case of missing `luks` we can imagine this to be the case

Missing `luks.Clevis` I do agree that it's odd this should be optional.

So for now, I will go with the comment you suggest

internal/host/hostcommands/tang_connectivity_check_cmd.go
```diff
@@ -38,6 +38,11 @@ func (c *tangConnectivityCheckCmd) getTangServersFromHostIgnition(host *models.H
 		if err != nil {
 			return nil, err
 		}
+		if luks == nil || luks.Clevis == nil {
+			// TODO: not sure how this could happen or whether its possibly valid, for now pretend encryption is disabled just so we avoid a nil pointer dereference
+			c.log.Warn("luks configuration is missing or incomplete in the host ignition, disk encryption will be assumed to be disabled.")
```
omertuc, line 43: luks being nil is pretty common, it doesn't warrant a warning

paul-maidment: OK, removed the log entry.

omertuc: My comment was about only `luks` being `nil`, which doesn't warrant a warning.

But `luks` being non-nil and `Clevis` nil does warrant a warning, that's the case I've been discussing in all my comments so far, the one I don't understand

paul-maidment: OK, updated to deal with that scenario.

rccrdpccl:
I'm OK with merging just to avoid crashing, but we need to get to the bottom of this to make sure we are not ignoring bigger flaws:
* Was this crashing because Luks was nil (does this mean a host doesn't have luks on a given disk and it's normal, or what else could it mean given the context we are running this check from?)
* Was this crashing because Clevis was nil?
* What does it mean having luks with no clevis and what implications does it have on the rest of the code?

Also I think the message we find in the logs should be `MGMT-XXXX: avoid crashing when luks not present` or something, ideally with the real reason

/hold

Holding for title change, once changed feel free to unhold

paul-maidment:
> I'm OK with merging just to avoid crashing, but we need to get to the bottom of this to make sure we are not ignoring bigger flaws:
> 
>     * Was this crashing because Luks was nil (does this mean a host doesn't have luks on a given disk and it's normal, or what else could it mean given the context we are running this check from?)
> 
>     * Was this crashing because Clevis was nil?
> 
>     * What does it mean having luks with no clevis and what implications does it have on the rest of the code?
> 
> 
> Also I think the message we find in the logs should be `MGMT-XXXX: avoid crashing when luks not present` or something, ideally with the real reason
> 
> /hold
> 
> Holding for title change, once changed feel free to unhold

I think we would need to get more detailed information than the stuff we have in the ticket in order to understand why these fields were nil. As I have alluded to in previous comments, having "luks" as nil is not abnormal, this just means the customer is not using disk encryption.

the check of luks.Clevis, I don't know why it would be nil, if it was nil in this case, I just know that it CAN be nil as someone designed it as a pointer. It's only good hygene to check for nil.

Not quite sure how to proceed to satisfy your concern here...

paul-maidment:
> > Not quite sure how to proceed to satisfy your concern here...
> 
> Short term: As requested, the log should contain a JIRA ticket reference which seeks to further explore this topic
> 
> Long term: The work on that ticket should include understanding the meaning of Clevis and whether it can be nil. Once we gain that understanding, we can decide how our service should behave in the face of it being nil
> 
> > As I say, only checking for nil because the API suggests the field can be nil, I have no idea why it would be.
> 
> This is why we need to gain understanding. Obviously it can be `nil`, cause it's a pointer, but what does it mean, and what should we do - should we block the installation because we don't know how to proceed? Should we ignore it and pretend there's no encryption on that cluster? Maybe it means there's different type of encryption our service is not familiar with, and so it should block the installation? etc

According to https://github.com/coreos/ignition/blob/main/config/v3_2/types/luks.go#L48
This field can be nil, there is an explicit check for this.

paul-maidment:
> > Not quite sure how to proceed to satisfy your concern here...
> 
> Short term: As requested, the log should contain a JIRA ticket reference which seeks to further explore this topic
> 
> Long term: The work on that ticket should include understanding the meaning of Clevis and whether it can be nil. Once we gain that understanding, we can decide how our service should behave in the face of it being nil
> 
> > As I say, only checking for nil because the API suggests the field can be nil, I have no idea why it would be.
> 
> This is why we need to gain understanding. Obviously it can be `nil`, cause it's a pointer, but what does it mean, and what should we do - should we block the installation because we don't know how to proceed? Should we ignore it and pretend there's no encryption on that cluster? Maybe it means there's different type of encryption our service is not familiar with, and so it should block the installation? etc

OK, to provide a more detailed response, I looked into clevis/tang when doing the original feature and the `clevis` component is responsible for making it possible to automate boot from an encrypted volume. It accepts the LUKS key and encrypts it, securely storing it using TPM. This is then used during boot to automate the entry to the decryption key.

The ignition specification makes it possible to make this nil. https://github.com/coreos/ignition/blob/main/config/v3_2/types/luks.go#L48

It seems like a reasonable conjecture that this is nil because the ignition could be used for the provisioning of a LUKS installation where automated login is not desirable (such as the provisioning of a work laptop where the key should be expected from the user, for obvious reasons)

So in the case of _us_ receiving a nil in this field, I suspect that we should raise an error because this means that we wont be able to boot the LUKS volume as we lack the key.

Does this make sense?

paul-maidment:
I have created ticket https://issues.redhat.com/browse/MGMT-16721 as an investigation ticket in response to this.

rccrdpccl:
/retitle MGMT-16721: handle hosts with no luks

internal/host/hostcommands/tang_connectivity_check_cmd.go
```diff
@@ -38,6 +38,15 @@ func (c *tangConnectivityCheckCmd) getTangServersFromHostIgnition(host *models.H
 		if err != nil {
 			return nil, err
 		}
+		if luks == nil {
+			return nil, nil
+		}
+		if luks.Clevis == nil {
+			// TODO: not sure how this could happen or whether its possibly valid, for now pretend encryption is disabled just so we avoid a nil pointer dereference
+			// Investigation of this will be carried out in https://issues.redhat.com/browse/MGMT-16721
+			c.log.Warn("luks clevis configuration is missing or incomplete in the host ignition, disk encryption will be assumed to be disabled.")
```
rccrdpccl, line 47: Please add the investigation issue ID inside the actual error message: this will save time for people looking at the logs

omertuc:
>So in the case of us receiving a nil in this field, I suspect that we should raise an error because this means that we wont be able to boot the LUKS volume as we lack the key.
>
>Does this make sense?

Not really, we're discussing validations for adding hosts to an existing cluster here. 

This entire ordeal of looking at the ignition to get tang details is just so we can perform a validation that the tang config specified in the target cluster actually works on this new day-2 host - we know how to recognize and validate tang, so we look into the clevis struct to get tang servers (when those are present), so we can send those details to our agent so it can attempt to connect to the tang servers, followed by the agent reporting about what it experienced doing that, so that our validation can ensure tang works.

If you see the cluster has encryption (i.e. has the luks struct) but it's not using using the `clevis` struct, and you throw an error because you don't recognize it, then you are implicitly making an assumption of how that encryption works (in particular, you're assuming that it wouldn't work and it deserves and error just because you have no way to validate that it works), but perhaps it just means that our validation should be skipped (both in the `tang_connectivity_check_cmd.go` and the corresponding validation code that awaits the results from the agent).

The reasoning of skipping instead of giving an error is that the validation was best-effort in the first place - we're doing it as a "favor" to the user. We can validate tang so we do. If the user uses some other scheme we don't recognize, we should still let them add the host to the cluster, and if they run into issues, that's unfortunate, but it's none of our business, because it just might work. We can't assume it won't because we don't know what it is

Also regardless of whether you simply `return nil` as you did, or return an error - a user's day-2 host installation will still hang - as you never modified the corresponding validation code that is waiting on those tang results from the agent

rccrdpccl:
/hold

Holding waiting for clarification on this comment:
```Also regardless of whether you simply return nil as you did, or return an error - a user's day-2 host installation will still hang - as you never modified the corresponding validation code that is waiting on those tang results from the agent```

Is this changing current behaviour, other than fixing the nil pointer case when luks not present?
Just want to make sure we do not add more issues

**Comment by paul-maidment** [comment]
> tang_connectivity_check_cmd

Regarding the validation of disk encryption requirements. 

It seems that firstly we fetch host ignition and validate that luks is not nil and contains at least one entry.
https://github.com/openshift/assisted-service/blob/master/internal/host/hostutil/host_utils.go#L326

https://github.com/openshift/assisted-service/blob/master/internal/host/validator.go#L422-L425
Then we check to see if the clevis setting is present, bypassing validation and marking it as successful if these settings are not present

```
if luks == nil || luks.Clevis == nil {
	// Disk encryption is disabled for workers on day1 cluster
	return ValidationSuccessSuppressOutput, ""
}
```
the comment `// Disk encryption is disabled for workers on day1 cluster` is a little confusing as a few lines above, we have already determined that we are looking at a day 2 host.

With this aside, we can see how a Day2 host without Clevis is treated. This is treated as though there is no disk encryption.
Which to me at least, looks like the same logic I implemented in this PR?

**Comment by omertuc** [comment]
>the comment // Disk encryption is disabled for workers on day1 cluster is a little confusing as a few lines above, we have already determined that we are looking at a day 2 host.

By "day 1 cluster" this comment actually means "target cluster"

**Comment by paul-maidment** [comment]
> > the comment // Disk encryption is disabled for workers on day1 cluster is a little confusing as a few lines above, we have already determined that we are looking at a day 2 host.
> 
> By "day 1 cluster" this comment actually means "target cluster"

Should we change the above comment to make this a lot clearer?

**Comment by omertuc** [comment]
>With this aside, we can see how a Day2 host without Clevis is treated. This is treated as though there is no disk encryption.
Which to me at least, looks like the same logic I implemented in this PR?

Yes, it's the same.

It's making the false statement that because clavis is nil then there's no encryption. Which is a wrong thing to say. But its behavior is still correct since its goal is to validate tang and clavis being nil does mean there's no tang to validate

**Comment by omertuc** [comment]
>Should we change the above comment to make this a lot clearer?

Sure

File: internal/host/hostcommands/tang_connectivity_check_cmd.go [hunk 3]
```diff
@@ -38,6 +38,15 @@ func (c *tangConnectivityCheckCmd) getTangServersFromHostIgnition(host *models.H
 		if err != nil {
 			return nil, err
 		}
+		if luks == nil {
+			return nil, nil
+		}
+		if luks.Clevis == nil {
+			// TODO: not sure how this could happen or whether its possibly valid, for now pretend encryption is disabled just so we avoid a nil pointer dereference
+			// Investigation of this will be carried out in https://issues.redhat.com/browse/MGMT-16721
+			c.log.Warn("luks clevis configuration is missing or incomplete in the host ignition, disk encryption will be assumed to be disabled. (MGMT-16721)")
```
omertuc, line 47: I believe after discussing it and understanding who's who and what are we even doing here and more importantly what we should be doing, we can rephrase this

```suggestion
// This cluster does not use tang, so we shouldn't perform a tang connectivity check
```

paul-maidment:
> /hold
> 
> Holding waiting for clarification on this comment: `Also regardless of whether you simply return nil as you did, or return an error - a user's day-2 host installation will still hang - as you never modified the corresponding validation code that is waiting on those tang results from the agent`
> 
> Is this changing current behavior, other than fixing the nil pointer case when luks not present? Just want to make sure we do not add more issues

In short...zero behavioral change. Our existing validations expect these fields to be set otherwise we assume that disk encryption is disabled. We still need an investigation into whether or not we should change some behavior but for now this will have no negative impact as far as I can tell.

internal/host/hostcommands/tang_connectivity_check_cmd.go
paul-maidment: and shall we discard the investigation, I think we know how things are used right?

omertuc: right

rccrdpccl:
Thanks for clarifying @paul-maidment 
/unhold
/lgtm
/approve

PR #5952, opened 2024-02-06
MGMT-16753: use ip=on to configure dual stack host

adriengentil:
`ip=<nic>:dhcp,dhcp6` is not valid according to [dracut documentation](https://man7.org/linux/man-pages/man7/dracut.cmdline.7.html), and it makes `systemd-network-generator` to complain with:
```
Feb 01 14:45:07 ai-saas-ci-1104-master-2 systemd-network-generator[1803]: Failed to parse kernel command line: Invalid argument
```

Fixing this issue by setting `ip=<nic>:on`.

### DIFF
--- a/internal/host/hostcommands/install_cmd.go
+++ b/internal/host/hostcommands/install_cmd.go
@@ -404,7 +404,7 @@ func getDHCPArgPerNIC(network *net.IPNet, nic *models.Interface, ipv6 bool, dual
 	}
 	if found {
 		if dualStack {
-			dhcp = "dhcp,dhcp6"
+			dhcp = "on"
 		}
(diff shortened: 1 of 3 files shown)


adriengentil:
> Have you considered the original motivation? ([ea56a99](https://github.com/openshift/assisted-service/commit/ea56a9934b3eb3264dd0bae85d70d01bf89f83f4))
> 
> Why is that no longer needed? Or did it never work in the first place (i.e. it's not just a new version issue)?

Thanks, I'll double check!

adriengentil:
/close
ip=dhcp,dhcp6 is support by `nm-initrd-generator` but it's not by `systemd-network-generator`

PR #6010, opened 2024-02-21
OCPBUGS-29772: All nodes fail to become managed when installConfigOverrides defined

tsorya:
OCPBUGS-29772: All nodes fail to become managed when installConfigOverrides defined
We are marshalling install config as yaml and this actually causing it to return with empty values, moving to json fixes the issue

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -67,7 +67,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
-	"gopkg.in/yaml.v2"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/validation"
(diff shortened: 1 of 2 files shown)


internal/bminventory/inventory.go
```diff
@@ -67,7 +67,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
-	"gopkg.in/yaml.v2"
```
zaneb, line 70: I've found it pays to always use `sigs.k8s.io/yaml` (which uses json under the hood) instead of `gopkg.in/yaml.v2`.

tsorya: On the other hand why not to use json as in all other places we just use it. Though agree that if already to use yaml better to use sigs.k8s.io/yaml in this case. 
Thanks

CrystalChun (approved):
/lgtm

PR #6025, opened 2024-02-26
MGMT-16263: Enable usage of custom OCP release images when we create cluster

danmanor:
## In General
This PR is part 2 of the Implementation of this enhancement proposal - https://github.com/openshift/assisted-service/blob/master/docs/enhancements/dynamic-fetching-of-openshift-versions.md, which is enabling usage of dynamic `OpenShift` releases. (part 1 - https://github.com/openshift/assisted-service/pull/5916)

### Main Flow
As mentioned in the description in part 1, the main changes are:

1. **In openshift-versions endpoint** - We retrieve the release images from the DB (from part (1), which contains static and dynamic release images), and in addition we filter by 2 new query parameters (only_latest, version_pattern) that will be used by the `UI` to present short list of the versions. We also filter by a new configuration parameter `IGNORED_OPENSHIFT_VERSIONS`, which contains a list of versions we don't want to retrieve (as one might have a blocker bug).

2. **We split the versions handler** to 2 different version handlers - one for `RestAPI` and one for `KubeAPI`. Each of the structs will implement version handling according to the running mode.

### Remarks

1. I decided to split the version handler into two separate implementations for `RestAPI` and `KubeAPI` because the changes necessitated distinct handling of versions in each mode.
2. This change will trigger release images syncer to start in `SaaS` .
3. The current process for subsystem testing begins with deploying `Wiremock` (to mock `OCM`), then launching the assisted service and other components, and finally starting the tests. Stubs are added to `Wiremock` after the test initiation, which poses a problem. This is because part 1 (see [part 1](https://github.com/openshift/assisted-service/pull/5916)) implements a utility that needs to interact with `OCM` immediately upon the service's startup. This necessitates having the stubs available in `Wiremock` from the moment the service begins (to retrieve releases). To solve this, I have implemented a Python script that adds the stubs to `Wiremock`, ensuring they are in place before the service starts.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


Dockerfile.assisted-service
```diff
@@ -37,6 +37,7 @@ RUN cd cmd && CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /build/assiste
 RUN cd ./cmd/operator && CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /build/assisted-service-operator
 RUN cd ./cmd/webadmission && CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /build/assisted-service-admission
 RUN cd ./cmd/agentbasedinstaller/client && CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /build/agent-installer-client
+RUN cd ./cmd/wiremock && CGO_ENABLED=1 GOFLAGS="" GO111MODULE=on go build -o /build/add-wiremock-stubs
```
danielerez, line 40: It's a bit non-conservative to include a binary for tests purposes in the production image. Not sure if we can easily avoid it though, but maybe just worth thinking about alternatives.

data/default_release_sources.json
```diff
@@ -0,0 +1,218 @@
+[
```
danielerez, line 1: I haven't followed the original change, but I just wonder why is this object separated from default_release_images.json? Is it just for simplicity, or is there some other consideration? Also, do we handle updating these values with the upgrade script (i.e. when a new OCP version is released, etc)?

internal/versions/rest_api_versions.go
```diff
@@ -0,0 +1,210 @@
...
+		return &releaseImage, err
+	}
+
+	// openshiftVersion must be major.minor, find the latest version matching
+
+	// In case openshiftVersion is in the format of major.minor-multi
+	newVersion = strings.TrimSuffix(openshiftVersion, "-multi")
+
+	// Escape SQL wildcards in case thay are approved as a version
+	newVersion = common.EscapePlaceHolders(newVersion)
```
danielerez, line 84: Instead of escaping the version and use sql, won't it be simpler to find relevant release images with something like gofunk? E.g. https://github.com/openshift/assisted-service/blob/cc56ded8ef0a9dc3f95e2c8a65d08e96546cd72b/internal/versions/versions.go#L175

danielerez (commented):
Partial review, add some comments for now.

Dockerfile.assisted-service
danmanor: I don't think it would be complicated, but I don't think it worth the maintaining of another image

cmd/main.go
```diff
@@ -273,6 +274,9 @@ func main() {
 			"Failed to parse RELEASE_SOURCES json %s", Options.ReleaseSourcesConfig.ReleaseSources)
 	}
 
+	var ignoredOpenshiftVersions = []string{}
+	parseIgnoredOpenshiftVersions(ignoredOpenshiftVersions, failOnError)
+
 	log.Println(fmt.Sprintf("Started service with OS images %v, Release images %v",
```
danielerez, line 280: Consider logging also the IgnoredOpenshiftVersions, could be useful for debugging,

internal/bminventory/inventory.go
```diff
@@ -466,10 +466,7 @@ func (b *bareMetalInventory) getNewClusterReleaseImage(ctx context.Context, para
 		return nil, errors.Wrapf(err, "Openshift version %s for CPU architecture %s is not supported",
 			swag.StringValue(params.OpenshiftVersion), arch)
 	}
-	if models.OpenshiftVersionSupportLevelMaintenance == releaseImage.SupportLevel {
-		return nil, errors.Errorf("Openshift version %s support level is: %s, and can't be used for creating a new cluster",
-			swag.StringValue(params.OpenshiftVersion), releaseImage.SupportLevel)
-	}
+
```
danielerez, line 469: We're no longer handling 'maintenance' support level?

danielerez (commented):
Looks good overall, added some comments.
Just a general comment, as discussed, please check disconnected scenarios (i.e. disable sync to db, etc).

internal/bminventory/inventory.go
danmanor: We do, but for some reason there was a limitation in the service code that prohibited using maintenance release images for installing cluster, I removed it because we want to support this option.

internal/versions/rest_api_versions.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// ValidateReleaseImageForRHCOS validates whether for a specified RHCOS version we have an OCP
+// version that can be used. This functions performs a very weak matching because RHCOS versions
+// are very loosely coupled with the OpenShift versions what allows for a variety of mix&match.
+func (h *restAPIVersionsHandler) ValidateReleaseImageForRHCOS(rhcosVersion, cpuArch string) error {
+	var releaseImages models.ReleaseImages
+
+	err := h.db.Model(&models.ReleaseImage{}).Find(&releaseImages).Error
+	if err != nil {
```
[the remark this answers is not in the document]
danmanor: In DB actions it is clearer for me to check the error in a different line to distinguish the action

internal/versions/rest_api_versions.go
```diff
@@ -0,0 +1,210 @@
...
+	return getLatestReleaseImage(releaseImages, h.ignoredOpenshiftVersions)
+}
+
+// GetReleaseImageByURL fetches a release image using the specified URL.
+// It returns an error if no matching image is found.
+func (h *restAPIVersionsHandler) GetReleaseImageByURL(_ context.Context, url, _ string) (*models.ReleaseImage, error) {
+	var releaseImage models.ReleaseImage
+
+	query := h.db.Session(&gorm.Session{})
+	err := query.Model(&models.ReleaseImage{}).Where("url = ?", url).Take(&releaseImage).Error
```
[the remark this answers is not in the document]
danmanor: I was thinking about that, but I think it I can't do much validation because urls can be in many forms (in the CI it is much different for example). I think if it will be empty it just won't find anything, do you think it worth checking it separately ?

danmanor:
@danielerez @gamli75 holding until the UI will merge the adjustments to this change

carbonin (changes requested):
Holding this to take a look at the wiremock situation.

I very much don't like the idea of including non-production code that can alter/intercept network communication in our production image.

carbonin:
Also just as a general note, splitting into separate commits is great, but try to make the commit messages usable.
All this prefixing is unnecessary:
image

Dockerfile.assisted-service
carbonin: We're already maintaining the image we're deploying for wiremock (or should be at least).
[This](https://github.com/openshift/assisted-service/blob/3d14207ff4155d3f982d06fd485c828b0f0594f4/deploy/wiremock/wiremock-deployment.yaml#L18) shows it's using `ocpmetal` repo which IIRC is something we used to maintain.

carbonin: FWIW I tried to get rid of all of these in b8aa0b95615847c42434c46b6a08678b3d5d4476 but I guess I missed some

carbonin: Also I'm not sure why converting to python is the only option.

In the past I was able to initialize the wiremock instance using a small go program. Could we not do the same as a part of deploying wiremock (before we start assisted-service)?

Example:
```
package main

import (
	"log"

	"github.com/kelseyhightower/envconfig"
	"github.com/openshift/assisted-service/subsystem"
)

var Options struct {
	OCMHost   string `envconfig:"OCM_HOST"`
	TestToken string `envconfig:"TEST_TOKEN"`
}

func main() {
	if err := envconfig.Process("subsystem", &Options); err != nil {
		log.Fatal(err.Error())
	}

	wiremock := &subsystem.WireMock{
		OCMHost:   Options.OCMHost,
		TestToken: Options.TestToken,
	}

	if err := wiremock.DeleteAllWiremockStubs(); err != nil {
		log.Fatalf("Failed to delete all wiremock stubs, %v", err)
	}

	if err := wiremock.CreateWiremockStubsForOCM(); err != nil {
		log.Fatalf("Failed to init wiremock stubs, %v", err)
	}
}
```

danmanor: @carbonin IIUC this is what I did exactly

danmanor:
@carbonin This code runs only when we run subsystem test though, then why can it interfere with the production flow ?

internal/versions/rest_api_versions_test.go
```diff
@@ -0,0 +1,786 @@
+package versions
```
carbonin, line 1: Can you ensure you retain the tests added in https://github.com/openshift/assisted-service/pull/6066/files#diff-83dbd9742dbbf03982097164da597f3c30dcda4ed0af066409af7f3ac7f67716R281-R330 to ensure we don't regress?

internal/versions/test_common.go
```diff
@@ -0,0 +1,91 @@
+package versions
```
carbonin, line 1: Should this be `common_test.go`? All test files should be `*_test.go`, right?

subsystem/agent_based_installer_client_test.go
```diff
@@ -26,7 +26,7 @@ var _ = Describe("RegisterClusterAndInfraEnv", func() {
 		Expect(registerClusterErr).NotTo(HaveOccurred())
 		Expect(network.GetApiVipById(&common.Cluster{Cluster: *modelCluster}, 0)).To(Equal("1.2.3.8"))
 		Expect(network.GetIngressVipById(&common.Cluster{Cluster: *modelCluster}, 0)).To(Equal("1.2.3.9"))
-		Expect(modelCluster.OpenshiftVersion).To(ContainSubstring(snoVersion))
+		Expect(modelCluster.OpenshiftVersion).To(ContainSubstring("4.15.0"))
```
carbonin, line 29: Why was this changed?

subsystem/operators_test.go
```diff
@@ -245,7 +246,7 @@ var _ = Describe("Operators endpoint tests", func() {
 				swag.String(models.ClusterCPUArchitectureS390x),
 				swag.Bool(false),
 			)
-			Expect(cluster.Payload.CPUArchitecture).To(Equal(models.ClusterCPUArchitectureMulti))
+			Expect(cluster.Payload.CPUArchitecture).To(Equal(models.ClusterCPUArchitectureS390x))
```
carbonin, line 249: Why change this?

swagger.yaml
```diff
@@ -3113,6 +3113,15 @@ paths:
         - userAuth: [admin, read-only-admin, user]
       operationId: v2ListSupportedOpenshiftVersions
       description: Retrieves the list of OpenShift supported versions.
+      parameters:
+        - in: query
+          name: version_pattern
+          description: Version pattern for the returned releases. E.g. '4.12'.
```
carbonin, line 3119: You would literally put `4.12` in here? Is that how that works?

Would it be better for this to be an enum with `X.Y` and `X.Y.Z` as members or are there more options?

tools/deploy_assisted_installer_configmap.py
```diff
@@ -107,8 +112,9 @@ def main():
     data = data.replace('REPLACE_HW_VALIDATOR_REQUIREMENTS', '"{}"'.format(deploy_options.hw_requirements))
     data = data.replace('REPLACE_DISABLED_HOST_VALIDATIONS', '"{}"'.format(deploy_options.disabled_host_validations))
     data = data.replace('REPLACE_DISABLED_STEPS', '"{}"'.format(deploy_options.disabled_steps))
-    data = data.replace('REPLACE_RELEASE_SOURCES', '"{}"'.format(RELEASE_SOURCES))
+    data = data.replace('REPLACE_RELEASE_SOURCES', "'{}'".format(RELEASE_SOURCES))
```
carbonin, line 115: Why change this? Every other field is the previous way.

carbonin (commented):
I would also like us to try to look more closely for solutions to tests that don't involve wiremock stuff in our prod image.

Dockerfile.assisted-service
carbonin: Why not do it from outside the assisted service image?

internal/versions/kube_api_versions_test.go
```diff
@@ -218,24 +226,6 @@ var _ = Describe("GetReleaseImage", func() {
 		Expect(*releaseImage.Version).Should(Equal("4.9-candidate"))
 	})
 
-	It("get from ReleaseImages", func() {
```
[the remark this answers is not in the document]
danmanor: I didn't understand what was this test trying to test, but I get it now. I will insert it back

internal/versions/kube_api_versions_test.go
```diff
@@ -447,31 +428,9 @@ var _ = Describe("ValidateReleaseImageForRHCOS", func() {
...
-		h, err := NewHandler(common.GetTestLog(), nil, models.ReleaseImages{}, nil, "", nil)
-		Expect(err).ShouldNot(HaveOccurred())
-
-		_, err = h.GetDefaultReleaseImage(common.TestDefaultConfig.CPUArchitecture)
-		Expect(err).Should(HaveOccurred())
-		Expect(err.Error()).Should(Equal("Default release image is not available"))
-	})
-})
-
 var _ = Describe("GetMustGatherImages", func() {
```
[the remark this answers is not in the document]
danmanor: I think we can be verbose in this one and keep it in separate tests as it uses `GetReleaseImage` for specific instance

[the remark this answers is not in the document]
carbonin: But the content is exactly the same.
I don't see value in having to change identical tests in two places when we make one code change unless there is significant differences between how the function is used, which is not the case here.

internal/versions/rest_api_versions.go
```diff
@@ -0,0 +1,245 @@
...
+	// it means that they should be the source for release images.
+	if h.releaseImages != nil {
+		for _, releaseImage := range h.releaseImages {
+			if *releaseImage.OpenshiftVersion == openshiftVersion &&
+				(funk.Contains(releaseImage.CPUArchitectures, cpuArchitecture) || *releaseImage.CPUArchitecture == cpuArchitecture) {
+				return releaseImage, nil
+			}
+		}
+
+		return nil, errors.Errorf("no release image found for openshiftVersion: '%s' and CPU architecture '%s'", openshiftVersion, cpuArchitecture)
```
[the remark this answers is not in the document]
danmanor: The thought process was as follows:
For KubeAPI mode, we want RELEASE_IMAGES to serve as cache.
For RestAPI mode, there are two scenarios:

1. In SaaS, we want to activate the release images gathering process, and use the DB as the source for release images.
2. In other scenarios (ABI for example), we don't want to activate the release images gathering, (by not supplying RELEASE_SOURCES, then we want to use RELEASE_IMAGES as the source for release images.

This is the way I implemented it. An alternative could be having the release images in the DB in rest mode anyway, and if the gathering is activated then add more releases, otherwise not. then you can only look at the DB for the source of release images in Rest mode, although I found this solution much bigger change then the current option.

internal/versions/rest_api_versions_test.go
danmanor: Yes I added in kubeAPI mode, in rest we don't use that code

subsystem/operators_test.go
danmanor: It needs to match the version inside `../docs/hive-integration/crds/clusterImageSet.yaml`, which I upgraded

swagger.yaml
danmanor: Currently it retrieves every release image such that `version_pattern` is substring of `ReleaseImage.Version`

subsystem/agent_based_installer_client_test.go
danmanor: It needs to match the version inside `../docs/hive-integration/crds/clusterImageSet.yaml`, which I upgraded

subsystem/operators_test.go
danmanor: We register a cluster with `s390x` arch, we should get this arch now

internal/versions/rest_api_versions.go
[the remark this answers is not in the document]
danmanor: In second thought, it might be better to go with the alternative and have one source of release images for rest API, I will try it out

internal/versions/kube_api_versions_test.go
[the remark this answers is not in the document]
danmanor: I tried to convert it, in my opinion it makes less sense in common as the test itself doesn't have much context but calling the function with different parameters, therefore it turns up as a bunch of restAPI calls (with restAPI.GetReleaseImage) and then a bunch of kubeAPI calls, same as if they were in separate files. In my opinion, it would make more sense to keep them as they are currently, but if you disagree I can merge the tests in common_test.go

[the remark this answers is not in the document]
carbonin: From meeting: You can "mock" the `GetReleaseImage` by passing in a function defined just for the test to return some test release image. Don't worry about the different real implementations of `GetReleaseImage`

swagger.yaml
carbonin: Okay if we want this then we should change the name.
I'd call it just `version` assuming it will return all the data for the versions matching whatever is sent.

subsystem/operators_test.go
carbonin: I'm not sure if this was being done intentionally. If the test wasn't failing before I'm worried it was behavior we were somehow expecting.

Dockerfile.assisted-service
carbonin: Lets try to just run the go code to initialize the mocks from the Makefile.
If that doesn't work we can explore other options.

cmd/main.go
```diff
@@ -434,6 +443,7 @@ func main() {
 	hostStateMonitor.Start()
 	defer hostStateMonitor.Stop()
 
+	addReleaseImagesToDBIfNeeded(db, releaseImagesArray, failOnError)
```
carbonin, line 446: Is this a problem that this runs on every pod (not just the leader)?
I assume we'll get duplicates or errors when we try to insert the same version from multiple replicas.

Maybe do this in the release syncer process?

cmd/main.go
```diff
@@ -951,3 +982,37 @@ func stopOpenshiftReleaseSyncerIfNeeded(openShiftReleaseSyncer *thread.Thread) {
 		openShiftReleaseSyncer.Stop()
 	}
 }
+
+func addReleaseImagesToDBIfNeeded(db *gorm.DB, releaseImages models.ReleaseImages, failOnError func(err error, msg string, args ...interface{})) {
+	if Options.EnableKubeAPI {
```
carbonin, line 987: I think we also need to check release sources here.
Because we want to just let the syncer thread add these if it's going to run rather than adding them here and then having the sync thread immediately delete and readd them.

deploy/wiremock/wiremock-deployment.yaml
```diff
@@ -16,6 +16,7 @@ spec:
       containers:
         - name: wiremock
           image: quay.io/ocpmetal/wiremock
```
carbonin, line 18: Just for reference, we should figure out how this was built and if we should update it or move it to `edge-infrastructure` I don't think we do anything with `ocpmetal` anymore.

cc @gamli75 what happened with that org? Do we even own it still?

docs/hive-integration/crds/clusterImageSet.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: hive.openshift.io/v1
 kind: ClusterImageSet
 metadata:
-  name: openshift-v4.9.0
+  name: openshift-v4.15.0
```
carbonin, line 4: Why update this now? Everything kubeapi related should remain the same, right?

internal/releasesources/clients.go
```diff
@@ -33,11 +33,11 @@ type openShiftSupportLevelAPIClientInterface interface {
 	getSupportLevels(majorVersion string) (ocpVersionSupportLevels, error)
 }
 
-type releaseGraph struct {
-	Nodes []node `json:"nodes"`
+type ReleaseGraph struct {
```
carbonin, line 36: If we are able to remove the wiremock stuff I don't think this should be exported.

internal/versions/api.go
```diff
@@ -73,25 +75,145 @@ func GetModelVersions(v Versions) models.Versions {
...
+
+	if *isPreRelease {
+		return swag.String(models.ReleaseImageSupportLevelBeta), nil
+	}
+
+	return swag.String(models.ReleaseImageSupportLevelProduction), nil
+}
+
 func (h *apiHandler) V2ListSupportedOpenshiftVersions(ctx context.Context, params operations.V2ListSupportedOpenshiftVersionsParams) middleware.Responder {
+	handler := h.versionsHandler.(*restAPIVersionsHandler)
```
carbonin, line 191: Instead of asserting the type here why not use the actual type in `apiHandler`?

internal/versions/rest_api_versions.go
```diff
@@ -0,0 +1,210 @@
...
+
+	query := h.db.Session(&gorm.Session{})
+	cleanQuery := h.db.Session(&gorm.Session{})
+	query = query.Where(&models.ReleaseImage{})
+	if cpuArchitecture == common.MultiCPUArchitecture {
+		query = query.Where("cardinality(cpu_architectures) > 1")
+	} else {
+		query = query.Where(cleanQuery.Where("? = ANY(cpu_architectures)", cpuArchitecture).Or("cpu_architecture = ?", cpuArchitecture))
+	}
+	newVersion := openshiftVersion
```
carbonin, line 55: Keep the variable in the scope where it's used.

internal/versions/rest_api_versions.go
```diff
@@ -0,0 +1,210 @@
...
+	newVersion = strings.TrimSuffix(openshiftVersion, "-multi")
+
+	var releaseImages models.ReleaseImages
+	err := query.Find(&releaseImages).Error
+	if err != nil {
+		return nil, err
+	}
+
+	releaseImages = funk.Filter(releaseImages, func(releaseImage *models.ReleaseImage) bool {
+		return strings.HasPrefix(*releaseImage.Version, fmt.Sprintf("%s.", newVersion))
```
carbonin, line 93: Can we do this in sql? If not this is fine.

internal/versions/rest_api_versions_test.go
```diff
@@ -0,0 +1,655 @@
...
+		}
+
+		err := db.Create(releaseImages).Error
+		Expect(err).ShouldNot(HaveOccurred())
+
+		handler.ignoredOpenshiftVersions = []string{"4.14.2"}
+		releaseImage, err := handler.GetReleaseImage(ctx, "4.14", common.X86CPUArchitecture, pullSecret)
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(*releaseImage.Version).To(Equal("4.14.1"))
+	})
```
carbonin, line 458: I would add a test case for when someone asks for an exact version and we don't have it.

subsystem/agent_based_installer_client_test.go
carbonin: I'm not even sure that needs to change.

deploy/wiremock/wiremock-deployment.yaml
gamli75: the org still exists but I agree that we should move it

internal/versions/api.go
danmanor: In second thought, I think it makes more sense to use Handler and assert, as it makes much more sense to test the api handler this way, because I can use a mock.

internal/versions/rest_api_versions.go
```diff
@@ -0,0 +1,210 @@
...
+	db                       *gorm.DB
+}
+
+// GetReleaseImage retrieves a release image based on a specified OpenShift version and CPU architecture.
+// If the provided OpenShift version includes a patch version, it attempts to retrieve an exact match for the release image if available.
+// For OpenShift versions specified as major.minor, it fetches the latest matching release image.
+// The function returns an error Returns an error for other formats of OpenShift version or if no matching image can be found.
+func (h *restAPIVersionsHandler) GetReleaseImage(_ context.Context, openshiftVersion, cpuArchitecture, _ string) (*models.ReleaseImage, error) {
+	// validations
+	if err := validateCPUArchitecture(cpuArchitecture); err != nil {
```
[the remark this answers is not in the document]
danmanor: Safer to validate as it goes into SQL queries don't you think ?

docs/hive-integration/crds/clusterImageSet.yaml
danmanor: OK I remember now, this is because of `agentBasedInstaller` test that uses it, which uses restAPI

internal/releasesources/clients.go
carbonin: Looks like it's required for the wiremock stuff in subsystem, this is fine then.

internal/versions/api.go
carbonin: Okay, but just in case we end up ever using these APIs in kube mode maybe check that the type assertion was successful here.

internal/versions/api.go
```diff
@@ -107,55 +229,52 @@ func (h *apiHandler) V2ListSupportedOpenshiftVersions(ctx context.Context, param
...
 			// JSON definition we do nothing, but in case it's missing there (because CVO is not
 			// reporting it), we add it ourselves.
-			displayName := *releaseImage.Version
-			if len(supportedArchs) > 1 && !strings.HasSuffix(displayName, "-multi") {
+			if len(releaseImage.CPUArchitectures) > 1 && !strings.HasSuffix(displayName, "-multi") {
 				displayName = displayName + "-multi"
 			}
 
-			openshiftVersion, exists := openshiftVersions[key]
+			openshiftVersion, exists := openshiftVersions[displayName]
```
carbonin, line 260: This changes the key from `releaseImage.OpenshiftVersion` to `releaseImage.Version`.
Was that intentional?

subsystem/authz_test.go
```diff
@@ -56,16 +56,24 @@ var _ = Describe("test authorization", func() {
...
-		capabilityReviewIgnoreValidationsNotallowedUserStubID, err = wiremock.createStubIgnoreValidationsCapabilityReview(fakePayloadUsername, OrgId1, false)
+		capabilityReviewIgnoreValidationsNotallowedUserStubID, err = wiremock.createStubIgnoreValidationsCapabilityReview(
+			fakePayloadUsername, OrgId1, false,
+		)
 		Expect(err).ShouldNot(HaveOccurred())
 
-		capabilityReviewIgnoreValidationsAllowedUserStubID, err = wiremock.createStubIgnoreValidationsCapabilityReview(fakePayloadUsername2, OrgId2, true)
+		capabilityReviewIgnoreValidationsAllowedUserStubID, err = wiremock.createStubIgnoreValidationsCapabilityReview(
+			fakePayloadUsername2, OrgId2, true,
+		)
```
carbonin, line 76: For the future just don't make changes like this unless you happen to be making an actual change to the code. It bloats the diff and wastes time for reviewers to look through irrelevant changes.

carbonin (commented):
Good work switching how we create the wiremock stubs :+1: 

Still looks like there's some work to do though between the last few comments and the failing tests.

carbonin:
/retest

Just to see if any of these were flakes.
Some are referencing OCP versions so that's a bit suspicious IMO.

internal/versions/api.go
danmanor: Yes, because `openshiftVersion` may be x.y.z or x.y while `Version` is always x.y.z, there are now two different versions with the same `openshiftVersion`

subsystem/authz_test.go
danmanor: Understood, sorry

subsystem/operators_test.go
```diff
@@ -402,29 +402,6 @@ var _ = Describe("Operators endpoint tests", func() {
 
 			Expect(operatorSubscriptionName).To(Equal(lvm.LvmsSubscriptionName))
 		})
-
-		It("should lvm have right subscription name on 4.11", func() {
-			cluster := registerNewCluster(
-				"4.11.0",
```
carbonin, line 408: Removing otherwise valid tests is not the correct way to deal with errors.
The error was that a release image wasn't present for 4.11.0 it feels like the correct solution would be to add the wiremock stubs to ensure a 4.11 image is present and/or update this version if necessary.

danmanor: I removed this test as 4.11 is `EOL` now and this test tests specific behavior of lvm in 4.11

danmanor: I believe this test is no longer relevant

carbonin: Let's keep it until we also remove the [related code](https://github.com/openshift/assisted-service/blob/7cac29ddf85d31912a091d6a9c4bd453ee6a0003/internal/operators/lvm/manifest.go#L40-L57)

danmanor: Yes you are right, I will change the version

danmanor: I just set it to 4.11, then it will take the latest 4.11 release image

internal/common/db.go
```diff
@@ -537,3 +538,10 @@ func CloseDB(db *gorm.DB) {
 	}
 	_ = sqlDB.Close()
 }
+
+func EscapePlaceHolders(message string) string {
```
carbonin, line 542: Can this be removed? I don't see it being used anywhere and it doesn't have tests.

internal/versions/rest_api_versions.go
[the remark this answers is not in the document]
carbonin: Fair enough

internal/common/db.go
danmanor: Yes, I will remove it

carbonin (approved):

danmanor:
@carbonin Thank you for your time and effort

PR #6037, opened 2024-02-28
MGMT-17080: enable infrastructure operator

rccrdpccl:
When MCE and a storage operator (LVM or ODF) are selected, generate manifest for `AgentServiceConfig`

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


rccrdpccl:
/hold
WIP

internal/operators/mce/manifest.go
```diff
@@ -34,6 +45,45 @@ func Manifests() (openshiftManifests map[string][]byte, customManifests []byte,
...
+	}
+	databasePVC, err := getPvcClaim(databaseStorageSize)
+	if err != nil {
+		return []byte{}, err
+	}
+	imagePVC, err := getPvcClaim(imageStorageSize)
+	if err != nil {
+		return []byte{}, err
+	}
+	agentServiceConfig := &aiv1beta1.AgentServiceConfig{
```
rccrdpccl, line 77: Will add more default values here

internal/operators/mce/manifest.go
```diff
@@ -34,6 +45,45 @@ func Manifests() (openshiftManifests map[string][]byte, customManifests []byte,
 	return openshiftManifests, mceManifest, nil
 }
 
+func getPvcClaim(storageSize string) (corev1.PersistentVolumeClaimSpec, error) {
```
rccrdpccl, line 48: Rename method to make sense

internal/operators/mce/manifest.go
```diff
@@ -3,6 +3,17 @@ package mce
...
 	"text/template"
+
+	aiv1beta1 "github.com/openshift/assisted-service/api/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	filesystemStorageSize = "20Gi"
```
rccrdpccl, line 14: Make sure sizes are matching storage requirements

internal/operators/mce/manifest.go
```diff
@@ -3,6 +3,17 @@ package mce
...
+	databaseStorageSize   = "20Gi"
+	imageStorageSize      = "20Gi"
```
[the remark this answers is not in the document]
carbonin: Related to this, should we be adding some kind of validation that the user has the required storage?

internal/operators/lvm/config.go
```diff
@@ -1,40 +1,17 @@
...
 
 type Config struct {
-	LvmCPUPerHost                 int64  `envconfig:"LVM_CPU_PER_HOST" default:"1"`
-	LvmMemoryPerHostMiB           int64  `envconfig:"LVM_MEMORY_PER_HOST_MIB" default:"400"`
-	LvmMemoryPerHostMiBBefore4_13 int64  `envconfig:"LVM_MEMORY_PER_HOST_MIB" default:"1200"`
-	LvmMinOpenshiftVersion        string `envconfig:"LVM_MIN_OPENSHIFT_VERSION" default:"4.11.0"`
-}
-
-// count all disks of drive type ssd or hdd
-func (o *operator) getValidDiskCount(disks []*models.Disk, installationDiskID string) int64 {
```
rccrdpccl, line 31: Although I dislike common, we've moved this function to common, as it's common to odf operator.
Long term I think it would be best to have `<package>utils` packages for this type of things (this probably would be `inventoryutils`)

internal/operators/manager.go
```diff
@@ -24,7 +26,7 @@ import (
 	"github.com/thoas/go-funk"
 )
 
-const customManifestFile = "custom_manifests.json"
+const controllerManifestFile = "custom_manifests.json"
```
rccrdpccl, line 29: Renamed "custom manifest" to controller manifest, as it was confusing to have 2 customs manifests.
They are now "controller manifest" and "install manifest"

rccrdpccl:
/cc @omertuc
PTAL

internal/cluster/cluster.go
```diff
@@ -688,12 +688,7 @@ func CanDownloadKubeconfig(c *common.Cluster) (err error) {
 }
 
 func (m *Manager) IsOperatorMonitored(c *common.Cluster, operatorName string) bool {
```
omertuc, line 690: Can we remove this function?

internal/operators/lvm/config.go
```diff
@@ -1,40 +1,17 @@
...
-	// LvmsMemoryRequirement           int64 = 400
-	// LvmsMemoryRequirementBefore4_13 int64 = 1200
-
 )
 
 type Config struct {
-	LvmCPUPerHost                 int64  `envconfig:"LVM_CPU_PER_HOST" default:"1"`
-	LvmMemoryPerHostMiB           int64  `envconfig:"LVM_MEMORY_PER_HOST_MIB" default:"400"`
-	LvmMemoryPerHostMiBBefore4_13 int64  `envconfig:"LVM_MEMORY_PER_HOST_MIB" default:"1200"`
-	LvmMinOpenshiftVersion        string `envconfig:"LVM_MIN_OPENSHIFT_VERSION" default:"4.11.0"`
```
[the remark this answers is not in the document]
omertuc: Note that this is not a simple constant, it's an envconfig. 

However [I couldn't find](https://github.com/search?type=code&q=%2FLVM_MIN_OPENSHIFT_VERSION%2F) anyone publicly using it, so it's probably fine to remove it. Not everything has to be an adjustable knob

internal/operators/mce/config.go
```diff
@@ -15,3 +20,28 @@ const (
...
+	return maxValue
+}
+
+func sum(storageSize map[string]int64) int64 {
+	var total int64
+	for _, v := range storageSize {
+		total += v
+	}
+	return total
+}
```
omertuc, line 47: Instead of max/sum we can use go-funk or [lo](https://github.com/samber/lo)

rccrdpccl: amazing, we probably should replace funk with lo, I don't really like having them both around.
Or maybe just wait to upgrade to go 1.22 where most uses will be covered by stdlib

omertuc: Which part of go 1.22 will cover most uses?

internal/operators/manager.go
```diff
@@ -120,10 +122,36 @@ func (mgr *Manager) GetRequirementsBreakdownForHostInCluster(ctx context.Context
...
+
+func getStorageClassName(cluster *models.Cluster) string {
+	if common.HasOperator(cluster, odf.Operator.Name) {
+		return odf.GetStorageClassName()
+	}
+	if common.HasOperator(cluster, lvm.Operator.Name) {
+		return lvm.GetStorageClassName()
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
rccrdpccl: I've tried a different approach now, but we still have a lengthy function to retrieve the storage operator.
The issue is that it could be the case that both are selected (unlikely but it's not forbidden), and we need the manager to pick one.
WDYT?

internal/operators/lvm/lvm_operator.go
```diff
@@ -21,6 +22,8 @@ type operator struct {
 	extracter oc.Extracter
 }
 
+const defaultStorageClassName = "lvms-vg1"
```
rccrdpccl, line 25: @omertuc do you know if there is a better way to have the default storage class name? It depends on the LVMcluster manifest specs, as it stands it generates as above, but IMO it's very fragile.

rccrdpccl: same is true, if not worse, for odf

omertuc:
Sorry, can't get around to having a look.

internal/common/common.go
```diff
@@ -143,6 +144,32 @@ func NormalizeCPUArchitecture(arch string) string {
...
+}
+
+func HasOperator(c *models.Cluster, operatorName string) bool {
+	for _, o := range c.MonitoredOperators {
+		if o.Name == operatorName {
+			return true
+		}
+	}
+	return false
+}
```
avishayt, line 171: Maybe it should be in internal/cluster/common.go?

internal/common/common.go
```diff
@@ -143,6 +144,32 @@ func NormalizeCPUArchitecture(arch string) string {
...
+		if (disk.DriveType == models.DriveTypeSSD || disk.DriveType == models.DriveTypeHDD) && installationDiskID != disk.ID && disk.SizeBytes != 0 {
+			if disk.SizeBytes >= conversions.GbToBytes(minSize) {
+				eligibleDisks++
+			} else {
+				availableDisks++
+			}
+		}
+	}
+	return eligibleDisks, availableDisks
+}
```
avishayt, line 162: Maybe it should be in internal/operators/common.go?

rccrdpccl: Cannot move it there because of circular dependency (assuming package would be `operators`)
To avoid this problem, I would suggest using `<package>utils` pattern for stateless functions that might be used outside the package. For example, I would move this function in `inventoryutils` package - as all needed data is inventory-related, and `HasOperator` function in `clusterutils` - as all necessary data is cluster-related. WDYT?

internal/common/common.go
rccrdpccl: this can cause circular dependencies, please see [the other related comment](https://github.com/openshift/assisted-service/pull/6037#discussion_r1523797814)

internal/common/common.go
avishayt: These methods are used in a single package, so they shouldn't be in common/common.go, but rather in the package they are used.

internal/operators/odf/validations.go
```diff
@@ -101,7 +100,7 @@ func (o *operator) computeResourcesAllNodes(cluster *models.Cluster, odfClusterR
 	return status, nil
 }
 
-func (o *operator) computeNodeResourceUtil(host *models.Host, odfClusterResources *odfClusterResourcesInfo, mode odfDeploymentMode) (string, error) {
+func (o *operator) computeNodeResourceUtil(cluster *models.Cluster, host *models.Host, odfClusterResources *odfClusterResourcesInfo, mode odfDeploymentMode) (string, error) {
```
CrystalChun, line 103: Is `mode` no longer being used in this function?

internal/operators/odf/odf_operator.go
```diff
@@ -101,6 +105,31 @@ func getODFDeploymentMode(numOfHosts int) odfDeploymentMode {
...
+	return defaultStorageClassName
+}
+
+func (o *operator) getMinDiskSizeGB(cluster *models.Cluster) int64 {
+	minDiskSizeGB := o.config.ODFMinDiskSizeGB
+	if common.HasOperator(cluster, mce.Operator.Name) {
+		minDiskSizeGB += mce.GetMinDiskSizeGB(cluster)
+	}
+	return minDiskSizeGB
+}
```
CrystalChun, line 118: This function seems common to both storage operators 🤔 could we potentially have this function defined once and use it in both?

internal/operators/odf/validations.go
rccrdpccl: Mode was used just for the error message for ODF, so it's now handled in a wrapper for odf only.

internal/operators/manager.go
```diff
@@ -120,10 +122,35 @@ func (mgr *Manager) GetRequirementsBreakdownForHostInCluster(ctx context.Context
 	return requirements, nil
 }
 
+func (mgr *Manager) getStorageOperator(cluster *models.Cluster) (api.StorageOperator, error) {
+	priorityOrder := []string{odf.Operator.Name, lvm.Operator.Name}
```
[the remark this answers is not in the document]
rccrdpccl: This is just selecting storage operator to work with MCE, to select the right storage class. If both ODF and LVM are selected, the storage class will be from ODF

internal/operators/lvm/lvm_operator.go
```diff
@@ -101,7 +108,12 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
 		return api.ValidationResult{Status: api.Failure, ValidationId: o.GetHostValidationID(), Reasons: []string{message}}, err
 	}
 
-	diskCount := o.getValidDiskCount(inventory.Disks, host.InstallationDiskID)
+	minSize := int64(0)
```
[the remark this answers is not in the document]
rccrdpccl: It seemed to be 0 before this PR so I've kept it that way

internal/common/common.go
```diff
@@ -143,6 +144,32 @@ func NormalizeCPUArchitecture(arch string) string {
 	}
 }
 
+// Return available disks and available disks that meet the minSize requirement
```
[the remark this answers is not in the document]
rccrdpccl: I meant "available" and "available that meet the minSize", looks like I need to rephrase it :)

internal/common/common.go
rccrdpccl: They are not a single package: "operators/lvm" and "operators/odf" are two separate packages. This said, I agree it shouldn't be in "common", but I'll move it to "operators/common" as moving to "operators" will cause circular dependency.

internal/operators/manager.go
```diff
@@ -120,10 +122,35 @@ func (mgr *Manager) GetRequirementsBreakdownForHostInCluster(ctx context.Context
...
+					return storageOperator, nil
+				}
+			}
+		}
+	}
+	return nil, fmt.Errorf("no storage operator found")
+}
+
+func hasMCEWithStorage(cluster *models.Cluster) bool {
+	return common.HasOperator(cluster, mce.Operator.Name) &&
```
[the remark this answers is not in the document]
rccrdpccl: I didn't see it :facepalm: 
But it's duplicated code, let's unify it

internal/operators/odf/odf_operator.go
rccrdpccl: Good point, I've also extracted mce operator dependency and done this operation from the manager

internal/operators/odf/validations.go
CrystalChun: Are we able to remove it from this function then?

internal/bminventory/inventory_test.go
```diff
@@ -18892,3 +18882,28 @@ var _ = Describe("V2UpdateHostIgnition unbound blabla", func() {
...
+}
+
+func createClusterWithMonitoredOperator(db *gorm.DB, operator models.MonitoredOperator) common.Cluster {
+	c := getDummyCluster()
+	c.MonitoredOperators = append(c.MonitoredOperators, &operator)
+	Expect(db.Create(&c).Error).ShouldNot(HaveOccurred())
+	return c
+}
+
+func createClusterWithoutMonitoredOperator(db *gorm.DB) common.Cluster {
```
CrystalChun, line 18905: Nit: Isn't this just a createCluster function?
If so, can we just reuse this one https://github.com/openshift/assisted-service/blob/9f84d8570761e0b4f4fe3b6bfdafbfabccb62243/internal/bminventory/inventory_test.go#L390

internal/operators/manager.go
```diff
@@ -203,6 +245,14 @@ func (mgr *Manager) AnyOLMOperatorEnabled(cluster *common.Cluster) bool {
 func (mgr *Manager) ValidateHost(ctx context.Context, cluster *common.Cluster, host *models.Host) ([]api.ValidationResult, error) {
 	results := make([]api.ValidationResult, 0, len(mgr.olmOperators))
 
+	if hasMCEWithStorage(cluster.Cluster.MonitoredOperators) {
+		storageOperator, err := mgr.getStorageOperator(&cluster.Cluster)
```
CrystalChun, line 249: Sorry about silly questions: 
What if both ODF and LVM is enabled? Would it matter if only ODF has the minimum disk requirement? Is this because if both are enabled, we're definitely expected to use ODF?

rccrdpccl: Not a silly question, infact you have just spotted a bug :)
Both enabled and one not fulfilling the minimum disk requirements would block the installation, however we can only bump disk requirements on the one that it's going to be used (this is the bug, at the moment we are bumping both storage operator's disk requirements if MCE is selected).
The decision of picking ODF over LVM is arbitrary, but I think we should define it to avoid undefined behaviour.

internal/bminventory/inventory_test.go
rccrdpccl: Nice, well spotted. I'll move it and reuse the function you mention

internal/operators/odf/validations.go
rccrdpccl: oops sorry, I've misread where you where commenting, I was referring to another function where mode was use.
You're right, is not used anymore, we can remove it and simplify this function

internal/operators/manager.go
rccrdpccl: Nevermind, I was wrong, we do bump only the correct operator's disk requirements

CrystalChun: Ah that makes sense! Thank you for the explanation!!

> The decision of picking ODF over LVM is arbitrary, but I think we should define it to avoid undefined behaviour.
Agreed! Maybe doc it in code and in the user-facing docs? I think we have a doc folder for operators

internal/bminventory/inventory_test.go
```diff
@@ -18892,3 +18882,22 @@ var _ = Describe("V2UpdateHostIgnition unbound blabla", func() {
...
+			ID:            &clusterID,
+			Name:          "my-cluster",
+			BaseDNSDomain: "my-domain",
+		},
+	}
+	return c
+}
+
+func createClusterWithMonitoredOperator(db *gorm.DB, operator models.MonitoredOperator) *common.Cluster {
+	c := getDummyCluster()
```
CrystalChun, line 18899: Would we be able to use `createCluster()` here too?

CrystalChun: Oh gotcha!

docs/user-guide/additional-operator-notes.md
```diff
@@ -4,3 +4,9 @@
 - When deploying CNV on Single Node OpenShift (SNO), [hostpath-provisioner](https://github.com/kubevirt/hostpath-provisioner) (part of the CNV product) storage is automatically opted in and set up to use, to enable persisting VM disks.  
 This is done with the thought in mind that most virtualization use cases require persistence.  
 The hostpath-provisioner is set up to utilize an LSO PV as the backing storage for provisioning dynamic hostPath volumes on.
+
+## Multi-Cluster Engine (MCE)
+
+- When deploying MCE together with a storage operator (ODF or LVM) Infrastructure Operator will be automatically
+enabled. This will require extra disk space availability requirements for the storage operator.
```
CrystalChun, line 11: Maybe a future note of just how much extra disk space availability is needed. Maybe we can just point to ACM docs of disk space required for MCE?

docs/user-guide/additional-operator-notes.md
```diff
@@ -4,3 +4,9 @@
...
+When selecting MCE with both ODF and LVM, ODF will have priority and its storage class will be selected to provision Infrastructure Operator's PVCs.
```
CrystalChun, line 12: Nice!

internal/operators/common/common.go
```diff
@@ -0,0 +1,32 @@
+package common
```
CrystalChun, line 1: Maybe a little too nit: but not sure if the name `common` is ok since the `internal/common` pkg exists already 😅 @avishayt @carbonin wdyt?

internal/operators/lvm/lvm_operator.go
```diff
@@ -57,6 +61,18 @@ func (o *operator) GetFullName() string {
 	return "Logical Volume Management"
 }
 
+func (o *operator) GetStorageClassName() string {
```
CrystalChun, line 64: Small nit: Getter functions don't need the word "Get" in for golang https://go.dev/doc/effective_go#Getters

internal/operators/common/common_test.go
```diff
@@ -0,0 +1,61 @@
...
+	func(disks []*models.Disk, diskID string, minSize, expectedEligibleDisks, expectedAvailableDisks int64) {
+		eligibleDisks, availableDisks := common.GetValidDiskCount(disks, diskID, minSize)
+		Expect(eligibleDisks).To(Equal(expectedEligibleDisks))
+		Expect(availableDisks).To(Equal(expectedAvailableDisks))
+	},
+	Entry("no disk provided", []*models.Disk{}, "", int64(0), int64(0), int64(0)),
+	Entry("no valid disk provided", []*models.Disk{
+		{SizeBytes: 20 * conversions.GB, DriveType: models.DriveTypeUnknown, ID: "/dev/disk/by-id/disk-1"},
+		{SizeBytes: 20 * conversions.GB, DriveType: models.DriveTypeVirtual, ID: "/dev/disk/by-id/disk-2"},
+	}, "", int64(0), int64(0), int64(0)),
```
carbonin, line 27: Why are these all `int64`? How many disks are you expecting?

internal/operators/lvm/lvm_operator.go
carbonin: Is this not something we can set when deploying the operators?

I don't know much about them but it seems like that should be a feature :shrug:

internal/operators/manager.go
```diff
@@ -120,10 +123,35 @@ func (mgr *Manager) GetRequirementsBreakdownForHostInCluster(ctx context.Context
...
+				if ok {
+					return storageOperator, nil
+				}
+			}
+		}
+	}
+	return nil, fmt.Errorf("no storage operator found")
+}
+
+func hasMCEWithStorage(operators []*models.MonitoredOperator) bool {
```
carbonin, line 145: Super minor, but maybe `hasMCEAndStorage`?

Currently it sounds like you're searching for some property of MCE saying it's configured with storage already.

internal/operators/manager_test.go
```diff
@@ -27,6 +32,59 @@ import (
...
+		expectedName,
+		expectedDecodedBodyRegexp,
+	}
+}
+
+func decodeInterface(x interface{}) ([]map[string]string, error) {
+	data := []map[string]string{}
+	jsonContentBytes, ok := x.([]byte)
+	if !ok {
+		return data, errors.New("wtf")
```
carbonin, line 46: While I understand this is very unlikely to happen perhaps create a more informative error string :sweat_smile:

internal/operators/mce/manifest.go
```diff
@@ -34,6 +48,24 @@ func Manifests() (openshiftManifests map[string][]byte, customManifests []byte,
...
+	}
+	return ""
+}
+
+func GetAgentServiceConfigWithPVCManifest(storageClassName string) ([]byte, error) {
+	vars := map[string]string{
+		"DATABASE_STORAGE_SIZE":   getStorageSize(databaseStorage),
+		"FILESYSTEM_STORAGE_SIZE": getStorageSize(filesystemStorage),
+		"IMAGE_STORAGE_SIZE":      getStorageSize(imageStorage),
+		"STORAGE_CLASS_NAME":      storageClassName,
```
carbonin, line 63: Does creating the storage operator create a default storage class in the cluster?
I assume it would make things easier if you didn't have to provide this, right?

internal/operators/mce/mce_operator_test.go
```diff
@@ -93,3 +93,26 @@ var _ = Describe("MCE Operator", func() {
 		)
 	})
 })
+
+var _ = Describe("Get right min disk size", func() {
```
carbonin, line 97:
```suggestion
var _ = Describe("GetMinDiskSizeGB", func() {
```

internal/operators/mce/mce_operator_test.go
```diff
@@ -93,3 +93,26 @@ var _ = Describe("MCE Operator", func() {
...
+		})
+	})
+	When("cluster is HA", func() {
+		cluster := &models.Cluster{HighAvailabilityMode: &fullHaMode}
+		It("should return the maximum value of any required PVCs", func() {
+			minDiskSize := GetMinDiskSizeGB(cluster)
+			expectedMinDiskSize := int64(50)
+			Expect(expectedMinDiskSize).To(Equal(minDiskSize))
+		})
+	})
```
carbonin, line 117: This all feels like it could be simplified, maybe something like this?
```suggestion
	It("should return the sum of all required PVCs for a SNO cluster", func() {
	        snoMode := models.ClusterHighAvailabilityModeNone
		minDiskSize := GetMinDiskSizeGB(&models.Cluster{HighAvailabilityMode: &snoMode})
		expectedMinDiskSize := int64(70)
		Expect(expectedMinDiskSize).To(Equal(minDiskSize))
	})
	It("should return the maximum value of any required PVCs for a HA cluster", func() {
	        fullHaMode := models.ClusterHighAvailabilityModeFull
		minDiskSize := GetMinDiskSizeGB(&models.Cluster{HighAvailabilityMode: &fullHaMode})
		expectedMinDiskSize := int64(50)
		Expect(expectedMinDiskSize).To(Equal(minDiskSize))
	})
```

internal/operators/odf/validations.go
```diff
@@ -78,20 +77,13 @@ func (o *operator) computeResourcesAllNodes(cluster *models.Cluster, odfClusterR
...
-				}
-			}
-		} else {
-			status, err = o.computeNodeResourceUtil(host, odfClusterResources, compactMode)
+		if role == models.HostRoleAutoAssign && !isCompactMode {
+			status = "For ODF Standard Mode, all host roles must be assigned to master or worker."
+			err = errors.New("Role is set to auto-assign for host ")
+			return status, err
+		}
+		if role == models.HostRoleWorker || isCompactMode {
```
carbonin, line 85: I think there's a difference between using the cluster to determine if we're standard or compact vs using the number of hosts.

It could be that this causes a difference in behavior if a cluster is set to HA, but we don't actually have the correct number of hosts yet, right?

carbonin: Oh was this only previously used for logging? Maybe it's not an issue, but worth confirming.

carbonin (commented):
Looking good, mostly minor comments

internal/operators/common/common.go
rccrdpccl: I totally agree that common is a crappy name, I've maintained it for consistency as it is used throughout the code.
Adding this to `operators` package would've caused circular dependency as `operators` package is requiring `operators/odf` and `operators/odf` would require this function(s).

As I discussed in another comment, I think the best approach to this would be using `<package>utils` pattern (see for example std lib `ioutils`). This way we can segregate in a package all related utilities and minimize circular dependency issues.
For example, all functions (not methods) related cluster object could go in `internal/clusterutils`. This would reduce dependencies to a minimum and can be used from the whole codebase without any issue.

internal/common/common_test.go
```diff
@@ -1,4 +1,4 @@
-package common
+package common_test
```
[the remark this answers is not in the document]
rccrdpccl: I think when possible it's best to change package name for testing, it has at least two major benefits:
* reduce dependencies: tests often need more dependencies
* only test public interface: it's good practice to only test public interfaces, so that we can change internal implementation as long as the public interface keeps the same behaviour

In this case it was a leftover because I moved away from common a function that we tested here (IIRC it was HasOperator) which would've caused circular dependencies. I could've reverted the package renaming but I have decided to leave the renamed package as IMO it only brings advantages. WDYT?

internal/operators/manager_test.go
rccrdpccl: Ooops I didn't notice I've left this in :sweat_smile:

internal/operators/mce/manifest.go
rccrdpccl: It would definitely make it easier, but unfortunately it doesn't create a default storage class :(

internal/operators/mce/mce_operator_test.go
rccrdpccl: Yes, I have the habit to make it human readable: "When cluster is SNO" (or maybe "context cluster is SNO" "it should return..."
I can simplify it as suggested though, no big difference

internal/operators/odf/validations.go
rccrdpccl: To be honest I've just refactored the code to a simpler non-duplicated one and the conditions should be the same. Nevertheless if we spot a bug I can take advantage of this PR to fix it

internal/operators/lvm/lvm_operator.go
rccrdpccl: Unfortunately I am not aware of any mechanism that would allow me to do that....
The manifest for `AgentServiceConfig` is being sent to the controller, together with `LVMCluster` manifest and similar resources for ODF operator. These latter two are the one responsible to eventually creating the storage classes, and we can't really know the storage class beforehand, it's worked out from the parameters we set.

It's a shitty solution (hardcoded values for storage class) but I haven't found a better way - WDYT?

internal/operators/common/common_test.go
rccrdpccl: You're right, I've just blindly kept the type that was already there but I guess we'll never have more than `int`

docs/user-guide/additional-operator-notes.md
rccrdpccl: I've left them out on purpose because the docs will diverge with the real number (I'd expect the exact space requirement could change), but I can add them if we think it's valuable information

internal/operators/common/common_test.go
rccrdpccl: Actually it's a bit of a mess, swagger sets default type for models to int64, we could change it to int within business logic and then convert it back to int64 when we're dealing with the generated models but it would be a bigger change - happy to do it even if it'd be a bit out of scope - WDYT?

docs/user-guide/additional-operator-notes.md
CrystalChun: That makes sense. I was just imagining customers asking what disk size to use if the validation fails 😆

rccrdpccl: Good point, ODF validation is already stating that, but I can definitely add it to LVM error message

internal/common/common_test.go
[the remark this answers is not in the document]
carbonin: I understand and agree with you, but it will probably be a bit confusing given the vast majority of tests are in the same package in this repo. If you don't need to change it I'd leave it in `common`.

internal/operators/common/common_test.go
carbonin: Ah alright if it's coming from swagger it's fine.

internal/operators/odf/validations.go
carbonin: I think the confusing thing is what these conditions mean.

Maybe we can create small functions that call out what we're actually doing then this could look like:
`if hostWillRunODF(host, mode) {`

```
func hostWillRunODF(...) bool {
  return host.role == models.HostRoleWorker || mode == CompactMode
}
```

internal/operators/lvm/lvm_operator.go
```diff
@@ -57,6 +61,18 @@ func (o *operator) GetFullName() string {
...
 
+func (o *operator) StorageClassName() string {
+	return defaultStorageClassName
+}
+
+func (o *operator) SetAdditionalDiskRequirements(additionalSizeGB int64) {
+	o.additionalDiskRequirementGB = additionalSizeGB
+}
+
+func (o *operator) getMinDiskSizeGB() int64 {
```
carbonin, line 72: I don't think a non-exported getter should ever really be needed.

internal/operators/manager.go
```diff
@@ -120,10 +123,35 @@ func (mgr *Manager) GetRequirementsBreakdownForHostInCluster(ctx context.Context
...
 
+func (mgr *Manager) getStorageOperator(cluster *models.Cluster) (api.StorageOperator, error) {
+	priorityOrder := []string{odf.Operator.Name, lvm.Operator.Name}
+	for _, operatorName := range priorityOrder {
+		for _, operator := range cluster.MonitoredOperators {
+			if operator.Name == operatorName {
+				o := mgr.olmOperators[operatorName]
+				storageOperator, ok := o.(api.StorageOperator)
+				if !ok {
+					mgr.log.Errorf("defined storage operator %s does not implement StorageOperator interface", operatorName)
```
carbonin, line 134: I don't think we should log here.
You'll definitely have monitored operators that are not storage operators, right?

internal/operators/mce/config.go
```diff
@@ -15,3 +21,10 @@ const (
 type Config struct {
 	MceMinOpenshiftVersion string `envconfig:"MCE_MIN_OPENSHIFT_VERSION" default:"4.10.0"`
 }
+
+func GetMinDiskSizeGB(cluster *models.Cluster) int64 {
```
rccrdpccl, line 25: move this to mce_operator.go

internal/operators/mce/manifest.go
```diff
@@ -2,9 +2,22 @@ package mce
 
 import (
 	"bytes"
+	"fmt"
 	"text/template"
 )
 
+var storageSizeGi = map[string]int64{
+	"filesystem": int64(10),
```
carbonin, line 10: You can probably use the const here.

PR #6047, opened 2024-03-04
MGMT-16814: Pass both IP family URLs to ironic agent

carbonin:
When deploying a single stack spoke cluster from a dual-stack hub it's hard for us to determine what callback URLs we should send to the ironic agent. If we get the choice wrong the agent simply never registers and the ironic agent gets stuck trying to send data to ironic on the hub.

As of https://issues.redhat.com/browse/OCPBUGS-24579 the ironic agent can accept a comma-separated list of URLs and will select the correct one based on the host's networking situation at runtime.

This PR provides that comma-separated list of URLs to the ironic agent if the agent is of a version that includes the fix (currently OCP 4.16+).

As a side effect we also need to track the version of the ironic agent we're embedding. This is so that we don't provide a list of URLs to an agent that is only expecting a single one. This is simple enough for now, but may become more complicated as the fix is backported (we may need to track multiple releases for each Y-stream depending on how far back the fix goes). This also means that we need users to provide the version of overrides and the env-configurable defaults.

https://issues.redhat.com/browse/MGMT-16814

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

I don't have the environment to test this, but QE is going to run my image through a test.

## Checklist

- [ ] This change does not require a documentation update (docstring, `docs`, README, etc) - **docs updated**

## Reviewers Checklist

### DIFF
--- a/docs/hive-integration/README.md
+++ b/docs/hive-integration/README.md
@@ -212,8 +212,8 @@ The converged flow is enabled by default, you can disable the converged flow by 
 The ironic agent image will be determined based on the hub release image set in the ClusterVersion resource.
 If the hub cluster architecture does not match that of the InfraEnv or if the release image cannot be retrieved for some reason a default ironic agent image will be used.
 The defaults are:
-- x86_64: `quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:d3f1d4d3cd5fbcf1b9249dd71d01be4b901d337fdc5f8f66569eb71df4d9d446`
-- arm64: `quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:cb0edf19fffc17f542a7efae76939b1e9757dc75782d4727fb0aa77ed5809b43`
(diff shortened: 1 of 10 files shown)


carbonin:
I'm a bit concerned about maintaining this ...

If, for example, the ironic fix gets backported to both 4.15 and 4.14 we'd need to maintain 2 different versions for this mode to be available.

I wonder if it would be acceptable to just fix the existing case that should have handled the dual-stack hub + ipv6-only spoke (I'm only assuming there is a bug there because there is no must-gather).
My guess is that since we're using the machine networks to determine if the cluster is dual-stack we're running into an issue because machine networks wouldn't be set before nodes are discovered with baremetal spokes, right?

@eranco74 @avishayt what do you think?

carbonin:
/hold

See previous comment for reasoning.

carbonin:
Closing this in favor of https://github.com/openshift/assisted-service/pull/6048.

PR #6054, opened 2024-03-05
OCPBUGS-30232: Handle skipping hostPrefix validation for IPv6

bfournie:
For non-OVN/SDN networkTypes, the hostPrefix validation is not required and it is skipped. This fixes a regression introduced in the fix for https://issues.redhat.com/browse/OCPBUGS-23069 in which IPv6 CIDRs were not using the correct default hostPrefix. In addition, all cases where the validation is used are now covered.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/validator.go
+++ b/internal/cluster/validator.go
@@ -25,10 +25,12 @@ import (
 type ValidationStatus string
 
 const (
-	ValidationSuccess ValidationStatus = "success"
-	ValidationFailure ValidationStatus = "failure"
-	ValidationPending ValidationStatus = "pending"
-	ValidationError   ValidationStatus = "error"
+	ValidationSuccess     ValidationStatus = "success"
+	ValidationFailure     ValidationStatus = "failure"
+	ValidationPending     ValidationStatus = "pending"
+	ValidationError       ValidationStatus = "error"
+	DefaultIPV4HostPrefix                  = 25
+	DefaultIPV6HostPrefix                  = 64
 )
 
 func (v ValidationStatus) String() string {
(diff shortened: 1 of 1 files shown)


internal/cluster/validator.go
```diff
@@ -600,3 +601,16 @@ func (v *clusterValidator) skipNetworkHostPrefixCheck(c *clusterPreprocessContex
 	}
 	return false
 }
+
+// getHostPrefix returns the default host prefix to use.
+func getHostPrefix(skipHostPrefix bool, clusterNetwork *models.ClusterNetwork) int {
+	hostPrefix := int(clusterNetwork.HostPrefix)
+	if skipHostPrefix {
+		if network.IsIPv6CIDR(string(clusterNetwork.Cidr)) {
+			hostPrefix = DefaultIPV6HostPrefix
```
ori-amizur, line 610: If the host prefix needs to be skipped, why does it have a value?
If it has a value then it is not skipped.

internal/cluster/validator.go
```diff
@@ -551,10 +548,14 @@ func (v *clusterValidator) networkPrefixValid(c *clusterPreprocessContext) (Vali
 
 	var err error
 	for _, clusterNetwork := range c.cluster.ClusterNetworks {
-		if err = network.VerifyNetworkHostPrefix(clusterNetwork.HostPrefix); err != nil {
-			return ValidationFailure, fmt.Sprintf("Invalid Cluster Network prefix: %s.", err.Error())
+		if !skipHostPrefix {
+			if err = network.VerifyNetworkHostPrefix(clusterNetwork.HostPrefix); err != nil {
```
ori-amizur, line 552: If the value was set, then why is the value not verified?

internal/cluster/validator.go
bfournie: The name of this function is confusing and I'll change it. The prefix is only needed for CIDR validation, if the prefix check is skipped we still need to use a default value for CIDR validation check.

internal/cluster/validator.go
```diff
@@ -600,3 +601,16 @@ func (v *clusterValidator) skipNetworkHostPrefixCheck(c *clusterPreprocessContex
 	}
 	return false
 }
+
+// getHostPrefixForCidrValidation returns the prefix to use when validating CIDR.
+func getHostPrefixForCidrValidation(skipHostPrefix bool, clusterNetwork *models.ClusterNetwork) int {
+	hostPrefix := int(clusterNetwork.HostPrefix)
+	if skipHostPrefix {
+		if network.IsIPv6CIDR(string(clusterNetwork.Cidr)) {
+			hostPrefix = DefaultIPV6HostPrefix
```
ori-amizur, line 610: Does `clusterNetwork.HostPrefix` have a meaning for your case? 
Is there any reason not to set it somewhere (not necessarily here) to the default-values, instead of overriding them just for validation?

bfournie: The hostPrefix isn't necessary in the case of a user defining a networkType other than OVN/SDN. The check is in here since it only affects validation, the networkType can be retrieved, and the CIDRvalidation still needs to run using a default prefix. That said, I'm fine with moving the setting of the hostprefix elsewhere but would like to handle that in a separate PR. This PR is only to fix the problem with the IPv6 prefix which is preventing customers from using networkTypes other than OVN/SDN in a IPv6 or dual-stack setup.

internal/cluster/validator.go
```diff
@@ -531,17 +533,12 @@ func (v *clusterValidator) networkPrefixValid(c *clusterPreprocessContext) (Vali
 		return ValidationPending, "The Cluster Network CIDR is undefined."
 	}
 
-	skipHostPrefix := v.skipNetworkHostPrefixCheck(c)
+	skipHostPrefixCheck := v.skipNetworkHostPrefixCheck(c)
 
 	validClusterNetworks := funk.Filter(c.cluster.ClusterNetworks, func(clusterNetwork *models.ClusterNetwork) bool {
-		var hostPrefixForCidrValidation int
```
ori-amizur, line 537: It seems to me that the code part from the `Filter` function to the code fragment
```
        if len(validClusterNetworks) == len(c.cluster.ClusterNetworks) {
		return ValidationSuccess, "The Cluster Network prefix is valid."
	}
```
Can be deleted with any harm.  In this case the function should return `ValidationSuccess` at the end.

PR #6062, opened 2024-03-06
AGENT-860: Update configure command to handle multiple HostConfigs

rwsu:
for day-2/addnodes workflow.

Updated the LoadHostConfigs function to only load the current host's HostConfig if the workflow is addnodes.

Only the current host's HostConfig is needed because each host in the addnodes workflow has its own assisted-service. Loading other node's HostConfigs would cause the apply-host-config service to continually wait for other hosts to register with assisted-service because each node is added in isolation.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
- [x] Manual (Elaborate on how it was tested) - Tested using ABI with the addnodes workflow currently being developed.
- [] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/agentbasedinstaller/client/main.go
+++ b/cmd/agentbasedinstaller/client/main.go
@@ -57,8 +57,9 @@ var RegisterOptions struct {
 }
 
 var ConfigureOptions struct {
-	InfraEnvID    string `envconfig:"INFRA_ENV_ID" default:""`
-	HostConfigDir string `envconfig:"HOST_CONFIG_DIR" default:"/etc/assisted/hostconfig"`
+	InfraEnvID    string                                `envconfig:"INFRA_ENV_ID" default:""`
+	HostConfigDir string                                `envconfig:"HOST_CONFIG_DIR" default:"/etc/assisted/hostconfig"`
+	WorkflowType  agentbasedinstaller.AgentWorkflowType `envconfig:"WORKFLOW_TYPE" default:"install"`
 }
 
 var ImportOptions struct {
@@ -219,7 +220,7 @@ func configure(ctx context.Context, log *log.Logger, bmInventory *client.Assiste
 		log.Fatal("No INFRA_ENV_ID specified")
 	}
 
-	hostConfigs, err := agentbasedinstaller.LoadHostConfigs(ConfigureOptions.HostConfigDir)
+	hostConfigs, err := agentbasedinstaller.LoadHostConfigs(ConfigureOptions.HostConfigDir, ConfigureOptions.WorkflowType)
 	if err != nil {
 		log.Fatal("Failed to load host configuration: ", err)
 	}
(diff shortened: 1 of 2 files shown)


cmd/agentbasedinstaller/client/main.go
```diff
@@ -59,6 +59,7 @@ var RegisterOptions struct {
 var ConfigureOptions struct {
 	InfraEnvID    string `envconfig:"INFRA_ENV_ID" default:""`
 	HostConfigDir string `envconfig:"HOST_CONFIG_DIR" default:"/etc/assisted/hostconfig"`
+	WorkflowType  string `envconfig:"WORKFLOW_TYPE" default:""`
```
andfasano, line 62: default could be "install"

cmd/agentbasedinstaller/host_config.go
```diff
@@ -192,10 +193,28 @@ func LoadHostConfigs(hostConfigDir string) (HostConfigs, error) {
 			}
 		}
 
-		configs = append(configs, &hostConfig{
-			configDir:    hostPath,
-			macAddresses: addresses,
-		})
+		addHostConfig := false
+		if workflowType == "addnodes" {
```
andfasano, line 197: For the sake of clarity, it could be worth redefining the workflow enum here, instead of using plain strings

cmd/agentbasedinstaller/host_config.go
```diff
@@ -192,10 +193,28 @@ func LoadHostConfigs(hostConfigDir string) (HostConfigs, error) {
...
+			// if multiple day-2 nodes are being added using the same day-2 ISO.
+			// Filter otu the other HostConfig entries because each day-2
+			// node is added in isolation using their own internal assisted-service
+			// instance.
+			addHostConfig, err = currentHostHasMACAddress(addresses)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			addHostConfig = true
```
andfasano, line 209: nit: to simplify a little bit the loop and drop the bool var, if workflowType is "addnodes" and the current entry does not match the current host, a `continue` before adding the `hostConfig` to `configs` should be enough

cmd/agentbasedinstaller/host_config.go
```diff
@@ -206,6 +225,28 @@ type hostConfig struct {
...
+	if err != nil {
+		return currentHostHasMatchingMACAddress, fmt.Errorf("failed to get this host's interfaces: %w", err)
+	}
+	for _, iface := range hostInterfaces {
+		if iface.HardwareAddr == nil {
+			continue
+		}
+		for _, hostConfigMac := range addresses {
+			if iface.HardwareAddr.String() == hostConfigMac {
+				currentHostHasMatchingMACAddress = true
```
andfasano, line 242: Could also have an immediate return here, no need to continue looping I guess

PR #6066, opened 2024-03-08
MGMT-16813: Prioritize finding release image with exact CPU architecture

CrystalChun:
https://issues.redhat.com/browse/MGMT-16813
Previously, the `getReleaseImageFromCache` would get all release
images that included the CPU architecture and return the first
instance it found that matched the desired Openshift version.

Since multi-arch images include the CPU architecture, it led
to undesireable selection of the multi-arch image over the
exact matching image with the specified CPU architecture.

This fix ensures that the release image will be found with
the following priority:
1. Exact CPU Architecture & OpenShift x.y.z version match
2. Exact CPU Architecture & OpenShift Major Minor (x.y) version match
3. Multi-arch CPU with specified CPU Architecture & OpenShift x.y.z version match
4. Multi-arch CPU with specified CPU Architecture & OpenShift Major Minor (x.y) version match

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

**note**: will push up code gen once it finishes

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -166,50 +166,80 @@ func (h *handler) GetReleaseImageByURL(ctx context.Context, url, pullSecret stri
 	return h.addReleaseImage(url, pullSecret)
 }
 
+// Finds a release image from the cache that matches the specified version and architecture in the following
+// priority:
+// 1. Exact CPU Architecture & OpenShift x.y.z version match
+// 2. Exact CPU Architecture & OpenShift Major Minor (x.y) version match
+// 3. Multi-arch CPU with specified CPU Architecture & OpenShift x.y.z version match
+// 4. Multi-arch CPU with specified CPU Architecture & OpenShift Major Minor (x.y) version match
 func (h *handler) getReleaseImageFromCache(openshiftVersion, cpuArchitecture string) (*models.ReleaseImage, error) {
 	if cpuArchitecture == "" {
 		// Empty implies default CPU architecture
 		cpuArchitecture = common.DefaultCPUArchitecture
 	}
+
 	// Filter Release images by specified CPU architecture.
-	releaseImages := funk.Filter(h.releaseImages, func(releaseImage *models.ReleaseImage) bool {
-		for _, arch := range releaseImage.CPUArchitectures {
-			if arch == cpuArchitecture {
-				return true
-			}
-		}
+	exactCPUArchReleaseImages := funk.Filter(h.releaseImages, func(releaseImage *models.ReleaseImage) bool {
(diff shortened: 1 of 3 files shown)


carbonin (changes requested):
I think the problem is actually [here](https://github.com/openshift/assisted-service/blob/eaab3736514e20a320733252f787a6f9b581179d/internal/versions/versions.go#L175-L182) in `getReleaseImageFromCache`.

It looks like it's considering the multi arch release images valid for when we've created a cluster with a specific architecture.

Really it should be preferring the exact matching arch while also allowing multi-arch if no exact match by architecture exists.

I think we should try to agree on some priority system for this function, put it in a comment, then try to fix the implementation.

CrystalChun:
/retitle MGMT-16813: Prioritize finding release image with exact CPU architecture

internal/versions/versions.go
```diff
@@ -166,50 +166,80 @@ func (h *handler) GetReleaseImageByURL(ctx context.Context, url, pullSecret stri
...
-		return swag.StringValue(releaseImage.CPUArchitecture) == cpuArchitecture
+		return false
 	})
-	if funk.IsEmpty(releaseImages) {
+
+	if funk.IsEmpty(exactCPUArchReleaseImages) && funk.IsEmpty(multiArchReleaseImages) {
 		return nil, errors.Errorf("The requested CPU architecture (%s) isn't specified in release images list", cpuArchitecture)
 	}
+
+	majorMinorVersion, err := toMajorMinor(openshiftVersion)
```
adriengentil, line 202: `majorMinorVersion` looks to be used only in `getReleaseImageByVersion` function maybe we can move this statement there?

internal/versions/versions_test.go
```diff
@@ -271,6 +271,56 @@ var _ = Describe("GetReleaseImage", func() {
 		Expect(*releaseImage.Version).Should(Equal("4.12.999-rc.4"))
 	})
 
+	It("returns the matching CPU architecture over multi-arch if it is present", func() {
```
adriengentil, line 274: would it make sense to test the x.y version matching cases as well? Or is it already tested?

internal/versions/versions.go
CrystalChun: Modified as suggested!

adriengentil (approved):

PR #6071, opened 2024-03-11
OCPBUGS-29975: Allow multiple machine networks

zaneb:
Until now, assisted-service has assumed that there would be either exactly one MachineNetwork specified for single-stack clusters, or for dual-stack clusters that there would be exactly one IPv4 and one IPv6 MachineNetwork specified (in that order).

None of these restrictions exist in OpenShift itself, which allows multiple MachineNetworks of each address family. This is necessary to support remote worker nodes on day 1, as well as distributed installations of a kind that are common in UPI deployments (where an external load balancer can easily balance across hosts in separate networks). ([OCPBUGS-29975](https://issues.redhat.com/browse/OCPBUGS-29975))

This change removes the assumptions about a single MachineNetwork per address family for clusters with UserManagedNetworking enabled, and reverts the change in #4867 that prevents users from specifying more networks at the API level.

(Clusters _without_ UserManagedNetworking will still use Layer 2 reachability checks in the `belongs-to-majority-group` host validation, which effectively prevents using remote worker nodes when creating these clusters. See [OCPBUGS-30730](https://issues.redhat.com/browse/OCPBUGS-30730).)

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] Agent-based installer
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


zaneb:
Infra issues with the CentOS repos
/retest

internal/bminventory/inventory.go
```diff
@@ -2145,45 +2145,32 @@ func (b *bareMetalInventory) updateNonDhcpNetworkParams(cluster *common.Cluster,
...
-				if err = network.VerifyVips(cluster.Hosts, secondaryMachineNetworkCidr, network.GetApiVipById(&targetConfiguration, 1), network.GetIngressVipById(&targetConfiguration, 1), log); err != nil {
+				if err = network.VerifyVips(cluster.Hosts, targetConfiguration.MachineNetworks, network.GetApiVipById(&targetConfiguration, 1), network.GetIngressVipById(&targetConfiguration, 1), log); err != nil {
 					log.WithError(err).Warnf("Verify VIPs")
 					return common.NewApiError(http.StatusBadRequest, err)
 				}
 			}
 
 		} else {
-			primaryMachineNetworkCidr, err = network.CalculateMachineNetworkCIDR(network.GetApiVipById(&targetConfiguration, 0), network.GetIngressVipById(&targetConfiguration, 0), cluster.Hosts, matchRequired)
+			primaryMachineNetworkCidr, err := network.CalculateMachineNetworkCIDR(network.GetApiVipById(&targetConfiguration, 0), network.GetIngressVipById(&targetConfiguration, 0), cluster.Hosts, matchRequired)
```
ori-amizur, line 2173: This might be wrong according the concept presented by this PR, because we may have to machine-network - one per vip and one of both vips.

internal/cluster/validator.go
```diff
@@ -458,13 +437,8 @@ func (v *clusterValidator) noCidrsOverlapping(c *clusterPreprocessContext) (Vali
...
 	}
-	if len(c.cluster.ClusterNetworks) != len(c.cluster.ServiceNetworks) {
-		// TODO MGMT-7587: Support any number of subnets
-		// Assumes that the number of cluster networks equal to the number of service networks
-		return ValidationError, "A mismatch between the number of Cluster and Service networks"
-	}
-	if err := checkCidrsOverlapping(c.cluster); err != nil {
-		return ValidationFailure, fmt.Sprintf("CIDRS Overlapping: %s.", err.Error())
+	if err := network.VerifyNoNetworkCidrOverlaps(c.cluster.ClusterNetworks, c.cluster.MachineNetworks, c.cluster.ServiceNetworks); err != nil {
+		return ValidationFailure, err.Error()
```
ori-amizur, line 441: The validation error messages need to be agreed upon.  They are presented by UI.  Here it is not clear what is expected as text of this error message

internal/host/validator.go
```diff
@@ -600,7 +600,7 @@ func (v *validator) belongsToMachineCidr(c *validationContext) (ValidationStatus
 	if c.inventory == nil || !network.IsMachineCidrAvailable(c.cluster) {
 		return ValidationPending, "Missing inventory or machine network CIDR"
 	}
-	if !network.IsHostInPrimaryMachineNetCidr(v.log, c.cluster, c.host) {
+	if !network.IsHostInMachineNetCidrs(v.log, c.cluster, c.host) {
 		return ValidationFailure, "Host does not belong to machine network CIDRs. Verify that the host belongs to every CIDR listed under machine networks"
```
ori-amizur, line 604: This error message is not true now.  Right ?

internal/network/machine_network_cidr.go
```diff
@@ -115,12 +115,25 @@ func VerifyVipFree(hosts []*models.Host, vip string, machineNetworkCidr string,
...
+			break
+		}
+	}
+
+	if !validMachineNetwork {
 		return models.VipVerificationUnverified, errors.Errorf("%s <%s> cannot be set if Machine Network CIDR is empty", vipName, vip)
 	}
-	if !ipInCidr(vip, machineNetworkCidr) {
-		return models.VipVerificationFailed, errors.Errorf("%s <%s> does not belong to machine-network-cidr <%s>", vipName, vip, machineNetworkCidr)
+	if machineNetworkCidr == "" {
```
ori-amizur, line 135: This test is irrelevant. The machine network cannot be valid and empty.

ori-amizur (commented):
It is not clear what are the restrictions here:
Do we allow multiple API VIPs per address family?
Do we have restriction per host-role to belong to a specific machine-network?  If yes, this needs to be added as validation.
What if we have only 3 host cluster (only masters). Can such cluster have multiple IPv4 machine-networks?
Can a machine-network be stale (without hosts)?

internal/cluster/validator.go
zaneb: Ack. I did change them, you can see the change in the tests here: https://github.com/openshift/assisted-service/pull/6071/commits/51e76b10d543c5868dcb250061c8b21250e45015#diff-854f83029709261bb4e532a5a6839b51ddeb2beb2f6e921056016a74340d06a3

internal/host/validator.go
zaneb: Ah yep, last part is not.

internal/network/machine_network_cidr.go
zaneb: Yes it can - validMachineNetwork will be true if any machine networks are defined, but machineNetworkCidr will only be set if the VIP is in one of them.

zaneb:
> Do we allow multiple API VIPs per address family?

No.

> Do we have restriction per host-role to belong to a specific machine-network? If yes, this needs to be added as validation.

No.

> What if we have only 3 host cluster (only masters). Can such cluster have multiple IPv4 machine-networks?

Yes (and in fact this is likely to be a common case for `platform: none`).

> Can a machine-network be stale (without hosts)?

I don't see why not.

internal/host/validator.go
zaneb: Wording is awkward here, because we still have to belong to both IPv4 and IPv6 CIDRs if both appear in the list. What do you think of "Verify that the host belongs to a CIDR for every IP version listed under machine networks"?

ori-amizur: That might be OK.  @avishayt what do you think?

avishayt: "Verify that the host belongs to a machine network CIDR for every used IP version"?

ori-amizur:
> > Do we have restriction per host-role to belong to a specific machine-network? If yes, this needs to be added as validation.
> 
> No.

So workers do not have to belong to machine-networks of ingress-vips? It will cause a mess.  The logic is already complicated and if we don't set rules, users will be confused.

> 
> > What if we have only 3 host cluster (only masters). Can such cluster have multiple IPv4 machine-networks?
> 
> Yes (and in fact this is likely to be a common case for `platform: none`).

None platform does not have machine-networks at all (At least how we implemented it).
Anyway if we have multiple machine-networks for masters, how can the API VIPs move between these hosts?  

> 
> > Can a machine-network be stale (without hosts)?
> 
> I don't see why not.

Why is it needed?

internal/network/machine_network_cidr.go
zaneb: This is maintaining the previous behaviour. If there are no machineNetworks specified, we return `VipVerificationUnverified`. If there are machineNetworks specified and the VIP isn't in any of them, we return `VipVerificationFailed`.

I think what's confusing in the diff is that `machineNetworkCidr` was previously a parameter, and now it's a local variable.

zaneb:
@ori-amizur I added a check in the VerifyVIPs that the machine network the API VIP is in is present on all master nodes, and the machine network the Ingress VIP is in is present on all worker nodes.
Note that for non-UserManagedNetworking, the l2_connectivity check is still going to require all the nodes to be on the same L2 network segment, effectively. There is a separate bug [OCPBUGS-30730](https://issues.redhat.com/browse/OCPBUGS-30730) for handling that, but this is a prerequisite for it.
For UserManagedNetworking clusters, hosts need only have l3_connectivity, so this change will allow them to pass multiple machine networks and have hosts connected to disjoint L2 domains, as long as they are routable. Since these clusters have no VIPs, this is what we want.

Note that in ABI we _always_ pass the MachineNetworks provided in the install-config. Also, for dual-stack the user must always pass the MachineNetworks explicitly to assisted-service.

internal/host/validator.go
zaneb: Changed to "Verify that the host belongs to a machine network CIDR listed for each IP stack in use"

internal/bminventory/inventory.go
zaneb: I added a comment indicating that this is something we'll want to fix as part of [OCPBUGS-30730](https://issues.redhat.com/browse/OCPBUGS-30730).

zaneb:
Tests are working now, so I believe this is ready for another round of review.

internal/network/cidr_validations.go
```diff
@@ -119,33 +120,59 @@ func VerifyClusterCidrSize(hostNetworkPrefix int, clusterNetworkCIDR string, num
...
-	if machineNetworkRequired {
-		err := VerifyNetworksNotOverlap(machineNetworkCidr, serviceNetworkCidr)
-		if err != nil {
-			return errors.Wrap(err, "MachineNetworkCIDR and ServiceNetworkCIDR")
+func VerifyNoNetworkCidrOverlaps(
+	clusterNetworks []*models.ClusterNetwork,
+	machineNetworks []*models.MachineNetwork,
+	serviceNetworks []*models.ServiceNetwork) error {
+	errs := []error{}
+	for imn, mn := range machineNetworks {
```
paul-maidment, line 134: High time complexity here O(m*n)
There might be something clever we could do not reduce this. 

Different language I know, but there are approaches using a data structure to memorize already searched ranges and to allow iteration across a set of networks in Order(n)

Note the use of a Radix tree in this implementation.

https://github.com/fgiuba/ipconflict/blob/master/ipconflict/subnet.py#L46-L61

Due to the limited number of networks, the consequences of not handling this may not be problematic. But it does look like something could be improved here!

paul-maidment (changes requested):

obockows:
This PR is really important - it enables multiple machineNetworks in agent-based installations (OCPBUGS-29975). Review comments from @paul-maidment are very helpful, but it looks like no one followed up.

Any ETA for  finalizing the suggested improvements?
Otherwise this will go stale and the fix will be lost.

PR #6094, opened 2024-03-18
OCPBUGS-30233: Filter IPs in majority group validation

CrystalChun:
https://issues.redhat.com/browse/OCPBUGS-30233
Filter the IP addresses when creating connectivity groups to
hosts that belong within the machine network CIDRs
(if they exist) to prevent failing "belongs-to-majority"
validation. Previously, if the IPs were not filtered,
the validation would fail if hosts had IPs on different
NICs that couldn't connect to other hosts. These IPs
were not used and caused the "belongs-to-majority"
validation to fail.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] Reviewer's test appreciated: My test environment isn't capable of installing more than a SNO 
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @zaneb

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -1400,7 +1400,7 @@ func (m *Manager) setConnectivityMajorityGroupsForClusterInternal(cluster *commo
 	}
 
 	for _, family := range []network.AddressFamily{network.IPv4, network.IPv6} {
-		majorityGroup, err := network.CreateL3MajorityGroup(hosts, family)
+		majorityGroup, err := network.CreateL3MajorityGroup(hosts, family, network.GetMachineNetworkCidrs(cluster))
 		if err != nil {
 			m.log.WithError(err).Warnf("Create L3 majority group for cluster %s failed", cluster.ID.String())
 		} else {
--- a/internal/network/connectivity_groups.go
+++ b/internal/network/connectivity_groups.go
@@ -436,7 +436,7 @@ func (l *l3Query) next() strfmt.UUID {
 				foundAddresses[l3.RemoteIPAddress] = true
 			}
 		}
-		if len(addresses) == len(foundAddresses) {
+		if len(foundAddresses) > 0 && len(addresses) == len(foundAddresses) {
 			return rh.HostID
 		}
 	}
@@ -458,7 +458,7 @@ func (l *l3QueryFactory) create(h *models.Host) (hostQuery, error) {
 	return &ret, nil
 }
 
-func newL3QueryFactory(hosts []*models.Host, family AddressFamily) (hostQueryFactory, error) {
+func newL3QueryFactory(hosts []*models.Host, family AddressFamily, machineNetworkCidrs []string) (hostQueryFactory, error) {
 	nodesAddresses := make(map[strfmt.UUID]map[string]bool)
 	for _, h := range hosts {
 		if h.Inventory == "" {
@@ -482,7 +482,10 @@ func newL3QueryFactory(hosts []*models.Host, family AddressFamily) (hostQueryFac
 				if err != nil {
 					return nil, err
 				}
-				value[ip.String()] = true
+
+				if ipInCidr := IpInCidrs(ip.String(), machineNetworkCidrs); len(machineNetworkCidrs) == 0 || ipInCidr {
+					value[ip.String()] = true
+				}
 			}
 		}
 		nodesAddresses[*h.ID] = value
@@ -552,7 +555,7 @@ func calculateMajorityGroup(hosts []*models.Host, factory hostQueryFactory) ([]s
 }
 
 /*
- * Crate majority for a cidr.  A majority group is a the largest group of hosts in a cluster that all of them have full mesh
+ * Create majority for a cidr.  A majority group is a the largest group of hosts in a cluster that all of them have full mesh
  * to the other group members.
  * It is done by taking a sorted connectivity group list according to the group size, and from this group take the
  * largest one
@@ -566,16 +569,16 @@ func CreateL2MajorityGroup(cidr string, hosts []*models.Host) ([]strfmt.UUID, er
 }
 
 /*
- * Crate majority for address family.  A majority group is a the largest group of hosts in a cluster that all of them have full mesh
+ * Create majority for address family.  A majority group is the largest group of hosts in a cluster that all of them have full mesh
  * to the other group members.
  * It is done by taking a sorted connectivity group list according to the group size, and from this group take the
  * largest one
  */
-func CreateL3MajorityGroup(hosts []*models.Host, family AddressFamily) ([]strfmt.UUID, error) {
+func CreateL3MajorityGroup(hosts []*models.Host, family AddressFamily, machineCidrs []string) ([]strfmt.UUID, error) {
 	if !funk.Contains([]AddressFamily{IPv4, IPv6}, family) {
 		return nil, errors.Errorf("Unexpected address family %+v", family)
 	}
-	factory, err := newL3QueryFactory(hosts, family)
+	factory, err := newL3QueryFactory(hosts, family, machineCidrs)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/network/connectivity_groups_test.go
+++ b/internal/network/connectivity_groups_test.go
@@ -530,6 +530,8 @@ func GenerateL3ConnectivityGroupTests(ipV4 bool, net1CIDR, net2CIDR string) {
 	} else {
 		family = IPv6
 	}
+	cidrs := []string{net1CIDR, net2CIDR}
+
 	Describe(fmt.Sprintf("connectivity groups %s", ipVersion), func() {
 		BeforeEach(func() {
 			if ipV4 {
@@ -560,7 +562,7 @@ func GenerateL3ConnectivityGroupTests(ipV4 bool, net1CIDR, net2CIDR string) {
 						Inventory:    makeInventory(nodes[2]),
 					},
 				}
-				ret, err := CreateL3MajorityGroup(hosts, family)
+				ret, err := CreateL3MajorityGroup(hosts, family, cidrs)
 				Expect(err).ToNot(HaveOccurred())
 				Expect(ret).To(Equal([]strfmt.UUID{}))
 			})
@@ -579,7 +581,7 @@ func GenerateL3ConnectivityGroupTests(ipV4 bool, net1CIDR, net2CIDR string) {
 						Inventory: makeInventory(nodes[2]),
 					},
 				}
-				ret, err := CreateL3MajorityGroup(hosts, family)
+				ret, err := CreateL3MajorityGroup(hosts, family, cidrs)
 				Expect(err).ToNot(HaveOccurred())
 				Expect(ret).To(Equal([]strfmt.UUID{}))
 			})
@@ -604,7 +606,7 @@ func GenerateL3ConnectivityGroupTests(ipV4 bool, net1CIDR, net2CIDR string) {
 					Inventory:    makeInventory(nodes[2]),
 				},
 			}
-			ret, err := CreateL3MajorityGroup(hosts, family)
+			ret, err := CreateL3MajorityGroup(hosts, family, cidrs)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(ret).To(Equal([]strfmt.UUID{}))
 		})
@@ -632,7 +634,7 @@ func GenerateL3ConnectivityGroupTests(ipV4 bool, net1CIDR, net2CIDR string) {
 					Inventory: makeInventory(nodes[2]),
 				},
 			}
-			ret, err := CreateL3MajorityGroup(hosts, family)
+			ret, err := CreateL3MajorityGroup(hosts, family, cidrs)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(ret).To(HaveLen(0))
 		})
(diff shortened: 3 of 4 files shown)


CrystalChun:
@zaneb would this change cover the issue? If not, may I request your advice?

internal/cluster/cluster.go
```diff
@@ -1399,8 +1399,12 @@ func (m *Manager) setConnectivityMajorityGroupsForClusterInternal(cluster *commo
 		majorityGroups[cidr] = majorityGroup
 	}
 
+	primaryMachineCIDR := ""
+	if network.IsMachineCidrAvailable(cluster) {
+		primaryMachineCIDR = network.GetMachineCidrById(cluster, 0)
+	}
```
zaneb, line 1405: Is there any reason to limit to a single MachineNetwork CIDR here?
This already will not work for dual stack clusters, because CreateL3MajorityGroup() is called for both IPv4 and IPv6, but this only passes the IPv4 MachineNetwork.
#6071 will allow multiple machine networks of each address family as well.

I would prefer to pass all of the MachineNetworks and check if each IP is in any of them.

internal/cluster/cluster.go
```diff
@@ -1399,8 +1399,12 @@ func (m *Manager) setConnectivityMajorityGroupsForClusterInternal(cluster *commo
 		majorityGroups[cidr] = majorityGroup
 	}
 
+	primaryMachineCIDR := ""
```
zaneb, line 1402: This will result in an error (from `IpInCidr()`) if the Machine CIDRs are not defined yet. But I presume what we actually want is for it to fall back to the previous behaviour in those cases.

zaneb:
> would this change cover the issue?

I have difficulty following the connectivity groups code, but it seems plausible. If we ignore the extra IP addresses at the point where we're reading the inventory then it should work as if those interfaces did not exist, which is what I think we want.

internal/network/connectivity_groups.go
```diff
@@ -482,7 +482,13 @@ func newL3QueryFactory(hosts []*models.Host, family AddressFamily) (hostQueryFac
 				if err != nil {
 					return nil, err
 				}
-				value[ip.String()] = true
+				ipInCidr, err := IpInCidr(ip.String(), primaryMachineCIDR)
```
ori-amizur, line 485: This misses the whole point of creating connectivity-groups.  The connectivity-groups should provide all connectivity information.  Then the validation should filter weather it needs L2, L3 or specific networks.

CrystalChun: I see...I couldn't find another place to filter ip addresses 😅 do you know of another place where we can filter them?

internal/cluster/cluster.go
CrystalChun: Gotcha, I modified it to check across multiple machine networks.

internal/cluster/cluster.go
CrystalChun: Good point, maybe if machine cidrs are not defined, then this validation shouldn't pass?

ori-amizur:
> /test edge-e2e-metal-assisted-none

It seems that the test passed,
But the machine-networks is empty: Here:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_assisted-service/6094/pull-ci-openshift-assisted-service-master-edge-e2e-metal-assisted-none/1770287733328056320/artifacts/e2e-metal-assisted-none/assisted-common-gather/artifacts/2024-03-20_03-51-58_82a1c33a-f02c-4afb-81e0-838e61b5069e/metadata.json

So I wonder where did the machine-networks come from? Needs to be checked.

ori-amizur:
It seems there is something wrong with the connectivity-groups.
The result of connectivity-groups is:
`{"192.168.127.0/24":["6043b820-6b7c-4b3b-a5c0-ca0f2b1a82dc","899a891c-ed6a-4e67-ba58-72588a95a9e0","c5895010-acda-4155-9b8d-7a4de74a82e3"],"192.168.145.0/24":[],"IPv4":["31b68c6d-cd3a-494c-8fb6-68784dda95ae","6043b820-6b7c-4b3b-a5c0-ca0f2b1a82dc","899a891c-ed6a-4e67-ba58-72588a95a9e0","c5895010-acda-4155-9b8d-7a4de74a82e3","feeccca3-46e8-45c6-9ca0-fbde91638fb7"],"IPv6":["31b68c6d-cd3a-494c-8fb6-68784dda95ae","6043b820-6b7c-4b3b-a5c0-ca0f2b1a82dc","899a891c-ed6a-4e67-ba58-72588a95a9e0","c5895010-acda-4155-9b8d-7a4de74a82e3","feeccca3-46e8-45c6-9ca0-fbde91638fb7"]}`

So there is IPv6 connectivity but there are no IPv6 addresses.

internal/network/connectivity_groups.go
ori-amizur: Actually you are right.  If we want to limit the addresses this is probably the place.
In general, in UMN any IP from any network can be part of the cluster.  Therefore, no machine-networks are required for multi-node UMN.
Since no machine-network should be provided in the install-config, OCP doesn't expect them either.
So even if we limit the addresses we use in connectivity-check, does OCP limit its activity to these addresses?

internal/network/connectivity_groups.go
```diff
@@ -431,9 +432,11 @@ func (l *l3Query) next() strfmt.UUID {
 		}
 		foundAddresses := make(map[string]bool)
 		for _, l3 := range rh.L3Connectivity {
-			_, foundAddress := addresses[l3.RemoteIPAddress]
-			if foundAddress && l3.Successful {
-				foundAddresses[l3.RemoteIPAddress] = true
+			if isInCidr := IpInCidrs(l3.RemoteIPAddress, l.machineNetworkCidrs); isInCidr {
```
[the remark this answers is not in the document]
CrystalChun: If this filter isn't added, will this condition still be true? 

https://github.com/openshift/assisted-service/pull/6094/files/e178b301e29d9b637beaaa8afbd5526af0d63222#diff-e0b0e8e6b4de6465aa429384eadda9ed02b2ceb1300e6b95a2f89b8a0858f658R442

internal/network/connectivity_groups.go
CrystalChun: Hmm not too sure, @zaneb do you know if OCP will use these addresses? Also if no machine networks are required for UMN then how do we filter them?

CrystalChun:
> It seems there is something wrong with the connectivity-groups. The result of connectivity-groups is: `{"192.168.127.0/24":["6043b820-6b7c-4b3b-a5c0-ca0f2b1a82dc","899a891c-ed6a-4e67-ba58-72588a95a9e0","c5895010-acda-4155-9b8d-7a4de74a82e3"],"192.168.145.0/24":[],"IPv4":["31b68c6d-cd3a-494c-8fb6-68784dda95ae","6043b820-6b7c-4b3b-a5c0-ca0f2b1a82dc","899a891c-ed6a-4e67-ba58-72588a95a9e0","c5895010-acda-4155-9b8d-7a4de74a82e3","feeccca3-46e8-45c6-9ca0-fbde91638fb7"],"IPv6":["31b68c6d-cd3a-494c-8fb6-68784dda95ae","6043b820-6b7c-4b3b-a5c0-ca0f2b1a82dc","899a891c-ed6a-4e67-ba58-72588a95a9e0","c5895010-acda-4155-9b8d-7a4de74a82e3","feeccca3-46e8-45c6-9ca0-fbde91638fb7"]}`
> 
> So there is IPv6 connectivity but there are no IPv6 addresses.


Oh I think it's because of the original condition here https://github.com/openshift/assisted-service/pull/6094/files#diff-e0b0e8e6b4de6465aa429384eadda9ed02b2ceb1300e6b95a2f89b8a0858f658R442 
Changed it so now it shouldn't return if there are no ipv6 addresses

internal/network/connectivity_groups.go
```diff
@@ -431,25 +432,29 @@ func (l *l3Query) next() strfmt.UUID {
...
+				_, foundAddress := addresses[l3.RemoteIPAddress]
+				if foundAddress && l3.Successful {
+					foundAddresses[l3.RemoteIPAddress] = true
+				}
 			}
 		}
-		if len(addresses) == len(foundAddresses) {
+		if len(foundAddresses) > 0 && len(addresses) == len(foundAddresses) {
```
ori-amizur, line 442: Nice.  Good catch

internal/network/connectivity_groups.go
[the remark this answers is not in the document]
ori-amizur: Yes, because the addresses were already filtered previously with the same criteria

ori-amizur (commented):
This change is incorrect because for multi-node UMN there are no machine-networks.  Even if we add it and it works we cannot force it because this change is not backward compatible.
If you still want it, you might need to either do not filter anything if there are no machine-networks, or create a new factory for this.

CrystalChun:
Looks like the connectivity groups are accurate now 
```
        "connectivity_majority_groups": "{\"192.168.127.0/24\":[\"7f4b73b0-14d4-4695-a185-2b62666ee175\",\"b1860fd5-3a8a-45c9-bcc5-8b41a7ed21be\",\"b8e6124e-ff3c-4958-86ae-f326607f56df\"],\"192.168.145.0/24\":[],\"IPv4\":[\"366d5130-0dd2-4543-b20b-9c518dbecc9e\",\"6593e35f-4ee2-41ad-bcb7-f88bf14bdd1b\",\"7f4b73b0-14d4-4695-a185-2b62666ee175\",\"b1860fd5-3a8a-45c9-bcc5-8b41a7ed21be\",\"b8e6124e-ff3c-4958-86ae-f326607f56df\"],\"IPv6\":[]}",

```

internal/network/connectivity_groups.go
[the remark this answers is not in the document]
CrystalChun: Gotcha, ok I removed this filter here

CrystalChun:
> This change is incorrect because for multi-node UMN there are no machine-networks. Even if we add it and it works we cannot force it because this change is not backward compatible.
If you still want it, you might need to either do not filter anything if there are no machine-networks, or create a new factory for this, and use it only in cases when there are available machine-networks. In this case it should not be used for UMN.

Modified the filter to only be used if there are machine networks! 
https://github.com/openshift/assisted-service/pull/6094/files#diff-e0b0e8e6b4de6465aa429384eadda9ed02b2ceb1300e6b95a2f89b8a0858f658R486

internal/network/connectivity_groups.go
zaneb: OCP doesn't really seem to use the machine-networks except for validations in the installer.

In UMN hosts are allowed to be in different networks. Because assisted has a limitation of only 1 machine-network, I guess we just allow none to be specified. #6071 removes that limitation. We still don't want any of the machine networks to overlap with e.g. the clusterNetwork or serviceNetwork, so this enables us to do all of the validations.

In any event, the agent installer always passes the machineNetwork(s), even with UMN.

internal/cluster/cluster.go
zaneb: No, I think changing it to fall back to the old behaviour (i.e. not filter IPs if there are no machine networks) was the right thing to do, for the reasons given in https://github.com/openshift/assisted-service/pull/6094#pullrequestreview-1950748577

CrystalChun: Great, this change should be in place then. Thank you for your input!

zaneb:
> In general the concept is that connectivity checks all possible connections.
> The belongs-to-majority-group checks if there is possible connections between hosts.

Right now it actually checks that there are no impossible connections between hosts. That's a very high bar.

> @zaneb are you sure that the requested change does not allow potential failures to occur?

Failures could occur if the actual routes set up don't result in traffic going over the network explicitly specified as the machine network, and also don't result in a successful connection over the networks the traffic is actually sent over. But as you said, in standard assisted installations the machine network is not provided anyway. So this change will only affect ABI.

zaneb:
It appears that when UMN is enabled (so [no VIPs](https://github.com/openshift/baremetal-runtimecfg/blob/master/cmd/runtimecfg/node-ip.go#L231)) the interface used for the Node IP is the one that has [the default route](https://github.com/openshift/baremetal-runtimecfg/blob/master/pkg/utils/addresses.go#L226).
If we had the full routing information we could limit to that interface and not have to rely on the user input being correct. Do we?

ori-amizur:
> It appears that when UMN is enabled (so [no VIPs](https://github.com/openshift/baremetal-runtimecfg/blob/master/cmd/runtimecfg/node-ip.go#L231)) the interface used for the Node IP is the one that has [the default route](https://github.com/openshift/baremetal-runtimecfg/blob/master/pkg/utils/addresses.go#L226). If we had the full routing information we could limit to that interface and not have to rely on the user input being correct. Do we?

I believe it needs to be tested

CrystalChun:
/hold
This change doesn't actually act correctly with UMN and multiple machine networks defined

CrystalChun:
Also there's currently a validation that would fail this too https://github.com/openshift/assisted-service/blob/17959954d606e1faac51baa251d5df3e090dd3f9/internal/cluster/validations/validations.go#L825-L828
No more than one machine network can be defined if it's not dual-stack

zaneb:
> > Right now it actually checks that there are no impossible connections between hosts. That's a very high bar.
> 
> It is all a matter of what OCP dictates.

If the nodes can talk to each other on their kubelet IPs then OCP will be happy.

PR #6115, opened 2024-03-27
MGMT-17361: Changing the addition of configuration release images to run when the instance becomes leader instead of skipping if it not

danmanor:
Currently, when an assisted-service instance starts, it first checks if it is the leader. If it is, the instance deletes all the release images from the database and adds the configuration ones. If not, it bypasses this procedure. Due to the deployment strategy of the service in SaaS, a scenario arises where none of the instances become the leader when executing this flow. This results in an empty database if the service is run for the first time, leading the `/openshift-versions` endpoint to return an empty list. Consequently, the UI is unable to display the new cluster page. To address this issue, I propose modifying the current flow to operate similarly to how migrations are handled: wait to become the leader, then execute the code, and finally release the lock.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -319,7 +319,7 @@ func main() {
 		maxDuration(Options.InstructionConfig.DiskCheckTimeout, Options.InstructionConfig.ImageAvailabilityTimeout)+1*time.Minute)
 	var lead leader.ElectorInterface
 	var k8sClient *kubernetes.Clientset
-	var autoMigrationLeader leader.ElectorInterface
+	var startupLeader leader.ElectorInterface
 
 	mirrorRegistriesBuilder := mirrorregistries.New()
 	releaseHandler := oc.NewRelease(&executer.CommonExecuter{},
@@ -382,7 +382,7 @@ func main() {
 		failOnError(cerr, "Failed to create kubernetes cluster config")
 		k8sClient = kubernetes.NewForConfigOrDie(cfg)
 
-		autoMigrationLeader = leader.NewElector(k8sClient, leader.Config{LeaseDuration: 5 * time.Second,
+		startupLeader = leader.NewElector(k8sClient, leader.Config{LeaseDuration: 5 * time.Second,
 			RetryInterval: 2 * time.Second, Namespace: Options.LeaderConfig.Namespace, RenewDeadline: 4 * time.Second},
 			"assisted-service-migration-helper",
 			log.WithField("pkg", "migrationLeader"))
@@ -398,7 +398,7 @@ func main() {
 	case deployment_type_onprem, deployment_type_ocp:
 
 		lead = &leader.DummyElector{}
-		autoMigrationLeader = lead
+		startupLeader = lead
 
 		if Options.DeployTarget == deployment_type_ocp {
 			ocpClient, err = k8sclient.NewK8SClient("", log)
@@ -409,7 +409,7 @@ func main() {
 		log.Fatalf("not supported deploy target %s", Options.DeployTarget)
 	}
 
-	failOnError(autoMigrationWithLeader(autoMigrationLeader, db, log), "Failed auto migration process")
+	failOnError(autoMigrationWithLeader(startupLeader, db, log), "Failed auto migration process")
 
 	Options.UploaderConfig.AssistedServiceVersion = versions.GetRevision()
 	Options.UploaderConfig.Versions = Options.Versions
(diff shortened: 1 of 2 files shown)


danmanor:
waiting for - https://github.com/openshift/assisted-service/pull/6116

cmd/main.go
```diff
@@ -962,11 +962,11 @@ func runOpenshiftReleaseSyncerIfNeeded(
 		if err != nil {
 			failOnError(err, "error occurred while creating release sources handler")
 		}
-		log.WithField("pkg", "releasesources").Info("Running Openshift Releases Syncer at startup")
-		err = releaseSourcesHandler.SyncReleaseImages()
-		if err != nil {
-			failOnError(err, "error occurred while syncing release images at startup")
-		}
+
+		failOnError(lead.RunWithLeader(context.Background(), func() error {
```
carbonin, line 966: `RunWithLeader` is a concurrency control. This means that every replica will run this code.

This is fine for automigration because no work is done once the database is fully migrated, but for this it will lead to us removing and re-adding all releases for every replica, right? Are we okay with that behavior?

cmd/main.go
```diff
@@ -382,7 +382,7 @@ func main() {
 		failOnError(cerr, "Failed to create kubernetes cluster config")
 		k8sClient = kubernetes.NewForConfigOrDie(cfg)
 
-		autoMigrationLeader = leader.NewElector(k8sClient, leader.Config{LeaseDuration: 5 * time.Second,
+		secondLeader = leader.NewElector(k8sClient, leader.Config{LeaseDuration: 5 * time.Second,
```
carbonin, line 385: I'm not sure reusing this is the best option. Now querying for release images will prevent migration from running.

Maybe a better solution on startup would be to use a database lock on the table.
Then check if the table is empty, if it is populate it and release the table lock, if not release the lock with no action.

This would lead to only one replica populating the table on first deploy then later updates would be done from the thread which uses the "regular" leader.

carbonin:
>Due to the deployment strategy of the service in SaaS, a scenario arises where none of the instances become the leader when executing this flow.

I thought we talked about this and decided that even if the current leader comes down last the last new replica should become the new leader on startup. I don't fully understand how we ran into this issue.

rccrdpccl:
> > Due to the deployment strategy of the service in SaaS, a scenario arises where none of the instances become the leader when executing this flow.
> 
> I thought we talked about this and decided that even if the current leader comes down last the last new replica should become the new leader on startup. I don't fully understand how we ran into this issue.

Do leaders only get elected on startup? If not maybe a pod that already started would then become leader thus avoiding the last replica getting leader lease

danmanor:
> I thought we talked about this and decided that even if the current leader comes down last the last new replica should become the new leader on startup. I don't fully understand how we ran into this issue.

It did became the new leader, but at the wrong time. The problem is that non of the pods was leader at the moment it executed my code

cmd/main.go
danmanor: I think this solution is a good fit for now for patching stage, then we can improve the solution by trying to use postgres resource protection mechanism instead

gamli75:
> Also, we used to have periodic (actually I think they may have run after each PR merge) sanity checks on the integration environment. Something like that would have caught this and prevented us from promoting to stage.
> 
> @gamli75 maybe it's time to bring something like that back?

Currently, we only have the option to run the prow job against production. 
There is an option to run e2e test in QE Jenkins - we can ask QE to run it daily or every couple of hours on integration but it means that someone will need to investigate failures :)

cmd/main.go
danmanor: Yes you are right, I tested it and it doesn't have performance impact as it will happen only once per instance. I agree this is not ideal but I don't see any other option for running this only by one instance, as the table need to be cleaned up anyway if this is not the there are records in the DB already

danmanor: I can try to explore a more ideal solution after

cmd/main.go
carbonin: Okay, if this is actually going to be temporary then it's fine for now, but please open an issue to revisit this.

cmd/main.go
```diff
@@ -952,21 +952,22 @@ func runOpenshiftReleaseSyncerIfNeeded(
...
 	releaseHandler oc.Release,
-	lead leader.ElectorInterface,
+	leader,
+	startupLeader leader.ElectorInterface,
 	failOnError func(err error, msg string, args ...interface{}),
 ) *thread.Thread {
 	if !Options.EnableKubeAPI && Options.ReleaseSourcesConfig.ReleaseSources != "" {
 		releaseSourcesHandler, err := releasesources.NewReleaseSourcesHandler(
-			releaseSources, releaseImages, log, db, Options.ReleaseSourcesConfig, lead,
+			releaseSources, releaseImages, log, db, Options.ReleaseSourcesConfig, leader,
```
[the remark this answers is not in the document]
danmanor: This is the old leader, sorry for the unnecessary diff (renaming), I didn't want to rerun the tests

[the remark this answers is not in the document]
carbonin: ah okay

PR #6130, opened 2024-03-28
WIP: MGMT-17317: Use IBI to provision SNO enhancement doc

adriengentil:


adriengentil:
/retitle WIP: MGMT-17317: Use IBI to provision SNO enhancement doc

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+### Open Questions
+
+When selecting Openshift version:
+- Do we plan to provide “curated” seed images?
+- Should we allow users to install their own seed?
+
+When restoring the image seed:
+- Is it worth to pre-cache the container images in SaaS flow? they will be
```
carbonin, line 110: Pre-cache would be done when creating the seed, right?

At install time I don't think it matters if we pull the images before we restore or after.

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+# Support Image-Based Installation in Assisted-Installer to install SNO
+
+## Summary
+
+Currently, installing an SNO cluster using the Assisted-Installer takes up to
+~40min using the current flow which installs OCP from the ground-up.
+
+We can reduce the installation time to ~10min by leveraging the Image-Based
+installation method, and as a way to increase the success rate of SNO
+installations as we restore previously working setup.
```
carbonin, line 18: Maybe we should get more exact timings for this.

Does that 40 minutes include all the pre-install validations we're running or is that from the time the user hits the install button? We're going to still be running all those validations during discovery right?

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+### Risks and Mitigations
+
+
+## Design Details [optional]
+
+
+### Open Questions
+
+When selecting Openshift version:
+- Do we plan to provide “curated” seed images?
```
carbonin, line 106: I don't think this is really an option.

Correct me if I'm wrong @javipolo , but the seed config needs to match the hardware of the install target, right?

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+- Should we allow users to install their own seed?
```
carbonin, line 107: Assuming this is the direction we go, we'll need some kind of UX for the user to tell us the image. Then we'll need to ensure we don't actually try to access it from our application as that could be a security risk.

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+When restoring the image seed:
+- Is it worth to pre-cache the container images in SaaS flow? they will be
+  downloaded anyway on reboot
+- Do we need to support the dedicated partition for /var/lib/containers?
+   - If yes, is there a sensible default for its size? Or do we need to ask the
+     user?
+
+Installation:
+- I don’t know the details, but I foresee some re-work of the around the states
+  the installation goes through?
```
carbonin, line 118: Assuming we continue to do all the same validations I'd guess this will mostly be around the install "stages".
We could probably reuse some (writing image to disk), but probably won't need others.

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+- Grow `/` partition
+- Mount `/`, `/boot` and `/var/lib/container`
+- Download and restore the seed on the disk using `lca-cli`
+
+[Reference](https://github.com/openshift-kni/lifecycle-agent/blob/main/ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh)
+
+#### Configuration
+
+After the image seed is restored on the disk, the `assisted-installer` will
+drop the configuration in `/opt/openshift`:
```
carbonin, line 73: We'll also need to consider static network configuration.

I think we should be able to continue to provide nmconnection files (as we are currently in assisted) but in "real" IBI we're using nmstate and running nmstatectl on the host so we'll want to make sure the nmconnection path continues to work.

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Installation
+
+When the user will hit the installation button, the `assisted-installer` will
+restore the image seed on the disk selected for installation:
+- Install RHCOS using `coreos-installer`
+- Partition the disk to create a dedicated `/var/lib/containers`
+- Grow `/` partition
+- Mount `/`, `/boot` and `/var/lib/container`
+- Download and restore the seed on the disk using `lca-cli`
```
carbonin, line 66: Where does `lca-cli` come from? How do we get it into the discovery image?

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+  other environments at this point
+
+## Proposal
+
+### As a user, I want to install a SNO using the IBI method
+
+#### REST API changes
+
+A new flag will be added to the cluster object [TBD], along with
+extra-parameters specific to IBI [TBD].
```
carbonin, line 47: I think it's worth coming up with at least a first pass on the API you'll add in the enhancement.

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+- assisted-controller deployment manifest, it will tell us if the installation
+  succeeded, and provide a way to get installation logs?
+- generate cryptographic materials in order to provide a kubeconfig file to the
+  user
+- SSH keys
+- Eventual extra manifests
+   - OLM operator manifests
+   - Custom manifests
+
+### Implementation Details/Notes/Constraints [optional]
```
carbonin, line 94: We should do some work to call out options/features that we currently support, but won't be able to for IBI clusters.

Things like host ignition customization, host install args (maybe these will work since we're still running `coreos-installer`?), cluster network (cidr), are some that come to mind immediately. Really anything that will be taken directly from the seed, but is already in our API we'll need to make optional and block the user from setting for IBI clusters.

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+   - BaseDomain: will be picked from the cluster definition 
+   - ClusterID: will be picked from the cluster definition 
+   - SSHkey: will be provided by the user
+   - KubeadminPasswordHash: will be generated by Assisted-Installer and
+     provided to the user
+- Pull secret
+- assisted-controller deployment manifest, it will tell us if the installation
+  succeeded, and provide a way to get installation logs?
+- generate cryptographic materials in order to provide a kubeconfig file to the
+  user
```
carbonin, line 88: @omertuc brought up a point recently about us baking the private keys into the ISO, it might be wise to have these go over an API request from the agent rather than sit in an artifact that the user might misplace.

docs/enhancements/install-sno-using-ibi.md
omertuc: I agree. I believe @eranco74 already has some numbers in a doc somewhere

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+## Proposal
+
+### As a user, I want to install a SNO using the IBI method
+
+#### REST API changes
+
+A new flag will be added to the cluster object [TBD], along with
+extra-parameters specific to IBI [TBD].
+
+`supported-versions` API with be amended to return only the version of OCP
```
omertuc, line 49:
```suggestion
`supported-versions` API will be amended to return only the version of OCP
```

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+available for installation with IBI.
```
omertuc, line 50: Under which circumstanaces?

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+A new feature support will be created for the IBI installation. 
+The feature will be gated by an OCM capability until we make it available to
+all users.
+
+#### Installation
+
+When the user will hit the installation button, the `assisted-installer` will
+restore the image seed on the disk selected for installation:
+- Install RHCOS using `coreos-installer`
+- Partition the disk to create a dedicated `/var/lib/containers`
```
omertuc, line 63: What's the point of this?

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+- Download and restore the seed on the disk using `lca-cli`
+
+[Reference](https://github.com/openshift-kni/lifecycle-agent/blob/main/ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh)
+
+#### Configuration
+
+After the image seed is restored on the disk, the `assisted-installer` will
+drop the configuration in `/opt/openshift`:
+- A
+  [manifest.json](https://github.com/openshift-kni/lifecycle-agent/blob/main/docs/post-pivot-configuration.md#user-specifications)
```
omertuc, line 75: Since this is a technical document, let's link to the SeedReconfiguration struct directly, as it's the source of truth and the linked doc may drift out of date

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+  with: 
+   - Hostname: will be picked from the inventory of host
+   - Cluster name: will be picked from the cluster definition 
+   - BaseDomain: will be picked from the cluster definition 
+   - ClusterID: will be picked from the cluster definition 
+   - SSHkey: will be provided by the user
+   - KubeadminPasswordHash: will be generated by Assisted-Installer and
+     provided to the user
+- Pull secret
```
omertuc, line 84: Why is this bullet indented separately?

docs/enhancements/install-sno-using-ibi.md
omertuc: Yeah, but let's keep things simple for now and have that in the back of ours minds as a ticket, until we figure out the security side of things and whether this is a threat that's relevant to protect from

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+- assisted-controller deployment manifest, it will tell us if the installation
+  succeeded, and provide a way to get installation logs?
+- generate cryptographic materials in order to provide a kubeconfig file to the
+  user
+- SSH keys
```
omertuc, line 89:
```suggestion
- Authorized SSH keys
```

SSH keys can mean many different things, so it's better to be explicit

docs/enhancements/install-sno-using-ibi.md
romfreiman: how useful would it be that u have to generate 1 seed to install 1 cluster? :) I assume it would be usecase only for repeatable users that run heavy automation - we can check in the telemetry maybe and talk to the relevant teams.

docs/enhancements/install-sno-using-ibi.md
javipolo: Nope. Precached images are not in the seed, this happens in a later stage, after installing the seed image to disk

The main reasons of running it BEFORE rebooting into the new system are:
- For IBU: Minimize downtime (time after reboot and until system becomes fully available)
- For IBI: Be able to pre-cache in factory, and then after dispatching to edge, just reconfigure the device as fast as possible

I don't think any of those apply for SaaS, but it does not hurt to do it either

docs/enhancements/install-sno-using-ibi.md
omertuc: Found a discussion about the first question [here](https://github.com/openshift/assisted-service/pull/6130#discussion_r1560018301)

docs/enhancements/install-sno-using-ibi.md
javipolo: Not so much. In the vDU case we needed to match hardware quite more because of the performance profile that is applied, in the case of a vanilla OCP I don't think the hardware needs to match so much .....

IIRC it needs to have enough memory and CPU, and be of the same architecture, but those requirements are already there for normal installations ....

If there are very special needs, then yes, a "vanilla" seed image wont fit

docs/enhancements/install-sno-using-ibi.md
omertuc: One huge benefit for precaching in SaaS during discovery is that major network problems in user environments are more likely to be surfaced and result in an explicit "I can't even get the images to pull" error, rather than obscure errors / timeouts during cluster initialization caused by slow network

docs/enhancements/install-sno-using-ibi.md
adriengentil: I'll add it for sure once get some time, the rough idea I had was something like that in the cluster object:
```
{
  installation_method: "ibi",
  ibi_cluster_image: "quay.io/...." # allowed to be set only when installation_method is "ibi"
  openshift_version: nil, # because installation_method is set to "ibi"
  ocp_release_image: nil # because installation_method is set to "ibi"
}
```

docs/enhancements/install-sno-using-ibi.md
adriengentil: The one I used comes from a container image. So I guess we can pull it, and execute it like any other commands? 
@javipolo can we always use the last version to restore any existing? Is there some constraints on the `lca-cli` version based on the OCP version we try to restore?

docs/enhancements/install-sno-using-ibi.md
adriengentil: Isn't static networking already based on nmstate in assisted (which generate nmconnection files)? or I miss something?
For sure, we need to consider this flow.

docs/enhancements/install-sno-using-ibi.md
adriengentil: Related to this topic, I understand we also have a way that works without the partition for `/var/lib/containers`. In the case we allow users to restore their own snapshots, should we be able to support both partition schemes? @javipolo what do you think?

Also, when `/var/lib/containers` is required, how can we determine the right size for it?

docs/enhancements/install-sno-using-ibi.md
eranco74: The actual SNO installation takes 30 minutes.

docs/enhancements/install-sno-using-ibi.md
adriengentil: As discussed, we decided to not pre-cache the images because it would not reduce the installation time. As @omertuc said the only benefit would be the triage on Red Hat side, but it should not be the only reason to enable the caching.

docs/enhancements/install-sno-using-ibi.md
adriengentil: After discussion we decided that in the scope of this enhancement, we will let the user to restore a seed they created. I'll add a section for future work with:
* add an option to install and create a seed in Assisted-Installer
* provide vanilla seed images

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+[Reference](https://github.com/openshift-kni/lifecycle-agent/blob/main/ib-cli/installationiso/data/install-rhcos-and-restore-seed.sh)
+
+#### Configuration
+
+After the image seed is restored on the disk, the `assisted-installer` will
+drop the configuration in `/opt/openshift`:
+- A
+  [manifest.json](https://github.com/openshift-kni/lifecycle-agent/blob/main/docs/post-pivot-configuration.md#user-specifications)
+  with: 
+   - Hostname: will be picked from the inventory of host
```
adriengentil, line 77: openshift-installer will provide some commands to generate the config artifacts

docs/enhancements/install-sno-using-ibi.md
adriengentil: I'll add the installation steps

docs/enhancements/install-sno-using-ibi.md
adriengentil: lca-cli is version per OCP release, so we should be able to pull it as long as we now the OCP version of the seed we are going to restore.

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+
+A new feature support will be created for the IBI installation. 
+The feature will be gated by an OCM capability until we make it available to
+all users.
+
+#### Installation
+
+When the user will hit the installation button, the `assisted-installer` will
+restore the image seed on the disk selected for installation:
+- Install RHCOS using `coreos-installer`
```
adriengentil, line 62: we should know in advance what version of OCP the user is going to restore

docs/enhancements/install-sno-using-ibi.md
```diff
@@ -0,0 +1,133 @@
...
+
+`supported-versions` API will be amended to return only the version of OCP
+available for installation with IBI.
+
+#### Feature support
+
+A new feature support will be created for the IBI installation. 
+The feature will be gated by an OCM capability until we make it available to
+all users.
+
```
adriengentil, line 57: Add word about Infra-env and RHCOS version that should be used

docs/enhancements/install-sno-using-ibi.md
javipolo: As discussed, let's add a label for the OCI container that says which method was done:
- partition
- directory
- none

this way we could support both methods, or having precaching at all

PR #6145, opened 2024-04-03
MGMT-17412: Change fallbck behavior when error occuring in OpenShift Release Syncer

danmanor:
If release images are already present in the database, the service should continue to use this stale data instead of failing. In addition, currently failure to fetch new release images from the API does not trigger an error, allowing the service to proceed, truncate the table, and insert only static images. Hence, I propose treating failures to fetch release images as errors as well.

## Failure Scenarios of `OpenShift Release Syncer`

| Scenario              | Currently                                                                                   | Target                                                                                       |
|-----------------------|---------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------|
| **Startup & Empty Table** | Service fails to start.                                                                     | Service should fail to start.                   |
| **Startup & Full Table** | Service fails to start.                                                                     | Service should start successfully with stale data.              |
| **Routine & Empty Table** | Shouldn't happen, though if it does then the service will continue using stale data.                                                                          | Shouldn't happen, though if it does then the service should continue using stale data.                                                                            |
| **Routine & Full Table**  | The service will continue using stale data. | The service should continue using stale data.                                 |

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -76,7 +76,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
-	"github.com/thoas/go-funk"
 	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
 	"gorm.io/gorm/logger"
@@ -261,7 +260,7 @@ func main() {
 	failOnError(err, "Failed to initialize OSImages")
 
 	var releaseImagesArray = models.ReleaseImages{}
-	parseReleaseImages(&releaseImagesArray, failOnError)
+	versions.ParseReleaseImages(&releaseImagesArray, Options.ReleaseImages, failOnError)
 
 	var releaseSourcesArray = models.ReleaseSources{}
 	if Options.ReleaseSourcesConfig.ReleaseSources != "" {
@@ -270,7 +269,7 @@ func main() {
 	}
 
 	var ignoredOpenshiftVersions = []string{}
-	parseIgnoredOpenshiftVersions(&ignoredOpenshiftVersions, failOnError)
+	versions.ParseIgnoredOpenshiftVersions(&ignoredOpenshiftVersions, Options.IgnoredOpenshiftVersions, failOnError)
 
 	log.Println(fmt.Sprintf("Started service with OS Images %v, Release Images %v, Release Sources %v, Ignored OpenShift Versions %v",
 		Options.OsImages, Options.ReleaseImages, Options.ReleaseSourcesConfig.ReleaseSources, Options.IgnoredOpenshiftVersions))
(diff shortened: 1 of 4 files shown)


cmd/main.go
```diff
@@ -966,7 +967,22 @@ func runOpenshiftReleaseSyncerIfNeeded(
 
 		failOnError(startupLeader.RunWithLeader(context.Background(), func() error {
```
rccrdpccl, line 968: should we extract this (and possibly other if applicable) anonymous function(s) to the relevant package?

danmanor: What will be the advantage of this move? IMHO when you see the anonymous functions in main you get a better sense on what they are doing, also I don't think it should be part of `releasesources` package, because it is about the behavior of the output  of its use case.

danmanor: I agree, it would be better to extract it. As to destination, I think it is part of the functionality of main, hence we should extract it to a function out side the main function, but in the same package. WDYT ?

rccrdpccl: No, I do not think so. Common it's a "socks drawer" right now, the issue is if we put everything there it soon will be a nightmare about circular dependencies.

We can create a package used to initialize stuff from main: this way the dependencies will be from such package, freeing main from complex dependencies.

maybe `internal/releasesources/startup` ? `internal/releasesources//init` ? WDYT?

internal/releasesources/release_sources.go
```diff
@@ -503,3 +506,58 @@ func getReleaseImageReference(version, cpuArchitecture string) string {
...
+) *thread.Thread {
+	if !enableKubeAPI && releaseSourcesConfig.ReleaseSources != "" {
+		releaseSourcesHandler, err := newReleaseSourcesHandler(
+			releaseSources, releaseImages, log, db, releaseSourcesConfig, leader,
+		)
+		if err != nil {
+			failOnError(err, "error occurred while creating release sources handler")
+		}
+
+		failOnError(startupLeader.RunWithLeader(context.Background(), func() error {
```
rccrdpccl, line 530: Is failOnError function panicking? shouldn't this return an error and let the main fail?

cmd/main.go
```diff
@@ -443,9 +442,14 @@ func main() {
 	hostStateMonitor.Start()
 	defer hostStateMonitor.Stop()
 
-	failOnError(addReleaseImagesToDBIfNeeded(db, releaseImagesArray, startupLeader, log), "error occured while adding configuration release images to the DB if needed")
-	openshiftReleaseSyncer := runOpenshiftReleaseSyncerIfNeeded(releaseSourcesArray, releaseImagesArray, db, log, releaseHandler, lead, startupLeader, failOnError)
-	defer stopOpenshiftReleaseSyncerIfNeeded(openshiftReleaseSyncer)
+	failOnError(
```
rccrdpccl, line 445: in my mind this could be something like:
```
failOnError(lead.RunWithLeader(releaseSources.UpdateImagesIfEmpty(releaseSources, releaseImages)))
```
main difference is we leave lead stuff in main.

danmanor: we can have `RunWithLeader` in main, but then we are back in the problem of having the inside function in `main.go` (which doesn't make it shorter) or placing it in an external package (which doesn't really fit in `versions` or `releasesources`, so it means we beed to create a new package for it probably, which I don't know if it not overstepping the purpose of this PR. WDYT ?

rccrdpccl:
/lgtm
/approve

we'll address the suggested changes at a later stage

cmd/main.go
rccrdpccl: I would see those functions in another package, but even if we had those kind of function in main and all other logic specific for some package extracted, main would still be shorter than now

PR #6150, opened 2024-04-04
MGMT-17438: bugfix, make sure ODF and LVM requirements do not change if MCE not selected

rccrdpccl:
When selecting MCE and ODF we will add additional storage requirements to the operator singleton, which was never reset. This mean that if we install once MCE and ODF, all installations after that will keep additional requirements even if not necessary. This PR fixes this issue.

/cc @bkopilov @carbonin 

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


carbonin:
>we will add additional storage requirements to the operator singleton, which was never reset

It feels to me like the answer isn't to reset this, but to not do it in the first place. Is that an option?
If we have some kind of global for these operators then that's bad enough, but trying to edit and reset data on them without any kind of sync mechanism is worse.

rccrdpccl:
> > we will add additional storage requirements to the operator singleton, which was never reset
> 
> It feels to me like the answer isn't to reset this, but to not do it in the first place. Is that an option? If we have some kind of global for these operators then that's bad enough, but trying to edit and reset data on them without any kind of sync mechanism is worse.

I'll try to think something, the main issue is that I do not want to make aware MCE (or any other operator) of other operators, that would cause a mess of dependencies.

rccrdpccl:
The change aims to pass to additional requirements when that can be understood only from the manager when validating the hosts. In this instance the additional disk requirement is *only* when a combinations of operators are activated.
As each operator is not aware of the others, only the manager can issue additional requirements.

internal/operators/api/api.go
```diff
@@ -38,7 +38,7 @@ type Operator interface {
 	// ValidateCluster verifies whether this operator is valid for given cluster
 	ValidateCluster(ctx context.Context, cluster *common.Cluster) (ValidationResult, error)
 	// ValidateHost verifies whether this operator is valid for given host
-	ValidateHost(ctx context.Context, cluster *common.Cluster, hosts *models.Host) (ValidationResult, error)
+	ValidateHost(ctx context.Context, cluster *common.Cluster, hosts *models.Host, interOperatorRequirements *models.ClusterHostRequirementsDetails) (ValidationResult, error)
```
carbonin, line 41: Maybe name this just `additionalOperatorRequirements`?
I could see this being useful for other runtime stuff that might come up.

carbonin (commented):
Looks good other than that name.

carbonin (approved):

PR #6158, opened 2024-04-07
MGMT-17413: Change pull secret validation on register/update cluster/infraenv to run only against the required images

danmanor:
Currently, assisted-service performs validation of the user pull secret during actions like registering / updating a cluster / infraenv, in which it checks that the pull secret contains the tokens for all release images.  Therefore, currently we can't add nightly release images to stage environment as it blocks users without the right token from installing clusters with different release images. To mitigate this, I am changing the validation on register/update cluster to run against the cluster's release image registry, and register/update Infraenv to run against images besides release images.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


carbonin (commented):
Just some nits, otherwise looks good.

FWIW this was added way back when we first started supporting multiple versions. So it's effectively been like this forever: https://github.com/openshift/assisted-service/pull/805

internal/cluster/validations/validations.go
```diff
@@ -171,34 +172,29 @@ func AddRHRegPullSecret(secret, rhCred string) (string, error) {
...
 	if v.authHandler.AuthType() == auth.TypeRHSSO {
-
 		r, ok := creds["cloud.openshift.com"]
 		if !ok {
 			return &PullSecretError{Msg: "pull secret must contain auth for \"cloud.openshift.com\""}
 		}
 
-		user, err := v.authHandler.AuthAgentAuth(r.AuthRaw)
+		var user interface{}
+		user, err = v.authHandler.AuthAgentAuth(r.AuthRaw)
```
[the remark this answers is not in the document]
danmanor: Lint, I was shadowing `err`

cmd/main.go
```diff
@@ -365,10 +365,6 @@ func main() {
 		Options.InstructionConfig.AgentImage,
 	}
 
-	for _, releaseImage := range releaseImagesArray {
-		images = append(images, *releaseImage.URL)
```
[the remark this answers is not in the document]
danmanor: Sure

carbonin (approved):

internal/cluster/validations/validations.go
```diff
@@ -208,8 +204,19 @@ func (v *registryPullSecretValidator) ValidatePullSecret(secret string, username
...
 
-	for registry := range *v.registriesWithAuth {
+	images := make([]string, len(v.images))
+	copy(images, v.images)
+
+	if releaseImageURL != "" {
+		images = append(images, releaseImageURL)
+	}
+
+	registriesWithAuth, err := getRegistriesWithAuth(v.config.PublicRegistries, ignoreListSeparator, images...)
```
tsorya, line 214: i assume it is not a big deal to go through images all the time but not sure why not to leave the logic we had + append registry to v.registriesWithAuth once you get releaseImageURL ?

danmanor: I agree with you going over the images once is better than 4 times, but I didn't want to add the release image url to that list because it is dynamic, which means we will have to delete it after. I can extract the login of `getRegistriesWithAuth` and run it once as instance creation and once again for each release image

tsorya: i assume i missed something here, can't we just copy the object and add dynamic to static when required?
or we don't have static objects anymore?

carbonin: I think what you're suggesting is to maintain a list of registries with auth based on the static images we check then still call `getRegistriesWithAuth` for the release image when it's passed. We's also still have to do the same `copy` + `append` dance but with registries rather than images.

Is that right @tsorya ?

I think it's a good idea.

tsorya: Yep :) 
Thanks @carbonin

danmanor: Done :)

internal/cluster/validations/validations.go
```diff
@@ -322,7 +338,6 @@ func ParseRegistry(image string) (string, error) {
 // on a list of used images and an ignore list. The ingore list comes as a string and a separator
 // to make it easier to read from a configuration variable
 func getRegistriesWithAuth(ignoreList string, ignoreSeparator string, images ...string) (*map[string]bool, error) {
```
carbonin, line 340: Why does this function have to return a map?
It seems like that's making your job in the caller much more difficult.

internal/cluster/validations/validations.go
```diff
@@ -208,8 +209,23 @@ func (v *registryPullSecretValidator) ValidatePullSecret(secret string, username
...
+	if releaseImageURL != "" {
+		MoreRegistriesWithAuth, err := getRegistriesWithAuth(v.config.PublicRegistries, ignoreListSeparator, releaseImageURL)
+		if err != nil {
+			return err
+		}
+
+		registriesWithAuth = append(registriesWithAuth, funk.Map(*MoreRegistriesWithAuth, func(registry string, _ bool) string {
+			return registry
+		}).([]string)...)
+	}
```
carbonin, line 226: Instead of doing this let's extract the logic from `getRegistriesWithAuth` that determines if a particular image's registry needs auth. That function can return (bool, string, error) where the bool indicates if the registry returned in string (parsed from the image) requires auth.

This and making the `v.registriesWithAuth` a slice instead of a map should simplify quite a lot of this.

internal/cluster/validations/validations.go
```diff
@@ -208,8 +209,23 @@ func (v *registryPullSecretValidator) ValidatePullSecret(secret string, username
...
+
+	for _, registry := range registriesWithAuth {
```
carbonin, line 228: It may also make sense to make the contents of this loop a function then just call it twice:
- Once in a loop for the common images
- An additional time if a release image was provided

This, I think would be a lot simpler than all the work you're doing above to create a single list from a global list and possibly an additional image.

carbonin (changes requested):
This still ended up quite a bit more complicated than I think it needs to be.

Can we also pull the pull secret validator into it's own file?
It's strange that it's buried in this file with a bunch of other cluster validation code.

PR #6257, opened 2024-05-02
MGMT-17700: Assign none platform node-ips based on connected addresses and etcd restrictions

ori-amizur:
When none platform is in use, if there is ambiguity in node-ip assignment, then incorrect assignment might lead to installation failure. This happens when etcd detects that the socket address from an etcd node does not match the expected address in the peer certificate. In this case etcd rejects such connection.

Example: assuming two networks - net1 and net2.
master node 1 has 1 address that belongs to net1.
master node 2 has 2 addresses. one that belongs to net 1, and another that belongs to net 2 master node 3 has 1 address that belongs to net 1. If the selected node-ip of master node 2 belongs to net 2, then when it will create a connection with any other master node, the socket address will be the address that belongs to net 1. Since etcd expects it to be the same as the node-ip, it will reject the connection.

This can be solved by node-ips selection that will not cause such a conflict. Node ips assignment should be done through ignition. To correctly set bootstrap ip, the machine-network for the cluster must be set to match the selected node-ip for that host.

[MGMT-17701](https://issues.redhat.com//browse/MGMT-17701): Add capability to calculate none platform node-ips based on L3 connected addresses and connectivity

Calculate the node-ips for none platform cluster.  Node ip calculation is actually calculation of the node-ip, hint, and cidr. Node-ip is the ip address that exists on the host that was selected as the node-ip.
Hint is actually an IP address that does not exist on the host, but must belong to the subnet of the node-ip.
Cidr is the subnet in cidr notation that the node-ip and hint belong to. Node-ip calculation is either done for all cluster hosts or none of them.

[MGMT-17702](https://issues.redhat.com//browse/MGMT-17702): Modify ignition to use calculated node-ips for none platform

In order to set node-ip, ignition is modified.  A file called /etc/default/nodeip-configuration contains the hint as was set by node-ip generation.
In addition the bootstrap-ip is set in ignition as was set in node-ip generation.  This is set as environment variable called "OPENSHIFT_INSTALL_BOOTSTRAP_NODE_IP".

[MGMT-17703](https://issues.redhat.com//browse/MGMT-17703): Modify machine-network based on calculated node-ips

The etcd in boostrap uses he machine-network to set the IP address. Therefore during install-config generation the machine-network may be replaced by the cidr from the node-ip generation.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @tsorya 
/cc @paul-maidment

### DIFF
--- a/internal/common/common.go
+++ b/internal/common/common.go
@@ -609,3 +609,7 @@ func IsExternalIntegrationEnabled(platform *models.Platform, platformName string
 func IsOciExternalIntegrationEnabled(platform *models.Platform) bool {
 	return IsExternalIntegrationEnabled(platform, ExternalPlatformNameOci)
 }
+
+func IsMultiNodeNonePlatformCluster(cluster *Cluster) bool {
+	return !IsSingleNodeCluster(cluster) && swag.BoolValue(cluster.UserManagedNetworking)
+}
--- a/internal/network/none_platform_node_ips_allocation.go
+++ b/internal/network/none_platform_node_ips_allocation.go
@@ -1,0 +1,321 @@
+package network
+
+import (
+	"encoding/json"
+	stderrors "errors"
+	"fmt"
+	"net"
+	"sort"
+	"strings"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/openshift/assisted-service/internal/common"
+	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
+	"github.com/samber/lo"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+// When none platform is in use, if there is ambiguity in node-ip assignment, then incorrect assignment might lead
+// to installation failure.  This happens when etcd detects that the socket address from an etcd node does not match
+// the expected address in the peer certificate. In this case etcd rejects such connection.
+// Example: assuming two networks - net1 and net2.
+// master node 1 has 1 address that belongs to net1.
+// master node 2 has 2 addresses.  one that belongs to net 1, and another that belongs to net 2
+// master node 3 has 1 address that belongs to net 1.
+// If the selected node-ip of master node 2 belongs to net 2, then when it will create a connection with any other master node,
+// the socket address will be the address that belongs to net 1. Since etcd expects it to be the same as the node-ip, it will
+// reject the connection.
+// A similar situation may happen with tunnel IPs (Geneve for OVN).  When setting tunnel, OVS configuration contains the
+// node-ips as endpoints.  When there is a collision, the receiving endpoint of a tunnel may receive a tunnel packet
+// from an IP address it does expect.  This leads OVS to ignore the packet.  Such situation may occur for master and worker
+// nodes.  If such collision happens for workers, there is a possibility that the cluster will still be functional.
+//
+// This file attempts to solve the issue automatically.
+
+// NodeIpAllocation is the allocation result for specific host
+type NodeIpAllocation struct {
+	NodeIp string
+	HintIp string
+	Cidr   string
+}
+
+// nodeIpCandidate is an assignment candidate which is used when attempting to find optimal solution
+type nodeIpCandidate struct {
+	hostId strfmt.UUID
+	ip     net.IP
+}
+
+// hostNetwork represents discovered network in a host
+type hostNetwork struct {
+	hostId strfmt.UUID
+	ipNet  *net.IPNet
+}
+
+// eligibleNodeIp is an ip with its network which eligible for allocation.  It means the IP is connected to all other hosts
+// and its interface has default route
+type eligibleNodeIp struct {
+	nodeIp      net.IP
+	ipNet       *net.IPNet
+	routeMetric int32
+}
+
+// hostCandidates gathers all eligible node IPs and all the discovered networks in that host
+type hostCandidates struct {
+	hostId          strfmt.UUID
+	hostNetworks    []*hostNetwork
+	eligibleNodeIps []*eligibleNodeIp
+}
+
+// allocationSolution contains all or partial set of the allocation.
+type allocationSolution struct {
+	routeMetric       int32
+	nodeIpAllocations map[strfmt.UUID]*NodeIpAllocation
+}
+
+func getDefaultMetricForInterface(interfaceName string, routes []*models.Route, isIpv4 bool) (bool, int32) {
+	for _, r := range routes {
+		if isIpv4 && r.Family != unix.AF_INET || !isIpv4 && r.Family != unix.AF_INET6 {
+			continue
+		}
+		isDefault, err := IsDefaultRoute(r)
+		if err != nil {
+			continue
+		}
+		if isDefault && r.Interface == interfaceName {
+			return true, r.Metric
+		}
+	}
+	return false, 0
+}
+
+// Create candidates for host.  For an address to be considered as eligible IP, it has to be connected, and its interface must
+// have default route.
+func createHostCandidates(host *models.Host, connectivity *Connectivity, isIpv4 bool) (*hostCandidates, error) {
+	connectedAddresses := connectivity.L3ConnectedAddresses[lo.FromPtr(host.ID)]
+	if len(connectedAddresses) == 0 {
+		return nil, errors.Errorf("no address found for host %s", lo.FromPtr(host.ID))
+	}
+	connectedIps := lo.Map(lo.Filter(connectedAddresses, func(addrStr string, _ int) bool { return isIpv4 == IsIPv4Addr(addrStr) }),
+		func(addrStr string, _ int) net.IP { return net.ParseIP(addrStr) })
+	inventory, err := common.UnmarshalInventory(host.Inventory)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal inventory for host %s", lo.FromPtr(host.ID))
+	}
+	var (
+		eligibleNodeIps []*eligibleNodeIp
+		hostNetworks    []*hostNetwork
+	)
+
+	for _, intf := range inventory.Interfaces {
+		addresses := intf.IPV4Addresses
+		if !isIpv4 {
+			addresses = intf.IPV6Addresses
+		}
+		hasDefaultMetric, metric := getDefaultMetricForInterface(intf.Name, inventory.Routes, isIpv4)
+		for _, addrStr := range addresses {
+			addrIp, ipNet, err := net.ParseCIDR(addrStr)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to parse address %s for host %s", addrStr, lo.FromPtr(host.ID))
+			}
+			hostNetworks = append(hostNetworks, &hostNetwork{
+				hostId: *host.ID,
+				ipNet:  ipNet,
+			})
+			if hasDefaultMetric && lo.ContainsBy(connectedIps, func(ip net.IP) bool { return ip.Equal(addrIp) }) {
+				eligibleNodeIps = append(eligibleNodeIps, &eligibleNodeIp{
+					nodeIp:      addrIp,
+					ipNet:       ipNet,
+					routeMetric: metric,
+				})
+			}
+		}
+	}
+	if len(eligibleNodeIps) == 0 {
+		return nil, errors.Errorf("failed to create candidates for host %s", *host.ID)
+	}
+	return &hostCandidates{
+		hostId:          lo.FromPtr(host.ID),
+		hostNetworks:    hostNetworks,
+		eligibleNodeIps: eligibleNodeIps,
+	}, nil
+}
+
+func createHostsCandidates(hosts []*models.Host, connectivity *Connectivity, isIpv4 bool) (candidates []*hostCandidates, err error) {
+	for _, h := range hosts {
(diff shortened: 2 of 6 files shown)


internal/network/none_platform_node_ips_allocation.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
+	"github.com/thoas/go-funk"
+	"golang.org/x/sys/unix"
+)
+
+// When none platform is in use, if there is ambiguity in node-ip assignment, then incorrect assignment might lead
+// to installation failure.  This happens when etcd detects that the socket address from an etcd node does not match
+// the expected address in the peer certificate. In this case etcd rejects such connection.
+// Example: assuming to networks - net1 and net2.
```
tsorya, line 20: nit: to networks -> two networks

internal/network/none_platform_node_ips_allocation.go
```diff
@@ -0,0 +1,261 @@
...
+	ipnet  *net.IPNet
+}
+
+type candidate struct {
+	nodeIp net.IP
+	cidr   *net.IPNet
+	metric int32
+}
+
+type hostCandidates struct {
```
tsorya, line 52: Hmm, what is hostCanditate? maybe some description should be added?

internal/ignition/ignition.go
```diff
@@ -496,6 +510,16 @@ func (g *installerGenerator) addBootstrapKubeletIpIfRequired(log logrus.FieldLog
 			return envVars, err
 		}
 		envVars = append(envVars, "OPENSHIFT_INSTALL_BOOTSTRAP_NODE_IP="+bootstrapIp)
+	} else if g.nodeIpAllocations != nil {
```
tsorya, line 513: g.nodeIpAllocations will always be part of MNO none platform right? 
In this case i am not sure that comment is right and i think better to change conditions in the function

internal/ignition/ignition.go
```diff
@@ -1530,6 +1554,12 @@ func (g *installerGenerator) writeSingleHostFile(host *models.Host, baseFile str
 			return errors.Wrapf(errP, "Failed to parse machine cidr for node ip hint content")
 		}
 		setFileInIgnition(config, nodeIpHintFile, fmt.Sprintf("data:,KUBELET_NODEIP_HINT=%s", ip), false, 420, true)
+	} else if g.nodeIpAllocations != nil {
```
tsorya, line 1557: can we have a function in common that will return true in case it is none platform multinode installation?

internal/network/none_platform_node_ips_allocation.go
```diff
@@ -0,0 +1,374 @@
...
+// to installation failure.  This happens when etcd detects that the socket address from an etcd node does not match
+// the expected address in the peer certificate. In this case etcd rejects such connection.
+// Example: assuming two networks - net1 and net2.
+// master node 1 has 1 address that belongs to net1.
+// master node 2 has 2 addresses.  one that belongs to net 1, and another that belongs to net 2
+// master node 3 has 1 address that belongs to net 1.
+// If the selected node-ip of master node 2 belongs to net 2, then when it will create a connection with any other master node,
+// the socket address will be the address that belongs to net 1. Since etcd expects it to be the same as the node-ip, it will
+// reject the connection.
+// A similar situation may happen with runnel IPs (Geneve for OVN).  When setting tunnel, OVS configuration contains the
```
tsorya, line 28: runnel -> tunnel

internal/network/none_platform_node_ips_allocation.go
```diff
@@ -0,0 +1,374 @@
...
+// node-ips as endpoints.  When there is a collision, the receiving endpoint of a tunnel may receive a tunnel packet
+// from an IP address it does expect.  This leads OVS to ignore the packet.  Such situation may occur for master and worker
+// nodes.  If such collision happens for workers, there is a possibility that the cluster will still be functional.
+//
+// This file attempts to solve the issue automatically.  Masters failures are considered non-recoverable, and therefore no
+// node-ip allocation will be done for them. For workers we try to find node-ips with the least number of conflicts as possible.
```
tsorya, line 34: You sure we need to handle workers ips at all? Not sure how workers are influenced at all, they don't reach etcd

ori-amizur: The solution is applicable for Geneve tunnel used for cluster networks.

internal/ignition/ignition.go
ori-amizur: What comment ?

internal/common/common.go
```diff
@@ -609,3 +609,7 @@ func IsExternalIntegrationEnabled(platform *models.Platform, platformName string
 func IsOciExternalIntegrationEnabled(platform *models.Platform) bool {
 	return IsExternalIntegrationEnabled(platform, ExternalPlatformNameOci)
 }
+
+func IsNonePlatformCluster(cluster *Cluster) bool {
+	return !IsSingleNodeCluster(cluster) && swag.BoolValue(cluster.UserManagedNetworking)
```
eranco74, line 614: Why exclude SNO? while it's not relevant to the fix in this PR SNO is a none platform cluster.
Consider renaming the function to `IsMultiNodeNonePlatformCluster`

internal/ignition/ignition.go
eranco74, line 513: Log if `g.nodeIpAllocations == nil`

internal/ignition/ignition.go
eranco74, line 1557: This `else if` statement is a bit tricky, Are relaying on `g.nodeIpAllocations` to be nil in case of MNO with platform other than none?
It would be clearer if you check for the case you want to handle (none platform + MNO)

internal/network/none_platform_node_ips_allocation.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+	var (
+		errs                []error
+		allocationSolutions []*allocationSolution
+		h                   = n.candidates[index]
+	)
+	for i := range h.eligibleNodeIps {
+		c := h.eligibleNodeIps[i]
+		localCollisionContext := newCollisionDetectorForNodeIp(h, c)
+		if collisionErrs := localCollisionContext.detectCollisions(parentContext); len(collisionErrs) > 0 {
```
eranco74, line 239: Add logging of the localCollisionContext of each host for debuging

internal/network/none_platform_node_ips_allocation.go
```diff
@@ -0,0 +1,312 @@
...
+	"github.com/pkg/errors"
+	"github.com/samber/lo"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+// When none platform is in use, if there is ambiguity in node-ip assignment, then incorrect assignment might lead
+// to installation failure.  This happens when etcd detects that the socket address from an etcd node does not match
+// the expected address in the peer certificate. In this case etcd rejects such connection.
+// Example: assuming two networks - net1 and net2.
```
eranco74, line 21: What will happen in the following scenario?
assuming two networks - net1 and net2.
// master node 1 has 1 address that belongs to net1.
// master node 2 has 2 addresses.  one that belongs to net 1, and another that belongs to net 2
// master node 3 has 1 address that belongs to net2.

ori-amizur: This is a failure to allocate.
According to my tests, the installation will fail too.
If we want, we can use this as a pre-install validation

internal/common/common.go
ori-amizur: OK

internal/ignition/ignition.go
ori-amizur: In case of failure to allocate like https://github.com/openshift/assisted-service/pull/6257#discussion_r1606317112 the `g.nodeIpAllocations` will be nil.  So we will continue installation as usual.
So in general, in case we have a valid allocation, it should be used.
I will try to make the code clearer.

ori-amizur: Changed to rely also on cluster being multi-node-none-platform-cluster

internal/network/none_platform_node_ips_allocation.go
ori-amizur: Bare in mind that it can be multiple times for the same host

internal/provider/none/installConfig.go
```diff
@@ -5,10 +5,36 @@ import (
...
 	"github.com/openshift/assisted-service/internal/installcfg"
+	"github.com/openshift/assisted-service/internal/network"
 	"github.com/openshift/assisted-service/internal/provider"
 	"github.com/openshift/assisted-service/models"
+	"github.com/samber/lo"
 )
 
+func (p noneProvider) replaceMachineNetworkIfNeeded(cluster *common.Cluster, cfg *installcfg.InstallerConfigBaremetal) {
+	bootstrapHost := common.GetBootstrapHost(cluster)
+	if bootstrapHost != nil {
```
tsorya, line 16: lets change to bootstrap == nil {return}
Should not we return error in this case?

internal/provider/none/installConfig.go
```diff
@@ -5,10 +5,36 @@ import (
...
+		nodeIpAllocations, err := network.GenerateNonePlatformAddressAllocation(cluster, p.Log)
+		if err != nil {
+			p.Log.WithError(err).Warnf("failed to get node ip allocations for cluster %s", *cluster.ID)
+		} else {
```
tsorya, line 20: lets return in case of error and remove else, imo will be easier to read

internal/provider/none/installConfig.go
```diff
@@ -5,10 +5,36 @@ import (
...
+			allocation, ok := nodeIpAllocations[lo.FromPtr(bootstrapHost.ID)]
+			if ok {
```
tsorya, line 22: same as with else, just return in case it is not ok

internal/network/none_platform_node_ips_allocation.go
```diff
@@ -0,0 +1,325 @@
...
+	return result, nil
+}
+
+func GenerateNonePlatformAddressAllocation(cluster *common.Cluster, log logrus.FieldLogger) (map[strfmt.UUID]*NodeIpAllocation, error) {
+	addressFamilies, err := GetClusterAddressFamilies(cluster)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed tp get address families for cluster %s", lo.FromPtr(cluster.ID))
+	}
+	isIpv4 := lo.Contains(addressFamilies, IPv4)
+	isIpv6 := lo.Contains(addressFamilies, IPv6)
```
tsorya, line 295: should not we check ipv6 too? I don't see we the place we create host candidates for ipv6?
What i miss?

ori-amizur: I think we do not support dual stack with IPv6 primary

internal/provider/none/installConfig.go
ori-amizur: I think we should a pre-install validation.
If we add a validation for it, then if we reached here, it means the validation was ignored.  So I don't want to stop installation in this case.

internal/provider/none/installConfig.go
ori-amizur: Same as here https://github.com/openshift/assisted-service/pull/6257#discussion_r1607998519

internal/provider/none/installConfig.go
ori-amizur: OK

ori-amizur: > lets change to bootstrap == nil {return} 

OK

> Should not we return error in this case?

OK.  Will return error

internal/network/none_platform_node_ips_allocation.go
tsorya: We have a validation that stops it? I mean not dual stack installation but ipv6 only when there are ipv4 ips on the node

internal/ignition/ignition.go
tsorya: https://github.com/openshift/assisted-service/pull/6257/files#diff-f0c8175849547a8e75a7442935a2ae99a16a72dfeee8efd9ea663fd2aa3c8c1bR504

internal/provider/none/installConfig.go
tsorya: Not sure i understand, there is not code running if !ok so why not to change the code 
if !ok {
  log.
  return
}
and run code only in case allocation exists?

internal/network/none_platform_node_ips_allocation.go
ori-amizur: In this case `GetClusterAddressFamilies` will return only IPv6.
It relies on ServiceNetworks and ClusterNetworks

internal/provider/none/installConfig.go
ori-amizur: Ahh sorry. I will change.

internal/ignition/ignition.go
ori-amizur: OK

PR #6260, opened 2024-05-05
MGMT-17194: Ensure releaseImage is not tag based when mirror registry enabled.

paul-maidment:
When mirror registries are enabled, it is not possible to use a tag based image (for example quay.io/openshift-release-dev/ocp-release:4.15.0-multi) for the release image of a cluster being deployed by assisted.
The user should use digest a based image (such as quay.io/openshift-release-dev/ocp-release@sha256:b86422e972b9c838dfdb8b481a67ae08308437d6489ea6aaf150242b1d30fa1c)

This PR ensures that we check for this scenario during the validation of the ClusterDeployment.

This will be propagated to the ClusterDeployments_Controller as a validation failure if the image is tag based.
A spec Sync failure will be indicated in the conditions of the controller.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/clusterdeployments_controller.go
+++ b/internal/controller/controllers/clusterdeployments_controller.go
@@ -50,9 +50,11 @@ import (
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/assisted-service/pkg/auth"
 	logutil "github.com/openshift/assisted-service/pkg/log"
+	"github.com/openshift/assisted-service/pkg/mirrorregistries"
 	"github.com/openshift/assisted-service/restapi/operations/installer"
 	operations "github.com/openshift/assisted-service/restapi/operations/manifests"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	imageReference "github.com/openshift/library-go/pkg/image/reference"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
@@ -104,9 +106,10 @@ type ClusterDeploymentsReconciler struct {
 	Manifests        manifestsapi.ClusterManifestsInternals
 	ServiceBaseURL   string
 	PullSecretHandler
-	AuthType              auth.AuthType
-	VersionsHandler       versions.Handler
-	SpokeK8sClientFactory spoke_k8s_client.SpokeK8sClientFactory
+	AuthType                      auth.AuthType
+	VersionsHandler               versions.Handler
+	SpokeK8sClientFactory         spoke_k8s_client.SpokeK8sClientFactory
+	MirrorRegistriesConfigBuilder mirrorregistries.MirrorRegistriesConfigBuilder
 }
 
 const minimalOpenShiftVersionForDefaultNetworkTypeOVNKubernetes = "4.12.0-0.0"
@@ -199,7 +202,7 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 	}
 
 	cluster, err := r.Installer.GetClusterByKubeKey(req.NamespacedName)
-	if err1 := r.validateClusterDeployment(clusterDeployment, clusterInstall); err1 != nil {
+	if err1 := r.validateClusterDeployment(ctx, clusterDeployment, clusterInstall); err1 != nil {
 		log.Error(err1)
 		return r.updateStatus(ctx, log, clusterInstall, clusterDeployment, cluster, err1)
 	}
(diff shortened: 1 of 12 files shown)


pkg/mirrorregistries/generator.go
```diff
@@ -69,6 +70,18 @@ func (m *mirrorRegistriesConfigBuilder) GetMirrorRegistries() ([]byte, error) {
...
 }
 
+// Determines whether or not this installation of assisted is using mirror registries
+func (m *mirrorRegistriesConfigBuilder) HasMirrorRegistries() (bool, error) {
+	hasMirrorRegistries := false
+	if _, err := m.GetMirrorRegistries(); err != nil {
+		if !os.IsNotExist(err) {
+			return false, err
+		}
+		hasMirrorRegistries = true
```
tsorya, line 80: Not sure i understand why in case of error we return true?

pkg/mirrorregistries/generator.go
```diff
@@ -69,6 +70,18 @@ func (m *mirrorRegistriesConfigBuilder) GetMirrorRegistries() ([]byte, error) {
 	return os.ReadFile(m.MirrorRegistriesConfigPath)
 }
 
+// Determines whether or not this installation of assisted is using mirror registries
+func (m *mirrorRegistriesConfigBuilder) HasMirrorRegistries() (bool, error) {
+	hasMirrorRegistries := false
+	if _, err := m.GetMirrorRegistries(); err != nil {
+		if !os.IsNotExist(err) {
+			return false, err
```
tsorya, line 78: maybe just return hasMirrorRegistries ? as it should be false in anycase? just for consistency

pkg/mirrorregistries/generator.go
```diff
@@ -69,6 +70,18 @@ func (m *mirrorRegistriesConfigBuilder) GetMirrorRegistries() ([]byte, error) {
...
+		}
+		hasMirrorRegistries = true
+	}
+	return hasMirrorRegistries, nil
```
tsorya, line 82: How it becomes true in case  m.GetMirrorRegistries() returns content?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -253,14 +256,42 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
...
 
+	// If mirror registries are enabled then we need to make sure that the release image is not tag based.
+	// A tagged release image will be rejected by oc when mirror registry configuration is present.
+	// Mirrored images are typically used in disconnected environments which can have trouble with tag based images.
+	hasMirrorRegistries, err := r.MirrorRegistriesConfigBuilder.HasMirrorRegistries()
+	if err != nil {
+		return errors.Wrap(err, "unable to validate release image due to error determining mirror registry status")
+	}
+	if hasMirrorRegistries && clusterInstall.Spec.ImageSetRef != nil {
+		clusterImageSet := &hivev1.ClusterImageSet{}
```
tsorya, line 275: We have this code in other place, maybe lets create function for it ?

paul-maidment: True that it's very similar to getReleaseImage but as the purpose of this fix is to cater for an issue affecting disconnected environments, there is a possibility that this will not work.

`func (r *ClusterDeploymentsReconciler) getReleaseImage( `

Makes an internal call to fetch further repository information, this is overkill for my use case as I only need the URL.
Also, for the problem I am trying to solve, I feel it will lead to misleading errors that do not advise the user as intended.

https://github.com/openshift/assisted-service/blob/9462a1af97cd3efb009894ce6dd0b75da06e65b9/internal/versions/kube_api_versions.go#L111

pkg/mirrorregistries/generator.go
paul-maidment: I got the order of my true/false mixed up here, good spot!

pkg/mirrorregistries/generator.go
paul-maidment: I mean, that assumes you just want to swallow that error, I don't think that's our decision to make.

What if there is some error, unrelated to the configuration of mirror registries, maybe some disk IO problem or something? We would just say "there are no mirror registries" instead of saying "I encountered an unexpected error"?

I think we are right to handle in the way I have chosen, any errors we do not expect, should be propagated upward and handled.

paul-maidment: OK, I went with your suggestion, though it does mean that we are swallowing errors in some cases.

A problem fetching mirror registries does not necessarily imply that we should treat this scenario as hasMirrorRegistries := false

Saves a few lines of code, looks more elegant but is that really just an illusion?

internal/controller/controllers/clusterdeployments_controller.go
paul-maidment: Done, I refactored.

pkg/mirrorregistries/generator.go
paul-maidment: Just found `r.MirrorRegistriesConfigBuilder.IsMirrorRegistriesConfigured()` which looks a lot more appropriate!, going to scrap this function altogether.

PR #6266, opened 2024-05-06
MGMT-17194: Ensure that we improve error messages for mirror registry tagged images

paul-maidment:
If the user relies on mirror registries, and clusterimageset is set to a tagged image (e.g. quay.io/openshift-release-dev/ocp-release:4.15.0-multi)
as opposed to a by digest image (e.g. quay.io/openshift-release-dev/ocp-release@sha256:b86422e972b9c838dfdb8b481a67ae08308437d6489ea6aaf150242b1d30fa1c)

then oc will fail to pull with:

--icsp-file only applies to images referenced by digest and will be ignored for tags

This PR adds some additional information to the SpecSynced condition of the AgentClusterInstall 
if Mirror Registries is enabled, a tag is detected in the registry reference and there was a problem with `oc`

This should make it clearer to the user that there might be an issue the tagged image reference.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/clusterdeployments_controller.go
+++ b/internal/controller/controllers/clusterdeployments_controller.go
@@ -50,9 +50,11 @@ import (
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/assisted-service/pkg/auth"
 	logutil "github.com/openshift/assisted-service/pkg/log"
+	"github.com/openshift/assisted-service/pkg/mirrorregistries"
 	"github.com/openshift/assisted-service/restapi/operations/installer"
 	operations "github.com/openshift/assisted-service/restapi/operations/manifests"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	imageReference "github.com/openshift/library-go/pkg/image/reference"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thoas/go-funk"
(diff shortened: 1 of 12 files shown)


internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1453,9 +1455,18 @@ func (r *ClusterDeploymentsReconciler) getReleaseImage(
 
 	releaseImage, err := r.VersionsHandler.GetReleaseImageByURL(ctx, clusterImageSet.Spec.ReleaseImage, pullSecret)
 	if err != nil {
+		errMsgSuffix := ""
+		if r.MirrorRegistriesConfigBuilder.IsMirrorRegistriesConfigured() {
+			// There is a special error case where we should check to see if the supplied image is tag based
+			// if so then we need to provide the user with a hint as to what might be happening.
+			ref, err2 := imageReference.Parse(clusterImageSet.Spec.ReleaseImage)
+			if err2 == nil && len(ref.Tag) > 0 {
+				errMsgSuffix = "as the image is tag based and this can cause issues with oc, please take this into consideration"
```
tsorya, line 1464: wondering if message is clear, @eranco74 @omertuc WDYT?

eranco74: Consider changing it to:
```suggestion
				errMsgSuffix = "as the image is tag based and this can cause issues with oc, please use image digest instead"
```

omertuc: `oc` is an implementation detail, our users should not be aware of it, so we should not mention it

Users also might not be familiar with pull-spec terminology such as "tag" so I would help them out with an example, by including the tag itself in the error message. So maybe something like:;

`fmt.Sprintf("Release image %q uses a tag (%q), this is usually unsupported in combination with mirror registries, please provide a digest-based image instead", clusterImageSet.Spec.ReleaseImage, ref.Tag)`

Feel free to drop "usually" if you think it's not a good fit, I'm not familiar with the technical details here so I'm not sure

PR #6290, opened 2024-05-08
MGMT-17771: Adds enhancement for FIPS with multiple RHEL installer versions

carbonin:
In order for an OpenShift cluster to be considered FIPS compliant the installer must be run on a system with FIPS compliant openssl libraries. This means using a dynamically linked `openshift-install` binary against the openssl libraries present on our container image. Today this is not a problem because all `openshift-install` binaries in use have been expecting to link to RHEL 8 based openssl libraries, but now OpenShift 4.16 will ship and installer that requires RHEL 9 libraries.

This will require assisted-service to maintain a way to run the `openshift-install` binary in a compatible environment for multiple openssl versions.

https://issues.redhat.com/browse/MGMT-17771

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+Shipping a new image is a non-trivial process. This may take more time than we
+have to set up. We could likely get away with using the existing
+assisted-service image with a different entrypoint for one of the runner images,
+but that still requires us to publish a new one for the architecture
+assisted-service will not be using.
+
+## Design Details [optional]
+
+The new runner containers will expose a HTTP server listening only for local
+connections. assisted-service will POST to this server when it needs manifests
```
mhrivnak, line 85: Speaking of local connections, it's worth considering a unix socket instead of the network stack. We did that for communicating with [ansible-runner](https://github.com/ansible/ansible-runner) from an operator process. It's a simpler approach when all communication is local; no TLS, no auth, no DNS, no network protocols, ...

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Risks and Mitigations
+
+Shipping a new image is a non-trivial process. This may take more time than we
+have to set up. We could likely get away with using the existing
+assisted-service image with a different entrypoint for one of the runner images,
+but that still requires us to publish a new one for the architecture
+assisted-service will not be using.
+
+## Design Details [optional]
```
mhrivnak, line 82: This service closely matches what we need to implement. It could almost be borrowed as-is, mostly modifying what action it takes when receiving a request. Its purpose is to receive events from ansible-runner on a local unix socket shared between the containers.

https://github.com/operator-framework/ansible-operator-plugins/blob/main/internal/ansible/runner/eventapi/eventapi.go

mhrivnak (changes requested):

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+- Allow for FIPS compliant clusters to be installed from the SaaS offering or
+  the on-prem offering
+
+### Non-Goals
+
+- Changing cluster install interfaces to accommodate new FIPS requirements
+  should be avoided
+
+- Dynamically determining a given release's RHEL version. The RHEL version will
+  be a part of configuration tied to the minor release
```
CrystalChun, line 44: Out of curiosity, how will this look? Will it be an index we maintain like the default_os/release images files?

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Proposal
+
+Two additional containers will run alongside assisted-service in the same pod.
+These "installer-runner" containers will, when triggered by assisted-service,
```
CrystalChun, line 49: Are these containers expected to always run? Like even during non-fips mode?

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+run a specified installer binary to generate the manifests required for a
+particular install. These manifests will then be uploaded to whatever storage
+is in use for this deployment (local for on-prem, or s3 for SaaS) and
+assisted-service will take over as usual from there.
+
+Each of these containers should maintain its own cache of installers as
+assisted-service does now and should be built with the required packages to run
```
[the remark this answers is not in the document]
CrystalChun: I would guess the `openshift-client` `oc` CLI would? Or is that something we'd extract from the release image specified?

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Summary
+
+In order for an OpenShift cluster to be considered FIPS compliant the installer
+must be run on a system with FIPS compliant openssl libraries. This means using
+a dynamically linked `openshift-install` binary against the openssl libraries
+present on our container image. Today this is not a problem because all
+`openshift-install` binaries in use have been expecting to link to RHEL 8 based
+openssl libraries, but now OpenShift 4.16 will ship and installer that requires
+RHEL 9 libraries.
```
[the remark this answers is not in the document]
carbonin: >The second sentence seems to imply that there might be a statically-linked binary available, or that it's an option, but is that the case?

Yes, there is. Actually I'm working on a bug right now that is solved by using the `openshift-install` binary from 4.16. https://issues.redhat.com/browse/OCPBUGS-33227

That said, it is that simple. I just figured it would be easier to do this split in any case.
If we also want to consider the static binary for non-fips cases we just add complexity (3 cases instead of just 2)

docs/enhancements/multi-rhel-fips.md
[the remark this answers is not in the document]
carbonin: I'm not actually sure. I'll add this to open questions, but I'd expect the required libraries it to be documented with the installer somewhere.

docs/enhancements/multi-rhel-fips.md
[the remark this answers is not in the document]
mhrivnak: Where would someone obtain a statically vs dynamically linked build?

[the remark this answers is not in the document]
carbonin: As of https://issues.redhat.com/browse/CORS-3418 `openshift-install` is statically linked and as of https://issues.redhat.com/browse/CORS-3419 it supports baremetal installs.

`openshift-baremetal-install` is now only intended for use with FIPS use-cases as that's the only time you'd need a dynamically linked binary. https://issues.redhat.com/browse/WRKLDS-1171 added a new command to `oc release extract` so it could be referenced as `openshift-install-fips` which more accurately represents its purpose.

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+creation-date: 2024-05-07
+last-updated: 2024-05-08
+---
+
+# Support FIPS for installers built for different RHEL releases
+
+## Summary
+
+In order for an OpenShift cluster to be considered FIPS compliant the installer
+must be run on a system with FIPS compliant openssl libraries. This means using
```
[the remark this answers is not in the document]
carbonin: Yes, but that's not really relevant for this topic. Do you still want it added?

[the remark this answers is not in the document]
mhrivnak: Without fips mode enabled, the binary won't use openssl at all, so I think it's a relevant detail to include.

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+## Design Details [optional]
+
+The new runner containers will expose a HTTP server listening only for local
+connections. assisted-service will POST to this server when it needs manifests
+generated. The runner container will respond with any error that occurred while
+generating the manifests.
+
+### Open Questions
+
+1. What does the API look like for the runner containers? What data should be
```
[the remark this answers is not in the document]
carbonin: I think the cache will be managed by the installer runner container and this container will also upload the resulting artifacts to the required location (either the shared data volume or s3).

I don't expect the operation to take too long. I think we could to a single http request for this. Even if we need to tweak the timeout parameters I'd rather that than build an async system.

docs/enhancements/multi-rhel-fips.md
carbonin: Yup, I agree. I think a socket would work well for this.

docs/enhancements/multi-rhel-fips.md
carbonin: Yup, looks about right to me, thanks for the reference

docs/enhancements/multi-rhel-fips.md
carbonin: I was thinking that, but now I'm rethinking it.

Maybe it would be better to just switch on if the version is 4.16 or later.
@mhrivnak mentioned OKD in this context. What do the cluster versions look like when OKD is in use?

docs/enhancements/multi-rhel-fips.md
carbonin: That's the idea, yeah.
My intention is to use these containers for both fips and non-fips cases as well since there shouldn't really be a difference and if we were to only use them for fips they would get tested very infrequently.

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+assisted-service will not be using.
+
+## Design Details [optional]
+
+The new runner containers will expose a HTTP server listening only for local
+connections. assisted-service will POST to this server when it needs manifests
+generated. The runner container will respond with any error that occurred while
+generating the manifests.
+
+### Open Questions
```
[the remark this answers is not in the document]
carbonin: I briefly called this out in `New Images` under `Implementation Details`, but if you think it would be good to put somewhere else or just talked about more prominently then we can do that.

I don't think it's an open question though. For this pass I'm going to create the service in `assisted-service` using the existing code to run the installer. Extracting and splitting everything up would definitely come later.

docs/enhancements/multi-rhel-fips.md
mhrivnak: Image list trimmed:

```
oc adm release info quay.io/openshift/okd@sha256:46b462be1e4c15ce5ab5fba97e713e8824bbb9f614ac5abe1be41fda916920cc
Name:           4.15.0-0.okd-2024-03-10-010116
Digest:         sha256:46b462be1e4c15ce5ab5fba97e713e8824bbb9f614ac5abe1be41fda916920cc
Created:        2024-03-10T06:53:44Z
OS/Arch:        linux/amd64
Manifests:      707
Metadata files: 1

Pull From: quay.io/openshift/okd@sha256:46b462be1e4c15ce5ab5fba97e713e8824bbb9f614ac5abe1be41fda916920cc

Release Metadata:
  Version:  4.15.0-0.okd-2024-03-10-010116
  Upgrades: <none>
  Metadata:

Component Versions:
  kubernetes 1.28.7        
  machine-os 39.20240210.3 Fedora CoreOS
```

carbonin: So we'd still get `4.15.0-0.okd-2024-03-10-010116` in this case.

I mean I suppose if no one is doing FIPS with OKD then we should be able to apply the same rule and it will work out. (4.16 OKD would go to the el9 runner)

romfreiman: why not always use fips compliant installer?

docs/enhancements/multi-rhel-fips.md
eranco74, line 49: Triggered how?

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+This is a complicated change in architecture something simpler might be more
+desirable.
+
+Creating two additional containers in a pod makes the assisted service more
+expensive to scale.
+
+Creating, maintaining, and releasing additional images is a non-trivial amount
+of additional work.
+
+## Alternatives
```
romfreiman, line 119: if we're saying that <4.16 is not fips complaint anyhow, can we say that for those releases we gonna use the statically linked one and for >=4.16, dynamically linked with rhel9? Can we use single existing container then?

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+
+Hive is investigating using the container image from the release payload to run
+the installer. This seems possible, but doesn't make much sense if we want to
+run a persistent service rather than a job per cluster. Running a job per
+cluster wouldn't scale particularly well and would also be an even larger
+architectural change.
```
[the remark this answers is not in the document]
eranco74: I don't think we support assisted-installer on podman.
Also at some point podman will get it: https://github.com/containers/podman/issues/17011.

The assisted-service can still be persistent it can use k8s jobs for executing the openshift-installer create-ignition manifests commands.
Why using a job is considered "a larger architectural change" ?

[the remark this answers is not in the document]
mhrivnak: We don't offer normal product support for it. But as a project: it works, we invite people to try it, and people use and depend on it.

docs/enhancements/multi-rhel-fips.md
mhrivnak: I don't think we're saying that. OpenShift can be installed in FIPS mode today, and that's been the case for a while.

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,137 @@
...
+
+FIPS compliance is important for our customers and assisted-service should be
+able to install FIPS compliant clusters.
+
+### Goals
+
+- Allow for a single installation of assisted-service to install FIPS-compliant
+  clusters using `openshift-install` binaries built against RHEL 8 or RHEL 9
+
+- Allow for FIPS compliant clusters to be installed from the SaaS offering or
```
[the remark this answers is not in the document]
mhrivnak: Yes.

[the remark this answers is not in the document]
gamli75: we have an open task for app-sre https://issues.redhat.com/browse/SDE-3692

docs/enhancements/multi-rhel-fips.md
romfreiman: in ACM/SaaS world? or in general?

romfreiman:
What about ABI? Wont it be affected by this change?

docs/enhancements/multi-rhel-fips.md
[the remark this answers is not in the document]
carbonin: Right, also this comment was more about the solution allowing for this possibility not necessarily that, once implmented, we will be able to install FIPS from the SaaS. The main goal is to make FIPS work for both rhel8 and rhel9 installer releases.

docs/enhancements/multi-rhel-fips.md
carbonin: We will, but which FIPS compliant installer we use (rhel8 or rhel9) will be determined by the release version.

docs/enhancements/multi-rhel-fips.md
carbonin: Plan right now is for an http API, I'll add it here.

docs/enhancements/multi-rhel-fips.md
carbonin: In general and in ACM IIUC

docs/enhancements/multi-rhel-fips.md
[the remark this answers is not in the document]
carbonin: >Why using a job is considered "a larger architectural change" ?

I suppose it's not that much larger really, but it makes the runtime more complicated.
Running jobs over the kube API feels heavier to me than making an http call to a persistently running container in the same pod (effectively an RPC).

carbonin:
>What about ABI? Wont it be affected by this change?

ABI doesn't actually have this problem since they build assisted-service on the base image that corresponds with the installer, but making this change might affect them if we don't also continue to support the current way we run the installer (in the assisted-service container).

Also it depends on what you mean by "affect". If, after this change, they also deploy assisted as described in the enhancement (with the sidecar containers) then things should just work for them, but this will affect them in that they will likely need to update the deployment code for their assisted-service instance.

cc @bfournie @pawanpinjarkar @zaneb

docs/enhancements/multi-rhel-fips.md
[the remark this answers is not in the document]
mhrivnak: It would be more difficult to arrange for a shared filesystem with a Job.

You also have to account for all of the things that could happen when scheduling the job, including that it might end up on a node that fails, or on a node that gets drained right as it starts, etc. It would definitely require an asynchronous approach to watching and waiting for the work to be completed.

docs/enhancements/multi-rhel-fips.md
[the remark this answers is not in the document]
zaneb: The three parts of the problem are:

1. you must use the dynamically-linked binary (`openshift-baremetal-install`/`openshift-install-fips`) b/c sometimes it will need to do FIPS
2. if you use a dynamically linked library, you must use a RHEL 9 userspace from 4.16 and may use one for pre-4.16
3. but if you want to do FIPS pre-4.16 you must use a RHEL 8 userspace

docs/enhancements/multi-rhel-fips.md
zaneb, line 119: Have you considered installing the RHEL 8 OpenSSL 1.1 into the RHEL 9 container alongside OpenSSL 3.0?
(Note: I haven't tried this to see what conflicts.)

zaneb (commented):
> What about ABI? Wont it be affected by this change?

We'll want to keep the same implementation, so we'll likely need to deploy the other container in ABI as well.
The actual issue doesn't affect ABI because we only deploy one version of OpenShift from each assisted-service, so we make sure it always matches.

docs/enhancements/multi-rhel-fips.md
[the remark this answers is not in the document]
carbonin: Also regarding the storage side of this.

The cache and the place where the generated files are stored after the installer is run is already a volume (specifically a PV) and this is what I intended to use for the shared storage, so I think that's a non-issue.

My question here was what exactly will we need to pass over the API call.

docs/enhancements/multi-rhel-fips.md
carbonin: I considered this, but I'm a bit worried about landing this in time and this approach feels like it has more unknown potential pitfalls than the approach described here.

mhrivnak: I did bring up the alternative, which should probably be added to this doc, of extracting an entire rhel 8 userspace within a rhel 9 userspace, and using chroot to run an installer that needs rhel 8. But that brings other complexities around image build and management that probably aren't worthwhile compared to just running two copies of a small process in two different container images.

romfreiman: I don't need chroot. We can use ldpreload env to point to the relevant path when installing <4.16

carbonin: Is it possible that this could be as simple as installing a separate ssl version somewhere else and setting some env vars when running the installer?

I can give that a try if there's a chance it might work.

Will ssl be the only library I need to override in this way? Everything else will be backward compatible? Will this actually be FIPS compliant? Are there any other requirements we would also need to satisfy?

carbonin: Running a rhel8 container (our current released image) the files provided by the older openssl rpm are these:

```
bash-4.4$ rpm -ql openssl-libs-1.1.1k-12.el8_9.x86_64
/etc/pki/tls
/etc/pki/tls/certs
/etc/pki/tls/ct_log_list.cnf
/etc/pki/tls/misc
/etc/pki/tls/openssl.cnf
/etc/pki/tls/private
/usr/lib/.build-id
/usr/lib/.build-id/14
/usr/lib/.build-id/14/0abbb0c09726652dd61128b74b8bb5010f5542
/usr/lib/.build-id/42
/usr/lib/.build-id/42/9301d8c47d78f29d7a39fe01d8076e7b656e4e
/usr/lib/.build-id/6f
/usr/lib/.build-id/6f/619a1566052cd522eb945d91197ad60852a8f8
/usr/lib/.build-id/bb
/usr/lib/.build-id/bb/0f1c2857f6f9e4f68c7c5dc36a41c441318eca
/usr/lib/.build-id/e1
/usr/lib/.build-id/e1/1d63787a0230c919507d7ebc73e8af7e8e8a2b
/usr/lib64/.libcrypto.so.1.1.1k.hmac
/usr/lib64/.libcrypto.so.1.1.hmac
/usr/lib64/.libssl.so.1.1.1k.hmac
/usr/lib64/.libssl.so.1.1.hmac
/usr/lib64/engines-1.1
/usr/lib64/engines-1.1/afalg.so
/usr/lib64/engines-1.1/capi.so
/usr/lib64/engines-1.1/padlock.so
/usr/lib64/libcrypto.so.1.1
/usr/lib64/libcrypto.so.1.1.1k
/usr/lib64/libssl.so.1.1
/usr/lib64/libssl.so.1.1.1k
/usr/share/licenses/openssl-libs
/usr/share/licenses/openssl-libs/LICENSE
```

I suppose I could use a multi-stage build to copy those files into a special directory on the final (rhel9) image then try use `LD_PRELOAD` to push those library versions instead of the ones installed on the system.

I'd need a fully FIPS enabled environment to test if it's working though.

zaneb: `LD_PRELOAD` (or rather, `LD_LIBRARY_PATH` which I think is what you're going for here) is not really the issue - libssl.so.1.1 and libssl.so.3.0 are different major versions of the shared library, so AIUI the binary will know to load the right one. This issue is what _other_ files the RPM installs that might conflict. Mostly just those .cnf files by the looks of it?

carbonin: I'm not sure you can install different versions of the same RPM on the same system.
Also I don't know if the older ssl package is even available in the el9 repos.

My idea was to use a multi-stage container build, install the ssl rpm on an el8 image they copy the files directly into a directory in later stage. Then when we know we need to run the el8 binary we'd set whatever envs we need to allow it to find the old so.

carbonin: Hopefully I'll have an environment to test this in this morning and then if it works I'll run it by the FIPS experts to figure out if this process would be considered compliant.

wking: Sources for "OpenShift can be installed in FIPS mode today" include [4.12 docs for FIPS installation][1].  And [ROSA FedRAMP docs][2]:

> ROSA only uses FIPS-validated modules to process cryptographic libraries.

[1]: https://docs.openshift.com/container-platform/4.12/installing/installing-fips.html
[2]: https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-rosa.html

docs/enhancements/multi-rhel-fips.md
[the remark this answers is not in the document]
wking: [Here's the cluster-version operator using Jobs to pull down release images][1], and it includes asynchronous watching (which you need for installer runs anyway?) and a shared volume (for getting data back from the Job's Pod to the Job-launching container).  Our volume-sharing is [`hostPath`][2], so we schedule to the same node where the Job-launching container is running, although obviously you could use a cloud-ier volume if you didn't want that requirement.  

[1]: https://github.com/openshift/cluster-version-operator/blob/0b3f507632ce4705702fc725614bd22d25d6686c/pkg/cvo/updatepayload.go#L186-L339
[2]: https://kubernetes.io/docs/concepts/storage/volumes/#hostpath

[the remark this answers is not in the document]
mhrivnak: Thanks for the suggestions. We have multiple non-k8s runtime environments that need to keep working, so that's a limiting factor.

We're only using the installer binary to generate artifacts, so we do not need to do asynchronous monitoring of its work or progress. We're not using it to carry out actual provisioning.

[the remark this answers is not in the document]
carbonin: If we do go with `Jobs` then both ABI and podman would either need to stick with the current method of running the installer on the service container (which might be fine for ABI since it will always be using the matching image, and podman because we'll never support a FIPS cluster that way) or find some other way to spin up the image from the release to run the installer. In any case we end up with needing to support and maintain two methods for running the installer which isn't ideal.

I'm also worried about just how many jobs we might end up running at once.
I don't know exactly what the peak number of installs at a time is on the SaaS, but I know it's at least several hundred simultaneously and several thousand in a few hours when the scale team is testing the on-prem deployment through MCE, @akrzos likely has those numbers.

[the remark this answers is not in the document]
mhrivnak: While we don't offer normal product support for the podman runtime method, I don't see a reason why someone couldn't use it on a FIPS-enabled system to install a FIPS-enabled cluster.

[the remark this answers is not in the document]
zaneb: > While we don't offer normal product support for the podman runtime method, I don't see a reason why someone couldn't use it on a FIPS-enabled system to install a FIPS-enabled cluster.
I don't disagree, but the question is more whether we need the upstream project to support installing both pre-4.16 and post-4.16 clusters with FIPS from the same podman deployment and not ask the user to create separate podman deployments. That sounds like the kind of requirement that products have but which upstream projects can dispense with.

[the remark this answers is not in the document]
carbonin: For what it's worth I added some more detail about these alternatives to the actual enhancement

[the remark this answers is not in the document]
mhrivnak: We don't build a different assisted-installer service to install one version of openshift vs. another. It's just one service that theoretically covers them all.

In theory we could make two parallel builds available and tell users to run one stack for version <=4.15 and the other for versions >4.16, but that seems like more work than just solving the problem properly for everyone whether they're running the upstream or downstream build.

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,217 @@
...
+It may be possible to install both versions of the shared libraries required by
+the installers (libcrypto and libssl?) for FIPS compliance on a single image.
+This would require much less change and should be significantly quicker to
+implement, but it's not clear if these would be possible or supportable.
+This could be achieved by any of the following methods:
+
+1. Create a separate userspace for el8 libraries and chroot when those libraries
+   are required.
+  - This seems a bit complicated and it will likely make our image quite a bit
+    larger than it already is (~1.3G).
```
carbonin, line 208: I tried this out and was able to get both installers to run in a container on a FIPS-enabled RHEL9 host using `chroot` for the rhel8 one. I only needed to copy `/usr` from the el8 image and create the top-level symlinks in the chroot target dir. Based on my tests with the centos images this adds about 200M to the image size.

I still need to get assurance that this would actually result in a FIPS-compliant cluster, but if it does it seems like the most promising option to me.

I think this is worth a POC. @mhrivnak @romfreiman What do you think?

carbonin: Asked about this in the FIPS-related internal channels and it turns out nothing like this would be considered FIPS-compliant so this disqualifies any approach that doesn't run the installer on a base container image that it is expecting to run on.

I'll add this to the enhancement text.

romfreiman: chroot :)

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,244 @@
...
+    only different in version.
+3. Use multi-layer container builds to copy the libraries from an el8 image to a
+   directory on the el9 image and use `LD_PRELOAD` or manipulate `LD_LIBRARY_PATH`
+   to point the el8 installer binaries to the correct libraries.
+
+The approach using chroot worked, but FIPS SMEs said that the container base
+image *must* match the installer for the resulting cluster to be considered
+FIPS-compliant so none of these multi-library options are valid.
+
+### Publish multiple assisted-service images
```
carbonin, line 220: @romfreiman does this look like what you were talking about?
Tried to capture it here.

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,244 @@
...
+
+It's likely that a user installing in FIPS mode will only be interested in
```
mhrivnak, line 222: Why do we think this is true?

Everyone starts with a single version. But existing clusters and their use cases lag behind new clusters and new use cases. I would expect that most users end up with a mix of openshift versions over time.

Keep in mind that users need the ability to re-provision their clusters at any time, for recovery from all kind of unexpected events. And a growing number of customers depend on regular "cluster as a service" provisioning. Telling such a user that they can only (re-)provision a subset of openshift versions would generally be a big disadvantage.

Even a large-scale user with relatively uniform versions is not going to upgrade all of their clusters at once. It would not be ideal to put them in a position where they can only provision clusters for the newest openshift versions in their fleet.

Is there reason to think that a FIPS user would have different patterns and expectations?

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,244 @@
...
+installing a single OCP version at a time. This means that a given version of
+assisted will need to still support both el8 and el9 FIPS installs, but a single
+deployment of assisted would not.
+
+To handle this the assisted-service image would be built twice; once based on
+el8 and again based on el9. Both images would be released and the operator would
+choose which to deploy based on configuration (likely an annotation since a more
+robust solution would be preferred in the future).
+
+For example, in the case that a user knew they wanted to deploy OCP 4.14 from a
```
mhrivnak, line 232: This assumes that the user who owns the configuration of the management cluster itself is the same user who is provisioning clusters. But often that role is separate. The "Cluster Creator" would need to work with the cluster admin to re-configure assisted-installer to switch it from el8 mode to el9 mode. Then what if they need to switch back? This could become a real hassle.

It would help to have automation that recognizes when the config doesn't work for a desired install operation, but resolving that could be a pain. Even for cases where it is the same person doing both personas, it's inconvenient.

docs/enhancements/multi-rhel-fips.md
carbonin: You're right that this is an assumption on my part, but my experience with even moderately conservative customers is that they have a single version of OCP that is vetted and allowed in the org (at least in production).

My guess was that any user interested in full FIPS compliance would be even more strict than that, but I have no actual users to back any of this up.

docs/enhancements/multi-rhel-fips.md
carbonin: This is already something that the application admin is choosing to a degree when they set the available OS images, so I don't think it's too much of a stretch for them to also communicate which OCP versions will be in use.

docs/enhancements/multi-rhel-fips.md
carbonin: As for upgrade and reprovisioning. I do consider this option a temporary solution that buys us time to properly think through and vet something that would work more generically. Ideally in the version following the one this alternative would be implemented in we'd implement the main body of this enhancement (or something similar).

mhrivnak: They start with a single vetted version, then that progresses over time, but upgrades lag.

I think it's still better to implement this enhancement as described, which would fully preserve the existing use cases and user experience, and then optimize it later as needed.

docs/enhancements/multi-rhel-fips.md
mhrivnak: Fair point. So then the time will come when the cluster creator asks the admin to add a new version that requires removal of all other versions. Then they'll have to negotiate an upgrade plan with all stakeholders to get the whole fleet safely over the hump, etc etc. Doesn't that seem like a pain? If we can avoid creating that burden for our customers, we'll all be better-off.

docs/enhancements/multi-rhel-fips.md
romfreiman: @mhrivnak all good. can we find such customer now? we'll have it as an rfe and improve in the next release. I dont think we should restructure the whole service 2 weeks before the release.
for next release, lets reconsider.

docs/enhancements/multi-rhel-fips.md
```diff
@@ -0,0 +1,244 @@
...
+- install config
+- release image
+
+API call output:
+- Appropriate HTTP response
+- Error message if the call was not successful
+
+### Installer Cache
+
+The installer cache directory will be shared (as it's currently on the PV), but
```
paul-maidment, line 113: "as it's currently on the PV" <- Do you mean that the installer cache is on a PV in non SAAS environments.
As far as I am aware, SAAS does not use a PV.
Could you clarify what you mean by this?

PR #6304, opened 2024-05-09
OCPBUGS-33227: Use openshift-install binary for releases >= 4.16

carbonin:
In 4.16 the installer binaries started to be built against EL9 libraries rather than EL8. Additionally support for baremetal installs was added to the `openshift-install` binary which was also made statically linked.

This means that using `openshift-baremetal-install` from a 4.16 release on an EL8-based image will cause errors like:

```
Failed to prepare the installation due to an unexpected error: failed generating install config for cluster 47ff23ae-012a-421e-89a9-8ae1ca04a67f: error running openshift-install manifests, /data/install-config-generate/installercache/quay.io/openshift-release-dev/ocp-release:4.16.0-rc.0-x86_64/ln_1715091595_openshift-baremetal-install: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /data/install-config-generate/installercache/quay.io/openshift-release-dev/ocp-release:4.16.0-rc.0-x86_64/ln_1715091595_openshift-baremetal-install)
```

To address this, this commit moves installs for clusters >= 4.16 to use `openshift-install` which will work correctly regardless of the container base image it is run on. This is only possible for 4.16 and greater releases because the changes that allow `openshift-install` to do baremetal installations (and to statically link it) were only made for 4.16 and will not be backported to older releases.

This is a temporary fix as installing FIPS compliant clusters will still require using `openshift-baremetal-install` so the ultimate goal is to be able to run `openshift-baremetal-install` for any supported OCP version.

Note that this is only relevant for the case where assisted-service is an el8-based image, which is not the case currently. This PR is mostly meant for backport to earlier versions which will still need to install 4.16.

Related to https://issues.redhat.com/browse/CORS-3024
Resolves https://issues.redhat.com/browse/OCPBUGS-33227

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

Deployed manually and installed both 4.15 and 4.16 SNO successfully.
Also provided the patched image to telco QE for testing.

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/oc/release.go
```diff
@@ -343,19 +343,35 @@ func (r *release) Extract(log logrus.FieldLogger, releaseImage string, releaseIm
...
 
-func (r *release) GetReleaseBinaryPath(releaseImage string, cacheDir string) (workdir string, binary string, path string) {
-	workdir = filepath.Join(cacheDir, releaseImage)
+func (r *release) GetReleaseBinaryPath(releaseImage string, cacheDir string, ocpVersion string) (workdir string, binary string, path string, err error) {
 	binary = "openshift-baremetal-install"
+
+	// use the statically linked binary for 4.16 and up since our container is el8
+	// based and the baremetal binary for those versions is dynamically linked
+	// against el9 libaries
+	staticLinkingRequiredVersion := version.Must(version.NewVersion("4.16.0-ec.6"))
```
tsorya, line 352: lets make 4.16.0-ec.6 consts with name like MinVersion....

internal/installercache/installercache_test.go
```diff
@@ -39,29 +39,29 @@ var _ = Describe("installer cache", func() {
...
 		mockRelease.EXPECT().GetReleaseBinaryPath(
-			gomock.Any(), gomock.Any()).
-			Return(workdir, releaseID, fname)
+			gomock.Any(), gomock.Any(), version).
+			Return(workdir, releaseID, fname, nil)
 		mockRelease.EXPECT().Extract(gomock.Any(), releaseID,
-			gomock.Any(), cacheDir, gomock.Any()).
-			DoAndReturn(func(log logrus.FieldLogger, releaseImage string, releaseImageMirror string, cacheDir string, pullSecret string) (string, error) {
+			gomock.Any(), cacheDir, gomock.Any(), version).
+			DoAndReturn(func(log logrus.FieldLogger, releaseImage string, releaseImageMirror string, cacheDir string, pullSecret string, versoin string) (string, error) {
```
dgonyier, line 51: @carbonin is there a typo in the last argument?

internal/oc/release.go
```diff
@@ -343,19 +343,35 @@ func (r *release) Extract(log logrus.FieldLogger, releaseImage string, releaseIm
...
+	v, err := version.NewVersion(ocpVersion)
```
tsorya, line 353: why not to provide it as version object? Just to remove error handling from here? WDYT?

internal/oc/release.go
```diff
@@ -343,19 +343,35 @@ func (r *release) Extract(log logrus.FieldLogger, releaseImage string, releaseIm
 	return path, err
 }
 
-func (r *release) GetReleaseBinaryPath(releaseImage string, cacheDir string) (workdir string, binary string, path string) {
-	workdir = filepath.Join(cacheDir, releaseImage)
+func (r *release) GetReleaseBinaryPath(releaseImage string, cacheDir string, ocpVersion string) (workdir string, binary string, path string, err error) {
```
tsorya, line 346: I wonder maybe we can use https://github.com/openshift/assisted-service/pull/6304/files#diff-3b8a020ef087b81b6b45d56bef6dd8d32fcf9d50dfe3d91fd369755b253cfff8R299 
in order to get release version?

carbonin: I thought about that, but this is called pretty frequently and I didn't want to be running `oc ...` every time

internal/oc/release.go
carbonin: Not sure it makes a difference. Then all the callers would have to do the same handling.

internal/oc/release_test.go
```diff
@@ -605,6 +605,34 @@ var _ = Describe("getIcspFileFromRegistriesConfig", func() {
...
+	})
+
+	It("returns the openshift-install binary for 4.16 pre release", func() {
+		_, bin, _, err := r.GetReleaseBinaryPath("image", "dir", "4.16.0-ec.6")
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(bin).To(Equal("openshift-install"))
+		_, bin, _, err = r.GetReleaseBinaryPath("image", "dir", "4.16.0-rc.0")
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(bin).To(Equal("openshift-install"))
+	})
```
lack, line 633: Should we future-proof this by adding a test case for 4.17 as well?

carbonin: Done

carbonin:
/test edge-e2e-metal-assisted

Operator didn't become ready, not an issue with this PR

carbonin:
/hold

Looks like this may indeed be causing the test failure.

carbonin:
/unhold

It wasn't causing the test failure. It's https://issues.redhat.com/browse/OCPBUGS-33493

carbonin:
Spoke with @gamli75 and we're going to override to get this one in since it's blocking telco QE and they already tested the fix an verified it works.

/override ci/prow/edge-e2e-metal-assisted

zaneb:
@carbonin doesn't this break FIPS on the agent-based installer?
I don't understand how this passed CI, because we should be testing with FIPS enabled.

PR #6305, opened 2024-05-09
MGMT-17618: NMStateConfig interfaces presence should be validated

ori-amizur:
In case there is static networking configuration without mac-interface mapping, it is created but the configuration is not applied.

This change validates that for every ethernet interface there is a corresponding mac-interface mapping when the static network configuration is set during infa-env registration or during infra-env update.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @gamli75

### DIFF
--- a/pkg/staticnetworkconfig/generator_test.go
+++ b/pkg/staticnetworkconfig/generator_test.go
@@ -2,6 +2,7 @@ package staticnetworkconfig
 
 import (
 	"encoding/json"
+	"fmt"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


carbonin (commented):
I'm a bit worried that parsing out the nmconnection file is a bit fragile.

I was imagining that we just ensure there is _something_ in the `HostStaticNetworkConfig.MacInterfaceMap`. Is there a use case for that being empty?

ori-amizur:
> I'm a bit worried that parsing out the nmconnection file is a bit fragile.

We are already doing it here: https://github.com/openshift/assisted-service/blob/904ee715faa05ede09fad39718fc64ece60a18d0/pkg/staticnetworkconfig/generator.go#L139-L165 - so it isn't very different.

> I was imagining that we just ensure there is _something_ in the `HostStaticNetworkConfig.MacInterfaceMap`. Is there a use case for that being empty?

I don't think there is use case for being empty - but this is a stronger validation. It validates that for every ethernet interface this is a corresponding mac-mapping.

carbonin:
>so it isn't very different. So if it was OK there, then it is OK here

Fair enough.

Are you sure that this isn't going to cause issues for bonds/vlans? Do we have a test anywhere for this?
Are those files formatted differently such that this would fail or not validate them?

ori-amizur:
> Are you sure that this isn't going to cause issues for bonds/vlans? 
It should be tested - like any other change.

> Do we have a test anywhere for this?
We have a test for bond.  We don't have a test for vlan yet.

> Are those files formatted differently such that this would fail or not validate them? 
They all look similar.  They all have a connection section with `interface-name` and `type`.
The unit-test already contain bond tests.  I can add also vlan tests.

pkg/staticnetworkconfig/generator_test.go
```diff
@@ -51,6 +52,234 @@ var _ = Describe("generateConfiguration", func() {
...
+
+[ipv4]
+method=disabled
+
+[ipv6]
+method=disabled
+
+[vlan]
+id=101
+parent=eth1
```
paul-maidment, line 242: Is it a valid scenario to reference a physical interface that does not exist when creating a vlan connection?

ori-amizur: Seems valid comment.  I will change that at least one interface will be matched.

paul-maidment (approved):
/lgtm

PR #6306, opened 2024-05-10
[WIP] Remove provisioning paused annotation

carbonin:
This was added to avoid a race between the provisioning stack running on the bootstrap control plane and the real one running in the installed cluster. During IPI this annotation is removed when bootstrap is finished, but for assisted and agent installs it was never getting removed.

This was causing baremetal operator to never start in the installed cluster which prevented the machine API from recognizing that all machines were running. This, in turn, resulted in the cluster never being able to finish finalizing with the machine API operator reporting an degraded condition like:

```
- lastTransitionTime: "2024-05-09T22:11:21Z"
  message: 'Failed when progressing towards operator: 4.16.0-0.nightly-2024-05-08-222442
    because minimum worker replica count (2) not yet met: current running replicas
    0, waiting for [test-infra-cluster-96-64zpj-worker-0-59bzh test-infra-cluster-96-64zpj-worker-0-cktvz]'
  reason: SyncingFailed
  status: "True"
  type: Degraded
```

Caused by https://issues.redhat.com/browse/OCPBUGS-33157
Related to https://issues.redhat.com/browse/OCPBUGS-33493

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

- [ ] Does this change include unit-tests (note that code changes require unit-tests) - in progress

## Reviewers Checklist


internal/ignition/ignition.go
```diff
@@ -885,13 +887,18 @@ func (g *installerGenerator) updateBootstrap(ctx context.Context, bootstrapPath
 				// drop this from the list of Files because we don't want to run BMO
 				continue
 			}
+			if provErr := removeProvisioningPausedAnnotation(&config.Storage.Files[i]); err != nil {
```
eranco74, line 890: Currently, whenever assisted is creating a BMH resource the BMH status is `unmanaged` (SaaS) or `externally provisioned` (ZTP). if assisted behavior change or ironic/metal3 will try to update the BMh status we will end up with the original issue (https://issues.redhat.com//browse/OCPBUGS-33157).
Perhaps removing this annotation from the assited-installer-controller after all control plane nodes are ready is preferable since it aligns with the desired CBO behavior.

zaneb: The annotation is purely to prevent BMO in the cluster interfering with the BMO on the baremetal bootstrap host while it is still doing its work.
In a system with no baremetal bootstrap host it is not required.

internal/ignition/ignition.go
```diff
@@ -963,23 +970,55 @@ func isBaremetalProvisioningConfig(file *config_latest_types.File) bool {
 	return strings.Contains(file.Node.Path, "baremetal-provisioning-config")
 }
 
-func fileToBMH(file *config_latest_types.File) (*bmh_v1alpha1.BareMetalHost, error) {
+func removeProvisioningPausedAnnotation(file *config_latest_types.File) error {
+	provisioning := &metal3iov1alpha1.Provisioning{}
+	if err := deserializeToObject(file, provisioning); err != nil {
```
zaneb, line 975: This will cause us problems in future, because the API definition will have to be kept up to date every time a field is added to the Provisioning resource, and we will certainly forget that this is here and then wonder why the new data is getting silently dropped.

This might still be the best thing to do in the short term, but it's not a great substitute for a long-term fix that doesn't require adding this annotation at all.

carbonin: How long do you think it will be until we get a proper long-term fix?
Do you think doing this in the controller against the actual cluster with a patch API call will be any more maintainable?

zaneb: We got a successful test on https://github.com/openshift/installer/pull/8391 but there's a few moving parts so it might take a minute to land.

zaneb: Using a patch API call would avoid the out-of-date-API-definition problem.

PR #6307, opened 2024-05-11
NO-ISSUE: Deprecate and disable support for OCM offline token

jhernand:
Currently or `ocm/client` package supports authentication to OCM using an offline token via the `OCM_SELF_TOKEN` environment variable. But the OCM team has deprecated this authentication mechanism and will remove it in the future. The alternative is to use a client identifier and client secret, which is what we use in the SaaS environment. This patch disables by default that support, and will make the server fail with an explicit error message if it is used. Users that really need to use it will need to explicitly enable it setting the `ACKNOWLEDGE_DEPRECATED_OCM_SELF_TOKEN` environment variable to `yes`. In that case the server will start, but a warning will be written to the log.

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/ocm/client.go
+++ b/pkg/ocm/client.go
@@ -2,6 +2,7 @@ package ocm
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	sdkClient "github.com/openshift-online/ocm-sdk-go"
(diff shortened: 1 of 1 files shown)


pkg/ocm/client.go
```diff
@@ -113,7 +125,25 @@ func (c *Client) newConnection() error {
 	if c.Config.ClientID != "" && c.Config.ClientSecret != "" {
 		builder = builder.Client(c.Config.ClientID, c.Config.ClientSecret)
 	} else if c.Config.SelfToken != "" {
-		builder = builder.Tokens(c.Config.SelfToken)
+		if strings.EqualFold(c.Config.AcknowledgeDeprecatedSelfToken, "yes") {
```
rccrdpccl, line 128: nit: if we invert this condition we can exit early and remove `else` condition

PR #6325, opened 2024-05-16
MGMT-17826: Terminate reconcile early if ManagedCluster CRD not present 

paul-maidment:
When the ManagedCluster CRD is not present during installation of the infrastructure operator, a failure is encountered

```
"failed to get API group resources: unable to retrieve the complete list of server APIs: cluster.open-cluster-management.io/v1: the server could not find the requested resource"
```

This results in a non functional operator.

This PR aims to address that by ensuring that we do not complete the reconcile loop (and therefore perfom the check for the presence of a ManagedCluster CR) if the CRD is not present.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -44,8 +44,10 @@ import (
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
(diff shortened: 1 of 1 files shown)


carbonin (commented):
Does this work? I'd expect this to fail when we set up our watches as well.

paul-maidment:
As per our discussion and some further testing.
It's more beneficial to us to leave things the way they are.
I'm going to close this PR as per that discussion.

paul-maidment:
Re-opening as I have discovered that the infrastructure operator will not function correctly unless we handle this issue.

carbonin:
>as I have discovered that the infrastructure operator will not function correctly unless we handle this issue

Can you elaborate a bit more? Ideally we'd have the failing behavior explained in the bug and also here in more detail.

carbonin:
While I think we should consider this, I'm still split.

All customers would install the operator through MCE and thus would definitely have ManagedCluster, but we also publish our operator in the community operators catalog to install standalone. This change would break that workflow.

That said, we do the same thing for BMH and treating the two resources the same would make sense. But on the other hand booting using a BMH is a much more common use case than managing the local cluster with assisted.

@avishayt what do you think?

paul-maidment:
> > as I have discovered that the infrastructure operator will not function correctly unless we handle this issue
> 
> Can you elaborate a bit more? Ideally we'd have the failing behavior explained in the bug and also here in more detail.

It is effectively a crash as soon as a watch is attempted on Managed cluster.

So in a way it is very similar to the behaviour I have placed in this PR, with the exception that the code in this PR makes it very clear why the error occurs.

paul-maidment:
> While I think we should consider this, I'm still split.
> 
> All customers would install the operator through MCE and thus would definitely have ManagedCluster, but we also publish our operator in the community operators catalog to install standalone. This change would break that workflow.
> 
> That said, we do the same thing for BMH and treating the two resources the same would make sense. But on the other hand booting using a BMH is a much more common use case than managing the local cluster with assisted.
> 
> @avishayt what do you think?

I think my original proposal to terminate the LocalClusterImportController reconcile loop early in the case of Managed cluster CRD not present is reasonable.

It prevents a hard crash while allowing the rest of the operator to be functional.

After creation of the Managed cluster CRD. The operator is able to start (though the user will need to force a reconcile)

PR #6347, opened 2024-05-21
MGMT-17896: Add separate Dockerfile for assisted-service with an el8 base

carbonin:
This is required to ensure that users have an option to install FIPS clusters for versions where the installer is linked against el8 crypto libraries.

The resulting image will be deployed by the operator when the user indicates that clusters requiring the el8 libraries will be installed in FIPS mode (versions < 4.16).

Resolves https://issues.redhat.com/browse/MGMT-17896

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


pastequo:
Diff with https://github.com/openshift/assisted-service/blob/master/Dockerfile.assisted-service:
- FROM quay.io/centos/centos:stream8 as builder
  FROM quay.io/centos/centos:stream9 as builder
- RUN dnf install --enablerepo=powertools -y gcc git nmstate-devel && dnf clean all
  RUN dnf install --enablerepo=crb -y gcc git nmstate-devel && dnf clean all
- FROM quay.io/centos/centos:stream8
  FROM quay.io/centos/centos:stream9

Is there a way to have a common dnf install for both image ? If it might be worth it to add an ARG for the stream8/stream9 variation instead of adding a new file to maintain

carbonin:
Using `ARG`s in `FROM` and especially before the first `FROM` is something I've avoided since the confusion in https://github.com/openshift/imagebuilder/pull/151

Ideally this will be temporary since we'll want to move toward the solution outlined in the body of https://github.com/openshift/assisted-service/pull/6290 in the long term.

That said it may need to be a problem we solve for the installer runner images anyway, but hopefully those will be significantly simpler to build than assisted-service.

pastequo:
I'm not sure to see what you refer to when you talk about "confusion"

- Is it related to the docker discussion ? In this case I think we are safe since this variable would only be used by FROM directives
- Is it related to the imagebuilder issue you linked ? In this case, it's not clear if you have fixed it or not, but it sounds like a bug (not implementing the confusing docker behavior). imagebuilder being one of many tool to build images, is there a reason to restrict ourselves from using a syntax because one implementation has some bug ? I don't know if/where imagebuilder is used

carbonin:
> imagebuilder being one of many tool to build images, is there a reason to restrict ourselves from using a syntax because one implementation has some bug ? I don't know if/where imagebuilder is used

Sorry, I should have given a bit more information. That library (imagebuilder) is what is behind `podman build`, [`buildah`](https://buildah.io/) as well as [OpenShift builds](https://docs.openshift.com/container-platform/4.15/rest_api/workloads_apis/build-build-openshift-io-v1.html).

The "confusion" I'm referencing is that there was difference in behavior between how imagebuilder was handling these "heading args" and how docker handled it. All of our automated and published images should be using the imagebuilder code in some way, but I'm sure there are still people using docker locally so I wanted to avoid any chance of a very difficult to debug issue coming up due to that difference.

That all said, I'll check to see if there's some straightforward way to parameterize this and use a single file. If not, I'd like to move forward with the different files for now given that this has a few other changes depending on it (all of which need to get in by the end of next week). We can then optimize later (or just wait until I remove them with the planned changes for the next release).

----------------------------------------------------

On a separate note, and while you're here @pastequo ... How to I rerun the Konflux PR check that's failing here? Or at least how do I see what failed? When I click into the status I can see that something called a `clamav-scan` failed due to a timeout, but following the link brings me to what looks like a generic "Get started with Red Hat Trusted Application Pipeline" page.

adriengentil:
/retest

should do the trick to re-run konflux pipeline @carbonin

carbonin:
Moved to using a build arg. @adriengentil @pastequo can you take another look?

adriengentil (approved):

PR #6359, opened 2024-05-23
MGMT-17763: Setting Machine network CIDR is forbidden" error is only present on cluster updates

ori-amizur:
When a cluster is single stack and DHCP VIP allocation is off, the machine-network is calculated by selecting a host network that the VIPs belong to. So machine network update is forbidden in this case. This is undesirable for KubeAPI where the machine-network is specified in a custom resource.
So this change allows machine-network update if the update arrived from KubeAPI (non interactive).

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @gamli75 
/cc @carbonin 
/cc @eranco74

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -2125,6 +2125,7 @@ func (b *bareMetalInventory) updateNonDhcpNetworkParams(cluster *common.Cluster,
 	}
 	if params.ClusterUpdateParams.MachineNetworks != nil &&
 		common.IsSliceNonEmpty(params.ClusterUpdateParams.MachineNetworks) &&
+		interactivity == Interactive &&
 		!reqDualStack {
 		err := errors.New("Setting Machine network CIDR is forbidden when cluster is not in vip-dhcp-allocation mode")
 		log.WithError(err).Warnf("Set Machine Network CIDR")
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -4299,6 +4299,26 @@ var _ = Describe("cluster", func() {
 					verifyApiErrorString(reply, http.StatusBadRequest,
 						"Setting Machine network CIDR is forbidden when cluster is not in vip-dhcp-allocation mode")
 				})
+				It("Machine network CIDR in non dhcp non interactive", func() {
+					mockClusterUpdatability(2)
+					mockHostApi.EXPECT().GetStagesByRole(gomock.Any(), gomock.Any()).Return(nil).AnyTimes()
+
+					setAndCheck := func(cidr models.Subnet) {
+						_, err := bm.UpdateClusterNonInteractive(ctx, installer.V2UpdateClusterParams{
+							ClusterID: clusterID,
+							ClusterUpdateParams: &models.V2ClusterUpdateParams{
+								MachineNetworks: []*models.MachineNetwork{{Cidr: cidr}},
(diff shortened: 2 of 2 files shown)


carbonin:
This fixes the problem, but I don't see the difference between the two cases.

Both kube and REST allow users to set the machine network and both also do the calculation to determine the machine network based on the VIPs so why treat them differently?

carbonin:
Would adding a check that the machine networks are not actually changing solve the problem?

ori-amizur:
> This fixes the problem, but I don't see the difference between the two cases.

I don't think there is an intention to allow setting machine-network in SaaS when not in VIPs DHCP mode.  The UI expects the machine-network to be inferred from the VIPs and the host-networks.  So it is not a bug for SaaS.
The flow in SaaS is different.   Hosts are discovered and the machine-network is never set.  When using KubeAPI machine-network is set before the hosts are discovered.  This leads to different flows.

> Both kube and REST allow users to set the machine network and both also do the calculation to determine the machine network based on the VIPs so why treat them differently?

Calculation and setting the machine-network is only done for REST.

ori-amizur:
> Would adding a check that the machine networks are not actually changing solve the problem?

IDK how the backup/restore works.  Maybe the machine-networks were different or empty before.

carbonin:
>Calculation and setting the machine-network is only done for REST.

I'm pretty sure we're calculating it for both. Machine network isn't a required field for kubeAPI and isn't set in our [documented ACI](https://github.com/openshift/assisted-service/blob/b8e6e5320b44423974f6081b7c0507792ac0b2df/docs/hive-integration/crds/agentClusterInstall.yaml), it's also conditional for the [ACI we use in CI tests](https://github.com/openshift/assisted-service/blob/b8e6e5320b44423974f6081b7c0507792ac0b2df/deploy/operator/ztp/agentClusterInstall.j2#L27-L32).

I assumed that means it must be calculated or is machine network not actually required in the install config?

ori-amizur:
The main difference between kube and rest API is the flow.  
In kube api we allow setting vips that do not belong to machine-network knowing that they will not pass the validations.
In reset API since it is interactive, there is expectation that in case there is an error it will be shown as a reply from the api call.  So this block https://github.com/openshift/assisted-service/blob/22fe321b2fddd60381a264c6ab7e3d5c0f1bacf7/internal/bminventory/inventory.go#L2145-L2202 is not relevant for kube api.
So the flow will not be the same.
Coming to think of it, maybe the machine-network is calculated here https://github.com/openshift/assisted-service/blob/22fe321b2fddd60381a264c6ab7e3d5c0f1bacf7/internal/cluster/cluster.go#L429
But I still think that since this change is urgent and we don't want to involve UI in this change, it should be kept minimal.

carbonin:
> But I still think that since this change is urgent and I don't want to involve UI in this change, it should be kept minimal.

Okay, yup, that makes sense. I didn't know this was larger than those few lines.

/lgtm

carbonin (approved):

gamli75:
/cherry-pick release-ocm-2.10 release-ocm-2.9 release-ocm-2.8

PR #6383, opened 2024-05-30
MGMT-17300: adding has partition field for disks

eifrach:
- Adding new field for disks
- `HasPartition` field will be used for new validation if disk not empty 

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -5909,6 +5909,8 @@ definitions:
         type: boolean
       removable:
         type: boolean
+      partitionTypes:
+        type: string
       is_installation_media:
         type: boolean
         description: Whether the disk appears to be an installation media or not
--- a/api/vendor/github.com/openshift/assisted-service/models/disk.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/disk.go
@@ -57,6 +57,9 @@ type Disk struct {
 	// name
 	Name string `json:"name,omitempty"`
 
+	// partition types
+	PartitionTypes string `json:"partitionTypes,omitempty"`
+
 	// path
 	Path string `json:"path,omitempty"`
 
--- a/client/vendor/github.com/openshift/assisted-service/models/disk.go
+++ b/client/vendor/github.com/openshift/assisted-service/models/disk.go
@@ -57,6 +57,9 @@ type Disk struct {
 	// name
 	Name string `json:"name,omitempty"`
 
+	// partition types
+	PartitionTypes string `json:"partitionTypes,omitempty"`
+
 	// path
 	Path string `json:"path,omitempty"`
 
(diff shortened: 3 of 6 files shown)


avishayt:
Can you describe what this is for in more detail?  What does it mean?  What will the validation report?  What will the user do with that information?

eifrach:
> What would the agent check? Obviously partitions, but what about LVs? Anything else?

no need to check anything else if you have a partition table it mean that the disk is not empty
so we just want to see if it set or not.

This keeps the check simple.

eifrach:
/hold
need to confirm some checks

eifrach:
> What would the agent check? Obviously partitions, but what about LVs? Anything else?

Found a solution using LVM package to check if there is VG configured
```golang
	path := "/dev/vda"

	err := lvm.PVScan(path)
	if err != nil {
		fmt.Printf("PV not found in %v.\n ", path)
	} else {
		data, _ := lvm.ListVolumeGroupNames()
		fmt.Println(data)
	}
```

swagger.yaml
```diff
@@ -5909,6 +5909,8 @@ definitions:
         type: boolean
       removable:
         type: boolean
+      has_partitions:
+        type: boolean
```
avishayt, line 5913: Maybe a string is better - you can pass more information to the user.  For example "The disk has a 4 partitions" or "The disk has 3 logical volumes", etc.  Either that, or change the boolean to an int and return only data about partitions.

```suggestion
      formatting:
        type: string
```

avishayt: It's not only about the number; it's about what it is.  The point is that a boolean conveys limited information, so it's usually not a good idea to use it in an API.  If we want to provide the user with more information, we would need to add another property or deprecate this one and replace it.  So, it is better to use a string, and then we can provide the user with more information.  For example, if we told them there were LVs on it, they might pause and check it out.

eifrach: ok but the formatting is kinda bad naming cause we have a field for `format` 
can we call it `partitionTableType` which will accept `[ basic , lvm, none]` and `none` value will be there passsing value for the validation

avishayt: How about `partitionTypes` that returns a string that is a comma-separated list of partition types?  So the user can see if the partitions are LVM, regular linux file systems, RAID, Ceph OSD, etc.

avishayt (approved):

PR #6385, opened 2024-05-30
MGMT-8356: Deploy assisted-service and its components to kind cluster instead of minikube for local subsystem testing

danmanor:
## Overview

This PR is is about using `kind` cluster using `podman` provider to deploy assisted-service and its components for local subsystem tests.

## Goals

Using `kind` instead of `Minikube` for local `k8s` development can offer several advantages:

- **Speed and Simplicity**: `kind` starts faster than `Minikube` because it runs `k8s` cluster components as `podman` containers. This can lead to quicker setup and teardown times.
- **Less Resource-Intensive**: Since `kind` runs within `podman`, it tends to use less system resources compared to `Minikube`, which creates a new VM for the `k8s` cluster. This is a big advantage for running subsystem tests on a PC.
- **Avoid using KVM**: `KVM` was causing several issues in our deployments.

## Major Changes

- **Network**: With `Minikube`, components we automatically exposed outside the cluster when their `LoadBalancer` `Service` resource was applied (by getting External IP). we utilized it to get traffic inside the cluster into the components. this is not the case with `kind`, so we expose static `nodePort` for each service and configure kind to forward the traffic from `localhost:target-port` to `cluster:nodePort`
- **Local Registry**: With `Minikube`, we were using local registry to store assisted-service image after it was built, for the pod to pull from later. Configuring local registry with `kind` using `podman` is not supported and redundant. Instead, I loaded the build image into kind for local use

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [x] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


Makefile
```diff
@@ -380,30 +365,35 @@ create-ocp-manifests:
...
-ci-deploy-for-subsystem: _verify_cluster generate-keys
-	export TEST_FLAGS=--subsystem-test && export AUTH_TYPE="rhsso" && export DUMMY_IGNITION=${DUMMY_IGNITION} && \
-	export IPV6_SUPPORT="True" && export ENABLE_ORG_TENANCY="True" && export ENABLE_ORG_BASED_FEATURE_GATES="True" && \
-	export RELEASE_SOURCES='$(or ${RELEASE_SOURCES},${DEFAULT_RELEASE_SOURCES})' && \
-	$(MAKE) deploy-wiremock deploy-all
+ci-deploy-for-subsystem: deploy-test
 
-patch-service: _verify_cluster update-local-image
+patch-service: _verify_cluster
+	$(MAKE) load-image SERVICE=localhost/assisted-service:latest
```
rccrdpccl, line 371: Why $(MAKE) ? Can't this be added as dependency as `_very_cluster` target is?

deploy/assisted-image-service-service.yaml
```diff
@@ -13,6 +13,7 @@ spec:
       port: 8080
       protocol: TCP
       targetPort: 8080
+      nodePort: 30001
```
rccrdpccl, line 16: Is this manifest only used by Kind?

deploy/assisted-image-service-service.yaml
```diff
@@ -13,6 +13,7 @@ spec:
...
   selector:
     app: assisted-image-service
   type: LoadBalancer
```
rccrdpccl, line 19: Do we want nodeport or LoadBalancer?

deploy/assisted-image-service-service.yaml
danmanor: Currently not as minikube is still used for e2e tests

Makefile
```diff
@@ -238,16 +229,10 @@ update-debug-minimal:
...
-
-_update-local-k8s-image:
-	# Temporary hack that updates the local k8s(e.g minikube) with the latest image.
-	# Should be replaced after installing a local registry
-	./hack/update_local_image.sh
-
-update-local-image: $(UPDATE_LOCAL_SERVICE)
+load-image: update-image
+	rm -f build/assisted_service_image.tar || true
+	$(CONTAINER_COMMAND) save -o build/assisted_service_image.tar $(SERVICE)
```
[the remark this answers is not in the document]
danmanor: AFAIK you can load docker image (but we use podman) or use image archives

deploy/assisted-image-service-service.yaml
danmanor: I need a static `nodePort` port, but I still want to get external IP when deployed on AWS/minikube, so I kept it LoadBalancer

rccrdpccl: couldn't we use some templating instead (kustomize/helm) to generate a cleaner manifest?

docs/dev/running-test.md
```diff
@@ -1,103 +1,65 @@
...
+- REST-API mode
+- Kube-API mode
 
-Assisted-service subsystem tests require that you first deploy the
-assisted-service on a k8s cluster (e.g. minikube).
+## Deployment for Subsystem Tests
 
-This document will explain how you can easily deploy the service in preperation
-for subsystem tests, and also how to run the tests themselves.
+Assisted-service components will be deployed in a Kind cluster using the Podman provider. Necessary tools like `kind`, `kustomize`, `strato-skipper`, and `waiting` will be installed automatically if not present.
```
rccrdpccl, line 11:
```# make deploy-service-for-subsystem-test
python3 ./tools/wait_for_cluster_info.py --namespace "assisted-installer"
  File "/root/assisted-service/./tools/wait_for_cluster_info.py", line 4, in <module>
    import utils
  File "/root/assisted-service/tools/utils.py", line 7, in <module>
    import waiting
ModuleNotFoundError: No module named 'waiting'
```

hack/kind/kind.sh
```diff
@@ -0,0 +1,58 @@
...
+	# for more information about contour, see: https://projectcontour.io
+	kubectl apply -f $__dir/contour.yaml
+	kubectl rollout status -n projectcontour daemonset envoy --timeout 2m
+	kubectl rollout status -n projectcontour deployment contour --timeout 2m
+}
+
+function create() {
+	check
+
+	if ! kind export kubeconfig &> /dev/null ; then
```
rccrdpccl, line 42: We need to define a cluster name.
It is likely that people will run different kind clusters for different projects, and not naming it could be problematic

hack/kind/kind.sh
```diff
@@ -0,0 +1,58 @@
...
+set -o nounset
+set -o errexit
+set -o pipefail
+
+KIND_VERSION="0.17.0"
+
+__dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+
+function check() {
+	if [ "$(kind --version)" == "kind version $KIND_VERSION" ]; then
```
rccrdpccl, line 12: # kind version
kind v0.17.0 go1.19.2 linux/amd64

rccrdpccl, line 12: Why? please add a comment with the reason

Also, this is a bit intrusive, it would (and it did in my case) swap the original's system kind with version 0.17.0

Can we at least prompt the user to accept this or not?

Makefile
rccrdpccl: can't we name it as the user specifies and use the global SERVICE variable instead? Is the problem only replacing the manifest's value?

rccrdpccl:
Looks good overall, but I think we should tidy up the manifest and move one step forward towards having a more flexible manifests.

For now I'd suggest to move the dynamic logic to the current python scripts, but probably for the longer term it'd be great to move towards something more standard (kustomize/helm)

deploy/assisted-image-service-service.yaml
danmanor: This change shouldn't have any affect in other deployments (not kind) as nodePort was given either way, we would just had to search for it :)

hack/kind/kind.sh
danmanor: Agreed

hack/kind/kind.sh
danmanor: Agreed

danmanor: Do you think writing it in the docs will suffice ?

Makefile
danmanor: I prefer not changing it globally as it is used in many targets and it is hard to anticipate the scope of its impact

rccrdpccl: But this way the user is forced to build an image  all the time, right? i.e. they wouldn't be able to quickly use something they pushed to quay.io without editing the config, right?

deploy/assisted-image-service-service.yaml
rccrdpccl: the problem I see is that keeping LoadBalancer will make users think something is wrong (pending external IP forever), also defining fixed nodePort for apparently no reason (in the minikube's case) can cause confusion.
I think it'd be cleaner if we define it properly according to each use case

hack/kind/kind.sh
rccrdpccl: I don't think so, users will most likely run this script without giving it much thought, and they can end up with a kind binary different from what they had in the host

eliorerz:
You have extra character in the commit message: `oMGMT-8356`

danmanor:
@eliorerz Thank you

tools/utils.py
```diff
@@ -69,7 +70,7 @@ def set_namespace_in_yaml_docs(docs, ns):
                 doc['metadata']['namespace'] = ns
         except KeyError:
             continue
-
+
```
rccrdpccl, line 73: :sweat_smile:

tools/utils.py
```diff
@@ -329,3 +336,22 @@ def get_cluster_server(cluster_name='default'):
         )
 
     return out
+
+
+def override_service_type_definition_and_node_port(
+    internal_definitions_path: str,
+    internal_target_definitions_path: str,
+    new_type: str,
```
rccrdpccl, line 344: `service_type`?

tools/utils.py
```diff
@@ -329,3 +336,22 @@ def get_cluster_server(cluster_name='default'):
         )
 
     return out
+
+
+def override_service_type_definition_and_node_port(
+    internal_definitions_path: str,
```
rccrdpccl, line 342: is this source manifest filaname and output filename?

tools/deploy_wiremock.py
```diff
@@ -11,18 +20,103 @@ def main():
...
         namespace=deploy_options.namespace,
         file=dst_file
     )
 
+    if deploy_options.target == "kind":
+        hostname = os.environ.get('IP', 'localhost')
+        port="8070"
+
+    else:
+        log.info("Waiting for Wiremock service to be ready...")
```
rccrdpccl, line 50: Maybe we can comment why we need to wait for wiremock service if not kind target? I take it it's because we're waiting for the load balancer to be provisioned, right?

tools/utils.py
danmanor: Yes

Makefile
```diff
@@ -380,30 +375,36 @@ create-ocp-manifests:
...
+	export TEST_FLAGS=--subsystem-test && \
 	export AUTH_TYPE="rhsso" && export DUMMY_IGNITION="True" && \
 	export IPV6_SUPPORT="True" && ENABLE_ORG_TENANCY="True" && ENABLE_ORG_BASED_FEATURE_GATES="True" && \
 	export RELEASE_SOURCES='$(or ${RELEASE_SOURCES},${DEFAULT_RELEASE_SOURCES})' && \
 	$(MAKE) deploy-wiremock deploy-all
 
-# An alias for the deploy-test target
-deploy-service-for-subsystem-test: deploy-test
+deploy-service-for-subsystem-test:
+	$(MAKE) create-hub-cluster
```
rccrdpccl, line 398: why `$(MAKE)` and not used as a dependency?

docs/dev/running-test.md
```diff
@@ -1,103 +1,73 @@
...
-### minikube
+## Deploying Components
 
-First we must prepare the minikube cluster -
+To deploy in REST-API mode, run:
 
 ```bash
-# Optionally delete the existing minikube cluster:
-# minikube delete
+make deploy-service-for-subsystem-test
```
rccrdpccl, line 25: Can we clarify this must be run from the host? This target runs skipper

Makefile
danmanor: I wanted to control the order of executions. to my understanding when you specify list of dependencies e.g:

```
deploy-service-for-subsystem-test: create-hub-cluster load-image
```

there is no guarantee that this is the order it will execute unless one of the dependencies is also a dependency of the other, which shouldn't be the case here

docs/dev/running-test.md
danmanor: I mentioned in the doc that it requires skipper

danmanor: Do you think we should add something else ?

tools/utils.py
rccrdpccl: Can we give them a better name? :sweat_smile: It's not really clear to me, I had to follow the code to guess it

Makefile
rccrdpccl: Dependencies are executed in order, unless further dependencies state otherwise. If you really need one step to be executed before another, you can define so in the same step:
```
deploy-service-for-subsystem-test: create-hub-cluster load-image
    ...

load-image: create-hub-cluster
   ...
```

docs/dev/running-test.md
rccrdpccl: That's not what I meant, I think we should explicitly say this command needs to be executed directly from the host, without skipper.
I think it's already not very clear (for pre-existing targets) which targets are to be executed in skipper and which ones are to be executed from the host.

danmanor: So where you think we should specify it ?

danmanor: I added a comment in the makefile and it is specified in the doc

Makefile
danmanor: Done

Makefile
```diff
@@ -380,30 +375,35 @@ create-ocp-manifests:
...
-	export SERVICE=${LOCAL_SERVICE_IMAGE} && export TEST_FLAGS=--subsystem-test && \
+	export TEST_FLAGS=--subsystem-test && \
 	export AUTH_TYPE="rhsso" && export DUMMY_IGNITION="True" && \
 	export IPV6_SUPPORT="True" && ENABLE_ORG_TENANCY="True" && ENABLE_ORG_BASED_FEATURE_GATES="True" && \
 	export RELEASE_SOURCES='$(or ${RELEASE_SOURCES},${DEFAULT_RELEASE_SOURCES})' && \
 	$(MAKE) deploy-wiremock deploy-all
 
-# An alias for the deploy-test target
-deploy-service-for-subsystem-test: deploy-test
+# execute on the host, without skipper
```
rccrdpccl, line 397: I meant more on the documentation's reference, I hope here is clear as two lines below there is a `skipper` command :sweat_smile:

Makefile
```diff
@@ -417,34 +417,43 @@ deploy-onprem:
...
-subsystem-test: test
+# execute on the host, without skipper
+subsystem-test:
+	TARGET=kind skipper $(MAKE) test
 
-# An alias for the test-kube-api target
-subsystem-test-kube-api: test-kube-api
+# execute on the host, without skipper
+subsystem-test-kube-api:
+	TARGET=kind skipper $(MAKE) test-kube-api
```
rccrdpccl, line 439: why skipper command has been introduced?
Can't this be executed `skipper make subsystem-test` ?

danmanor: you mean without skipper ?

Makefile
danmanor: The docs specify to tun this without skipper, isn't that what you meant ?

Makefile
rccrdpccl: skipper supports makefile target, why are we executing skipper within makefile targets? I think it should be the other way round, unless we have some mixed approach (i.e. within one target, one command needs skipper and one needs to run without). WDYT?

PR #6410, opened 2024-06-06
MGMT-17794: Make sure that uploaded manifest files have valid filename before extensions

paul-maidment:
The validation of manifest filenames as it stands has a bug

any file without a name before the extension starts is considered to be valid.

`.yaml.patch` is considered valid when it should not be for example

This fix remedies that by ensuring that, in addition to the other checks we already peform, that file filename does not begin with a `.`

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/manifests/manifests.go
+++ b/internal/manifests/manifests.go
@@ -453,7 +453,7 @@ func (m *Manifests) fetchManifestContent(ctx context.Context, clusterID strfmt.U
 func (m *Manifests) validateManifestFileNames(ctx context.Context, clusterID strfmt.UUID, fileNames []string) error {
 	for _, fileName := range fileNames {
 		fileNameWithoutExtension := strings.TrimSuffix(fileName, filepath.Ext(fileName))
-		if len(strings.TrimSpace(fileNameWithoutExtension)) == 0 {
+		if fileName[0] == '.' || len(strings.TrimSpace(fileNameWithoutExtension)) == 0 {
 			return m.prepareAndLogError(
 				ctx,
 				http.StatusUnprocessableEntity,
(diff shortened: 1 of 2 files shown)


internal/manifests/manifests_test.go
```diff
@@ -269,7 +269,7 @@ var _ = Describe("ClusterManifestTests", func() {
 			It("Does not accept a filename that does not contain a name before the extension", func() {
 				clusterID := registerCluster().ID
 				content := "{}"
-				fileName := ".yaml"
+				fileName := ".yaml.patch"
```
eifrach, line 272: shouldn't we test both cases? 
`.yaml` and `.yaml.something`

avishayt (approved):

eifrach (changes requested):

internal/manifests/manifests_test.go
paul-maidment: The point is to check for the filename starting with period.
The regex already takes care of these cases.

We can possibly update the test to cover lots of cases but I'm not sure it's necessary in this case?

internal/manifests/manifests.go
```diff
@@ -453,7 +453,7 @@ func (m *Manifests) fetchManifestContent(ctx context.Context, clusterID strfmt.U
 func (m *Manifests) validateManifestFileNames(ctx context.Context, clusterID strfmt.UUID, fileNames []string) error {
 	for _, fileName := range fileNames {
 		fileNameWithoutExtension := strings.TrimSuffix(fileName, filepath.Ext(fileName))
```
rccrdpccl, line 455: Wouldn't a regex approach be simpler?

internal/manifests/manifests_test.go
eifrach: /lgtm

PR #6431, opened 2024-06-13
MGMT-18163: Enhancement - Transactional Custom Manifests

paul-maidment:
We would like to improve the custom manifest feature to make it more consistent, more reliable and more user friendly. By addressing the following points

We have had a recent bug report about an edge case

  The following bug was contributed to by this issue: https://issues.redhat.com/browse/MGMT-18019

In this edge case, if there is a failure in S3 with unfortunate timing then it is possible to store a manifest with inconsistent metadata.
This is not a situation that is acceptable moving forward and we need to do something.

There are some problems with the existing implementation of manifest storage when it comes to metadata and we need to change the implementation to ensure that any manifest creation or update is performed in a single operation to the underlying storage.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/transactional-custom-manifests.md
+++ b/docs/enhancements/transactional-custom-manifests.md
@@ -1,0 +1,260 @@
+---
+title: Transactional Custom Manifests
+authors:
+  - "@pmaidmen"
+creation-date: 2024-06-09
+last-updated: 2024-06-09
+---
+
+# Transactional Custom Manifests
+
+# Background:
+
+It is important to start with a description of how the current functionality works;
+
+Custom manifests are stored in an "object store", this may be within an object store container or within the Amazon S3 service, depending on where and how the assisted service is deployed.
+
+Manifest storage is divided into two directories
+
+/manifests and /openshift
+
+Manifests may be uploaded to either directory via the assisted service.
+
+
+Manifests are configuration files that are applied by openshift during the installation phase. These are in .yaml, yml or json format. It is also possible to upload "patch" files that will alter yaml or yml files within the openshift or manifest directories.
+
+Manifest and patch files may be marked as either "system generated" or "user generated", by default a file is considered "system generated". 
+
+To mark a file as user generated, a metadata file is created under the subtree `manifest-attributes` to match the file and path in the subtree `manifests`
+
+So if we have a file:
+`manifests/openshift/user-defined-manifest.yaml`
+
+The metadata file to mark this as user-supplied would be:
+`manifest-attributes/openshift/user-defined-manifest.yaml/user-supplied`
+
+As we are using the filesystem as a means of storing metadata about a file, this needs to be performed in two operations, one to write the file and one to write the metadata.
+
+This has a weakness; by performing two writes, we risk that the second write may not take place in the case of an S3 failure. 
+
+# Summary
+
+We would like to improve the custom manifest feature to make it more consistent, more reliable and more user friendly. By addressing the following points
+
+* Find a more reliable and "transactional" way in which the upload of manifest files may be handled.
+
+# Motivation
+
+We have had a recent bug report about an edge case
+
+  The following bug was contributed to by this issue: https://issues.redhat.com/browse/MGMT-17802
+
+In this edge case, if there is a failure in S3 with unfortunate timing then it is possible to store a manifest with inconsistent metadata.
+This is not a situation that is acceptable moving forward and we need to do something.
+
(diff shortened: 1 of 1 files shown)


docs/enhancements/transactional-custom-manifests.md
```diff
@@ -0,0 +1,252 @@
...
+	#### Uploads
+
+	S3Wrapper API:
+	Calls to Upload, UploadFile and UploadStream will need to be updated to support metadata for an object
+	https://github.com/openshift/assisted-service/blob/master/pkg/s3wrapper/client.go#L42-L43
+	
+	##### Filesystem implementation:
+	
+	Implementation changes to the filesystem S3Wrapper client will be required in order to implement metadata support.
+	We should attempt to keep this atomic by using XATTRs if supported.
```
jhernand, line 84: Take into account that setting XATTRs requires a separate `setxattr` system call. So you will need first to create the file and then set the extended attributes. One of those two operations may fail, or never be attempted (if you crash in between) so it won't give you the transactional behavior that you are looking for.

I think that our filesystem implementation creates a temporary file and them, when it is completely written, it moves it to its definitive location. So to get that transactional behavior you will need to make sure that the extended attribute is set on the temporary file before moving it.

paul-maidment: This is a very good point, I will be sure to keep this in mind when working on this.
One thing we do get rid of is distributed + non transactional, easier to deal with rollbacks and issues such as this when the filesystem is local (the original file can be deleted before returning an error)

Not an option in remote filesystems such as S3 of course.

avishayt: Yes I thought of this, but we can get around it by creating the file with the xattr, then renaming it.  The rename operation is transactional.
Though now that I'm reading this document, why not just have separate directories for user-generated manifests?

Instead of
manifests/openshift/user-defined-manifest.yaml
Upload it to
manifests/openshift.userdefined/manifest.yaml

Then you don't need any additional file or attribute, because anything under there is user-defined.

paul-maidment: I actually think that using the AWS metadata is the right approach

I think it's the right approach because using a folder to "hack" some metadata is not a scalable approach and certainly not pleasant to work with or write tests for

using AWS metadata + Xattr (for the local implementation) makes more sense to me as it can be extended in future with minimal effort.

docs/enhancements/transactional-custom-manifests.md
```diff
@@ -0,0 +1,252 @@
...
+	For each cluster
+		For each manifest in cluster (as determined by database + storage)
+			Ensure that appropriate manifest S3 metadata record exists for any files that are "user-generated"
+			Remove the database record for the files that have been processsed
+
+There are approx 3500 live clusters in SAAS, if we allow a conservative estimate of 10 manifests per cluster, this will involve no S3 writes for a "roll forward" and up to 1 write per file for a "roll forward"
+
+### Risks and Mitigations
+
+We must deal with the migration of historical data in order for this to work.
```
avishayt, line 203: Need to see how to execute the migration in the context of ACM.

avishayt (approved):

docs/enhancements/transactional-custom-manifests.md
```diff
@@ -0,0 +1,260 @@
...
+	https://github.com/openshift/assisted-service/blob/master/pkg/s3wrapper/client.go#L42-L43
+	
+	##### Filesystem implementation:
+	
+	Implementation changes to the filesystem S3Wrapper client will be required in order to implement metadata support.
+	We should attempt to keep this atomic by using XATTRs if supported.
+	
+	Implementation of Upload in S3Wrapper API (filesystem):
+	Need to be updated to support metadata for an object (possibly based on XATTRs
+	https://github.com/openshift/assisted-service/blob/master/pkg/s3wrapper/filesystem.go#L55-L70
```
carbonin, line 88: FWIW these links will be inaccurate very soon since they reference `master`.

It would be better to use a commit for these to capture the state of the code when the enhancement was written (generally this also applies to everywhere else you put a GH link like emails, IMs, and comments).

PR #6435, opened 2024-06-13
NO-ISSUE: Refactor uploader pkg

pastequo:
* Flatten pyramidal code
* Explicit data modification
* Use different creds for data enrichment & push authent

Currently, the logic to get pull secret is hard to read (and more complex than it seems). While flattening the nested logic, it highlights some shortcuts in the err check. I tried to explicit the different scenario, so there is a change of behavior. Previously the Hub cluster credentials were used by default and if not, the spoke cluster creds were used (if any). I think spoke cluster creds should be used to enrich the data (since it seems to be more accurate to reflect the cluster owner identity. Any insights/opinions about this part are welcome)

Since the code is able to use the spoke cluster credentials to push data, should we disable by default this feature if there is no creds in the hub creds ? I kept the check, but we can discuss it

Otherwise, since the cluster entity is passed by as a pointer, and the code is modifying its values, it can be error prone in the future for the callers. I isolated the part that changes the data. Introducing a deepcopy library would be the way to go IMHO

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/uploader/auth_utils.go
+++ b/internal/uploader/auth_utils.go
@@ -1,52 +1,130 @@
 package uploader
 
 import (
+	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/openshift/assisted-service/internal/cluster/validations"
 	"github.com/openshift/assisted-service/pkg/k8sclient"
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 const (
-	openshiftTokenKey = "cloud.openshift.com"
+	openshiftNamespace  = "openshift-config"
+	openshiftSecretName = "pull-secret"
+	openshiftTokenKey   = "cloud.openshift.com"
 )
 
-func getPullSecret(clusterPullSecret string, k8sclient k8sclient.K8SClient, key string) (*validations.PullSecretCreds, error) {
-	data, exists := getOCMPullSecret(k8sclient)
-	if !exists {
-		data = clusterPullSecret
-	}
-	if data == "" {
-		return nil, errors.Errorf("failed to find pull secret for %s", key)
-	}
-	pullSecret, err := validations.ParsePullSecret(data)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to parse pull secret")
-	}
-	if auth, ok := pullSecret[key]; ok {
-		return &auth, nil
-	}
-	return nil, errors.Errorf("pull secret doesn't contain authentication information for %s", key)
+type Identity struct {
+	Username    string
+	EmailDomain string
 }
 
-func getOCMPullSecret(k8sclient k8sclient.K8SClient) (string, bool) {
-	if k8sclient != nil {
-		secret, err := k8sclient.GetSecret("openshift-config", "pull-secret")
-		if err != nil {
-			err = client.IgnoreNotFound(err)
-			// A "not found" error (err == nil) should return false
-			// indicating that the secret does not exist
-			return "", err != nil
-		}
-		if pullSecret, ok := secret.Data[corev1.DockerConfigJsonKey]; ok {
-			return string(pullSecret), true
-		}
-		return "", true
+type APIAuth struct {
+	AuthRaw string
+}
+
+type PullSecret struct {
+	Identity Identity
+	APIAuth  APIAuth
+}
+
+func getPullSecret(clusterPullSecret string, k8sclient k8sclient.K8SClient) (PullSecret, error) {
+	ret := PullSecret{}
+
+	managementCreds, managementErr := getManagementCreds(k8sclient)
+	workloadCreds, workloadErr := getCloudOpenshiftCreds(clusterPullSecret)
+
+	if managementErr != nil && workloadErr != nil {
+		return ret, errors.Join(
+			fmt.Errorf("failed to get management creds: %w", managementErr),
+			fmt.Errorf("failed to get workload creds: %w", workloadErr),
+		)
 	}
-	return "", false
+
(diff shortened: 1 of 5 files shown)


pastequo:
cc @CrystalChun

internal/uploader/auth_utils.go
```diff
@@ -1,73 +1,128 @@
...
+
+// Checks if cloud.openshift.com is present in the OCM pull secret to see if the user
+// is still opted-in to sending data
+func isOCMPullSecretOptIn(k8sclient k8sclient.K8SClient) (bool, error) {
+	creds, err := getManagementCreds(k8sclient)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to parse pull secret")
+		if apierrors.IsNotFound(err) {
+			// Indicates the pull secret doesn't exist, so we'll assume they're opted-in
+			return true, nil
```
CrystalChun, line 55: I believe this is the opposite: if the pull secret is not found, then we shouldn't assume they're opted-in. It's actually how they can opt-out (by removing these credentials or not even having this pull secret [1])

[1] https://docs.openshift.com/container-platform/4.15/support/remote_health_monitoring/opting-out-of-remote-health-reporting.html#insights-operator-new-pull-secret_opting-out-remote-health-reporting

internal/uploader/auth_utils.go
```diff
@@ -1,73 +1,128 @@
...
+
+	return pullSecrets.ManagementCreds
+}
+
+func selectPullSecretForAuthent(pullSecrets PullSecretCreds) *validations.PullSecretCreds {
+	if pullSecrets.ManagementCreds != nil {
+		return pullSecrets.ManagementCreds
+	}
+
+	return pullSecrets.WorkloadCreds
```
CrystalChun, line 118: Don't these do the same thing and are used the same way?

internal/uploader/auth_utils.go
```diff
@@ -1,73 +1,128 @@
...
+	openshiftTokenKey   = "cloud.openshift.com"
 )
 
-func getPullSecret(clusterPullSecret string, k8sclient k8sclient.K8SClient, key string) (*validations.PullSecretCreds, error) {
-	data, exists := getOCMPullSecret(k8sclient)
-	if !exists {
-		data = clusterPullSecret
+type PullSecretCreds struct {
+	ManagementCreds *validations.PullSecretCreds
+	WorkloadCreds   *validations.PullSecretCreds
```
CrystalChun, line 22: Alternatively, 
```suggestion
type Credentials struct {
	Username string
	Email string 
	Upload   *validations.PullSecretCreds
```
So you wouldn't have to call `selectCredentials` [here](https://github.com/openshift/assisted-service/pull/6435/files#r1638404061) and can use each field individually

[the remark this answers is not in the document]
rccrdpccl: I think the management cluster's creds should have priority to authenticate the upload, and the workload cluster should have priority to extract identity data.

rccrdpccl: Good idea, maybe `Identity { Email, Username }` and `AuthToken` or something?

internal/uploader/auth_utils.go
pastequo: I'm totally fine changing this, but let's double check I haven't changed that logic and it seemed to be there on purpose https://github.com/openshift/assisted-service/blob/master/internal/uploader/auth_utils.go#L71

internal/uploader/auth_utils.go
pastequo: Those 2 methods gathers the logic of choosing the creds for enrichment vs choosing the creds for authN. They just differ on what is the first value (~default value) to use

I like to isolate this kind of logic in dedicated methods so that we are clear about the expected behavior

internal/uploader/auth_utils.go
rccrdpccl: To clarify:

opt-in (accept that we will handle user data):
* implicit: presence of pull-secret
* explicit: DATA_COLLECTION env var ENABLED

opt-out (disagree sending us user data):
* explicit: DATA_COLLECTION env var DISABLED
* implicit: absence of pull-secret key "cloud.openshift.com" (by extension absence of pull secret too)


as long as we respect this logic, we should be fine

internal/uploader/auth_utils.go
pastequo: I pushed something, let me know if it fits what you had in mind

CrystalChun: Looks great to me, thanks!

CrystalChun (changes requested):
Modifying this piece https://github.com/openshift/assisted-service/pull/6435/files#r1638392761, but otherwise looks good

internal/uploader/auth_utils.go
rccrdpccl: As for ENABLE_DATA_COLLECTION I don't think we should consider "unset", the env var will get a default value (we can debate over the default value if necessary)

If ENABLE_DATA_COLLECTION is true, we should try to do our best to send the data (even with spoke cluster creds IMO)., but if we have no pull secret nor key we wouldn't be able to send it anyway (can throw a warning but the bottom line is that we cannot even attempt sending the data)

pastequo: In this case, presence/absence of pull secret doesn't matter when we check if events uploading is enabled (while starting the service). It just matters when we tried to upload.

I'm changing the logic in https://github.com/openshift/assisted-service/blob/master/internal/uploader/events_uploader.go#L65 and removing this method that is not called anymore https://github.com/openshift/assisted-service/blob/master/internal/uploader/auth_utils.go#L62

CrystalChun (approved):
/lgtm

PR #6437, opened 2024-06-13
NO-ISSUE: Differentiate between monitoring failures and component issues

carbonin:
Currently the function for monitoring the status of components deployed by the agentserviceconfig controller doesn't differentiate between a failure to monitor and a component that is failing.

This means that the reconciliation function actually returns an error when reconcile succeeded, but some component isn't healthy. This isn't correct and causes issues when testing basic success cases.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -288,14 +288,10 @@ func (r *AgentServiceConfigReconciler) Reconcile(origCtx context.Context, req ct
 
 	}
 
-	if err := updateConditions(ctx, log, asc); err != nil {
-		return ctrl.Result{Requeue: true}, err
-	}
-
-	return ctrl.Result{}, nil
+	return updateConditions(ctx, log, asc)
 }
 
-func updateConditions(ctx context.Context, log *logrus.Entry, asc ASC) error {
+func updateConditions(ctx context.Context, log *logrus.Entry, asc ASC) (ctrl.Result, error) {
 	msg := "AgentServiceConfig reconcile completed without error."
 	conditionsv1.SetStatusConditionNoHeartbeat(asc.conditions, conditionsv1.Condition{
 		Type:    aiv1beta1.ConditionReconcileCompleted,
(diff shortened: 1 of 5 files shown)


internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -507,49 +515,41 @@ func monitorOperands(ctx context.Context, log logrus.FieldLogger, asc ASC) error
 	for _, deployName := range []string{"agentinstalladmission", "assisted-service"} {
 		deployment := &appsv1.Deployment{}
 		if err := asc.Client.Get(ctx, types.NamespacedName{Name: deployName, Namespace: asc.namespace}, deployment); err != nil {
-			return err
+			return "", err
```
CrystalChun, line 518: Would this still count as a Monitoring failure?

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -507,49 +515,41 @@ func monitorOperands(ctx context.Context, log logrus.FieldLogger, asc ASC) error
...
-	checkReplicas := func(replicas int32, name string) error {
+	for kind, replicas := range map[string]int32{
+		"created": ss.Status.Replicas,
+		"ready":   ss.Status.ReadyReplicas,
+		"current": ss.Status.CurrentReplicas,
+		"updated": ss.Status.UpdatedReplicas,
+	} {
 		if replicas != desiredReplicas {
-			return fmt.Errorf("StatefulSet %s %s replicas does not match desired replicas", imageServiceName, name)
+			return fmt.Sprintf("StatefulSet %s %s replicas does not match desired replicas", imageServiceName, kind), nil
```
CrystalChun, line 548: Should this return an error to indicate a monitoring failure?

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: Yeah, so what I mean by "monitoring failure" is that we failed to actually monitor a component. So this would include API issues.

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: This isn't really a "monitoring failure" this is a situation where we successfully monitored the stateful set and determined it wasn't in a good state so I'm reporting that there's an issue, but not returning an error from the function.

CrystalChun: Gotcha, thanks for the explanation!

internal/controller/controllers/agentserviceconfig_controller.go
CrystalChun: Ah I see, thanks for explaining!

CrystalChun (approved):
/lgtm

PR #6446, opened 2024-06-18
MGMT-18131: Make application work on non-OCP kubernetes

carbonin:
This PR makes some changes to the operator code to ensure the application can deploy properly to non-OCP kubernetes distributions.

The following changes are made to the deployment when the operator detects it is not running on an OpenShift cluster:

- Routes are dropped in favor of Ingress
    - Routes are an OpenShift specific resource.
    - To use ingress fields on AgentServiceConfig were added for the user to supply the desired hostname and ingress class name
- No TLS is enabled
    - In OCP TLS certs are generated by the `service-ca-operator` which is not available on non-OCP clusters
    - Looking into cert management options will be done in https://issues.redhat.com/browse/MGMT-18137
- Webhooks are not deployed
    - The webhook server requires a certificate to be deployed so this is also blocked on https://issues.redhat.com/browse/MGMT-18137

Resolves https://issues.redhat.com/browse/MGMT-18131

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- Deployed the operator on a kind cluster
- created an agentserviceconfig and validated all components were created correctly
- created an infraenv and downloaded an ISO to confirm the application was working correctly

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

- [ ] This change does not require a documentation update (docstring, `docs`, README, etc) - added dev docs

## Reviewers Checklist


api/v1beta1/agentserviceconfig_types.go
```diff
@@ -122,6 +122,23 @@ type AgentServiceConfigSpec struct {
...
+	// Has no effect when running on an OpenShift cluster.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Assisted Image Service Ingress hostname"
+	// +optional
+	ImageServiceIngressHost string `json:"imageServiceIngressHost,omitempty"`
+
+	// IngressClassName is the name of the ingress class to be used when configuring ingress resources.
+	// Has no effect when running on an OpenShift cluster.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Ingress Class Name"
+	// +optional
+	IngressClassName string `json:"ingressClassName,omitempty"`
```
carbonin, line 141: Does anyone have an issue with adding these to the top level of agentserviceconfig?
Should we nest them in some kind of kube-specific type?

@avishayt @mhrivnak any thoughts?

CrystalChun (commented):
Tested it out locally on Kind with Nick's instructions and looks good! Will let others weigh in too.

carbonin:
>with Nick's instructions

Sent these to @CrystalChun in a DM, but do you think they're worth documenting separately? Maybe in dev docs somewhere?

CrystalChun:
> > with Nick's instructions
> 
> Sent these to @CrystalChun in a DM, but do you think they're worth documenting separately? Maybe in dev docs somewhere?

Yeah that's a good idea!

eranco74:
Does this handle the imageDownloadURL on the infraEnv CR?

carbonin:
> Does this handle the imageDownloadURL on the infraEnv CR and the asisted-service URL in the agent config?

Yup

api/v1beta1/agentserviceconfig_types.go
mhrivnak: Generally this looks like a good opportunity to nest them. Then it's clear that the fields together represent a specific feature. You either create the nested struct and fill it out appropriately, or you are free to ignore the whole thing. Validation is a bit simpler that way too, especially for inter-field dependencies.

carbonin:
/hold

While we decide on how to define the API

api/v1beta1/agentserviceconfig_types.go
avishayt: I agree, nested

carbonin: How about `IngressConfig` as the field and type name with fields `ClassName`, `ImageServiceHostname`, and `AssistedServiceHostname`?

mhrivnak: Are any of these fields more of a "config" than everything else in the Spec? Assuming not, I'd default to `Ingress.ClassName`, `Ingress.ImageServiceHostname`, etc.

api/v1beta1/agentserviceconfig_types.go
```diff
@@ -122,6 +122,31 @@ type AgentServiceConfigSpec struct {
...
+	// +optional
+	Ingress Ingress `json:"ingress,omitempty"`
+}
+
+type Ingress struct {
+	// AssistedServiceHostname is the hostname to be assigned to the assisted-service ingress.
+	// Has no effect when running on an OpenShift cluster.
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Assisted Service hostname"
+	// +optional
+	AssistedServiceHostname string `json:"assistedServiceHostname,omitempty"`
```
mhrivnak, line 137: Are these really optional? What happens when they're not provided? I see that the string gets plumbed through to the Ingress, so if it's `""`, is that ok?

carbonin: Ah I missed removing that when nesting it. Thanks for catching this.

carbonin:
Fixing test failures in https://github.com/openshift/release/pull/53804

carbonin:
Any more comments on this one @mhrivnak @avishayt ?

docs/dev/operator-on-kind.md
```diff
@@ -0,0 +1,102 @@
...
+
+```yaml
+apiVersion: agent-install.openshift.io/v1beta1
+kind: AgentServiceConfig
+metadata:
+  name: agent
+spec:
+  ingressClassName: nginx
+  assistedServiceIngressHost: assisted-service.example.com
+  imageServiceIngressHost: image-service.example.com
```
jhernand, line 78: I think this isn't in sync with the CRD. Should be like this?

```yaml
ingress:
  className: nginx
  assistedServiceHostname: assisted-service.example.com
  imageServiceHostname: image-service.example.com
```

jhernand:
Would it make sense to use `Ingress` also for OpenShift?

docs/dev/operator-on-kind.md
carbonin: Ah yea, let me update that. Thanks

jhernand:
> I don't think so. At least not right now.
> We'd need to investigate if it handles tls in the same way routes do in the openshift case.
> For now I think it's easier to get something working for kubernetes, then we can followup with deduplication if it makes sense.

I agree, but had ask.

For future reference: OpenShift translates `Ingress` objects into the corresponding `Route` objects. It even copies all the annotations from the ingress to the route. You can use the `route.openshift.io/termination` annotation, for example, to configure the TLS termination. The most relevant difference is that with routes you don't have to specify the host name: it is generated automatically. With ingress it is mandatory. More information about all that here:

https://docs.openshift.com/container-platform/4.16/networking/routes/route-configuration.html#nw-ingress-creating-a-route-via-an-ingress_route-configuration

I think that what would be more complicated if we decide to use ingress for OpenShift would be to migrate existing environments that use routes to use ingress: we would need to explicitly delete the existing routes and then create the new ingress.

carbonin:
>I think that what would be more complicated if we decide to use ingress for OpenShift would be to migrate existing environments that use routes to use ingress: we would need to explicitly delete the existing routes and then create the new ingress.

For sure, and the switch we have now isn't all that bad IMO. Probably not worth trying to migrate.

mhrivnak (approved):

PR #6465, opened 2024-06-24
OCPBUGS-35932: Allow auth in docker credentials to be empty

pastequo:
Allow auth field to be empty in docker pull secret

The impacts are limited to 
* uploader/ package (for on-prem usecase)
* Saas. The code being called by https://github.com/openshift/assisted-service/blob/v2.32.0/internal/cluster/auth.go#L45
  * it only looks for "cloud.openshift.com"
  * this func is only called if the authType is `auth.TypeRHSSO`

I will appreciate to have the opinion of Assisted-Installer gurus to be sure I'm not missing something and it's safe to allow such entry without additional checks. For example, should I add some check here https://github.com/openshift/assisted-service/blob/v2.32.0/internal/cluster/auth.go#L32 or somewhere else to ensure that for SaaS the user:password is specified ?

Also 
  * removed Password that wasn't used https://github.com/openshift/assisted-service/blob/v2.32.0/internal/cluster/validations/pull_secret_validation.go#L51
  * removed credsStore logic, that didn't look functional

credsStore didn't seem functional because:
* The wrong field was parsed (it is at the same level of `auths`, the code was looking at it in `auths.*`. See https://github.com/docker/cli/blob/master/cli/config/configfile/file.go#L28
* The logic looked wrong at first sight, so it probably never worked ?
  * check if one of `credsStore` or `auth` is valuated https://github.com/openshift/assisted-service/blob/v2.32.0/internal/cluster/validations/pull_secret_validation.go#L89
  * if auth is not present take the default value of it (empty string)  - why ? https://github.com/openshift/assisted-service/blob/v2.32.0/internal/cluster/validations/pull_secret_validation.go#L97
  * Assume from here, auth is valuated and try to parse it and extract info from it. This should fail all the time if auth is not present

So I think it's safe to assume that credsStore can be removed. Let me know what you think

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/validations/pull_secret_validation.go
+++ b/internal/cluster/validations/pull_secret_validation.go
@@ -68,7 +68,7 @@ func (e *PullSecretError) Unwrap() error {
 	return e.Cause
 }
 
-// ParsePullSecret validates the format of a pull secret and converts the secret string into individual credentail entries
+// ParsePullSecret validates the format of a pull secret and converts the secret string into individual credential entries
 func ParsePullSecret(secret string) (map[string]PullSecretCreds, error) {
 	result := make(map[string]PullSecretCreds)
 	var s imagePullSecret
@@ -83,11 +83,9 @@ func ParsePullSecret(secret string) (map[string]PullSecretCreds, error) {
 	}
 
 	for d, a := range s.Auths {
-
 		_, authPresent := a["auth"]
-		_, credsStorePresent := a["credsStore"]
-		if !authPresent && !credsStorePresent {
-			return nil, &PullSecretError{Msg: fmt.Sprintf("invalid pull secret: %q JSON-object requires either 'auth' or 'credsStore' field", d)}
+		if !authPresent {
+			return nil, &PullSecretError{Msg: fmt.Sprintf("invalid pull secret: %q JSON-object requires 'auth' field", d)}
 		}
 
 		var authRaw string
--- a/internal/cluster/validations/pull_secret_validation.go
+++ b/internal/cluster/validations/pull_secret_validation.go
@@ -77,47 +77,19 @@ func ParsePullSecret(secret string) (map[string]PullSecretCreds, error) {
 	if err != nil {
 		return nil, &PullSecretError{Msg: "pull secret must be a well-formed JSON", Cause: err}
 	}
-
 	if len(s.Auths) == 0 {
 		return nil, &PullSecretError{Msg: "pull secret must contain 'auths' JSON-object field"}
 	}
 
 	for d, a := range s.Auths {
-		_, authPresent := a["auth"]
-		if !authPresent {
-			return nil, &PullSecretError{Msg: fmt.Sprintf("invalid pull secret: %q JSON-object requires 'auth' field", d)}
-		}
-
-		var authRaw string
-		if auth, ok := a["auth"].(string); authPresent && !ok {
-			return nil, &PullSecretError{Msg: fmt.Sprintf("invalid pull secret: 'auth' field of %q is %v but should be a string", d, a["auth"])}
-		} else {
-			authRaw = auth
-		}
-		data, err := base64.StdEncoding.DecodeString(authRaw)
+		pullSecretCreds, err := parseAuthConfig(d, a)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


gamli75:
@jhernand @carbonin  please review

jhernand:
I agree with the removal of `credsStore`: the implementation was wrong, and it is only meaningful for docker, which we don't really use. It would be nice if that can go in a different pull commit or pull request, would be easier to review.

Allowing an empty password looks good as well. My only objection would be about removing the `Password` field. I think that we don't use it today, but it seems logical to have both the user name and the password, and we may want to use it in the future.

internal/cluster/validations/pull_secret_validation.go
```diff
@@ -83,44 +82,54 @@ func ParsePullSecret(secret string) (map[string]PullSecretCreds, error) {
...
+	}
+
+	ret.AuthRaw = authRaw
+
+	data, err := base64.StdEncoding.DecodeString(authRaw)
+	if err != nil {
+		return ret, &PullSecretError{Msg: fmt.Sprintf("invalid pull secret: 'auth' field of %q is not base64-encoded", registry)}
+	}
+
+	res := bytes.SplitN(data, []byte(":"), 2)
```
jhernand, line 125: I know this code was already here, but it isn't safe to split a string this way. It is encoded in UTF-8 so the user name may (very unlikely) contain multi-byte characters that happen to have the `:` byte as part of their representation. It would be better to convert `data` to a string and then use `strings.SplitN`.

pastequo: I think they can't by construction of UTF-8 (all UTF8 char encoded one more than 1 byte, have their bytes starting by one or several 1. And ':' starts with 0)

But I think it's fair enough to make the change, it makes more sense

jhernand:
> I would be more in favor to add it in the code once we need it. The username field exist not as credentials, but rather to add this information (after hashing it) as metadata in the events. We don't want to send the password (even hashed) in the events. So logically I see it more like "user identification metadata" rather than "user credentials"

I don't have a strong opinion about this. Would have preferred a separate commit or pull request for the removal of `credsStore`, but looks good to me. Lets see what @carbonin thinks before merging.

carbonin (changes requested):
I agree with @jhernand. This PR is doing too much for a bug fix.

I think as a general rule we should keep PRs intended to fix bugs (especially critical ones that may need to be backported) limited to only the required changes to fix the bug.

If you find additional enhancements that could be made along the way please open a followup PR.

pastequo:
@carbonin I generally agree but here the logic are tied (except for Password. Password can definitively be done in another PR)
Consider 
```golang
		_, authPresent := a["auth"]
		_, credsStorePresent := a["credsStore"]
		if !authPresent && !credsStorePresent {
			return nil, &PullSecretError{Msg: fmt.Sprintf("invalid pull secret: %q JSON-object requires either 'auth' or 'credsStore' field", d)}
		}
```

credsStorePresent is always false. So if I keep it, the code will raise an error if "auth" is not present. Meaning for the initial issue

OK:
```bash
[root@k8sbm-1220389 ~]# cat .docker/config.json
{
  "auths": {
    "dockerhub.site.domain.com:8413": {
      "auth": "",
      "email": "you@example.com"
    }
  }
}
```

Not OK:
```bash
[root@k8sbm-1220389 ~]# cat .docker/config.json
{
  "auths": {
    "dockerhub.site.domain.com:8413": {
      "email": "you@example.com"
    }
  }
}
```

And they are supposed to be semantically equivalent

pastequo:
I remove the removal of Password :upside_down_face: . It will come in another PR

@jhernand I splitted the commits into 3 more granular commits. Let me know what you think
@carbonin See comment above

jhernand:
Thanks @pastequo, the separate commits make the change easier to understand and review.

/lgtm

carbonin (approved):
Looks good to me now, thanks for removing those changes and splitting things up.

I wonder though if the original bug would be solved by just not providing an entry in the pull secret at all if your registry doesn't require auth.

PR #6499, opened 2024-06-27
MGMT-18269: Reset finalizing stage in DB while resetting cluster

pastequo:
FinalizingStage is not reset while resetting a cluster, leading to skipping an update for the progress field (https://github.com/openshift/assisted-service/blob/v2.31.7/internal/cluster/cluster.go#L1811) 
Then the timeout (set from the 1st install try) is more likely to be reached, flagging the installation as failed while it's not

Strictly speaking this value can be updated by this route https://github.com/openshift/assisted-service/blob/v2.31.7/swagger.yaml#L333-L338. While the body is required, the internal param is not https://github.com/openshift/assisted-service/blob/v2.31.7/swagger.yaml#L5713-L5717. That means that this API could be called with something empty (and then break the idea of setting it to empty to be able to reset the start time)
But good news, this check here will do the trick https://github.com/openshift/assisted-service/blob/v2.31.7/internal/cluster/cluster.go#L1793

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/transition.go
+++ b/internal/cluster/transition.go
@@ -31,7 +31,9 @@ import (
 
 var resetLogsField = []interface{}{"logs_info", "", "controller_logs_started_at", strfmt.DateTime(time.Time{}), "controller_logs_collected_at", strfmt.DateTime(time.Time{})}
 var resetProgressFields = []interface{}{"progress_finalizing_stage_percentage", 0, "progress_installing_stage_percentage", 0,
-	"progress_preparing_for_installation_stage_percentage", 0, "progress_total_percentage", 0}
+	"progress_preparing_for_installation_stage_percentage", 0, "progress_total_percentage", 0,
+	"progress_finalizing_stage_timed_out", false,
+	"progress_finalizing_stage", "", "progress_finalizing_stage_started_at", strfmt.DateTime(time.Time{})}
 
 var resetFields = append(append(resetProgressFields, resetLogsField...), "openshift_cluster_id", "")
 
--- a/internal/cluster/transition_test.go
+++ b/internal/cluster/transition_test.go
@@ -475,6 +475,9 @@ var _ = Describe("Cancel cluster installation", func() {
 })
 
 var _ = Describe("Reset cluster", func() {
+	timeZero := time.Time{}
+	timeZeroLocal := timeZero.Local()
+
 	var (
 		ctx               = context.Background()
 		dbName            string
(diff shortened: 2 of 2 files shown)


pastequo:
@CrystalChun wdyt ?

internal/cluster/transition.go
```diff
@@ -31,7 +31,9 @@ import (
 
 var resetLogsField = []interface{}{"logs_info", "", "controller_logs_started_at", strfmt.DateTime(time.Time{}), "controller_logs_collected_at", strfmt.DateTime(time.Time{})}
 var resetProgressFields = []interface{}{"progress_finalizing_stage_percentage", 0, "progress_installing_stage_percentage", 0,
-	"progress_preparing_for_installation_stage_percentage", 0, "progress_total_percentage", 0}
+	"progress_preparing_for_installation_stage_percentage", 0, "progress_total_percentage", 0,
+	"progress_finalizing_stage_timed_out", false,
+	"progress_finalizing_stage", "", "progress_finalizing_stage_started_at", strfmt.DateTime(time.Time{})}
```
pastequo, line 36: This is unfortunately not equivalent to something unset but equivalent of what we get from the db when it's unset

internal/cluster/transition_test.go
```diff
@@ -521,14 +524,31 @@ var _ = Describe("Reset cluster", func() {
 		t := t
 		clusterId := strfmt.UUID(uuid.New().String())
 		cluster := common.Cluster{
-			Cluster: models.Cluster{ID: &clusterId, Status: swag.String(t.state)},
+			Cluster: models.Cluster{
+				ID:     &clusterId,
+				Status: swag.String(t.state),
+				Progress: &models.ClusterProgressInfo{
+					FinalizingStage:                         models.FinalizingStageWaitingForClusterOperators,
```
carbonin, line 531: Should we only set these in the test case where we're in `ClusterStatusFinalizing`?

internal/cluster/transition.go
CrystalChun: Looks good 👍

internal/cluster/transition_test.go
CrystalChun: Agreed, let's make sure the fields match the expected stage the cluster is in

PR #6532, opened 2024-07-04
MGMT-17893: Don't destroy cluster on detach

jhernand:
Currently the procedure to detach a cluster that was created using hosts from a late binding pool is to first delete the `ManagedCluster` object, then add the `preserveOnDelete: true` field to the `ClusterDeployment` and then delete that `ClusterDeployment`. But the cluster deployment controller doesn't look at the `preserveOnDelete` field at all. As a result the hosts of the cluster are returned to the pool and they are provisioned again, which effectively destroys the cluster.

This patch changes the deployment manager controller so that in that case it will check the `preserveOnDelete` field and if it is `true` it will delete the corresponding `Agent` objects. The result of that is that the hosts will go back to the pool, but marked the will still be marked as provisioned and they will not be provisioned again, thus avoiding the destruction of the cluster.

Note that the `BareMetalHosts` will not be removed, but they will stay detached.

Related: https://issues.redhat.com/browse/MGMT-17893

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

Tested manually with MCE 2.5.4, replacing the assisted service image with a modified one containing this change.

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/clusterdeployments_controller_test.go
+++ b/internal/controller/controllers/clusterdeployments_controller_test.go
@@ -4157,6 +4157,15 @@ var _ = Describe("unbindAgents", func() {
 		}
 		Expect(c.Create(ctx, infraEnv)).To(Succeed())
 
+		clusterDeployment := &hivev1.ClusterDeployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: clusterReference.Namespace,
+				Name:      clusterReference.Name,
+			},
+			Spec: hivev1.ClusterDeploymentSpec{},
+		}
+		Expect(c.Create(ctx, clusterDeployment)).To(Succeed())
+
 		cdName := types.NamespacedName{
 			Name:      clusterReference.Name,
 			Namespace: clusterReference.Namespace,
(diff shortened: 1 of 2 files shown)


internal/controller/controllers/clusterdeployments_controller_test.go
```diff
@@ -4198,6 +4207,54 @@ var _ = Describe("unbindAgents", func() {
 		err = c.Get(ctx, types.NamespacedName{Name: "test-agent2", Namespace: testNamespace}, agent)
 		Expect(k8serrors.IsNotFound(err)).To(BeTrue())
 	})
+
+	It("deletes late bound agents when cluster deployment is preserved", func() {
+		infraEnv := &aiv1beta1.InfraEnv{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "InfraEnv",
+				APIVersion: fmt.Sprintf("%s/%s", aiv1beta1.GroupVersion.Group, aiv1beta1.GroupVersion.Version),
```
carbonin, line 4215: You generally shouldn't need to set TypeMeta like this. Is there a reason you're doing it?

internal/controller/controllers/clusterdeployments_controller_test.go
```diff
@@ -4198,6 +4207,54 @@ var _ = Describe("unbindAgents", func() {
...
+		clusterDeployment := &hivev1.ClusterDeployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: clusterReference.Namespace,
+				Name:      clusterReference.Name,
+			},
+			Spec: hivev1.ClusterDeploymentSpec{
+				PreserveOnDelete: true,
+			},
+		}
+		Expect(c.Create(ctx, clusterDeployment)).To(Succeed())
```
carbonin, line 4234: Instead of recreating these could you use the existing ones already defined in the test setup?

carbonin (commented):
Looks good. Just a few minor test questions.

internal/controller/controllers/clusterdeployments_controller_test.go
jhernand: Copy & paste, from the test above. Will remove it.

internal/controller/controllers/clusterdeployments_controller_test.go
jhernand: I don't see this created by the test setup, rather individual tests create it explicitly. I can change this to use the `newClusterDeployment` function, if that is what you mean.

jhernand: It would be something like this, which isn't much better:

```diff
diff --git a/internal/controller/controllers/clusterdeployments_controller_test.go b/internal/controller/controllers/clusterdeployments_controller_test.go
index ed612a28e..ba71eaba9 100644
--- a/internal/controller/controllers/clusterdeployments_controller_test.go
+++ b/internal/controller/controllers/clusterdeployments_controller_test.go
@@ -4218,15 +4218,13 @@ var _ = Describe("unbindAgents", func() {
                }
                Expect(c.Create(ctx, infraEnv)).To(Succeed())
 
-               clusterDeployment := &hivev1.ClusterDeployment{
-                       ObjectMeta: metav1.ObjectMeta{
-                               Namespace: clusterReference.Namespace,
-                               Name:      clusterReference.Name,
-                       },
-                       Spec: hivev1.ClusterDeploymentSpec{
+               clusterDeployment := newClusterDeployment(
+                       clusterReference.Name,
+                       clusterReference.Namespace,
+                       hivev1.ClusterDeploymentSpec{
                                PreserveOnDelete: true,
                        },
-               }
+               )
                Expect(c.Create(ctx, clusterDeployment)).To(Succeed())
 
                agent := &aiv1beta1.Agent{
```

carbonin: Ah sorry I misread the setup. I guess it's only creating the agent. What you have is fine then. You could use one of the agents that was created in the setup, but it's not a big deal.

jhernand: Done, the test uses now the agents created in the setup.

carbonin (approved):

PR #6533, opened 2024-07-04
MGMT-17478: Use ICC config secret for converged flow info when available

adriengentil:
When available, read the ICC secret set by BMO in order to retrieve the
agent image, the agent URL(s), and the inspector URL(s).

This this change, we wil try to:
1/ Get the agent image from the user override
2/ if 1/ is not found, get the image from the ICC configuration
3/ if 2/ is not found, get the image from the HUB cluster
4/ if 3/ is not found, return the default image

### DIFF
--- a/internal/controller/controllers/bmo_utils.go
+++ b/internal/controller/controllers/bmo_utils.go
@@ -31,7 +31,7 @@ const (
 type BMOUtils interface {
 	ConvergedFlowAvailable() bool
 	GetIronicIPs() ([]string, []string, error)
-	GetICCConfig() (*ICCConfig, error)
+	GetICCConfig(ctx context.Context) (*ICCConfig, error)
 }
 
 type ICCConfig struct {
@@ -106,12 +106,12 @@ func (r *bmoUtils) GetIronicIPs() ([]string, []string, error) {
 	return ironicIPs, inspectorIPs, nil
 }
 
-func (r *bmoUtils) GetICCConfig() (*ICCConfig, error) {
+func (r *bmoUtils) GetICCConfig(ctx context.Context) (*ICCConfig, error) {
 	const configKeyNotFoundError string = "Failed to get '%s' key from secret %s in namespace %s"
 
 	secret := &corev1.Secret{}
 	namespacedName := types.NamespacedName{Name: iccSecretName, Namespace: iccNamespace}
-	if err := r.c.Get(context.TODO(), namespacedName, secret); err != nil {
+	if err := r.c.Get(ctx, namespacedName, secret); err != nil {
 		return nil, fmt.Errorf("Failed to get secret %s in namespace %s: %w", iccSecretName, iccNamespace, err)
 	}
 
--- a/internal/controller/controllers/bmo_utils.go
+++ b/internal/controller/controllers/bmo_utils.go
@@ -12,17 +12,32 @@ import (
 	"github.com/openshift/cluster-baremetal-operator/provisioning"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const MinimalVersionForConvergedFlow = "4.12.0-0.alpha"
+const (
+	MinimalVersionForConvergedFlow = "4.12.0-0.alpha"
+	iccNamespace                   = "openshift-machine-api"
+	iccSecretName                  = "metal3-image-customization-config" // #nosec G101
+	ironicBaseURLKey               = "IRONIC_BASE_URL"
+	ironicInspectorBaseURLKey      = "IRONIC_INSPECTOR_BASE_URL"
+	ironicAgentImageKey            = "IRONIC_AGENT_IMAGE"
+)
 
 //go:generate mockgen --build_flags=--mod=mod -package=controllers -destination=mock_bmo_utils.go . BMOUtils
 type BMOUtils interface {
 	ConvergedFlowAvailable() bool
 	GetIronicIPs() ([]string, []string, error)
+	GetICCConfig() (*ICCConfig, error)
+}
+
+type ICCConfig struct {
+	IronicBaseURL          string
+	IronicInspectorBaseUrl string
+	IronicAgentImage       string
 }
 
 type bmoUtils struct {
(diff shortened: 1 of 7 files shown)


adriengentil:
@carbonin can you have a first look? :pray: 
I use the agent/inspector URL from the ICC config only if we use the agent image from the ICC config, I was thinking that the format of the URLs may be tied to some version of the agent image (for example, we can have a list of an IPv4 and an IPv6 in the URL field). Does it makes sense... or not?

internal/controller/controllers/bmo_utils.go
```diff
@@ -91,6 +106,37 @@ func (r *bmoUtils) GetIronicIPs() ([]string, []string, error) {
 	return ironicIPs, inspectorIPs, nil
 }
 
+func (r *bmoUtils) GetICCConfig() (*ICCConfig, error) {
+	const configKeyNotFoundError string = "Failed to get '%s' key from secret %s in namespace %s"
+
+	secret := &corev1.Secret{}
+	namespacedName := types.NamespacedName{Name: iccSecretName, Namespace: iccNamespace}
+	if err := r.c.Get(context.TODO(), namespacedName, secret); err != nil {
```
carbonin, line 114: Is there a good reason to not pass a context to this function?

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -409,17 +411,42 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
...
+	var iccConfig *ICCConfig
+	if ironicAgentImage == "" {
+		iccConfig, err = r.BMOUtils.GetICCConfig()
+		if err != nil {
+			log.WithError(err).Info("ICC configuration is not available, will continue without it")
+		}
+	}
+	if iccConfig != nil {
+		var architectures []string
+		architectures, err = r.OcRelease.GetReleaseArchitecture(log, iccConfig.IronicAgentImage, r.ReleaseImageMirror, infraEnvInternal.PullSecret)
```
carbonin, line 431: Does this work? I'm surprised this isn't specific to release images.

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -409,17 +411,42 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
 		log.WithError(err).Error("failed to get corresponding infraEnv")
 		return false, err
 	}
+
+	// 1 - Check if agent image image is overriden
```
carbonin, line 415: Can we maybe put all of this tiered logic into a separate function?

Maybe `getIronicConfig` and return the new ironic config struct even if we get the data from one of the existing sources?

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -482,6 +511,7 @@ func (r *PreprovisioningImageReconciler) getIronicServiceURLs(ctx context.Contex
 
 	// default to the first IP returned
 	// v4 for dualstack hub or whatever family the single stack is
+
```
carbonin, line 514: Bonus space?

internal/controller/controllers/preprovisioningimage_controller.go
adriengentil: I did not get any issues in my cluster so far. I'll double check.

adriengentil: The [code](https://github.com/openshift/assisted-service/blob/master/internal/oc/release.go#L249) basically does a `oc image info --output json quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:e3813103687092e9016edf7d40ea7394d8cddbc773a750f5a67317f751d3eee7 --registry-config ~/pull-secret | jq .config.architecture` (it's a bit more complicated if the image is multi-arch as is uses skopeo).

I didn't realize that `Release` in function name meant "OCP release", but the code behind it looks quite generic.

carbonin: Yeah, that was supposed to be for release images ... I mean I suppose this is an image in the release so it _could_ make sense. I just worry that in the future someone might change this function so that it does only work for release images and break this use.

What do you think about creating a new function that just calls the old one called `GetImageArchitecture`? Maybe we can then fully move over to the new name in a followup?

adriengentil: I agree, I thought about the same thing this morning.

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -396,6 +396,104 @@ func (r *PreprovisioningImageReconciler) mapInfraEnvPPI() func(ctx context.Conte
...
+	log.Infof("CPU architecture (%v) of Ironic agent image (%s) from ICC config is not available for infraEnv %v with arch (%s)",
+		architectures,
+		iccConfig.IronicAgentImage,
+		infraEnvInternal.Name,
+		infraEnvInternal.CPUArchitecture)
+	return nil
+}
+
+func (r *PreprovisioningImageReconciler) getIronicConfigFromHUB(ctx context.Context, log logrus.FieldLogger, infraEnvInternal *common.InfraEnv) *ICCConfig {
+	ironicAgentImage, err := r.getIronicAgentImageByRelease(ctx, log, infraEnvInternal)
```
carbonin, line 435: Assuming `getIronicAgentImageByRelease` isn't used anywhere else you can combine these two functions. Not sure there's any reason to have both.

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -396,6 +396,104 @@ func (r *PreprovisioningImageReconciler) mapInfraEnvPPI() func(ctx context.Conte
...
+		ironicAgentImage = r.Config.BaremetalIronicAgentImage
+	}
+
+	log.Infof("Setting default ironic agent image (%s) for infraEnv %v", ironicAgentImage, infraEnvInternal.Name)
+	return &ICCConfig{
+		IronicAgentImage: ironicAgentImage,
+	}
+}
+
+func (r *PreprovisioningImageReconciler) getIronicConfig(ctx context.Context, log logrus.FieldLogger, infraEnv *aiv1beta1.InfraEnv, infraEnvInternal *common.InfraEnv) (*ICCConfig, error) {
```
carbonin, line 461: No change needed, I'm just really happy with how much easier this is to read now. :tada:

internal/controller/controllers/preprovisioningimage_controller.go
adriengentil: merged both methods!

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -396,9 +396,133 @@ func (r *PreprovisioningImageReconciler) mapInfraEnvPPI() func(ctx context.Conte
...
+		iccConfig.IronicAgentImage,
+		iccConfig.IronicBaseURL,
+		iccConfig.IronicInspectorBaseUrl)
+	return iccConfig, nil
+}
+
 // AddIronicAgentToInfraEnv updates the infra-env in the database with the ironic agent ignition config if required
 // returns true when the infra-env was updated, false otherwise
 func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Context, log logrus.FieldLogger, infraEnv *aiv1beta1.InfraEnv) (bool, error) {
+	log = log.WithFields(logrus.Fields{
```
carbonin, line 521: Should we maybe do this earlier for all functions as soon as we know the infraenv?

internal/controller/controllers/bmo_utils.go
```diff
@@ -91,6 +106,37 @@ func (r *bmoUtils) GetIronicIPs() ([]string, []string, error) {
 	return ironicIPs, inspectorIPs, nil
 }
 
+func (r *bmoUtils) GetICCConfig(ctx context.Context) (*ICCConfig, error) {
```
carbonin, line 109: This doesn't need to be exported if we're only using it from this package.

internal/controller/controllers/preprovisioningimage_controller.go
adriengentil: changed it!

internal/controller/controllers/bmo_utils.go
adriengentil: I changed it, not sure if having a private method in BMOUtils interface is very idiomatic, but it looks like it's mainly here to get a mock out of it. Let me know if we should do differently :thinking:

adriengentil:
/retest
equinix issues

adriengentil:
/retest
prow is flacky

internal/controller/controllers/bmo_utils.go
```diff
@@ -91,6 +106,37 @@ func (r *bmoUtils) GetIronicIPs() ([]string, []string, error) {
...
+	}
+
+	ironicBaseURL, ok := secret.Data[ironicBaseURLKey]
+	if !ok {
+		return nil, fmt.Errorf(configKeyNotFoundError, ironicBaseURLKey, secret.Name, secret.Namespace)
+	}
+
+	ironicInspectorBaseURL, ok := secret.Data[ironicInspectorBaseURLKey]
+	if !ok {
+		return nil, fmt.Errorf(configKeyNotFoundError, ironicInspectorBaseURLKey, secret.Name, secret.Namespace)
```
dtantsur, line 125: Late to this party, sorry. This is not quite correct: the inspector URL can and actually will be missing.

adriengentil: In that case, how we get the inspector URL? From the preprovisioning info?

adriengentil: Ok, if understand well, we can leave the inspector URL empty, and it will be auto-magically filled using the ICC lib https://github.com/openshift/image-customization-controller/commit/26ce5d6c64ee4ff120a8f7f56eb44397ca7661f1

dtantsur: This is correct.

carbonin: :confused: but we don't use any of the icc stuff, right?

adriengentil: we use it here: https://github.com/openshift/assisted-service/blob/8703b59882a2f9d5730ece7c2c47cb025ab42a8f/internal/ignition/ironic.go#L22

PR #6546, opened 2024-07-08
MGMT-18160: Use cert-manager to handle tls certs for webhooks

carbonin:
When not running in an OpenShift cluster we don't have access to [service serving certs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.15/html/security_and_compliance/configuring-certificates#understanding-service-serving_service-serving-certificate) and [our webhooks _require_ https configuration](https://github.com/openshift/generic-admission-server/blob/8dcc3c9b298fefaf4d31c60001907222bf2c3f83/README.md#why-cant-i-write-a-simple-http-webhook-server) to run.

This PR uses cert-manager to configure keys and certs for our webhook services and enables them in non-OpenShift environments.

As a followup cert-manager will also be used to create certificates to handle https ingress as well as communication between services.

https://issues.redhat.com/browse/MGMT-18160

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Deployed an operator image built from this branch to a kind server and verified the webhooks worked as expected.

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


carbonin:
/hold

Looking into the test failures

carbonin:
Test failures are due to the controller-runtime version bump.
Still working on a fix.

internal/controller/controllers/hypershiftagentserviceconfig_controller.go
```diff
@@ -189,8 +189,18 @@ func (hr *HypershiftAgentServiceConfigReconciler) Reconcile(origCtx context.Cont
 		return ctrl.Result{}, nil
 	}
 
+	supportsCertManager := false
+	if !asc.rec.IsOpenShift {
+		var certErr error
+		supportsCertManager, certErr = serverSupportsCertManager(ctx, asc.Client)
+		if err != nil {
```
CrystalChun, line 196:
```suggestion
		if certErr != nil {
```
Should this check for `certErr`?

internal/controller/controllers/kube_compat.go
```diff
@@ -18,3 +30,168 @@ func ServerIsOpenShift(ctx context.Context, c client.Client) (bool, error) {
...
+	return []component{
+		{"SelfSignedIssuer", aiv1beta1.ReasonCertificateFailure, newSelfSignedIssuer},
+		{"CAIssuer", aiv1beta1.ReasonCertificateFailure, newCAIssuer},
+		{"CACert", aiv1beta1.ReasonCertificateFailure, newCACert},
+		{"WebhookCert", aiv1beta1.ReasonCertificateFailure, newWebhookCert},
+	}
+}
+
+// newSelfSignedIssuer describes how to create an Issuer for the self-signed root issuer
+func newSelfSignedIssuer(ctx context.Context, log logrus.FieldLogger, asc ASC) (client.Object, controllerutil.MutateFn, error) {
```
CrystalChun, line 95: very nit: can all the names be constants?
E.g. `Name: "assisted-installer-selfsigned-ca",`

carbonin: Yeah, I was debating this so I don't care much either way. I'll change it

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -380,6 +389,8 @@ func getComponents(spec *aiv1beta1.AgentServiceConfigSpec, isOpenshift bool) []c
 				component{"AgentIPXERoute", aiv1beta1.ReasonAgentRouteFailure, newAgentIPXERoute},
 			)
 		}
+	} else {
+		components = append(components, certManagerComponents()...)
```
rccrdpccl, line 393: If we extract this we can return early and avoid the nesting and else condition:

```golang
components = append(components, getExtraComponents(spec))
```

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -2716,28 +2693,36 @@ func validate(ctx context.Context, log logrus.FieldLogger, asc ASC) (bool, error
...
+					Reason:  reason,
+					Message: message,
+				},
+			)
+			err = asc.Client.Status().Update(ctx, asc.Object)
+			if err != nil {
+				log.WithError(err).Error("Failed to update status")
+				return false, err
+			}
+			return false, nil
```
rccrdpccl, line 2724: If we extract this error handling we can return earlier and avoid else condition, i.e.
```golang
if asc.spec.Ingress == nil || asc.spec.Ingress.AssistedServiceHostname == "" || asc.spec.Ingress.ImageServiceHostname == "" {
		return handleError("ingress configuration is required for non-OpenShift deployment", aiv1beta1.ReasonKubernetesIngressMissing)
}
if !supportsCertManager {
		return handleError("cert-manager is a required dependency for non-OpenShift deployments", aiv1beta1.ReasonCertificateFailure)
}
```

internal/controller/controllers/kube_compat.go
```diff
@@ -18,3 +33,168 @@ func ServerIsOpenShift(ctx context.Context, c client.Client) (bool, error) {
...
 	return false, client.IgnoreNotFound(err)
 }
+
+func serverSupportsCertManager(ctx context.Context, c client.Client) (bool, error) {
+	certCRD := apiextensionsv1.CustomResourceDefinition{}
+	err := c.Get(ctx, types.NamespacedName{Name: certificateCRDName}, &certCRD)
+	if err == nil {
+		return true, nil
+	}
+	return false, client.IgnoreNotFound(err)
```
rccrdpccl, line 43: nit: this could probably be inlined by reversing the condition

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: Not sure I see the benefit. Then `getExtraComponents` will still have the `if isOpenshift ... else ...`, right? Plus the logic isn't particularly complicated and the new function wouldn't be reused so it's not something I would think of breaking out for these reasons.

internal/controller/controllers/agentserviceconfig_controller.go
carbonin: Again not sure I see the upside in extracting to a function like `handleError` that only handles errors for this one particular part of validation.

I could be convinced to extract this entire block into something like `validateForKube` if you think this `validate` function is just becoming too complex.

internal/controller/controllers/kube_compat.go
carbonin: Not sure what you mean but I guess something like this would work:
```suggestion
	err := c.Get(ctx, types.NamespacedName{Name: certificateCRDName}, &certCRD)
	return err == nil, client.IgnoreNotFound(err)
```

Is that what you're suggesting?

internal/controller/controllers/kube_compat.go
```diff
@@ -2,19 +2,195 @@ package controllers
...
+	return crdExists(ctx, c, certificateCRDName)
+}
+
+func crdExists(ctx context.Context, c client.Client, crdName string) (bool, error) {
+	crd := apiextensionsv1.CustomResourceDefinition{}
+	err := c.Get(ctx, types.NamespacedName{Name: crdName}, &crd)
+	return err == nil, client.IgnoreNotFound(err)
+}
+
+func newIngress(ctx context.Context, log logrus.FieldLogger, asc ASC, name string, host string, port int32) (client.Object, controllerutil.MutateFn, error) {
```
CrystalChun, line 42: Does `ctx context.Context, log logrus.FieldLogger` need to be in the argument list? :o

internal/controller/controllers/kube_compat.go
```diff
@@ -2,19 +2,195 @@ package controllers
...
+	return []component{
+		{"SelfSignedIssuer", aiv1beta1.ReasonCertificateFailure, newSelfSignedIssuer},
+		{"CAIssuer", aiv1beta1.ReasonCertificateFailure, newCAIssuer},
+		{"CACert", aiv1beta1.ReasonCertificateFailure, newCACert},
+		{"WebhookCert", aiv1beta1.ReasonCertificateFailure, newWebhookCert},
+	}
+}
+
+// newSelfSignedIssuer describes how to create an Issuer for the self-signed root issuer
+func newSelfSignedIssuer(ctx context.Context, log logrus.FieldLogger, asc ASC) (client.Object, controllerutil.MutateFn, error) {
```
CrystalChun, line 94: Same here, does `ctx context.Context, log logrus.FieldLogger` need to be in the argument list? :o

internal/controller/controllers/kube_compat.go
```diff
@@ -2,19 +2,195 @@ package controllers
...
+				SelfSigned: &certtypes.SelfSignedIssuer{},
+			},
+		}
+		return nil
+	}
+	return is, mutateFn, nil
+}
+
+// newCACert describes how to create a Certificate for the root CA
+func newCACert(ctx context.Context, log logrus.FieldLogger, asc ASC) (client.Object, controllerutil.MutateFn, error) {
```
CrystalChun, line 116: Same here, does `ctx context.Context, log logrus.FieldLogger` need to be in the argument list? :o

internal/controller/controllers/kube_compat.go
```diff
@@ -2,19 +2,195 @@ package controllers
...
+			fmt.Sprintf("%s.%s.svc", serviceName, asc.namespace),
+			fmt.Sprintf("%s.%s.svc.cluster.local", serviceName, asc.namespace),
+		}
+		return nil
+	}
+
+	return cert, mutateFn, nil
+}
+
+func newWebhookCert(ctx context.Context, log logrus.FieldLogger, asc ASC) (client.Object, controllerutil.MutateFn, error) {
```
CrystalChun, line 194: Same here, does `ctx context.Context, log logrus.FieldLogger` need to be in the argument list? :o

internal/controller/controllers/kube_compat.go
```diff
@@ -2,19 +2,195 @@ package controllers
...
+			Name: selfSignedIssuerName,
+		}
+		return nil
+	}
+
+	return cert, mutateFn, nil
+}
+
+// newCAIssuer describes how to create an Issuer for creating certificates signed by the root CA cert
+func newCAIssuer(ctx context.Context, log logrus.FieldLogger, asc ASC) (client.Object, controllerutil.MutateFn, error) {
```
CrystalChun, line 142: Same here, does `ctx context.Context, log logrus.FieldLogger` need to be in the argument list? :o

carbonin: Yes, because this is a [`NewComponentFn`](https://github.com/openshift/assisted-service/blob/d2bba26c003c543612de88883f936d54aceb4ca2/internal/controller/controllers/agentserviceconfig_controller.go#L187)

internal/controller/controllers/kube_compat.go
carbonin: Yes, because this is a [`NewComponentFn`](https://github.com/openshift/assisted-service/blob/d2bba26c003c543612de88883f936d54aceb4ca2/internal/controller/controllers/agentserviceconfig_controller.go#L187)

internal/controller/controllers/kube_compat.go
carbonin: Yes, because this is a [`NewComponentFn`](https://github.com/openshift/assisted-service/blob/d2bba26c003c543612de88883f936d54aceb4ca2/internal/controller/controllers/agentserviceconfig_controller.go#L187)

internal/controller/controllers/kube_compat.go
carbonin: Yes, because this is a [`NewComponentFn`](https://github.com/openshift/assisted-service/blob/d2bba26c003c543612de88883f936d54aceb4ca2/internal/controller/controllers/agentserviceconfig_controller.go#L187)

internal/controller/controllers/kube_compat.go
carbonin: Nope, I suppose not. I'll remove them

internal/controller/controllers/kube_compat.go
CrystalChun: Ah gotcha!

CrystalChun (approved):
/lgtm

internal/controller/controllers/kube_compat.go
rccrdpccl: I meant
```golang
if err := c.Get(ctx, types.NamespacedName{Name: certificateCRDName}, &certCRD); err != nil {
    return false, client.IgnoreNotFound(err)
}
return return true, nil
```

internal/controller/controllers/agentserviceconfig_controller.go
rccrdpccl: This function is almost 100 lines and the "handle error block" is duplicated in lines 2679-2693
Extracting this function would make this validate function way more readable IMO.
It's not too complex, but we would make it easier to follow by exiting earlier and factoring out common verbose code.

PR #6564, opened 2024-07-11
MGMT-18305: Add certs to ingress when deploying in non-OCP clusters

carbonin:
Resolves https://issues.redhat.com/browse/MGMT-18305

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agentserviceconfig_controller.go
+++ b/internal/controller/controllers/agentserviceconfig_controller.go
@@ -1061,8 +1061,8 @@ func newImageServiceConfigMap(ctx context.Context, log logrus.FieldLogger, asc A
 func urlForRoute(ctx context.Context, asc ASC, routeName string) (string, error) {
 	var hostname, scheme string
 
+	scheme = "https"
 	if asc.rec.IsOpenShift {
-		scheme = "https"
 		route := &routev1.Route{}
 		err := asc.Client.Get(ctx, types.NamespacedName{Name: routeName, Namespace: asc.namespace}, route)
 		if err != nil || route.Spec.Host == "" {
@@ -1076,7 +1076,6 @@ func urlForRoute(ctx context.Context, asc ASC, routeName string) (string, error)
 		if asc.spec.Ingress == nil {
 			return "", fmt.Errorf("ingress config is required for non-OpenShift deployments")
 		}
-		scheme = "http"
 		switch routeName {
 		case serviceName:
 			hostname = asc.spec.Ingress.AssistedServiceHostname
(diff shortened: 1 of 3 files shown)


internal/controller/controllers/kube_compat.go
```diff
@@ -39,6 +40,10 @@ func crdExists(ctx context.Context, c client.Client, crdName string) (bool, erro
 	return err == nil, client.IgnoreNotFound(err)
 }
 
+func ingressTLSSecretName(ingressName string) string {
+	return fmt.Sprintf("%s-ingress", ingressName)
```
rccrdpccl, line 44: Where is this naming convention coming from? Not very important, but if we have control over id, should we add something to the name that would make it recognizable that's about TLS specifically? i.e. `%s-ingress-tls`

rccrdpccl:
/hold
/lgtm
/approve

Just minor comment, please unhold if we can't/won't address it right now

internal/controller/controllers/kube_compat.go
carbonin: >Where is this naming convention coming from?

Just what I happened to think of.

>should we add something to the name that would make it recognizable that's about TLS specifically? i.e. %s-ingress-tls

The secrets are created with type tls so I don't think we really need to repeat it in the name. Plus I'm not sure what other ingress-related secrets we're planning to create:

```
$ kubectl -n assisted-installer get secret
assisted-image-service-ingress   kubernetes.io/tls   3      2d23h
assisted-service-ingress         kubernetes.io/tls   3      2d23h
...
```

internal/controller/controllers/agentserviceconfig_controller.go
```diff
@@ -1195,16 +1194,18 @@ func newAssistedCM(ctx context.Context, log logrus.FieldLogger, asc ASC) (client
...
 			"DATA_UPLOAD_ENDPOINT":   "https://console.redhat.com/api/ingress/v1/upload",
 
-			"NAMESPACE":       asc.namespace,
-			"INSTALL_INVOKER": "assisted-installer-operator",
+			"NAMESPACE":              asc.namespace,
+			"INSTALL_INVOKER":        "assisted-installer-operator",
+			"SKIP_CERT_VERIFICATION": "False",
 		}
-		// enable https only on OCP
+		// serve https only on OCP
```
CrystalChun, line 1201: Sorry, for my understanding - is this comment still true if we're always enabling https here https://github.com/openshift/assisted-service/pull/6564/files#diff-c619bb36181b5e1dd76fb1fcc1f54068b16a8761f77106052fcbf5e2f5f9d356R1064

carbonin: Yeah, so the ingress will be https, but tls will terminate there. Meaning assisted-service will serve http between itself and the ingress controller.

That's why I changed this comment to say that we're enabling https in that the agent will expect https and we should provide the agent with certs to trust the ingress, but the assisted service itself is not serving https.

CrystalChun (approved):
/lgtm

internal/controller/controllers/agentserviceconfig_controller.go
CrystalChun: Thanks for the explanation!

PR #6591, opened 2024-07-18
MGMT-18411: Add NTP sources to discovery environment

jhernand:
Currently when the user provides additional NTP sources for the installation they are used by the installed cluster, but not by the discovery environment. That means that the system clock of the discovery environment may be wrong if the default NTP servers can't be reached, and as result it may fail to pull images because the system clock may be before the validity date of the certificates of the registry server. To avoid that this patch changes the ignition of the discovery environment so that it will also use the NTP sources provided by the user.

Related: https://issues.redhat.com/browse/MGMT-18411

- [ ] New Feature 
- [X] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Tested creating a libvirt virtual machine that has the clock fixed in the past:

```bash
# date -d '01/01/2000' +"%s"
946681200
```

```xml
<clock offset='absolute' start='946681200'>
  ...
</clock>
```

The blocked access to all NTP servers, except one which is not in the default pool used by CoreOS:

```bash
# iptables --table filter --insert FORWARD 1 --protocol udp --dport 123 --jump REJECT
# iptables --table filter --insert FORWARD 1 --destination 162.159.200.123 --protocol udp --dport 123 --jump ACCEPT
```

Then verified that when the NTP sources aren't specified the discovery environment has the date incorrect, and when 162.159.200.123 is specified it is added to `/etc/chrony.conf` and the date is set correctly.

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/ignition/discovery.go
+++ b/internal/ignition/discovery.go
@@ -242,6 +242,13 @@ func (ib *ignitionBuilder) FormatDiscoveryIgnitionFile(ctx context.Context, infr
 		return "", err
 	}
 
+	// If the list of additional NTP sources is empty then we want to pass an empty list to the
+	// template, but the Split method returns a slice with one empty element in that case.
+	additionalNtpSources := strings.Split(infraEnv.AdditionalNtpSources, ",")
+	if len(additionalNtpSources) == 1 && additionalNtpSources[0] == "" {
+		additionalNtpSources = []string{}
+	}
+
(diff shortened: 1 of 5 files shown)


CrystalChun (approved):
/lgtm

PR #6602, opened 2024-07-22
MGMT-17560: Append kargs for iSCSI boot

adriengentil:
Set DHCP on the network interface used for the iSCSI network when
installing RHCOS on the disk. This is required to allow boot from an
iSCSI volume.


internal/host/hostcommands/install_cmd.go
```diff
@@ -7,6 +7,7 @@ import (
 	"net"
 	"strings"
 
+	"github.com/asaskevich/govalidator"
```
eifrach, line 10: this was last released in 2021 - and not look like it been mantained in a while 

maybe it's better to use the builtin net package
https://pkg.go.dev/net/netip#Addr.Is4

```go
package main

import (
	"fmt"
	"net/netip"
)

func main() {
	addrString := "192.168.1.1"
	addr, err := netip.ParseAddr(addrString)
	if err != nil {
		fmt.Println("Error parsing")
	}
	fmt.Printf("%v", addr.Is4())
	// output: true
	fmt.Printf("%v", addr.Is6())
	// output: false

}

```

eifrach (changes requested):

internal/host/hostcommands/install_cmd.go
```diff
@@ -305,15 +308,22 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
...
-			if disk.DriveType == models.DriveTypeMultipath {
-				installerArgs = append(installerArgs, "--append-karg", "root=/dev/disk/by-label/dm-mpath-root", "--append-karg", "rw", "--append-karg", "rd.multipath=default")
-			} else if disk.DriveType == models.DriveTypeISCSI {
-				// Currently only allowed on the OCI platform
-				installerArgs = append(installerArgs, "--append-karg", "rd.iscsi.firmware=1")
-			}
+			installationDisk = disk
+			break
+		}
+	}
```
eifrach, line 317: 1.  what happens if no installation disk has been found? 
2. maybe we should create a common function to get installation disk and error like `common.GetInstallatioDisk(disks []models.Disk) models.Disk error`

internal/host/hostcommands/install_cmd.go
```diff
@@ -305,15 +308,22 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
 		}
 	}
 
+	var installationDisk *models.Disk
 	for _, disk := range inventory.Disks {
 		if disk.ID == host.InstallationDiskID {
```
adriengentil, line 313: we have a helper to get the installation disk, let's use it

internal/host/hostcommands/install_cmd.go
adriengentil: old behavior was just ignoring the missing installation disk. But a warning would be nice

internal/host/hostcommands/install_cmd.go
```diff
@@ -305,15 +308,22 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
...
+
+	if installationDisk != nil && installationDisk.DriveType == models.DriveTypeMultipath {
+		installerArgs = append(installerArgs, "--append-karg", "root=/dev/disk/by-label/dm-mpath-root", "--append-karg", "rw", "--append-karg", "rd.multipath=default")
+	} else if installationDisk != nil && installationDisk.DriveType == models.DriveTypeISCSI {
+		installerArgs, err = appendISCSIArgs(installerArgs, installationDisk, inventory, hasUserConfiguredIP)
+		if err != nil {
+			return "", err
 		}
+
 	}
```
eifrach, line 327: first check only once if the disk is nil - no need to go throw another if 

```suggestion
	if installationDisk != nil {
		if installationDisk.DriveType == models.DriveTypeMultipath {
			installerArgs = append(installerArgs, "--append-karg", "root=/dev/disk/by-label/dm-mpath-root", "--append-karg", "rw", "--append-karg", "rd.multipath=default")
		} else if  installationDisk.DriveType == models.DriveTypeISCSI {
			installerArgs, err = appendISCSIArgs(installerArgs, installationDisk, inventory, hasUserConfiguredIP)
			if err != nil {
				return "", err
			}
		}
	}
```

adriengentil: will propably refactor to avoid too many levels of ifs

internal/host/hostcommands/install_cmd.go
```diff
@@ -325,6 +335,45 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
...
+
+	if nic == nil {
+		return nil, fmt.Errorf("Cannot find the interface belonging to iSCSI host IP %s", iSCSIHostIP.String())
+	}
+
+	dhcp := "dhcp"
+	if isIPv6 {
+		dhcp = "dhcp6"
+	}
+	installerArgs = append(installerArgs, "--append-karg", fmt.Sprintf("ip=%s:%s", nic.Name, dhcp))
```
eifrach, line 372: not sure, but please validate we don't append multiple statements of "--append-karg"
lines 340, 321, 372

internal/host/hostcommands/install_cmd.go
```diff
@@ -365,11 +414,6 @@ func appends390xArgs(inventory *models.Inventory, installerArgs []string, log lo
 }
 
 func appendDHCPArgs(cluster *common.Cluster, host *models.Host, inventory *models.Inventory, installerArgs []string, log logrus.FieldLogger) ([]string, error) {
-
-	if hasUserConfiguredIP(installerArgs) {
-		return installerArgs, nil
-	}
```
eifrach, line 371: why is this removed? if there is an IP we should exit

internal/host/hostcommands/install_cmd_test.go
```diff
@@ -743,13 +746,17 @@ var _ = Describe("construct host install arguments", func() {
 				{
 					"name": "eth1",
 					"ipv4_addresses":["10.56.20.80/25"]
+				},
+				{
+					"name": "eth2",
+					"ipv4_addresses":["10.56.21.80/25"]
 				}
 			]
 		}`, models.DriveTypeISCSI, models.DriveTypeSSD)
```
eifrach, line 755: you check if `installationDisk.DriveType == models.DriveTypeMultipath` in line 319
but I don't see a test for it

eifrach (changes requested):
I think there are more unit-test we can add here
not sure if you still working on it or not

+ can you validate if we have subsystem tests that may need changes/ update

internal/host/hostcommands/install_cmd.go
```diff
@@ -305,15 +308,22 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
...
-				// Currently only allowed on the OCI platform
-				installerArgs = append(installerArgs, "--append-karg", "rd.iscsi.firmware=1")
-			}
+			installationDisk = disk
+			break
+		}
+	}
+
+	if installationDisk != nil && installationDisk.DriveType == models.DriveTypeMultipath {
+		installerArgs = append(installerArgs, "--append-karg", "root=/dev/disk/by-label/dm-mpath-root", "--append-karg", "rw", "--append-karg", "rd.multipath=default")
```
linoyaslan, line 320: Can you please explain how it's possible for the installation disk to be multipath? AFAIK, multipath is not eligible as an installation disk (we have blocked it), so how could it be selected as the installation disk?

adriengentil: This is the existing code, AFAIK, multipath is blocked for iSCSI, but not fiber channel.

internal/host/hostcommands/install_cmd_test.go
adriengentil: it's existing behavior, I didn't check if there are tests for it

internal/host/hostcommands/install_cmd.go
adriengentil: it's not removed, this check has moved to L.295, to ensure we don't add any `ip=...` kargs if the user has already defined some.

internal/host/hostcommands/install_cmd.go
linoyaslan: Right. Thank you :)

adriengentil:
> can you validate if we have subsystem tests that may need changes/ update

I don't see any subsystem tests related to installation drives. We may add some later, it you think we need them.

I added more unit tests to improve the coverage of this PR.

internal/host/hostcommands/install_cmd.go
adriengentil: we will append multiple statements depending on the needs. I performed a cluster installation to check that.

PR #6603, opened 2024-07-22
MGMT-17560: Workaround missing DNS on iSCSI

adriengentil:
When iSCSI boot is enabled, DNS configuration will be missing on first
boot, even though the information was advertised by DHCP:
https://issues.redhat.com/browse/OCPBUGS-26580.

The workaround consists at re-applying the network configuration on all
the network interfaces when we detect if `/sysroot` is mounted from an
iSCSI block volume.

PRs part of this story:
- Validate iSCSI volume eligibility on hosts: https://github.com/openshift/assisted-service/pull/6434
- Report iSCSI host ip address: https://github.com/openshift/assisted-installer-agent/pull/727
- Configure DHCP on iSCSI network interface: https://github.com/openshift/assisted-service/pull/6602
- Workaround DNS issue on first boot: https://github.com/openshift/assisted-service/pull/6603
- Filter out machine networks: TODO

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -1632,6 +1632,9 @@ func (m *Manager) GenerateAdditionalManifests(ctx context.Context, cluster *comm
 		return errors.Wrap(err, "failed to add disk encryption manifest")
 	}
 
+	if err := m.manifestsGeneratorAPI.AddNicReapply(ctx, log, cluster); err != nil {
+		return errors.Wrap(err, "failed to add nic reapply manifest")
+	}
 	return nil
 }
 
--- a/internal/network/manifests_generator_test.go
+++ b/internal/network/manifests_generator_test.go
@@ -659,7 +659,7 @@ var _ = Describe("nic reaaply manifest", func() {
 
 	Context("CreateClusterManifest success", func() {
 		It("CreateClusterManifest success", func() {
-			manifestsApi.EXPECT().CreateClusterManifestInternal(gomock.Any(), gomock.Any(), false),Times(2).Return(&models.Manifest{
+			manifestsApi.EXPECT().CreateClusterManifestInternal(gomock.Any(), gomock.Any(), false).Times(2).Return(&models.Manifest{
 				FileName: "manifest.yaml",
 				Folder:   models.ManifestFolderOpenshift,
 			}, nil)
--- a/internal/network/manifests_generator_test.go
+++ b/internal/network/manifests_generator_test.go
@@ -633,6 +633,8 @@ var _ = Describe("nic reaaply manifest", func() {
 		dbName                string
 		clusterId             strfmt.UUID
 		cluster               common.Cluster
+		hostWithiSCSI         models.Host
+		hostWithSSD           models.Host
 	)
 
 	BeforeEach(func() {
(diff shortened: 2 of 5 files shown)


internal/cluster/cluster.go
```diff
@@ -1632,6 +1632,9 @@ func (m *Manager) GenerateAdditionalManifests(ctx context.Context, cluster *comm
 		return errors.Wrap(err, "failed to add disk encryption manifest")
 	}
 
+	if err := m.manifestsGeneratorAPI.AddNicReapply(ctx, log, cluster); err != nil {
+		return errors.Wrap(err, "failed to add nic reapply manifest")
+	}
```
eifrach, line 1637: This will run on all nodes  regardless if there is an ISCSI drive.
 
does it happens on the first boot? during installation ( reboot )?

adriengentil: > This will run on all nodes regardless if there is an ISCSI drive.

Yes, because we can have mixed configurations (e.g.: masters on VMs with non-iSCSI and workers on BM with iSCSI drives). That's why the script checks for an iSCSI boot volume.

> does it happens on the first boot? during installation ( reboot )?

The manifest is added to the ignition of the machine, then on reboot, the unit is create an enabled.

adriengentil: @avishayt do you think it's acceptable to drop such machine config file?

avishayt: I don't know enough to say what the effect will be

adriengentil: ack, maybe in would be better to be conservative, and just add this manifest only in one of the nodes is using iSCSI boot

adriengentil: done!

internal/network/manifests_generator_test.go
```diff
@@ -621,3 +621,117 @@ var _ = Describe("disk encryption manifest", func() {
...
+							{
+								"id": "install-id",
+								"drive_type": "%s"
+							}
+						]
+						}`, models.DriveTypeSSD),
+					InstallationDiskID: "install-id",
+				},
+			}
+			Expect(manifestsGeneratorApi.AddNicReapply(ctx, log, &cluster)).ShouldNot(HaveOccurred())
```
eifrach, line 717: can you just add a vlidatation for the error message

internal/network/manifests_generator_test.go
```diff
@@ -621,3 +621,117 @@ var _ = Describe("disk encryption manifest", func() {
...
+							{
+								"id": "install-id",
+								"drive_type": "%s"
+							}
+						]
+						}`, models.DriveTypeISCSI),
+					InstallationDiskID: "install-id",
+				},
+			}
+			Expect(manifestsGeneratorApi.AddNicReapply(ctx, log, &cluster)).Should(HaveOccurred())
```
eifrach, line 734: can you just add a vlidatation for the error message

internal/network/manifests_generator_test.go
adriengentil: this one is one an error message

internal/network/manifests_generator_test.go
```diff
@@ -621,3 +621,117 @@ var _ = Describe("disk encryption manifest", func() {
...
+		common.DeleteTestDB(db, dbName)
+	})
+
+	Context("CreateClusterManifest", func() {
+		It("added when one of the host installs on an iSCSI drive", func() {
+			manifestsApi.EXPECT().CreateClusterManifestInternal(gomock.Any(), gomock.Any(), false).Times(2).Return(&models.Manifest{
+				FileName: "manifest.yaml",
+				Folder:   models.ManifestFolderOpenshift,
+			}, nil)
+			cluster.Cluster.Hosts = []*models.Host{
```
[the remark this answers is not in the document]
adriengentil: done! Let me know!

internal/network/manifests_generator_test.go
adriengentil: done!

PR #6605, opened 2024-07-22
AGENT-930: Verify token expiration

pawanpinjarkar:
For the agent installer, parse the expiration time from the token and verify if the token is valid.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/auth/agent_local_authenticator.go
+++ b/pkg/auth/agent_local_authenticator.go
@@ -4,6 +4,7 @@ import (
 	"crypto"
 	"encoding/base64"
 	"net/http"
+	"time"
 
 	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/runtime/security"
@@ -42,7 +43,6 @@ func NewAgentLocalAuthenticator(cfg *Config, log logrus.FieldLogger) (*AgentLoca
 	if err != nil {
 		return nil, err
 	}
-
 	a := &AgentLocalAuthenticator{
 		log:       log,
 		publicKey: key,
@@ -73,19 +73,31 @@ func (a *AgentLocalAuthenticator) AuthAgentAuth(token string) (interface{}, erro
 	}
 	claims, ok := t.Claims.(jwt.MapClaims)
 	if !ok {
-		err := errors.Errorf("failed to parse JWT token claims")
+		err = errors.Errorf("failed to parse JWT token claims")
 		a.log.Error(err)
 		return nil, common.NewInfraError(http.StatusUnauthorized, err)
 	}
 
 	infraEnvID, infraEnvOk := claims[string(gencrypto.InfraEnvKey)].(string)
 	if !infraEnvOk {
-		err := errors.Errorf("claims are incorrectly formatted")
+		err = errors.Errorf("claims are incorrectly formatted")
 		a.log.Error(err)
 		return nil, common.NewInfraError(http.StatusUnauthorized, err)
 	}
 	a.log.Infof("Authenticating infraEnv %s JWT", infraEnvID)
 
+	exp, found := claims["exp"].(float64)
+	if !found {
+		// exp claim is not found in the case of install workflow
+		return ocm.AdminPayload(), nil
+	}
+	// in the case of addnodes workflow, check if the token is expired
+	expTime := time.Unix(int64(exp), 0)
+	if expTime.Before(time.Now().UTC()) {
+		err = errors.Errorf("The provided authentication token has expired. Please generate a new token and try again.")
+		a.log.Error(err)
+		return nil, common.NewInfraError(http.StatusUnauthorized, err)
+	}
 	return ocm.AdminPayload(), nil
 }
 
--- a/pkg/auth/agent_local_authenticator_test.go
+++ b/pkg/auth/agent_local_authenticator_test.go
@@ -4,8 +4,10 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"strings"
+	"time"
 
 	"github.com/go-openapi/strfmt"
+	"github.com/golang-jwt/jwt/v4"
 	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -16,14 +18,15 @@ import (
 
 var _ = Describe("AuthAgentAuth", func() {
 	var (
-		a     *AgentLocalAuthenticator
-		token string
+		a                 *AgentLocalAuthenticator
+		token, privateKey string
 	)
 
 	BeforeEach(func() {
 		infraEnvID := strfmt.UUID(uuid.New().String())
 
 		pubKey, privKey, err := gencrypto.ECDSAKeyPairPEM()
+		privateKey = privKey
 		Expect(err).ToNot(HaveOccurred())
 
 		// Encode to Base64 (Standard encoding)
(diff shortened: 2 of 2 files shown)


pkg/auth/agent_local_authenticator.go
```diff
@@ -66,6 +68,20 @@ func (a *AgentLocalAuthenticator) EnableOrgBasedFeatureGates() bool {
 }
 
 func (a *AgentLocalAuthenticator) AuthAgentAuth(token string) (interface{}, error) {
+	// Verify is token is expired or not in case of add nodes workflow
+	if a.WorkflowType == AgentWorkflowTypeAddNodes {
```
rwsu, line 72: I wonder if the check can be simplified by removing the WorkflowType check. The token contains the expiration time only in the addnodes workflow. So if the expiration time exist, then check if it has expired. If it doesn't have the expiration time, then it is the install workflow and the expiration time check is skipped.

pawanpinjarkar: `ParseExpirationFromToken` is an internal function of assisted and is used in other cases. If we always parse the token for the expiration time then in the case of install workflow, the existing check in the `ParseExpirationFromToken` will fail and error with `token missing 'exp' claim` https://github.com/openshift/assisted-service/blob/8703b59882a2f9d5730ece7c2c47cb025ab42a8f/internal/gencrypto/token.go#L107-L110 i.e. it will incorrectly expect the exp claim in the case of install workflow hence the check on workflowType

pkg/auth/agent_local_authenticator.go
```diff
@@ -66,6 +66,19 @@ func (a *AgentLocalAuthenticator) EnableOrgBasedFeatureGates() bool {
...
 func (a *AgentLocalAuthenticator) AuthAgentAuth(token string) (interface{}, error) {
+	expiryDateTime, err := gencrypto.ParseExpirationFromToken(token)
+	// it is expected that agent installer sends the correct token
+	// and will contain the exp claim when it is required.
+	if err != nil && err.Error() != gencrypto.TokenMissingExpClaim {
+		a.log.WithError(err).Error("failed to parse token expiration time")
+	}
+	expiryTime := time.Time(*expiryDateTime)
+	if expiryTime.Before(time.Now()) {
+		err := errors.Errorf("Auth token is expired. Re-run 'add-nodes' command to create new image files(ISO/PXE files)")
```
andfasano, line 77:
```suggestion
		err := errors.Errorf("The provided authentication token has expired. Please generate a new token and try again.")
```

pkg/auth/agent_local_authenticator.go
```diff
@@ -66,6 +66,20 @@ func (a *AgentLocalAuthenticator) EnableOrgBasedFeatureGates() bool {
...
 func (a *AgentLocalAuthenticator) AuthAgentAuth(token string) (interface{}, error) {
+	expiryDateTime, err := gencrypto.ParseExpirationFromToken(token)
+	// it is expected that agent installer sends the correct token
+	// and will contain the exp claim when it is required.
+	if err != nil && err.Error() != gencrypto.TokenMissingExpClaim {
+		a.log.WithError(err).Error("failed to parse token expiration time")
+	}
+	if expiryDateTime != nil {
+		expiryTime := time.Time(*expiryDateTime)
+		if expiryTime.Before(time.Now()) {
```
andfasano, line 77: IINW the [stored token](https://github.com/openshift/installer/pull/8717/files#diff-174c7f0a2bed95144baa2b38b752cb3a48aab7bc4380c16028faed7dd97924f9R91) is in UTC?

```suggestion
		if expiryTime.Before(time.Now().UTC()) {
```

pkg/auth/agent_local_authenticator.go
```diff
@@ -66,6 +66,20 @@ func (a *AgentLocalAuthenticator) EnableOrgBasedFeatureGates() bool {
 }
 
 func (a *AgentLocalAuthenticator) AuthAgentAuth(token string) (interface{}, error) {
+	expiryDateTime, err := gencrypto.ParseExpirationFromToken(token)
+	// it is expected that agent installer sends the correct token
+	// and will contain the exp claim when it is required.
+	if err != nil && err.Error() != gencrypto.TokenMissingExpClaim {
+		a.log.WithError(err).Error("failed to parse token expiration time")
+	}
+	if expiryDateTime != nil {
```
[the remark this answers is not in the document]
pawanpinjarkar: In the install workflow (also known as Day 1), the authentication token does not expire, and therefore, no expiration claim is needed. However, in the add-nodes workflow (Day 2), the token does expire. Our goal was to keep the code in the assisted process agnostic to the specific workflow type, as this is specific to the installer.

Initially, I proposed parsing the token expiration only for the add-nodes workflow, but it was suggested that we make the code more loosely coupled.

cc @andfasano  @rwsu

pkg/auth/agent_local_authenticator.go
```diff
@@ -86,6 +86,21 @@ func (a *AgentLocalAuthenticator) AuthAgentAuth(token string) (interface{}, erro
 	}
 	a.log.Infof("Authenticating infraEnv %s JWT", infraEnvID)
 
+	expiryDateTime, err := gencrypto.ParseExpirationFromToken(token)
```
carbonin, line 89: Instead of calling this method, just do what ever you need to do with the existing `claims` variable here.
That should make the logic a bit more clear and prevent you from having to match error strings.

pkg/auth/agent_local_authenticator.go
```diff
@@ -86,6 +86,21 @@ func (a *AgentLocalAuthenticator) AuthAgentAuth(token string) (interface{}, erro
...
+	// it is expected that agent installer sends the correct token
+	// and will contain the exp claim when it is required.
+	if err != nil && err.Error() != gencrypto.TokenMissingExpClaim {
```
carbonin, line 92: Additionally this error handling is flawed. If we know we don't have the exp claim, then it's safe to proceed, but if we have an error that isn't the missing claim we should error.

pkg/auth/agent_local_authenticator.go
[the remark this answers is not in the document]
andfasano: Given that we are considering the presence of the exp field optional, only a missing exp error should be ignored. Agree that the other types of error returned by `ParseExpirationFromToken` shouldn't be ignored

pkg/auth/agent_local_authenticator.go
```diff
@@ -78,6 +78,19 @@ func (a *AgentLocalAuthenticator) AuthAgentAuth(token string) (interface{}, erro
 		return nil, common.NewInfraError(http.StatusUnauthorized, err)
 	}
 
+	exp, found := claims["exp"].(float64)
+	if !found {
+		// exp claim is not found in the case of install workflow
+		return ocm.AdminPayload(), nil
```
carbonin, line 84: Even if `exp` is not found you should still test for the infraenv, right? I'd expect this checking for the expiration to happen after the infraenv check.

carbonin (changes requested):

pkg/auth/agent_local_authenticator.go
pawanpinjarkar: ah , right

PR #6606, opened 2024-07-22
MGMT-18448: Allow local cluster import to be disabled.

paul-maidment:
We need to make sure that the local cluster import may be disabled.
The import should be switched off by default and only overridden with an annotation.

This PR implements this functionality.

If the local cluster import functionality is disabled then all local cluster entities controlled
by assisted will be left intact, we will simply skip the reconcile loop.

Customers who need to enable the feature may do so by placing an annotation on the AgentServiceConfig.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/local_cluster_import_controller.go
+++ b/internal/controller/controllers/local_cluster_import_controller.go
@@ -35,6 +35,7 @@ const (
 	hubKubeConfigName                                 = "node-kubeconfigs"
 	hubPullSecretNamespace                            = "openshift-config" // #nosec G101
 	hubPullSecretName                                 = "pull-secret"      // #nosec G101
+	importLocalClusterEnabledAnnotation               = "agent-install.openshift.io/enable-local-cluster-import"
 )
 
 type LocalClusterImportReconciler struct {
@@ -123,6 +124,16 @@ func (r *LocalClusterImportReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 		// Stop reconciliation as the item is being deleted
 		r.log.Infof("Finalizer removed by local cluster import controller (local cluster CRs are cleared)")
+		return ctrl.Result{}, nil
+	}
+
+	// Enable local cluster import if annotation allows this.
+	_, importLocalClusterEnabled := instance.GetAnnotations()[importLocalClusterEnabledAnnotation]
+	if !importLocalClusterEnabled {
+		err := r.setReconciliationStatus(ctx, false, aiv1beta1.ReasonLocalClusterImportNotEnabled, "Local cluster import is not enabled")
+		if err != nil {
+			return ctrl.Result{}, errors.Wrap(err, "Unable to set reconciliation status of LocalClusterImport on AgentServiceConfig")
+		}
 		return ctrl.Result{}, nil
 	}
 
--- a/api/v1beta1/agentserviceconfig_types.go
+++ b/api/v1beta1/agentserviceconfig_types.go
@@ -156,7 +156,8 @@ const (
 	ConditionReconcileCompleted conditionsv1.ConditionType = "ReconcileCompleted"
 	// ConditionDeploymentsHealthy reports whether deployments are healthy.
 	ConditionDeploymentsHealthy conditionsv1.ConditionType = "DeploymentsHealthy"
-
+	// ReasonLocalClusterImportNotEnabled when the import of local cluster is not enabled.
+	ReasonLocalClusterImportNotEnabled string = "Local cluster import is not enabled"
 	// ReasonLocalClusterEntitiesCreated when the local cluster is managed.
 	ReasonLocalClusterManaged string = "Local cluster is managed."
 	// ReasonLocalClusterEntitiesRemoved when the local cluster is not managed.
--- a/internal/controller/controllers/local_cluster_import_controller_test.go
+++ b/internal/controller/controllers/local_cluster_import_controller_test.go
@@ -96,6 +96,9 @@ var _ = Describe("Reconcile", func() {
 		agentServiceConfig = &aiv1beta1.AgentServiceConfig{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "agent",
+				Annotations: map[string]string{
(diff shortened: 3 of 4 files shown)


internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -134,7 +141,7 @@ func (r *LocalClusterImportReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 		return ctrl.Result{}, errors.Wrap(err, "error while attempting to determine presence of managed cluster")
 	}
-	if hasManagedCluster {
+	if hasManagedCluster && importLocalClusterEnabled == "true" {
```
gamli75, line 144:
```suggestion
	if hasManagedCluster && importLocalClusterEnabled {
```

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -126,6 +127,12 @@ func (r *LocalClusterImportReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return ctrl.Result{}, nil
 	}
 
+	// Enable local cluster import if annotation allows this.
+	importLocalClusterEnabled, ok := instance.GetAnnotations()[importLocalClusterEnabledAnnotation]
```
gamli75, line 131:
```suggestion
	importLocalClusterEnabled, err := instance.GetAnnotations()[importLocalClusterEnabledAnnotation]
	if err {
```

internal/controller/controllers/local_cluster_import_controller.go
paul-maidment: This suggestion will not work, the annotation value is not a Boolean, it is a string.

internal/controller/controllers/local_cluster_import_controller.go
paul-maidment: The OK syntax is actually the correct way to check for the presence of an element in a map. It's not an error if you don't find the element. !ok just implies that the element is not present.

paul-maidment: The annotations are stored as map[string]string meaning that if the value is present, the setting will be a string value, which can be "true" or "false", this code explicitly looks for a value of "true" as an indicator the annotation is enabled.
I feel it would be incorrect to base this on the presence of the annotation (using the "OK" to assume presence of the annotation means "active") as I can forsee a scenario where a user might use it and say "false"

internal/controller/controllers/local_cluster_import_controller.go
paul-maidment: The annotations are stored as map[string]string meaning that if the value is present, the setting will be a string value, which can be "true" or "false", this code explicitly looks for a value of "true" as an indicator the annotation is enabled.
I feel it would be incorrect to base this on the presence of the annotation (using the "OK" to assume presence of the annotation means "active") as I can forsee a scenario where a user might use it and say "false"

paul-maidment: I have had a look at the way others approach annotations that are true/false as Openshift/Kubernetes does not support these natively (at least any more from what I understand) every annotation must have a string value, the value can be empty but is a string.

I am going to change this code so that the "presence test" "importLocalClusterEnabled" and it will become a boolean.

This means that the presence of the annotation with any value will be considered to be "activation" of the feature.

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -35,6 +35,7 @@ const (
 	hubKubeConfigName                                 = "node-kubeconfigs"
 	hubPullSecretNamespace                            = "openshift-config" // #nosec G101
 	hubPullSecretName                                 = "pull-secret"      // #nosec G101
+	importLocalClusterEnabledAnnotation               = "enable-local-cluster-import"
```
carbonin, line 38: All of our other [annotation names](https://github.com/openshift/assisted-service/blob/151356e84ce032e003793fa41bf563b5f33d46d5/internal/controller/controllers/agentserviceconfig_controller.go#L89-L93) are prefixed with `agent-install.openshift.io/`. We should do the same here.

carbonin (commented):
>If the local cluster import functionality is disabled then all local cluster entities controlled by assisted will be deleted.

Regarding this and the issue we're trying to avoid ...
Are we sure there is no ClusterDeployment for the `local-cluster` out of the box? It feels like the changes we're seeing in the UI mean we might be editing an existing resource rather than just creating a new one, but I'm not sure.

carbonin (approved):

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -134,7 +138,7 @@ func (r *LocalClusterImportReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 		return ctrl.Result{}, errors.Wrap(err, "error while attempting to determine presence of managed cluster")
 	}
-	if hasManagedCluster {
+	if hasManagedCluster && importLocalClusterEnabled {
```
carbonin, line 141: After some input from @berenss it seems like it's possible for some users to have created these resources (effectively importing the local cluster) before this controller existed.

Making this change would delete their resources which we should avoid.

Instead of this let's just not reconcile at all unless we're enabled. Then users can recover the situation based on their individual situation.

carbonin (changes requested):

internal/controller/controllers/local_cluster_import_controller.go
paul-maidment: Updated to force early exit of reconcile with no changes/deletions of CRs.

carbonin (approved):

carbonin:
hm, strange. Looks like we are still fast forwarding changes to 2.12.

PR #6612, opened 2024-07-23
MGMT-18231: Block iSCSI as an installation disk when its holder is multipath

linoyaslan:
Currently, customers can select an iSCSI disk as an installation disk when its holder is multipath. We have decided to block this option for now.

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [X] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/hardware/validator.go
+++ b/internal/hardware/validator.go
@@ -28,6 +28,7 @@ const (
 	tooSmallDiskTemplate       = "Disk is too small (disk only has %s, but %s are required)"
 	wrongDriveTypeTemplate     = "Drive type is %s, it must be one of %s."
 	wrongMultipathTypeTemplate = "Multipath device has path of type %s, it must be %s"
+	iSCSIWithMultipathHolder   = "iSCSI disk with a multipath holder is not eligible"
 )
 
 //go:generate mockgen -source=validator.go -package=hardware -destination=mock_validator.go
@@ -146,6 +147,25 @@ func (v *validator) DiskIsEligible(ctx context.Context, disk *models.Disk, infra
 						fmt.Sprintf(wrongMultipathTypeTemplate, inventoryDisk.DriveType, string(models.DriveTypeFC)))
 					break
 				}
+			}
+		}
+	}
+
+	// We are not allow iSCSI disk with multipath holder
+	if disk.DriveType == models.DriveTypeISCSI {
+		multipathDiskNamesMap := make(map[string]struct{})
+		for _, inventoryDisk := range allDisks {
+			if inventoryDisk.DriveType == models.DriveTypeMultipath {
+				multipathDiskNamesMap[inventoryDisk.Name] = struct{}{}
+			}
+		}
+
+		// Check if the iSCSI disk has any holders that are multipath disks
+		holders := strings.Split(disk.Holders, ",")
+		for _, holder := range holders {
+			if _, exists := multipathDiskNamesMap[holder]; exists {
+				notEligibleReasons = append(notEligibleReasons, iSCSIWithMultipathHolder)
+				break
 			}
 		}
 	}
--- a/internal/hardware/validator_test.go
+++ b/internal/hardware/validator_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"math"
 	"os"
 	"testing"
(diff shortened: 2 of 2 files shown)


internal/hardware/validator.go
```diff
@@ -150,6 +151,28 @@ func (v *validator) DiskIsEligible(ctx context.Context, disk *models.Disk, infra
...
+			if inventoryDisk.DriveType == models.DriveTypeMultipath {
+				multipathDiskName = inventoryDisk.Name
+				break // No need to continue searching once the multipath disk is found
+			}
+		}
+
+		if multipathDiskName != "" {
+			for _, inventoryDisk := range allDisks {
+				if inventoryDisk.Name == disk.Name {
+					if funk.ContainsString(strings.Split(inventoryDisk.Holders, ","), multipathDiskName) {
```
adriengentil, line 167: what contains `holders` exactly? If it's always a multipath drive, can't we just do something like:
```
	if disk.DriveType == models.DriveTypeISCSI && disk.Holders != "" {
               notEligibleReasons = append(notEligibleReasons, iSCSIWithMultipathHolder)
        }
```

internal/hardware/validator.go
```diff
@@ -150,6 +151,28 @@ func (v *validator) DiskIsEligible(ctx context.Context, disk *models.Disk, infra
 		}
 	}
 
+	// We are not allow iSCSI disk with multipath holder
+	if disk.DriveType == models.DriveTypeISCSI {
+		var multipathDiskName string
+		for _, inventoryDisk := range allDisks {
+			if inventoryDisk.DriveType == models.DriveTypeMultipath {
+				multipathDiskName = inventoryDisk.Name
+				break // No need to continue searching once the multipath disk is found
```
adriengentil, line 160: can't we have multiple multipath disks on the machine?

internal/hardware/validator.go
```diff
@@ -150,6 +151,28 @@ func (v *validator) DiskIsEligible(ctx context.Context, disk *models.Disk, infra
...
+			}
+		}
+
+		if multipathDiskName != "" {
+			for _, inventoryDisk := range allDisks {
+				if inventoryDisk.Name == disk.Name {
```
adriengentil, line 166: just use `disk`?

internal/hardware/validator.go
linoyaslan: @adriengentil The reason I didn't implement that approach is because I wasn't sure if an iSCSI disk can have holders other than multipath. If you can confirm that it cannot, I can use your suggestion, as it would simplify everything.

adriengentil: We can look into the code to understand what fills the `Holders` field

linoyaslan: @adriengentil https://github.com/openshift/assisted-installer-agent/blob/master/src/inventory/disks.go#L262-L277
It doesn't imply what drive type the holders are.

internal/hardware/validator_test.go
```diff
@@ -220,6 +220,19 @@ var _ = Describe("Disk eligibility", func() {
 		Expect(eligible).ToNot(BeEmpty())
 	})
 
+	It("Check that iSCSI is not eligible when its holder is Multipath.", func() {
```
eifrach, line 223: can you add some other cases?
- iSCSI and is eligible
- ISCSI support / not-support on different OCP versions

linoyaslan: Isn't that what you mean? - https://github.com/openshift/assisted-service/blob/master/internal/hardware/validator_test.go#L163-L186 ?

internal/hardware/validator.go
adriengentil: safer to check then :+1:

internal/hardware/validator_test.go
eifrach: yes but in multipath context 
for example:
- disk is not multipath but one of it's holders is 
- validate  that multipath not supported on all version "4.10","4.12","4.31" 
- validate the returned reason `iSCSIWithMultipathHolder`

we want to cover as much uses cases that can happen in the added if statement

eifrach (changes requested):

internal/hardware/validator_test.go
linoyaslan: > disk is not multipath but one of it's holders is

If the disk itself isn't multipath but its holder is, and we're talking about iSCSI, isn't that what I already tested?

eifrach: yes, I just gave an examples on different thing to test.

linoyaslan: > validate that multipath not supported on all version "4.10","4.12","4.31"
AFAIK, iSCSI support was introduced only after version 4.15, so the test should fail with a "wrongDriveTypeTemplate" error, similar to the "Check iSCSI on older version is not eligible" test. Do you still want me to test on these specific versions with multipath? Isn't it sufficient to test on iSCSI alone?
In addition, could you please explain why these three specific versions?

> validate the returned reason iSCSIWithMultipathHolder
Done

avishayt:
@adriengentil Please make sure this doesn't break OCI

internal/hardware/validator_test.go
eifrach: well I'm not sure about the support for it, you can test whatever version you think should be /not be supported
this was just an example

adriengentil (approved):
We are good :+1:

PR #6639, opened 2024-07-31
MGMT-18505: Fix installation from a 4.17 hub with converged flow

carbonin:
This allows ironic inspector URL to be missing in ICC config secret.

In 4.17 this is expected to be missing as the inspector service as been
removed.

If this URL is provided the agent will attempt to contact the inspector
service when it shouldn't causing the install to fail.

In earlier versions the secret will not be present so the controller
will continue to provide the inspector service URL as before.

## List all the issues related to this PR

Resolves https://issues.redhat.com/browse/MGMT-18505

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

Deployed a 4.17 hub cluster and installed a 4.17 SNO from assisted installed on it with this branch image.

- [ ] assisted-test-infra environment
- [x] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

PR #6641, opened 2024-08-01
NO-ISSUE: Use 4.17 builder image for OCP

adriengentil:
Since master tracks OCP 4.17, we should build with the corresponding
builder image.


adriengentil:
/cc @pawanpinjarkar @andfasano  @carbonin 
The upgrade of the icc dependency in [this fix](https://github.com/openshift/assisted-service/pull/6639) requires go 1.22.

PR #6646, opened 2024-08-05
[WIP] MGMT-18013: Add kube api support for adding mirror registry in agent cluster install

eliorerz:

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

/cc @eliorerz


api/v1beta1/infraenv_types.go
```diff
@@ -114,9 +115,15 @@ type InfraEnvSpec struct {
...
-	// Note: OSImageVersion can't be specified along with ClusterRef. 
+	// Note: OSImageVersion can't be specified along with ClusterRef.
 	// +optional
 	OSImageVersion string `json:"osImageVersion,omitempty"`
+
+	// MirrorRegistryRef is a reference to ClusterMirrorRegistry ConfigMap that holds the registries toml
+	// data
+	// Set per cluster mirror registry
+	// +optional
+	MirrorRegistryRef *v1beta1.MirrorRegistryConfigMapReference `json:"mirrorRegistryRef,omitempty"`
```
CrystalChun, line 126: What do you think about using the `corev1.ObjectReference`? Similar to https://github.com/openshift-assisted/cluster-api-agent/blob/9abff0c472c9156bde1cbc936fcc5bb03db7700f/bootstrap/api/v1alpha1/agentbootstrapconfig_types.go#L85

eliorerz: I though so too but after reading `corev1.ObjectReference` documentation I decided to create a new reference type. LMK if still you think we should use `corev1.ObjectReference` 

``` go
// Instead of using this type, create a locally provided and used type that is well-focused on your reference.
// For example, ServiceReferences for admission registration: https://github.com/kubernetes/api/blob/release-1.17/admissionregistration/v1/types.go#L533 .
```

pkg/mirrorregistries/cluster_image_registry.go
```diff
@@ -0,0 +1,228 @@
...
+		RegistriesConf:                  registriesConf,
+	}, nil
+}
+
+// getUserTomlConfigMapData get registries.conf and ca-bundle.crt if exist in the provided configmap inside AgentClusterInstall
+func getUserTomlConfigMapData(ctx context.Context, log logrus.FieldLogger, c client.Client, ref *hiveext.MirrorRegistryConfigMapReference) (string, string, error) {
+	userTomlConfigMap := &corev1.ConfigMap{}
+	err := c.Get(ctx, types.NamespacedName{Name: ref.Name, Namespace: ref.Namespace}, userTomlConfigMap)
+	if err != nil {
+		log.Error(err, "Failed to get ConfigMap", "ConfigMapName", ref.Name, "ConfigMapNamespace", ref.Namespace)
```
danielerez, line 147: do we have a validation for a missing CM? probably need some condition error in status for that.

internal/controller/controllers/infraenv_controller.go
```diff
@@ -499,7 +505,12 @@ func (r *InfraEnvReconciler) createInfraEnv(ctx context.Context, log logrus.Fiel
 		createParams.InfraenvCreateParams.StaticNetworkConfig = staticNetworkConfig
 	}
 
-	return r.Installer.RegisterInfraEnvInternal(ctx, key, createParams)
+	mirrorRegistryConfiguration, err := mirrorregistries.ProcessMirrorRegistryConfig(ctx, log, r.Client, infraEnv.Spec.MirrorRegistryRef)
```
danielerez, line 508: Do we need to handle somewhere a conflict between ACI and InfraEnv? I.e. do we want to settle a potential mismatch in config?

danielerez (commented):
Just a partial review.
A general suggestion - we probably need to document the new API (e.g. in /docs). Mainly to explain the support for both ACI and InfraEnv.

internal/controller/controllers/infraenv_controller.go
eliorerz: Nope, by @avishayt definition the user is able to set different mirror registry for InfraEnv and ACI

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -201,14 +201,22 @@ func (r *ClusterDeploymentsReconciler) Reconcile(origCtx context.Context, req ct
 		return ctrl.Result{Requeue: true}, err
 	}
 
+	mirrorRegistryConfiguration := &hiveext.MirrorRegistryConfiguration{}
+	if clusterInstall.Spec.MirrorRegistryRef != nil {
+		mirrorRegistryConfiguration, err = r.processMirrorRegistryConfig(ctx, log, clusterInstall)
+		if err != nil {
+			return r.updateStatus(ctx, log, clusterInstall, clusterDeployment, nil, newInputError(err.Error()))
```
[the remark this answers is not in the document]
eliorerz: The log is printed inside `processMirrorRegistryConfig`

pkg/mirrorregistries/cluster_image_registry.go
eliorerz: We update the status [here](https://github.com/openshift/assisted-service/pull/6646/files#diff-40883156aab921707c5a73dfa3b451bbd3accb4c0636d715d0a7768db8904b57R208). Is that enough?

internal/bminventory/inventory.go
```diff
@@ -144,12 +145,12 @@ type OCPClusterAPI interface {
 
 //go:generate mockgen --build_flags=--mod=mod -package bminventory -destination mock_installer_internal.go . InstallerInternals
 type InstallerInternals interface {
-	RegisterClusterInternal(ctx context.Context, kubeKey *types.NamespacedName, params installer.V2RegisterClusterParams) (*common.Cluster, error)
+	RegisterClusterInternal(ctx context.Context, kubeKey *types.NamespacedName, mirrorRegistryConfiguration *v1beta1.MirrorRegistryConfiguration, params installer.V2RegisterClusterParams) (*common.Cluster, error)
```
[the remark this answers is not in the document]
eliorerz: Yeah, One of the definitions in this task is to not affect the REST API. Updating the `ClusterCreateParams`  can only be done with adding it to the swagger definition which will affect the REST API.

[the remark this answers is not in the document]
eliorerz: I don't like that implementation also, from my perspective there are few solution for passing the `mirrorRegistryConfiguration`:
 - Change the internal API (as done here) and pass `mirrorRegistryConfiguration` if needed
 - Change `ClusterCreateParams` and include `mirrorRegistryConfiguration` which affect the REST API
 - Create a new structs that holds kubeapi parameters (e.g. `kubeKey`, `mirrorRegistryConfiguration` ), this struct will replace `kubeKey` and will prevent us from adding new parameters to the function declaration in the future. The main problem with this method is that it will require us to refactor the code and it a bit out of this PR scope

internal/controller/controllers/infraenv_controller.go
CrystalChun: I forgot to click submit on my comment yesterday 😆 but I had the same thought! 

Do we handle if user sets different mirror registry for both? Do we merge the mirror registries when they start install?

CrystalChun:
Just a head's up that Nick is putting in a change for the operator to always include the hub's OCP cluster's CA certificate bundle https://github.com/openshift/assisted-service/pull/6649

Another consideration is if the user supplies an overall Hub Cluster mirror registry CA certificate (that will now be merged with the Hub Cluster's CA bundle from the above PR). We should handle the case where they provide both a per spoke cluster mirror registry CA and a mirror registry CA in the hub cluster.

internal/controller/controllers/infraenv_controller.go
eliorerz: Basically, We don't want to check if the mirror registries are different, as long as they are exists. On failure the user will see the status update.
Now that I'm thinking of it I need to add the InfraEnv status update, it's currently updating only the ACI status

internal/bminventory/inventory.go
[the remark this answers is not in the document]
eliorerz: You basically right, but the moment we expose the mirror registry configuration on swagger we will need to support some logic on the REST side (even if this is just to raise an error when the user supplying mirror registry configuration on non kube api mode).
@avishayt WDYT?

PR #6660, opened 2024-08-07
MGMT-16212: Allow Assisted-Installer to tolerate a slightly lower memory than required

giladravid16:
We want to allow hosts to have slightly less memory than in the official documentation because some setups (like VMs) might give slightly less than the required amount. This PR changes the host validator to tolerate a slightly lower memory amount.

Apart from automatic tests this was also tested in the following way:
1. deploy a k8s cluster with assisted installer with an image for assisted-service from this branch
2.  create a vm with 16284 mb (100 mb less than 16 gb)
3. try to install an sno and add the vm as a host
4. check that the ui shows that minimum memory requirement is 16 gb even though the host is valid
5. finish the installation to make sure everything is fine
6. try the same with a vm that has 16283 mb
7. check that the host is insufficient with the error message saying it does not have the minimum memory which is 16 gb 

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -47,6 +47,11 @@ const (
 	maxHostAheadOfServiceTimeDiff                    = 1 * time.Hour
 	maxHostTimingMetrics                             = 4
 	maxPingCommandExamples                           = 4
+
+	// How much we allow the user to deviate from our official
+	// minimum. Some setups (e.g. VMs) might give just a bit less
+	// than requested, so we shouldn't be too strict about it
+	HostMemoryRequirementToleranceMiB int64 = 100
 )
 
 const FailedToFindAction = "failed to find action for step"
@@ -347,7 +352,7 @@ func (v *validator) hasMinMemory(c *validationContext) (ValidationStatus, string
 	if c.inventory == nil {
 		return status, "Missing inventory"
 	}
-	status = boolValue(c.inventory.Memory.PhysicalBytes >= conversions.MibToBytes(c.minRAMMibRequirement))
+	status = boolValue(c.inventory.Memory.PhysicalBytes >= conversions.MibToBytes(c.minRAMMibRequirement-HostMemoryRequirementToleranceMiB))
 	if status == ValidationSuccess {
 		return status, "Sufficient minimum RAM"
 	}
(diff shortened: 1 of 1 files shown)


internal/operators/mce/config.go
```diff
@@ -10,6 +10,9 @@ const (
 	// Memory value provided in GiB
 	SNOMinimumMemory int64 = 32
 	SNOMinimumCpu    int64 = 8
+
+	// Memory value provided in MiB
+	MaximumMemoryRequirementOffset int64 = 100
```
omertuc, line 15:
```suggestion
	// How much we allow the user to deviate from our official
	// minimum. Some setups (e.g. VMs) might give just a bit less
	// than requested, so we shouldn't be too strict about it
	MemoryRequirementToleranceMiB int64 = 100
```

omertuc (changes requested):

rccrdpccl:
[MGMT-16212](https://issues.redhat.com//browse/MGMT-16212) talks about generic requirements, but the change seems specific to the MCE operator. @giladravid16 could you please clarify what are the intentions of this PR by updating description and title accordingly?

giladravid16:
[MGMT-16212](https://issues.redhat.com//browse/MGMT-16212) is specifically for the assisted-installer component and the MCE operator handles the logic of validating the hosts before installation

rccrdpccl:
> [MGMT-16212](https://issues.redhat.com//browse/MGMT-16212) is specifically for the assisted-installer component and the MCE operator handles the logic of validating the hosts before installation

What do you mean by "specifically for the assisted-installer component"? Is this PR intention to be flexible with requirements under all circumstances, or only when deploying MCE?
Please do change the description to of this PR to describe what are we trying to do exactly here

The MCE operator should handle logic of additional requirements due to the MCE operator deployment. For generic/total requirements I think [this](https://github.com/openshift/assisted-service/blob/885904784c71d90ab4014c096e3348d08b174849/internal/hardware/validator.go#L297) is the place that handles them

omertuc:
I missed that the code change was in the MCE requirements, that was not the intention. The ticket wants us to address the final memory requirement, not just for the MCE operator

eliorerz:
@giladravid16 Nice work, few things:
- Please fix the commit messages.
Make sure the commit message start with `MGMT-16212: <Your commit message here>`.  Also remove the 'merge' commit, looks like it's not needed.
- Add description to this PR explaining the changes done here and the decision made (if needed)

giladravid16:
I'll write here what I understood the flow is and what should be done.

There are 4 function that are used to calculate whether the host has enough memory and they are divided into 2 groups:
1. are used to check if the host has enough memory for it's role (role memory requirement + operators' host memory requirement)
2. are used to check if the host has enough memory in general by comparing it's memory to the minimum of worker and master (or SNO if it is SNO).

But all 4 functions use the [versioned host requirements](https://github.com/openshift/assisted-service/blob/master/internal/hardware/validator.go#L511) so it is best to change what is returned from that to already take the tolerance into account. So the function that should be updated is [this.](https://github.com/openshift/assisted-service/blob/master/internal/hardware/versioned-requirements.go#L96)

Is this correct?

omertuc:
Please make sure to include your summary inside the commit message as well, so that the motivation is accessible through `git blame`

image

eliorerz (approved):
LGTM

PR #6661, opened 2024-08-07
MGMT-18514: Calculate machine networks in external platform

adriengentil:
Since external platform relies on user managed networking, we need the
feature introduced in
https://github.com/openshift/assisted-service/pull/6257 to determine the
machine networks.

This change is useful when using iSCSI boot drive on Oracle Cloud
Infrastructure, in order to select the machine network that doesn't
belong to the iSCSI subnet.

### DIFF
--- a/internal/provider/base.go
+++ b/internal/provider/base.go
@@ -6,6 +6,7 @@ import (
 	"github.com/openshift/assisted-service/internal/network"
 	"github.com/openshift/assisted-service/internal/usage"
 	"github.com/openshift/assisted-service/models"
+	"github.com/samber/lo"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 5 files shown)


adriengentil:
@ori-amizur would it make sense to call this new function `ConfigureUserManagedNetworkingInInstallConfig` from [Nutanix](https://github.com/openshift/assisted-service/blob/master/internal/provider/nutanix/installConfig.go#L63-L66) and [VSphere](https://github.com/openshift/assisted-service/blob/master/internal/provider/vsphere/installConfig.go#L77-L80) ?

ori-amizur:
> would it make sense to call this new function `ConfigureUserManagedNetworkingInInstallConfig` from [Nutanix](https://github.com/openshift/assisted-service/blob/master/internal/provider/nutanix/installConfig.go#L63-L66) and [VSphere](https://github.com/openshift/assisted-service/blob/master/internal/provider/vsphere/installConfig.go#L77-L80) ?


I am not familiar so much with Vsphere and Nutanix.  If these platforms behave the same as none-platform then yes.

PR #6665, opened 2024-08-08
MGMT-18121: Configure networking when using ISCSI over OCI

adriengentil:
A secondary network interface is required in OCI in order to install clusters on iSCSI boot volumes. Extra interfaces in OCI must be statically set, this change adds a script that configures the secondary network interface on OCI during discovery. We ensure that we execute the script only for machines in OCI with an iBFT configuration.

Since the secondary interface is the one that will be part of the machine networks, we will prevent its configuration with DHCP though the kernel arguments.

Unfortunately, we cannot propagate the network configuration set during discovery to the installed system because `--copy-network` won't work with iSCSI boot volumes: https://github.com/coreos/coreos-installer/issues/1389. As an alternative, the user will need to pass extra custom manifests to configure the network on next boot.

For the moment, we'll keep this script to configure OCI networking in assisted-service, but in the future we may want Oracle to own this script and inject it at discovery time. This will be refined in a later stage.

### DIFF
--- a/internal/host/hostcommands/install_cmd.go
+++ b/internal/host/hostcommands/install_cmd.go
@@ -294,6 +294,25 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
 
 	hasUserConfiguredIP := hasUserConfiguredIP(installerArgs)
 
+	// append kargs depending on installation drive type
+	installationDisk := hostutil.GetDiskByInstallationPath(inventory.Disks, hostutil.GetHostInstallationPath(host))
+	if installationDisk != nil {
+		installerArgs = appendMultipathArgs(installerArgs, installationDisk)
+		installerArgs, err = appendISCSIArgs(installerArgs, installationDisk, inventory, hasUserConfiguredIP)
+		if err != nil {
+			return "", err
+		}
+	} else {
+		log.Warnf("No installation disk found for host ID %s", host.ID)
+	}
+
+	if installationDisk.DriveType == models.DriveTypeISCSI && common.IsOciExternalIntegrationEnabled(cluster.Platform) {
+		// when using ISCSI along OCI, we expect the user (via a
+		// script) to configure the network statically on the nodes as
+		// DHCP is not available in this case
+		hasStaticNetwork = true
+	}
+
 	// set DHCP args only if no IP config override was specified (only for LPAR and zVM nodes on s390x)
 	if !hasStaticNetwork && !hasIPConfigOverride && !hasUserConfiguredIP {
 		// The set of ip=<nic>:dhcp kernel arguments should be added only if there is no static
@@ -306,18 +325,6 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
 		if err != nil {
 			return "", err
 		}
-	}
-
-	// append kargs depending on installation drive type
-	installationDisk := hostutil.GetDiskByInstallationPath(inventory.Disks, hostutil.GetHostInstallationPath(host))
-	if installationDisk != nil {
-		installerArgs = appendMultipathArgs(installerArgs, installationDisk)
-		installerArgs, err = appendISCSIArgs(installerArgs, installationDisk, inventory, hasUserConfiguredIP)
-		if err != nil {
-			return "", err
-		}
-	} else {
-		log.Warnf("No installation disk found for host ID %s", host.ID)
 	}
 
 	if hasStaticNetwork && !lo.Contains(installerArgs, "--copy-network") {
--- a/internal/host/hostcommands/install_cmd.go
+++ b/internal/host/hostcommands/install_cmd.go
@@ -282,6 +282,10 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
 	var err error
 	var hasStaticNetwork bool
 
+	if inventory == nil {
+		return "", fmt.Errorf("Missing inventory")
+	}
+
 	if host.InstallerArgs != "" {
 		err = json.Unmarshal([]byte(host.InstallerArgs), &installerArgs)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


internal/host/hostcommands/install_cmd.go
```diff
@@ -280,8 +280,11 @@ controls DHCP depending on the IP stack being used.
 func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inventory *models.Inventory, infraEnv *common.InfraEnv, log logrus.FieldLogger) (string, error) {
 	var installerArgs []string
 	var err error
+	var hasStaticNetwork bool
 
-	hasStaticNetwork := (infraEnv != nil && infraEnv.StaticNetworkConfig != "") || cluster.StaticNetworkConfigured
+	if inventory == nil {
```
adriengentil, line 285: added this check because we have a unit test for it: https://github.com/openshift/assisted-service/blob/master/internal/host/hostcommands/install_cmd_test.go#L1275-L1281
Before it was failing in `appendDHCPArgs` function, but if we have the inventory set to `nil` along with static  networking, we would get a panic.

internal/ignition/templates/iscsi-oci-configure-secondary-nic.sh
```diff
@@ -0,0 +1,66 @@
...
+  exit 0
+fi
+
+if [ ! -d "/sys/firmware/ibft" ]
+then
+  echo "No IBFT configuration found. Skipping."
+  exit 0
+fi
+
+MTU=9000
```
eifrach, line 19: you can't hardcoded this. 

there are network with smallar / and larger packest and on iSCSI frame you have `DF` bit set. which can cause issues.

you can take it from the Static network config

internal/ignition/templates/iscsi-oci-configure-secondary-nic.sh
```diff
@@ -0,0 +1,66 @@
...
+#     "nicIndex": 1,
+#     "privateIp": "10.0.32.210",
+#     "subnetCidrBlock": "10.0.32.0/20",
+#     "virtualRouterIp": "10.0.32.1",
+#     "vlanTag": 0,
+#     "vnicId": "ocid1.vnic.oc1.us-sanjose-1.abzwuljrsndaptsyq5mppfsoaqoun3gbvnpngcaspybo2nbpcmrozx25jenq"
+#   }
+# ]
+
+vnics=$(curl --silent -H "Authorization: Bearer Oracle" -L http://169.254.169.254/opc/v2/vnics/)
```
eifrach, line 49: there isn't error handling here - what if the API isn't avilable - the next commands should be ommited

internal/ignition/templates/iscsi-oci-configure-secondary-nic.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Configure secondary NIC on OCI when using iSCSI boot drive
+After=NetworkManager.service
+
+[Service]
```
eifrach, line 5: please add logs 
```suggestion
[Service]
StandardOutput=syslog
StandardError=syslog
```

internal/ignition/templates/iscsi-oci-configure-secondary-nic.sh
```diff
@@ -0,0 +1,66 @@
...
+secondary_if_mac_address=$(jq -r '.[1].macAddr' <<< "${vnics}")
+secondary_if_name=$(get_if_name_from_mac_address "${secondary_if_mac_address}")
+secondary_if_ip_address=$(jq -r '.[1].privateIp' <<< "${vnics}")
```
eifrach, line 52: just making sure there is not way we will get IPv6 address, right ?

internal/host/hostcommands/install_cmd.go
```diff
@@ -290,10 +293,35 @@ func constructHostInstallerArgs(cluster *common.Cluster, host *models.Host, inve
...
+		hasIPConfigOverride = hasIPConfigOverride ||
+			(installationDisk.DriveType == models.DriveTypeISCSI && common.IsOciExternalIntegrationEnabled(cluster.Platform))
+	} else {
+		log.Warnf("No installation disk found for host ID %s", host.ID)
+	}
+
+	hasStaticNetwork := (infraEnv != nil && infraEnv.StaticNetworkConfig != "") || cluster.StaticNetworkConfigured
+
 	// set DHCP args only if no IP config override was specified (only for LPAR and zVM nodes on s390x)
 	if !hasStaticNetwork && !hasIPConfigOverride && !hasUserConfiguredIP {
```
eifrach, line 326: This `hasStaticNetwork` shown twice should be moved to a common 

also line 326 look smilar maybe we can combine this to a single check

internal/ignition/templates/iscsi-oci-configure-secondary-nic.sh
adriengentil: this is the configuration for the other interface, that will be connected to the load balancer.

OCI instances use a MTU of 9000 by default:
https://docs.oracle.com/en-us/iaas/Content/Network/Troubleshoot/connectionhang.htm#Overview
```
All OCI compute instances use an MTU of 9000 by default
```

If the default value (1500 I guess), will PMTUD increase the size of the MTU if needed?

> you can take it from the Static network config

what do you mean? It's not specified in the config send by the API.

internal/ignition/templates/iscsi-oci-configure-secondary-nic.sh
adriengentil: not in the scope yet, that's why I would like Oracle to take case of this logic in the future.

internal/ignition/templates/iscsi-oci-configure-secondary-nic.sh
adriengentil: I would expect the script to fail because of `set -e`, and be restarted by systemd (`Restart=on-failure`)

internal/host/hostcommands/install_cmd.go
adriengentil: refactored, let me know!

internal/ignition/templates/iscsi-oci-configure-secondary-nic.sh
eifrach: just wanted to be more generic

PR #6683, opened 2024-08-15
MGMT-18589: Enhancement doc for backup/restore support

danielerez:
Enhancement document for improving the backup/restore support.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/backup-restore-support.md
+++ b/docs/enhancements/backup-restore-support.md
@@ -1,0 +1,154 @@
+---
+title: backup-restore-support
+authors:
+  - "@danielerez"
+creation-date: 2024-08-08
+last-updated: 2024-08-08
+---
+
+# Backup/Restore Support
+
+## Summary
+
+The Assisted Installer should support backup/restore and disaster recovery scenarios, either using OADP (OpenShift API for Data Protection) for ACM (Advanced Cluster Management), or, using Kube-API.
+I.e. the assisted-service should be resilient in such scenarios which, for this context and effort, means that restored/moved spoke clusters should keep the same state and behave the same on the new hub cluster. 
+
+## Motivation
+
+Provide resiliency in the assisted-service for safe backup/restore flows, allowing spoke clusters to be used without any restriction after DR scenarios or moving between hubs.
+
+### Goals
+
+Spoke clusters (with/without BMHs) should have a consistent state and function properly, including the ability to add/remove nodes, after these scenarios: 
+* Spoke clusters restored on a passive hub with ACM/OADP functionality.
+* Spoke clusters moved between hubs using Kube-API.
+
+### Non-Goals
+
+* Improve import day2 cluster solution by getting full host data (e.g. using an agent on existing nodes).
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As an Infrastructure Admin, I want to backup an installed spoke cluster from an active hub, including the associated BMHs (e.g. by using ACM/Kube-API/GitOps/etc).
+Then, I want to safely restore the cluster on a passive hub, including the BMHs that should keep the same state. I.e. the restored cluster should behave exactly as in the active hub with no restrictions (support day2 operations, etc).
+Note: if unbound hosts are included, they should still be automatically reprovisioned using a new discovery ISO.
+
+#### Story 2
+
+As an Infrastructure Admin, I want to backup an installed spoke cluster that doesn't have associated BMHs. Then, I want to safely restore the cluster on a passive hub.
+Note: if unbound hosts are included, they should be manually booted using a new discovery ISO. 
+
+#### Story 3
+
+As an Infrastructure Admin, I want to restore a spoke cluster in non-installed status (draft/in-progress/etc). The cluster on target hub should, ideally, keep the same status. Or, at least, fail gracefully and let the user retry the installation.
+Unbound hosts should be treated as previous stories (depends on BMHs).
+
+#### Story 4
+
+As an Infrastructure Admin, I want to safely move installed spoke clusters between hubs.
+Concerns are similar to previous stories.
+
+### Implementation Details/Notes/Constraints
+
+#### Solution for BMHs reconcile (due to a missing status)
+
+According to BMO documentation, the suggested approach for moving BareMetalHost from one cluster to another is using the `baremetalhost.metal3.io/status` [annotation](https://github.com/metal3-io/metal3-docs/blob/6a656b3eb195c1b09ba35fcad4d011c6cb02dbc2/docs/user-guide/src/bmo/status_annotation.md).
+I.e. Extract the current status of the BMH, store it in an annotation, create the BMH on the new hub:
+```
+# Save the status in json format to a file
+kubectl get bmh <name-of-bmh> -o jsonpath="{.status}" > status.json
+# Save the BMH and apply the status annotation to the saved BMH.
+kubectl -n metal3 annotate bmh <name-of-bmh> \
+  baremetalhost.metal3.io/status="$(cat status.json)" \
+  --dry-run=client -o yaml > bmh.yaml
+# Apply the saved BMH on the target hub
+kubectl --kubeconfig <target-hub-kubeconfig> apply -f bmh.yaml
+```
+This flow ensures that the BMH on the new hub has status (BMO applies the status using the annotation). Hence, as the BMH should be in state 'Provisioned', the BMH reconcile is avoided.
+I.e. prevent destruction of restored spoke clusters.
+
+##### Suggested solution for automating the flow (make it transparent to the user):
+
+In assisted-service -> `bmh_agent_controller`:
+* If a BMH is in 'Provisioned' state (and, doesn't have `status`+`paused` annotations):
+  * Extract BMH's status and add `baremetalhost.metal3.io/status` to the object metadata.
+  * Add also, add the `baremetalhost.metal3.io/paused` annotation:
+    * This is required since once the annotation is added, BMO updates the BMH's status and delete the annotation.
+    * The [paused](https://github.com/metal3-io/baremetal-operator/blob/ac05af9fb9a0b444c4441a758c731e6f518c21df/docs/api.md#pausing-reconciliation) annotation, is similar to the `detached` annotation we're already using - but, it's also pausing the reconcile loop. Thus, keeping the status annotation intact.
+* If a BMH is missing 'Status' (and, has `status`+`paused` annotation):
+  * To restore the original Status:
+    * Remove the `baremetalhost.metal3.io/paused` annotation
+      * Status is restored, but `baremetalhost.metal3.io/status` annotation is removed.
+      * The `status` and `paused` annotation will be restored (see previous bullet). 
+
+#### Solution for deleted Agent CRs
+
+1. Don't delete the Agent CR when an associated host doesn't exits on DB.
+   * Remove relevant [code](https://github.com/openshift/assisted-service/blob/5bbf696fc0cc5a220fd8a0d1e408ab0bcc889e3f/internal/controller/controllers/agent_controller.go#L143-L151) from AgentController.
+
+2. Restore the Host record in DB according to properties from the Agent CR.
+   * Seems we can infer more than the mandatory properties:
+     * host.ID <-- agent.Name
+     * host.Role <-- agent.Role
+     * host.InfraEnvID <-- agent.metadata.annotations['infraenvs.agent-install.openshift.io']
+     * host.Kind <-- 'Host' or 'AddToExistingClusterHost' if cluster is already installed
+     * host.InstallerArgs <-- agent.InstallerArgs
+     * host.RequestedHostname <-- agent.Hostname
+   * The entire inventory exists in Agent.Status. But, need to ensure it exists also post restore. If not, we can consider adding an annotation for it (i.e something as `status` in BMH).
+   * Note: some of the inventory data can also be restored from existing annotations.
+     E.g.:
+     * inventory.agent-install.openshift.io/cpu-architecture: x86_64
+     * inventory.agent-install.openshift.io/cpu-virtenabled: "true"
+     * inventory.agent-install.openshift.io/host-isvirtual: "true"
+     * inventory.agent-install.openshift.io/host-manufacturer: RedHat
+     * inventory.agent-install.openshift.io/host-productname: KVM
+     * inventory.agent-install.openshift.io/storage-hasnonrotationaldisk: "false"
+   
+### Risks and Mitigations
+
+* In general, these changes might affect flows on spoke cluster (before and after restore), need to ensure no regressions on all scenarios.
+* Adding the 'paused' annotation means a stale BMH status, need to check if it's crucial.
+* Keeping the Agent CR and restoring values into Host in DB could affect current flows (e.g. BindHost, UnbindHost, etc).
+* Restoring Hosts will result in partial records in DB. E.g. missing Inventory - could be challenging to tackle (I.e. probably need to add annotations to Agent).
+
+### Open Questions
+
+* Any specific concerns for the ConvergedFlow/Non-ConvergedFlow?
+* How seamlessly should we support restoring of 'in-progress' clusters (as the focus is on 'installed' clusters)?
+  * Need to consider conflicts when restoring from stale a backup. E.g. a backup with an unbound agent, that has already been installed since the backup was taken.
+  * We could just ensure that such clusters have some indication that an error has occurred. I.e. so the user could retry the installation.
+
+### UI Impact
+
+* No UI changes are required.
+
+* The ACM UI should display the correct status for spoke clusters, etc.
+
+### Test Plan
+
+* All existing E2E tests should pass on restored/moved spoke clusters.
+  * Specifically, should test day2 flows on clusters.
+* Should be tested both with ACM/OADP Backup/Restore feature and with the ZTP flow.
+
+## Current Alternatives
+
+* Using ACM/OADP, the current workaround for including BMHs safely:
+  * Forcing BMHs in restore by adding an annotation: `cluster.open-cluster-management.io/backup: cluster-activation`
+  * Include BMHs original status during the restore (to avoid reprovisioning of the BMHs).
+    I.e. add 'restoreStatus' in the Restore CR
+    ```
+    apiVersion: cluster.open-cluster-management.io/v1beta1
+    kind: Restore
+    ...
+    spec:
+      restoreStatus:
+        includedResources:
+        - BareMetalHost
+    ```
+
+* For moving a spoke cluster between hubs:
+  * Current workaround is to use [import an existing cluster](../hive-integration/import-installed-cluster.md) flow.


docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+  - "@danielerez"
+creation-date: 2024-08-08
+last-updated: 2024-08-08
+---
+
+# Backup/Restore Support
+
+## Summary
+
+The Assisted Installer should support backup/restore and disaster recovery scenarios, either using OADP (OpenShift API for Data Protection) for ACM (Advanced Cluster Management), or, using ZTP (Zero Touch Provisioning) flows.
```
avishayt, line 13: By "using ZTP", do you mean copying the relevant resources used by ZTP in any way?  (e.g., gitops)

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+I.e. the assisted-service should be resilient in such scenarios which, for this context and effort, means that restored/moved spoke clusters should keep the same state and behave the same on the new hub cluster. 
+
+## Motivation
+
+Provide resiliency in the assisted-service for safe backup/restore flows, allowing spoke clusters to be used without any restriction after DR scenarios or moving between hubs.
+
+### Goals
```
avishayt, line 20: I would add the ability to add/remove nodes via the new/restored hub.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+#### Story 1
+
+As an ACM/OADP user, I want to backup an installed spoke cluster from an active hub, including the associated BMHs. Then, I want to safely restore the cluster on a passive hub, including the BMHs that should keep the same state. I.e. the restored cluster should behave exactly as in the active hub with no restrictions (support day2 operations, etc).
+
+#### Story 2
+
+As an ACM/OADP user, I want to backup an installed spoke cluster from an active hub.
+Then, I want to safely restore the cluster on a passive hub, without including the BMHs, and enable a GitOps tool (e.g ArgoCD) on the passive hub. If the GitOps tool is configured to apply the BMHs CRs, it should not affect the spoke cluster.
+
+#### Stroy 3
```
avishayt, line 43:
```suggestion
#### Story 3
```

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+* Spoke clusters restored on a passive hub with ACM/OADP functionality.
+* Spoke clusters moved between hubs using ZTP flows.
+
+### Non-Goals
+
+* Improve import day2 cluster solution by getting full host data (e.g. using an agent on existing nodes).
+
+## Proposal
+
+### User Stories
```
avishayt, line 32: All stories are "as an ACM user".  Why do we care about how resources are backed up and restored?  OADP, ArgoCD, or typing each CRD by hand should be the same, right?

I think the important variations are:
1. Are they using BMHs?  This affects installed hosts getting rebooted with our ISO (bad) and uninstalled hosts getting rebooted with our ISO (good, so we can use them in the new hub).  If they aren't using BMHs, they need to manually boot uninstalled hosts if the hub changed.
2. What is the cluster state?  Installation not started, installing, or installed.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+* When the BMH is applied, the status will be missing - since reconcile is paused.
+  * To restore the original status:
+    * Remove the `baremetalhost.metal3.io/paused` annotation
+      * Status is restored, but `baremetalhost.metal3.io/status` annotation is removed.
+    * Hence, add the status and paused annotation back
+      * By detecting a BMH with `provisioned` state and a missing status annotation.  
+
+#### Solution for deleted Agent CRs
+
+1. Don't delete the Agent CR when an associated host doesn't exits on DB.
```
CrystalChun, line 87: Are we worried about a user creating an Agent CR manually? 

If we are, how do we prevent a user from adding an Agent CR that we never registered before? 


An idea I had was to label or annotate the Agent that it has been created by the service, that way we know it was an Agent that we created. This might not prevent the user from adding this themselves, but it's at least an extra layer that could prevent this scenario.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+### Risks and Mitigations
+
+* In general, these changes might affect flows on spoke cluster (before and after restore), need to ensure no regressions on all scenarios.
+* Adding the 'paused' annotation means a stale BMH status, need to check if it's crucial.
+* Keeping the Agent CR and restoring values into Host in DB could affect current flows (e.g. BindHost, UnbindHost, etc).
+* Restoring Hosts will result in partial records in DB. E.g. missing Inventory - could be challenging to tackle (I.e. probably need to add annotations to Agent).
+
+### Open Questions
+
+* Any specific concerns for the ConvergedFlow/Non-ConvergedFlow?
```
CrystalChun, line 117: If it was the converged flow, would we be concerned about how the node is registered in Ironic? Like if the state of the node in Ironic somehow wasn't preserved after restoring and we can't deprovision (or scale down) the host

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+      * Status is restored, but `baremetalhost.metal3.io/status` annotation is removed.
+    * Hence, add the status and paused annotation back
+      * By detecting a BMH with `provisioned` state and a missing status annotation.  
+
+#### Solution for deleted Agent CRs
+
+1. Don't delete the Agent CR when an associated host doesn't exits on DB.
+   * Remove relevant [code](https://github.com/openshift/assisted-service/blob/5bbf696fc0cc5a220fd8a0d1e408ab0bcc889e3f/internal/controller/controllers/agent_controller.go#L143-L151) from AgentController.
+
+2. Restore the Host record in DB according to properties from the Agent CR.
```
CrystalChun, line 90: Nick had a good idea before which was to run the assisted-installer-agent to recollect the discovery information. Do you think it's worth pursuing?

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+     * inventory.agent-install.openshift.io/host-isvirtual: "true"
+     * inventory.agent-install.openshift.io/host-manufacturer: RedHat
+     * inventory.agent-install.openshift.io/host-productname: KVM
+     * inventory.agent-install.openshift.io/storage-hasnonrotationaldisk: "false"
+   * Note: We might need to add more annotations for additional inventory data.
+   
+### Risks and Mitigations
+
+* In general, these changes might affect flows on spoke cluster (before and after restore), need to ensure no regressions on all scenarios.
+* Adding the 'paused' annotation means a stale BMH status, need to check if it's crucial.
```
CrystalChun, line 111: Also what if we want to deprovision using BMH - will this pause annotation affect that?

docs/enhancements/backup-restore-support.md
carbonin: I think OADP and the ACM DR solution generally are totally unaware of if someone is using gitops or not so we just need to make it work in both cases.

Or are you suggesting gitops is an alternative to OADP in some way?

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+### User Stories
+
+#### Story 1
+
+As an ACM/OADP user, I want to backup an installed spoke cluster from an active hub, including the associated BMHs. Then, I want to safely restore the cluster on a passive hub, including the BMHs that should keep the same state. I.e. the restored cluster should behave exactly as in the active hub with no restrictions (support day2 operations, etc).
+
+#### Story 2
+
+As an ACM/OADP user, I want to backup an installed spoke cluster from an active hub.
+Then, I want to safely restore the cluster on a passive hub, without including the BMHs, and enable a GitOps tool (e.g ArgoCD) on the passive hub. If the GitOps tool is configured to apply the BMHs CRs, it should not affect the spoke cluster.
```
carbonin, line 41: I think with or without the BMHs in the backup is an implementation detail. I don't think the use cares what we back up or not as long as everything is usable in the end.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+#### Stroy 3
+
+As a ZTP user, I want to safely move installed spoke clusters between hubs. Either manually (by applying CRs), or using a GitOps tool (e.g. ArgoCD). The spoke cluster on the target hub should behave the same and keep the status of BMHs.
```
carbonin, line 45: Manually moving CRs from one hub to another should be allowed whether someone is using ZTP or not.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+#### Story 4
+
+As an ACM/OADP or ZTP user, I want to restore/move a spoke cluster in non-installed status (draft/in-progress/etc). The cluster on target hub should, ideally, keep the same status. Or, at least, fail gracefully.
```
carbonin, line 49: Does this include unbound hosts?

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+I.e. prevent destruction of restored spoke clusters.
+
+##### Suggested solution for automating the flow (make it transparent to the user):
+
+For BMHs in the active hub: (In assisted-service -> `bmh_agent_controller`)
+* Once the BMH is in 'Provisioned' state:
+  * Extract BMH's status and add `baremetalhost.metal3.io/status` to the object metadata.
+  * Add also, add the `baremetalhost.metal3.io/paused` annotation:
+    * This is required since once the annotation is added, BMO updates the BMH's status and delete the annotation.
+    * The [paused](https://github.com/metal3-io/baremetal-operator/blob/main/docs/api.md#pausing-reconciliation) annotation, is similar to the `detached` annotation we're already using - but, it's also pausing the reconcile loop. Thus, keeping the status annotation intact.
```
carbonin, line 75: Do we need both paused and detached?

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+For BMHs in the passive hub: (In assisted-service -> `bmh_agent_controller`)
```
carbonin, line 77: I think I understand this section, but starting it with knowledge about which hub we're on is a bit confusing since a controller would not know this.

It might be better to format this as explaining the logic from the controller point of view. Something like:

If a BMH is finished installing and has status, apply the paused annotation and apply the status in the status annotation.
If a BMH is missing status and has the status and paused annotations, remove the paused annotation to allow the status to apply.

Does that cover all the logic?

docs/enhancements/backup-restore-support.md
carbonin: I'm not sure it's worth the extra effort.
If someone is already creating the CR they'll just create whatever they need to.

In any case such an agent wouldn't be able to do anything. If we try to install with it it will stay disconnected unless an actual host matching this agent is booted with a discovery ISO.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+  * To restore the original status:
+    * Remove the `baremetalhost.metal3.io/paused` annotation
+      * Status is restored, but `baremetalhost.metal3.io/status` annotation is removed.
+    * Hence, add the status and paused annotation back
+      * By detecting a BMH with `provisioned` state and a missing status annotation.  
+
+#### Solution for deleted Agent CRs
+
+1. Don't delete the Agent CR when an associated host doesn't exits on DB.
+   * Remove relevant [code](https://github.com/openshift/assisted-service/blob/5bbf696fc0cc5a220fd8a0d1e408ab0bcc889e3f/internal/controller/controllers/agent_controller.go#L143-L151) from AgentController.
```
carbonin, line 88: Are we ever relying on this code to delete agents for us in other cases?

For example if a user deletes an infraenv, does that delete the hosts internally and rely on this to remove the agent CRs or do we have finalizer logic that does this all already?

IIRC when a cluster with an infraenv bound directly to it (early binding) is removed, we should also remove the agents in that cluster.

docs/enhancements/backup-restore-support.md
carbonin: I think this was already called out as a non goal for this proposal.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+2. Restore the Host record in DB according to properties from the Agent CR.
+   * Seems we can infer more than the mandatory properties:
+     * host.ID <-- agent.Name
+     * host.Role <-- agent.Role
+     * host.InfraEnvID <-- agent.metadata.annotations['infraenvs.agent-install.openshift.io']
+     * host.Kind <-- 'Host' or 'AddToExistingClusterHost' if cluster is already installed
+     * host.InstallerArgs <-- agent.InstallerArgs
+     * host.RequestedHostname <-- agent.Hostname
+   * Some of the inventory data can be restored from annotations.
```
carbonin, line 98: Do we not have the entire inventory in agent.status? Are we expecting that to not be recreated on restore?

If not, should we consider taking BMHs example and create a similar status annotation?

docs/enhancements/backup-restore-support.md
carbonin: We already have the detached annotation which our controller removes when we need to deprovision (unbind or delete). I imagine we'll do the same for paused. Especially if it fully replaces detached.

docs/enhancements/backup-restore-support.md
carbonin: I think everything they need should be in the BMH. So as long as that is restored correctly it should be fine. :pray:

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+
+* In general, these changes might affect flows on spoke cluster (before and after restore), need to ensure no regressions on all scenarios.
+* Adding the 'paused' annotation means a stale BMH status, need to check if it's crucial.
+* Keeping the Agent CR and restoring values into Host in DB could affect current flows (e.g. BindHost, UnbindHost, etc).
+* Restoring Hosts will result in partial records in DB. E.g. missing Inventory - could be challenging to tackle (I.e. probably need to add annotations to Agent).
+
+### Open Questions
+
+* Any specific concerns for the ConvergedFlow/Non-ConvergedFlow?
+* How seamlessly should we support restoring of 'in-progress' clusters (as the focus is on 'installed' clusters)?
```
carbonin, line 118: For this I think we just need a clear answer as to what will happen along with the ability to recover.

DR is a necessarily uncontrolled scenario and some things will be in an unstable state when it happens. As long as nothing gets "stuck" on the restore side we should be fine.

Also something to think about is that the most recent backup might not reflect the current state of the world.
Imagine a scenario where the backup has an unbound agent, but that agent had actually been installed since the backup was taken. Also the opposite might be true. Obviously there's not much we can do to make this "correct" in the restore, but it's worth thinking about.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+* To reduce risks, maybe worth adding an API (annotation) for manually pausing BMHs?
```
carbonin, line 119: What does this mean? Isn't the BMH pause already an annotation?

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,151 @@
...
+    I.e. add 'restoreStatus' in the Restore CR
+    ```
+    apiVersion: cluster.open-cluster-management.io/v1beta1
+    kind: Restore
+    ...
+    spec:
+      restoreStatus:
+        includedResources:
+        - BareMetalHost
+    ```
```
carbonin, line 146: As a first pass/patch would replacing this with the BMH status and paused annotation be a net improvement even without all the rest? Seems like we could avoid this restore status stuff if we handle that internally.

carbonin (changes requested):
IMO user stories should be more like:

1. As a user of ACM I want clusters restored to a new ACM hub cluster using ACM DR (OADP) to continue to allow management capabilities, including adding and removing nodes, after the restore.
2. As a user of ACM I want to the ability to manually move an existing cluster's management from one hub to another by applying the applicable CRs while preserving the ability to add and remove nodes.
3. As a user of the infrastructure operator, no DR or hub cluster change should affect my installed clusters (restore should not reboot hosts).
4. As a user of ACM, hosts in my host inventory should automatically re-register to the new hub after a DR event (restore) if they are associated with BMHs.

ZTP, and gitops don't really matter unless we're relying on something within gitops to solve some problem.

docs/enhancements/backup-restore-support.md
CrystalChun: I hope so 😅 I just recalled times where the BMH would either be stuck "deprovisioning" or "inspecting" when ironic was reporting "clean wait" or something else

docs/enhancements/backup-restore-support.md
CrystalChun: Sweet yeah just want to make sure it's called out that's something in our scope too!

docs/enhancements/backup-restore-support.md
CrystalChun: Oh I see 🤦‍♀️ thanks!

avishayt: I would say that the information in the agent status shouldn't be relied on after the host is installed anyway.  Some things won't change, but other things might.  Better to use the OpenShift node info.

docs/enhancements/backup-restore-support.md
avishayt: We shouldn't be writing status - that should only be done by the controllers.  We can consider a status annotation, but:
- We'll have status for uninstalled hosts anyway once they boot with an ISO for the new hub
- The status for installed hosts may be stale and they should use information collected from OpenShift on the spoke

docs/enhancements/backup-restore-support.md
avishayt: We should probably ask @racedo about the requirements here.  My opinion is that we don't need to continue with the installations that are in progress.  Ideally, the cluster and agent resources should either indicate that an error occurred or allow the user to retry.  I'm guessing that if we do no work here then the agents will show up as disconnected, the cluster will show up as waiting for the agents, and installation will start once agents reboot with the new ISO.  This is also acceptable IMO.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,153 @@
...
+#### Story 2
+
+As an Infrastructure Admin, I want to backup an installed spoke cluster from an active hub,
+without including the BMHs. 
+Then, I want to safely restore the cluster on a passive hub.
+Note: since no BMHs are included, unbound hosts should be manually booted using a new discovery ISO. 
+
+#### Story 3
+
+As an Infrastructure Admin, I want to restore a spoke cluster in non-installed status (draft/in-progress/etc). The cluster on target hub should, ideally, keep the same status. Or, at least, fail gracefully and let the user retry the installation.
```
gamli75, line 49: Question: If we back up and restore a spoke cluster where the installation hasn't started yet and we don't have the full inventory, will it impact our ability to start the installation? will we get a validation failure?

avishayt: Once the user restores all resources, the agents will be in "disconnected" state.  Once the hosts boot with a new ISO, the agents will populate the inventory, and installation can proceed.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,152 @@
...
+#### Story 4
+
+As an Infrastructure Admin, I want to safely move installed spoke clusters between hubs.
+Concerns are similar to previous stories.
+
+### Implementation Details/Notes/Constraints
+
+#### Solution for BMHs reconcile (due to a missing status)
+
+According to BMO documentation, the suggested approach for moving BareMetalHost from one cluster to another is using the `baremetalhost.metal3.io/status` [annotation](https://github.com/metal3-io/metal3-docs/blob/main/docs/user-guide/src/bmo/status_annotation.md).
```
carbonin, line 59: Minor, but it would probably be good to reference a specific commit for this rather than `main` just in case the doc moves in the future.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,152 @@
...
+I.e. Extract the current status of the BMH, store it in an annotation, create the BMH on the new hub:
+```
+# Save the status in json format to a file
+kubectl get bmh <name-of-bmh> -o jsonpath="{.status}" > status.json
+# Save the BMH and apply the status annotation to the saved BMH.
+kubectl -n metal3 annotate bmh <name-of-bmh> \
+  baremetalhost.metal3.io/status="$(cat status.json)" \
+  --dry-run=client -o yaml > bmh.yaml
```
carbonin, line 67: I assume  you shouldn't include this `--dry-run` bit when doing this in a real environment, right?

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,152 @@
...
+```
+This flow ensures that the BMH on the new hub has status (BMO applies the status using the annotation). Hence, as the BMH should be in state 'Provisioned', the BMH reconcile is avoided.
+I.e. prevent destruction of restored spoke clusters.
+
+##### Suggested solution for automating the flow (make it transparent to the user):
+
+In assisted-service -> `bmh_agent_controller`:
+* If a BMH is in 'Provisioned' state (and, doesn't have `status`+`paused` annotations):
```
carbonin, line 75: In the non-converged flow the BMH moves to `Provisioned` directly after the ISO is booted IIRC

I know this is in open questions, but I figured it would be better to mention it here.

I'm not actually sure if this is a problem, but we should be sure to test both converged and non-converged if we intend this to work with both. Otherwise I think it would be okay to only support new features in converged flow, but I'm curious what @avishayt thinks about that one.

docs/enhancements/backup-restore-support.md
carbonin: Yup, that looks like it deletes the CR which would allow the agent controller to do the delete for the db record.

docs/enhancements/backup-restore-support.md
carbonin: I'm not sure what you mean by your comment @avishayt.

Are you suggesting @danielerez not use the agent status for populating the database or replying to @CrystalChun's original comment?

FWIW I think it's fine to populate the DB on the restore side using the agent status as that's what would have been in the DB on the backup side.

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,152 @@
...
+2. Restore the Host record in DB according to properties from the Agent CR.
+   * Seems we can infer more than the mandatory properties:
+     * host.ID <-- agent.Name
+     * host.Role <-- agent.Role
+     * host.InfraEnvID <-- agent.metadata.annotations['infraenvs.agent-install.openshift.io']
+     * host.Kind <-- 'Host' or 'AddToExistingClusterHost' if cluster is already installed
+     * host.InstallerArgs <-- agent.InstallerArgs
+     * host.RequestedHostname <-- agent.Hostname
+   * The entire inventory exists in Agent.Status. But, need to ensure it exists also post restore. If not, we can consider adding an annotation for it (i.e something as `status` in BMH).
+   * Note: some of the inventory data can also be restored from existing annotations.
```
carbonin, line 100: Is there any reason to use these annotations over the inventory in the status?

docs/enhancements/backup-restore-support.md
carbonin: We do have some people complaining about the BMH status not being accurate.

Do we want to make this entire procedure (pausing BMH and saving status annotation) opt-in? Maybe it would be good to put this behind a feature flag in any case to start?

docs/enhancements/backup-restore-support.md
```diff
@@ -0,0 +1,152 @@
...
+     * inventory.agent-install.openshift.io/host-manufacturer: RedHat
+     * inventory.agent-install.openshift.io/host-productname: KVM
+     * inventory.agent-install.openshift.io/storage-hasnonrotationaldisk: "false"
+   
+### Risks and Mitigations
+
+* In general, these changes might affect flows on spoke cluster (before and after restore), need to ensure no regressions on all scenarios.
+* Adding the 'paused' annotation means a stale BMH status, need to check if it's crucial.
+* Keeping the Agent CR and restoring values into Host in DB could affect current flows (e.g. BindHost, UnbindHost, etc).
+* Restoring Hosts will result in partial records in DB. E.g. missing Inventory - could be challenging to tackle (I.e. probably need to add annotations to Agent).
```
carbonin, line 114: I don't think we'd be missing inventory, but do we know of any other gaps here?

docs/enhancements/backup-restore-support.md
danielerez: God catch, will change.

docs/enhancements/backup-restore-support.md
danielerez: So according to the doc, the idea is to first just create the bmh.yaml file. Then, apply it to the target hub.
I'll add a note to clarify that.

docs/enhancements/backup-restore-support.md
danielerez: Just in case we discover there's an issue with using the Agent status.

docs/enhancements/backup-restore-support.md
danielerez: > We do have some people complaining about the BMH status not being accurate.
> 
> Do we want to make this entire procedure (pausing BMH and saving status annotation) opt-in? Maybe it would be good to put this behind a feature flag in any case to start?

Good point. As an alternative, it could be opt-out perhaps? I mean, disabling this logic when required.
I.e. Not sure how crucial it is to keep the BMH status updated once already provisioned:/

@avishayt @gamli75 @oourfali WDYT?

docs/enhancements/backup-restore-support.md
danielerez: Not yet, would need to investigate.

docs/enhancements/backup-restore-support.md
carbonin: Ah I just missed this was creating a file then applying it. Nevermind, you can ignore this.

docs/enhancements/backup-restore-support.md
carbonin: As long as we have some flag I'm fine.

I'm not as concerned with it being enabled or not.

docs/enhancements/backup-restore-support.md
danielerez: Also, shouldn't we consider a scenario in which Agents are applied without status?

carbonin: Right, I forgot the DR doesn't apply status by default.

Maybe we can check if we can enable status out of the box for agents rather than in the `Restore` CR?

carbonin:
This is looking better.

@avishayt is there anything here you can resolve or clear up?

Also how many of the open questions and remaining discussions do we want to fully settle before merging. I imagine some of these will only be resolved with some spike/dev work.

avishayt:
I agree with @carbonin .  Looks good enough to proceed.

avishayt (approved):

PR #6686, opened 2024-08-16
MGMT-18510: Add an annotation to override ironic IP family to provide in dual-stack hubs

carbonin:
Currently when a hub cluster is dual-stack the only way the preprovisioning image controller can determine which callback IP family to provide to the ironic agent is to use the cluster reference in the infraenv. If there is no cluster reference (for late binding cases) then the controller always provides the primary IP family of the hub cluster.

This prevents users from using late-binding to deploy ipv6 only hosts from an ipv4-primary dual-stack hub (or vice versa).

This commit adds an annotation on the infraenv that can be used to override the ip family used when discovering hosts with a particular discovery image. `infraenv.agent-install.openshift.io/ip-family` can be set to `v4`, `v6`, or `v4,v6` to indicate to the controller which family should be used.

In newer OCP versions the ironic agent supports multiple callback URLs in its config, but the infrastructure-operator still supports running on hubs that don't have this change so this annotation is still required for those situations.

Partially resolves https://issues.redhat.com/browse/MGMT-18510

- [ ] New Feature 
- [x] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Unit tests only - I don't have a dual-stack hub cluster to test this on

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

- [ ] This change does not require a documentation update (docstring, `docs`, README, etc) - docs added

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/preprovisioningimage_controller.go
+++ b/internal/controller/controllers/preprovisioningimage_controller.go
@@ -507,7 +507,7 @@ func (r *PreprovisioningImageReconciler) getIronicConfig(ctx context.Context, lo
 	}
 
 	if iccConfig.IronicBaseURL == "" && iccConfig.IronicInspectorBaseUrl == "" {
-		err := r.fillIronicServiceURLs(ctx, infraEnvInternal, iccConfig)
+		err := r.fillIronicServiceURLs(ctx, infraEnv, infraEnvInternal, iccConfig)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


internal/controller/controllers/preprovisioningimage_controller_test.go
```diff
@@ -233,6 +233,31 @@ var _ = Describe("PreprovisioningImage reconcile", func() {
...
 		})
+		It("Adds the ironic IPv6 address to the ignition when the infraEnv has no cluster and is annotated for IPv6", func() {
+			metav1.SetMetaDataAnnotation(&infraEnv.ObjectMeta, infraEnvIPFamilyAnnotation, ipv6Family)
+			Expect(c.Create(ctx, infraEnv)).To(BeNil())
+			backendInfraEnv.ClusterID = ""
+			mockInstallerInternal.EXPECT().GetInfraEnvByKubeKey(gomock.Any()).Return(backendInfraEnv, nil)
+			mockInstallerInternal.EXPECT().UpdateInfraEnvInternal(gomock.Any(), gomock.Any(), gomock.Any()).
+				Do(func(ctx context.Context, params installer.UpdateInfraEnvParams, internalIgnitionConfig *string) {
+					Expect(params.InfraEnvID).To(Equal(*backendInfraEnv.ID))
+					Expect(*internalIgnitionConfig).Should(ContainSubstring(url.QueryEscape(ironicServiceIPs[1])))
```
paul-maidment, line 244: I made the assumption here that `ironicServiceIPs[0]` is an IPv4 address and that `ironicServiceIPs[1]` is an IPv6 address. Though it's not directly obvious at first glance in the test.

Could we have a comment here that points out that we are using IPv6 addresses here?

carbonin: The test name is "Adds the ironic IPv6 address to the ignition when the infraEnv has no cluster and is annotated for IPv6" do you really think more commentary is required?

Also the `ironicServiceIPs` variable is just a hardcoded list [here](https://github.com/openshift/assisted-service/pull/6686/files#diff-139e1f8c8d05d26082afdbf307f7e25ad325febb81fc20e6de3136b048daa885R100)

carbonin:
@adriengentil can you take a look at this one when you get a chance?

internal/controller/controllers/preprovisioningimage_controller_test.go
paul-maidment: I saw the title of the test and I knew of the hard coded IP slice.

I thought it would be more convenient for future readers of this code to not have to look up the hard coded list.

adriengentil (commented):
Just one question, otherwise it looks good!

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -566,18 +566,23 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
...
 
-func (r *PreprovisioningImageReconciler) getIPFamilyForInfraEnv(ctx context.Context, infraEnv *common.InfraEnv) (v4 bool, v6 bool, err error) {
-	if infraEnv.ClusterID == "" {
+func (r *PreprovisioningImageReconciler) getIPFamilyForInfraEnv(ctx context.Context, infraEnv *aiv1beta1.InfraEnv, internalInfraEnv *common.InfraEnv) (v4 bool, v6 bool, err error) {
+	ipFamilyAnnotation := infraEnv.GetAnnotations()[infraEnvIPFamilyAnnotation]
+	if ipFamilyAnnotation != "" {
+		families := strings.Split(ipFamilyAnnotation, ",")
+		return funk.Contains(families, ipv4Family), funk.Contains(families, ipv6Family), nil
+	}
+	if internalInfraEnv.ClusterID == "" {
```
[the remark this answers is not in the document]
carbonin: I changed the name to be more consistent with the other functions.
Now `internalInfraEnv` is the db object and `infraEnv` is the CR.

In this case I needed both because `infraEnv` has the annotations.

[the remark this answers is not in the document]
adriengentil: oh right! I missed that `infraEnv` was referring to the DB object :man_facepalming:, it's a very good change :+1:

PR #6695, opened 2024-08-23
MGMT-18579: Inject nmpolicy captures into the provided YAML and place it with INI file under the host-specific path

linoyaslan:
The purpose of this PR is to change the current method of creating nmconnection files from the nmstate YAML provided by users. Currently, we rely on pre-generated nmconnection files and a complex script running on the node to replace temporary interface names with actual ones. These modifications aim to simplify the process by using nmpolicy and the nmstate service, offering a more straightforward approach to generating nmconnection files. For more detailed information, please refer to the documentation here - [Doc](https://docs.google.com/document/d/1JPfGhfx2pkfAn14XORIR5zpZXJ8I5VLVYdjdhouY25g/edit)

- [ ] New Feature 
- [X] Enhancement 
- [ ] Bug fix
- [X] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/bminventory/inventory.go
```diff
@@ -3700,8 +3700,27 @@ func (b *bareMetalInventory) DownloadMinimalInitrd(ctx context.Context, params i
...
+		ocpVersion, err := version.NewVersion(infraEnv.OpenshiftVersion)
+		if err != nil {
+			return common.GenerateErrorResponder(errors.Errorf("Failed to parse OCP version"))
+		}
+		minimalVersionForNmstatectl, err := version.NewVersion(constants.MinimalVersionForNmstatectl)
+		if err != nil {
+			return common.GenerateErrorResponder(errors.Errorf("Failed to parse minimal ocp version for nmstatectl"))
+		}
+		// TODO: Remove the architecture condition after fetching the nmstatectl binary from rootfs.
+		if ocpVersion.GreaterThanOrEqual(minimalVersionForNmstatectl) && infraEnv.CPUArchitecture == "x86_64" {
```
carbonin, line 3715: We have a helper for this [here](https://github.com/openshift/assisted-service/blob/17927d8dfe54d1e5c0bde382f36278b2f9f91119/internal/common/version.go#L33)

carbonin: Also we have constants for the architectures [here](https://github.com/openshift/assisted-service/blob/17927d8dfe54d1e5c0bde382f36278b2f9f91119/internal/common/common.go#L44-L52)

internal/bminventory/inventory.go
```diff
@@ -5838,8 +5857,23 @@ func (b *bareMetalInventory) V2DownloadInfraEnvFiles(ctx context.Context, params
...
+			ocpVersion, err := version.NewVersion(infraEnv.OpenshiftVersion)
+			if err != nil {
+				return common.GenerateErrorResponder(errors.Errorf("Failed to parse OCP version"))
+			}
+			minimalVersionForNmstatectl, err := version.NewVersion(constants.MinimalVersionForNmstatectl)
+			if err != nil {
+				return common.GenerateErrorResponder(errors.Errorf("Failed to parse minimal ocp version for nmstatectl"))
+			}
+			// TODO: Remove the architecture condition after fetching the nmstatectl binary from rootfs.
+			if ocpVersion.GreaterThanOrEqual(minimalVersionForNmstatectl) && infraEnv.CPUArchitecture == "x86_64" {
```
carbonin, line 5872: Since this is being used multiple times lets put it in a helper function in the `staticnetworkconfig` package.

Then we have a good home for the version constant as well.

internal/constants/files.go
```diff
@@ -19,3 +19,5 @@ const DNSWildcardFalseDomainName = "validateNoWildcardDNS"
 
 // Plain http machine config server port
 const InsecureMCSPort = 22624
+
+const MinimalVersionForNmstatectl = "4.13"
```
carbonin, line 23: As I said in the previous comment lets put this in the `staticnetworkconfig` package with a helper that uses it.

internal/constants/script-with-nmstatectl.go
```diff
@@ -0,0 +1,129 @@
...
+
+# A map of host mac addresses to interface names
+declare -A host_macs_to_hw_iface
+
+# Find destination directory based on ISO mode
+if [[ -f ${PATH_PREFIX}/etc/initrd-release ]]; then
+  ETC_NETWORK_MANAGER="${PATH_PREFIX}/etc/coreos-firstboot-network"
+else
+  ETC_NETWORK_MANAGER="${PATH_PREFIX}/etc/NetworkManager/system-connections"
+fi
```
carbonin, line 56: Some of this and some of the below script looks reused. Could we co-locate it somehow so that if/when we need to fix bugs we don't have to remember to fix them in two places?

internal/ignition/discovery.go
```diff
@@ -297,13 +298,33 @@ func (ib *ignitionBuilder) FormatDiscoveryIgnitionFile(ctx context.Context, infr
...
+		ocpVersion, err := version.NewVersion(infraEnv.OpenshiftVersion)
+		if err != nil {
+			return "", errors.Errorf("Failed to parse OCP version")
+		}
+		minimalVersionForNmstatectl, err := version.NewVersion(constants.MinimalVersionForNmstatectl)
+		if err != nil {
+			return "", errors.Errorf("Failed to parse OCP version")
+		}
+		// TODO: Remove the architecture condition after fetching the nmstatectl binary from rootfs.
+		if ocpVersion.GreaterThanOrEqual(minimalVersionForNmstatectl) && infraEnv.CPUArchitecture == "x86_64" {
```
carbonin, line 314: Same, lets get a helper function for this.

internal/ignition/templates/discovery.ign
```diff
@@ -176,6 +181,24 @@
         },
         "contents": { "source": "data:text/plain;base64,{{.PreNetworkConfigScript}}"}
     }{{end}}{{range .StaticNetworkConfig}},
+    {
+      "path": "{{.FilePath}}",
```
carbonin, line 185: Is this duplicated starting on line 201?

carbonin (changes requested):

internal/ignition/templates/discovery.ign
linoyaslan: I've separated the current flow from the new implementation, so when we're ready to fully transition to the new approach, it will be easy to remove the related parts.

internal/bminventory/inventory.go
linoyaslan: After updating with the VersionGreaterOrEqual you suggested:
`		// backward compatibility - nmstate.service has been available on RHCOS since version 4.13+, therefore, we should maintain both flows
		ok, err := common.VersionGreaterOrEqual(infraEnv.OpenshiftVersion, staticnetworkconfig.MinimalVersionForNmstatectl)
		if err != nil {
			return "", err
		}

		// TODO: Remove the architecture condition after fetching the nmstatectl binary from rootfs.
		if ok && infraEnv.CPUArchitecture == common.X86CPUArchitecture {`
		
I don't see a need for a helper function. what do you think?

pkg/staticnetworkconfig/generator.go
```diff
@@ -316,6 +317,66 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
...
+	var matched bool
+	Interfaces := config["interfaces"].([]interface{})
+
+	for _, iface := range Interfaces {
+		nic := iface.(map[interface{}]interface{})
+		interfaceName := nic["name"]
+		interfaceType := nic["type"]
+		switch interfaceType {
+		case "802-3-ethernet", "ethernet":
+			if !lo.Contains(interfaceNames, interfaceName.(string)) {
```
AlonaKaplan, line 342: Please notice https://github.com/openshift/assisted-service/pull/6715 is changing this validation.

linoyaslan:
/test edge-e2e-metal-assisted-bond
/test edge-e2e-metal-assisted-bond-4-12
//test edge-e2e-metal-assisted-bond-4-13
/test edge-e2e-metal-assisted-bond-4-14
/test edge-e2e-metal-assisted-bond-4-15
/test edge-e2e-metal-assisted-bond-4-16

/test edge-e2e-metal-assisted-static-ip-suite
/test edge-e2e-metal-assisted-static-ip-suite-4-12
/test edge-e2e-metal-assisted-static-ip-suite-4-13
/test edge-e2e-metal-assisted-static-ip-suite-4-14
/test edge-e2e-metal-assisted-static-ip-suite-4-15
/test edge-e2e-metal-assisted-static-ip-suite-4-16

pkg/staticnetworkconfig/backward_compatibility.go
```diff
@@ -0,0 +1,5 @@
+package staticnetworkconfig
+
+const MinimalVersionForNmstatectl = "4.13"
```
danielerez, line 3: As discussed offline, we probably want '4.14' as the minimal supported version. This is due to an issue in nmstatectl with using 'identifier: mac-address' - which is resolved in version [2.2.35](https://github.com/nmstate/nmstate/releases/tag/v2.2.35).
This nmstate version should be backported to z-streams of rhcos>=4.14.

subsystem/infra_env_test.go
```diff
@@ -350,9 +353,12 @@ var _ = Describe("Infra_Env", func() {
 		contents := buf.String()
 		Expect(len(contents)).ShouldNot(Equal(0))
 		Expect(contents).To(ContainSubstring("/etc/assisted/network/host0"))
-		Expect(contents).To(ContainSubstring("192.0.2.1/24"))
+		//Expect(contents).To(ContainSubstring("192.0.2.1/24"))
```
danielerez, line 356: can be removed if unneeded

subsystem/infra_env_test.go
```diff
@@ -350,9 +353,12 @@ var _ = Describe("Infra_Env", func() {
...
 		Expect(contents).To(ContainSubstring("eth0"))
-	})
+	},
+		Entry("ocp versions greater than/ equal to 4.13", staticnetworkconfig.MinimalVersionForNmstatectl),
```
danielerez, line 359: should be updated to 4.14, if we decide not to support this flow for 4.14

danielerez (commented):
Looks good, just a couple of minor comments.
Also, the PR needs a rebase (resolve conflicts).

pkg/staticnetworkconfig/generator.go
```diff
@@ -26,11 +27,27 @@ type StaticNetworkConfigData struct {
...
+	VLAN *VLAN  `yaml:"vlan,omitempty"`
+}
+
+type VLAN struct {
+	BaseIface string `yaml:"base-iface"`
+	ID        int    `yaml:"id"`
+}
+type InterfaceConfig struct {
+	Interfaces []Interface `yaml:"interfaces"`
+}
```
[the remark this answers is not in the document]
carbonin: Also should these be models and added as defined with `HostStaticNetworkConfig` rather than being parsed out here?

pkg/staticnetworkconfig/generator_test.go
```diff
@@ -61,17 +62,17 @@ var _ = Describe("generateConfiguration", func() {
...
+		escapedYamlContent, err := escapeYAMLForJSON(hostYAML)
+		Expect(err).NotTo(HaveOccurred())
 		config, err := staticNetworkGenerator.GenerateStaticNetworkConfigData(context.Background(), fmt.Sprintf(hostsYAML, escapedYamlContent))
 		fileContent := config[0].FileContents
 		Expect(err).NotTo(HaveOccurred())
 		Expect(fileContent).To(ContainSubstring("address0=192.0.2.1/24"))
 	})
 })
 
+// tests related to the flow involving nmpolicy and the nmstate service
```
carbonin, line 75: I don't think this is totally true, right? Better to just remove the comment or put the information in the describe/context

pkg/staticnetworkconfig/generator_test.go
```diff
@@ -284,9 +306,104 @@ var _ = Describe("validate mac interface mapping", func() {
...
+							MacAddress:     "f8:75:a4:a4:00:fe",
+						},
+					},
+					NetworkYaml: withPhysicalInterface,
+				},
+			}
+			err := validateStaticConfigMethod(staticNetworkConfig)
+			Expect(err).ToNot(HaveOccurred())
+		},
+			//table.Entry("ocp versions less than 4.13", staticNetworkGenerator.ValidateStaticConfigParams),
```
carbonin, line 372: Is this needed or not? If not let's get rid of the `DescribeTable` when there's only one entry.

pkg/staticnetworkconfig/generator_test.go
```diff
@@ -284,9 +306,104 @@ var _ = Describe("validate mac interface mapping", func() {
...
+							MacAddress:     "f8:75:a4:a4:00:fe",
+						},
+					},
+					NetworkYaml: withMacIdentifier,
+				},
+			}
+			err := validateStaticConfigMethod(staticNetworkConfig)
+			Expect(err).ToNot(HaveOccurred())
+		},
+			//table.Entry("ocp versions less than 4.13", staticNetworkGenerator.ValidateStaticConfigParams),
```
carbonin, line 404: Same here, if this isn't needed then there's no need for `DescribeTable`

pkg/staticnetworkconfig/generator_test.go
```diff
@@ -424,6 +544,74 @@ var _ = Describe("StaticNetworkConfig", func() {
...
+	{
+		"network_yaml": "%s"
+	}
+]`
+		hostsYAMLAndIni = `[
+	{
+		"network_yaml": "%s",
+        "mac_interface_map": %s
+	}
+]`
```
carbonin, line 560: For values that are so short I'd rather see this in one line than break up the indentation like this.

pkg/staticnetworkconfig/generator_test.go
```diff
@@ -424,6 +544,74 @@ var _ = Describe("StaticNetworkConfig", func() {
...
+        }
+      ]`
+	)
+	escapedYamlContent, err := escapeYAMLForJSON(hostYAML)
+	Expect(err).NotTo(HaveOccurred())
+
+	It("Fail with an empty host YAML", func() {
+		_, err := staticNetworkGenerator.GenerateStaticNetworkConfigDataYAML(`[
+    {
+        "network_yaml": ""
```
carbonin, line 584: Same here, a single line for this is fine

internal/ignition/templates/discovery.ign
carbonin: Okay I think the diff is just a bit strange. I guess only one of `StaticNetworkConfig` or `StaticNetworkConfigWithNmstatectl` will be defined, right?

internal/bminventory/inventory.go
```diff
@@ -3700,8 +3700,25 @@ func (b *bareMetalInventory) DownloadMinimalInitrd(ctx context.Context, params i
...
+
+		// TODO: Remove the architecture condition after fetching the nmstatectl binary from rootfs.
+		if ok && infraEnv.CPUArchitecture == common.X86CPUArchitecture {
+			netFiles, err = b.staticNetworkConfig.GenerateStaticNetworkConfigDataYAML(infraEnv.StaticNetworkConfig)
+			scriptContent = constants.PreNetworkConfigScriptWithNmstatectl
+			serviceContent = constants.MinimalISONetworkConfigServiceNmstatectl
+		} else {
+			netFiles, err = b.staticNetworkConfig.GenerateStaticNetworkConfigData(ctx, infraEnv.StaticNetworkConfig)
+			scriptContent = constants.PreNetworkConfigScript
+			serviceContent = constants.MinimalISONetworkConfigService
```
eifrach, line 3720: maybe we should have a log message here to know which config type was generated

pkg/staticnetworkconfig/backward_compatibility.go
```diff
@@ -0,0 +1,5 @@
...
+
+const MaximalVersionForNmstatectl = "4.17"
```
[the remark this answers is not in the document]
linoyaslan: It was decided to use a temporary version range from 4.13 to 4.17 until we fetch the binary from rootfs. Later, we will adjust this to include version 4.13 and above.

pkg/staticnetworkconfig/backward_compatibility.go
linoyaslan: According to Gris, version 2.2.35 will be released for RHEL 9.2 on October 1. It may take 1-2 weeks to see an OCP release that includes it. Since the mac-identifier has been broken in production for some time, we can continue working with version 4.13 until we receive the OCP versions from Gris. In addition, currently, all versions are affected by the mac-identifier issue anyway. what do you think?

internal/ignition/templates/discovery.ign
linoyaslan: correct, but since I created a new file named `pre-network-manager-config-with-nmstatectl.service`, I need distinct variables for each file to differentiate between them.

subsystem/infra_env_test.go
linoyaslan: According to Gris, version 2.2.35 will be released for RHEL 9.2 on October 1. It may take 1-2 weeks to see an OCP release that includes it. Since the mac-identifier has been broken in production for some time, we can continue working with version 4.13 until we receive the OCP versions from Gris. In addition, currently, all versions are affected by the mac-identifier issue anyway. what do you think?

pkg/staticnetworkconfig/backward_compatibility.go
danielerez: Ok, so if we don't know to which exact version the fix is going to land, then we can just change when needed.
I.e. if 4.13 won't get the fix, we can decide to not support it, or mention it in docs, etc.

subsystem/infra_env_test.go
danielerez: We can handle it later on: https://github.com/openshift/assisted-service/pull/6695/files#r1756393055

pkg/staticnetworkconfig/generator.go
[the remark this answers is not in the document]
linoyaslan: These are internal structs, so I can simply change the first character to lowercase.

internal/ignition/discovery_test.go
```diff
@@ -79,17 +79,19 @@ var _ = Describe("proxySettingsForIgnition", func() {
...
+		log                                 logrus.FieldLogger
+		builder                             IgnitionBuilder
+		mockStaticNetworkConfig             *staticnetworkconfig.MockStaticNetworkConfig
+		mockMirrorRegistriesConfigBuilder   *mirrorregistries.MockMirrorRegistriesConfigBuilder
+		infraEnvID                          strfmt.UUID
+		mockOcRelease                       *oc.MockRelease
+		mockVersionHandler                  *versions.MockHandler
+		ignitionConfig                      IgnitionConfig
+		ocpVersionInvolvingGenerateKeyfiles = "4.12"
+		ocpVersionInvolvingNmstateService   = "4.13"
```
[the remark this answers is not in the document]
linoyaslan: this version granularity shouldn’t matter for this feature

internal/isoeditor/rhcos_test.go
```diff
@@ -20,7 +20,15 @@ func TestIsoEditor(t *testing.T) {
 }
 
 var _ = Describe("RamdiskImageArchive", func() {
-	It("adds a new archive correctly", func() {
+	var (
+		scriptContentInvolvingGenerateKeyfiles  = constants.PreNetworkConfigScript
+		serviceContentInvolvingGenerateKeyfiles = constants.MinimalISONetworkConfigService
+
+		scriptContentInvolvingNmstateService  = constants.PreNetworkConfigScriptWithNmstatectl
+		serviceContentInvolvingNmstateService = constants.MinimalISONetworkConfigServiceNmstatectl
```
carbonin, line 28: Why redefine these as variables here?

pkg/staticnetworkconfig/backward_compatibility.go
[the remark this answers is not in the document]
linoyaslan: @AlonaKaplan correct me if I'm not accurate, but basically we decided on using a version range as a precaution in case we don't continue with this project in future versions. We're working in phases, and fetching the binary will be part of the next phase. For now, we're setting a temporary range, and later we can adjust it to include version 4.13 and above.

@AlonaKaplan  what do you think?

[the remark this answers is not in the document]
danielerez: As @carbonin mentioned, I don't think it worth maintaining a max version just to remove it later on. I.e. as we install the x86 version of nmstate when building the image-service, the arch check should be enough. If 4.18 is failing with the new flow for some weird reason, we should catch it early on anyway with the pre-release builds. Which sounds better than 'hiding' it with this condition.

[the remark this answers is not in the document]
AlonaKaplan: Sounds reasonable to remove the max version.

carbonin (approved):
Looks good. Unhold when you've got all the feedback from others you want.

pkg/staticnetworkconfig/generator.go
```diff
@@ -308,6 +309,67 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
...
+	Interfaces := config["interfaces"].([]interface{})
+
+	for _, iface := range Interfaces {
+		nic := iface.(map[interface{}]interface{})
+		interfaceName := nic["name"]
+		interfaceType := nic["type"]
+		switch interfaceType {
+		case "802-3-ethernet", "ethernet":
+			if !lo.Contains(interfaceNames, interfaceName.(string)) {
+				if !predicatableIfaceNamePattern.MatchString(interfaceName.(string)) {
```
zaneb, line 335: We also need to ignore the missing name when `nic["identifier"] == "mac-address"` (or perhaps `identifier, ok := nic["identifier"]; ok && identifier != "name"`)

pkg/staticnetworkconfig/generator.go
```diff
@@ -308,6 +309,67 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
...
+				} else {
+					s.log.Infof("Interface %s has no mac-interface mapping but matches a physical interface", interfaceName)
+				}
+			}
+		case "bond":
+			if val, exists := nic["link-aggregation"].(map[interface{}]interface{}); exists {
+				if ports, exists := val["port"].([]interface{}); exists {
+					for _, port := range ports {
+						if !lo.Contains(interfaceNames, port.(string)) {
+							if !predicatableIfaceNamePattern.MatchString(interfaceName.(string)) {
```
zaneb, line 346: We also need to ignore the missing name when the interface name appears as an interface in the NMState itself.

pkg/staticnetworkconfig/generator.go
```diff
@@ -308,6 +309,67 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
...
+
+	// Unmarshal the JSON string into the config struct
+	err := yaml.Unmarshal([]byte(networksYaml), &config)
+	if err != nil {
+		s.log.WithError(err).Errorf("Error unmarshalling JSON")
+		return err
+	}
+
+	// See 'man systemd.net-naming-scheme' for interface naming protocol
+	predicatableIfaceNamePattern := regexp.MustCompile("^en[PsvxXbucaipod]")
```
zaneb, line 325: If we really must have two separate functions doing the validation, at least make this global so it can't get out of sync.

internal/bminventory/inventory.go
```diff
@@ -4774,9 +4774,22 @@ func (b *bareMetalInventory) validateInfraEnvCreateParams(ctx context.Context, p
...
 	if params.InfraenvCreateParams.StaticNetworkConfig != nil {
-		if err = b.staticNetworkConfig.ValidateStaticConfigParams(params.InfraenvCreateParams.StaticNetworkConfig); err != nil {
+		// backward compatibility - nmstate.service has been available on RHCOS since version 4.14+, therefore, we should maintain both flows
+		var ok bool
+		ok, err = staticnetworkconfig.NMStatectlServiceSupported(params.InfraenvCreateParams.OpenshiftVersion, common.X86CPUArchitecture)
+		if err != nil {
 			return err
 		}
+
+		if ok {
```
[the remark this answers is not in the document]
linoyaslan: I agree. I'll keep the validation directly on the YAML since we don't want our code handling both YAML and nmconnection formats. The whole point of using nmstate is that we don't need to worry about its internal translation. Right now, we're using the generator for validation, but we're aware that the nmstate team is working on Go structures and adding validation, so that aspect will change in a later phase.

pkg/staticnetworkconfig/generator.go
linoyaslan: As I agree with you, so I'll stick with the validation on YAML, so no need to make it global

pkg/staticnetworkconfig/generator.go
AlonaKaplan: I believe that the following won't work when using nmstate.service.

```
 ---
 interfaces:
 - name: eth0
   type: ethernet
   state: up
   identifier: mac-address
   mac-address: 00:23:45:67:89:1a
 - name: eth1
   type: ethernet
   state: up
   identifier: mac-address
   mac-address: 00:23:45:67:89:1b
 - name: bond0
   type: bond
   state: up
   link-aggregation:
     mode: balance-rr
     port:
       - eth0
       - eth1
 ```

If it is indeed the case, the bond validation should remain as is.

Comment (linoyaslan) [reply pkg/staticnetworkconfig/generator.go:317]: @AlonaKaplan  As discussed offline, the above example indeed doesn't work with nmstate, but it also fails with a VerificationError when the ports are physical interface names. I've reached out to Gris to check if this is a known issue with nmstate.

Comment (zaneb) [reply pkg/staticnetworkconfig/generator.go:317]: Gris said [here](https://github.com/nmstate/nmstate/issues/2380#issuecomment-2316805780) that it would work for `nmstatectl apply`. Is there something different about `nmstatectl service` that would make it not work there? It seems to me like an issue with nmstate if this doesn't work. Forbidding this config would certainly be a regression for ABI.

Comment (AlonaKaplan) [reply pkg/staticnetworkconfig/generator.go:317]: @cathay4t would the above config work using nmstate apply?

Comment (linoyaslan) [reply pkg/staticnetworkconfig/generator.go]: @zaneb I’ve added a temporary condition to prevent regressions for ABI while blocking non-ABI users from using the mac-identifier - https://github.com/openshift/assisted-service/pull/6695/commits/f426f0976d656a7dde08f1693b8cc528e95b7e52#diff-8a1899b4c0132785a200ded4829d37bdd9c27e5f182f9edd91879a228649a45cR327-R329

File: pkg/staticnetworkconfig/generator.go [hunk 25]
```diff
@@ -166,49 +276,80 @@ func (s *StaticNetworkConfigGenerator) formatNMConnection(nmConnection string) (
...
+		if !exists {
+			return errors.Errorf("interface name not found in networks configuration")
 		}
-		interfaceType, err := connectionSection.GetKey("type")
-		if err != nil {
-			return errors.Wrapf(err, "failed to get interface type for interface %s", interfaceName.String())
+
+		// TODO: This is a temporary validation. It should be removed once the mac-identifier bug in nmstate is resolved.
+		identifier, exists := nic["identifier"]
+		if installerInvoker != "agent-installer" && versions.GreaterThanOrEqualTo(ocpVersion, MinimalVersionForNmstatectl) && exists && identifier == "mac-address" {
```
danielerez, line 305: can we use NMStatectlServiceSupported helper os something similar here?

danielerez, line 305: maybe worth simplifying it a bit just for readability. something like:
```
identifier, exists := nic["identifier"]
isMacAddressIdentifier := exists && identifier == "mac-address"
if isMacAddressIdentifier && NMStatectlServiceSupported... 
```

pkg/staticnetworkconfig/generator.go
```diff
@@ -275,49 +275,97 @@ func (s *StaticNetworkConfigGenerator) formatNMConnection(nmConnection string) (
...
+
+		identifier, exists := nic["identifier"]
+		isMacAddressIdentifier := exists && identifier == "mac-address"
+		isVersionOK, err := NMStatectlServiceSupported(ocpVersion, arch)
 		if err != nil {
-			return errors.Wrapf(err, "failed to get interface type for interface %s", interfaceName.String())
+			return err
+		}
+
+		// TODO: This is a temporary validation. It should be removed once the mac-identifier bug in nmstate is resolved.
```
zaneb, line 322: I suggest linking to an actual bug report, otherwise nobody in the future is ever going to know if it is resolved or not.

zaneb:
As far as I can see this should be OK for ABI now.

linoyaslan:
/retitle MGMT-18579: Inject nmpolicy captures into the provided YAML and place it with INI file under the host-specific path

pkg/staticnetworkconfig/generator.go
cathay4t: YAML at https://github.com/openshift/assisted-service/pull/6695#discussion_r1759656407 will not works as expected, the bond still refer the bond port by name eth1 and eth2.

You may use `interface.controller` property for now:
```yml
---
interfaces:
  - name: bond0
    type: bond
    link-aggregation:
      mode: balance-rr
  - name: bond0-port1
    type: ethernet
    state: up
    identifier: mac-address
    mac-address: 00:23:45:67:89:1a
    controller: bond0
```

Another approach would be wait https://github.com/nmstate/nmstate/pull/2710

AlonaKaplan: @zaneb as Gris approved, the [yaml](https://github.com/openshift/assisted-service/pull/6695#discussion_r1759656407) is not supported by nmstate and not the current/planned API to refer bond ports with mac-identifier.
I understand it works with nmstate generator, but it is kind of a hack.
I think both ABI and AI should be aligned with the official nmstate API and not support this format.

PR #6696, opened 2024-08-25
MGMT-18659: Local cluster name should be picked up from ACM ManagedCluster label

paul-maidment:
When performing a local cluster import - the local cluster name is hardwired to "local-cluster", this is not flexible enough and needs to be changed.
We should be picking up the name from the ManagedCluster that is labelled as "local-cluster" - this should then be used wherever we use the present local-cluster name.
An additional check is performed to ensure that the clusterID referenced in the labels of ManagedCluster matches the clusterID in the labels found in the
clusterVersion of the hub on which the LocalClusterImport is running.

This should be sufficient at this stage as we do not need to handle scenarios such as a "rename", mainly because ACM will be enforcing some rules

From ACM-DDR-022:

```
To ensure backward compatibility, we will not directly allow renaming the local cluster from MCE or ACM installer at the first stage. The user can choose a customized local cluster name with the following step:
Disable local cluster management in MCE or ACM when install
Manually create a ManagedCluster resource with the label “local-cluster=true” on the hub
```

I have also added some entity cleanup that was missed in previous versions, mainly just making sure that unused secrets are deleted.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment

1. Tested and verified to be correctly working in a dev-scripts environment with a custom operator bundle.
2. AgentServiceConfig was created in a way that renders LocalClusterInstall active
3. ManagedCluster was given the clusterID of the hub
4. Expectation was that local managed cluster would have CR's created for it in the namespace specified by the name of ManagedCluster. This expectation was met.

- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/local_cluster_import_controller.go
+++ b/internal/controller/controllers/local_cluster_import_controller.go
@@ -36,6 +36,8 @@ const (
 	hubPullSecretNamespace                            = "openshift-config" // #nosec G101
 	hubPullSecretName                                 = "pull-secret"      // #nosec G101
 	importLocalClusterEnabledAnnotation               = "agent-install.openshift.io/enable-local-cluster-import"
+	localClusterLabel                                 = "local-cluster"
+	clusterIDLabel                                    = "clusterID"
 )
 
 type LocalClusterImportReconciler struct {
@@ -45,10 +47,9 @@ type LocalClusterImportReconciler struct {
 	agentServiceConfigName string
 }
 
-func NewLocalClusterImportReconciler(client client.Client, localClusterName string, agentServiceConfigName string, log *logrus.Logger) *LocalClusterImportReconciler {
+func NewLocalClusterImportReconciler(client client.Client, agentServiceConfigName string, log *logrus.Logger) *LocalClusterImportReconciler {
 	return &LocalClusterImportReconciler{
 		client:                 client,
-		localClusterName:       localClusterName,
 		log:                    log,
 		agentServiceConfigName: agentServiceConfigName,
 	}
@@ -137,16 +138,29 @@ func (r *LocalClusterImportReconciler) Reconcile(ctx context.Context, req ctrl.R
 		return ctrl.Result{}, nil
 	}
 
-	hasManagedCluster, err := r.hasLocalManagedCluster(ctx)
+	clusterVersion := &configv1.ClusterVersion{}
+	namespacedName := types.NamespacedName{
+		Name: "version",
+	}
+	err := r.client.Get(ctx, namespacedName, clusterVersion)
+	if err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "unable to fetch local cluster version")
+	}
+
+	err = r.getLocalManagedClusterName(ctx, clusterVersion)
 	if err != nil {
 		err = r.setReconciliationStatus(ctx, false, aiv1beta1.ReasonUnableToDetermineLocalClusterManagedStatus, err.Error())
 		if err != nil {
 			return ctrl.Result{}, errors.Wrap(err, "Unable to set reconciliation status of LocalClusterImport on AgentServiceConfig")
 		}
-		return ctrl.Result{}, errors.Wrap(err, "error while attempting to determine presence of managed cluster")
+		return ctrl.Result{}, errors.Wrap(err, "error while attempting to determine local cluster name")
 	}
-	if hasManagedCluster {
-		if err = r.importLocalCluster(ctx, instance); err != nil {
+	hasLocalManagedCluster, err := r.hasLocalManagedCluster(ctx)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if hasLocalManagedCluster {
+		if err = r.importLocalCluster(ctx, instance, clusterVersion); err != nil {
 			return ctrl.Result{}, errors.Wrap(err, "failed to create managed cluster CRs")
 		}
 		err = r.setReconciliationStatus(ctx, true, aiv1beta1.ReasonLocalClusterManaged, "")
@@ -169,8 +183,9 @@ func (r *LocalClusterImportReconciler) Reconcile(ctx context.Context, req ctrl.R
 	return ctrl.Result{}, nil
 }
 
-func (r *LocalClusterImportReconciler) checkManagedClusterName(obj metav1.Object) bool {
-	return obj.GetName() == r.localClusterName
+func checkIsLocalManagedCluster(obj metav1.Object) bool {
+	value, ok := obj.GetLabels()[localClusterLabel]
+	return strings.ToLower(value) == "true" && ok
 }
 
 func checkSecretName(obj metav1.Object) bool {
@@ -180,10 +195,10 @@ func checkSecretName(obj metav1.Object) bool {
 // SetupWithManager sets up the controller with the Manager.
 func (r *LocalClusterImportReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	managedClusterPredicates := builder.WithPredicates(predicate.Funcs{
-		CreateFunc:  func(e event.CreateEvent) bool { return r.checkManagedClusterName(e.Object) },
-		UpdateFunc:  func(e event.UpdateEvent) bool { return r.checkManagedClusterName(e.ObjectNew) },
-		DeleteFunc:  func(e event.DeleteEvent) bool { return r.checkManagedClusterName(e.Object) },
-		GenericFunc: func(e event.GenericEvent) bool { return r.checkManagedClusterName(e.Object) },
+		CreateFunc:  func(e event.CreateEvent) bool { return checkIsLocalManagedCluster(e.Object) },
+		UpdateFunc:  func(e event.UpdateEvent) bool { return checkIsLocalManagedCluster(e.ObjectNew) },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return checkIsLocalManagedCluster(e.Object) },
+		GenericFunc: func(e event.GenericEvent) bool { return checkIsLocalManagedCluster(e.Object) },
 	})
 	secretPredicates := builder.WithPredicates(predicate.Funcs{
 		CreateFunc:  func(e event.CreateEvent) bool { return checkSecretName(e.Object) },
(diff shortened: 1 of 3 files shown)


internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -169,8 +170,9 @@ func (r *LocalClusterImportReconciler) Reconcile(ctx context.Context, req ctrl.R
 	return ctrl.Result{}, nil
 }
 
-func (r *LocalClusterImportReconciler) checkManagedClusterName(obj metav1.Object) bool {
-	return obj.GetName() == r.localClusterName
+func (r *LocalClusterImportReconciler) checkIsLocalManagedCluster(obj metav1.Object) bool {
+	_, isLocalManagedCluster := obj.GetLabels()[localClusterLabel]
```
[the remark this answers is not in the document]
paul-maidment: Good question, we can verify this with the ACM team. 
The pattern (within our project) for Annotations appears to be based on the simple presence of an annotation, I always thought this was a bit dodgy, but everyone (yourself included) uses this approach.

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -263,19 +265,31 @@ func (r *LocalClusterImportReconciler) deleteInfraEnv(ctx context.Context, names
...
-			return false, nil
+	if err := r.client.Delete(ctx, secret); err != nil && !k8serrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete Secret %s in namespace %s", name, namespace)
+	}
+	return nil
+}
+
+func (r *LocalClusterImportReconciler) getLocalManagedClusterName(ctx context.Context) (*string, error) {
+	managedClusters := &clusterv1.ManagedClusterList{}
+	err := r.client.List(ctx, managedClusters)
```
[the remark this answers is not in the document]
paul-maidment: Actually, let's go further than that...

Let's use the label selector to also match on clusterID from the clusterVersion

We just need to move the retrieval of that a bit, then we can take advantage of the fact that the ManagedCluster contains a clusterID that can be verified against the running cluster.

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -284,17 +298,28 @@ func (r *LocalClusterImportReconciler) hasLocalManagedCluster(ctx context.Contex
...
 		return err
 	}
 	err = r.deleteInfraEnv(ctx, r.localClusterName, r.localClusterName)
 	if err != nil && !k8serrors.IsNotFound(err) {
-		r.log.Errorf("could not delete local InfraEnv due to error %s", err.Error())
+		r.log.Errorf("could not delete local cluster InfraEnv due to error %s", err.Error())
+		return err
+	}
+	adminKubeConfigSecretName := fmt.Sprintf(adminKubeConfigStringTemplate, r.localClusterName)
+	err = r.deleteSecret(ctx, r.localClusterName, adminKubeConfigSecretName)
```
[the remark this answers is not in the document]
paul-maidment: I could see the value if we considered the previous releases "stable" enough for the customer to use, but given that we have deactivated the feature pending further refinements, I don't think backports will be necessary here.

The aim here was to fill a gap that was previously missed and to use this as an opportunity to fill that gap without too much bureaucracy.

So I will opt to properly comment the PR instead.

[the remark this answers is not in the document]
paul-maidment: Note: I did actually call this out in a one line comment at the bottom of the commit message, but it can easily be missed.

paul-maidment:
/hold Even if the review is approved, I would like to hold in order to account for PX requirements etc.

Please review as normal

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -137,16 +138,26 @@ func (r *LocalClusterImportReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 		err = r.setReconciliationStatus(ctx, false, aiv1beta1.ReasonUnableToDetermineLocalClusterManagedStatus, err.Error())
 		if err != nil {
 			return ctrl.Result{}, errors.Wrap(err, "Unable to set reconciliation status of LocalClusterImport on AgentServiceConfig")
 		}
-		return ctrl.Result{}, errors.Wrap(err, "error while attempting to determine presence of managed cluster")
+		return ctrl.Result{}, errors.Wrap(err, "error while attempting to determine local cluster name")
 	}
-	if hasManagedCluster {
-		if err = r.importLocalCluster(ctx, instance); err != nil {
+	if len(localClusterName) > 0 {
```
carbonin, line 158: I suppose this works, but I think the convention is to compare directly to `""` rather than check the length of the string.

My immediate reaction to seeing this was to check why you were returning a slice type from `getLocalManagedClusterName`

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -137,16 +138,26 @@ func (r *LocalClusterImportReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		r.localClusterName = localClusterName
```
carbonin, line 159: What is the value in setting this?

I think we either need to lean into this being a cache and only query for the local cluster name when it's not already set or query it whenever we need it.

Without too much additional thought I think I'd prefer the former. Make `getLocalManagedClusterName` set the field if it's not already set rather than returning it then have everywhere else use the reconciler struct field.

This way we don't get inconsistent behavior if the actual managed cluster does change names.

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -263,19 +275,31 @@ func (r *LocalClusterImportReconciler) deleteInfraEnv(ctx context.Context, names
...
+	}
+	return nil
+}
+
+func (r *LocalClusterImportReconciler) getLocalManagedClusterName(ctx context.Context, clusterVersion *configv1.ClusterVersion) (string, error) {
+	managedClusters := &clusterv1.ManagedClusterList{}
+	// This should result in a single ManagedCluster that matches our clusterID
+	err := r.client.List(ctx, managedClusters, client.MatchingLabels{localClusterLabel: "true", clusterIDLabel: string(clusterVersion.Spec.ClusterID)})
+	count := len(managedClusters.Items)
+	if err != nil || count == 0 {
```
carbonin, line 296: I think this should handle only the case where we have an error.

The `count == 0` case should create a new error since `err` will be nil.

internal/controller/controllers/local_cluster_import_controller_test.go
```diff
@@ -29,7 +30,9 @@ import (
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
-const localClusterName = "local-cluster"
+const localClusterName = "arbitrary-cluster-name"
+
+var localClusterID = uuid.NewString()
```
carbonin, line 35: Any reason we can't put this in the per-test-lifecycle? (i.e. generate a new one in `BeforeEach`)?

internal/controller/controllers/local_cluster_import_controller.go
paul-maidment: I don't follow. 

If we make getLocalManagedClusterName(....) act as a cache then we have to consider scenarios such as invalidation.

This approach is simpler, set once per reconcile, not used before the point at which we set it.

I think it's fine the way it is.

paul-maidment: Cache suggestion taken and applied.

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -263,16 +278,49 @@ func (r *LocalClusterImportReconciler) deleteInfraEnv(ctx context.Context, names
...
+	if r.localClusterName != "" {
+		return nil
+	}
+	managedClusters := &clusterv1.ManagedClusterList{}
+	// This should result in a single ManagedCluster that matches our clusterID
+	err := r.client.List(ctx, managedClusters, client.MatchingLabels{localClusterLabel: "true", clusterIDLabel: string(clusterVersion.Spec.ClusterID)})
+	count := len(managedClusters.Items)
+	if err != nil {
+		return err
+	}
```
carbonin, line 304: Not sure it matters a ton, but it's strange to see something between when we set `err` and check it.

Probably better to check `err` before using `managedCluster.Items`

internal/controller/controllers/local_cluster_import_controller.go
```diff
@@ -263,16 +278,49 @@ func (r *LocalClusterImportReconciler) deleteInfraEnv(ctx context.Context, names
...
-		Name: r.localClusterName,
+	err := r.client.Get(ctx, types.NamespacedName{Name: r.localClusterName}, managedCluster)
+	if k8serrors.IsNotFound(err) {
+		return false, nil
 	}
-	err := r.client.Get(ctx, namespacedName, managedCluster)
 	if err != nil {
-		if k8serrors.IsNotFound(err) {
-			return false, nil
-		}
```
carbonin, line 275: None of this behavior has actually changed, right? Maybe just leave it like it was?

internal/controller/controllers/local_cluster_import_controller.go
paul-maidment: Yes, seems I made a typo here, meant to do the check immediately.

I'll change this.

internal/controller/controllers/local_cluster_import_controller.go
paul-maidment: Yes fair enough. Think I was just trying to save some lines of code by inlining the namespace. I can change it back.

paul-maidment: OK either way I suppose but I'd argue that the code is a bit cleaner and neater this way. 

As you say, not really a behavioural change and nothing risky happening here?

carbonin: Fair enough. I'm fine with this

carbonin (approved):

PR #6715, opened 2024-09-02
MGMT-18694, MGMT-18575, OCPBUGS-34849: Don't require mapping for names matching physical interfaces

bfournie:
For static network interface names that match physical interfaces don't require mac-address mapping. This allows the names to be in the networkConfig section only to handle, for example, disabled interfaces.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/staticnetworkconfig/generator.go
+++ b/pkg/staticnetworkconfig/generator.go
@@ -170,7 +170,6 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 	interfaceNames := lo.Map(macInterfaceMap, func(m *models.MacInterfaceMapItems0, _ int) string { return m.LogicalNicName })
 	// See 'man systemd.net-naming-scheme' for interface naming protocol
 	predicatableIfaceNamePattern := regexp.MustCompile("^en[PsvxXbucaipod]")
-	var matched bool
 	for _, nc := range networksConfig {
 		cfg, err := ini.LoadSources(ini.LoadOptions{IgnoreInlineComment: true}, []byte(nc.FileContents))
 		if err != nil {
@@ -196,20 +195,7 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 					s.log.Infof("Interface %s has no mac-interface mapping but matches a physical interface", interfaceName.String())
 				}
 			}
-			matched = true
-		case "vlan":
-			vlanSection := cfg.Section("vlan")
-			parent, err := vlanSection.GetKey("parent")
-			if err != nil {
-				return errors.Wrapf(err, "failed to get parent of vlan %s", interfaceName.Name())
-			}
-			if lo.Contains(interfaceNames, parent.String()) {
-				matched = true
-			}
 		}
-	}
-	if !matched {
-		return errors.Errorf("no mac address mapping can be associated with the available network interfaces %v", interfaceNames)
 	}
 	return nil
 }
@@ -228,6 +214,9 @@ func (s *StaticNetworkConfigGenerator) ValidateStaticConfigParams(staticNetworkC
 }
 
 func (s *StaticNetworkConfigGenerator) validateMacInterfaceName(hostIdx int, macInterfaceMap models.MacInterfaceMap) error {
+	if len(macInterfaceMap) == 0 {
+		return fmt.Errorf("at least one interface for host %d must be provided", hostIdx)
+	}
 	interfaceCheck := make(map[string]struct{}, len(macInterfaceMap))
 	macCheck := make(map[string]struct{}, len(macInterfaceMap))
 	for _, macInterface := range macInterfaceMap {
--- a/pkg/staticnetworkconfig/generator.go
+++ b/pkg/staticnetworkconfig/generator.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -167,6 +168,8 @@ func (s *StaticNetworkConfigGenerator) formatNMConnection(nmConnection string) (
 
 func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterfaceMap models.MacInterfaceMap, networksConfig []StaticNetworkConfigData) error {
 	interfaceNames := lo.Map(macInterfaceMap, func(m *models.MacInterfaceMapItems0, _ int) string { return m.LogicalNicName })
+	// See 'man systemd.net-naming-scheme' for interface naming protocol
+	predicatableIfaceNamePattern := regexp.MustCompile("^en[PsvxXbucaipod]")
 	var matched bool
 	for _, nc := range networksConfig {
 		cfg, err := ini.LoadSources(ini.LoadOptions{IgnoreInlineComment: true}, []byte(nc.FileContents))
@@ -186,7 +189,11 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 		switch interfaceType.String() {
 		case "802-3-ethernet", "ethernet":
 			if !lo.Contains(interfaceNames, interfaceName.String()) {
-				return errors.Errorf("mac-interface mapping for interface %s is missing", interfaceName.String())
+				if !predicatableIfaceNamePattern.MatchString(interfaceName.String()) {
+					return errors.Errorf("mac-interface mapping for interface %s is missing and not a physical interface", interfaceName.String())
+				} else {
+					s.log.Infof("Interface %s has no mac-interface mapping but matches a physical interface", interfaceName.String())
+				}
 			}
 			matched = true
 		case "vlan":
--- a/pkg/staticnetworkconfig/generator.go
+++ b/pkg/staticnetworkconfig/generator.go
@@ -180,7 +180,8 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 		connectionSection := cfg.Section("connection")
 		interfaceName, err := connectionSection.GetKey("interface-name")
 		if err != nil {
-			return errors.Wrapf(err, "failed to get interface-name for file %s", nc.FilePath)
+			// When the id field is provided the interface-name will not be present
+			continue
 		}
 		interfaceType, err := connectionSection.GetKey("type")
 		if err != nil {
--- a/pkg/staticnetworkconfig/generator_test.go
+++ b/pkg/staticnetworkconfig/generator_test.go
@@ -306,6 +306,15 @@ var _ = Describe("validate mac interface mapping", func() {
     ipv4:
       enabled: false
       dhcp: false`
+		withMacIdentifier := `interfaces:
+  - name: eth0
+    type: ethernet
+    state: up
+    identifier: mac-address
+    mac-address: f8:75:a4:a4:00:fe
+    ipv4:
+      enabled: false
+      dhcp: false`
 		It("no mapping needed for physical interface", func() {
 			err := staticNetworkGenerator.ValidateStaticConfigParams([]*models.HostStaticNetworkConfig{
 				{
@@ -316,6 +325,20 @@ var _ = Describe("validate mac interface mapping", func() {
 						},
 					},
 					NetworkYaml: withPhysicalInterface,
+				},
+			})
+			Expect(err).ToNot(HaveOccurred())
+		})
+		It("mac-identifier field is allowed", func() {
+			err := staticNetworkGenerator.ValidateStaticConfigParams([]*models.HostStaticNetworkConfig{
+				{
+					MacInterfaceMap: []*models.MacInterfaceMapItems0{
+						{
+							LogicalNicName: "eth0",
+							MacAddress:     "f8:75:a4:a4:00:fe",
+						},
+					},
+					NetworkYaml: withMacIdentifier,
 				},
 			})
 			Expect(err).ToNot(HaveOccurred())
--- a/pkg/staticnetworkconfig/generator_test.go
+++ b/pkg/staticnetworkconfig/generator_test.go
@@ -238,17 +238,6 @@ var _ = Describe("validate mac interface mapping", func() {
     vlan:
       base-iface: eth1
       id: 101`
-		It("vlan without underlying interface - no mapping", func() {
-			err := staticNetworkGenerator.ValidateStaticConfigParams([]*models.HostStaticNetworkConfig{
-				{
-					MacInterfaceMap: nil,
-					NetworkYaml:     withoutUnderlyingInterface,
-				},
-			})
-			Expect(err).To(HaveOccurred())
-			Expect(err.Error()).To(ContainSubstring("no mac address mapping can be associated with the available network interfaces"))
-		})
-
 		It("vlan with underlying interface - no mapping", func() {
 			err := staticNetworkGenerator.ValidateStaticConfigParams([]*models.HostStaticNetworkConfig{
 				{
@@ -306,6 +295,19 @@ var _ = Describe("validate mac interface mapping", func() {
     ipv4:
       enabled: false
       dhcp: false`
+		withNoMappedInterfaces := `interfaces:
+  - name: eno12345
+    type: ethernet
+    state: up
+    ipv4:
+      enabled: false
+      dhcp: false
+  - name: eno12399np0
+    type: ethernet
+    state: up
+    ipv4:
+      enabled: false
+      dhcp: false`
 		withMacIdentifier := `interfaces:
   - name: eth0
     type: ethernet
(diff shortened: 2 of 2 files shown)


zaneb:
We also need to get rid of the error on line 181.

pkg/staticnetworkconfig/generator.go
```diff
@@ -186,7 +187,12 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 		switch interfaceType.String() {
 		case "802-3-ethernet", "ethernet":
 			if !lo.Contains(interfaceNames, interfaceName.String()) {
-				return errors.Errorf("mac-interface mapping for interface %s is missing", interfaceName.String())
+				physInt := regexp.MustCompile("^en[PsvXbucaipod]")
```
linoyaslan, line 190: What if a user selects a logical name that matches the regular expression, but in practice, it's just a randomly chosen logical name? Isn't that a situation that could happen?

ori-amizur:
Previously we had problems (not in ABI), that users failed to provide mapping.  This caused the definition to be stale (i.e the definition cannot be associated with any host).
This PR adds the possibility again that host directory cannot be mapped to any host.

AlonaKaplan:
Please change the PR title to "MGMT-18694: Don't require mapping for names matching physical interfaces", so our release automation will be able to track it.

bfournie:
> We also need to get rid of the error on line 181.

I think we'd always find the interface-name here even if there is no mapping.

pkg/staticnetworkconfig/generator.go
bfournie: In that case, yes, the name would pass validation.

pkg/staticnetworkconfig/generator.go
```diff
@@ -186,7 +187,12 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
...
+				if !physInt.MatchString(interfaceName.String()) {
+					return errors.Errorf("mac-interface mapping for interface %s is missing and not a physical interface", interfaceName.String())
+				} else {
+					s.log.Infof("Interface %s has no mac-interface mapping but matches a physical interface", interfaceName.String())
+				}
 			}
 			matched = true
```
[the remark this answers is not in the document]
bfournie: OK, will fix. Thanks.

pkg/staticnetworkconfig/generator.go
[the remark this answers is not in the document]
bfournie: > "eth*" may be a name of physical interface as well. I would call the variable - 
> predicatableIfaceNamePattern

Will change

> Can you please add a comment with a reference to an official doc specifying this > is the pattern of a predictable interface name.

I will add a note. Per @zaneb comment here [https://issues.redhat.com/browse/OCPBUGS-34849?focusedId=25452379&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25452379] the pattern was obtained from `man systemd.net-naming-scheme`. To summarize, the following are valid fields after the  `en` prefix:
```
x - for ID_NET_NAME_MAC
P or s - ID_NET_NAME_SLOT
v - VIO slot number
X - VIF interface number
b - Broadcom bus
u - USB port number
c - CCW device identifier
a - ACPI path name
i - simulated networking device
p - PCI geogpraphical location bus
o - PCI ethernet card
d - PCI geogpraphical location dev_port
```

pkg/staticnetworkconfig/generator_test.go
```diff
@@ -288,6 +288,39 @@ var _ = Describe("validate mac interface mapping", func() {
...
+		It("no mapping needed for physical interface", func() {
+			err := staticNetworkGenerator.ValidateStaticConfigParams([]*models.HostStaticNetworkConfig{
+				{
+					MacInterfaceMap: []*models.MacInterfaceMapItems0{
+						{
+							LogicalNicName: "eth0",
+							MacAddress:     "f8:75:a4:a4:00:fe",
+						},
+					},
+					NetworkYaml: withPhysicalInterface,
```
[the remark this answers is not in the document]
bfournie: Changed to require one mapping and added a test.

AlonaKaplan:
> > We also need to get rid of the error on line 181.
> 
> I think we'd always find the interface-name here even if there is no mapping.

In case the mac-identifier is used, there won't be interface-name in the keyfile.
If you fix it, please do in a separate PR.

AlonaKaplan:
> Previously we had problems (not in ABI), that users failed to provide mapping. This caused the definition to be stale (i.e the definition cannot be associated with any host). This PR adds the possibility again that host directory cannot be mapped to any host.

We still.require at least one physical interface from the yaml to be associated with the MAC map.

zaneb:
> In case the mac-identifier is used, there won't be interface-name in the keyfile.

Yep, for example:

```
cat <<EOF | nmstatectl gc -
interfaces:
- name: eth0       
  type: ethernet
  state: up
  identifier: mac-address
  mac-address: 00:23:45:67:89:1a
EOF
NetworkManager:
- - eth0.nmconnection
  - |
    [connection]
    autoconnect=true
    autoconnect-slaves=-1
    id=eth0
    type=802-3-ethernet
    uuid=dfd202f5-562f-5f07-8f2a-a7717756fb70

    [ipv4]
    dhcp-timeout=2147483647
    method=disabled

    [ipv6]
    dhcp-timeout=2147483647
    method=disabled

    [ethernet]
    mac-address=00:23:45:67:89:1A
```

> If you fix it, please do in a separate PR.

That would complicate backporting, I think we can just add it to this PR.

zaneb:
/retitle MGMT-18694, OCPBUGS-34849: Don't require mapping for names matching physical interfaces

AlonaKaplan:
> > In case the mac-identifier is used, there won't be interface-name in the keyfile.
> 
> Yep, for example:
> 
> ```
> cat <<EOF | nmstatectl gc -
> interfaces:
> - name: eth0       
>   type: ethernet
>   state: up
>   identifier: mac-address
>   mac-address: 00:23:45:67:89:1a
> EOF
> NetworkManager:
> - - eth0.nmconnection
>   - |
>     [connection]
>     autoconnect=true
>     autoconnect-slaves=-1
>     id=eth0
>     type=802-3-ethernet
>     uuid=dfd202f5-562f-5f07-8f2a-a7717756fb70
> 
>     [ipv4]
>     dhcp-timeout=2147483647
>     method=disabled
> 
>     [ipv6]
>     dhcp-timeout=2147483647
>     method=disabled
> 
>     [ethernet]
>     mac-address=00:23:45:67:89:1A
> ```
> 
> > If you fix it, please do in a separate PR.
> 
> That would complicate backporting, I think we can just add it to this PR.

If it is done in this PR, please add "MGMT-18575" to the title as well.

pkg/staticnetworkconfig/generator.go
```diff
@@ -186,9 +187,16 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
...
-				return errors.Errorf("mac-interface mapping for interface %s is missing", interfaceName.String())
+				// See 'man systemd.net-naming-scheme' for interface naming protocol
+				predicatableIfaceNamePattern := regexp.MustCompile("^en[PsvxXbucaipod]")
+				if !predicatableIfaceNamePattern.MatchString(interfaceName.String()) {
+					return errors.Errorf("mac-interface mapping for interface %s is missing and not a physical interface", interfaceName.String())
+				} else {
+					s.log.Infof("Interface %s has no mac-interface mapping but matches a physical interface", interfaceName.String())
+				}
+			} else {
+				matched = true
```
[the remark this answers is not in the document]
AlonaKaplan: From the API perspective it feels wrong that the "logical name" key in the map has no meaning and can be anything with no association with the yaml.
@linoyaslan can you please verifying that changing the validation to the one Zane suggested won't break the node [script](https://github.com/openshift/assisted-service/blob/9aef3554fca1af297850c359785121fb5b0183f1/internal/constants/scripts.go#L169)?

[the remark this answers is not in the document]
zaneb: We did verify this before building a feature that depended on it.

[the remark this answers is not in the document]
linoyaslan: Currently, in SaaS, there's a validation that checks for a match between the names in the YAML and the names in the map. If we plan to change this, we'll need to modify the UI accordingly.
image

pkg/staticnetworkconfig/generator.go
```diff
@@ -186,9 +187,16 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 		switch interfaceType.String() {
 		case "802-3-ethernet", "ethernet":
 			if !lo.Contains(interfaceNames, interfaceName.String()) {
-				return errors.Errorf("mac-interface mapping for interface %s is missing", interfaceName.String())
+				// See 'man systemd.net-naming-scheme' for interface naming protocol
+				predicatableIfaceNamePattern := regexp.MustCompile("^en[PsvxXbucaipod]")
```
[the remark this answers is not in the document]
bfournie: Fixed.

pkg/staticnetworkconfig/generator.go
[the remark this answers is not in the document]
zaneb: Why? If the SaaS UI has always had a validation that is stricter than the API, and everyone has been happy with that, I don't know of any reason to change it.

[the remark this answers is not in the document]
linoyaslan: My bad,  the validation is actually done on the backend only.

pkg/staticnetworkconfig/generator.go
```diff
@@ -177,7 +180,13 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 		connectionSection := cfg.Section("connection")
 		interfaceName, err := connectionSection.GetKey("interface-name")
 		if err != nil {
-			return errors.Wrapf(err, "failed to get interface-name for file %s", nc.FilePath)
+			interfaceID, err := connectionSection.GetKey("id")
+			if err == nil {
+				s.log.Infof("interface %s found using identifier, mapping not required", interfaceID.String())
+				matched = true
```
AlonaKaplan, line 186: 1. Please move the change to a separate commit with informative commit message explaining why it is done (support mac-identifier). BTW, I think the change in this case should be just ignoring the error and `continue`.
2. Regarding the `matched = true` added here-  `matched` is used to mark that at least one interface in the yaml/connection files matches an entry in the MAC map. Here there was no such matching.
3. Regarding `match` logic in general - As @zaneb mentioned, the "matched" logic is breaking ABI (which in some cases has only a [single](https://github.com/openshift/installer/blob/master/pkg/asset/agent/agentconfig/agenthosts.go#L260-L267) MAC address in the MAC map, that has no corresponding interface in the yaml). We need to remove the `matched` logic from the method here and add a separate validation, verifying the MAC map is not empty. Same here, separate commit please.

bfournie: Moved the check for a entry in interface table into a separate function. Split changes into separate commits.

pkg/staticnetworkconfig/generator.go
```diff
@@ -197,21 +195,14 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 					s.log.Infof("Interface %s has no mac-interface mapping but matches a physical interface", interfaceName.String())
 				}
 			}
-			matched = true
 		case "vlan":
 			vlanSection := cfg.Section("vlan")
-			parent, err := vlanSection.GetKey("parent")
+			_, err := vlanSection.GetKey("parent")
 			if err != nil {
 				return errors.Wrapf(err, "failed to get parent of vlan %s", interfaceName.Name())
```
zaneb, line 202: We're no longer doing any real validation with the VLANs, so we probably don't really need to look at it them at all.

zaneb (commented):
/lgtm

pkg/staticnetworkconfig/generator.go
```diff
@@ -180,7 +180,9 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 		connectionSection := cfg.Section("connection")
 		interfaceName, err := connectionSection.GetKey("interface-name")
 		if err != nil {
-			return errors.Wrapf(err, "failed to get interface-name for file %s", nc.FilePath)
+			// When the id field is provided the interface-name will not be present
+			matched = true
```
AlonaKaplan: There is no match here.

pkg/staticnetworkconfig/generator.go
```diff
@@ -167,7 +168,8 @@ func (s *StaticNetworkConfigGenerator) formatNMConnection(nmConnection string) (
 
 func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterfaceMap models.MacInterfaceMap, networksConfig []StaticNetworkConfigData) error {
 	interfaceNames := lo.Map(macInterfaceMap, func(m *models.MacInterfaceMapItems0, _ int) string { return m.LogicalNicName })
-	var matched bool
+	// See 'man systemd.net-naming-scheme' for interface naming protocol
+	predicatableIfaceNamePattern := regexp.MustCompile("^en[PsvxXbucaipod]")
```
paul-maidment, line 172: I think this will catch most physical interface names, we should note however that some invalid physical interface names will be matched by this, some examples...

`enxABC` - The `enx` prefix is used to denote a physical interface with a MAC address, MAC addresses should have exactly 12 characters.

`enp0` - PCI bus specified but no device specified

and other similar combinations where we include any of the parameter names used for physical interfaces.

Should we make the regex more specific?

bfournie: I believe the intent was to detect names that would be used as placeholders and require the interface-mapping in that case, not to validate all potential physical interface names as discussed in https://issues.redhat.com/browse/OCPBUGS-34849?focusedId=25452379&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25452379.

pkg/staticnetworkconfig/generator.go
bfournie: matched setting has been removed

pkg/staticnetworkconfig/generator.go
bfournie: Removed vlan check.

bfournie:
The edge-e2e-ai-operator-ztp seems to be permafailing for all PRs and is unrelated to this change.
Can we get this overriden?

pkg/staticnetworkconfig/generator.go
AlonaKaplan: I still see it in this commit... I know you removed it eventually but this commit has to be fixed as well.

bfournie: Got it. Updated commit to remove that line.

pkg/staticnetworkconfig/generator.go
linoyaslan: @zaneb  @bfournie  In that case, users can provide a YAML with a VLAN where the base interface is a logical name, and with no mapping of the name associated with a MAC address. Why don’t we want a validation for this scenario?

zaneb: They could also have done that before, because this validation wasn't checking for that. All it did was make sure that an entry in the Name/MAC mapping counted even if the interface wasn't defined in the NMState but a VLAN based on it was.

You _could_ validate something interesting about VLANs - that the parent interface of the VLAN is either defined in the NMState (this would require two passes through the list), is defined in the Name/MAC mapping list, or has a plausible physical name. That may well be worth doing as a separate ticket. But the code being removed did not validate anything comparable to that.

linoyaslan: But previously, if the base interface wasn't included in the mapping, `matched = false`, and the error "no mac address mapping can be associated with the available network interfaces" would be raised, isn't it?

pkg/staticnetworkconfig/generator.go
eifrach: setting `biosdevname=1` will also can break thios match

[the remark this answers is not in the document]
zaneb: OK, but it's strictly better than before this change, right? Previously it failed all of the time when the logical name was not in the mapping, and now it only fails when this pattern is not matched.
ABI does not have an API for adding kernel arguments to the Agent ISO and nor does IPI (which we're trying to maintain config compatibility with). Do we still need to care about biosdevname? Is it automatically enabled on Dell systems in RHCOS?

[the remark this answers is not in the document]
eifrach: it's not affecting only ABI, so we  need to  make  it compatible with all types of installation.

so, biosdevname is set to 0 by default, but in some production environments  where  you have automation for multi-node / multi-vendor you usually set it  to 1 to make the confiugration / scripts  consistent across the datacenter.

You can set it using custom manifest pre installation or with machineConfig post installation

pkg/staticnetworkconfig/generator.go
```diff
@@ -186,23 +189,14 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistence(macInterf
 		switch interfaceType.String() {
 		case "802-3-ethernet", "ethernet":
```
eifrach, line 190: This check is not needed if we  matching only `enXXX` it will be always `ethernet`

pkg/staticnetworkconfig/generator.go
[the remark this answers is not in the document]
zaneb: > You can set it using custom manifest pre installation or with machineConfig post installation

This check is running against the discovery ISO, so it cannot possibly be affected by either custom manifests nor machineConfigs.

pkg/staticnetworkconfig/generator.go
zaneb: I believe they're aliases of each other.

zaneb:
Not sure why the bot removed my LGTM. Weird.
/lgtm

eifrach (approved):

PR #6728, opened 2024-09-05
OCPBUGS-32803: Remove setting of staticNetwork in agent-based installer registration

bfournie:
When registering the agent-based installer with assisted-service, remove the setting of the nmstate info and thus the validations run by assisted-service. Its not required for this field to be set and can cause problems if the nmstate versions used for installer and assisted-service are different.

Confirmed agent-based installer installations succeed when this setting is removed. 
Note that this does not affect the use of static networking in the agent-based installer.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
- [] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/agentbasedinstaller/register.go
+++ b/cmd/agentbasedinstaller/register.go
@@ -154,14 +154,9 @@ func RegisterInfraEnv(ctx context.Context, log *log.Logger, bmInventory *client.
 			return nil, nmStateErr
 		}
 
-		staticNetworkConfig, processErr := processNMStateConfig(log, infraEnv, nmStateConfig)
+		_, processErr := processNMStateConfig(log, infraEnv, nmStateConfig)
 		if processErr != nil {
 			return nil, processErr
-		}
-
-		if len(staticNetworkConfig) > 0 {
-			log.Infof("Added %d nmstateconfigs", len(staticNetworkConfig))
-			infraEnvParams.InfraenvCreateParams.StaticNetworkConfig = staticNetworkConfig
 		}
 	}
 
--- a/subsystem/agent_based_installer_client_test.go
+++ b/subsystem/agent_based_installer_client_test.go
@@ -37,7 +37,6 @@ var _ = Describe("RegisterClusterAndInfraEnv", func() {
 		Expect(registerInfraEnvErr).NotTo(HaveOccurred())
 		Expect(*modelInfraEnv.Name).To(Equal("myinfraenv"))
 		Expect(modelInfraEnv.ClusterID).To(Equal(*modelCluster.ID))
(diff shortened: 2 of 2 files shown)


paul-maidment:
Excuse my ignorance, but the documentation for Agent Based Installer appears to instruct end users on how to configure static networking.

https://docs.openshift.com/container-platform/4.16/installing/installing_with_agent_based_installer/preparing-to-install-with-agent-based-installer.html#static-networking

Why do we not need static networking in this case? I am sure there is a valid reason, I just don't understand why?

andfasano:
> Excuse my ignorance, but the documentation for Agent Based Installer appears to instruct end users on how to configure static networking.
> 
> https://docs.openshift.com/container-platform/4.16/installing/installing_with_agent_based_installer/preparing-to-install-with-agent-based-installer.html#static-networking
> 
> Why do we not need static networking in this case? I am sure there is a valid reason, I just don't understand why?

In ABI, the .nmconnection files are generated when the image is created and stored in the ISO in the `/etc/assisted/network`. Then, the (very early) boot service [pre-network-manager-config](https://github.com/openshift/installer/blob/d61300fbc2107236d85ecb7e60bd1d2fabc8ed51/pkg/asset/agent/manifests/network-scripts.go#L87) takes care of applying them.

bfournie:
> > Excuse my ignorance, but the documentation for Agent Based Installer appears to instruct end users on how to configure static networking.
> > https://docs.openshift.com/container-platform/4.16/installing/installing_with_agent_based_installer/preparing-to-install-with-agent-based-installer.html#static-networking
> > Why do we not need static networking in this case? I am sure there is a valid reason, I just don't understand why?
> 
> In ABI, the .nmconnection files are generated when the image is created and stored in the ISO in the `/etc/assisted/network`. Then, the (very early) boot service [pre-network-manager-config](https://github.com/openshift/installer/blob/d61300fbc2107236d85ecb7e60bd1d2fabc8ed51/pkg/asset/agent/manifests/network-scripts.go#L87) takes care of applying them.

Thanks Andrea. Yes I think the title is misleading Paul as this doesn't affect the use of static networking in the agent-based installer. Its only affect is to not pass the parameter into assisted-service where another validation is done, potentially using a version of nmstate that differs with the one the agent-installer used when generating the ISO.

bfournie:
Closing this as it causes problems with static networking not being set after booting into the final image (see the note in the bug).
/close

PR #6730, opened 2024-09-08
MGMT-18635: Restore missing Host by Agent CR

danielerez:
When applying an Agent CR with a missing associated host, the Agnet CR is currently being deleted by the reconciler. This scenario would happen on restore to a new hub flows since the host is a DB only entity.

Thus, this change introduces handling of a missing host by restoring the object into DB according to properties from the Agent CR.
For restoring the host's status, added 'state' annotation to the agent which is set on reconcile.

See full flow and more details in the enhancement: https://github.com/openshift/assisted-service/blob/master/docs/enhancements/backup-restore-support.md#solution-for-deleted-agent-crs

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -179,6 +179,7 @@ type InstallerInternals interface {
 	ValidatePullSecret(secret string, username string, releaseImageURL string) error
 	GetInfraEnvInternal(ctx context.Context, params installer.GetInfraEnvParams) (*common.InfraEnv, error)
 	V2UpdateHostInstallProgressInternal(ctx context.Context, params installer.V2UpdateHostInstallProgressParams) error
+	CreateHostInKubeKeyNamespace(ctx context.Context, kubeKey types.NamespacedName, host *models.Host) error
 }
 
 //go:generate mockgen --build_flags=--mod=mod -package bminventory -destination mock_crd_utils.go . CRDUtils
@@ -5230,6 +5231,20 @@ func (b *bareMetalInventory) GetInfraEnvByKubeKey(key types.NamespacedName) (*co
 		return nil, err
 	}
 	return infraEnv, nil
+}
+
+func (b *bareMetalInventory) CreateHostInKubeKeyNamespace(ctx context.Context, kubeKey types.NamespacedName, host *models.Host) error {
+	log := logutil.FromContext(ctx, b.log)
+	hostToCreate := &common.Host{
+		Host:                    *host,
+		KubeKeyNamespace:        kubeKey.Namespace,
+		TriggerMonitorTimestamp: time.Now(),
+	}
+	log.Infof("Create host %s infra env %s", host.ID.String(), host.InfraEnvID)
+	if err := b.db.Create(hostToCreate).Error; err != nil {
+		return err
+	}
+	return nil
 }
 
 func (b *bareMetalInventory) V2RegisterHost(ctx context.Context, params installer.V2RegisterHostParams) middleware.Responder {
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -32,6 +32,7 @@ import (
 	bmh_v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	. "github.com/openshift/assisted-service/api/common"
+	"github.com/openshift/assisted-service/api/v1beta1"
 	aiv1beta1 "github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/internal/bminventory"
 	"github.com/openshift/assisted-service/internal/common"
@@ -65,6 +66,7 @@ import (
 const (
 	AgentFinalizerName                   = "agent." + aiv1beta1.Group + "/ai-deprovision"
 	AgentSkipSpokeCleanupAnnotation      = "agent." + aiv1beta1.Group + "/skip-spoke-cleanup"
+	AgentStateAnnotation                 = "agent." + aiv1beta1.Group + "/state"
 	BaseLabelPrefix                      = aiv1beta1.Group + "/"
 	InventoryLabelPrefix                 = "inventory." + BaseLabelPrefix
 	AgentLabelHasNonrotationalDisk       = InventoryLabelPrefix + "storage-hasnonrotationaldisk"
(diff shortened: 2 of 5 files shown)


internal/controller/controllers/agent_controller.go
```diff
@@ -164,6 +173,10 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 		log.WithError(err).Warnf("failed to set infraEnv name label on agent %s/%s", agent.Namespace, agent.Name)
 	}
 
+	if err = r.setStateAnnotation(ctx, log, h, agent); err != nil {
```
carbonin, line 176: Should we maybe do this in the same place we update the status?

internal/bminventory/inventory.go
```diff
@@ -5211,6 +5212,22 @@ func (b *bareMetalInventory) GetInfraEnvByKubeKey(key types.NamespacedName) (*co
 	return infraEnv, nil
 }
 
+func (b *bareMetalInventory) CreateHostInKubeKeyNamespace(ctx context.Context, kubeKey types.NamespacedName, host *models.Host) error {
+	log := logutil.FromContext(ctx, b.log)
+	log.Infof("Register host: %+v (KubeKey: %+v)", host, kubeKey)
```
carbonin, line 5217: I think the `NamespacedName` type has a `String()` method so you should be able to just print it with `%s`, but also I'd imagine this message is just for debugging since we probably don't want to print the whole host as well?

internal/controller/controllers/agent_controller.go
```diff
@@ -1761,3 +1767,99 @@ func (r *AgentReconciler) updateHostInstallProgress(ctx context.Context, host *m
...
+	if err != nil {
+		r.Log.WithError(err).Errorf("Failed to create Host with name '%s'", agent.Name)
+		return err
+	}
+
+	return r.Installer.CreateHostInKubeKeyNamespace(ctx, key, host)
+}
+
+func createNewHost(agent *v1beta1.Agent, clusterID, infraEnvID *strfmt.UUID) (*models.Host, error) {
+	if infraEnvID == nil {
```
carbonin, line 1819: Why not just pass this by value instead of reference?

internal/controller/controllers/agent_controller.go
```diff
@@ -1761,3 +1767,99 @@ func (r *AgentReconciler) updateHostInstallProgress(ctx context.Context, host *m
...
+			hostStatus = models.HostStatusKnownUnbound
+		}
+	}
+
+	// Create Host model
+	hostID := strfmt.UUID(agent.Name)
+	url := installer.V2GetHostURL{InfraEnvID: *infraEnvID, HostID: hostID}
+	host := &models.Host{
+		ID:            &hostID,
+		Href:          swag.String(url.String()),
```
carbonin, line 1852: Is this set for the kubeAPI? I suppose it isn't expose to the user in any case, but I feel like this API wouldn't be available.

internal/bminventory/inventory.go
```diff
@@ -5211,6 +5212,22 @@ func (b *bareMetalInventory) GetInfraEnvByKubeKey(key types.NamespacedName) (*co
...
+
+	hostToCreate := &common.Host{
+		Host:                    *host,
+		KubeKeyNamespace:        kubeKey.Namespace,
+		TriggerMonitorTimestamp: time.Now(),
+	}
+	log.Infof("Create host %s infra env %s", host.ID.String(), host.InfraEnvID)
```
CrystalChun, line 5224: Should we validate the infra env actually exists before creating the host or was that done elsewhere? 
Is it valid to create a host that references an infra env that doesn't exist anymore?

internal/controller/controllers/agent_controller.go
carbonin: Sorry on my first pass I thought this was an entire copy of the status (like BMH) and then it made more sense to save it to an annotation at the same time we updated it in the CR.

danielerez: np, I'll revert to the previous approach:)

internal/controller/controllers/agent_controller.go
```diff
@@ -145,7 +147,14 @@ func (r *AgentReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (
 	h, err := r.Installer.GetHostByKubeKey(req.NamespacedName)
 	if err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
-			return r.deleteAgent(ctx, log, req.NamespacedName)
+			// Restoring the Host according to values from Agent
+			if err = r.restoreHostByAgent(ctx, agent); err != nil {
```
carbonin, line 151: IIRC we gated the BMH restore stuff with a flag, do we want to do the same here or should this be mostly harmless?

internal/controller/controllers/agent_controller.go
```diff
@@ -1712,6 +1712,13 @@ func (r *AgentReconciler) setInfraEnvNameLabel(ctx context.Context, log logrus.F
 	return nil
 }
 
+func (r *AgentReconciler) setStateAnnotation(ctx context.Context, log logrus.FieldLogger, h *common.Host, agent *aiv1beta1.Agent) error {
```
carbonin, line 1715: Do you really think this needs to be a separate function?

I'm not sure how much we're gaining by splitting this out.

carbonin (commented):
Just some minor stuff, looks good otherwise.

internal/controller/controllers/agent_controller.go
danielerez: So it seems the logic was originally added merely for cleanup purposes. I.e. the removal of agent without an associated host shouldn't have affected the behavior. So since a missing host shouldn't be a common scenario other than in restore flows, I'm not sure we should have a feature gate for it. I mean, it doesn't seem to have much impact on current flows. Unless you think it's safer to still have a flag to disable it?

internal/controller/controllers/agent_controller.go
danielerez: Well the Reconcile func is getting quite large:/ So for readability sake, wanted to keep it clean.
I'll try to remove it and let's see how it looks.

internal/controller/controllers/agent_controller.go
carbonin: Okay let's leave it as is then.

carbonin (approved):

PR #6764, opened 2024-09-17
MGMT-17230: Use IDMS when possible

jhernand:
Currently when we run the 'oc' command in disconnected environments whe generate a `ImageContentSourcePolicy` object inside a YAML file and reference it with the `--icsp-file` command line flag. That is deprecated since OpenShift 4.14, and replaced by the `ImageDigestMirrorSet` object and the `--idms-file` command line flag.

This patch changes the code so that it checks if the `--idms-file` command line flag is available, and if it is it generates a `ImgeDigestMirrorSet` and uses the `--idms-file` flag instead of the deprecated `--icsp-file`.

The previous behaviour is preserved when the `--idms-file` flag isn't available in order to keep supporting older OpenShift versions.

Related: https://issues.redhat.com/browse/MGMT-17230
Related: https://github.com/openshift/enhancements/blob/master/enhancements/api-review/add-new-CRD-ImageDigestMirrorSet-and-ImageTagMirrorSet-to-config.openshift.io.md

- [ ] New Feature 
- [X] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/oc/release.go
+++ b/internal/oc/release.go
@@ -5,12 +5,16 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
+	"slices"
+	"sort"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/buger/jsonparser"
 	"github.com/hashicorp/go-version"
+	configv1 "github.com/openshift/api/config/v1"
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/system"
@@ -33,6 +37,14 @@ const (
 	DefaltRetryDelay               = time.Second * 5
 	staticInstallerRequiredVersion = "4.16.0-0.alpha"
 )
+
+// icspFileFlagName is the name of the command line flag used to indicate a file containg an ImageContentSourcePolicy
+// object. Note that this is deprecated since OpenShift 4.14.
+const icspFileFlagName = "icsp-file"
+
+// idmsFileFlagName is the name of the command line flag used to indicate a file containg an ImageDigestMirrorSet
+// object, which is the recommended way to configure image mirrors since OpenShift 4.14.
+const idmsFileFlagName = "idms-file"
 
 type Config struct {
 	MaxTries   uint
@@ -79,14 +91,11 @@ func NewRelease(executer executer.Executer, config Config, mirrorRegistriesBuild
 }
 
 const (
-	templateGetImage              = "oc adm release info --image-for=%s --insecure=%t %s"
-	templateGetImageWithIcsp      = "oc adm release info --image-for=%s --insecure=%t --icsp-file=%s %s"
-	templateGetVersion            = "oc adm release info -o template --template '{{.metadata.version}}' --insecure=%t %s"
-	templateGetVersionWithIcsp    = "oc adm release info -o template --template '{{.metadata.version}}' --insecure=%t --icsp-file=%s %s"
-	templateExtract               = "oc adm release extract --command=%s --to=%s --insecure=%t %s"
-	templateExtractWithIcsp       = "oc adm release extract --command=%s --to=%s --insecure=%t --icsp-file=%s %s"
-	templateImageInfo             = "oc image info --output json %s"
-	templateImageInfoWithIcsp     = "oc image info --output json --icsp-file=%s %s"
+	templateReleaseInfo           = "oc adm release info"
+	templateGetImage              = "oc adm release info --image-for=%s --insecure=%t %s %s"
+	templateGetVersion            = "oc adm release info -o template --template '{{.metadata.version}}' --insecure=%t %s %s"
+	templateExtract               = "oc adm release extract --command=%s --to=%s --insecure=%t %s %s"
+	templateImageInfo             = "oc image info --output json %s %s"
 	templateSkopeoDetectMultiarch = "skopeo inspect --raw --no-tags docker://%s"
 	ocAuthArgument                = " --registry-config="
 	skopeoAuthArgument            = " --authfile "
@@ -121,21 +130,21 @@ func (r *release) getImageByName(log logrus.FieldLogger, imageName, releaseImage
 		return "", errors.New("neither releaseImage, nor releaseImageMirror are provided")
 	}
 
-	icspFile, err := r.getIcspFileFromRegistriesConfig(log)
+	mirrorsFlag, err := r.getMirrorsFlagFromRegistriesConfig(log)
 	if err != nil {
-		return "", errors.Wrap(err, "failed to create file ICSP file from registries config")
+		return "", errors.Wrap(err, "failed to get mirrors info from registries config")
 	}
-	defer removeIcspFile(icspFile)
+	defer mirrorsFlag.Delete()
 
 	if releaseImageMirror != "" {
 		//TODO: Get mirror registry certificate from install-config
-		image, err = r.getImageFromRelease(log, imageName, releaseImageMirror, pullSecret, icspFile, true)
+		image, err = r.getImageFromRelease(log, imageName, releaseImageMirror, pullSecret, mirrorsFlag, true)
 		if err != nil {
 			log.WithError(err).Errorf("failed to get %s image from mirror release image %s", imageName, releaseImageMirror)
 			return "", err
 		}
 	} else {
-		image, err = r.getImageFromRelease(log, imageName, releaseImage, pullSecret, icspFile, false)
+		image, err = r.getImageFromRelease(log, imageName, releaseImage, pullSecret, mirrorsFlag, false)
 		if err != nil {
 			log.WithError(err).Errorf("failed to get %s image from release image %s", imageName, releaseImage)
 			return "", err
@@ -151,21 +160,21 @@ func (r *release) GetOpenshiftVersion(log logrus.FieldLogger, releaseImage strin
 		return "", errors.New("no releaseImage nor releaseImageMirror provided")
 	}
 
-	icspFile, err := r.getIcspFileFromRegistriesConfig(log)
+	mirrorsFlag, err := r.getMirrorsFlagFromRegistriesConfig(log)
 	if err != nil {
-		return "", errors.Wrap(err, "failed to create file ICSP file from registries config")
+		return "", errors.Wrap(err, "failed to create file mirrors file from registries config")
 	}
-	defer removeIcspFile(icspFile)
+	defer mirrorsFlag.Delete()
 
 	if releaseImageMirror != "" {
 		//TODO: Get mirror registry certificate from install-config
-		openshiftVersion, err = r.getOpenshiftVersionFromRelease(log, releaseImageMirror, pullSecret, icspFile, true)
+		openshiftVersion, err = r.getOpenshiftVersionFromRelease(log, releaseImageMirror, pullSecret, mirrorsFlag, true)
 		if err != nil {
 			log.WithError(err).Errorf("failed to get image openshift version from mirror release image %s", releaseImageMirror)
 			return "", err
 		}
 	} else {
-		openshiftVersion, err = r.getOpenshiftVersionFromRelease(log, releaseImage, pullSecret, icspFile, false)
+		openshiftVersion, err = r.getOpenshiftVersionFromRelease(log, releaseImage, pullSecret, mirrorsFlag, false)
 		if err != nil {
 			log.WithError(err).Errorf("failed to get image openshift version from release image %s", releaseImage)
 			return "", err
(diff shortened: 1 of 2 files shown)


jhernand:
@CrystalChun please take a look.

internal/oc/release_test.go
```diff
@@ -711,8 +890,113 @@ var _ = Describe("GetReleaseBinaryPath", func() {
...
+		ctrl = gomock.NewController(GinkgoT())
+		mockExecuter = executer.NewMockExecuter(ctrl)
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
+
+	DescribeTable(
+		"With correct help text",
```
paul-maidment, line 909: I think we should have a negative test case here too

Especially:
- What if we supply a help command that returns an error (maybe the executable is not installed for example?)
- What happens if the output is not as expected, do we get the correct results?

internal/oc/release_test.go
```diff
@@ -570,45 +572,222 @@ var _ = Describe("getIcspFileFromRegistriesConfig", func() {
...
+			Expect(mirrorsFlag).ShouldNot(BeNil())
+			Expect(mirrorsFlag.file).ShouldNot(BeEmpty())
+			data, err := os.ReadFile(mirrorsFlag.file)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(data).Should(MatchYAML(dedent.Dedent(`
+				apiVersion: config.openshift.io/v1
+				kind: ImageDigestMirrorSet
+				metadata:
+				  creationTimestamp: null
+				  name: image-policy
```
paul-maidment, line 654: Am I being picky to suggest that even in the test, we should not call this a policy?

Same comment for all other naming within ImageDigestMirrorSet tests

internal/oc/release.go
```diff
@@ -513,8 +652,62 @@ func getIcspContents(mirrorConfig []mirrorregistries.RegistriesConf) ([]byte, er
...
 }
 
-func removeIcspFile(filename string) {
-	if filename != "" {
-		os.Remove(filename)
-	}
+// getFlagsFromHelp gets the names of the flags of the given command. To do so it removes all the flags from the
+// command, executes it with the '--help' flag and parses the output, assuming that it has the format used by the 'oc'
+// command. It returns an array with the names of the flags in alphabetical order.
+func getFlagsFromHelp(log logrus.FieldLogger, executer executer.Executer, command string) (result []string, err error) {
```
paul-maidment, line 658: I think I would like to see more tests around extraction of flags from help specifically, just so we can see that edge cases have been covered. 

I know that we potentially cover this in tests later though, so not sure if this is excessive.

This might be a case of testing this function and then non longer having to worry about edge cases in tests later.

paul-maidment (changes requested):
Looks mostly good actually, I just have some questions around tests.

internal/oc/release_test.go
jhernand: Will rename it to "image-mirror-set". Does that look better? Any other suggestion? Note that these (both ICSP and IDMS) are written to temporary files, only for use while running the various "oc" commands, after that they are deleted. So the name isn't really important (unless I miss something).

internal/oc/release_test.go
jhernand: I added a few tests, including one to verify the behavior when the command doesn't exist and another to verify the behavior when the command returns an exit code other than zero. Let me know what you think.

internal/oc/release.go
jhernand: I added some more cases, please take a look again.

jhernand:
/hold

Holding while I manually verify it.

jhernand:
@paul-maidment I had to rebase this in order to use the `Dedent` function introduced in a a separate pull request (see #6781), so you `/lgtm` is gone. Would you be so kind as to review again?

internal/oc/release.go
```diff
@@ -396,12 +393,7 @@ func (r *release) extractFromRelease(log logrus.FieldLogger, releaseImage, cache
 		return "", err
 	}
 
-	var cmd string
-	if icspFile == "" {
-		cmd = fmt.Sprintf(templateExtract, binary, workdir, insecure, releaseImage)
-	} else {
-		cmd = fmt.Sprintf(templateExtractWithIcsp, binary, workdir, insecure, icspFile, releaseImage)
-	}
+	cmd := fmt.Sprintf(templateExtract, binary, workdir, insecure, mirrorsFlag, releaseImage)
```
CrystalChun, line 396: Sorry, I've never seen it like this before, but does `Sprintf` expand `mirrorsFlag`? In other words, you don't need to do 
```
cmd := fmt.Sprintf(templateExtract, binary, workdir, insecure, mirrorsFlag.String(), releaseImage)
```

internal/oc/release.go
```diff
@@ -442,21 +444,158 @@ func execute(log logrus.FieldLogger, executer executer.Executer, pullSecret stri
...
 
-	mirrorRegistriesConfig, err := r.mirrorRegistriesBuilder.ExtractLocationMirrorDataFromRegistries()
+	idmsFile, err := os.CreateTemp("", "idms-file")
 	if err != nil {
-		log.WithError(err).Errorf("Failed to get the mirror registries needed for ImageContentSources")
 		return "", err
 	}
+	log.Debugf("Building IDMS file from registries.conf with contents %s", contents)
+	if _, err := idmsFile.Write(contents); err != nil {
+		idmsFile.Close()
```
CrystalChun, line 547: What do you think about doing this in the function body?
```
defer idmsFile.Close()
```

internal/oc/release.go
```diff
@@ -442,21 +444,158 @@ func execute(log logrus.FieldLogger, executer executer.Executer, pullSecret stri
...
+	if err != nil {
+		return
+	}
+	var (
+		name string
+		file string
+	)
+	if suportsIdmsFileFlag {
+		name = idmsFileFlagName
+		file, err = r.getIdmsFileFromRegistriesConfig(log, mirrorsRegistriesConfig)
```
CrystalChun, line 477: If `file` comes back empty like [here](https://github.com/openshift/assisted-service/pull/6764/files#diff-3b8a020ef087b81b6b45d56bef6dd8d32fcf9d50dfe3d91fd369755b253cfff8R538) and error is nil, will that affect the formatting of the final [command](https://github.com/openshift/assisted-service/pull/6764/files#diff-3b8a020ef087b81b6b45d56bef6dd8d32fcf9d50dfe3d91fd369755b253cfff8R396)?

CrystalChun:
This is looking good so far! Just a few questions from my end

internal/oc/release.go
jhernand: No, that explicit conversion isn't needed, the printf-like strings do that conversion automatically. From https://pkg.go.dev/fmt:

> 5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

Se an example here: https://go.dev/play/p/3m2nGM1qsIc

internal/oc/release.go
jhernand: That would result in `--icsp-file=`, without a value. But I believe that can't happen due to how the `getIdmsFileFromRegistrationConfig` is written: it can only return an empty file if the `os.CreateTemp` function returns an empty file and no error. Anyhow I can explicitly check it here if you think it is convenient.

internal/oc/release.go
jhernand: Yes, makes sense. I tried to not change the existing "getIcspFileFromRegistriesConfig" function, and this is just a copy. I will change it.

internal/oc/release.go
jhernand: Added a check in this commit: https://github.com/openshift/assisted-service/pull/6764/commits/49952f2106fde101384028e215a4ccaf8ac2236a . Will squash when this is ready to merge.

internal/oc/release.go
jhernand: Actually after reading this more carefully I think that doing `defer idmsFile.Close()` will not work correctly in all cases: if writing to the file fails it would then try to close the file after removing it, which would fail.

jhernand: We could do something like this if we wanted to avoid repeating the `Close` call:

```diff
diff --git a/internal/oc/release.go b/internal/oc/release.go
index a37a6d7b1..b50bb113d 100644
--- a/internal/oc/release.go
+++ b/internal/oc/release.go
@@ -546,13 +546,17 @@ func (r *release) getIdmsFileFromRegistriesConfig(log logrus.FieldLogger,
        if err != nil {
                return "", err
        }
+       defer func() {
+               _ = idmsFile.Close()
+               if err != nil {
+                       _ = os.Remove(idmsFile.Name())
+                       result = ""
+               }
+       }()
        log.Debugf("Building IDMS file from registries.conf with contents %s", contents)
        if _, err := idmsFile.Write(contents); err != nil {
-               idmsFile.Close()
-               os.Remove(idmsFile.Name())
                return "", err
        }
-       idmsFile.Close()
 
        return idmsFile.Name(), nil
 }
```

But that is probably more obscure than the current code.

internal/oc/release.go
CrystalChun: Very nice, thanks for the explanation!

internal/oc/release.go
CrystalChun: Hmm what about having a check here instead: https://github.com/openshift/assisted-service/pull/6764/files#diff-3b8a020ef087b81b6b45d56bef6dd8d32fcf9d50dfe3d91fd369755b253cfff8R515
To see if file and name are both set to a string? and return "" if not

E.g.
```
func (i *mirrorsFlagInfo) String() string {
	if i == nil {
		return ""
	}
         if i.name == "" || i.file == "" {
		return ""
	}
	return fmt.Sprintf("--%s=%s", i.name, i.file)
}
```

I'm a bit worried that returning in error will cause the function [here](https://github.com/openshift/assisted-service/pull/6764/files#diff-3b8a020ef087b81b6b45d56bef6dd8d32fcf9d50dfe3d91fd369755b253cfff8R333) to error out, which shouldn't be the case because sometimes there are no mirror registries configured and the command should still run

jhernand: No problem, done.

internal/oc/release.go
```diff
@@ -442,21 +444,162 @@ func execute(log logrus.FieldLogger, executer executer.Executer, pullSecret stri
...
+	name string
+
+	// file is the name of a temporary file containing the serialized ImageDigestMirrorSet or
+	// ImageContentSourcePolicy.
+	file string
+}
+
+// String returns the complete text of the flag, something like '--idms-file=/tmp/my-idms.yaml'.
+func (i *mirrorsFlagInfo) String() string {
+	if i == nil || i.name == "" || i.file == "" {
```
jhernand, line 509: @CrystalChun here ^ is the additional check you suggested.

internal/oc/release.go
CrystalChun: Oh I see, thanks for looking into it! The original way you had it looks good then

internal/oc/release.go
CrystalChun: Looks great!

internal/oc/release.go
```diff
@@ -442,21 +444,158 @@ func execute(log logrus.FieldLogger, executer executer.Executer, pullSecret stri
...
 
-// Create a temporary file containing the ImageContentPolicySources
-func (r *release) getIcspFileFromRegistriesConfig(log logrus.FieldLogger) (string, error) {
+// whiteSpaceRE is the regular expression used to split commands into arguments, taking into account that the separator
+// can be one or multiple whilte spaces.
+var whiteSpaceRE = regexp.MustCompile(`\s+`)
 
+// getMirrorsFromRegistriesConfig returns the information needed to generate the command line flag that specifies the
+// mirrors configuration used by the 'oc' command.
+func (r *release) getMirrorsFlagFromRegistriesConfig(log logrus.FieldLogger) (result *mirrorsFlagInfo, err error) {
```
CrystalChun, line 453:
```suggestion
// getMirrorsFlagFromRegistriesConfig returns the information needed to generate the command line flag that specifies the
// mirrors configuration used by the 'oc' command.
func (r *release) getMirrorsFlagFromRegistriesConfig(log logrus.FieldLogger) (result *mirrorsFlagInfo, err error) {
```

internal/oc/release_test.go
```diff
@@ -570,45 +571,222 @@ var _ = Describe("getIcspFileFromRegistriesConfig", func() {
...
-		mockMirrorRegistriesConfigBuilder.EXPECT().ExtractLocationMirrorDataFromRegistries().Return(regData, nil).Times(1)
-		expected := "apiVersion: operator.openshift.io/v1alpha1\nkind: ImageContentSourcePolicy\nmetadata:\n  creationTimestamp: null\n  name: image-policy\nspec:\n  repositoryDigestMirrors:\n  - mirrors:\n    - host1.example.org:5000/localimages\n    - host1.example.org:5000/openshift\n    source: registry.ci.org\n  - mirrors:\n    - host1.example.org:5000/localimages\n    source: quay.io\n"
-		icspFile, err := oc.getIcspFileFromRegistriesConfig(log)
-		Expect(err).ShouldNot(HaveOccurred())
-		Expect(icspFile).ShouldNot(Equal(""))
-		data, err := os.ReadFile(icspFile)
-		Expect(err).ShouldNot(HaveOccurred())
-		Expect(string(data)).Should(Equal(expected))
-	})
+	Context("With IDSM support", func() {
```
CrystalChun, line 574:
```suggestion
	Context("With IDMS support", func() {
```

internal/oc/release_test.go
```diff
@@ -570,45 +571,222 @@ var _ = Describe("getIcspFileFromRegistriesConfig", func() {
...
 	})
 
-	It("mirror_registries_invalid", func() {
-		mockMirrorRegistriesConfigBuilder.EXPECT().IsMirrorRegistriesConfigured().Return(true).Times(1)
-		mockMirrorRegistriesConfigBuilder.EXPECT().ExtractLocationMirrorDataFromRegistries().Return(nil, fmt.Errorf("extract failed")).Times(1)
-		icspFile, err := oc.getIcspFileFromRegistriesConfig(log)
-		Expect(err).Should(HaveOccurred())
-		Expect(err).Should(MatchError("extract failed"))
-		Expect(icspFile).Should(Equal(""))
+	Context("Without IDSM support, but with deprecated ICSP support", func() {
```
CrystalChun, line 684:
```suggestion
	Context("Without IDMS support, but with deprecated ICSP support", func() {
```

internal/oc/release.go
jhernand: Done.

internal/oc/release_test.go
jhernand: Done.

internal/oc/release_test.go
jhernand: Done.

CrystalChun (approved):
/lgtm

Thanks Juan!

PR #6773, opened 2024-09-18
MGMT-18696: OCI: Download network configuration script from user data

adriengentil:
Currently, the network configuration script on OCI is duplicated on Red
Hat side to configure the instance during discovery, and on Oracle side
to configure the network on first boot (since this script is planned to
be run on machine with iSCSI boot volumes, `--copy-network` is not
supported in this case when installing RHCOS).

This is not convenient for maintenance as it requires both teams to be
in sync if a change happens.

Instead, we plan to inject the network configuration script into the
instance metadata using terraform. We now just need a unit to download
the script, and execute it during discovery and first boot. The network
configuration script would be maintained by the Oracle team.


internal/ignition/templates/discovery.ign
```diff
@@ -87,13 +87,14 @@
       "contents": { "source": "data:text/plain;charset=utf-8;base64,{{ executeTemplate "multipath.conf" . | toBase64 }}" }
     },
     {
-      "path": "/usr/local/bin/iscsi-oci-configure-secondary-nic.sh",
```
TDBank007, line 90: PLS RECONFIGURE THE SUBDIVISION

PR #6784, opened 2024-09-19
AGENT-949 AGENT-951: Add new authz handler and swagger security defn.

pawanpinjarkar:
The agent-based installer needs a new security definition to perform the read-only operations for the user persona `watcher`. The agent-based installer commands such as `wait-for` and `monitor` internally use the read-only persona `watcher` which can only make read requests to the endpoints annotated with `watcherAuth` security definition.

Other existing authenticators do not support this agent-based installer specific `watcherAuth` security definition.

- Read and set USER_AUTH_TOKEN
- Implememnt a new authorization handler agent_local_authz_handler
- Implement new security definition AuthWatcherAuth

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
Tested locally with dev-scripts using the installer PR https://github.com/openshift/installer/pull/9039
- [x] No tests needed

## Checklist

## Reviewers Checklist


carbonin:
So this looks okay, but is it really what you want to do?

As far as I can tell this just adds another way to provide the same token, so you're not really adding any additional security because someone with that token could provide it using a different header and still use other endpoints.

For this to actually achieve authorization you'd need to issue separate tokens that have some kind of claim that identifies them as read-only. Then in the token verification process somewhere ensure that a read-only token is only valid for the endpoints it's assigned to.

pawanpinjarkar:
> So this looks okay, but is it really what you want to do?
> 
> As far as I can tell this just adds another way to provide the same token, so you're not really adding any additional security because someone with that token could provide it using a different header and still use other endpoints.
> 
> For this to actually achieve authorization you'd need to issue separate tokens that have some kind of claim that identifies them as read-only. Then in the token verification process somewhere ensure that a read-only token is only valid for the endpoints it's assigned to.

yes, so my plan is to have the swagger changes from this PR. Then in a different PR, implement a new authz handler in assisted service for ABI and update the installer to create 3 seperate tokens. Code changes from all the 3 PRs then will do the intended operation.

carbonin:
I think it would be fine to add the authz stuff in this same PR. It gives context that makes these changes worthwhile.

Also I wouldn't want to have to go back and change the swagger again if we run into issues later with the next patch.

swagger.yaml
```diff
@@ -53,6 +53,10 @@ securityDefinitions:
     type: apiKey
     in: query
     name: image_token
+  watcherAuth:
+    type: apiKey
+    in: header
+    name: Watcher-Authorization
 
 security:
   - userAuth: [admin, user]
```
carbonin, line 62: I also think this might be the slightly more idiomatic way of achieving what we're doing with an additional security definition, but I don't see much real upside to putting this in the roles rather than a separate definition.

If you want to investigate it, then go ahead, but I won't hold the PR up for it.

pkg/auth/agent_local_authz_handler.go
```diff
@@ -0,0 +1,84 @@
...
+	"gorm.io/gorm"
+)
+
+/* AgentLocalAuthzHandler is the authorizer middleware that is being used for
+ * agent based installer (ABI) authentication cases
+ */
+type AgentLocalAuthzHandler struct {
+	cfg *Config
+	log logrus.FieldLogger
+	db  *gorm.DB
```
carbonin, line 23: Do you actually use any of these anywhere?

carbonin (commented):
One small comment, otherwise looks good to me assuming it's passing your manual testing with the ABI.

pkg/auth/agent_local_authz_handler.go
pawanpinjarkar: Good point, not needed. Removed!

pawanpinjarkar:
/hold
need to make sure the installer PR https://github.com/openshift/installer/pull/9039  lands simultaneously.

carbonin (approved):

swagger.yaml
```diff
@@ -53,6 +53,10 @@ securityDefinitions:
     type: apiKey
     in: query
     name: image_token
+  watcherAuth:
+    type: apiKey
+    in: header
+    name: Watcher-Authorization
```
zaneb, line 59: `Authorization` is an [officially-defined](https://datatracker.ietf.org/doc/html/rfc9110#section-11.6.2) header, but when you're creating your own you would typically add an `X-` prefix.

pawanpinjarkar: Looks like usage of `X-` is deprecated as per https://datatracker.ietf.org/doc/html/rfc6648

The name `Watcher-Authorization` aligns with the below recommendations -
Avoidance of X- Prefix, descriptive and unique naming, naming based on application context

This approach fits within the broader guidance from MDN’s [HTTP Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers), where unique, non-conflicting names are emphasized.

rwsu:
/test e2e-agent-compact-ipv4

Last failure appears to be infrastructure related:
```
Trying to pull quay.io/openshifttest/squid-proxy:multiarch...
Getting image source signatures
Copying blob sha256:c5fccdba25e8a523e654a2ee169250141a5741e8844dee063cbbbf0e5e2d4bb9
Copying blob sha256:970e121bff396815d64dc4e784589369b310fb6630281804ca684a7d34df7c63
Copying blob sha256:0d1baa9a1e65a2a5993ffaad35ff54cfb88cd20846309f34746d67cc421dbfe9
Copying blob sha256:3662bf1aea374e1e6919b47fac27665af0b8c1d86189ae43b5527c663ef79ce4
Copying blob sha256:07222971243684a01bce78451c72c5b314109024de2d66bf7c31386f6d68d72b
Copying blob sha256:70fb9965a23f2226fef622992fdf507b8333c61d68259766d4721cc4ba1e5dae
Copying blob sha256:dbb1afa921ac3695aad1cad01d9f2d53e2a983b8ee52aa34109463de403e4fcc
Copying config sha256:4f135017348a2539121a6476c0f8bc32b99ff8008f205a5eed5541a98d971989
Writing manifest to image destination
Error: OCI runtime error: crun: unknown version specified
{"component":"entrypoint","error":"wrapped process failed: exit status 126","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:84","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2024-11-13T00:42:12Z"}
error: failed to execute wrapped command: exit status 126
```

PR #5365, opened 2023-07-18
MGMT-14923: add OSImageVersion to InfraEnvSpec

danielerez:
Introduced a new property to InfraEnvSpec: `OSImageVersion`.
The property can be used for creating an InfraEnv through kube-api with a specific OS image version.

The `OSImageVersion` is useful for HyperShift flows which use late binding, hence, the InfraEnv isn't created according to a cluster version (thus, the latest OS image is used as a fallback).

Notes:
* The specified version should refer to an OSImage from the list in AgentServiceConfig, otherwise the condition is set to failure.
* When a ClusterRef is provided, we still use the cluster's version for creating the InfraEnv. Thus, The property should not be provided along with OSImageVersion to avoid a conflict (validated with a WebHook).

## List all the issues related to this PR

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [x] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #2319, opened 2021-08-02
MGMT-7088: Add InfraEnv CRUD API

ravidbro:
# Assisted Pull Request

## Description

This PR adds the initial API (CRUD) for the new InfraEnv resource via the REST API

## List all the issues related to this PR

- [X] New Feature
- [ ] Bug fix
- [X] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [X] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [X] Waiting for CI to do a full test run
- [X] Unit tests
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Assignees

/assign @yevgeny-shnaidman 

/cc @machacekondra 
/cc @ori-amizur 

## Checklist

- [ ] Title and description added to both, commit and PR.
- [ ] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [ ] Reviewers have been listed
- [ ] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- [ ] Are the title and description (in both PR and commit) meaningful and clear?
- [ ] Is there a bug required (and linked) for this change?
- [ ] Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #6808, opened 2024-09-25
ACM-14502: Allow updates to InfraEnv version

carbonin:
Allow updating the InfraEnv OpenshiftVersion in both the REST and kube APIs.

This is useful when after initial install a user wants to add a new node to a cluster, but the cluster has been upgraded between install and the scale out. Previously a user would need to create a new infraenv with identical settings to the old one and a newer version.

With this change users will be able to update the infraenv os image version just like other attributes.

For some background:

The field was added to the InfraEnv CR in https://github.com/openshift/assisted-service/pull/5365 for https://issues.redhat.com//browse/MGMT-14923 and I don't see any discussion there about updates.

Based on the PR that added infraenv update (https://github.com/openshift/assisted-service/pull/2319) the update params were originally created without the version with no discussion I could find.

Based on these it doesn't seem like this was an intentional omission, so it should be safe to add it here.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

Deployed assisted service kube-api, created an infraenv, updated `osImageVersion` several times.
Observed the URL change accordingly each time and an error in the infraenv status when a non-existent version was requested.

Additionally subsystem tests were added to cover this.

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4309,12 +4309,12 @@ func validateProxySettings(httpProxy, httpsProxy, noProxy, ocpVersion *string) e
 	return nil
 }
 
-// validateArchitectureAndVersion validates if architecture specified inside Infraenv matches one
+// validateClusterArchitectureAndVersion validates if architecture specified inside Infraenv matches one
 // specified for the cluster. For single-arch clusters the validation needs to only compare values
 // of the params. For multiarch cluster we want to see if the multiarch release image contains the
 // the architecture specifically requested by the InfraEnv. We don't need to explicitly validate if
 // the OS image exists because if not, this will be detected by the function generating the ISO.
-func validateArchitectureAndVersion(v versions.Handler, c *common.Cluster, cpuArch, ocpVersion string) error {
+func validateClusterArchitectureAndVersion(v versions.Handler, c *common.Cluster, cpuArch, ocpVersion string) error {
 	// For late-binding we don't know the cluster yet
 	if c == nil {
 		return nil
@@ -4752,7 +4752,7 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(
 func (b *bareMetalInventory) validateInfraEnvCreateParams(ctx context.Context, params installer.RegisterInfraEnvParams, cluster *common.Cluster) error {
 	var err error
 
-	if err = validateArchitectureAndVersion(b.versionsHandler, cluster, params.InfraenvCreateParams.CPUArchitecture, params.InfraenvCreateParams.OpenshiftVersion); err != nil {
+	if err = validateClusterArchitectureAndVersion(b.versionsHandler, cluster, params.InfraenvCreateParams.CPUArchitecture, params.InfraenvCreateParams.OpenshiftVersion); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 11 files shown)


subsystem/infra_env_test.go
```diff
@@ -525,4 +527,15 @@ var _ = Describe("Infra_Env", func() {
 		_, err := userBMClient.Installer.UpdateInfraEnv(ctx, updateParams)
 		Expect(err).To(HaveOccurred())
 	})
+
+	It("fails when given invalid version", func() {
+		updateParams := &installer.UpdateInfraEnvParams{
+			InfraEnvID: infraEnvID,
+			InfraEnvUpdateParams: &models.InfraEnvUpdateParams{
+				OpenshiftVersion: swag.String("5.99"),
```
CrystalChun, line 535: Sorry, I might've missed this, but where is the version validated?

carbonin: The check already existed here https://github.com/openshift/assisted-service/blob/0fde766336a2e6854cdfb758583f15ec40380b25/internal/bminventory/inventory.go#L4991-L4993

CrystalChun: Gotcha, thanks for the pointer! 

That seems to check the infraenv's current cpu arch and openshift version, and not the new updated param's openshift version. Is that alright with your changes?

carbonin: Hm I missed that. But this test is passing ... Let me figure it out.

carbonin: Ah okay, it's failing in the ISO generation code here https://github.com/openshift/assisted-service/blob/0fde766336a2e6854cdfb758583f15ec40380b25/internal/bminventory/inventory.go#L5027-L5029 and later here https://github.com/openshift/assisted-service/blob/0fde766336a2e6854cdfb758583f15ec40380b25/internal/versions/osimages.go#L160

That's after we've already updated the database. I'll add a new validation call for this. Thanks for catching this @CrystalChun!

FWIW This is the message I get currently when I update the infraenv to a non-existent version:
```
    message: 'Failed to create image: No OS image for Openshift version 4.22 and architecture
      x86_64: The requested OS image for version (4.22) and CPU architecture (x86_64)
      isn''t specified in OS images list'
```

CrystalChun: Ah, so it fails down the line. That's good to know too!

carbonin: Added some new commits to validate this earlier.

internal/bminventory/inventory.go
```diff
@@ -4988,7 +4998,7 @@ func (b *bareMetalInventory) UpdateInfraEnvInternal(ctx context.Context, params
 				cluster = nil
 			}
 		}
-		if err = validateArchitectureAndVersion(b.versionsHandler, cluster, infraEnv.CPUArchitecture, infraEnv.OpenshiftVersion); err != nil {
+		if err = validateClusterArchitectureAndVersion(b.versionsHandler, cluster, infraEnv.CPUArchitecture, openshiftVersion); err != nil {
```
CrystalChun, line 5001: What if it's a late-binding infraEnv? Would we still want to validate the version with [this function](https://github.com/openshift/assisted-service/pull/6808/files#diff-747a429a0e12307ef65d754c83b6898ce3858b621005e1e7af6e420da1f1657cR4330) or does it not matter if the infraEnv isn't bound to a cluster?

carbonin: I don't think so. IIRC we don't actually require someone to have defined cluster image sets before defining the infraenv. I'd rather leave that validation to the cluster API rather than tie infraenv and cluster concerns more tightly together.

CrystalChun: Right, we don't require a cluster image set for InfraEnv creation. That makes sense. Thanks!

CrystalChun (approved):
/lgtm

PR #6843, opened 2024-09-29
MGMT-18980: BMH finalizer - remove paused annotation

danielerez:
In order to ensure BMH deprovision by Ironic, removing the 'paused' annotation (if exists) as part of the finalizer.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -334,12 +334,16 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
 
 // handleBMHFinalizer adds a finalizer to the BMH if it isn't being deleted and the finalizer isn't already present
 // If the BMH is being deleted, the associated node is drained and the agent (if any) is annotated and deleted before the BMH is allowed to be removed
-// Finalizer handling is only run if the user has annotated the BMH and converged flow is enabled
+// Finalizer handling is only run if the user has annotated the BMH and converged flow is enabled,
+// or, when the BMH is annotated with 'paused' annotation.
 func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent) reconcileResult {
-	if _, has_annotation := bmh.GetAnnotations()[BMH_DELETE_ANNOTATION]; !has_annotation || !r.ConvergedFlowEnabled {
-		return reconcileComplete{}
+	var removeAgentOnDelete bool
+	if _, has_annotation := bmh.GetAnnotations()[BMH_DELETE_ANNOTATION]; has_annotation && r.ConvergedFlowEnabled {
+		removeAgentOnDelete = true
 	}
+	_, bmhPaused := bmh.GetAnnotations()[BMH_PAUSED_ANNOTATION]
 
+	// Register a finalizer if it's absent
 	bmhHasFinalizer := funk.ContainsString(bmh.GetFinalizers(), BMH_FINALIZER_NAME)
 	if bmh.ObjectMeta.DeletionTimestamp.IsZero() {
 		if !bmhHasFinalizer {
@@ -351,7 +355,11 @@ func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.Fiel
 
 	// BMH is being deleted and finalizer is gone, ensure detached and paused are gone and return
 	if !bmhHasFinalizer {
-		dirty := removeBMHDetachedAndPausedAnnotations(log, bmh)
+		if removeAgentOnDelete {
+			removeBMHDetachedAnnotation(log, bmh)
+		}
+		removeBMHPausedAnnotation(log, bmh)
+		dirty := removeAgentOnDelete || bmhPaused
 		return reconcileComplete{stop: true, dirty: dirty}
 	}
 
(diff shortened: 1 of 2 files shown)


internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -334,9 +334,20 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
...
+	if _, has_annotation := bmh.GetAnnotations()[BMH_DELETE_ANNOTATION]; has_annotation && r.ConvergedFlowEnabled {
+		removeAgentOnDelete = true
+	}
+
+	var bmhPaused bool
+	if _, has_annotation := bmh.GetAnnotations()[BMH_PAUSED_ANNOTATION]; has_annotation {
+		bmhPaused = true
+	}
+
+	if !removeAgentOnDelete && !bmhPaused {
```
CrystalChun, line 350: How come we don't set a finalizer if it's missing both the remove agent annotation and the paused annotation? :o

carbonin: The two situations where we have work to do when the BMH is deleted are:
1. The user set the remove agent on delete annotation, so we need to deprovision then remove the agent.
2. The BMH is paused, so we need to unpause the BMH to allow BMO to finalize it.

If neither is set there's no work for us to do so I don't think we need a finalizer.

That all said, might indeed be safer to just set it unconditionally then make the cleanup work conditional on what the resource looks like when it's deleted.

I mention that because we're not removing the finalizer if the BMH used to be paused but it now (when it is deleted) not, right? I could see that becoming a bug. And this is probably not the only problem. Surely there are more corner cases like this that I haven't thought of.

CrystalChun: Ah I see that we don't need to do work if there's no pause/delete annotation, thanks for the explanation! I agree that we should set it unconditionally too since there could be other cases like what you've mentioned.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -334,9 +334,20 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
...
-	if _, has_annotation := bmh.GetAnnotations()[BMH_DELETE_ANNOTATION]; !has_annotation || !r.ConvergedFlowEnabled {
+	var removeAgentOnDelete bool
+	if _, has_annotation := bmh.GetAnnotations()[BMH_DELETE_ANNOTATION]; has_annotation && r.ConvergedFlowEnabled {
+		removeAgentOnDelete = true
+	}
+
+	var bmhPaused bool
+	if _, has_annotation := bmh.GetAnnotations()[BMH_PAUSED_ANNOTATION]; has_annotation {
+		bmhPaused = true
+	}
```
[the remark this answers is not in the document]
danielerez: Thanks, indeed looks better:)

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -351,40 +362,54 @@ func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.Fiel
...
 	// BMH is being deleted and finalizer is gone, ensure detached and paused are gone and return
 	if !bmhHasFinalizer {
-		dirty := removeBMHDetachedAndPausedAnnotations(log, bmh)
+		var dirty bool
+		if removeAgentOnDelete {
+			dirty = removeBMHDetachedAnnotation(log, bmh)
+		}
+		if bmhPaused {
+			dirty = removeBMHPausedAnnotation(log, bmh) || dirty
+		}
```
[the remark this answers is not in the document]
danielerez: Yeah, I can simplify it.

internal/controller/controllers/bmh_agent_controller.go
danielerez: Good points, will set in unconditionally then.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -351,40 +355,54 @@ func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.Fiel
 
 	// BMH is being deleted and finalizer is gone, ensure detached and paused are gone and return
 	if !bmhHasFinalizer {
-		dirty := removeBMHDetachedAndPausedAnnotations(log, bmh)
+		if removeAgentOnDelete {
+			removeBMHDetachedAnnotation(log, bmh)
+		}
+		if bmhPaused {
```
carbonin, line 361: I still don't think you need to be checking this at all.
`removeBMHPausedAnnotation` is already a no-op if there's no annotation present.

internal/controller/controllers/bmh_agent_controller.go
```diff
@@ -351,40 +355,54 @@ func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.Fiel
...
 	log.Info("Handling BMH delete")
 	// agent could be nil here if it wasn't created yet, or if we deleted it, then failed to remove the finalizer for some reason
 	// if the agent is missing, just allow the BMH to be removed
 	if agent == nil {
-		removeBMHDetachedAndPausedAnnotations(log, bmh)
+		var dirty bool
+		if removeAgentOnDelete {
+			dirty = removeBMHDetachedAnnotation(log, bmh)
+		}
+		if bmhPaused {
```
carbonin, line 376: Same here, no real need to check this condition

carbonin (approved):

PR #6846, opened 2024-09-30
MGMT-18863: Fix bug in host event list endpoint

paul-maidment:
There is a scenario in which a user will use the same host for multiple installation attempts, in this scenario the cluster, infra env and original host will be deleted each time an installation is attempted.

When a new cluster is created and the host is added, it will have the same ID as the previously installed host. The events will contain entries for all previous incarnations of that host.

The problem occurs when the user attempts to fetch "host bound" events for that host, providing the host ID exclusively.The existing implementation of the events endpoint will return all events for a given host ID.

This creates an excessively large number of results, many of which are invalid.

The solution implemented in this PR
is to ensure that when fetching host exclusive events we condsider only events for which there is an infra env and a host, this will ensure that the events are for the "current" version of that host.

Additionally, some changes have been made to the way in which we fetch events from the ORM (gorm), the previous implementation focussed on binding to models of Hosts, Clusters or InfraEnvs, when in fact the core fields we use are event related.The queries have been rephrased to emphasize this "event centric" nature and to hopefully make the queries clearer to understand.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/events/event.go
+++ b/internal/events/event.go
@@ -355,14 +355,13 @@ func (e Events) prepareEventsTable(ctx context.Context, tx *gorm.DB, clusterID *
 	//host bound to a cluster or registered to a bound infra-env) check the access permission
 	//relative to the cluster ownership
 	if clusterBoundEvents() {
-		tx = tx.Model(&common.Cluster{}).
-			Select("events.*, clusters.user_name, clusters.org_id").
-			Joins("INNER JOIN events ON clusters.id = events.cluster_id")
+		tx = tx.Model(&common.Event{}).Select("events.*, clusters.user_name, clusters.org_id").
+			Joins("INNER JOIN clusters ON clusters.id = events.cluster_id")
 
 		// if deleted hosts flag is true, we need to add 'deleted_at' to know whether events are related to a deleted host
 		if swag.BoolValue(deletedHosts) {
 			tx = tx.Select("events.*, clusters.user_name, clusters.org_id, hosts.deleted_at").
-				Joins("LEFT JOIN hosts ON events.host_id = hosts.id")
+				Joins("LEFT JOIN hosts ON hosts.id = events.host_id")
 		}
 		return tx
 	}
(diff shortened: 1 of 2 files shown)


internal/events/event.go
```diff
@@ -355,32 +355,30 @@ func (e Events) prepareEventsTable(ctx context.Context, tx *gorm.DB, clusterID *
...
-	//host table does not contain an org_id we can not perform a join on that table and has to go
-	//through the infra-env table which is good because authorization is done on the infra-level
+	// Events must be linked to the infra_envs table and then to the hosts table
+	// The hosts table does not hold an org_id, so permissions related fields must be supplied by the infra_env
 	if hostOnlyEvents() {
-		return tx.Model(&common.Host{}).Select("events.*, infra_envs.user_name, infra_envs.org_id").
-			Joins("INNER JOIN infra_envs ON hosts.infra_env_id = infra_envs.id").Joins("INNER JOIN events ON events.host_id = hosts.id")
+		return tx.Model(&common.Event{}).Select("events.*, infra_envs.user_name, infra_envs.org_id").
+			Joins("INNER JOIN infra_envs ON infra_envs.id = events.infra_env_id").
+			Joins("INNER JOIN hosts ON hosts.id = events.host_id")
```
rccrdpccl, line 381: why are we joining on the host table? As we are not selecting, the only effect we would have is not returning any event where `events.host_id` has no match in the `hosts` table. Am I missing something here?

paul-maidment: The event must have a match in the hosts table + a match in the infra envs table. This is the purpose of this join. We don't need fields from the host for this as you can see in the `Select` nothing has changed.

I think this is reasonable, the other option is to do a `WHERE events.host_id IN ...` type of subquery but I don't think this provides any benefit. Arguably we return less data I guess?

WDYT do you think we should change this to check for an existing host by using a subquery?

rccrdpccl: I think it's good as it is, maybe let's make the comment slightly more explicit ("ensuring host is there" or something along those lines, rather than "linking to host table")

paul-maidment: Good point.

I added a comment `// This join is here to ensure that only events for a host that exists are fetched` to make the intent of the join without select a bit clearer.

internal/events/event.go
```diff
@@ -355,32 +355,31 @@ func (e Events) prepareEventsTable(ctx context.Context, tx *gorm.DB, clusterID *
...
-	//through the infra-env table which is good because authorization is done on the infra-level
+	// Events must be linked to the infra_envs table and then to the hosts table
+	// The hosts table does not hold an org_id, so permissions related fields must be supplied by the infra_env
 	if hostOnlyEvents() {
-		return tx.Model(&common.Host{}).Select("events.*, infra_envs.user_name, infra_envs.org_id").
-			Joins("INNER JOIN infra_envs ON hosts.infra_env_id = infra_envs.id").Joins("INNER JOIN events ON events.host_id = hosts.id")
+		return tx.Model(&common.Event{}).Select("events.*, infra_envs.user_name, infra_envs.org_id").
+			Joins("INNER JOIN infra_envs ON infra_envs.id = events.infra_env_id").
+			Joins("INNER JOIN hosts ON hosts.id = events.host_id"). // This join is here to ensure that only events for a host that exists are fetched
+			Where("hosts.deleted_at IS NULL")                       // Only interested in active hosts
```
rccrdpccl, line 382: Is this a new behaviour? Would we be able to see deleted host right now?

paul-maidment: tx.Model usually takes care of fields such as "deleted at" because we are dealing with entities, however, the joins themselves, are SQL and the only entity in this statement is Event. So the only filtering applied will be "events that are not deleted".

So it's not new behaviour but it is something we need to take care of when changing the model from `Host` to `Event`.

PR #6839, opened 2024-09-29
NO-ISSUE: Remove failing NMState kube api test until it is fixed

danmanor:
Due to a temporary issue, the kube-api subsystem tests passed falsely. Thus, masking the failure of `deploy clusterDeployment and infraEnv and with an invalid NMState config YAML`  in https://github.com/openshift/assisted-service/pull/6695. See [job](https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-ci-openshift-assisted-service-master-edge-subsystem-kubeapi-aws-periodic). To avoid blocking PRs, I am temporarily removing this test until it is fixed.

A follow up issue - https://issues.redhat.com/browse/MGMT-19027

## List all the issues related to this PR

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [x] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #6847, opened 2024-09-30
MGMT-19027: Fix failing kubeapi subsystem test invalid NMstate YAML

linoyaslan:
This PR addresses the failing kubeapi test, follow-up [this](https://github.com/openshift/assisted-service/pull/6839) PR.

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/staticnetworkconfig/generator.go
+++ b/pkg/staticnetworkconfig/generator.go
@@ -71,8 +71,15 @@ func (s *StaticNetworkConfigGenerator) injectIPV4V6FieldsIfNeeded(networkYaml st
 		s.log.WithError(err).Errorf("Error unmarshalling yaml")
 		return "", err
 	}
-	Interfaces := config["interfaces"].([]interface{})
-	for _, iface := range Interfaces {
+	interfaces, exists := config["interfaces"]
+	if !exists || interfaces == nil {
+		return "", nil
+	}
+	interfacesSlice, ok := interfaces.([]interface{})
+	if !ok {
+		return "", nil
+	}
+	for _, iface := range interfacesSlice {
 		nic := iface.(map[interface{}]interface{})
 		if val, exists := nic["ipv4"].(map[interface{}]interface{}); exists {
 			if _, exists := val["dhcp"]; !exists {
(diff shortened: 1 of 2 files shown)


pkg/staticnetworkconfig/generator.go
```diff
@@ -71,8 +71,15 @@ func (s *StaticNetworkConfigGenerator) injectIPV4V6FieldsIfNeeded(networkYaml st
...
 	}
-	Interfaces := config["interfaces"].([]interface{})
-	for _, iface := range Interfaces {
+	interfaces, exists := config["interfaces"]
+	if !exists || interfaces == nil {
+		return "", nil
+	}
+	interfacesSlice, ok := interfaces.([]interface{})
+	if !ok {
+		return "", nil
```
danielerez, line 80: don't we want to return an error here?

pkg/staticnetworkconfig/generator.go
```diff
@@ -325,10 +332,18 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistenceYAML(macIn
...
 	interfaceWithmacIdentifier := make(map[string]struct{})
 
-	for _, iface := range Interfaces {
+	interfaces, exists := config["interfaces"]
+	if !exists || interfaces == nil {
+		return nil
+	}
+	interfacesSlice, ok := interfaces.([]interface{})
+	if !ok {
+		return nil
```
danielerez, line 343: worth returning a new error here?

pkg/staticnetworkconfig/generator.go
```diff
@@ -340,8 +355,7 @@ func (s *StaticNetworkConfigGenerator) validateInterfaceNamesExistenceYAML(macIn
 			interfaceWithmacIdentifier[interfaceName.(string)] = struct{}{}
 		}
 	}
-
-	for _, iface := range Interfaces {
+	for _, iface := range interfacesSlice {
```
danielerez, line 358: not directly related to this PR, but why do we need to loop over the interfaces twice?

pkg/staticnetworkconfig/generator.go
linoyaslan: The generator should be the authority—if it passes through the generator, it's not our responsibility to determine whether the YAML is valid or not.

pkg/staticnetworkconfig/generator.go
linoyaslan: Because of that:
`if _, exists := interfaceWithMacIdentifier[port.(string)]; exists { continue }`

When iterating over the interfaces, and there is a bond, we want to check if its ports are already listed in the YAML as interfaces with a MAC identifier. If they are, we don't need to include them in the MAC map, and we can skip them.

danielerez: Still I suppose we could simplify the logic with a single pass:/ But it's an unrelated refactor to this PR any way.

PR #6865, opened 2024-10-04
MGMT-18913: Add support for external load balancer

jhernand:
This patch adds support for using an external load balancer. It will only be available for OpenShift 4.16 or newer. To enable it the user will have to use the new `load_balancer.type` field of the `cluster` type, for example:

```json
{
  ...
  "load_balancer": {
    "type": "user-managed"
  }
}
```

To preserve backwards compatibility this will be optional, and the
default value will be `cluster-managed`.

When the value is `user-managed` the following will be added to the
generated `install-config.yaml` file:

```yaml
platform:
  baremetal:
    loadBalancer:
      type: UserManaged
```

The database will be automatically migrated so that for existing
clusters the value of the `load_balancer_type` column will automatically
be set to `cluster-managed`.
    
This patch adds this support only for the `baremetal` and `vsphere`
platforms.
    
Note that when using an external load balancer the VIPs will not be
used, but we are still requiring them, and using them to calculate the
machine CIDR. Insted we should validate that no VIPs are given, and we
should calculate the machine CIDR in some other way. That is not part of
this patch.

https://issues.redhat.com/browse/MGMT-18913

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


jhernand:
Replaced by #7096.

PR #6917, opened 2024-10-15
MGMT-19080, MGMT-18590: Enable installation of 3-5 control plane cluster in day1

danmanor:
# Introduction

currently assisted-service enables only SNO or 3CP clusters in day1. With the new feature in OCP 4.18 enabling installation of 4/5 control planes as well, this PR enables it in assisted-service.

More info - https://issues.redhat.com/browse/OCPSTRAT-1199

### Cluster Validations

currently assisted-service validates that the amount of potential control planes is exactly 3 in HA clusters, we change it to 3-5 and the validations message as well.

### Requirements To Move To Finalizing state

Currently assisted service moves to `finalizing` stage if one of the masters failed to install but you still have at least 3. We change it to validate that all expected masters installed.

### Masters Schedulability

Currently assisted-service decides whether masters will be schedulable or not by checking if their are at least 5 hosts in the cluster (which means at least 2 workers). We change it to at least 2 workers, and refresh again after all hosts are assigned before installation.

### Infrastructure Topology

Currently assisted patches Infrastructure CR to `HighlyAvailable` in clusters with 4 hosts (1 worker). We change it to 1 worker.

### Operator Validations

- ODF: We limit ODF to 3 master cluster for now until it will be clear how will it work on stretched clusters.
- LVM: We change the way we decide whether the cluster's masters will be schedulable or not.

### AgentClusterInstall ProvisionRequirements

Currently, ACI controller uses `ProvisionRequirements` for two purposes - 

- Setting HA / None availability in the cluster (if `ControlPlaneAgents` == 1 and `WorkerAgents` == 0)
- Validating that the number of approved agents equals `ControlPlaneAgents` + `WorkerAgents`, relying on the fact that there is exactly 3 masters. This will no longer be valid as the user can specify for example 4 masters and 1 worker. We change the controller to reconcile to the desired state by host roles. 

### Cross Flow Role assignment

Currently, each flow has its own way of setting manual host roles:

- *ABI* - Host roles are set via host config files by the client.
- *KubeAPI* - Host roles are set via ACI spec.
- *RestAPI* - Host roles are set via update request.

The default role assignment in each of the flows above will change to rely on host monitoring, which will reconcile and designate host roles for each cluster. This process will be based on a new field in the database cluster structure called `ControlPlaneCount`, which represents the desired number of masters in the cluster. This solution will enable kubeAPI and `ABI` to create or update clusters with provisioning requirements, while the RestAPI will support more seamless, optional automatic assignment.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/host/host.go
```diff
@@ -1264,7 +1264,7 @@ func (m *Manager) selectRole(ctx context.Context, h *models.Host, db *gorm.DB) (
 		return autoSelectedRole, err
 	}
 
-	if len(masters) < common.MinMasterHostsNeededForInstallation {
+	if len(masters) < common.MinMasterHostsNeededForInstallationInHaMode {
```
rwsu, line 1267: I've started to test this change for [AGENT-989](https://issues.redhat.com/browse/AGENT-989) and the role assignment appeared to not work correctly for ABI. If I set the control plane replicas to 5 (in ABI would be in agent-cluster-install). 3 nodes would be set to master, and remaining 2 nodes would be assigned workers. 

I think the issue is the number of control plane replicas isn't flowing from from [AgentClusterInstall to CreateClusterParams](https://github.com/openshift/assisted-service/blob/818d92f978db0f0bbfa8e911affdaf31c02dd177/cmd/agentbasedinstaller/register.go#L90).

Rather it is used to [determine HighAvailabilityMode](https://github.com/openshift/assisted-service/blob/818d92f978db0f0bbfa8e911affdaf31c02dd177/internal/controller/controllers/clusterdeployments_controller.go#L1318). There isn't a field in CreateClusterParams to set the number control plane replicas. 

In order to install a cluster with 5 control plane nodes, should ABI be setting the hosts roles to master before starting cluster installation?

danmanor: Commented on https://redhat-internal.slack.com/archives/C02SPBZ4GPR/p1730014829899899?thread_ts=1729626990.388869&cid=C02SPBZ4GPR

swagger.yaml
```diff
@@ -6853,7 +6853,7 @@ definitions:
       - 'ingress-vips-defined'
       - 'ingress-vips-valid'
       - 'all-hosts-are-ready-to-install'
-      - 'sufficient-masters-count'
+      - 'sufficient-masters-and-workers-count'
```
gamli75, line 6856: I'm not sure it's a good idea to change it - it can break the UI and also partners that are using the API
@avishayt WDYT?

swagger.yaml
```diff
@@ -6865,6 +6865,7 @@ definitions:
       - 'mce-requirements-satisfied'
       - 'network-type-valid'
       - 'platform-requirements-satisfied'
+      - 'cluster-preparation-succeeded'
```
gamli75, line 6868: what is the meaning of this new validation? in which wizard step the UI will display it?

internal/host/host.go
rwsu: Hi @danmanor. I tested your latest updates using ABI and setting 5 controlplane agents in AgentClusterInstall still result in 3 masters and 2 workers. I think CreateClusterParams needs to be updated, please see my comment here: https://redhat-internal.slack.com/archives/C02SPBZ4GPR/p1730231838967219?thread_ts=1729626990.388869&cid=C02SPBZ4GPR

danmanor: Thanks for the update Richard. It shouldn't relate to `CreateClusterParams` as it is used for cluster registration but not for host registration. I will investigate further and update

swagger.yaml
danmanor: This validations are only exposed to the user via `validations` field in the cluster object. The UI pops a message if one of the cluster validations is failing, regardless of its name. Therefore, it shouldn't break the UI (not breaking from my tests as well). It could break QE / UI tests, but this could be addressed in advance. I am not sure we want to keep using this name as it is confusing. We can probably keep the old one as well just for back compatibility

gamli75: the UI has a mapping between this value and the string that is being displayed to the user.
I think it's better to keep the old one and add a new one - I also recommended to change it to sufficient-hosts-count because we will have more host roles in the near future.

danmanor: Ack. I will put the old name back

swagger.yaml
danmanor: I will remove this one as well, thanks

internal/cluster/validator.go
```diff
@@ -305,49 +305,52 @@ func (v *clusterValidator) areIngressVipsValid(c *clusterPreprocessContext) (Val
...
 	status := ValidationSuccess
-	var message string
+	var (
+		message                         string
+		minMastersNeededForInstallation int
+		maxMastersNeededForInstallation int
+		mastersRange                    bool
+	)
+
+	isGreaterThan, err := common.BaseVersionGreaterOrEqual(common.MinimumVersionForStretchedControlPlanesCluster, c.cluster.OpenshiftVersion)
```
CrystalChun, line 321: nit: `isGreaterThan` is a bit ambiguous, could we name it something related to the feature about stretch clusters?

internal/cluster/validator.go
```diff
@@ -305,49 +305,52 @@ func (v *clusterValidator) areIngressVipsValid(c *clusterPreprocessContext) (Val
...
 }
 
-// conditions to have a valid number of masters
-// 1. have exactly three masters
-// 2. have less then 3 masters but enough to auto-assign hosts that can become masters
-// 3. have at least 2 workers or auto-assign hosts that can become workers, if workers configured
-// 4. having more then 3 known masters is illegal
-func (v *clusterValidator) sufficientMastersCount(c *clusterPreprocessContext) (ValidationStatus, string) {
+// SufficientMastersCount validates there there is sufficient amount of hosts to satisfy requirments of both masters and workers.
+// The requirments are -
```
CrystalChun, line 309:
```suggestion
// SufficientMastersCount validates that there is a sufficient amount of hosts to satisfy requirements of both masters and workers.
// The requirements are -
```

api/vendor/github.com/openshift/assisted-service/models/cluster_create_params.go
```diff
@@ -42,6 +42,9 @@ type ClusterCreateParams struct {
 	// Cluster networks that are associated with this cluster.
 	ClusterNetworks []*ClusterNetwork `json:"cluster_networks"`
 
+	// The amount of control planes which should be part of the cluster in high availability 'Full' mode.
+	ControlPlaneCount *int64 `json:"control_plane_count,omitempty"`
```
CrystalChun, line 46: How come we're storing as `int64` instead of just `int`? Just curious because of all the conversions I see like this one https://github.com/openshift/assisted-service/pull/6917/files#diff-40883156aab921707c5a73dfa3b451bbd3accb4c0636d715d0a7768db8904b57R683

api/vendor/github.com/openshift/assisted-service/models/v2_cluster_update_params.go
```diff
@@ -273,6 +282,22 @@ func (m *V2ClusterUpdateParams) validateClusterNetworks(formats strfmt.Registry)
...
+	if err := validate.MinimumInt("control_plane_count", "body", *m.ControlPlaneCount, 3, false); err != nil {
+		return err
+	}
+
+	if err := validate.MaximumInt("control_plane_count", "body", *m.ControlPlaneCount, 5, false); err != nil {
+		return err
+	}
+
+	return nil
+}
```
CrystalChun, line 299: What about if it's a SNO cluster? Couldn't the ControlPlaneCount be 1?

internal/bminventory/inventory.go
```diff
@@ -304,6 +306,35 @@ func (b *bareMetalInventory) updatePullSecret(pullSecret string, log logrus.Fiel
 	return pullSecret, nil
 }
 
+func getDefaultHighAvailabilityAndMasterCountParams(highAvailabilityMode *string, controlPlaneCount *int64) (*string, *int64) {
```
eifrach, line 309: I think this should be a common check we we can use it in other places like operators / feature support etc..

internal/bminventory/inventory.go
```diff
@@ -599,6 +643,11 @@ func (b *bareMetalInventory) RegisterClusterInternal(
 		kubeKey = &types.NamespacedName{}
 	}
 
+	var controlPlaneCount *int64 = params.NewClusterParams.ControlPlaneCount
+	if controlPlaneCount == nil {
+		controlPlaneCount = swag.Int64(common.MinMasterHostsNeededForInstallationInHaMode)
+	}
+
```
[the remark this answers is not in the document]
danmanor: Yes I did double check, it is redundant. thanks

internal/bminventory/inventory.go
```diff
@@ -304,6 +306,35 @@ func (b *bareMetalInventory) updatePullSecret(pullSecret string, log logrus.Fiel
...
+func getDefaultHighAvailabilityAndMasterCountParams(highAvailabilityMode *string, controlPlaneCount *int64) (*string, *int64) {
+	// Both not set, multi node by default
+	if highAvailabilityMode == nil && controlPlaneCount == nil {
+		return swag.String(models.ClusterCreateParamsHighAvailabilityModeFull),
+			swag.Int64(common.MinMasterHostsNeededForInstallationInHaMode)
+	}
+
+	// only highAvailabilityMode set
+	if controlPlaneCount == nil {
+		if *highAvailabilityMode == models.ClusterHighAvailabilityModeNone {
```
eifrach, line 318: here you are checking if the Masters are schedulable 
I think you should use / modify the current function for it instead of re writing something new - also this need to be applied to it as well casue we do use it in a lot of places
https://github.com/openshift/assisted-service/blob/master/internal/common/common.go#L193

internal/cluster/validator.go
danmanor: Thank you

internal/bminventory/inventory.go
```diff
@@ -1911,9 +1971,80 @@ func (b *bareMetalInventory) validateAndUpdateClusterParams(ctx context.Context,
...
+		}
+	}
+
 	return *params, nil
 }
 
+func validateHighAvailabilityWithControlPlaneCount(highAvailabilityMode string, controlPlaneCount int64, openshiftVersion string) error {
+	validControlPlaceCountValues := []int64{1, 3, 4, 5}
+	// Updating to invalid control plane count value
+	if !lo.Contains(validControlPlaceCountValues, controlPlaneCount) {
```
eifrach, line 1990: can we use builtin function instead  ?
```suggestion
	if !slices.Contains(validControlPlaceCountValues, controlPlaneCount) {
```

internal/operators/odf/odf_operator.go
```diff
@@ -129,6 +129,13 @@ func (o *operator) getValidDiskCount(disks []*models.Disk, installationDiskID st
 
 // ValidateHost verifies whether this operator is valid for given host
 func (o *operator) ValidateHost(_ context.Context, cluster *common.Cluster, host *models.Host, additionalOperatorRequirements *models.ClusterHostRequirementsDetails) (api.ValidationResult, error) {
+	// temporary disabling ODF for stretched clusters until it will be clear how ODF will work in this scenario.
+	// We pass host validation to avoid false validation messages. The cluster validation will fail
+	masters, _, _ := common.GetHostsByEachRole(&cluster.Cluster, true)
+	if masterCount := len(masters); masterCount > 3 {
```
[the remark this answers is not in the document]
danmanor: No, we want to validate the host when there are 3 masters

internal/ignition/installmanifests.go
```diff
@@ -433,14 +433,20 @@ func (g *installerGenerator) applyManifestPatches(ctx context.Context) error {
...
-	// and the three masters are schedulable.
-	if len(g.cluster.Hosts) != 4 {
-		log.Debugf("number of hosts is different than 4, no need to patch the Infrastructure CR %d", len(g.cluster.Hosts))
+	// hosts roles are known at this stage
+	_, workers, _ := common.GetHostsByEachRole(&g.cluster.Cluster, false)
+	// Patch the InfrastructureCR only if there is exactly one worker.
+	// For multiple workers, the OpenShift installer will handle assigning 'infrastructureTopology: HighlyAvailable'.
+	// When there are no workers, the control plane nodes also act as workers, so 'infrastructureTopology: HighlyAvailable' is set automatically.
+	// Explicitly set 'infrastructureTopology: HighlyAvailable' for a single-worker setup, as specified in
+	// https://github.com/openshift/assisted-service/blob/master/docs/enhancements/4-nodes-cluster-deployment.md
```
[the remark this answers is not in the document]
danmanor: AFAIU yes, at least this is the behaviour we had until now

internal/common/validations.go
```diff
@@ -5,15 +5,28 @@ package common
...
+	string(models.HostValidationIDHostnameUnique),
+	string(models.HostValidationIDHostnameValid),
+}
+var NonIgnorableClusterValidations []string = []string{
+	string(models.ClusterValidationIDAPIVipsDefined),
+	string(models.ClusterValidationIDIngressVipsDefined),
+	string(models.ClusterValidationIDAllHostsAreReadyToInstall),
+	string(models.ClusterValidationIDSufficientMastersCount),
+	string(models.ClusterValidationIDPullSecretSet),
+}
```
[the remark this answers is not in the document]
danmanor: Forgotten, nice catch !

api/vendor/github.com/openshift/assisted-service/models/cluster_create_params.go
danmanor: This is generated by swagger as int64, so I thought it would result in less conversions. If you think we should change it, LMK

api/vendor/github.com/openshift/assisted-service/models/v2_cluster_update_params.go
danmanor: Yes this is a mistake I forgot to change it the update params as well, thanks

internal/bminventory/inventory.go
danmanor: Do you mean the validation func ?

internal/bminventory/inventory.go
danmanor: Sure, but we are using `lo` and `funk` everywhere, so I guess it won't matter much

internal/common/db.go
```diff
@@ -81,6 +82,9 @@ type Cluster struct {
 
 	// A JSON blob in which cluster UI settings will be stored.
 	UISettings string `json:"ui_settings"`
+
+	// The amount of control planes which should be part of the cluster in high availability 'Full' mode.
+	ControlPlaneCount int64 `json:"control_plane_count"`
```
danmanor, line 87: @rccrdpccl

internal/host/host.go
```diff
@@ -1229,13 +1229,13 @@ func (m *Manager) AutoAssignRole(ctx context.Context, h *models.Host, db *gorm.D
...
-// This function recommends a role for a given host based on these criteria:
-//  1. if there are not enough masters and the host has enough capabilities to be
+// selectRole recommends a role for a given host based on these criteria:
+//   - if there are not enough masters and the host has enough capabilities to be
 //     a master the function select it to be a master
-//  2. if there are enough masters, or it is a day2 host, or it has not enough capabilities
-//     to be a master the function select it to be a  worker
-//  3. in case of missing inventory or an internal error the function returns auto-assign
-func (m *Manager) selectRole(ctx context.Context, h *models.Host, db *gorm.DB) (models.HostRole, error) {
+//   - if there are enough masters, or it is a day2 host, or it does not not have enough capabilities
```
rwsu, line 1235: Just FYI, in the future, it is possible some day2 hosts will be masters.

rwsu (commented):
Looks good from ABI perspective. Can deploy with 4 and 5 control plane replicas. Thanks!

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1929,31 +1973,57 @@ func clusterSpecSynced(cluster *hiveext.AgentClusterInstall, syncErr error) {
...
+		// We count directly from the DB because the hosts are not loaded, and we want to avoid loading all hosts
+		mastersCountPtr, err := clusterAPI.GetHostCountByRole(*cluster.ID, models.HostRoleMaster, true)
+		if err != nil {
+			return err
+		}
+
+		workersCountPtr, err := clusterAPI.GetHostCountByRole(*cluster.ID, models.HostRoleWorker, true)
+		if err != nil {
+			return err
+		}
```
CrystalChun, line 2002: Is it possible to do this logic outside of this function? If it errors here, then we won't see any status updates on the AgentClusterInstall, but I think we should always try to set status updates

api/vendor/github.com/openshift/assisted-service/models/cluster_create_params.go
CrystalChun: Oh ok, that's alright then

internal/ignition/installmanifests.go
```diff
@@ -433,14 +433,20 @@ func (g *installerGenerator) applyManifestPatches(ctx context.Context) error {
...
+	numberOfWorkers := len(workers)
+	if len(workers) != 1 {
+		log.Debugf("There are '%d' workers, no need to patch the Infrastructure CR", numberOfWorkers)
```
CrystalChun, line 445: minor nit
```suggestion
	if len(workers) != 1 {
		log.Debugf("There are '%d' workers, no need to patch the Infrastructure CR", len(workers))
```
or
```suggestion
	numberOfWorkers := len(workers)
	if numberOfWorkers != 1 {
		log.Debugf("There are '%d' workers, no need to patch the Infrastructure CR", numberOfWorkers)
```

internal/cluster/validator.go
```diff
@@ -305,49 +305,30 @@ func (v *clusterValidator) areIngressVipsValid(c *clusterPreprocessContext) (Val
...
-		default:
-			//auto-assign hosts and other types go to the candidate list
-			candidates = append(candidates, host)
-		}
+	var (
+		message string
+	)
+
+	// Might be the case for already existing records (default value). In this case high availablity mode is set, we will get
+	// the corresponding control planens count
```
CrystalChun, line 319:
```suggestion
	// the corresponding control planes count
```

CrystalChun (commented):
Few minor comments and a question, but looks great otherwise!

internal/bminventory/inventory.go
```diff
@@ -1435,9 +1458,11 @@ func (b *bareMetalInventory) InstallSingleDay2HostInternal(ctx context.Context,
 	if h, err = b.getHost(ctx, infraEnvId.String(), hostId.String()); err != nil {
 		return err
 	}
+
 	// auto select host roles if not selected yet.
 	err = b.db.Transaction(func(tx *gorm.DB) error {
-		if _, err = b.hostApi.AutoAssignRole(ctx, &h.Host, tx); err != nil {
+		// no need to specify expected master count for day2 hosts, as their suggested role will always be worker
```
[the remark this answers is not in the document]
danmanor: In day2, assisted doesn't always have a way to know the roles of the other hosts from day1. Therefore, the usage of the new `controlPlaneCount` is problematic. I believe for this scenario it would be best to set the role manually in case of master

internal/host/host.go
danmanor: https://github.com/openshift/assisted-service/pull/6917#discussion_r1839668842

internal/controller/controllers/clusterdeployments_controller.go
danmanor: Makes sense. We need to know the amount of masters / workers in order to show more accurate message, so we need to count during `updateStatus`. We can take it one level up to `updateStatus` (as I see we already have some functions returning errors there), but I am not sure it will help because in case of error we will still not show any status no ? WDYT ?

internal/ignition/installmanifests.go
danmanor: Thanks !

internal/bminventory/inventory.go
```diff
@@ -1911,9 +1936,81 @@ func (b *bareMetalInventory) validateAndUpdateClusterParams(ctx context.Context,
...
+		)
+	}
+
+	// Updating SNO cluster to amount != 1
+	if highAvailabilityMode == models.ClusterCreateParamsHighAvailabilityModeNone &&
+		controlPlaneCount != 1 {
+		return common.NewApiError(
+			http.StatusBadRequest,
+			fmt.Errorf(
+				"there should be exactly %d control plane for high availability mode %s",
```
gamli75, line 1969: I think we should change the error message to something more clear such as:  there should be exactly %d control plane for Single node cluster

internal/bminventory/inventory.go
```diff
@@ -1911,9 +1936,81 @@ func (b *bareMetalInventory) validateAndUpdateClusterParams(ctx context.Context,
...
+		); err != nil {
+			return installer.V2UpdateClusterParams{}, err
+		}
+	}
+
 	return *params, nil
 }
 
+func validateHighAvailabilityWithControlPlaneCount(highAvailabilityMode string, controlPlaneCount int64, openshiftVersion string) error {
+	validControlPlaceCountValues := []int64{1, 3, 4, 5}
```
gamli75, line 1954: why do we need this validation at the beginning of the function? the user should get informative data and if we passed the other conditions than we should return this error

internal/bminventory/inventory.go
```diff
@@ -1911,9 +1936,81 @@ func (b *bareMetalInventory) validateAndUpdateClusterParams(ctx context.Context,
...
+	// Updating to invalid control plane count value
+	if !slices.Contains(validControlPlaceCountValues, controlPlaneCount) {
+		return common.NewApiError(
+			http.StatusBadRequest,
+			fmt.Errorf("valid values for control_plane_count are 1, 3, 4 or 5"),
+		)
+	}
+
+	// Updating SNO cluster to amount != 1
```
gamli75, line 1963: small comment - we are not updating anything here 
```suggestion
	// validating the number of control planes for SNO 
```

internal/bminventory/inventory.go
danmanor: You are right, I merged it to the others

internal/controller/controllers/clusterdeployments_controller.go
CrystalChun: I think taking it up one level up is a good idea. If there's an error, could we log it but still also update status?

danmanor: you think it would be better to update status even though the other functions one level up return 
```
```
?

I can maybe show this error as the status message in case of error, WDYT ?
I mean in the case of error -

```
condStatus = corev1.ConditionFalse
reason = hiveext.ClusterInsufficientAgentsReason
msg = "failed to count the amount of control plane agents and worker agents bound to the cluster"
```

CrystalChun: Yeah showing an error message would be better! I just don't think it should fail setting the status if the function GetHostCountByRole errors out

danmanor: Ack.

danmanor: I implemented the changes, Let me know what you think.

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -655,9 +694,23 @@ func (r *ClusterDeploymentsReconciler) isReadyForInstallation(ctx context.Contex
...
-	expectedHosts := clusterInstall.Spec.ProvisionRequirements.ControlPlaneAgents +
-		clusterInstall.Spec.ProvisionRequirements.WorkerAgents
-	return approvedHosts == expectedHosts && registered == approvedHosts && unsyncedHosts == 0, nil
+
+	expectedMasterCount := clusterInstall.Spec.ProvisionRequirements.ControlPlaneAgents
+	expectedWorkerCount := clusterInstall.Spec.ProvisionRequirements.WorkerAgents
+	expectedHostCount := expectedMasterCount + expectedWorkerCount
+
+	masterCountPrt, workerCountPtr, err := getHostSuggestedRoleCount(r.ClusterApi, *c.ID)
+	if err != nil {
```
CrystalChun, line 703:
```suggestion
	masterCountPtr, workerCountPtr, err := getHostSuggestedRoleCount(r.ClusterApi, *c.ID)
	if err != nil {
```

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -655,9 +694,23 @@ func (r *ClusterDeploymentsReconciler) isReadyForInstallation(ctx context.Contex
...
+		// will be shown as a SpecSynced error
+		log.WithError(err).Error("failed to fetch host suggested role count")
+		return false, err
+	}
+
+	return approvedHosts == expectedHostCount &&
+		registered == expectedHostCount &&
+		int(swag.Int64Value(masterCountPrt)) == expectedMasterCount &&
+		int(swag.Int64Value(workerCountPtr)) == expectedWorkerCount &&
```
CrystalChun, line 712: If `workerCountPtr` or `masterCountPrt` are nil, will these cause any issues?

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1766,31 +1831,47 @@ func (r *ClusterDeploymentsReconciler) updateStatus(ctx context.Context, log log
...
 			}
-			clusterRequirementsMet(clusterInstall, status, registeredHosts, approvedHosts, unsyncedHosts)
+
+			clusterRequirementsMet(
+				clusterInstall,
+				status,
+				approvedHosts,
+				unsyncedHosts,
+				int(swag.Int64Value(mastersCountPtr)),
+				int(swag.Int64Value(workersCountPtr)),
```
CrystalChun, line 1873: Will these cause any issues if `masterCountPtr` or `workersCountPtr` is nil?

danmanor: If these are nil then they will be 0, will probably show something like - `"The cluster currently requires exactly 3 master agents and 2 worker agents, but currently registered 0 master agents and 0 worker agents"`. I son't see any other option if we want to update status in this case, as we don't want to show custom message in this case right ?

CrystalChun: Cool just confirming it doesn't cause issues. That sounds good, thanks!

CrystalChun (approved):
/lgtm

PR #6918, opened 2024-10-15
OCPBUGS-42705: Use icsp/idms for oc commands instead of releaseMirror

bfournie:
Currently, the local image mirror url is used when running the 'oc' commands. This can cause a problem if the mirror configuration has multiple mirrors, since there will be no failover if the url is unreachable. Instead, the actual releaseImage should be used when the icsp or idms configuration is provided to ensure proper failover.

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [X] dev-scripts environment
- [] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/installercache/installercache.go
+++ b/internal/installercache/installercache.go
@@ -70,11 +70,7 @@ func (i *Installers) Get(releaseID, releaseIDMirror, pullSecret string, ocReleas
 	var workdir, binary, path string
 	var err error
 
-	releaseImageLocation := releaseID
-	if releaseIDMirror != "" {
-		releaseImageLocation = releaseIDMirror
-	}
-	workdir, binary, path, err = ocRelease.GetReleaseBinaryPath(releaseImageLocation, i.cacheDir, ocpVersion)
+	workdir, binary, path, err = ocRelease.GetReleaseBinaryPath(releaseID, i.cacheDir, ocpVersion)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/oc/release.go
+++ b/internal/oc/release.go
@@ -129,19 +129,15 @@ func (r *release) getImageByName(log logrus.FieldLogger, imageName, releaseImage
 		return "", errors.New("neither releaseImage, nor releaseImageMirror are provided")
 	}
 
-	if releaseImageMirror != "" {
-		//TODO: Get mirror registry certificate from install-config
-		image, err = r.getImageFromRelease(log, imageName, releaseImageMirror, pullSecret, true)
-		if err != nil {
+	//TODO: Get mirror registry certificate from install-config
+	image, err = r.getImageFromRelease(log, imageName, releaseImage, releaseImageMirror, pullSecret)
+	if err != nil {
+		if releaseImageMirror != "" {
 			log.WithError(err).Errorf("failed to get %s image from mirror release image %s", imageName, releaseImageMirror)
-			return "", err
-		}
-	} else {
-		image, err = r.getImageFromRelease(log, imageName, releaseImage, pullSecret, false)
-		if err != nil {
+		} else {
 			log.WithError(err).Errorf("failed to get %s image from release image %s", imageName, releaseImage)
-			return "", err
 		}
+		return "", err
 	}
 	return image, err
 }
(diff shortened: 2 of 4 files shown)


andfasano:
/retest

since sno/ha agent jobs failed

internal/oc/release.go
```diff
@@ -129,19 +129,11 @@ func (r *release) getImageByName(log logrus.FieldLogger, imageName, releaseImage
...
-	} else {
-		image, err = r.getImageFromRelease(log, imageName, releaseImage, pullSecret, false)
-		if err != nil {
-			log.WithError(err).Errorf("failed to get %s image from release image %s", imageName, releaseImage)
-			return "", err
-		}
+	//TODO: Get mirror registry certificate from install-config
+	image, err = r.getImageFromRelease(log, imageName, releaseImage, releaseImageMirror, pullSecret)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get %s image from mirror release image %s or mirror %s", imageName, releaseImage, releaseImageMirror)
```
andfasano, line 135: Wouldn't be a misleading log traces in case of empty `releaseImageMirror` (previous implementation seems to provide a different logging message in that case)

internal/oc/release.go
```diff
@@ -153,19 +145,11 @@ func (r *release) GetOpenshiftVersion(log logrus.FieldLogger, releaseImage strin
...
-	} else {
-		openshiftVersion, err = r.getOpenshiftVersionFromRelease(log, releaseImage, pullSecret, false)
-		if err != nil {
-			log.WithError(err).Errorf("failed to get image openshift version from release image %s", releaseImage)
-			return "", err
-		}
+	//TODO: Get mirror registry certificate from install-config
+	openshiftVersion, err = r.getOpenshiftVersionFromRelease(log, releaseImage, releaseImageMirror, pullSecret)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get image openshift version from release image %s or mirror %s", releaseImage, releaseImageMirror)
```
andfasano, line 151: Also here

internal/oc/release.go
bfournie: Updated log messages to be similar to previous implementation.

internal/installercache/installercache.go
```diff
@@ -70,11 +70,7 @@ func (i *Installers) Get(releaseID, releaseIDMirror, pullSecret string, ocReleas
 	var workdir, binary, path string
 	var err error
 
-	releaseImageLocation := releaseID
-	if releaseIDMirror != "" {
-		releaseImageLocation = releaseIDMirror
-	}
-	workdir, binary, path, err = ocRelease.GetReleaseBinaryPath(releaseImageLocation, i.cacheDir, ocpVersion)
+	workdir, binary, path, err = ocRelease.GetReleaseBinaryPath(releaseID, i.cacheDir, ocpVersion)
```
CrystalChun, line 73: Sorry, I'm not sure I understand this change. 

The path here might not match the path generated in the `Extract` function since the `Extract` function could use the mirror image instead, if it's provided. https://github.com/openshift/assisted-service/pull/6918/files#diff-3b8a020ef087b81b6b45d56bef6dd8d32fcf9d50dfe3d91fd369755b253cfff8R372

How come we're only using the release image instead of the mirror image for the path here?

CrystalChun (commented):
Just a question, but changes look good otherwise! Great idea to bubble up which image is being used.

internal/installercache/installercache.go
bfournie: The mirror image should be in the ICSP (or IDMS) that is provided (if that is not provided then the explicit mirror image will be used https://github.com/openshift/assisted-service/pull/6918/files#diff-3b8a020ef087b81b6b45d56bef6dd8d32fcf9d50dfe3d91fd369755b253cfff8R484  but the intent is to have the mirror image in the ICSP).

The way its currently implemented is that if the ICSP has multiple entries, all of the entries will not be tried if only one value of mirror image is used.

CrystalChun: Sure, but sorry, shouldn't this path calculated here match the one in here? https://github.com/openshift/assisted-service/pull/6918/files#diff-3b8a020ef087b81b6b45d56bef6dd8d32fcf9d50dfe3d91fd369755b253cfff8R372

bfournie: OK, I understand the question now. Yes, the call with GetReleaseBinaryPath() will always pass in the releaseImage, not the mirror. Since this path is only used for the cache location its not necessary that it include the mirror string, only that it is the same whenever it is referenced.

CrystalChun: Ah thanks! I wasn't looking carefully and thought that they were mismatched. Sorry about that!

CrystalChun (approved):
/lgtm

PR #6920, opened 2024-10-16
MGMT-19141: Ensure no finalizer is set for BMHs without the infraenv label

carbonin:
Previously BMAC was setting our finalizer for every BMH in the cluster and only removing them when we either saw the delete agent annotation or the paused annotation. This meant that no other BMHs could be removed without manually removing this finalizer.

This commit ensures the finalizer handling sees the finalizer being removed as a change and sets the "dirty" return flag accordingly and also ensures we never set out finalizer on a BMH without an infraenv label (one we don't own).

Resolves https://issues.redhat.com/browse/MGMT-19141

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Manually tested by deploying to an environment with BMHs for multiple purposes.
Saw that finalizers were no longer being set for BMHs without the infraenv label.

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/bmh_agent_controller.go
+++ b/internal/controller/controllers/bmh_agent_controller.go
@@ -337,21 +337,23 @@ func (r *BMACReconciler) Reconcile(origCtx context.Context, req ctrl.Request) (c
 // Finalizer handling is only run if the user has annotated the BMH and converged flow is enabled,
 // or, when the BMH is annotated with 'paused' annotation.
 func (r *BMACReconciler) handleBMHFinalizer(ctx context.Context, log logrus.FieldLogger, bmh *bmh_v1alpha1.BareMetalHost, agent *aiv1beta1.Agent) reconcileResult {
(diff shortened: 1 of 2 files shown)


carbonin:
/hold

Still doing some testing, but wanted to get this out to get CI running.

carbonin:
I'm going to change this to trigger based on the presence of the infraenv label on the BMH rather than whether or not the infraenv actually exists.

That's mainly because we don't care about the infraenv, we only really want to know if this is a BMH that is meant for us or not.

I mention this because I deployed an image with this change and I ended up with a BMH that references a missing infraenv having no finalizer but also having the paused annotation. This means that it would be impossible to delete regularly.

CrystalChun (approved):
/lgtm

PR #6941, opened 2024-10-23
NO-ISSUE: Fix unit tests that fail locally

jhernand:
This patch fixes several tests that fail when executed locally with a recent version of PostgreSQL, in particular version 16.

- _Test that DHCP VIPs were cleared when switching to UserManagedNetworking: true_.

  This test fails when executed individually because it tries to update
  the VIPs in the database doing `db.Save(&cluster)`, but that only
  saves the cluster itself, and not other referenced tables, like the
  VIPs tables. I believe this works in CI today because of the state of
  the database generated by previous tests. The patch changes it to
  explicitly save the VIPs using `db.Save(&models.APIVip{...})`.

- _single SNO vsphere host_ and other related tests.

  These fail because in recent versions of PostgreSQL it is not the OK to update the database with zero timestamps: the database server rejects the update, but _Gorm_ ignores it and just does nothing. The net result is that nothing is updated. The patch changes those tests to update only the required columns, so there is no problem with zero time stamps.

- _Deregister inactive cluster_.

  This test fails because it uses `time.Now()` as the last activity date, and depending on the speed of the machine the date may change between that call to `time.Now()` and the actual check. The patch changes it to use a data that is a second in the future.

- Operator handler tests.

  These tests fail because they compare the string representation of time stamps. But depending on the machine and database configuration the time zone may be `nil` or different to what is stored in the database, so the string representation may be different even if the time stamp is the same. To avoid that the patch adds a new `EqualTime` Gomega matcher that compares time stamps converting them to the same time stamp.

No related issue.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [X] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Executed unit tests locally.

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory_test.go
+++ b/internal/bminventory/inventory_test.go
@@ -3963,11 +3963,16 @@ var _ = Describe("cluster", func() {
 					Expect(reply).To(BeAssignableToTypeOf(installer.NewV2UpdateClusterCreated()))
 
 					By("Update VIPS in DB cluster to mimic DHCP allocation")
-					cluster, err := common.GetClusterFromDB(db, clusterID, common.SkipEagerLoading)
-					Expect(err).ToNot(HaveOccurred())
-					cluster.APIVips = []*models.APIVip{{IP: models.IP(apiVip)}}
-					cluster.IngressVips = []*models.IngressVip{{IP: models.IP(ingressVip)}}
-					db.Save(&cluster)
+					db.Save(&models.APIVip{
+						ClusterID: clusterID,
+						IP:        models.IP(apiVip),
+					})
+					Expect(db.Error).ToNot(HaveOccurred())
+					db.Save(&models.IngressVip{
+						ClusterID: clusterID,
+						IP:        models.IP(ingressVip),
+					})
+					Expect(db.Error).ToNot(HaveOccurred())
 
 					By("Verify VIPs updates")
 					replay := bm.V2GetCluster(ctx, installer.V2GetClusterParams{ClusterID: clusterID}).(*installer.V2GetClusterOK)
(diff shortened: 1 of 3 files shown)


internal/testing/time.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"time"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/onsi/gomega/types"
+)
+
+// EqualTime creates a matcher that checks that the actual value is equal to the given expected value, even if they are
+// expressed using different time zones. Actual and expected values should be time.Time or strfmt.DateTime.
+func EqualTime(expected any) types.GomegaMatcher {
```
carbonin, line 13: Do we have enough uses of `strfmt.DateTime` to justify not using https://onsi.github.io/gomega/#betemporallycomparator-string-compareto-timetime-threshold-timeduration?

Alternatively does it make sense to wrap the existing matcher with a type conversion?

jhernand: I'd say we have many uses of `strfmt.DateTime` because that is what the OpenAPI code generator uses: all our "date-time" fields will be `strfmt.DateTime`. However in this particular pull request the new marcher is used only three times. I think it is convenient to have this kind of matchers to simplify tests, but it isn't a strong opinion. I will replace it with explicit type conversions and time zone changes.

jhernand: Done.

internal/bminventory/inventory_test.go
```diff
@@ -12511,18 +12522,23 @@ var _ = Describe("GetSupportedPlatformsFromInventory", func() {
...
-		c.OpenshiftVersion = "4.12"
-		db.Save(c)
+		db.Model(&models.Cluster{}).
+			Where("id = ?", clusterID).
+			Update("openshift_version", "4.12")
+		Expect(db.Error).ToNot(HaveOccurred())
 		mockProviderRegistry.EXPECT().GetSupportedProvidersByHosts(gomock.Any()).Return(expectedPlatforms, nil)
 		platformReplay = bm.GetClusterSupportedPlatforms(ctx, installer.GetClusterSupportedPlatformsParams{ClusterID: clusterID})
 		platforms = platformReplay.(*installer.GetClusterSupportedPlatformsOK).Payload
+		fmt.Fprintf(GinkgoWriter, "platforms: %v\n", platforms)
```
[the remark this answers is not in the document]
jhernand: Yes, done.

jhernand:
@CrystalChun , @carbonin this is ready for your re-review.

CrystalChun (commented):
Looks good to me! Will wait for Nick's review

CrystalChun (approved):
/lgtm

PR #6963, opened 2024-11-02
MGMT-19218: Remove getting OKD for late-binding InfraEnv

CrystalChun:
An InfraEnv that's not attached to a cluster will not know what release image to use so it'll be unable to find an OKD image. This removes getting the OKD image unless the InfraEnv is attached to a Cluster.

Mentioned in the original introduction that this flow is not broken https://github.com/openshift/assisted-service/pull/4650/files#r1044872735

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/ignition/discovery.go
+++ b/internal/ignition/discovery.go
@@ -209,6 +209,10 @@ func (ib *ignitionBuilder) shouldAppendOKDFiles(ctx context.Context, infraEnv *c
 	if cfg.OKDRPMsImage != "" {
 		return cfg.OKDRPMsImage, true
 	}
+	// Late-binding InfraEnvs will not have release images associated with them
+	if infraEnv.ClusterID == "" {
+		return "", false
+	}
 	// Check if selected payload contains `okd-rpms` image
 	releaseImage, err := ib.versionHandler.GetReleaseImage(ctx, infraEnv.OpenshiftVersion, infraEnv.CPUArchitecture, infraEnv.PullSecret)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


CrystalChun:
An alternative solution would be to at least removing caching all the cluster image sets in the GetReleaseImage function as that has caused a lot of errors.

gamli75 (approved):

internal/ignition/discovery.go
```diff
@@ -209,6 +209,10 @@ func (ib *ignitionBuilder) shouldAppendOKDFiles(ctx context.Context, infraEnv *c
 	if cfg.OKDRPMsImage != "" {
 		return cfg.OKDRPMsImage, true
 	}
+	// Late-binding InfraEnvs will not have release images associated with them
+	if infraEnv.ClusterID == "" {
+		return "", false
+	}
 	// Check if selected payload contains `okd-rpms` image
 	releaseImage, err := ib.versionHandler.GetReleaseImage(ctx, infraEnv.OpenshiftVersion, infraEnv.CPUArchitecture, infraEnv.PullSecret)
```
carbonin, line 217: This should also probably be changed to use the cluster release versions, right?

CrystalChun: Good question, but wouldn't a bound infra-env's openshift version/arch match the cluster it's bound to or is that an incorrect assumption?

carbonin: I think you're right, but this just looked weird to me.

carbonin: To be clear, now that you mentioned the matching for bound clusters it's probably not worth changing it (especially because it looks like we'd need to plumb the db client into this function somehow).

CrystalChun: Yeah I thought getting the cluster from the db would be excessive, especially given that the version/arch should match. But I do agree that it does look funny.

PR #6868, opened 2024-10-06
MGMT-18561: MGMT-18562:  Add kube api support for adding per cluster mirror registry in AgentClusterInstall

eliorerz:
## List all the issues related to this PR

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested)
- [ ] No tests needed

/cc @eliorerz

PR #6965, opened 2024-11-03
MGMT-18561: MGMT-18562: Add kube api support for adding per cluster mirror registry in AgentClusterInstall

eliorerz:

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

/cc @eliorerz


CrystalChun:
Retested and verified it's pulling from the mirror registry 👍 
```
  Warning  Failed          1s (x2 over 15s)   kubelet            Failed to pull image "quay.io/cchun/testing@sha256:e1f8ac471c4cd63333590309829aa8f1210a62571179d39fc6989ab6a9c09e2e": (Mirrors also failed: [10.1.178.25:5000/cchun/testing@sha256:e1f8ac471c4cd63333590309829aa8f1210a62571179d39fc6989ab6a9c09e2e: pinging container registry 10.1.178.25:5000: Get "https://10.1.178.25:5000/v2/": http: server gave HTTP response to HTTPS client]): quay.io/cchun/testing@sha256:e1f8ac471c4cd63333590309829aa8f1210a62571179d39fc6989ab6a9c09e2e: reading manifest sha256:e1f8ac471c4cd63333590309829aa8f1210a62571179d39fc6989ab6a9c09e2e in quay.io/cchun/testing: unauthorized: access to the requested resource is not authorized
```

internal/bminventory/inventory.go
```diff
@@ -1279,7 +1281,7 @@ func (b *bareMetalInventory) integrateWithAMSClusterPreInstallation(ctx context.
 	return nil
 }
 
-func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params installer.V2InstallClusterParams) (*common.Cluster, error) {
+func (b *bareMetalInventory) InstallClusterInternal(ctx context.Context, params installer.V2InstallClusterParams, mirrorRegistryConfiguration *v1beta1.MirrorRegistryConfiguration) (*common.Cluster, error) {
```
carbonin, line 1284: Looks like this was added and isn't used, can we remove it?

internal/bminventory/inventory.go
```diff
@@ -1670,6 +1672,13 @@ func (b *bareMetalInventory) UpdateClusterInstallConfigInternal(ctx context.Cont
 			return common.NewApiError(http.StatusBadRequest, err)
 		}
 
+		if mirrorregistries.IsMirrorConfigurationSet(mirrorRegistryConfiguration) {
```
carbonin, line 1675: Why do this check?

internal/bminventory/inventory_test.go
```diff
@@ -13188,6 +13178,65 @@ var _ = Describe("TestRegisterCluster", func() {
...
+			mockClusterRegisterSuccess(true)
+			mockAMSSubscription(ctx)
+			conf, _ := getMirrorRegistryConfigurations(getSecureRegistryToml(), mirrorRegistryCertificate)
+			params := getClusterCreateParams()
+			c, err := bm.RegisterClusterInternal(ctx, nil, conf, installer.V2RegisterClusterParams{NewClusterParams: params})
+			Expect(err).ShouldNot(HaveOccurred())
+
+			var clusterObj common.Cluster
+			Expect(db.First(&clusterObj, "id = ?", c.ID).Error).ShouldNot(HaveOccurred())
+			Expect(clusterObj.MirrorRegistryConfiguration).ShouldNot(BeNil())
```
carbonin, line 13236: Is there anything about this content you can verify?

internal/bminventory/inventory_test.go
```diff
@@ -13188,6 +13178,65 @@ var _ = Describe("TestRegisterCluster", func() {
 		}
 	})
 
+	Context("Cluster Mirror Registry", func() {
```
carbonin, line 13181: Can you add a test for the infraenv functions as well as the cluster update one?

internal/bminventory/inventory.go
```diff
@@ -1647,7 +1649,7 @@ func (b *bareMetalInventory) GetArchitecturesSupportLevelListInternal(_ context.
 	return featuresupport.GetCpuArchitectureSupportList(params.OpenshiftVersion), nil
 }
 
-func (b *bareMetalInventory) UpdateClusterInstallConfigInternal(ctx context.Context, params installer.V2UpdateClusterInstallConfigParams) (*common.Cluster, error) {
+func (b *bareMetalInventory) UpdateClusterInstallConfigInternal(ctx context.Context, params installer.V2UpdateClusterInstallConfigParams, mirrorRegistryConfiguration *v1beta1.MirrorRegistryConfiguration) (*common.Cluster, error) {
```
carbonin, line 1652: Why have this as a parameter to install config and also to the cluster itself. I would think it would be better to keep it in one place (just in the cluster).

internal/common/db.go
```diff
@@ -537,3 +589,27 @@ func CloseDB(db *gorm.DB) {
...
+	err := json.Unmarshal(data, &config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal MirrorRegistryConfiguration: %w", err)
+	}
+
+	return &config, nil
+}
+
+// ConvertMirrorRegistryConfigToBytes converts a MirrorRegistryConfiguration struct to a byte slice
+func ConvertMirrorRegistryConfigToBytes(config *v1beta1.MirrorRegistryConfiguration) ([]uint8, error) {
```
carbonin, line 607: Does this need to be exported?

internal/common/db.go
```diff
@@ -191,6 +218,9 @@ type InfraEnv struct {
 	// Json formatted string containing internal overrides for the default ignition config.
 	// This is used for adding ironic ignition config to the assisted ignition config
 	InternalIgnitionConfigOverride string `json:"internal_ignition_config_override,omitempty"`
+
+	// A JSON blob in which holds a mirror registry configurations if set
+	MirrorRegistryConfiguration []uint8 `gorm:"type:BYTEA"`
```
carbonin, line 223: Why byte and not just serialized json?

api/hiveextension/v1beta1/agentclusterinstall_types.go
```diff
@@ -505,3 +512,20 @@ type ManifestsConfigMapReference struct {
...
+// MirrorRegistryConfigMapReference contains reference to a ConfigMap for mirror registry
+type MirrorRegistryConfigMapReference struct {
+	// Name is the name of the ConfigMap that this refers to
+	Name string `json:"name"`
+	// Namespace of the ConfigMap
+	Namespace string `json:"namespace"`
+}
+
+// MirrorRegistryConfiguration holds the given mirror registry configuration
+type MirrorRegistryConfiguration struct {
```
carbonin, line 525: Given that this is an internal representation of the mirror config (and that it's shared between infraenv and cluster) I think it should be somewhere internal rather than in the hiveextension API package.

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -2325,3 +2343,19 @@ func getClusterDeploymentAdminKubeConfigSecretName(cd *hivev1.ClusterDeployment)
 	}
 	return fmt.Sprintf(adminKubeConfigStringTemplate, cd.Name)
 }
+
+// processMirrorRegistryConfig retrieves the mirror registry configuration from the referenced ConfigMap
+func (r *ClusterDeploymentsReconciler) processMirrorRegistryConfig(ctx context.Context, log logrus.FieldLogger, clusterInstall *hiveext.AgentClusterInstall) (*hiveext.MirrorRegistryConfiguration, error) {
+	mirrorRegistryConfiguration, userTomlConfigMap, err := mirrorregistries.ProcessMirrorRegistryConfig(ctx, log, r.Client, clusterInstall.Spec.MirrorRegistryRef)
```
carbonin, line 2349: I'd prefer we read the configmap data in the controllers package here and keep the knowledge of the kube API out of the internal packages.

internal/controller/controllers/clusterdeployments_controller_test.go
```diff
@@ -254,6 +353,293 @@ var _ = Describe("cluster reconcile", func() {
 		mockCtrl.Finish()
 	})
 
+	Context("Mirror Registry", func() {
```
carbonin, line 356: I'd prefer this be something more like `Describe("processMirrorRegistryConfig", func() {`

That way we can separate the tests that are specifically for this function from the ones that are testing `Reconcile` (which I think should probably be with the other tests).

Generally I don't feel like there's more value in grouping tests by feature rather than the function they're testing.

internal/controller/controllers/clusterdeployments_controller_test.go
```diff
@@ -254,6 +353,293 @@ var _ = Describe("cluster reconcile", func() {
...
+
+			return getTestClusterInstall()
+		}
+
+		createUserMirrorRegistryConfigmap := func(registryToml string, certificate string) {
+			registryCM := newUserProvidedRegistryCM(registryToml, certificate)
+			Expect(c.Create(ctx, registryCM)).ShouldNot(HaveOccurred())
+		}
+
+		When("the user-provided valid image registry contains correct data", func() {
```
carbonin, line 373: Nit, but single example contexts with no setup code can just inline the context in the test name. I think it will look better in failed test output.

internal/controller/controllers/clusterdeployments_controller_test.go
```diff
@@ -254,6 +353,293 @@ var _ = Describe("cluster reconcile", func() {
...
+			Expect(c.Create(ctx, cluster)).ShouldNot(HaveOccurred())
+			defaultAgentClusterInstallSpec.MirrorRegistryRef = &hiveext.MirrorRegistryConfigMapReference{Name: providedMirrorRegistryCMName, Namespace: testNamespace}
+
+			aci := newAgentClusterInstall(agentClusterInstallName, testNamespace, defaultAgentClusterInstallSpec, cluster)
+			Expect(c.Create(ctx, aci)).ShouldNot(HaveOccurred())
+
+			return cluster, aci
+		}
+
+		getClusterInstallFailure := func(registryToml, certificate string, createMirrorRegistryCM bool) error {
```
carbonin, line 489: If this is only used in a single context I'd prefer you define it there.

internal/controller/controllers/infraenv_controller_test.go
```diff
@@ -134,6 +135,53 @@ var _ = Describe("infraEnv reconcile", func() {
...
 
+	Context("Mirror registry", func() {
+
+		createUserMirrorRegistryConfigmap := func(registryToml string, certificate string) {
+			registryCM := newUserProvidedRegistryCM(registryToml, certificate)
+			Expect(c.Create(ctx, registryCM)).ShouldNot(HaveOccurred())
+		}
+
+		When("the user-provided valid image registry contains correct data", func() {
+			It("Successfully parse and set the given user configmap", func() {
```
carbonin, line 146: Since this is all just one big test, it can just be one big test with much less indentation.

internal/common/db.go
```diff
@@ -101,6 +107,27 @@ func (c *Cluster) Payload() any {
...
+func (c *Cluster) GetMirrorRegistryConfiguration() (*v1beta1.MirrorRegistryConfiguration, error) {
+	if c.MirrorRegistryConfiguration == nil {
+		return nil, nil
+	}
+	return convertBytesToMirrorRegistryConfig(c.MirrorRegistryConfiguration)
+}
+
+func (c *Cluster) SetMirrorRegistryConfiguration(configuration *v1beta1.MirrorRegistryConfiguration) error {
+	if configuration == nil {
+		return nil
```
carbonin, line 119: This makes it look like there's no way to explicitly remove the mirror configuration. Do we want to deal with that scenario?

internal/ignition/discovery.go
```diff
@@ -222,6 +223,36 @@ func (ib *ignitionBuilder) shouldAppendOKDFiles(ctx context.Context, infraEnv *c
 	return okdRpmsImage, true
 }
 
+func (ib *ignitionBuilder) setIgnitionMirrorRegistry(ignitionParams *map[string]interface{}, configuration *hiveext.MirrorRegistryConfiguration) error {
+	mirrorRegistriesConfigKey := "MirrorRegistriesConfig"
+	mirrorRegistriesCAConfigKey := "MirrorRegistriesCAConfig"
+
+	if mirrorregistries.IsMirrorConfigurationSet(configuration) {
+		ib.log.Infof("Setting ignition cluster mirror registry to %s", configuration.RegistriesConf)
```
carbonin, line 231: How large is this string? Should this maybe be a debug message?

internal/ignition/discovery.go
```diff
@@ -325,19 +356,13 @@ func (ib *ignitionBuilder) FormatDiscoveryIgnitionFile(ctx context.Context, infr
...
-			return "", mirrorsErr
-		}
-		ignitionParams["MirrorRegistriesConfig"] = base64.StdEncoding.EncodeToString(registriesContents)
-		ignitionParams["MirrorRegistriesCAConfig"] = base64.StdEncoding.EncodeToString(caContents)
+	mirrorRegistryConfiguration, err := infraEnv.GetMirrorRegistryConfiguration()
+	if err != nil {
+		return "", err
+	}
+
+	if err = ib.setIgnitionMirrorRegistry(&ignitionParams, mirrorRegistryConfiguration); err != nil {
```
carbonin, line 364: Why pass a reference here? Maps are already references.

https://go.dev/play/p/MOgHmWF1LLp

internal/ignition/discovery_test.go
```diff
@@ -728,6 +730,93 @@ var _ = Describe("IgnitionBuilder", func() {
 		})
 	})
 
+	const (
```
carbonin, line 733: Given that you have a bunch of setup/helpers that are specific to your tests, maybe here is a good place to create a separate context for them.

internal/installcfg/builder/builder.go
```diff
@@ -132,52 +133,93 @@ func (i *installConfigBuilder) getBasicInstallConfig(cluster *common.Cluster) (*
...
-	imageDigestSourceList := make([]installcfg.ImageDigestSource, len(mirrorRegistriesConfigs))
-	for i, mirrorRegistriesConfig := range mirrorRegistriesConfigs {
-		imageDigestSourceList[i] = installcfg.ImageDigestSource{Source: mirrorRegistriesConfig.Location, Mirrors: mirrorRegistriesConfig.Mirror}
+	configuration, err := cluster.GetMirrorRegistryConfiguration()
+	if err != nil {
+		return err
 	}
-	cfg.ImageDigestSources = imageDigestSourceList
+
+	i.log.Infof("found OpenShift version %s, setting mirror registry %+v", cluster.OpenshiftVersion, configuration)
```
carbonin, line 153: Same here, is this a large string? Maybe debug would be better?

internal/installcfg/builder/builder.go
```diff
@@ -132,52 +133,93 @@ func (i *installConfigBuilder) getBasicInstallConfig(cluster *common.Cluster) (*
...
+
 	return nil
 }
 
-func (i *installConfigBuilder) setImageContentSources(cfg *installcfg.InstallerConfigBaremetal) error {
-	mirrorRegistriesConfigs, err := i.mirrorRegistriesBuilder.ExtractLocationMirrorDataFromRegistries()
-	if err != nil {
-		i.log.WithError(err).Errorf("Failed to get the mirror registries conf need for ImageContentSources")
-		return err
+func (i *installConfigBuilder) setImageDigestMirrorSet(cfg *installcfg.InstallerConfigBaremetal, configuration *v1beta1.MirrorRegistryConfiguration) error {
```
carbonin, line 168: Is there a way we can deduplicate these functions?

pkg/mirrorregistries/mirror_registry.go
```diff
@@ -0,0 +1,189 @@
+package mirrorregistries
```
carbonin, line 1: Maybe a better file name would be `parser.go`?
Seems a bit weird to just have this be the package name in a different format.

internal/bminventory/inventory.go
eliorerz: Good catch! We also need to set it if it’s nil (in case the mirror configuration was removed)

internal/bminventory/inventory_test.go
eliorerz: Sure, added verifications

internal/bminventory/inventory_test.go
eliorerz: Added

internal/common/db.go
eliorerz: Yes, it is used in the update cluster

internal/common/db.go
eliorerz: No reason actually, I'll change it to `serialized json` for consistency

api/hiveextension/v1beta1/agentclusterinstall_types.go
eliorerz: Moved to api/common

internal/controller/controllers/clusterdeployments_controller.go
eliorerz: Not sure I understood what you meant here

internal/common/db.go
eliorerz: Fixed that in last commit

internal/ignition/discovery_test.go
eliorerz: It was suupposed to part of `mirror registries config` context. I'll fix taht

internal/controller/controllers/clusterdeployments_controller.go
carbonin: You're sending a configmap reference to the `mirrorregistries` package, but I would expect that package to take the data as input rather than the configmap.

For example `ProcessMirrorRegistryConfig` should also be usable from the REST API functions (if we were ever to implement them)

internal/installcfg/builder/builder.go
eliorerz: Not really; both methods, `setImageContentSources` and `setImageDigestMirrorSet` loop over `ImageDigestMirrors` and set different elements within this loop. One sets the `imageContentSourceList`, and the other sets the `imageDigestSourceList`.
If we want to combine them, we would need to loop over this data twice—once for extracting the data and another time for setting it.
In my opinion, this approach is better and more readable than the combined one.

api/common/common_types.go
```diff
@@ -16,3 +18,12 @@ type ValidationsStatus map[string]ValidationResults
 
 // +kubebuilder:object:generate=true
 type ValidationResults []ValidationResult
+
+// MirrorRegistryConfiguration holds the given mirror registry configuration
+type MirrorRegistryConfiguration struct {
```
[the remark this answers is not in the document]
eliorerz: Right, when I removed it from aci status it became relevent there

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -548,7 +548,7 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
 
 	updated := false
 	if string(conf) != infraEnvInternal.InternalIgnitionConfigOverride {
-		_, err = r.Installer.UpdateInfraEnvInternal(ctx, installer.UpdateInfraEnvParams{InfraEnvID: *infraEnvInternal.ID, InfraEnvUpdateParams: &models.InfraEnvUpdateParams{}}, swag.String(string(conf)))
+		_, err = r.Installer.UpdateInfraEnvInternal(ctx, installer.UpdateInfraEnvParams{InfraEnvID: *infraEnvInternal.ID, InfraEnvUpdateParams: &models.InfraEnvUpdateParams{}}, swag.String(string(conf)), nil)
```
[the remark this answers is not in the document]
eliorerz: In a previous [comment](https://github.com/openshift/assisted-service/pull/6868#discussion_r1819651450), you mentioned that it is safe to remove this logic from here (or did I misunderstand your comment?).

internal/ignition/discovery.go
```diff
@@ -14,9 +14,11 @@ import (
 	"text/template"
 	"time"
 
+	common_api "github.com/openshift/assisted-service/api/common"
 	clusterPkg "github.com/openshift/assisted-service/internal/cluster"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/constants"
+	"github.com/openshift/assisted-service/internal/controller/controllers/mirror_registry"
```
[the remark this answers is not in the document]
eliorerz: When I moved `MirrorRegistryConfiguration` to `common/db` I moved this method alongside with it

pkg/mirrorregistries/generator.go
```diff
@@ -78,34 +78,15 @@ func (m *mirrorRegistriesConfigBuilder) ExtractLocationMirrorDataFromRegistries(
...
-		for i := range mirrorTree {
-			currentMirror, ok := mirrorTree[i].Get("location").(string)
-			if !ok {
-				return nil, fmt.Errorf("failed to cast mirror location key to string, registriesConfToml: %s", registriesConfToml)
-			}
-			mirrors = append(mirrors, currentMirror)
-		}
-		registriesConfList[i] = RegistriesConf{Location: location, Mirror: mirrors}
+	for _, i := range idmsMirrors {
+		mirrors := *(*[]string)(unsafe.Pointer(&i.Mirrors))
```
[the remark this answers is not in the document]
eliorerz: As discussed I created a conversion function for that

internal/controller/controllers/preprovisioningimage_controller.go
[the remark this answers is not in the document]
carbonin: That was before we moved the mirror registries to the database.

Ideally it still true, we shouldn't have to care about setting the mirror registries when we're just trying to update the internal override.

I'm not sure how to handle this while also allowing the user to clear these registries. Maybe the internal override and/or the mirror registries should be set with a function separate from this update one?

Alternatively maybe we need some kind of test to make sure `UpdateInfraEnvInternal` doesn't clear the mirror registries when a nil is passed like this if that's the intention (because it sure looks like it will to me).

internal/controller/controllers/preprovisioningimage_controller.go
```diff
@@ -546,9 +547,18 @@ func (r *PreprovisioningImageReconciler) AddIronicAgentToInfraEnv(ctx context.Co
 		return false, err
 	}
 
+	var mirrorRegistryConfiguration *common.MirrorRegistryConfiguration
+	if infraEnvInternal.MirrorRegistryConfiguration != "" {
```
carbonin, line 551: If we're only doing this for the internal ignition override, can we move this into the conditional?

Also generally I'm not thrilled with this approach, but I don't have a better alternative for now without making some significant unrelated changes to how we save the internal ignition so I suppose we'll just leave it for now.

eliorerz:
/override ci/prow/edge-e2e-ai-operator-disconnected-capi 
Due to it's broken

PR #6978, opened 2024-11-08
iscsi machine networks

adriengentil:
- **MGMT-18513: machine networks must not be part of iSCSI subnet**
- **MGMT-18513: Add machine networks validation when iSCSI boot disk is in use**


internal/host/validator.go
```diff
@@ -1829,3 +1831,62 @@ func (v *validator) inventoryHasIP(inventory *models.Inventory, ipAddress string
...
+			machineNetworkPrefix, err = netip.ParsePrefix(string(machineNetwork.Cidr))
+			if err != nil {
+				return true
+			}
+			return machineNetworkPrefix.Contains(nicIP.Addr())
+		})
+		return found
+	})
+
+	if err != nil {
```
rccrdpccl, line 1882: where is this error produced? seems to be from `network.FindInterfaceByIPString` which it's already handled

internal/host/validations_test.go
```diff
@@ -2339,4 +2339,255 @@ var _ = Describe("Validations test", func() {
...
+					{
+						Cidr:      "192.168.2.1/24",
+						ClusterID: clusterID,
+					},
+					{
+						Cidr:      "2001:0db8:85a3::/48",
+						ClusterID: clusterID,
+					},
+				},
+				[]*models.Interface{
```
rccrdpccl, line 2581: Is it worth adding a test case with an unparseable IP prefix? For both NIC and machine network?

rccrdpccl: Actually if test separately and additionally only the function it will be way easier, and we could add all these edge cases

internal/host/validator.go
```diff
@@ -1829,3 +1831,62 @@ func (v *validator) inventoryHasIP(inventory *models.Inventory, ipAddress string
...
+
+	nic, err := network.FindInterfaceByIPString(installationDisk.Iscsi.HostIPAddress, c.inventory.Interfaces)
+	if err != nil {
+		v.log.WithError(err).Warn("Cannot find network interface associated to iSCSI host IP address")
+		return ValidationError, "Cannot find network interface associated to iSCSI host IP address"
+	}
+
+	// Look if one of the IP set on the interface connected to the iSCSI disk belong to machine network CIDRs
+	ips := append(nic.IPV4Addresses, nic.IPV6Addresses...)
+	_, found := lo.Find(ips, func(ip string) bool {
```
rccrdpccl, line 1864: Maybe we could extract this function to make it both clearer to the reader and more testeable (as commented in the test we could test edge cases in the function test, and more generic tests in the validation test)

swagger.yaml
```diff
@@ -6503,6 +6503,7 @@ definitions:
       - 'no-skip-installation-disk'
       - 'no-skip-missing-disk'
       - 'no-ip-collisions-in-network'
+      - 'iscsi-host-network-interface-does-not-belong-to-machine-cidr'
```
eifrach, line 6506: let make it a bit shorter -- It still OK leaving it as is 
also better you have a validation be a `True` if it's good and `False` when it fails - just a bit easier to understand when seen/debugging 

@AlonaKaplan what to you think ?
```suggestion
      - 'iscsi-net-belong-to-machine-cidr'
```

internal/host/validator.go
```diff
@@ -1829,3 +1831,62 @@ func (v *validator) inventoryHasIP(inventory *models.Inventory, ipAddress string
...
+		_, found := lo.Find(c.cluster.MachineNetworks, func(machineNetwork *models.MachineNetwork) bool {
+			var machineNetworkPrefix netip.Prefix
+			machineNetworkPrefix, err = netip.ParsePrefix(string(machineNetwork.Cidr))
+			if err != nil {
+				return true
+			}
+			return machineNetworkPrefix.Contains(nicIP.Addr())
+		})
+		return found
+	})
```
eifrach, line 1880: If I understand, you are trying to check if the host IP in the CIDR
let take this part and this check 
https://github.com/openshift/assisted-service/blob/master/internal/network/machine_network_cidr.go#L519
and create a common function to serve them both

make less repeating code

internal/host/hostcommands/install_cmd.go
```diff
@@ -382,20 +382,9 @@ func appendISCSIArgs(installerArgs []string, installationDisk *models.Disk, inve
...
-			return err == nil && iSCSIHostIP.Compare(prefix.Addr()) == 0
-		})
-		return ok
-	})
-
-	if !ok {
-		return nil, fmt.Errorf("Cannot find the interface belonging to iSCSI host IP %s", iSCSIHostIP.String())
+	nic, err := network.FindInterfaceByIP(iSCSIHostIP, inventory.Interfaces)
+	if err != nil {
+		return nil, err
```
eifrach, line 387: let's keep the error message - helps debuging and better logging 
```suggestion
		return nil, fmt.Errorf("Cannot find the interface belonging to iSCSI host IP %s", iSCSIHostIP.String())
```

internal/network/utils_test.go
```diff
@@ -1012,3 +1013,40 @@ var _ = Describe("GetVips", func() {
...
+			},
+			&searchedInterface,
+		}
+
+	})
+	DescribeTable("FindInterfaceByIP", func(allInterfaces *[]*models.Interface, lookForIp string, expectedInterface *models.Interface, expectedError string) {
+		int, err := FindInterfaceByIPString(lookForIp, *allInterfaces)
+		if expectedError == "" {
+			Expect(err).To(BeNil())
+			Expect(int).To(Equal(expectedInterface))
```
eifrach, line 1040: let's add debug output for the expect command
```suggestion
			Expect(err).To(BeNil(), fmt.Sprintf(err))
			Expect(int).To(Equal(expectedInterface),
			    fmt.Sprintf("expected: %s to be in interface %s", int , expectedInterface ))
```

internal/network/utils_test.go
```diff
@@ -1012,3 +1013,40 @@ var _ = Describe("GetVips", func() {
...
+		} else {
+			Expect(err).ToNot(BeNil())
+			Expect(err.Error()).To(Equal(expectedError))
```
eifrach, line 1043: same

internal/host/transition_test.go
```diff
@@ -6708,6 +6709,19 @@ var _ = Describe("State machine test - refresh transition", func() {
...
+
+			refreshHostArgs.conditions[string(IscsiHostNetworkInterfaceDoesNotBelongToMachineCidr)] = false
+
+			Expect(stateMachine.Run(TransitionTypeRefresh, testState, &refreshHostArgs)).To(Succeed())
+			Expect(string(testState.State())).To(Equal(models.HostStatusInsufficient))
+
+			refreshHostArgs.conditions[string(IscsiHostNetworkInterfaceDoesNotBelongToMachineCidr)] = true
+
+			Expect(stateMachine.Run(TransitionTypeRefresh, testState, &refreshHostArgs)).To(Succeed())
+			Expect(string(testState.State())).To(Equal(models.HostStatusKnown))
```
eifrach, line 6723: same as before let make it a practice to add debug message to Expect .
This will help see the root cause and not some generic failure

eifrach (changes requested):
Great Work, small changes just  to make it easier to read / deubg

I  also want you to add deubg output for the `Expect` I didn't comment on all of them.

swagger.yaml
adriengentil: maybe `iscsi-nic-belong-to-machine-cidr` WDYT?

internal/host/validations_test.go
adriengentil: MachineNetwork is hard to test because swagger already validates that the field is already a CIDR. So the test don't even comes to this validation.

adriengentil: agree that if I may extract a unit test if I split the function

internal/host/transition_test.go
rccrdpccl: I think it's a great idea to add error expectations, but I think it should be done by type and not by error/debug message. Otherwise tests will be even more fragile (text change will make us fix tests) , and our codebase is already struggling with test fragility.

internal/network/utils_test.go
adriengentil: the issue is to avoid the truncation ?

internal/host/hostcommands/install_cmd.go
pastequo: Not sure it's a good idea to lose the original error. Should be wrapped instead

internal/network/utils_test.go
eifrach: so the message will only be shown if the test will fail -- so it just for debug on failure

internal/network/machine_network_cidr.go
```diff
@@ -516,25 +516,58 @@ func GetDefaultRouteNetworkByFamily(h *models.Host, networks map[AddressFamily][
...
+		}
+		parsedCidr = append(parsedCidr, machineIpnet)
+	}
 
+	return parsedCidr, nil
+}
+
+// Check if a host belongs to all the networks specified as Machine Networks.
+func IsHostInPrimaryMachineNetCidr(log logrus.FieldLogger, cluster *common.Cluster, host *models.Host) bool {
 	// TODO(mko) This rule should be revised as soon as OCP supports multiple machineNetwork
```
eifrach, line 535: let's check if it's ClusterManagedNetworking retrun true
cause we allow only one network

internal/network/utils_test.go
adriengentil: I don't understand the point, the test wil already output the diff with the name of the test. Why adding one more output `expected: %s to be in interface %s"` ?

adriengentil: just had a look, the report looks like that in case of failure:
```
$ SKIP_UT_DB=1 go test ./internal/network -ginkgo.focus "FindInterfaceByIP"
Running Suite: Machine network cider Suite
==========================================
Random Seed: 1731427695
Will run 5 of 218 specs

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS•
------------------------------
• Failure [0.000 seconds]
FindInterfaceByIP
/home/agentil/Projects/openshift/assisted-service/internal/network/utils_test.go:1016
  FindInterfaceByIP
  /home/agentil/Projects/openshift/assisted-service/internal/network/utils_test.go:1036
    Look for IPv6 [It]
    /home/agentil/Projects/openshift/assisted-service/internal/network/utils_test.go:1047

    Expected
        <*errors.errorString | 0xc000583d10>: 
        Cannot find the network interface on which the IP 1001:db8:0:200::75 is set
        {
            s: "Cannot find the network interface on which the IP 1001:db8:0:200::75 is set",
        }
    to be nil

    /home/agentil/Projects/openshift/assisted-service/internal/network/utils_test.go:1039
------------------------------
•••SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Summarizing 1 Failure:

[Fail] FindInterfaceByIP FindInterfaceByIP [It] Look for IPv6 
/home/agentil/Projects/openshift/assisted-service/internal/network/utils_test.go:1039

Ran 5 of 218 Specs in 0.001 seconds
FAIL! -- 4 Passed | 1 Failed | 0 Pending | 213 Skipped
--- FAIL: TestMachineNetworkCidr (0.03s)
FAIL
FAIL	github.com/openshift/assisted-service/internal/network	0.105s
FAIL

```

internal/network/machine_network_cidr.go
eifrach: ignore it - It OK
you can have different host network form ISCSI network

internal/network/utils_test.go
eifrach: in this case it's not required but, when using `tables` or loops than you don't know what vaules have failed 
but let's take it offline i'll show some examples

internal/host/validator.go
adriengentil: it got refactored with functions in network packages

internal/host/validator.go
adriengentil: from `netip.ParsePrefix` because `err` is captured in the closure

internal/host/transition_test.go
adriengentil: let's take it offline

internal/network/utils_test.go
adriengentil: let's take it offline

PR #6983, opened 2024-11-11
MGMT-19230: Handle installation preparation status in statemachine

paul-maidment:
There is a bug that manifests when the installation endpoint for a cluster is called while the cluster is in an invalid state for install.
The installation request is correctly rejected, however the "Last Installation Preparation Status" fields are prematurely reset to LastInstallationPreparationStatusNotStarted.
These should only be reset when the cluster transitions between the "ready" and "preparing for installation" states.

This PR introduces a test for this assertion and additionally changes behaviour to match.

I introduced a test for "reset" cluster and functionality to apply the Last Installation Preparation Status reset when the cluster installation is reset.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/transition.go
+++ b/internal/cluster/transition.go
@@ -37,6 +37,8 @@ var resetProgressFields = []interface{}{"progress_finalizing_stage_percentage", 
 
 var resetFields = append(append(resetProgressFields, resetLogsField...), "openshift_cluster_id", "")
 
+var resetInstallationPreparationStatusFields = []interface{}{"last_installation_preparation_status", models.LastInstallationPreparationStatusNotStarted, "last_installation_preparation_reason", constants.InstallationPreparationReasonNotPerformed}
+
 type transitionHandler struct {
 	log                 logrus.FieldLogger
 	db                  *gorm.DB
@@ -175,7 +177,7 @@ func (th *transitionHandler) PostResetCluster(sw stateswitch.StateSwitch, args s
 			return err
 		}
 	}
-
+	extra = append(extra, resetInstallationPreparationStatusFields...)
 	return th.updateTransitionCluster(params.ctx, logutil.FromContext(params.ctx, th.log), params.db, sCluster, params.reason, extra...)
 }
 
@@ -201,6 +203,7 @@ func (th *transitionHandler) PostPrepareForInstallation(sw stateswitch.StateSwit
 		return errors.New("PostPrepareForInstallation invalid argument")
 	}
 	extra := append(append(make([]interface{}, 0), "install_started_at", strfmt.DateTime(time.Now())), resetLogsField...)
+	extra = append(extra, resetInstallationPreparationStatusFields...)
 	err = th.updateTransitionCluster(params.ctx, logutil.FromContext(params.ctx, th.log), th.db, sCluster,
 		statusInfoPreparingForInstallation, extra...)
 	if err != nil {
--- a/internal/constants/installation_preparation_status_reasons.go
+++ b/internal/constants/installation_preparation_status_reasons.go
@@ -1,0 +1,6 @@
+package constants
+
+const (
+	InstallationPreparationReasonNotPerformed = "Installation preparation not performed"
+	InstallationPreparationReasonSuccess      = "Installation preparation successful"
+)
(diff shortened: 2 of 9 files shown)


gamli75:
@tsorya please review

paul-maidment:
@gamli75 
> Not sure we can test all the cases in unit tests.
" can we make sure that we are not resetting the reason and time after few seconds?" -> By moving the "reset" code to the statemachine, we actually make this impossible, the state machine will not transition to an illegal state and the reset is only handled once...per transition.

I wrote a subsystem test that works to prove this anyway, you should see it in the PR. In this test, 

1: I prepare an installation to the point of success, which places the cluster in an installing state.
2: I then make a call to "install" the cluster while in this state (expecting to receive a rejection of the request)
* It was at this point that the illegal reset would have taken place...
3: I then check that LastInstallationPreparationStatus has remained in the same condition.

paul-maidment:
Can we test reset the installation and verify that we are resetting the last preparing reason and time?

I didn't put this into the reset transition, it's not actually necessary. Though I can do this for the sake of making things more intuitive to the end user.

paul-maidment:
> Do we have tests for failures in "preparing for installation" stage? 
Yes we do, these make the most sense in transition tests. 
I made one exception, for the subsystem test in this PR. A red/green style test to support the move of the problematic code into the transitions.

paul-maidment:
> can we make sure that we are not resetting the reason and time after few seconds? 
The reason that this bug was discovered is that a user was invoking the `install` endpoint of the cluster API every 10 seconds. As we were incorrectly handling the reset of the last installation status there, that caused this bug.

I attempt to re-create that scenario in my subsystem test and I feel that it is well enough covered.

paul-maidment:
> Can we test reset the installation and verify that we are resetting the last preparing reason and time?

I will add this as it makes some sense to avoid user confusion around reset.

paul-maidment:
@gamli75 I have added tests for cluster reset. Performed some other tidy up.

internal/cluster/transition.go
```diff
@@ -176,7 +176,11 @@ func (th *transitionHandler) PostResetCluster(sw stateswitch.StateSwitch, args s
 		}
 	}
 
-	return th.updateTransitionCluster(params.ctx, logutil.FromContext(params.ctx, th.log), params.db, sCluster, params.reason, extra...)
+	err := th.updateTransitionCluster(params.ctx, logutil.FromContext(params.ctx, th.log), params.db, sCluster, params.reason, extra...)
```
tsorya, line 179: Can we do resetLastInstallationPreparationStatus as part of updateTransitionCluster with extra ?

paul-maidment: Makes sense, I will change

internal/host/refresh_status_preprocessor_test.go
```diff
@@ -114,7 +114,7 @@ var _ = Describe("Cluster Refresh Status Preprocessor", func() {
 		cluster.Hosts = []*models.Host{&hostToAdd}
 		cluster.LastInstallationPreparation = models.LastInstallationPreparation{
 			Status: models.LastInstallationPreparationStatusFailed,
-			Reason: "",
+			Reason: "There was a failure during preparation",
```
gamli75, line 117: why do you add this description? I didn't see that you use it in the PR

internal/constants/installation_preparation_status_reasons.go
```diff
@@ -0,0 +1,6 @@
+package constants
+
+const (
+	InstallationPreparationReasonNotPerformed = "Installation preparation not performed"
+	InstallationPreparationReasonSuccess      = "Installation preparation successful"
```
gamli75, line 5: please check with QE if they are verifying the reason in their automation test because they will need to update it

tsorya:
looks good as for me

internal/host/refresh_status_preprocessor_test.go
paul-maidment: An empty string never gets stored. If you want the field to change you have to give it a value.
The "Reason" field is a "Free text" field when the status is `LastInstallationPreparationStatusFailed`
This is to allow a detailed error message to be stored. 

In this particular case, it only matters that the field is present, not it's value as we don't use it in any test case apart from one which is covered elsewhere that deals with installation preparation failures.

paul-maidment: Yes, looking at this specific test, I don't see how we use `LastInstallationPreparation` at all,
I will remove this altogether.

gamli75 (approved):

PR #6962, opened 2024-11-01
MGMT-19158: Always try to cache the ClusterImageSet

CrystalChun:
Previously, the `ClusterImageSet` defined by a `ClusterDeployment` was only cached if the cluster was being created. This is problematic in the long run if assisted is restarted and the cluster is not recreated. Errors "getting the release image" will occur due to this.

To prevent this, we'll now always try to cache the ClusterImageSet defined for a cluster even if it's already installed. This also allows us to ensure that the image the user specified in the ClusterImageSet as the release image to be installed.

## List all the issues related to this PR

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [x] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [ ] No tests needed

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [x] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #6993, opened 2024-11-13
MGMT-19159: Removing caching all ClusterImageSets from GetReleaseImage

CrystalChun:
There have been several instances where caching all the ClusterImageSets causes issues because of how long it can take if there are a lot of CRs. This moves that caching outside of the GetReleaseImage function.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/versions/kube_api_versions.go
+++ b/internal/versions/kube_api_versions.go
@@ -49,21 +49,16 @@ func (h *kubeAPIVersionsHandler) GetMustGatherImages(openshiftVersion, cpuArchit
 }
 
 // GetReleaseImage attempts to retrieve a release image that matches a specified OpenShift version and CPU architecture.
-// It first tries to find the image in the local cache. If the image is not found in the cache and a Kubernetes client is available (KubeAPI mode),
-// it then fetches all cluster image sets, updates the cache with these image sets, and attempts the cache lookup again.
 func (h *kubeAPIVersionsHandler) GetReleaseImage(ctx context.Context, openshiftVersion, cpuArchitecture, pullSecret string) (*models.ReleaseImage, error) {
 	cpuArchitecture = common.NormalizeCPUArchitecture(cpuArchitecture)
-	image, err := h.getReleaseImageFromCache(openshiftVersion, cpuArchitecture)
-	if err == nil || h.kubeClient == nil {
-		return image, err
-	}
+	return h.getReleaseImageFromCache(openshiftVersion, cpuArchitecture)
+}
 
-	// The image doesn't exist in the cache.
-	// Fetch all the cluster image sets, cache them, then search the cache again
-
+func (h *kubeAPIVersionsHandler) CacheAllReleaseImages(ctx context.Context, pullSecret string) error {
+	// Fetch all the cluster image sets and caches them
 	clusterImageSets := &hivev1.ClusterImageSetList{}
 	if err := h.kubeClient.List(ctx, clusterImageSets); err != nil {
-		return nil, err
+		return err
 	}
 	var wg sync.WaitGroup
 	for _, clusterImageSet := range clusterImageSets.Items {
@@ -93,8 +88,7 @@ func (h *kubeAPIVersionsHandler) GetReleaseImage(ctx context.Context, openshiftV
 		}(clusterImageSet)
 	}
 	wg.Wait()
-
-	return h.getReleaseImageFromCache(openshiftVersion, cpuArchitecture)
+	return nil
 }
 
 // GetReleaseImageByURL retrieves a release image based on its URL.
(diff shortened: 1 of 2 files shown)


internal/versions/kube_api_versions.go
```diff
@@ -49,21 +49,16 @@ func (h *kubeAPIVersionsHandler) GetMustGatherImages(openshiftVersion, cpuArchit
...
-	if err == nil || h.kubeClient == nil {
-		return image, err
-	}
-
-	// The image doesn't exist in the cache.
-	// Fetch all the cluster image sets, cache them, then search the cache again
+	return h.getReleaseImageFromCache(openshiftVersion, cpuArchitecture)
+}
 
+func (h *kubeAPIVersionsHandler) CacheAllReleaseImages(ctx context.Context, pullSecret string) error {
```
carbonin, line 57: When does this get called?

internal/versions/kube_api_versions.go
```diff
@@ -49,21 +49,16 @@ func (h *kubeAPIVersionsHandler) GetMustGatherImages(openshiftVersion, cpuArchit
...
 func (h *kubeAPIVersionsHandler) GetReleaseImage(ctx context.Context, openshiftVersion, cpuArchitecture, pullSecret string) (*models.ReleaseImage, error) {
 	cpuArchitecture = common.NormalizeCPUArchitecture(cpuArchitecture)
-	image, err := h.getReleaseImageFromCache(openshiftVersion, cpuArchitecture)
-	if err == nil || h.kubeClient == nil {
-		return image, err
-	}
-
-	// The image doesn't exist in the cache.
-	// Fetch all the cluster image sets, cache them, then search the cache again
+	return h.getReleaseImageFromCache(openshiftVersion, cpuArchitecture)
```
carbonin, line 54: Are we okay if this just fails even when someone created a cluster image set for this image that hasn't been cached? How do we deal with this situation in that case?

internal/versions/kube_api_versions.go
CrystalChun: It doesn't for now, I've just kept this function in case we ever want to use it again. I have a few ideas for caching, but would want to run them by you (and we can take this offline too):

1. Cache when assisted starts - this should be a non-blocking op. However, this can cause the issue we saw previously where if a user specifies a ClusterImageSet, but we end up using a release image that's in the cache that doesn't exactly match the specified one.
2. Provide a field in an existing API (maybe in `AgentClusterInstall`) that tells assisted which cluster image sets to cache (instead of providing just one clusterimageset, maybe allow a list?)
3. Provide another trigger - a completely separate API - that the user can set to tell assisted to cache all image sets

internal/versions/kube_api_versions.go
CrystalChun: Theoretically, I think this shouldn't happen now due to the change I've put into https://github.com/openshift/assisted-service/pull/6962 

If a cluster exists, there will always be a cluster image set that's cached, and any flow that calls this function should have a cluster associated with it. If it does fail, then that would indicate there's no cluster and there should probably be one, right? I might've missed something too, so let me know!

carbonin: I don't think that's the right way to think about this since the cache isn't persisted across container restarts.

We'd need to ensure that we get to this part of the cluster deployment controller for every cluster every time our container starts and before we need that release image for anything else.

Situations like this are why I created the release image cache the way I did.

https://issues.redhat.com/browse/MGMT-8279 for example was one of the bugs I had in mind at the time.

Just having fetched that release image one time when the cluster was first created (or even updated) isn't really good enough. We need to be sure, every time we ask for that release image, that it has been added if it does exist as a ClusterImageSet.

Maybe something changed in the mean time that makes us run the check for the release image more frequently now, but I'm not convinced yet that it's safe to remove the cache miss handling just yet. Have you thought about cases like this? Are we sure this is okay for all current callers of `GetReleaseImage`?

If not (or if it's going to be hard to track in the future) maybe we can investigate some other way of making these failing scenarios faster? We have an epic to create a controller specifically for `ClusterImageSets` that I think needs some more thought and maybe we could do something like annotate `ClusterImageSets` that we know we can't pull and speed up cache miss processing that way.

carbonin:
cc @danmanor you might also be interested in this one

danmanor:
@CrystalChun As @carbonin mentioned, we have an epic https://issues.redhat.com/browse/MGMT-18175 that should eventually remove the caching and move data to the DB (as done in RESTAPI flow) using a new controller, this way ensuring data persistence across restarts and avoiding the processing during the http request which might take too much time. WDYT ?

PR #6998, opened 2024-11-14
MGMT-18986: Allow ISO type configuration from kube API

eliorerz:
### Feature goal
Allow users to configure to type of ISO they are given from the kubernetes API.
Today, only the minimal ISO is available, but this doesn't work for all use cases (for example, booting from a USB device, or for debugging networking issues).

### DoD
User will be able to set the ISO image type using supported APIs.

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @eliorerz


internal/controller/controllers/infraenv_controller.go
```diff
@@ -502,7 +500,14 @@ func (r *InfraEnvReconciler) createInfraEnv(ctx context.Context, log logrus.Fiel
...
 	}
-	createParams := CreateInfraEnvParams(infraEnv, r.getImageType(ctx, infraEnv), pullSecret, clusterID, osImageVersion)
+	imageType := r.getImageType(log, infraEnv)
+	infraEnv.Spec.ImageType = imageType
+	createParams := CreateInfraEnvParams(infraEnv, imageType, pullSecret, clusterID, osImageVersion)
+	if err = r.Update(ctx, infraEnv); err != nil {
+		err = errors.Wrapf(err, "failed to update InfraEnv %s with ImageType %s", infraEnv.Name, imageType)
+		log.WithError(err)
+		return nil, err
+	}
```
CrystalChun, line 510: I don't think it's preferable to update the InfraEnv's spec. The spec is there for the user to spec-ify their own configs and I don't think our controller should set this field, even if it was empty.

internal/controller/controllers/infraenv_controller.go
```diff
@@ -176,6 +172,9 @@ func (r *InfraEnvReconciler) updateInfraEnv(ctx context.Context, log logrus.Fiel
 	if len(infraEnv.Spec.AdditionalNTPSources) > 0 {
 		updateParams.InfraEnvUpdateParams.AdditionalNtpSources = swag.String(strings.Join(infraEnv.Spec.AdditionalNTPSources[:], ","))
 	}
+	if infraEnv.Spec.ImageType != common.ImageTypeValue(internalInfraEnv.Type) {
+		updateParams.InfraEnvUpdateParams.ImageType = infraEnv.Spec.ImageType
+	}
```
CrystalChun, line 177: For the update, do you think it's a good idea to get the image type like how you did it in the [create function](https://github.com/openshift/assisted-service/pull/6998/files#diff-402c091b3839c89d4be13bcb028e7bfaf13c9380e01d430556b7e7b27d356f1eR503) and compare that to what's stored in the db? 
```suggestion
        imageType := r.getImageType(log, infraEnv)
	if imageType != common.ImageTypeValue(internalInfraEnv.Type) {
		updateParams.InfraEnvUpdateParams.ImageType = imageType
	}
```

internal/controller/controllers/infraenv_controller_test.go
```diff
@@ -785,6 +786,63 @@ var _ = Describe("infraEnv reconcile", func() {
 		Expect(infraEnvImage.Status.InfraEnvDebugInfo.EventsURL).To(HavePrefix(eventURL))
 	})
 
+	It("create and update infraenv image type", func() {
+		clusterDeployment := newClusterDeployment("clusterDeployment", testNamespace, getDefaultClusterDeploymentSpec("clusterDeployment-test", "test-cluster-aci", "pull-secret"))
+		Expect(c.Create(ctx, clusterDeployment)).To(BeNil())
+		mockInstallerInternal.EXPECT().GetInfraEnvByKubeKey(gomock.Any()).Return(nil, gorm.ErrRecordNotFound)
+		//mockInstallerInternal.EXPECT().GetInfraEnvByKubeKey(gomock.Any()).Return(backendInfraEnv, nil)
```
CrystalChun, line 793: nit: uncomment or delete commented out code

internal/controller/controllers/infraenv_controller_test.go
```diff
@@ -785,6 +786,63 @@ var _ = Describe("infraEnv reconcile", func() {
...
+		mockInstallerInternal.EXPECT().GetClusterByKubeKey(gomock.Any()).Return(backEndCluster, nil)
+		//mockInstallerInternal.EXPECT().DeregisterInfraEnvInternal(gomock.Any(), gomock.Any()).Return(nil)
```
CrystalChun, line 795: nit: uncomment or delete commented out code

CrystalChun (changes requested):
This is a great change!! Thanks for doing this! Just a couple of small requests, but looks great overall

internal/controller/controllers/infraenv_controller.go
eliorerz: So leave it empty? Even if we updated the image type to some default value?

internal/controller/controllers/infraenv_controller_test.go
eliorerz: Thanks! Missed that :)

internal/controller/controllers/infraenv_controller.go
eliorerz: `getImageType` calls to `GetInfraEnvIsoImageType` that return default values on creating an infraenv

internal/controller/controllers/infraenv_controller.go
CrystalChun: Yes, if you want, the status can have what we set it to. The default value is expected normally as the user couldn't request the type before

internal/controller/controllers/infraenv_controller.go
CrystalChun: Yes and that should either match the value stored in the DB or if it's different, then we should update it

subsystem/kubeapi_test.go
```diff
@@ -1370,6 +1370,41 @@ location = "%s"
...
+		By("Verify default image type requested on creation of infra-env")
+		infraEnvKey := types.NamespacedName{
+			Namespace: Options.Namespace,
+			Name:      infraNsName.Name,
+		}
+		var infraEnv *common.InfraEnv
+		Eventually(func() bool {
+			infraEnv = getInfraEnvFromDBByKubeKey(ctx, db, infraEnvKey, waitForReconcileTimeout)
+			infraEnvCrd := getInfraEnvCRD(ctx, kubeClient, infraEnvKey)
+			return *infraEnv.InfraEnv.Type == models.ImageTypeMinimalIso && models.ImageTypeMinimalIso == infraEnvCrd.Spec.ImageType
```
CrystalChun, line 1390: I think this might need to be modified because of the recent changes
```suggestion
			return *infraEnv.InfraEnv.Type == models.ImageTypeMinimalIso
```

eliorerz: which recent changes are you referring to?

CrystalChun: Related to this comment: https://github.com/openshift/assisted-service/pull/6998#discussion_r1860919042
You're no longer setting the infraenv spec.imageType in the code

CrystalChun (approved):
/lgtm

carbonin:
/hold

I think you're missing CRD changes and bundle updates for this.

Make sure you run `make generate`

PR #7001, opened 2024-11-15
MGMT-19237: Update validation hook to support external platform

adriengentil:
Update the conversion function that maps hive palatform to AI platform
to handle the external platform spec.


adriengentil:
/hold
until I test it

PR #7012, opened 2024-11-19
MGMT-19166: Implement MTU validations for CMN and UMN

linoyaslan:
This PR introduces new validations for MTU. There are two separate validations: one for the CMN and another for the UMN. For CMN, we focus on reports from the agent that related specifically to the machine network. For UMN, we consider reports related to all interfaces. You can find the MTU checker added to the agent here - https://github.com/openshift/assisted-installer-agent/pull/821

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/refresh_status_preprocessor.go
+++ b/internal/host/refresh_status_preprocessor.go
@@ -220,6 +220,10 @@ func newValidations(v *validator) []validation {
 			condition: v.belongsToMajorityGroup,
 		},
 		{
+			id:        IsMtuValid,
+			condition: v.isMtuValid,
+		},
+		{
 			id:        IsPlatformNetworkSettingsValid,
 			condition: v.isValidPlatformNetworkSettings,
 		},
--- a/internal/host/validation_id.go
+++ b/internal/host/validation_id.go
@@ -60,6 +60,7 @@ const (
 	AreServerLessRequirementsSatisfied             = validationID(models.HostValidationIDServerlessRequirementsSatisfied)
 	AreOpenShiftAIRequirementsSatisfied            = validationID(models.HostValidationIDOpenshiftAiRequirementsSatisfied)
 	AreAuthorinoRequirementsSatisfied              = validationID(models.HostValidationIDAuthorinoRequirementsSatisfied)
+	IsMtuValid                                     = validationID(models.HostValidationIDMtuValid)
 )
 
 func (v validationID) category() (string, error) {
@@ -84,6 +85,7 @@ func (v validationID) category() (string, error) {
 		NonOverlappingSubnets,
 		IsReleaseDomainNameResolvedCorrectly,
 		NoIPCollisionsInNetwork,
+		IsMtuValid,
 		NoIscsiNicBelongsToMachineCidr:
 		return "network", nil
 	case HasInventory,
--- a/internal/host/validations_test.go
+++ b/internal/host/validations_test.go
@@ -2605,4 +2605,64 @@ var _ = Describe("Validations test", func() {
 			Entry("is imported", &models.Cluster{ID: &clusterID, Imported: swag.Bool(true)}),
 		)
 	})
+	Context("Is MTU Valid", func() {
+		var (
+			cluster common.Cluster
+			host    models.Host
+		)
+		BeforeEach(func() {
+			cluster = hostutil.GenerateTestCluster(clusterID)
+			hostId, infraEnvId := strfmt.UUID(uuid.New().String()), strfmt.UUID(uuid.New().String())
+			mockProviderRegistry.EXPECT().IsHostSupported(commontesting.EqPlatformType(models.PlatformTypeVsphere), gomock.Any()).Return(false, nil).AnyTimes()
+			host = hostutil.GenerateTestHostByKind(hostId, infraEnvId, &clusterID, models.HostStatusKnown, models.HostKindHost, models.HostRoleMaster)
+			host.Inventory = hostutil.GenerateMasterInventory()
+		})
+		DescribeTable("MTU validations across CMN and UMN", func(validationID validationID, mtuReports []*models.MtuReport, validationStatus ValidationStatus, isUMN bool) {
+			cluster.UserManagedNetworking = &isUMN
+
+			mNet := []*models.MachineNetwork{
+				{
+					Cidr:      "1.2.3.1/24",
+					ClusterID: clusterID,
+				},
+			}
+			cluster.MachineNetworks = mNet
+
+			connectivityReport := &models.ConnectivityReport{}
+			connectivityReport.RemoteHosts = append(connectivityReport.RemoteHosts, &models.ConnectivityRemoteHost{
+				HostID:    hostID,
+				MtuReport: mtuReports,
+			})
+
+			report, err := hostutil.MarshalConnectivityReport(connectivityReport)
+			Expect(err).To(BeNil())
+			host.Connectivity = report
+			Expect(db.Create(&cluster).Error).ToNot(HaveOccurred())
+			Expect(db.Create(&host).Error).ShouldNot(HaveOccurred())
+
+			mockAndRefreshStatus(&host)
+
+			refreshedHost := hostutil.GetHostFromDB(*host.ID, host.InfraEnvID, db).Host
+
+			status, _, ok := getValidationResult(refreshedHost.ValidationsInfo, validationID)
+			Expect(ok).To(BeTrue(), fmt.Sprintf("debuuging info: status %s, isOK: %t host: %#v cluster: %#v", status, ok, host, cluster))
+			Expect(status).To(Equal(validationStatus))
+		},
+			Entry("isMtuValidInMachineNetwork (CMN) - Happy flow - MTU reports in the machine network are successful", IsMtuValid, []*models.MtuReport{
+				{MtuSuccessful: true, RemoteIPAddress: "1.2.3.5", OutgoingNic: "eth0"},
+				{MtuSuccessful: false, RemoteIPAddress: "1.2.4.5", OutgoingNic: "eth0"},
+			}, ValidationSuccess, false),
+			Entry("isMtuValidInMachineNetwork (CMN) - Bad flow - MTU reports in the machine network are unsuccessful", IsMtuValid, []*models.MtuReport{
+				{MtuSuccessful: false, RemoteIPAddress: "1.2.3.5", OutgoingNic: "eth0"},
+			}, ValidationFailure, false),
+			Entry("isMtuValidAllInterfaces (UMN) - Happy flow - all Mtu reports successful", IsMtuValid, []*models.MtuReport{
+				{MtuSuccessful: true, RemoteIPAddress: "1.2.3.5", OutgoingNic: "eth0"},
+				{MtuSuccessful: true, RemoteIPAddress: "1.2.4.5", OutgoingNic: "eth0"},
+			}, ValidationSuccess, true),
+			Entry("isMtuValidAllInterfaces (UMN) - Bad flow - all MTU reports are unsuccessful", IsMtuValid, []*models.MtuReport{
+				{MtuSuccessful: false, RemoteIPAddress: "1.2.3.5", OutgoingNic: "eth0"},
+				{MtuSuccessful: false, RemoteIPAddress: "1.2.4.5", OutgoingNic: "eth0"},
+			}, ValidationFailure, true),
+		)
+	})
 })
--- a/internal/host/validator.go
+++ b/internal/host/validator.go
@@ -47,15 +47,14 @@ const (
 	maxHostAheadOfServiceTimeDiff                    = 1 * time.Hour
 	maxHostTimingMetrics                             = 4
 	maxPingCommandExamples                           = 4
-
 	// How much we allow the user to deviate from our official
 	// minimum. Some setups (e.g. VMs) might give just a bit less
 	// than requested, so we shouldn't be too strict about it
 	HostMemoryRequirementToleranceMiB int64 = 100
+	FailedToFindAction                      = "failed to find action for step"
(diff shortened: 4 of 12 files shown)


internal/host/refresh_status_preprocessor.go
```diff
@@ -93,6 +94,16 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *validationConte
 			message = validationDisabledByConfiguration
 			conditions[v.id.String()] = true
 		} else {
+			// We want to differentiate between CMN and UMN. If it's UMN, we will execute a IsMtuOKAllInterfaces validation; otherwise, IsMtuOKInMachineNetwork validation will be executed.
```
pastequo, line 97: Shouldn't those check be part of the validator function itself (and return early a ValidationSuccessSuppressOutput if the test is not relevant) ?

I'm not familiar with this part of the code, but it would keep this function generic

internal/host/validator.go
```diff
@@ -305,6 +305,57 @@ func (v *validator) isMediaConnected(c *validationContext) (ValidationStatus, st
...
 
+// isMtuOKAllInterfaces - this validation is intended for UMN, where the MTU report is checked across all interfaces.
+func (v *validator) isMtuOKAllInterfaces(c *validationContext) (ValidationStatus, string) {
+	if c.infraEnv != nil {
+		return ValidationSuccessSuppressOutput, ""
+	}
+
+	connectivityReport, err := hostutil.UnmarshalConnectivityReport(c.host.Connectivity)
+	if err != nil {
+		return ValidationError, ""
```
pastequo, line 316: Is there a reason to hide the failure reason in this case (and below when it's not a ValidationSuccessSuppressOutput) ? If this information could be display to the end user, we shouldn't display the internal error (but it would be convenient to understand the problem, too bad to just lose it), but we can have something that gives hints where the check fail at least

internal/host/refresh_status_preprocessor.go
linoyaslan: Agree. Thanks!

internal/host/validation_id.go
```diff
@@ -59,6 +59,9 @@ const (
 	AreServiceMeshRequirementsSatisfied            = validationID(models.HostValidationIDServicemeshRequirementsSatisfied)
 	AreServerLessRequirementsSatisfied             = validationID(models.HostValidationIDServerlessRequirementsSatisfied)
 	AreOpenShiftAIRequirementsSatisfied            = validationID(models.HostValidationIDOpenshiftAiRequirementsSatisfied)
+
+	IsMtuOKAllInterfaces    = validationID(models.HostValidationIDMtuTestBetweenHostsAllInterfaces)
+	IsMtuOKInMachineNetwork = validationID(models.HostValidationIDMtuTestBetweenHostsInMachineNetwork)
```
eifrach, line 64: formating ?

internal/host/validations_test.go
```diff
@@ -2605,4 +2605,134 @@ var _ = Describe("Validations test", func() {
...
+
+			mockAndRefreshStatus(&host)
+
+			refreshedHost := hostutil.GetHostFromDB(*host.ID, host.InfraEnvID, db).Host
+
+			status, _, ok := getValidationResult(refreshedHost.ValidationsInfo, IsMtuOKInMachineNetwork)
+			Expect(ok).To(BeTrue(), fmt.Sprintf("debuuging info: status %s, isOK: %t host: %#v cluster: %#v", status, ok, host, cluster))
+			Expect(status).To(Equal(ValidationFailure))
+		})
+	})
```
eifrach, line 2674: moving this to Table can shorten this code

internal/host/validations_test.go
```diff
@@ -2605,4 +2605,134 @@ var _ = Describe("Validations test", func() {
...
+	Context("Is MTU OK - UMN (Verify All Reports)", func() {
+		var (
+			cluster common.Cluster
+			host    models.Host
+		)
+		BeforeEach(func() {
+			cluster = hostutil.GenerateTestCluster(clusterID)
+			hostId, infraEnvId := strfmt.UUID(uuid.New().String()), strfmt.UUID(uuid.New().String())
+			mockProviderRegistry.EXPECT().IsHostSupported(commontesting.EqPlatformType(models.PlatformTypeVsphere), gomock.Any()).Return(false, nil).AnyTimes()
```
eifrach, line 2683: why we need to check the provider type? 
if we remove it we can merge both CMN and UNM to same table test

internal/host/validations_test.go
```diff
@@ -2605,4 +2605,134 @@ var _ = Describe("Validations test", func() {
...
+			isUMN := true
+			cluster.UserManagedNetworking = &isUMN
+		})
+		It("Happy flow - all Mtu reports successful", func() {
+			connectivityReport := &models.ConnectivityReport{}
+			connectivityReport.RemoteHosts = append(connectivityReport.RemoteHosts, &models.ConnectivityRemoteHost{
+				HostID: hostID,
+				MtuReport: []*models.MtuReport{
+					{MtuSuccessful: true, RemoteIPAddress: "192.168.127.31", OutgoingNic: "ens3"},
+					{MtuSuccessful: true, RemoteIPAddress: "192.168.145.31", OutgoingNic: "ens4"},
```
eifrach, line 2695: I would like to see also  IPv6 and Dual stack tests

internal/host/validations_test.go
```diff
@@ -2605,4 +2605,134 @@ var _ = Describe("Validations test", func() {
...
 	})
+	Context("Is MTU OK - CMN (Verify Reports Related to Machine Networks Only)", func() {
+		var (
+			cluster common.Cluster
+			host    models.Host
+		)
+		BeforeEach(func() {
+			cluster = hostutil.GenerateTestCluster(clusterID)
+			hostId, infraEnvId := strfmt.UUID(uuid.New().String()), strfmt.UUID(uuid.New().String())
+			mockProviderRegistry.EXPECT().IsHostSupported(commontesting.EqPlatformType(models.PlatformTypeVsphere), gomock.Any()).Return(false, nil).AnyTimes()
```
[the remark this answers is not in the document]
linoyaslan: Since the `isVSphereDiskUUIDEnabled` validation is executed in each unit test (each unit test runs all the validations), we specifically need the mock of `IsHostSupported` for the vSphere platform - https://github.com/openshift/assisted-service/blob/dffaec3a63661db97437147c2d0987dde482bfa5/internal/host/validator.go#L1674

internal/host/validations_test.go
linoyaslan: Agree. Thanks

internal/host/validations_test.go
linoyaslan: I answered this in the comment above.

internal/host/validator.go
```diff
@@ -52,10 +52,11 @@ const (
 	// minimum. Some setups (e.g. VMs) might give just a bit less
 	// than requested, so we shouldn't be too strict about it
 	HostMemoryRequirementToleranceMiB int64 = 100
-)
 
-const FailedToFindAction = "failed to find action for step"
-const OpenStackPlatform = "OpenStack Compute"
+	FailedToFindAction = "failed to find action for step"
+	OpenStackPlatform  = "OpenStack Compute"
+	LateBindingMsg     = "Host is not bound to a cluster"
```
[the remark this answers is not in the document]
linoyaslan: I'm not sure I understand

internal/host/validations_test.go
linoyaslan: Just to clarify my understanding—why is this important here? I agree that checking IPv6 on the agent side is necessary, but in this case, we aren't concerned with whether it's IPv6 or IPv4. What matters here is whether the report  successful or not.

ori-amizur (commented):
What about state machine changes? How will this validation fail if there are no state machine changes?

linoyaslan:
> What about state machine changes? How will this validation fail if there are no state machine changes?

As discussed, in the first phase, we will treat the MTU validation as a warning rather than a requirement for installation.

swagger.yaml
```diff
@@ -6548,6 +6548,8 @@ definitions:
       - 'servicemesh-requirements-satisfied'
       - 'serverless-requirements-satisfied'
       - 'openshift-ai-requirements-satisfied'
+      - 'mtu-test-between-hosts-in-machine-network'
```
ori-amizur, line 6551: Why do we need 2 validation ids?

internal/host/refresh_status_preprocessor.go
```diff
@@ -219,6 +219,14 @@ func newValidations(v *validator) []validation {
 			id:        BelongsToMajorityGroup,
 			condition: v.belongsToMajorityGroup,
 		},
+		{
+			id:        IsMtuOKAllInterfaces,
```
ori-amizur, line 223: Why not have a single validation that will verify according to the current cluster setting ?

swagger.yaml
linoyaslan: We want to separate the validations for CMN and UMN, so that we can disable one while keeping the other enabled, or make one required for installation while the other remains a warning only etc..

internal/host/refresh_status_preprocessor.go
linoyaslan: As I commented above
>  we want to separate the validations for CMN and UMN, so that we can disable one while keeping the other enabled, or make one required for installation while the other remains a warning only.

ori-amizur: You don't need to leave a warning for UMN if the current setting is CMN.
Having both validations seems confusing to me.
For connectivity check the belongs-to-majority-group is a single validation even though it is checked differently for CMN and UMN.

linoyaslan: If the validation isn't relevant to CMN/UMN, it will be skipped. See the example: https://github.com/openshift/assisted-service/pull/7012/files#diff-d7db3d1eb3c67167cb6aaffa4ed66cd1dcd7d2097d63648ecd03410a36f1283eR315-R316

linoyaslan: What I mean is that it was decided to separate the validations instead of combining them and separate in a condition, because we may want to make the one for CMN a requirement for installation, while for UMN, it might not be mandatory (for example). That's why they need to be separated. As I mentioned earlier, I added a condition to each validation so that if it's not relevant, it will be skipped.

ori-amizur: We can make it optional in the validation function by asking what type of cluster do we validate.
Is this separation a requirement ?

linoyaslan: > Is this separation a requirement ?
Yes
@AlonaKaplan

linoyaslan: As discussed, it was decided to implement a single validation for both CMN and UMN. If differentiation is needed in the future, it will be handled through a new rule.

internal/host/validator.go
```diff
@@ -305,6 +304,62 @@ func (v *validator) isMediaConnected(c *validationContext) (ValidationStatus, st
...
+// isMtuOKInMachineNetwork - intended for CMN, focusing exclusively on the MTU reports for the machine network.
+func (v *validator) isMtuOKInMachineNetwork(c *validationContext, connectivityReport *models.ConnectivityReport) (ValidationStatus, string) {
+	for _, machineNet := range c.cluster.MachineNetworks {
+		_, mNetwork, err := net.ParseCIDR(string(machineNet.Cidr))
+		if err != nil {
+			return ValidationError, "Internal error - failed to parse machine network CIDR"
+		}
+		for _, r := range connectivityReport.RemoteHosts {
+			for _, mtuReport := range r.MtuReport {
+				ip := net.ParseIP(mtuReport.RemoteIPAddress)
```
ori-amizur, line 350: Don't you need to check that the source interface has an address that belongs to the machine network ?

linoyaslan: I only consider IPs that are on the machine network. Take a look at the line below.
`if mNetwork.Contains(ip) `

ori-amizur: That is correct for the destination, but what about the source?

internal/host/validator.go
```diff
@@ -305,6 +304,81 @@ func (v *validator) isMediaConnected(c *validationContext) (ValidationStatus, st
...
+func (v *validator) isMtuOKInMachineNetwork(c *validationContext, connectivityReport *models.ConnectivityReport) (ValidationStatus, string) {
+	for _, machineNet := range c.cluster.MachineNetworks {
+		_, mNetwork, err := net.ParseCIDR(string(machineNet.Cidr))
+		if err != nil {
+			return ValidationError, "Internal error - failed to parse machine network CIDR"
+		}
+		for _, r := range connectivityReport.RemoteHosts {
+			for _, mtuReport := range r.MtuReport {
+				var sourceIP string
+				for _, nic := range c.inventory.Interfaces {
```
ori-amizur, line 354: Maybe put the test against the inventory in a function ?
Can you first check if the remoteIP is in the machineNetwork and then the local IP ?

linoyaslan: > Maybe put the test against the inventory in a function ?
Yes, I agree

> Can you first check if the remoteIP is in the machineNetwork and then the local IP ?
Can you explain why it is important?

eifrach:
/override Red Hat Konflux / assisted-service-rhel8-mce-downstream-main-on-pull-request

internal/host/validator.go
ori-amizur: > Can you explain why it is important?

Actually it is OK if you leave the order as is

PR #7029, opened 2024-11-26
OCPBUGS-44849: Provide fallback Xattr method where not supported in kernel. 

paul-maidment:
Some customers are expereincing breaking issues when their OS does not support xattr, most notably NFS + Xattr
This causes installations not to proceed and needs to be addressed urgently.

This change provides a means of 'falling back' to a filesystem based implementation in the case that xattr is unavailable in the kernel.

This is achieved by the creation of an XattrClient interface and three implementations of this client

OSxattrClient - which wraps the kernel native xattr functions and is responsible for determining if xattr is available or not.
FilesystemBasedXattrClient - Which uses a directory parallel to the root to maintain metadata keys as individual files - one file per key

CompositeXattrClient - Which encapsulates the other two clients and dependent on the availability of xattr in the kernel will call the right client as applicable.

If we need to disable the fallback for any reason then the correct way to do this is to create a custom config map for assisted service.
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-config-map
  namespace: multicluster-engine
data:
  ENABLE_XATTR_FALLBACK: "false"
```

This should then be added as an annotation to the AgentServiceConfig
```
kind: AgentServiceConfig
metadata:
  annotations:
    unsupported.agent-install.openshift.io/assisted-service-configmap: custom-config-map

```

Once you have updated the AgentServiceConfig, redeploy the service
```
oc rollout restart deployment -n multicluster-engine assisted-service
```

After the service redeploys, you should see an entry in the assisted service log to indicate that the setting has been picked up

```
oc logs -f assisted-service-66dfcbf689-95v5f -c asisted-service | grep -i FALLBACK
time="2024-12-03T09:46:36Z" level=info msg="Options.EnableXattrFallback:false" func=main.main file="/assisted-service/cmd/main.go:332"
```


pkg/s3wrapper/filesystem.go
```diff
@@ -55,22 +55,62 @@ func (f *FSClient) CreateBucket() error {
...
+		if err != nil && !isUnsupportedXattrError(err) {
 			return errors.Wrapf(err, "unable to store metadata key %s = %s", attributeName, attributeValue)
 		}
+		// Warn that Xattr is unsupported in the filesystem for the path - then ignore the error
+		f.log.Warnf(
+			"Error while setting xattr %s=%s for path %s, the filesystem does not support extended attributes (xattr) - consider upgrading to one that supports xattr",
+			fmt.Sprintf("%s%s", xattrUserAttributePrefix, attributeName),
+			attributeValue,
+			filePath,
+		)
```
jhernand, line 89: This will generate a log message each time we try to write or read a file, right? Won't that be too much logging? Can we instead do a check when the FSClient object is created, write the log message, save the result in flag, and then avoid the xattr operation (and the log message) when it is true?

paul-maidment: I think the main issue with this is that the way we detect Xattr enablement ism with the use of an actual file/path, we respond to the exception when we attempt to write.

On creation of the FSClient, we don't have the context required to check this.

paul-maidment: I went for adding a flag to limit the warning messages to a single message.

pkg/s3wrapper/filesystem_xattr_client.go
```diff
@@ -0,0 +1,117 @@
...
+) *FilesystemBasedXattrClient {
+	return &FilesystemBasedXattrClient{
+		log:           log,
+		rootDirectory: rootDirectory,
+	}
+}
+
+const (
+	filesystemXattrMetaDataDirectoryName = "metadata"
+	delimiter                            = "::"
```
jhernand, line 25: I wanted to test this with a real filesystem that doesn't support extended attributes, so I used VFAT. Then I noticed that `:` (and many other punctuation characters) aren't allowed in VFAT file names, so creating the files failed with `Invaild argument`. I doubt any real user will use VFAT, but just for my testing, can we change this to something that will be allowed in VFAT, for example `;`?

pkg/s3wrapper/filesystem_xattr_client.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+type FilesystemBasedXattrClient struct {
+	log           logrus.FieldLogger
+	rootDirectory string
+}
+
+func (xattrClient *FilesystemBasedXattrClient) getPathWithKey(path string, key string) string {
```
jhernand, line 33: Would be nice to have a shorter receiver name here, for example `c`.

pkg/s3wrapper/filesystem_xattr_client.go
```diff
@@ -0,0 +1,117 @@
...
+// getMetaFilenameForManifestFile Finds the base of the filename and the directory for the manifest metadata files
+func (xattrClient *FilesystemBasedXattrClient) getMetaFilenamePrefixForManifestFile(path string) (fileName string, directory string) {
+	relativePath := path[len(xattrClient.rootDirectory):]
+	fileName = filepath.Join(xattrClient.rootDirectory, filesystemXattrMetaDataDirectoryName, relativePath)
+	directory = filepath.Dir(fileName)
+	return
+}
+
+func (c *FilesystemBasedXattrClient) SetRootDirectory(path string) {
+	c.rootDirectory = path
```
jhernand, line 51: Why do we need to set and get this? It is passed in the `New...` function, and should never be changed.

pkg/s3wrapper/xattr_client.go
```diff
@@ -0,0 +1,170 @@
...
+	// Set applies an extended attribute to a given path
+	// parameter: key - The key of the extended attribute.
+	// parameter: value - The value of the extended attribute.
+	// returns: an error if this fails, otherwise nil.
+	Set(tempFileName string, fileName string, key string, value string) error
+
+	// Get returns a value for an extended attribute key on a given given path.
+	// parameter: path - The path for which the key is to be fetched.
+	// parameter: key - The key of the extended attribute.
+	// returns the attribute value as a string, a boolean to indicate whether the value is ok, returns an error if there was one
```
jhernand, line 33: Can we make it clearer that `ok` will be `true` if the attribute existed and `false` otherwise?

pkg/s3wrapper/xattr_client.go
```diff
@@ -0,0 +1,170 @@
...
+	// List obtains a list of extended attribute keys for a given path
+	// parameter: path - The path for which extended attributes are to be fetched.
+	// returns: a list of extended attribute keys for the path or an error if this fails.
+	List(path string) ([]string, error)
+
+	// Removes an extended attribute for the file at the given path.
+	// parameter: path - The path for which the extended attribute is to be deleted.
+	// parameter: key - The key of the extended attribute.
+	// returns an error if there was an error
+	Remove(path string, key string) error
```
jhernand, line 45: Can we make it clear in the documentation that removing an attribute that doesn't exist does nothing and returns `nil`.

pkg/s3wrapper/xattr_client.go
```diff
@@ -0,0 +1,170 @@
...
+
+	// Removes all extended attributes for the file at the given path.
+	// parameter: path - The path for which the extended attributes are to be deleted.
+	// returns an error if there was an error
+	RemoveAll(path string) error
```
jhernand, line 50: Can we make it clear in the documentation what happens when the file doesn't exist?

pkg/s3wrapper/composite_xattr_client.go
```diff
@@ -0,0 +1,122 @@
...
+	"slices"
+
+	"github.com/sirupsen/logrus"
+)
+
+func NewCompositeXattrClient(
+	log logrus.FieldLogger,
+	oSxattrClient XattrClient,
+	filesystemBasedXattrClient XattrClient,
+	rootDirectory string,
```
jhernand, line 13: I think this `rootDirectory` isn't needed.

cmd/main.go
```diff
@@ -325,8 +326,16 @@ func main() {
 	failOnError(err, "failed to create ignition builder")
 	installConfigBuilder := installcfg.NewInstallConfigBuilder(log.WithField("pkg", "installcfg"), mirrorRegistriesBuilder, providerRegistry)
 
+	compositeXattrClient, err := s3wrapper.NewCompositeXattrClient(
+		log,
+		s3wrapper.NewOSxAttrClient(log, filepath.Join(Options.WorkDir, Options.S3Config.S3Bucket)),
+		s3wrapper.NewFilesystemBasedXattrClient(log, filepath.Join(Options.WorkDir, Options.S3Config.S3Bucket)),
```
jhernand, line 332: You may want to avoid repeating the construction of the root directory in these ^ tool lines. Maybe do it before calling the `New...` function?

pkg/s3wrapper/filesystem_xattr_client.go
paul-maidment: The short answer is that we cannot construct until we know the baseDir and we only know this at runtime for our tests

See here:
https://github.com/openshift/assisted-service/blob/745923e17a5dfab66cb92c8c9fc975aac9008649/pkg/s3wrapper/filesystem_test.go#L44-L58

pkg/s3wrapper/xattr_client.go
```diff
@@ -0,0 +1,170 @@
+package s3wrapper
```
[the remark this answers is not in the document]
paul-maidment: We should split out os OSxattr_client and write tests for that, otherwise this is largely an interface

PR #7030, opened 2024-11-26
MGMT-19360: do not monitor hosts with status installed

adriengentil:
Performance test in stage showed that the assisted-service continued to
monitor hosts even after the cluster installed. This behavior consumes
a lot of CPU, this PR disables it by stopping monitoring  hosts in "Installed" 
state when cluster state is "Installed".

Also, move the logic from `SkipMonitoring` function into the SQL query so
we avoid to retrieve hosts that we be skipped anyway (and save some
un-marshaling operations).

### DIFF
--- a/internal/host/monitor.go
+++ b/internal/host/monitor.go
@@ -14,13 +14,6 @@ import (
 	"gorm.io/gorm"
 )
 
-func (m *Manager) SkipMonitoring(h *models.Host) bool {
-	skipMonitoringStates := []string{string(models.LogsStateCompleted), string(models.LogsStateTimeout), ""}
-	result := ((swag.StringValue(h.Status) == models.HostStatusError || swag.StringValue(h.Status) == models.HostStatusCancelled) &&
-		funk.Contains(skipMonitoringStates, h.LogsInfo))
-	return result
-}
-
 func (m *Manager) initMonitoringQueryGenerator() {
 	if m.monitorClusterQueryGenerator == nil {
 		buildInitialQuery := func(db *gorm.DB) *gorm.DB {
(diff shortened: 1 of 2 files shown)


internal/host/monitor.go
```diff
@@ -14,13 +14,6 @@ import (
 	"gorm.io/gorm"
 )
 
-func (m *Manager) SkipMonitoring(h *models.Host) bool {
```
tsorya, line 17: nice !!!

internal/host/monitor.go
```diff
@@ -14,13 +14,6 @@ import (
...
-	skipMonitoringStates := []string{string(models.LogsStateCompleted), string(models.LogsStateTimeout), ""}
-	result := ((swag.StringValue(h.Status) == models.HostStatusError || swag.StringValue(h.Status) == models.HostStatusCancelled) &&
-		funk.Contains(skipMonitoringStates, h.LogsInfo))
-	return result
-}
-
 func (m *Manager) initMonitoringQueryGenerator() {
```
tsorya, line 17: Looks great , do we have a unit-test for it? if yes lets push

internal/host/monitor.go
```diff
@@ -35,14 +28,22 @@ func (m *Manager) initMonitoringQueryGenerator() {
...
-				models.HostStatusError,     // for limited time, until log collection finished or timed-out
 			}
 
-			dbWithCondition := common.LoadTableFromDB(db, common.HostsTable, "status in (?)", monitorStates)
+			// monitor following states for limited time, until log collection finished or timed-out
+			monitorStatesUntilLogCollection := []string{
+				models.HostStatusCancelled,
+				models.HostStatusError,
+			}
+			logCollectionEndStates := []string{
```
tsorya, line 40: can't we do it another way? like bring only those that are not in completed state?

internal/host/monitor.go
adriengentil: yeah, noticed that yestarday that there were not unit tests, I added some!

rccrdpccl:
amazing :heart: 

/lgtm

internal/host/monitor.go
```diff
@@ -35,16 +28,29 @@ func (m *Manager) initMonitoringQueryGenerator() {
...
+				string(models.LogsStateCompleted),
+				string(models.LogsStateTimeout),
+				string(models.LogsStateEmpty),
+			}
+
+			dbWithCondition := common.LoadClusterTablesFromDB(db)
```
rccrdpccl, line 46: How is this working? How is this selecting hosts? Am I missing something here? :thinking:

adriengentil: this query selects clusters, not hosts (then the hosts are embeded in the cluster object). I fought with gorm to understand what all this does.
Now you tell me that, I wonder if we select distinct clusters. I need to check that.

adriengentil: By the why, I see we batch queries here: https://github.com/openshift/assisted-service/blob/21d36c5b6b81d480fbfdb39666175abdbf75c964/internal/common/monitor_query_generator.go#L41

Wouldn't it make sense to use a cursor instead? (I don't know how it works with gorm though)

rccrdpccl: Oh, you're right - I got confused by the deleted `"exists (select 1 from hosts where clusters.id = hosts.cluster_id)"`. As we are using subqueries and selecting from clusters it'll be bound to be distinct, right?

adriengentil: yep, no problems

internal/host/monitor.go
```diff
@@ -35,16 +28,29 @@ func (m *Manager) initMonitoringQueryGenerator() {
...
+			dbWithCondition = dbWithCondition.Where(
+				"id IN (SELECT cluster_id FROM hosts WHERE hosts.status in (?) OR (hosts.status in (?) AND hosts.logs_info not in (?)))",
+				monitorStates, monitorStatesUntilLogCollection, logCollectionEndStates)
+			dbWithCondition = dbWithCondition.Or(
+				"id IN (SELECT cluster_id FROM hosts WHERE clusters.id = hosts.cluster_id AND hosts.status = ? AND clusters.status <> ?)",
```
eranco74, line 51: This adds clusters with status != installed if they have an installed host.
note that prior to this change the clusters selected by this query would go through the `SkipMonitoring` filter but after this PR they don't

PR #7036, opened 2024-11-28
MGMT-19272: Enable ODF for non standard HA OCP control plane cluster

danmanor:

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/operators/manager_test.go
```diff
@@ -543,13 +543,20 @@ var _ = Describe("Operators manager", func() {
...
 		It("should be not valid if not enough disk space available for mce and odf", func() {
-			clusterHost = getMockHostWithDisks(int64(20), int64(20))
+			clusterHost = getMockHostWithDisks(int64(50), int64(50))
 
 			cluster.MonitoredOperators = []*models.MonitoredOperator{
 				&odf.Operator,
 				&mce.Operator,
 			}
 
+			// To get the right deployment mode
```
rccrdpccl, line 553: Does `To get the right deployment mode` mean `target compact mode`?

internal/operators/manager_test.go
```diff
@@ -543,13 +543,20 @@ var _ = Describe("Operators manager", func() {
 		})
 
 		It("should be not valid if not enough disk space available for mce and odf", func() {
-			clusterHost = getMockHostWithDisks(int64(20), int64(20))
+			clusterHost = getMockHostWithDisks(int64(50), int64(50))
```
rccrdpccl, line 546: Why did this change?

danmanor: This test wants to test a scenario where there is mce + odf => there should be at least 25 (ODF requirement) + 50 (+MCE requirement) = 75 GB. The 20 GB would fail for ODF even without MCE, so I increased it to 50

internal/operators/odf/common.go
```diff
@@ -0,0 +1,42 @@
...
+		return standardMode
+	}
+
+	// To be determined, the cluster is not yet in a valid form.
+	return unknown
+}
+
+// To keep compatability with the behaviour until now.
+// TODO - remove this once two control plane nodes OpenShift is implemented
+func isAutoAssignmentAllowed(cluster *models.Cluster) bool {
```
rccrdpccl, line 35: Shouldn't this part of the following function (`willHostPotentiallyRunODF`)?

internal/operators/odf/common.go
```diff
@@ -0,0 +1,42 @@
...
+	return len(cluster.Hosts) == common.MinMasterHostsNeededForInstallationInHaMode
+}
+
+func willHostPotentiallyRunODF(mode odfDeploymentMode, hostEffectiveRole models.HostRole) bool {
```
rccrdpccl, line 39: wouldnt `canRunODF`  or `canHostRunODF` better (naming-wise)? Also maybe we can retrieve `hostEffectiveRole` within this function?

internal/operators/odf/odf_operator.go
```diff
@@ -218,47 +241,30 @@ func (o *operator) GetHostRequirements(_ context.Context, cluster *common.Cluste
 
 		/* getValidDiskCount counts the total number of valid disks in each host and return a error if we don't have the disk of required size,
 		we ignore the error as its treated as 500 in the UI */
-		diskCount, _ = o.getValidDiskCount(inventory.Disks, host.InstallationDiskID, &cluster.Cluster, nil)
+		diskCount, _ = o.getValidDiskCount(inventory.Disks, host.InstallationDiskID, nil, mode)
 	}
 
-	role := common.GetEffectiveRole(host)
-
+	// master / auto-assign in compact mode
```
rccrdpccl, line 247: nit: probably redundant comment

internal/operators/odf/odf_operator.go
```diff
@@ -218,47 +241,30 @@ func (o *operator) GetHostRequirements(_ context.Context, cluster *common.Cluste
...
-	}
 
-	// worker and auto-assign
-	if diskCount > 0 {
-		// for each disk odf requires 2 CPUs and 5 GiB RAM
+		// Each ODF disk requires 2 CPUs and 5 GiB RAM
 		return &models.ClusterHostRequirementsDetails{
-			CPUCores: o.config.ODFPerHostCPUStandardMode + (diskCount * o.config.ODFPerDiskCPUCount),
-			RAMMib:   conversions.GibToMib(o.config.ODFPerHostMemoryGiBStandardMode + (diskCount * o.config.ODFPerDiskRAMGiB)),
+			CPUCores: o.config.ODFPerHostCPUCompactMode + (reqDisks * o.config.ODFPerDiskCPUCount),
```
rccrdpccl, line 256: What's the difference between host in compact mode and worker in standard mode?
Why here we set it as `reqDisks` and below as `diskCount`? Is it because we can not run ODF if a node has no disk?

In this case, wouldn't something like this be clearer?

```suggestion
// Workers can have 0 requirements if no disk because reasons
d := diskCount
cpuPerHost = o.config.ODFPerHostCPUCompactMode
perHostRAMGiB = o.config.ODFPerHostMemoryGiBCompactMode

if mode == compactMode {
   d = math.Max(reqDisks, diskCount)
   cpuPerHost = o.config.ODFPerHostCPUStandardMode
   perHostRAMGiB = o.config.ODFPerHostMemoryGiBStandardMode
}

return &models.ClusterHostRequirementsDetails{
		CPUCores: cpuPerHost + (d * o.config.ODFPerDiskCPUCount),
		RAMMib:   conversions.GibToMib(o.config.perHostRAMGiB + (d * o.config.ODFPerDiskRAMGiB)),
	}, nil
```

internal/operators/odf/common.go
danmanor: It made more sense to me to split it to a different function, as it is more of a validation utility rather then a function who checks a condition

internal/operators/odf/common.go
danmanor: The check is not if the host is capable of running the ODF workloads, but will it eventually run them. if it will eventually run them, then we want to verify that it is capable of doing that. We can change it to `willHostRunODF` or something like that

danmanor: WDYT ?

internal/operators/odf/odf_operator.go
danmanor: This is an excellent question. This was the behaviour before my changes, and as I couldn't get to the bottom of this and This is not really related to my PR, I preferred leaving this as is.

internal/operators/odf/common.go
rccrdpccl: What's the difference between `can host eventually run ODF` vs `can host  run ODF`? (honest question)

internal/operators/manager_test.go
rccrdpccl: I see, however the test states `"should be not valid if not enough disk space available for mce and odf"`

internal/operators/manager_test.go
rccrdpccl: I think a comment like `// make sure it's compact mode` or something along those lines would be clearer

internal/operators/odf/common.go
rccrdpccl: But the code usage is always one after the other used in the same way.
If we group these functions (can be a third one just grouping them) it will make the code neater and DRY-er. Don't you think?

internal/operators/odf/odf_operator.go
rccrdpccl: I understand, however we are already simplifying the code that was there before by removing unnecessary if in the standard mode, we probably could go a step further and simplify it even more, WDYT?

internal/operators/manager_test.go
rccrdpccl: I see now, we want ODF to pass but ODF+MCE to fail.
Nice catch

internal/operators/odf/odf_operator.go
```diff
@@ -129,14 +124,30 @@ func (o *operator) getValidDiskCount(disks []*models.Disk, installationDiskID st
...
-	// We pass host validation to avoid false validation messages. The cluster validation will fail
-	masters, _, _ := common.GetHostsByEachRole(&cluster.Cluster, true)
-	if masterCount := len(masters); masterCount > 3 {
+	mode := getODFDeploymentMode(&cluster.Cluster, o.config.ODFNumMinimumHosts)
+	hostEffectiveRole := common.GetEffectiveRole(host)
+	shouldHostRunODF, err := shouldHostRunODF(&cluster.Cluster, mode, hostEffectiveRole)
+
+	// host role is auto-assign in standard mode.
+	if err != nil {
+		status := "For ODF with more than 3 hosts, role must be assigned to master or worker."
```
rccrdpccl, line 133: We are assuming internals of the function `shouldHostRunODF`. Maybe we can just print error message as status? `status := err.Error()`

internal/operators/odf/common.go
```diff
@@ -0,0 +1,58 @@
...
+	return unknown
+}
+
+// To keep compatability with the behaviour until now.
+// TODO - remove this once two control plane nodes OpenShift is implemented
+func isAutoAssignmentAllowed(cluster *models.Cluster) bool {
+	return len(cluster.Hosts) == common.MinMasterHostsNeededForInstallationInHaMode
+}
+
+// shouldHostRunODF returns whether a host will eventually run ODF worloads given the cluster it belongs to,
```
rccrdpccl, line 42: I think we should specify that the function will return:
* nil if the ability of running ODF cannot be evaluated
* &true if the host  should run ODF
* &false otherwise

* error for invalid hosts

internal/operators/odf/odf_operator.go
```diff
@@ -129,14 +124,30 @@ func (o *operator) getValidDiskCount(disks []*models.Disk, installationDiskID st
...
+	}
+
+	// unknown mode, we can't validate hosts yet. We pass the validation until the
+	// cluster will reach a state which we can validate. In the meanwhile, it will fail the cluster validations.
+	if shouldHostRunODF == nil {
+		return api.ValidationResult{Status: api.Success, ValidationId: o.GetHostValidationID(), Reasons: []string{}}, nil
+	}
+
+	// The host will not ODF workloads, therefore nothing to validate in ODF context.
+	if !*shouldHostRunODF {
```
rccrdpccl, line 147:
```suggestion
	if shouldHostRunODF == nil || !*shouldHostRunODF {
```

internal/operators/odf/odf_operator.go
```diff
@@ -199,15 +213,26 @@ func (o *operator) GetMonitoredOperator() *models.MonitoredOperator {
...
+	// can't tell what role will the host get eventually. We return no extra requirements, it will fail host validation.
+	if err != nil {
 		return &models.ClusterHostRequirementsDetails{CPUCores: 0, RAMMib: 0}, nil
 	}
 
-	numOfHosts := len(cluster.Hosts)
-	mode := getODFDeploymentMode(numOfHosts)
+	// unknown mode, we can't decide what the host requirements are yey. We return no furthr requirements
+	// until the cluster will reach a state which we can validate. In the meanwhile, it will fail the cluster validations.
+	if shouldHostRunODF == nil {
```
rccrdpccl, line 228: Same as before, this condition can be grouped with the one below

internal/operators/odf/validations.go
```diff
@@ -133,30 +96,23 @@ func (o *operator) computeNodeResourceUtil(cluster *models.Cluster, host *models
...
 	return status, nil
 }
 
 // used to validate resource requirements for ODF
-func (o *operator) canODFBeDeployed(hosts []*models.Host, odfClusterResources *odfClusterResourcesInfo) (bool, string) {
+func (o *operator) canODFBeDeployed(
+	odfClusterResources *odfClusterResourcesInfo,
+	mode odfDeploymentMode,
+) (bool, string) {
 	var status string
```
rccrdpccl, line 108: nit: we can remove status variable and just directly return the output of `o.setStatusInsufficientResources` and `fmt.Sprintf` functions.

internal/operators/odf/odf_operator.go
danmanor: I have an issue with the error message can't start with capital but the status should. I don't want to add extra code func to capitalize the first letter

rccrdpccl: I think we should just capitalize programmatically the first letter. This will make the code less fragile (more flexible to change) and less confusing (we won't ask ourselves "how come we know it's more than 3 hosts at this point?")

linoyaslan:
@rccrdpccl  already approved, just fixing his typo
/lgtm

PR #7052, opened 2024-12-03
MGMT-19362: Explicitly add OpenShift AI incompatibilities

jhernand:
The OpenShift AI requires ODF, and that means that it is indirectly incompatible with LVM and SNO. But our feature support mechanism doesn't currently understand dependencies between operators. As a result it is possible to update a cluster so that it will have OpenShift AI and SNO, for example, enabled simultaneously. To avoid that this patch adds those indirect incompatibilities to the OpenShift AI operator, like they were direct incompatibilities.

https://issues.redhat.com/browse/MGMT-19362

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Tested manually creating a cluster with SNO enabled and then patching it to add OpenShift AI. Same for LVM.

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/featuresupport/features_olm_operators.go
+++ b/internal/featuresupport/features_olm_operators.go
@@ -84,6 +84,7 @@ func (feature *LvmFeature) getIncompatibleFeatures(OCPVersion string) *[]models.
 		models.FeatureSupportLevelIDNUTANIXINTEGRATION,
 		models.FeatureSupportLevelIDVSPHEREINTEGRATION,
 		models.FeatureSupportLevelIDODF,
+		models.FeatureSupportLevelIDOPENSHIFTAI,
 	}
 	if isEqual, _ := common.BaseVersionLessThan("4.15", OCPVersion); isEqual {
 		incompatibleFeatures = append(incompatibleFeatures,
@@ -582,7 +583,12 @@ func (f *OpenShiftAIFeature) getIncompatibleArchitectures(_ *string) *[]models.A
 }
 
 func (f *OpenShiftAIFeature) getIncompatibleFeatures(string) *[]models.FeatureSupportLevelID {
-	return &[]models.FeatureSupportLevelID{}
+	return &[]models.FeatureSupportLevelID{
+		// These aren't directly incompatible with OpenShift AI, rather with ODF, but the feature support
+		// mechanism doesn't currently understand operator dependencies, so we need to add these explicitly.
+		models.FeatureSupportLevelIDLVM,
+		models.FeatureSupportLevelIDSNO,
+	}
 }
 
 func (f *OpenShiftAIFeature) getFeatureActiveLevel(cluster *common.Cluster, _ *models.InfraEnv, clusterUpdateParams *models.V2ClusterUpdateParams, _ *models.InfraEnvUpdateParams) featureActiveLevel {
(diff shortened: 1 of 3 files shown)


paul-maidment:
> The OpenShift AI requires ODF, and that means that it is indirectly incompatible with LVM and SNO.
Why does the requirement for ODF mean that Openshift AI is incompatible with LVM and SNO, what is the reason?
More of a curiosity thing on my part but interested in the reason.

internal/featuresupport/features_olm_operators.go
```diff
@@ -582,7 +583,12 @@ func (f *OpenShiftAIFeature) getIncompatibleArchitectures(_ *string) *[]models.A
 }
 
 func (f *OpenShiftAIFeature) getIncompatibleFeatures(string) *[]models.FeatureSupportLevelID {
-	return &[]models.FeatureSupportLevelID{}
+	return &[]models.FeatureSupportLevelID{
+		// These aren't directly incompatible with OpenShift AI, rather with ODF, but the feature support
+		// mechanism doesn't currently understand operator dependencies, so we need to add these explicitly.
```
paul-maidment, line 588: Should we raise a ticket to investigate how to factor in operator dependencies for the purpose of determining feature support?

It seems like this is an area that should be explored further, is there already an issue for this?

internal/featuresupport/features_olm_operators.go
```diff
@@ -84,6 +84,7 @@ func (feature *LvmFeature) getIncompatibleFeatures(OCPVersion string) *[]models.
 		models.FeatureSupportLevelIDNUTANIXINTEGRATION,
 		models.FeatureSupportLevelIDVSPHEREINTEGRATION,
 		models.FeatureSupportLevelIDODF,
+		models.FeatureSupportLevelIDOPENSHIFTAI,
```
paul-maidment, line 87: Should we put a comment - perhaps to an issue or otherwise providing some context on why the LVM and SNO features are incompatible with Openshift AI. What causes the problem?

paul-maidment (changes requested):
I think we need a little more context on what causes Openshift AI to be incompatible with LVM and SNO, at the very least a link to a more detailed explanation.

internal/featuresupport/features_olm_operators.go
jhernand: Yes @paul-maidment , we have https://issues.redhat.com/browse/MGMT-19471 to address that.

internal/featuresupport/features_olm_operators.go
jhernand: @paul-maidment OpenShift AI itself isn't explicitly incompatible with LVM or SNO. But it requires at least two worker nodes (I believe that is a soft limitation that will be soon lifted) and some kind of storage for the most basic things, like creating a workbench or deploying a model. We decided to include ODF in the bundle because that provides block storage (used for workbenches) and also S3-like storage (used to deploy models).

LVM would also work out of the box for workbenches, but doesn't provide directly the S3-like storage: it would require also Nooba.

ODF is incompatible itself with LVM, and that is not new with OpenShift AI: it has been like that since the ODF operator was introduced. I am not sure about the reasons for that, but I guess is related to the fact that with the default configuration both will try to use the same disks for storage, and the result will be unpredictable.

I have the feeling that explaining all these details doesn't belong in this patch, but I will add them if you think it is really needed.

paul-maidment: No, I think it's OK, if there is a link to a discussion of the issue, maybe that would add context for those who want to look deeper into this?

paul-maidment (approved):

PR #7062, opened 2024-12-04
ACM-6053: Add NTP to machineconfig even if no sync

jhernand:
The server instructs the agent to perform NTP synchronization, and when the results are received it stores the results in the `ntp_sources` column of the `hosts` table. But in the ZTP flow installation may start before that response is received, so that `ntp_sources` column may be empty. The value of this column is used to build the `MachineConfig` that configures NTP for the hosts, and we have a fallback that intends to use the NTP servers provided by the user in the `additionalNTPSources` field of the `InfraEnv` object. But the current code skips that fallback logic when the `ntp_sources` is empty. This patch fixes that.

Related: https://issues.redhat.com/browse/ACM-6053

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Tested removing the NTP synchronization step, adding the additional NTP sources to the infrastructure environment and trying to install the cluster. Before the patch no NPT sources was added to the generated machine config. After the patch the additional NTP sources are added.

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/network/manifests_generator.go
+++ b/internal/network/manifests_generator.go
@@ -193,18 +193,16 @@ func (m *ManifestsGenerator) createChronyManifestContent(c *common.Cluster, role
 	sources := make([]string, 0)
 
 	for _, host := range c.Hosts {
-		if host.NtpSources == "" {
-			continue
-		}
+		if host.NtpSources != "" {
+			var ntpSources []*models.NtpSource
+			if err := json.Unmarshal([]byte(host.NtpSources), &ntpSources); err != nil {
+				return nil, errors.Wrapf(err, "Failed to unmarshal %s", host.NtpSources)
+			}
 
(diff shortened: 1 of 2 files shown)


CrystalChun (commented):
Nice catch, looks great! Seems like a unit test is failing though, will approve once it's fixed.

omertuc:
ugh good catch... I remember staring at this code a few months ago looking for how this bug could possibly happen and I completely missed that

jhernand:
@CrystalChun @omertuc @tsorya fixed the failing tests, and reworked them to include the use cases that we want to cover. Please take another look.

PR #7090, opened 2024-12-13
MGMT-19120: Use service net to connect to hosted API server

jhernand:
There are several situations where assisted service needs to connect to the API server of a spoke cluster. To do so it uses the kubeconfig generated during the installation, and that usually contains the external URL of the API server, and that means that the cluster where assisted service runs needs to be configured with a proxy that allows that. But for HyperShift clusters this can be avoided: assisted service can instead connect via the service network, using the `kube-apiserver.my-cluster.svc` host name, as the API server runs as a pod in the same cluster. Doing that reduces the number of round trips and the potential proxy configuration issues. In order to achieve that this patch changes the spoke client factory so that it checks if the cluster is a HyperShift cluster, and then it replaces the API server URL with `https://kube-apiserver.my-cluster.svc:6443`.

https://issues.redhat.com/browse/MGMT-19120

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


gamli75:
@eranco74 can you review this PR?

internal/spoke_k8s_client/factory.go
```diff
@@ -1,90 +1,255 @@
...
+}
+
+// SetLogger sets the object that will be used to write to the log. This is mandatory.
+func (b *SpokeK8sClientFactoryBuilder) SetLogger(value logrus.FieldLogger) *SpokeK8sClientFactoryBuilder {
+	b.logger = value
+	return b
+}
+
+// SetHubClient sets the client that will be used to call the API of the hub cluster. This is mandatory.
+func (b *SpokeK8sClientFactoryBuilder) SetHubClient(value ctrlclient.Client) *SpokeK8sClientFactoryBuilder {
```
tsorya, line 52: why we can't provide client while creating the object? 
we have it here https://github.com/openshift/assisted-service/pull/7090/files#diff-891ba2cfffd82a8ae4131c88beb092d2a88149f579c931e3a1f7ca77fbfc82a5L166 no?

tsorya: sorry my fault missed 
https://github.com/openshift/assisted-service/pull/7090/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R625

internal/spoke_k8s_client/factory.go
```diff
@@ -1,90 +1,255 @@
...
+	// Create and populate the object:
+	result = &spokeK8sClientFactory{
+		logger:            b.logger,
+		hubClient:         b.hubClient,
+		transportWrappers: slices.Clone(b.transportWrappers),
+	}
+	return
+}
+
+func (f *spokeK8sClientFactory) CreateFromSecret(ctx context.Context, secret *corev1.Secret) (result SpokeK8sClient, err error) {
```
tsorya, line 85: I wonder why better to return result this way? 
(result SpokeK8sClient, err error) ?
I believe 99% of the code doesn't do it this way and i just wonder why it is better

internal/spoke_k8s_client/factory.go
```diff
@@ -1,90 +1,255 @@
...
 	}
+	result, err = clientConfig.ClientConfig()
+	return
+}
 
-	clientset, err := kubernetes.NewForConfig(restConfig)
-	if err != nil {
-		cf.log.WithError(err).Warnf("Getting kuberenetes config for cluster")
-		return nil, nil, err
+func (f *spokeK8sClientFactory) kubeConfigFromSecret(secret *corev1.Secret) (result []byte, err error) {
```
tsorya, line 153: Can we make it common function? We have at least 2 more place that do the same

internal/spoke_k8s_client/factory.go
```diff
@@ -1,90 +1,255 @@
...
+
+// isFromHostedCluster determines if this cluster corresponding to the given kubeconfig secret is a hosted cluster.
+// Returns a boolean flag and, when that flag is true, the namespace and name of the cluster deployment of the hosted
+// cluster.
+func (f *spokeK8sClientFactory) isFromHostedCluster(ctx context.Context, kubeconfigSecret *corev1.Secret) (result bool,
+	clusterDeploymentKey types.NamespacedName, err error) {
+	// Try to find the cluster deployment. If we can't, for whatever the reason, explain it in the log and assume
+	// it isn't a hosted cluster.
+	clusterDeployment, err := f.findClusterDeploymentForKubeconfigSecret(ctx, kubeconfigSecret)
+	if err != nil || clusterDeployment == nil {
```
tsorya, line 191: Error and not having clusterDeployment seems to be different issues, maybe we should split the logging at least?

internal/spoke_k8s_client/factory.go
jhernand: This is just a way to make the code cleaner, avoiding long lists of parameters. We could pass the the logger, the client (and the transport wrapper, only used currently for tests) as parameters to the "New..." function, but over time that results in long lists of parameters like this.

```go
api = NewManager(common.GetTestLog(), db, testing.GetDummyNotificationStream(ctrl), mockEventApi, nil, nil, nil, nil, &config, &leader.DummyElector{}, nil, nil, true, nil, nil, false)
```

It is already useful to avoid the setting the transport wrapper parameter to nil.

internal/spoke_k8s_client/factory.go
```diff
@@ -1,90 +1,255 @@
...
+// findClusterDeploymentForKubeconfigSecret finds the cluster deployment that corresponds to the given kubeconfig
+// secret. It returns nil if there is no such cluster deployment.
+func (f *spokeK8sClientFactory) findClusterDeploymentForKubeconfigSecret(ctx context.Context,
+	kubeconfigSecret *corev1.Secret) (result *hivev1.ClusterDeployment, err error) {
+	// The cluster deployment should be in the same namespace than the secret, because the cluster deployment
+	// references the secret from the 'spec.clusterMetadata.adminKubeconfigSecretRef' field, and that is a local
+	// object reference. So to find the cluster deployment we can get all the instances inside the namespace of the
+	// secret and then select the first one that references it.
+	clusterDeploymentList := &hivev1.ClusterDeploymentList{}
+	err = f.hubClient.List(ctx, clusterDeploymentList, ctrlclient.InNamespace(kubeconfigSecret.Namespace))
```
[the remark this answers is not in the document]
jhernand: Not sure what you mean, can you elaborate? Note that the search criteria here is `spec.clusterMetadata.adminKubeconfigSecretRef.Name == ...`, I think searching by that field isn't supported by the API.

internal/spoke_k8s_client/factory.go
jhernand: It is probably a matter of taste. I like to have the names of the return parameters: helps understand what to expect. Not very important in this case as the meaning is very clear. I can change it if you want.

internal/spoke_k8s_client/factory.go
jhernand: Well, at least we have one less place now: I removed similar logic from the spoke client cache in a previous patch. I will try to find where we are doing this.

internal/spoke_k8s_client/factory.go
jhernand: OK, will do.

internal/spoke_k8s_client/factory.go
tsorya: But this is required params and in this case you left them  as optional so i don't understand actually why it is good

internal/spoke_k8s_client/factory.go
tsorya: I agree that it is taste issue :) just most of the code have another style so why to have different styles?

internal/spoke_k8s_client/factory.go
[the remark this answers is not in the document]
tsorya: I believe it can be only one cluster deployment per namespace actually. Don't we have owner ref in the secret for clusterdeployment?

internal/spoke_k8s_client/factory.go
jhernand: I believe it is good for several reasons:

1. It is consistent: all the parameters (required or optional) are provided in the same way.

3. It makes it clearer what each parameter means. Not in this case, but if you had two parameters that are strings it is not the same to see this:

```go
whatever, err := NewWhatever("foo", "bar")
```

Than this:

```go
whatever, err := NewWhatever(().
        SetUserName("foo").
        SetPassword("bar").
        Build()
```

In the first case you have to deep digger to find out what is the meaning of the parameters, and in the second it is explicit.

4. It gives room for documenting each parameter separately: the documentation goes in the "Set..." method of the builder.

5. It simplifies building the object in multiple steps, if needed, for example:

```go
builder := NewWhatever()
builder.SetUserName("foo")
if shouldUsePassword {
        builder.SetPassword("bar")
}
whatever, err := builder.Build()
```

6. It simplifies adding multiple values for the same parameter:

```go
whatever, err := NewWahtever().
        SetUserName("foo").
        SetUserName("foo-alias").
        Build()
```

7. It allows adding new optional parameters without having to change the call sites.

I don't want to bore you with my opinions about this. If you find this unacceptable I will change it.

internal/spoke_k8s_client/factory.go
jhernand: Fair enough, I will change it.

internal/spoke_k8s_client/factory.go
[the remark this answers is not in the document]
jhernand: I'd say we don't need to rely on that here.

internal/spoke_k8s_client/factory.go
jhernand: Done.

internal/spoke_k8s_client/factory.go
jhernand: I will do this in a different patch.

internal/spoke_k8s_client/factory.go
```diff
@@ -1,90 +1,255 @@
...
 
-	spokeClient := &spokeK8sClient{
-		Client:      targetClient,
-		csrClient:   clientset.CertificatesV1().CertificateSigningRequests(),
-		sarClient:   clientset.AuthorizationV1().SelfSubjectAccessReviews(),
-		nodesClient: clientset.CoreV1().Nodes(),
-		log:         cf.log,
+// findClusterDeploymentForKubeconfigSecret finds the cluster deployment that corresponds to the given kubeconfig
+// secret. It returns nil if there is no such cluster deployment.
+func (f *spokeK8sClientFactory) findClusterDeploymentForKubeconfigSecret(ctx context.Context,
```
carbonin, line 216: Are we ever in a situation where the caller if this factory doesn't already have a reference to the cluster deployment?

Since (based on the naming) we're talking about "spoke" clusters it seems likely that this could be simplified by either the caller supplying the cluster deployment or by this logic living outside this factory (then we would have an option like "useHubServiceNetwork" or something when creating the client).

jhernand: Yes, here we don't know what is the cluster deployment: https://github.com/openshift/assisted-service/blob/9a1b9eceb81c43fb0655022d393fdf9911dd537c/internal/controller/controllers/hypershiftagentserviceconfig_controller.go#L342 .

carbonin: Okay, thanks.

Side note though ... can we delete the HASC CRD and controller yet?
@gamli75 that effort isn't happening now, right?

gamli75: I'm not familiar with that effort. maybe @CrystalChun

CrystalChun: I'm not familiar with it either, maybe @danielerez?

internal/spoke_k8s_client/factory.go
tsorya: Idk, maybe it just me, I just believe that if parameter is required it should be provided as part of function call another way if someone will write 
`whatever, err := NewWahtever().Build()`
it will pass compilation but will fail on the run an i think better to find such error in compilation.
Though it is my personal opinion

tsorya: Just to be sure, i like your proposition i just don't think it should be that way with required params

jhernand: I understand your point of view, and still think that the benefits outweigh the drawbacks. As that is not the key point of this pull request I am changing it to a plain list of parameters. We can have this discussion another time.

jhernand:
@carbonin @tsorya I think I addressed your concerns. Anything else that needs to be changed before merging?

gamli75:
> /test edge-e2e-ai-operator-disconnected-capi

this job is broken again, see: https://issues.redhat.com/browse/MGMT-19358

internal/spoke_k8s_client/factory.go
```diff
@@ -1,90 +1,224 @@
...
+		clusterDeploymentKey.Namespace = clusterDeployment.Namespace
+		clusterDeploymentKey.Name = clusterDeployment.Name
 	}
-	return spokeClient, clientset, nil
+	return result, clusterDeploymentKey, nil
+}
+
+// findClusterDeploymentForKubeconfigSecret finds the cluster deployment that corresponds to the given kubeconfig
+// secret. It returns nil if there is no such cluster deployment.
+func (f *spokeK8sClientFactory) findClusterDeploymentForKubeconfigSecret(ctx context.Context,
```
CrystalChun, line 185: What do you think about passing in a `ClusterDeployment` pointer as a parameter instead of the `ctx`?

That way we don't need to know the hub client and the ctx, and we don't need this spoke client factory to find the ClusterDeployment. The work will be on the caller to pass in the ClusterDeployment.

So even if the function caller doesn't have a ClusterDeployment, we can already assume it's not a HostedCluster and just use the secret as we already have.

internal/spoke_k8s_client/factory.go
```diff
@@ -1,90 +1,224 @@
...
+	client, _, err := f.ClientAndSetFromSecret(ctx, secret)
 	return client, err
 }
 
-func (cf *spokeK8sClientFactory) ClientAndSetFromSecret(secret *corev1.Secret) (SpokeK8sClient, *kubernetes.Clientset, error) {
-	kubeconfig, err := kubeconfigFromSecret(secret)
+func (f *spokeK8sClientFactory) ClientAndSetFromSecret(ctx context.Context, secret *corev1.Secret) (SpokeK8sClient,
+	*kubernetes.Clientset, error) {
+	// Create the REST configuration from the content of the secret:
+	restConfig, err := f.restConfigFromSecret(secret)
```
CrystalChun, line 61:
```suggestion
func (f *spokeK8sClientFactory) ClientAndSetFromSecret(clusterDeployment *hivev1.ClusterDeployment, secret *corev1.Secret) (SpokeK8sClient,
	*kubernetes.Clientset, error) {
	// Create the REST configuration from the content of the secret:
	restConfig, err := f.restConfigFromSecret(secret)
```

internal/spoke_k8s_client/factory.go
CrystalChun: To clarify my point the function `ClientAndSetFromSecret` would be 
```
func (f *spokeK8sClientFactory) ClientAndSetFromSecret(clusterDeployment *hivev1.ClusterDeployment, secret *corev1.Secret) (SpokeK8sClient,
	*kubernetes.Clientset, error) {
```

Then in `isFromHostedCluster`
```
func (f *spokeK8sClientFactory) isFromHostedCluster(clusterDeployment *hivev1.ClusterDeployment, kubeconfigSecret *corev1.Secret) bool {
        if clusterDeployment == nil { 
                return false
        }
	// We assume this is a hosted cluster if it has the 'agentClusterRef' label, no matter the value.
	_, result := clusterDeployment.Labels["agentClusterRef"]
	return result
}
```

Just thought maybe this could make it simpler and offload the responsibility of getting a ClusterDeployment to the caller as opposed to this client factory.

Let me know your thoughts or if you want me to clarify more offline!

paul-maidment: I think that specifically for context the idiomatic way to pass context is to ensure that it is the first parameter of a function. It is expected that context will be passed from function to function this way.

As this is so specific and idiomatic, I don't think we should hide the context inside a struct.

I do agree with the other encapsulations though,  just not the hiding of context.

hack/start_db.sh
```diff
@@ -8,4 +8,11 @@ mkdir -p /tmp/postgres/data
 mkdir -p /tmp/postgres/sockets
 
 initdb -D /tmp/postgres/data -U postgres
+
+# We store the data files in the `/tmp` directory assuming that it will be a `tmpfs` and therefore
+# very fast. But in some environments it may not be, as it may be mapped to a real persistent file
+# system. Disabling `fsync` speeds things up in those environment, at the cost of risking data loss,
```
paul-maidment, line 14: Maybe add a tiny bit more detail on the purpose of fysnc to this comment.

The context that this is ensuring a file is fully written even in the case of a node crash/power failure/reboot is relevant here as _we don't expect or need our data to survive a reboot_. This is the real reason we don't care about "data loss"

jhernand: The meaning of the `fsync` parameter is described in the documentation of PostgreSQ. For example, for version 12 it is here: https://www.postgresql.org/docs/12/runtime-config-wal.html . I was trying to explain that we don't care about data loss because this is used only in unit tests, and crash, power failure or reboot is not a concern in unit tests. Anyhow, this is in this pull request by accicent, not really needed. I will remove it in the next version to avoid noiise.

internal/spoke_k8s_client/factory.go
jhernand: @CrystalChun what you suggest makes sense. My only concern would be that hat adding another optional parameter (the pointer to the cluster deployment) means that it will be easier for someone to write `ClientAndSetFromSecret(nil, mySecret)`, maybe without thinking if the cluster deployment is available or not. That would result in wrong results if that missing cluster deployment was actually a HyperShift cluster. But I am going to apply your suggestion anyhow: seems like a good way to move this forward.

@paul-maidment I agree with you, but I guess that @CrystalChun means that if I apply her suggestion then context parameter isn't strictly needed, as there will be no operation in those functions that supports cancelation.

jhernand:
@CrystalChun I applied your suggestion.

@carbonin @tsorya @paul-maidment you may want to review again.

internal/controller/controllers/agent_controller.go
```diff
@@ -571,7 +571,26 @@ func (r *AgentReconciler) spokeKubeClient(ctx context.Context, clusterRef *aiv1b
...
+		Namespace: clusterRef.Namespace,
+		Name:      clusterRef.Name,
+	}
+	clusterDeployment := &hivev1.ClusterDeployment{}
+	err = r.Client.Get(ctx, clusterDeploymentKey, clusterDeployment)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			clusterDeployment = nil
+			err = nil
+		}
```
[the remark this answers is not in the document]
jhernand: I'd prefer to not change the behavior of this code, as I am not sure if the existence of a cluster deployment is a requirement here. If it is I assume that it would be checked before this point.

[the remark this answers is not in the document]
CrystalChun: Gotcha, didn't realize this was the behavior before. That's alright then!

CrystalChun:
Looks good to me! Thanks Juan! Will wait for others to review

hack/start_db.sh
paul-maidment: Sure. I think my point was just to mention what you say about unit tests, the original text doesn't fully explain this and leaves some room for interpretation.

hack/start_db.sh
```diff
@@ -8,4 +8,5 @@ mkdir -p /tmp/postgres/data
 mkdir -p /tmp/postgres/sockets
 
 initdb -D /tmp/postgres/data -U postgres
+
```
paul-maidment, line 11: Not much going on in this file anymore, I suspect you can just remove this newline to omit it from the commit

jhernand: That is completely right, I will.

jhernand: Done.

paul-maidment:
Looks good to me, awaiting, will leave this for further review by @tsorya @carbonin

PR #7096, opened 2024-12-15
MGMT-19500: Enable the installation of a baremetal cluster with a user-managed load balancer, where nodes are in the same subnet, and the load balancer IP is within the subnet

danmanor:
This patch adds support for using an user managed load balancer. It will only be available for OpenShift 4.16 or newer. To enable it the user will have to use the new `load_balancer.type` field of the `cluster` type, for example:

```json
{
  "load_balancer": {
    "type": "user-managed"
  }
}
```

To preserve backwards compatibility this will be optional, and the
default value will be `cluster-managed`.

When the value is `user-managed` the following will be added to the
generated `install-config.yaml` file:

```yaml
platform:
  baremetal:
    loadBalancer:
      type: UserManaged
    apiVIPs:
    - <load_balancer_ip>
    ingressVIPs:
    - <load_balancer_ip>
```
__Note__ that when using an user managed load balancer the VIPs will be
used to specify the load balancer IP (not actually virtual).

More info can be found here - https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/deploying_installer-provisioned_clusters_on_bare_metal/ipi-install-installation-workflow#nw-osp-configuring-external-load-balancer_ipi-install-installation-workflow

The database will be automatically migrated so that for existing
clusters the value of the `load_balancer_type` column will automatically
be set to `cluster-managed`.
    
This patch adds this support only for the `baremetal` platform. a follow-up PR should implement this logic for `vsphere` as well.

### Notes

- This PR still require all hosts to be on the same subnet (same broadcast range). There is going to be a follow-up PR to enable hosts in different subnets as well. 

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


swagger.yaml
```diff
@@ -7607,3 +7614,24 @@ definitions:
...
     - Disabled
+
+  load_balancer:
+    type: object
+    x-go-custom-tag: gorm:"embedded;embeddedPrefix:load_balancer_"
+    properties:
+      type:
+        x-go-custom-tag: gorm:"not null;check:load_balancer_type in ('cluster-managed', 'user-managed');default:'cluster-managed'"
+        description: |
+          Indicates if the load balancer will be managed by the cluster or by the user. This is optional and The
```
paul-maidment, line 7625: Is this only to be used in "User Managed Networking" mode or does this somehow work in "cluster managed" networking, with the only exception being load balancer config?

paul-maidment:
The title of this issue contains "....with user managed load balancer and nodes are part of the same subnet".
I notice that the validation code skips any kind of validation of this when the LB is user managed.

Is there a validation performed before the installation starts to ensure that the user LB is reachable?

swagger.yaml
danmanor: Yes

paul-maidment: I assume that's a "yes" to "This only works for UMN"

Do we verify that the networking is UMN when allowing this to be set?

danmanor: Oh no I meant this work in "cluster managed" networking, with the only exception being load balancer config

danmanor: I blocked it

danmanor: It is simpler to manage this feature as part of cluster-managed-networking rather then part of user-managed-networking

paul-maidment:
> @paul-maidment It doesn't skip all network validations, just ones that try to validate that -
>  
>     * the VIPs are not the same
> 
>     * the VIPs are part of the machine network.
> 
> 
> all other validations are performed as usual

One of the requirements to use this type of load balancer configuration is that each of the hosts being 'balanced' resides on the same subnet. Is there anything we need to do to validate this or due to the 'user defined' nature of the load balancer, is this not possible?

Do we verify that the user load balancer works? (while remaining unconcerned about `why` or `how` it works?)

danmanor:
/override ci/prow/edge-e2e-oci-assisted-4-14 ci/prow/edge-e2e-oci-assisted-4-18 ci/prow/edge-e2e-nutanix-assisted-4-14 ci/prow/edge-e2e-nutanix-assisted-4-18

danmanor:
/override ci/prow/edge-e2e-oci-assisted-4-14 ci/prow/edge-e2e-oci-assisted-4-18

danmanor:
/override ci/prow/edge-e2e-nutanix-assisted-4-14 ci/prow/edge-e2e-nutanix-assisted-4-18

danmanor:
@paul-maidment We don't validate that the load balancer is accessible, same as we do with user managed networking

danmanor:
The following jobs are broken: 

- disconnected capi
- nutanix platform
- oci platform

danmanor:
/override ci/prow/edge-e2e-oci-assisted-4-14 ci/prow/edge-e2e-oci-assisted-4-18

danmanor:
/override ci/prow/edge-e2e-nutanix-assisted-4-14 ci/prow/edge-e2e-nutanix-assisted-4-18

eranco74 (commented):
/lgtm

danmanor:
/override ci/prow/edge-e2e-oci-assisted-4-18 ci/prow/edge-e2e-oci-assisted-4-14

internal/network/machine_network_cidr.go
```diff
@@ -126,6 +134,11 @@ func VerifyVip(hosts []*models.Host, machineNetworkCidr string, vip string, vipN
 	if ipIsBroadcast(vip, machineNetworkCidr) {
 		return models.VipVerificationFailed, errors.Errorf("%s <%s> is the broadcast address of machine-network-cidr <%s>", vipName, vip, machineNetworkCidr)
 	}
+
+	if !verifyVipFree {
```
eranco74, line 138: Consider renameing to `shouldVerifyIfVipIsFree`, current name is too similar to  `VerifyVipFree`

danmanor:
/override ci/prow/edge-e2e-ai-operator-disconnected-capi ci/prow/edge-e2e-nutanix-assisted-4-18 ci/prow/okd-scos-e2e-aws-ovn

paul-maidment (commented):
Assuming that the user accepts responsibility for setting up their load balancer and that we do not need to verify the provided VIP then I think this works fine.

Do E2E tests currently make use of the API VIP in any way, if so then would it be worth the extra step of setting up a "user defined" load balancer for the sake of making sure everything still works when the VIP is not the one we have already set up in the test?

internal/network/machine_network_cidr.go
paul-maidment: Agree

danmanor:
@paul-maidment I created two new tests that test this exact scenario - 
- `edge-e2e-metal-assisted-kube-api-umlb-4-18`
- `edge-e2e-metal-assisted-umlb-4-18`

They both passed, there is a small patch I need to add to `kube-api-umlb-4-18` which currently fails the conformance tests (after the installation)

PR #7156, opened 2025-01-02
MGMT-19573: Track release stats in installercache

paul-maidment:
To support Ephemeral storage improvement efforts in [MGMT-13917](https://issues.redhat.com//browse/MGMT-13917) It is desirable to have some statistics from the installer cache

	ReleaseId                 The release being downloaded
	Cached                    Was the release found in the cache (true) or did it need to be downloaded/extracted? (false)
	StartTime                 The start time of the install cache request
	EndTime                   The time at which the caller finished using the request
	ExtractDuration           The time taken to extract the file in seconds (zero if no extraction took place)

This wil be sent in a single metrics event, the event time of the event represents the time that the release was no longer needed by callers. All of the above should be sufficient to give visibility on the lifespan and concurrency of intstaller cache entries.

In addition, it was requested that we are able to enable/disable the recording of metrics for this feature so I have implemented this. Implemented through the metricsManager as a list of metrics that we would like to block.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/ignition/installmanifests.go
+++ b/internal/ignition/installmanifests.go
@@ -22,6 +22,7 @@ import (
 	bmh_v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/constants"
+	eventsapi "github.com/openshift/assisted-service/internal/events/api"
 	"github.com/openshift/assisted-service/internal/host/hostutil"
 	"github.com/openshift/assisted-service/internal/installercache"
 	"github.com/openshift/assisted-service/internal/manifests"
@@ -111,7 +112,7 @@ var fileNames = [...]string{
 // NewGenerator returns a generator that can generate ignition files
 func NewGenerator(workDir string, installerDir string, cluster *common.Cluster, releaseImage string, releaseImageMirror string,
 	serviceCACert string, installInvoker string, s3Client s3wrapper.API, log logrus.FieldLogger, providerRegistry registry.ProviderRegistry,
-	installerReleaseImageOverride, clusterTLSCertOverrideDir string, storageCapacityLimit int64, manifestApi manifestsapi.ManifestsAPI) Generator {
+	installerReleaseImageOverride, clusterTLSCertOverrideDir string, storageCapacityLimit int64, manifestApi manifestsapi.ManifestsAPI, eventsHandler eventsapi.Handler) Generator {
 	return &installerGenerator{
 		cluster:                       cluster,
 		log:                           log,
@@ -126,7 +127,7 @@ func NewGenerator(workDir string, installerDir string, cluster *common.Cluster, 
 		providerRegistry:              providerRegistry,
 		installerReleaseImageOverride: installerReleaseImageOverride,
 		clusterTLSCertOverrideDir:     clusterTLSCertOverrideDir,
-		installerCache:                installercache.New(installerDir, storageCapacityLimit, log),
+		installerCache:                installercache.New(installerDir, storageCapacityLimit, eventsHandler, log),
 		manifestApi:                   manifestApi,
 	}
 }
@@ -173,13 +174,13 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte)
 		system.NewLocalSystemInfo(),
 	)
 
-	release, err := g.installerCache.Get(g.installerReleaseImageOverride, g.releaseImageMirror,
-		g.cluster.PullSecret, ocRelease, g.cluster.OpenshiftVersion)
+	release, err := g.installerCache.Get(ctx, g.installerReleaseImageOverride, g.releaseImageMirror,
+		g.cluster.PullSecret, ocRelease, g.cluster.OpenshiftVersion, *g.cluster.ID)
 	if err != nil {
 		return errors.Wrap(err, "failed to get installer path")
 	}
 	//cleanup resources at the end
-	defer release.Release()
+	defer release.Cleanup(ctx)
 
 	installerPath := release.Path
 	installConfigPath := filepath.Join(g.workDir, "install-config.yaml")
--- a/internal/installercache/installercache.go
+++ b/internal/installercache/installercache.go
@@ -1,6 +1,7 @@
 package installercache
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path"
@@ -9,7 +10,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-openapi/strfmt"
+	eventsapi "github.com/openshift/assisted-service/internal/events/api"
 	"github.com/openshift/assisted-service/internal/oc"
+	"github.com/openshift/assisted-service/models"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -28,7 +32,8 @@ type Installers struct {
 	// total capcity of the allowed storage (in bytes)
 	storageCapacity int64
 	// parent directory of the binary cache
-	cacheDir string
+	cacheDir      string
+	eventsHandler eventsapi.Handler
 }
 
 type fileInfo struct {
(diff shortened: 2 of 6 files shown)


carbonin:
Are you intending to track this through events or a metric?

I could see something like installer cache misses and release usage duration as good metrics, but what you're doing here with timestamps feels more like an event.

If you're intending to target only events you can probably implement it as an event only rather than adding something to the metrics manager.

@rccrdpccl what do you think? Should these be events or actual prometheus metrics?

carbonin:
My personal preference would be to use actual metrics as they would be more usable through more tools (like grafana), but maybe we also do something clever there with events that I don't know about :shrug:

paul-maidment:
> My personal preference would be to use actual metrics as they would be more usable through more tools (like grafana), but maybe we also do something clever there with events that I don't know about 🤷


I chatted through the various means of gathering stats with @rccrdpccl before he went on vacation.
There are a limited number of installation preparations and it is preferable to use events as the metrics don't allow enough precision. With events I can determine the concurrency as well as release sizes etc.

The volume is low enough to not cause any concern.

My intention is to use the SQL features of starburst enterprise to extract some reports.

rccrdpccl:
> @rccrdpccl what do you think? Should these be events or actual prometheus metrics?

It really depends on what question are we trying to answer.
As I understand, we are trying to answer the following:
* How many "releases" happen concurrently?
* How long does a "release" normally take? What about if it's "cached"?

Answering the first question with Prometheus can get tricky: if it's a gauge we might miss scraping the data due to scraping interval, if it's a counter it'd get complicated to handle and represent. The second question would also be really awkward to answer through Prometheus. However by querying events both implementation and actual answer should be straightforward.

Another factor that we should consider when choosing between metrics and events are alerts: we can alert on metrics, we cannot on events. In this case I think it's not relevant.

> If you're intending to target only events you can probably implement it as an event only rather than adding something to the metrics manager.

100% agree. Metrics manager should serve as a facade for Prometheus metrics, and although it's coupled with events handler, IMO it really shouldn't and we should try to avoid that approach.

> In addition, it was requested that we are able to enable/disable the recording of metrics for this feature so I have implemented this. Implemented through the metricsManager as a list of metrics that we would like to block.

@paul-maidment why do we need to enable/disable this metric? I do not see the benefit in doing so

paul-maidment:
> @paul-maidment why do we need to enable/disable this metric? I do not see the benefit in doing so
In a 1:1 discussion around this PR, @gamli75 requested the ability to disable this metric.

paul-maidment:
> 100% agree. Metrics manager should serve as a facade for Prometheus metrics, and although it's coupled with events handler, IMO it really shouldn't and we should try to avoid that approach.

I can understand this perspective. I will update this to use the events api directly.

internal/installercache/installercache_test.go
```diff
@@ -149,8 +191,64 @@ var _ = Describe("installer cache", func() {
...
+		Expect(l.releaseID).To(Equal(releaseID))
+		Expect(l.clusterID).To(Equal(clusterId))
+		Expect(l.startTime).ShouldNot(BeZero())
+		Expect(l.cached).To(BeFalse())
+		Expect(l.extractDuration).Should(BeZero())
+		Expect(l.Path).Should(BeEmpty())
+		testMetricsSentOnRelease(l)
+	})
+
+	It("Should have correct metrics on release object if error extracting release", func() {
```
carbonin, line 227: I wouldn't expect this to be a valid case. We'll never call `Release` if `Get` returns an error.

internal/installercache/installercache.go
```diff
@@ -41,53 +48,96 @@ func (fi *fileInfo) Compare(other *fileInfo) bool {
...
+		startTime:                   time.Now(),
+	}
+
 	var workdir, binary, path string
 	var err error
 
 	workdir, binary, path, err = ocRelease.GetReleaseBinaryPath(releaseID, i.cacheDir, ocpVersion)
 	if err != nil {
-		return nil, err
+		return release, err
```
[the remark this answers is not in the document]
paul-maidment: In this case, if the release has partial information (some of the fields have zero values) this tells us about the stage we reached in progress of the release.

In the ".Release(..)" function, we return the state of the release even if there was an error. 

This is intentional.

internal/installercache/installercache_test.go
paul-maidment: I don't see this as an error or an anti pattern as it supports the capture of data for the event and also ensures that we always send an event.

Please see my comment above.

paul-maidment: Please see comment above.

rccrdpccl:
> > @paul-maidment why do we need to enable/disable this metric? I do not see the benefit in doing so
>  In a 1:1 discussion around this PR, @gamli75 requested the ability to disable this metric.

Can we estimate how many events we will have per cluster install on average? We can later confirm hypothesis in stage, but I think volume will not be an issue.

paul-maidment:
> > > @paul-maidment why do we need to enable/disable this metric? I do not see the benefit in doing so
> > > In a 1:1 discussion around this PR, @gamli75 requested the ability to disable this metric.
> 
> Can we estimate how many events we will have per cluster install on average? We can later confirm hypothesis in stage, but I think volume will not be an issue.

Based on the investigation I performed for the presentation I gave before Christmas, approximately 1000 events per week. I don't think this is a massive volume.

paul-maidment:
> > > @paul-maidment why do we need to enable/disable this metric? I do not see the benefit in doing so
> > > In a 1:1 discussion around this PR, @gamli75 requested the ability to disable this metric.
> 
> Can we estimate how many events we will have per cluster install on average? We can later confirm hypothesis in stage, but I think volume will not be an issue.

Approximately 1000 events per week based on what I have observed.

internal/installercache/installercache.go
```diff
@@ -41,53 +48,103 @@ func (fi *fileInfo) Compare(other *fileInfo) bool {
...
+		return
+	}
+	rl.eventsHandler.V2AddMetricsEvent(
+		ctx, &rl.clusterID, nil, nil, "", models.EventSeverityInfo,
+		metricEventInstallerCacheRelease,
+		time.Now(),
+		"release-id", rl.releaseID,
+		"start-time", rl.startTime.Format(time.RFC3339),
+		"cached", rl.cached,
+		"extract-duration", rl.extractDuration,
```
rccrdpccl, line 91: Why start-time and extract-duration? Would we need to work out from event-time the total release process duration? Wouldn't an end date help, especially if we need to calculate which releases are concurrent?

internal/installercache/installercache_test.go
```diff
@@ -52,16 +108,27 @@ var _ = Describe("installer cache", func() {
...
+		Expect(l.clusterID).To(Equal(clusterID))
+		Expect(l.startTime).ShouldNot(BeZero())
+		Expect(l.cached).To(Equal(expectCached))
+		Expect(l.eventsHandler).To(Equal(eventsHandler))
+		if !expectCached {
+			Expect(l.extractDuration).ShouldNot(BeZero())
+		}
+		Expect(l.Path).ShouldNot(BeEmpty())
 
 		time.Sleep(1 * time.Second)
```
rccrdpccl, line 123: I know this is already here, but why is this needed?

paul-maidment: Actually, that's a very interesting point. 

I think what they were trying to do was simulate the passage of time. The way we structure calls to `time.Now()` and such (throughout the application and not just here) does not make them very testable. The same applies to filesystem calls such as `os.Stat(..)`

Ideally, we should wrap calls to "non mockabe" calls in an interface that can be mocked. 

This would allow us to avoid all of this awkwardness and test things like filesystem errors or the advancement of time without hacks like this.

I personally think that we need to address this as part of the installer cache fixes and I was planning on this.
But that it does not need to be done in this PR.

internal/ignition/installmanifests.go
```diff
@@ -173,13 +174,13 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte)
...
-	release, err := g.installerCache.Get(g.installerReleaseImageOverride, g.releaseImageMirror,
-		g.cluster.PullSecret, ocRelease, g.cluster.OpenshiftVersion)
+	release, err := g.installerCache.Get(ctx, g.installerReleaseImageOverride, g.releaseImageMirror,
+		g.cluster.PullSecret, ocRelease, g.cluster.OpenshiftVersion, *g.cluster.ID)
 	if err != nil {
 		return errors.Wrap(err, "failed to get installer path")
 	}
 	//cleanup resources at the end
-	defer release.Release()
+	defer release.Release(ctx)
```
rccrdpccl, line 183: I think naming Release a cleanup method in this context it's confusing (it was already named that way)
Also, I think sending events when successful within this method, and when not from within `installerCache.Get` it's also confusing.
Couldn't we defer `sendReleaseMetricsEvents` before calling `installerCache.Get`? Am I missing anything here?

rccrdpccl: Btw if we do this, as clusterID is needed only for informational purposes, we could avoid changing the `installerCache.Get` signature

internal/installercache/installercache.go
paul-maidment: It feels counter intuitive in a way but the end time is actually the event time. 
We send the event as the very last action.

So start time represents the time the event was requested (via GET) 
event time (on the event itself) is the end time.

rccrdpccl: I see that, do you think that somebody discovering this event (either via kibana or ccx data) would find it helpful to have and endTime, or do you think it can be quickly figured out that event_time is infact endTime as well?

internal/installercache/installercache.go
```diff
@@ -41,53 +48,103 @@ func (fi *fileInfo) Compare(other *fileInfo) bool {
...
+
+func (rl *Release) sendReleaseMetricEvent(ctx context.Context) {
+	if !rl.enableInstallerCacheEvent {
+		return
+	}
+	rl.eventsHandler.V2AddMetricsEvent(
+		ctx, &rl.clusterID, nil, nil, "", models.EventSeverityInfo,
+		metricEventInstallerCacheRelease,
+		time.Now(),
+		"release-id", rl.releaseID,
```
rccrdpccl, line 88: please be consistent with casing. IIRC we use snake case, and can be prefixed using dots (i.e. release.id, release.start_time, etc)

internal/installercache/installercache.go
```diff
@@ -41,53 +48,103 @@ func (fi *fileInfo) Compare(other *fileInfo) bool {
...
+		releaseID:                 releaseID,
+		startTime:                 time.Now(),
+	}
+
 	var workdir, binary, path string
 	var err error
 
 	workdir, binary, path, err = ocRelease.GetReleaseBinaryPath(releaseID, i.cacheDir, ocpVersion)
 	if err != nil {
+		release.sendReleaseMetricEvent(ctx)
```
carbonin, line 126: I don't think sending an event with incomplete data is a good idea.

This event for example will say that the release wasn't cached but we don't really know that at this point.

I don't feel like we have enough failures in releases to gain anything meaningful from these and even then they might skew the data.

If you really think there's value in these error case events lets create a way to filter them out maybe add an error bool or something :shrug:

internal/ignition/installmanifests.go
rccrdpccl: To be fair I am not sure `Done` is much more descriptive than `Release`. I was thinking something like `release.Remove`, `release.Cleanup` or something that allows us to understand what's that for without having to look inside. I would definitely split it from the event sending though

paul-maidment: Cleanup it is then...

internal/installercache/installercache.go
paul-maidment: I'll add clear start and end time metrics to the event, this will make it way clearer for CCX users.

internal/installercache/installercache.go
```diff
@@ -41,32 +46,68 @@ func (fi *fileInfo) Compare(other *fileInfo) bool {
...
+	extractDuration float64
 }
 
-func (rl *Release) Release() {
+// Cleanup is called to signal that the caller has finished using the relase and that resources may be released.
+func (rl *Release) Cleanup(ctx context.Context) {
 	if err := os.Remove(rl.Path); err != nil {
 		logrus.New().WithError(err).Errorf("Failed to delete release link %s", rl.Path)
 	}
+	rl.eventsHandler.V2AddMetricsEvent(
```
rccrdpccl, line 75: Why should we hide sending a message in the Cleanup method?

paul-maidment: I don't see this as hiding, I see this as the release sending stats on cleanup.

paul-maidment:
To clarify, 1000 events per week limit was derived from the following query against production Grafana.

```
SELECT count(*) FROM events WHERE name='cluster_prepare_installation_started' AND event_time >= NOW() - INTERVAL '7 days';
```

For the last 7 days, that count is 804

PR #7170, opened 2025-01-08
MGMT-19599: Extend the operator API to indicate to which bundle the operator belongs

eranco74:
Part of [Assisted Installer Operator marketplace](https://issues.redhat.com/browse/MGMT-19311)
Jira task: https://issues.redhat.com/browse/MGMT-19599

Added support for operator bundles, introducing two initial bundles: OpenShift AI and Virtualization.
New API endpoints to retrieve bundle information:
* List available operator bundles.
* Retrieve operators associated with a specific bundle.
* Update the API documentation (Swagger/OpenAPI) to reflect these changes.

These changes provide a more structured approach to managing and discovering operators, making it easier for users to find and select relevant operators during cluster installation.

---
- Is this PR relying on CI for an e2e test run? for regression
- Should this PR be tested in a specific environment? no
```
$ curl http://10.100.102.50:8090/api/assisted-install/v2/operators/bundles
["virtualization","openshiftai"]
 $ curl http://10.100.102.50:8090/api/assisted-install/v2/operators/bundles/openshiftai
{"name":"openshiftai","operators":["serverless","authorino","odf","servicemesh","nvidia-gpu","pipelines","openshift-ai"]}
$ curl http://10.100.102.50:8090/api/assisted-install/v2/operators/bundles/virtualization
{"name":"virtualization","operators":["cnv","mtv","odf"]}

```-->

- [x] New Feature <!-- new functionality -->
- [ ] Enhancement <!-- refactor, code changes, improvement, that won't add new features -->
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD <!-- Notice that changes for Dockerfiles/Jenkinsfiles aren't tested in CI due to a known bug. -->

- [x] Cloud
- [x] None

## How was this code tested?

<!-- Please, select one or more if needed: -->

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### CODE REVIEW

**Comment by eranco74** [comment]
The e2e-ai-operator-disconnected-capi job is broken
https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-assisted-service-master-edge-e2e-ai-operator-disconnected-capi

**Comment by eranco74** [comment]
> > The e2e-ai-operator-disconnected-capi job is broken https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-assisted-service-master-edge-e2e-ai-operator-disconnected-capi
> 
> you can override it for now or wait for https://issues.redhat.com/browse/MGMT-19358

This https://github.com/openshift/hypershift/pull/5353 should fix it

File: swagger.yaml [hunk 1]
```diff
@@ -2716,6 +2716,55 @@ paths:
...
+      summary: Get operator properties for a bundle
+      description: Retrieves an array of operator properties for the specified bundle.
+      operationId: V2GetBundleOperators
+      parameters:
+        - in: path
+          name: bundle_name
+          description: The name of the bundle.
+          required: true
+          type: string
+          enum: ['virtualization', 'openshiftai']
```
jhernand, line 2751: Making this a enum means that every time that we add a new bundle we will need to modify the swagger.yaml file and regenerate all the code. I think it would be better to make this a plain string, maybe with a regular expression to disallow characters like `/`.

swagger.yaml
```diff
@@ -4082,6 +4143,11 @@ definitions:
     items:
       $ref: '#/definitions/monitored-operator'
 
+  bundle:
+    type: string
+    enum: ['virtualization', 'openshiftai']
+    description: The name of the bundle.
```
jhernand, line 4149: Same comment about using plain strings for bundle names.

swagger.yaml
```diff
@@ -2716,6 +2716,55 @@ paths:
...
+          required: true
+          type: string
+          enum: ['virtualization', 'openshiftai']
+      responses:
+        "200":
+          description: Success
+          schema:
+            type: array
+            items:
+              type: string
```
jhernand, line 2758: This leaves little room for expanding the definitions of bundles without breaking backwards compatibility. Can we instead the define the new `bundle` type as an object. It can then have a field that contains the names of the operators, but we can also extend it in the future if needed.

docs/enhancements/operator-marketplace.md
```diff
@@ -0,0 +1,114 @@
...
+- OpenShift AI (2.37)
+- Nvidia GPU (2.37) - Not currently in the UI
+- Pipelines (2.37) - Not currently in the UI
+- Serverless (2.37) - Not currently in the UI
+- ServiceMesh (2.37) - Not currently in the UI
+
+#### Work In Progress (WIP) Operators
+
+- OpenShift Service Connect (OSC, CoCo)
+- Authorino
```
jhernand, line 81: I believe these two have already been added, they are no longer work in progress.

docs/enhancements/operator-marketplace.md
```diff
@@ -0,0 +1,114 @@
...
+  - **Path**: `/v2/operators/v1/bundles`
+  - **Response**: JSON array of bundles names.
+
+- **List Operators by Bundle Endpoint**:
+
+  - **Method**: GET
+  - **Path**: `/v2/operators/v1/bundles/{bundle_name}`
+  - **Parameters**:
+    - `bundle_name` (path parameter): Identifier of the bundle.
+  - **Response**: JSON array of operators namees.
```
jhernand, line 54: I'd suggest to make the response a list of JSON objects. One of the fields of those objects can be the list of operator names. That way we have room for adding other bundle attributes in the future.

internal/operators/api/api.go
```diff
@@ -55,6 +56,8 @@ type Operator interface {
 	GetPreflightRequirements(ctx context.Context, cluster *common.Cluster) (*models.OperatorHardwareRequirements, error)
 	// GetFeatureSupportID returns the operator unique feature-support ID
 	GetFeatureSupportID() models.FeatureSupportLevelID
+	// GetBundleLabels returns the bundles associated with the operator
+	GetBundleLabels() pq.StringArray
```
jhernand, line 60: I think we should avoid using types from the `pq` library in the contracts: it ties us to that library more than needed. Actually at some point we should migrate from `pq` to `pgx`, as `pq` is no longer maintaned, see https://github.com/lib/pq?tab=readme-ov-file#status . If we want a list of strings then I'd say this should be `[]string`.

It feels a bit strange that in the REST API we represent the relationship of bundles as a list of operator names in the bundle representation, but internally it is a list of labels in operators. Would it be unreasonable to use same representation in the REST API and internally?

eranco74: This will go away as well if we change the bundle to be an object instead of a string.

swagger.yaml
eranco74: I actually like that it's part of the API and not a plain string. it means that this field is validated at the API level.
Generating code is free

swagger.yaml
```diff
@@ -2716,6 +2716,52 @@ paths:
...
+      summary: Get list of avaliable bundles 
+      description: Retrieves a list of avaliable bundles.
+      operationId: V2ListBundles
+      responses:
+        "200":
+          description: Success
+          schema:
+            type: array
+            items:
+              type: string
```
jhernand, line 2732: Should this return a list of `bundle` objects to make it more consistent with the `.../bundles/{bundle_name}` endpoint?

swagger.yaml
```diff
@@ -4041,6 +4087,18 @@ definitions:
...
+        description: List of bundles associated with the operator. Can be empty.
+        items:
+          type: string
+        x-go-custom-tag: gorm:"type:text[]"
+        x-go-type:
+          type: StringArray
+          import:
+            package: github.com/lib/pq
+          hints:
+            noValidation: true
```
jhernand, line 4101: What is the reason to have this list for each cluster? Can't that be inferred from the list of operators of each bundle?

I assume you are doing this to store somewhere the list of bundles that are enabled in a particular cluster, but wouldn't it be more straightforward to have a separate table (similar to the `monitored_operators` table) to store this explicitly? I mean something like this:

```sql
create table enabled_operator_bundles (
  cluster_id text not null,
  bundle_name text not null,
  primary key (cluster_id, bundle_name)
)
```

eranco74: Hmm, the reason we are doing this is because that's what the ticket asked.
"adding a list of bundle labels to each operator"
Moreover, we don't keep track of bundles that are enabled in a particular cluster.
I added the bundles API endpoints because it makes a bit more sense to me (and the UI dev) 
I agree that we could keep a map with the bundle to operator list mapping and it should be just fine (for the bundles API endpoint) but than you won't get the list of bundles the operator belongs to when you get the operator.
Overall, the ticket asked to add bundles list to the operator I did the extra step of exposing what I belive the UI cares about in a dedicated endpoint.

jhernand: OK, I get it, thanks for explaining.

PR #7178, opened 2025-01-12
MGMT-19529: Send metrics about disk usage

paul-maidment:
When the installer cache is used (by calling get) we would like to take a sample of the disk usage in the cache directory. This PR introduces two metrics, one which is a gauge of the percentage of disk space used and the other which is a histogram of usage in GiB

As part of this PR, I have introduced a `filesystem helper` class to mock some filesystem functions.

The broader intent is to do this more universally but for now I am just mocking the functions necessary to gather disk usage information.

The ticket for the broader effort is here https://issues.redhat.com/browse/MGMT-19683

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


internal/installercache/installercache.go
```diff
@@ -99,6 +112,8 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
 	i.Lock()
 	defer i.Unlock()
 
+	i.reportDiskUsageMetric()
```
rccrdpccl, line 115: Shouldn't this be generated on each prometheus scrape rather than here?

paul-maidment: I didn't think so actually.

I think that we are interested in changes in storage caused by the installer cache.

paul-maidment: I actually don't think that gathering this metric on _every_ endpoint call is the way to go.
The directory we are interested in is `WorkDir` and we are only interested in areas where the contents of `WorkDir` may be altered.

This can occur in one of 3 places, as we can see in `cmd/main.go`

1: In the "fallback xattr" client
During a `set`
https://github.com/openshift/assisted-service/blob/master/pkg/s3wrapper/filesystem_xattr_client.go#L52-L59
During a `remove`
https://github.com/openshift/assisted-service/blob/master/pkg/s3wrapper/filesystem_xattr_client.go#L97-L105

2: In the file system based object store
In any location where we already store filesystem usage metrics
https://github.com/openshift/assisted-service/blob/master/pkg/s3wrapper/filesystem.go#L471 for example

3: During the generation of install config
https://github.com/openshift/assisted-service/blob/master/pkg/generator/generator.go#L58

These are the only locations in which we would need to gather the disk metrics as they are the only location in which they can be mutated.

To generically register disk usage with every http request is excessive because

1: Not every request will affect the work dir
2: We know where the work dir content is changed so we can focus on that
3: There is no clean generic mechanism to register this metric for every request, the request based middleware is quite solidly focussed on things like duration of request and is called from libraries that use the metrics.Request interface. None of the callers expect a new method to gather disk stats.

internal/metrics/directory_usage_collector.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"github.com/alecthomas/units"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	directoryUsageGbName                = "assisted_installer_directory_usage_gb"
+	directoryUsageGbDescription         = "The amount of the free space in a directory that has been used in Gb"
```
rccrdpccl, line 12: Wouldn't be best to export it in bytes, and let the UI layer do the transformation to whatever unit it's most comfortable?

internal/metrics/directory_usage_collector.go
```diff
@@ -0,0 +1,60 @@
...
+	directoryUsagePercentageName        = "assisted_installer_directory_usage_percentage"
```
rccrdpccl, line 13: Why used GB and percentage? If we have used and free or used and total (which I think it's easier in this case) I think it's good enough. If % is needed, we can calculate it from PromQL. WDYT?

rccrdpccl:
Nice approach, do you think there is an easy way to add a small test?

internal/metrics/os_disk_stats_helper_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	It("Should retrieve correct stats for an empty directory", func() {
+		usageBytes, freeBytes, err := diskStatsHelper.GetDiskUsage(tempDir)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(usageBytes).To(BeEquivalentTo(0))
+		Expect(freeBytes).To(BeNumerically(">", 0))
+	})
+
+	It("should correctly retrieve stats for a directory with some space used", func() {
+		writeDummyFile(5 * int64(units.GiB))
```
rccrdpccl, line 45: Please reduce the size for the test. As this is real, it is adding a requisite of having 7 GB of free space in order to run the tests. Although in modern system should mainly be OK, I don't see the need for it.

internal/metrics/directory_usage_collector.go
paul-maidment: I ended up discovering a problem with this approach anyway and had to make things closer to your suggestion.

Now we gather directory usage and directory free space. 

The usage comes from the file content of the target directory.
The free space comes from the backing storage.

internal/metrics/directory_usage_collector.go
```diff
@@ -0,0 +1,45 @@
...
+
+	directoryLabelKey = "directory"
+)
+
+type directoryUsageCollector struct {
+	// The directory for which to report disk usage stats
+	directory            string
+	diskStatsHelper      DiskStatsHelper
+	directoryUsageUsedGB *prometheus.Desc
+	directoryUsageFreeGB *prometheus.Desc
```
rccrdpccl, line 21: nit: fsFreeBytes would be consistent with the metric name

internal/metrics/directory_usage_collector.go
```diff
@@ -0,0 +1,45 @@
...
+	fsUsageFreeBytesDescription        = "The amount of free space that is remaining for the filesystem backing this directory"
+
+	directoryLabelKey = "directory"
+)
+
+type directoryUsageCollector struct {
+	// The directory for which to report disk usage stats
+	directory            string
+	diskStatsHelper      DiskStatsHelper
+	directoryUsageUsedGB *prometheus.Desc
```
rccrdpccl, line 20: I think we could remove GB from the name now

internal/metrics/directory_usage_collector.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (c *directoryUsageCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.directoryUsageUsedGB
+	ch <- c.directoryUsageFreeGB
+}
+
+func (c *directoryUsageCollector) Collect(ch chan<- prometheus.Metric) {
+	usedBytes, freeBytes, err := c.diskStatsHelper.GetDiskUsage(c.directory)
+	if err != nil {
+		return
```
rccrdpccl, line 41: I think logging this error might help us find out potential issues

internal/metrics/metricsManager.go
```diff
@@ -119,16 +119,16 @@ type MetricsManager struct {
 	serviceLogicFilesystemUsagePercentage              *prometheus.GaugeVec
 	serviceLogicMonitoredHosts                         *prometheus.GaugeVec
 	serviceLogicMonitoredClusters                      *prometheus.GaugeVec
+	directoryUsageCollectors                           []*directoryUsageCollector
 }
 
 var _ API = &MetricsManager{}
 
-func NewMetricsManager(registry prometheus.Registerer, eventsHandler eventsapi.Handler) *MetricsManager {
+func NewMetricsManager(registry prometheus.Registerer, eventsHandler eventsapi.Handler, diskStatsHelper DiskStatsHelper, directoryUsageMonitorDirs ...string) *MetricsManager {
```
rccrdpccl, line 127: Instead of a variadic function, I think it'd be best to have an options parameter, where we could have a list of dirs in it.

internal/metrics/metricsManager.go
```diff
@@ -282,6 +282,10 @@ func NewMetricsManager(registry prometheus.Registerer, eventsHandler eventsapi.H
 		}, []string{hosts}),
 	}
 
+	for _, directory := range directoryUsageMonitorDirs {
```
rccrdpccl, line 285: I think this job of looping with dirs can be done by the collector directly, WDYT?

rccrdpccl: The advantage would be not registering potentially many collectors

rccrdpccl (changes requested):

internal/metrics/directory_usage_collector.go
```diff
@@ -0,0 +1,69 @@
...
+		directories:              directories,
+		diskStatsHelper:          diskStatsHelper,
+		directoryUsageBytesDescs: directoryUsageBytesDescs,
+		fsFreeByteDescs:          fsFreeByteDescs,
+		log:                      log,
+	}
+}
+
+func (c *directoryUsageCollector) Describe(ch chan<- *prometheus.Desc) {
+	for _, desc := range c.directoryUsageBytesDescs {
```
rccrdpccl, line 51: As `directoryUsageBytesDescs` and `fsFreeByteDescs` are maps we shouldn't rely on its order to do this, as golang randomizes access.

internal/metrics/directory_usage_collector.go
```diff
@@ -0,0 +1,69 @@
...
+		ch <- desc
+	}
+	for _, desc := range c.fsFreeByteDescs {
+		ch <- desc
+	}
+}
+
+func (c *directoryUsageCollector) Collect(ch chan<- prometheus.Metric) {
+	for _, directory := range c.directories {
```
rccrdpccl, line 60: As `directories`  is a map we shouldn't rely on its order to do this, as golang randomizes access.

internal/metrics/directory_usage_collector_test.go
```diff
@@ -0,0 +1,61 @@
...
+		server.Close()
+	})
+
+	var expectMetricValue = func(metric string, value string) {
+		metricParts := strings.Split(metric, "}")
+		metricValue := strings.Trim(metricParts[1], " ")
+		Expect(metricValue).To(Equal(value))
+	}
+
+	It("should collect metrics about disk usage on retrieval of metrics", func() {
```
rccrdpccl, line 46: as we implemented handling multiple directories, we should probably test it

cmd/main.go
```diff
@@ -312,7 +312,8 @@ func main() {
 	eventsHandler := createEventsHandler(crdEventsHandler, db, authzHandler, notificationStream, log)
 
 	prometheusRegistry := prometheus.DefaultRegisterer
-	metricsManager := metrics.NewMetricsManager(prometheusRegistry, eventsHandler)
+	directoryUsageMonitorConfig := metrics.NewDirectoryUsageCollectorConfig([]string{Options.WorkDir})
+	metricsManager := metrics.NewMetricsManager(prometheusRegistry, eventsHandler, metrics.NewOSDiskStatsHelper(), directoryUsageMonitorConfig, log)
```
rccrdpccl, line 316: nit: not very important, however by passing `directoryUsageMonitorConfig` we're hinting to add one parameter each config that might be needed by collectors/metrics registered within the manager. I would've envisioned this as `MetricManagerOptions.DirectoryUsageMonitorConfig`

cmd/main.go
```diff
@@ -312,7 +312,8 @@ func main() {
 	eventsHandler := createEventsHandler(crdEventsHandler, db, authzHandler, notificationStream, log)
 
 	prometheusRegistry := prometheus.DefaultRegisterer
-	metricsManager := metrics.NewMetricsManager(prometheusRegistry, eventsHandler)
+	directoryUsageMonitorConfig := metrics.NewDirectoryUsageCollectorConfig([]string{Options.WorkDir})
```
rccrdpccl, line 315: nit: I do not see any advantages with creating a config with a new function over instantiating a public struct config instead, in my opinion the code should be as simple as possible

internal/metrics/disk_stats_helper.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		if !fileInfo.IsDir() {
+			totalBytes += uint64(fileInfo.Size())
+		}
+		return nil
+	})
+	return totalBytes, err
+}
+
+func (c *OSDiskStatsHelper) GetDiskUsage(directory string) (uint64, uint64, error) {
```
rccrdpccl, line 37: Maybe named return values or a godoc block describing the function would help understand what are those unit64 that we are returning without reading the whole function.

internal/metrics/directory_usage_collector.go
paul-maidment: The order here is not actually important, only that each directory is handled.

I don't see this as a problem in this case

internal/metrics/directory_usage_collector.go
paul-maidment: I'm not depending on order, just expecting a sequence of directories, the order of that sequence is unimportant. Only that they are all handled.

internal/metrics/directory_usage_collector.go
rccrdpccl: How can each metric be associated with its description? Isn't that based on order?

paul-maidment: It's part of the map, directory is the key and there are two maps,

One for the descs regarding disk usage
One for the descs regarding free space

In both cases, looking up the directory by key will give you the correct desc for that directory.

internal/metrics/directory_usage_collector.go
paul-maidment: Also, we are building a map here, the key of the map is the directory, which we address later in the same way.

At no point do we depend on the order of the map.

rccrdpccl (approved):

internal/metrics/disk_stats_helper.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (c *OSDiskStatsHelper) getUsedBytesInDirectory(directory string) (uint64, error) {
+	var totalBytes uint64
+	err := filepath.Walk(directory, func(path string, fileInfo os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !fileInfo.IsDir() {
+			totalBytes += uint64(fileInfo.Size())
```
rccrdpccl, line 30: @paul-maidment Would this count twice hard links?

paul-maidment: I think we need to account for it

https://github.com/openshift/assisted-service/pull/7217/files

PR #7190, opened 2025-01-14
WIP: Prototype installercache fixes

paul-maidment:
The existing installer cache uses a single sync.Mutex to regulate filesystem access to the shared working directory. Also the lifespan of the mutex does not cover the lifespan of possible filesystem mutations.

It would appear that the mutual exclusion is not correctly configured for the use case.

It would be far better to maintain a collection of *sync.Mutex associated with the file path generated for a given release. The collection of mutexes is protected by a single mutex to ensure consistency in the collection of mutexes. The intention being to lock this mutex from the moment we know the path, until the moment the mutex is no longer required.

* If an error occurs, release the path mutex immediately.
* If the calling code (the code that fetches a release) has finished with the release, free up the mutex at this point.
* If the eviction code wishes to evict a specific path, obtain the mutex for this path before taking action.

There was a 'hard link' system in place, the purpose of which appears to be to manage mutual exclusivity around multiple users of a release, I have removed this code and all management of hard links. This is because I don't think we need them, if two users request the same release, one user will wait a short while for release of the lock, I think this is acceptable; The first fetch of a release will extract the release, this may take a little longer, subsequent calls will be faster due to cacheing.

Presently, all release requests in parallel on the same node will be blocked by the single mutex, this is far worse than what I propose to do.

This PR is WIP and has not even been tested yet, it exists as a starting point for changes to come

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/installercache/installercache.go
+++ b/internal/installercache/installercache.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


paul-maidment:
Closing as there is a better approach

PR #7193, opened 2025-01-15
MGMT-19708: Ignore mirrored registries when validating pull secrets

CrystalChun:
When a mirror registry configuration is provided to assisted, we will assume that the registry does not require authentication. This is a better user experience so that the user will not have to provide credentials for a registry if a mirror registry will be used in its place.

I've broken up the commits so that the first commit is for mirror configuration in assisted (globally) and the following commits are for mirror configurations per cluster.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -379,6 +379,15 @@ func main() {
 
 	publicRegistries := map[string]bool{}
 	validations.ParsePublicRegistries(publicRegistries, Options.ValidationsConfig.PublicRegistries)
+	if mirrorRegistriesBuilder.IsMirrorRegistriesConfigured() {
+		var mirrorRegistries []mirrorregistries.RegistriesConf
+		if mirrorRegistries, err = mirrorRegistriesBuilder.ExtractLocationMirrorDataFromRegistries(); err != nil {
+			log.WithError(err).Warnf("failed to parse mirror registries provided to assisted-service, check your mirror registry config map")
+		} else {
+			validations.ParseMirrorRegistries(log, publicRegistries, mirrorRegistries)
+		}
+	}
+
 	pullSecretValidator, err := validations.NewPullSecretValidator(
 		publicRegistries,
 		authHandler,
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -59,6 +59,7 @@ import (
 	"github.com/openshift/assisted-service/pkg/k8sclient"
 	"github.com/openshift/assisted-service/pkg/leader"
 	logutil "github.com/openshift/assisted-service/pkg/log"
+	"github.com/openshift/assisted-service/pkg/mirrorregistries"
 	"github.com/openshift/assisted-service/pkg/ocm"
 	"github.com/openshift/assisted-service/pkg/s3wrapper"
 	"github.com/openshift/assisted-service/pkg/staticnetworkconfig"
@@ -176,7 +177,7 @@ type InstallerInternals interface {
 	GetKnownHostApprovedCounts(clusterID strfmt.UUID) (registered, approved int, err error)
 	HostWithCollectedLogsExists(clusterId strfmt.UUID) (bool, error)
 	GetKnownApprovedHosts(clusterId strfmt.UUID) ([]*common.Host, error)
-	ValidatePullSecret(secret string, username string, releaseImageURL string) error
+	ValidatePullSecret(additionalPublicRegistries []string, secret string, username string, releaseImageURL string) error
 	GetInfraEnvInternal(ctx context.Context, params installer.GetInfraEnvParams) (*common.InfraEnv, error)
 	V2UpdateHostInstallProgressInternal(ctx context.Context, params installer.V2UpdateHostInstallProgressParams) error
 	CreateHostInKubeKeyNamespace(ctx context.Context, kubeKey types.NamespacedName, host *models.Host) error
@@ -288,8 +289,8 @@ func NewBareMetalInventory(
 	}
 }
 
-func (b *bareMetalInventory) ValidatePullSecret(secret string, username string, releaseImageURL string) error {
-	return b.secretValidator.ValidatePullSecret(secret, username, releaseImageURL)
+func (b *bareMetalInventory) ValidatePullSecret(additionalPublicRegistries []string, secret string, username string, releaseImageURL string) error {
+	return b.secretValidator.ValidatePullSecret(additionalPublicRegistries, secret, username, releaseImageURL)
 }
 
 func (b *bareMetalInventory) updatePullSecret(pullSecret string, log logrus.FieldLogger) (string, error) {
@@ -668,8 +669,9 @@ func (b *bareMetalInventory) RegisterClusterInternal(ctx context.Context, kubeKe
 		return nil, common.NewApiError(http.StatusBadRequest, err)
 	}
 
+	mirroredRegistries := extractMirroredRegistriesFromConfig(log, mirrorRegistryConfiguration)
 	pullSecret := swag.StringValue(params.NewClusterParams.PullSecret)
-	err = b.ValidatePullSecret(pullSecret, ocm.UserNameFromContext(ctx), *releaseImage.URL)
+	err = b.ValidatePullSecret(mirroredRegistries, pullSecret, ocm.UserNameFromContext(ctx), *releaseImage.URL)
 	if err != nil {
 		err = errors.Wrap(secretValidationToUserError(err), "pull secret for new cluster is invalid")
 		return nil, common.NewApiError(http.StatusBadRequest, err)
@@ -1881,7 +1883,15 @@ func (b *bareMetalInventory) validateAndUpdateClusterParams(ctx context.Context,
 	log := logutil.FromContext(ctx, b.log)
 
 	if swag.StringValue(params.ClusterUpdateParams.PullSecret) != "" {
-		if err := b.ValidatePullSecret(*params.ClusterUpdateParams.PullSecret, ocm.UserNameFromContext(ctx), cluster.OcpReleaseImage); err != nil {
+		var mirroredRegistries []string
+		if cluster.MirrorRegistryConfiguration != "" {
+			mirrorRegistriesConfig, err := cluster.GetMirrorRegistryConfiguration()
+			if err != nil {
+				b.log.WithError(err).Warnf("failed to get saved cluster mirror registry configuration while updating cluster params")
+			}
+			mirroredRegistries = extractMirroredRegistriesFromConfig(log, mirrorRegistriesConfig)
+		}
+		if err := b.ValidatePullSecret(mirroredRegistries, *params.ClusterUpdateParams.PullSecret, ocm.UserNameFromContext(ctx), cluster.OcpReleaseImage); err != nil {
 			log.WithError(err).Errorf("Pull secret for cluster %s is invalid", params.ClusterID)
 			return installer.V2UpdateClusterParams{}, err
 		}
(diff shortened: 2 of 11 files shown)


internal/cluster/validations/pull_secret_validation.go
```diff
@@ -35,6 +36,16 @@ func ParsePublicRegistries(publicRegistries map[string]bool, publicRegistriesLit
 	}
 }
 
+func ParseMirrorRegistries(publicRegistries map[string]bool, mirrorRegistries []mirrorregistries.RegistriesConf) {
+	for _, mirrorRegistry := range mirrorRegistries {
+		registry, err := ParseRegistry(mirrorRegistry.Location)
+		if err != nil {
+			continue
```
paul-maidment, line 43: Should we be swallowing such an error or should this be logged/reported somehow?

cmd/main.go
```diff
@@ -379,6 +379,15 @@ func main() {
 
 	publicRegistries := map[string]bool{}
 	validations.ParsePublicRegistries(publicRegistries, Options.ValidationsConfig.PublicRegistries)
+	if mirrorRegistriesBuilder.IsMirrorRegistriesConfigured() {
+		mirrorRegistries, err := mirrorRegistriesBuilder.ExtractLocationMirrorDataFromRegistries()
+		if err != nil {
+			log.WithError(err).Warnf("failed to parse mirror registries provided to assisted-service, check your mirror registry config map")
+		} else {
+			validations.ParseMirrorRegistries(publicRegistries, mirrorRegistries)
```
paul-maidment, line 387: I can see that `validations.ParseMirrorRegistries` will parse into publicRegistries. 
What happens if any of the mirror registries fail to parse, should we handle this somehow?

internal/bminventory/inventory.go
```diff
@@ -6751,3 +6782,16 @@ func getDefaultHighAvailabilityAndMasterCountParams(highAvailabilityMode *string
...
 }
+
+func extractMirroredRegistriesFromConfig(log logrus.FieldLogger, mirrorRegistryConfig *common.MirrorRegistryConfiguration) []string {
+	var ignorableRegistries []string
+	mirroredRegistries, err := mirrorregistries.ExtractLocationMirrorDataFromRegistriesFromToml(mirrorRegistryConfig.RegistriesConf)
+	if err != nil {
+		log.WithError(err).Warnf("failed to extract location from saved cluster mirror registry configuration while updating cluster params")
+	}
+	for _, registry := range mirroredRegistries {
+		log.Infof("CRYSTAL adding registry to ignorable %s", registry)
```
paul-maidment, line 6793: This looks like a debug message that needs to be removed.

internal/bminventory/inventory.go
```diff
@@ -6751,3 +6782,16 @@ func getDefaultHighAvailabilityAndMasterCountParams(highAvailabilityMode *string
 	// both are set
 	return highAvailabilityMode, controlPlaneCount
 }
+
+func extractMirroredRegistriesFromConfig(log logrus.FieldLogger, mirrorRegistryConfig *common.MirrorRegistryConfiguration) []string {
+	var ignorableRegistries []string
+	mirroredRegistries, err := mirrorregistries.ExtractLocationMirrorDataFromRegistriesFromToml(mirrorRegistryConfig.RegistriesConf)
+	if err != nil {
+		log.WithError(err).Warnf("failed to extract location from saved cluster mirror registry configuration while updating cluster params")
```
paul-maidment, line 6790: If we fail to extract the mirror registries because of an error, should we swallow this error?

internal/bminventory/inventory.go
```diff
@@ -5319,12 +5346,16 @@ func (b *bareMetalInventory) updateInfraEnvData(infraEnv *common.InfraEnv, param
...
-func (b *bareMetalInventory) validateAndUpdateInfraEnvParams(ctx context.Context, params *installer.UpdateInfraEnvParams) (installer.UpdateInfraEnvParams, error) {
+func (b *bareMetalInventory) validateAndUpdateInfraEnvParams(ctx context.Context, params *installer.UpdateInfraEnvParams, mirrorRegistryConfig *common.MirrorRegistryConfiguration) (installer.UpdateInfraEnvParams, error) {
 
 	log := logutil.FromContext(ctx, b.log)
 
 	if params.InfraEnvUpdateParams.PullSecret != "" {
-		if err := b.ValidatePullSecret(params.InfraEnvUpdateParams.PullSecret, ocm.UserNameFromContext(ctx), ""); err != nil {
+		var mirroredRegistries []string
+		if mirrorRegistryConfig != nil {
+			mirroredRegistries = extractMirroredRegistriesFromConfig(log, mirrorRegistryConfig)
```
paul-maidment, line 5356: What happens if extraction fails?

internal/bminventory/inventory.go
```diff
@@ -5113,15 +5133,33 @@ func (b *bareMetalInventory) UpdateInfraEnvInternal(ctx context.Context, params
...
 			log.WithError(err).Errorf("failed to get infraEnv: %s", params.InfraEnvID)
 			return common.NewApiError(http.StatusNotFound, err)
 		}
+
+		var cluster *common.Cluster
+		clusterId := infraEnv.ClusterID
+		if clusterId != "" {
+			cluster, err = b.GetClusterInternal(ctx, installer.V2GetClusterParams{ClusterID: clusterId})
+			if err != nil {
+				// We don't want to fail here if cluster is not found. It's not responsability of this place
```
paul-maidment, line 5147: But maybe we should log a warning anyway? Something that mentions the cluster not being found when it was expected, this could help with debugging.


I am aware the problem might be caught elsewhere too but anything that can help debugging?

internal/bminventory/inventory.go
```diff
@@ -4868,8 +4883,13 @@ func (b *bareMetalInventory) RegisterInfraEnvInternal(ctx context.Context, kubeK
 			infraEnv.ProxyHash = infraEnvProxyHash
 		}
 
+		var mirroredRegistries []string
+		if mirrorRegistryConfiguration != nil {
+			mirroredRegistries = extractMirroredRegistriesFromConfig(log, mirrorRegistryConfiguration)
```
paul-maidment, line 4888: What happens if extraction fails?

internal/bminventory/inventory.go
```diff
@@ -1881,7 +1887,16 @@ func (b *bareMetalInventory) validateAndUpdateClusterParams(ctx context.Context,
...
 
 	if swag.StringValue(params.ClusterUpdateParams.PullSecret) != "" {
-		if err := b.ValidatePullSecret(*params.ClusterUpdateParams.PullSecret, ocm.UserNameFromContext(ctx), cluster.OcpReleaseImage); err != nil {
+		var mirroredRegistries []string
+		if cluster.MirrorRegistryConfiguration != "" {
+			mirrorRegistriesConfig, err := cluster.GetMirrorRegistryConfiguration()
+			if err != nil {
+				b.log.WithError(err).Warnf("failed to get saved cluster mirror registry configuration while updating cluster params")
+			} else {
+				mirroredRegistries = extractMirroredRegistriesFromConfig(log, mirrorRegistriesConfig)
```
paul-maidment, line 1896: What happens if extraction fails?

internal/bminventory/inventory.go
```diff
@@ -668,8 +669,13 @@ func (b *bareMetalInventory) RegisterClusterInternal(ctx context.Context, kubeKe
 		return nil, common.NewApiError(http.StatusBadRequest, err)
 	}
 
+	var mirroredRegistries []string
+	if mirrorRegistryConfiguration != nil {
+		mirroredRegistries = extractMirroredRegistriesFromConfig(log, mirrorRegistryConfiguration)
```
paul-maidment, line 674: What happens if extraction fails, should we handle this error?

internal/bminventory/inventory.go
```diff
@@ -6751,3 +6782,16 @@ func getDefaultHighAvailabilityAndMasterCountParams(highAvailabilityMode *string
 	// both are set
 	return highAvailabilityMode, controlPlaneCount
 }
+
+func extractMirroredRegistriesFromConfig(log logrus.FieldLogger, mirrorRegistryConfig *common.MirrorRegistryConfiguration) []string {
+	var ignorableRegistries []string
+	mirroredRegistries, err := mirrorregistries.ExtractLocationMirrorDataFromRegistriesFromToml(mirrorRegistryConfig.RegistriesConf)
```
paul-maidment, line 6788: What happens if extraction fails?

internal/bminventory/inventory.go
CrystalChun: This part of the code was already there before, but I agree that a debug message here can help. Will add a log message in, thanks!

internal/bminventory/inventory.go
CrystalChun: Good catch, thanks! Removed it

cmd/main.go
CrystalChun: Good idea, I added a log message inside the parse mirror registries function. What do you think?

internal/cluster/validations/pull_secret_validation.go
CrystalChun: Added a log! What do you think?

internal/bminventory/inventory.go
CrystalChun: Thanks for bringing this up! I think if extraction fails, we should definitely have a log for it (which is done in the `extractMirroredRegistriesFromConfig`). The mirror registry validation is done in another part and should be handled there, should I leave a comment about that here?

Main reason is that I don't think registering a cluster and validating its pull secret should fail at this point because the mirror registry is incorrect. What do you think?

paul-maidment: We should just put a comment here to acknowledge that we are happy this is handled elsewhere

internal/bminventory/inventory.go
paul-maidment: We should just put a comment here to acknowledge that we are happy this is handled elsewhere

internal/bminventory/inventory.go
paul-maidment: As agreed below let's put the comment on extractMirroredRegistriesFromConfig, that way we should only need to do it once.
I think it's reasonable for the reader to look at the function and see the comment.

internal/bminventory/inventory.go
paul-maidment: As agreed below let's put the comment on extractMirroredRegistriesFromConfig, that way we should only need to do it once.
I think it's reasonable for the reader to look at the function and see the comment.

internal/bminventory/inventory.go
paul-maidment: As agreed below let's put the comment on extractMirroredRegistriesFromConfig, that way we should only need to do it once.
I think it's reasonable for the reader to look at the function and see the comment.

internal/bminventory/inventory.go
paul-maidment: Logging within the function makes sense.

There is an interesting edge case here though...

We are within a block of code where we expect that the mirror registry configuration has been set.
So if NO registries come back from this call I could argue that this is unexpected behaviour.

We later rely on this result to decide on the registries that will be ignored, I am not sure whether a problem here could have an unintended side effect?

Perhaps we want to consider this case to be a full error?

internal/bminventory/inventory.go
CrystalChun: Added a comment! What do you think?

internal/bminventory/inventory.go
CrystalChun: Added the comment, thanks!

internal/bminventory/inventory.go
CrystalChun: Added the comment below, thanks!

internal/bminventory/inventory.go
CrystalChun: Added the comment below, thanks!

internal/bminventory/inventory.go
CrystalChun: Thanks for the review and discussion, Paul!

Discussed with Paul offline:

- If there's a failure with parsing mirror registries here, then it should not fail at this point in the function. 
    - The failure to parse mirror registries will also fail the validation of the pull secret. 
    - So the failure will show up to the user.
- Also if there's a failure to parse the mirror registries, it should also fail elsewhere. 
    - Want to verify with @eliorerz that this is the case too.

Will also add the ignored registries passed in to the logging to help debug.

CrystalChun: Thanks Paul! I've also added the ignored registries in the error message during pull secret validation so we can more easily see what registries were passed in.

CrystalChun:
/cherry-pick release-ocm-2.12 release-ocm-2.11 release-ocm-2.10

PR #7197, opened 2025-01-16
MGMT-19100: Allow install to the existing root filesystem

carbonin:
In order to allow for new CAPI install methods this PR enables the installer to write to a new stateroot on the existing root filesystem as opposed to installing to an entire device using `coreos-installer`.

This PR creates an environment variable (`INSTALL_TO_EXISTING_ROOT`) which when set will fetch the coreos image content from the selected release and provide it to the installer. This (along with the changes in https://github.com/openshift/assisted-installer/pull/1003) will trigger the installer to write the image to a new stateroot on the existing disk rather than installing to a target device.

This functionality is disabled by default and should only be used when the host is not running in the live-iso environment.

In order to enable (and succeed when using) the new install flow the following must be done:
1. Set `INSTALL_TO_EXISTING_ROOT` to `true` in the assisted-service deployment
2. Boot a host using a disk image and provide the discovery ignition to that host using a platform metadata service
3. Proceed with installation as normal

In the future (https://issues.redhat.com/browse/MGMT-19102) we will detect if the host is booted using a live-iso or not and use the corresponding install strategy, but this PR is just a first pass to get CAPI integration work unstuck.

This PR can be merged now, but the full functionality also depends on the following PRs:
- https://github.com/openshift/assisted-installer/pull/1003
- [todo] agent PR

https://issues.redhat.com/browse/MGMT-19100

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Manually deployed to a dev-scripts environment and succeeded in an install using a disk image rather than the live-iso.

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

- [ ] This change does not require a documentation update (docstring, `docs`, README, etc) - Will follow up with docs

## Reviewers Checklist

### DIFF
--- a/api/vendor/github.com/openshift/assisted-service/models/install_cmd_request.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/install_cmd_request.go
@@ -38,6 +38,9 @@ type InstallCmdRequest struct {
 	// Pattern: ^(([a-zA-Z0-9\-\.]+)(:[0-9]+)?\/)?[a-z0-9\._\-\/@]+[?::a-zA-Z0-9_\-.]+$
 	ControllerImage *string `json:"controller_image"`
 
+	// CoreOS container image to use if installing to the local device
+	CoreosImage string `json:"coreos_image,omitempty"`
+
 	// List of disks to format
 	DisksToFormat []string `json:"disks_to_format"`
 
--- a/client/vendor/github.com/openshift/assisted-service/models/install_cmd_request.go
+++ b/client/vendor/github.com/openshift/assisted-service/models/install_cmd_request.go
@@ -38,6 +38,9 @@ type InstallCmdRequest struct {
 	// Pattern: ^(([a-zA-Z0-9\-\.]+)(:[0-9]+)?\/)?[a-z0-9\._\-\/@]+[?::a-zA-Z0-9_\-.]+$
 	ControllerImage *string `json:"controller_image"`
 
+	// CoreOS container image to use if installing to the local device
+	CoreosImage string `json:"coreos_image,omitempty"`
+
 	// List of disks to format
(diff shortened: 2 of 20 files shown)


rccrdpccl:
> This PR creates an environment variable (`INSTALL_TO_EXISTING_ROOT`) which when set will fetch the coreos image content from the selected release and provide it to the installer. This (along with the changes in [openshift/assisted-installer#1003](https://github.com/openshift/assisted-installer/pull/1003)) will trigger the installer to write the image to a new stateroot on the existing disk rather than installing to a target device.

Would you see any value in making this parameter "per cluster" rather than per AI instance? This way we could run both types of install flow with the same AI instance. Of course in such case in order for it to be useful we'd need to advertise the userdata address, so maybe could be a later iteration?

carbonin:
@rccrdpccl I address this later in the description and with a separate task.
This PR is just a first pass to get yall unstuck.

>In the future (https://issues.redhat.com/browse/MGMT-19102) we will detect if the host is booted using a live-iso or not and use the corresponding install strategy, but this PR is just a first pass to get CAPI integration work unstuck.

rccrdpccl:
> @rccrdpccl I address this later in the description and with a separate task. This PR is just a first pass to get yall unstuck.
> 
> > In the future (https://issues.redhat.com/browse/MGMT-19102) we will detect if the host is booted using a live-iso or not and use the corresponding install strategy, but this PR is just a first pass to get CAPI integration work unstuck.

Great, I didn't realize the automatic flow detection was effectively the same (or even better) of what I proposed. In any case totally agree to tackle it separately

PR #7205, opened 2025-01-20
MGMT-14453: Make sure installercache can actually run Mutex correctly

paul-maidment:
Presently we instantiate a new `installercache.Installers` instance per execution of the code generator. This renders any mutex handling pointless as we will never have any mutex conflict over a single request.

This PR moves the instantiation into the main and effectively makes it a singleton so that mutexes will function as intended.

This is the root cause behind a couple of reported issues

https://issues.redhat.com/browse/MGMT-14452 - Installer cache removes in-used cached image when out of space https://issues.redhat.com/browse/MGMT-14453 - INSTALLER_CACHE_CAPACITY small value cause to assisted-service crash

Both of these issues are reproducible on master and do not reproduce when running on code based on this PR.

If the cache size is set to a low value such as 1 then the current release will be deleted once the caller has finished with the release.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -39,6 +39,7 @@ import (
 	"github.com/openshift/assisted-service/internal/ignition"
 	"github.com/openshift/assisted-service/internal/infraenv"
 	installcfg "github.com/openshift/assisted-service/internal/installcfg/builder"
+	"github.com/openshift/assisted-service/internal/installercache"
 	internaljson "github.com/openshift/assisted-service/internal/json"
 	"github.com/openshift/assisted-service/internal/manifests"
 	"github.com/openshift/assisted-service/internal/metrics"
@@ -476,7 +477,10 @@ func main() {
 	failOnError(err, "failed to create valid bm config S3 endpoint URL from %s", Options.BMConfig.S3EndpointURL)
 	Options.BMConfig.S3EndpointURL = newUrl
 
-	generator := generator.New(log, objectHandler, Options.GeneratorConfig, Options.WorkDir, providerRegistry, manifestsApi, eventsHandler)
+	installGeneratorWorkDir := filepath.Join(Options.WorkDir, "install-config-generate")
+	installerCacheDir := filepath.Join(installGeneratorWorkDir, "installercache")
+	installerCache := installercache.New(installerCacheDir, 5, eventsHandler, log)
+	generator := generator.New(log, objectHandler, Options.GeneratorConfig, Options.WorkDir, providerRegistry, manifestsApi, eventsHandler, installGeneratorWorkDir, installerCache)
 	var crdUtils bminventory.CRDUtils
 	if ctrlMgr != nil {
 		crdUtils = controllers.NewCRDUtils(ctrlMgr.GetClient(), hostApi)
--- a/internal/ignition/installmanifests.go
+++ b/internal/ignition/installmanifests.go
@@ -85,7 +85,6 @@ type installerGenerator struct {
 	cluster                       *common.Cluster
 	releaseImage                  string
 	releaseImageMirror            string
-	installerDir                  string
 	serviceCACert                 string
 	encodedDhcpFileContents       string
 	s3Client                      s3wrapper.API
@@ -110,16 +109,15 @@ var fileNames = [...]string{
 }
 
 // NewGenerator returns a generator that can generate ignition files
-func NewGenerator(workDir string, installerDir string, cluster *common.Cluster, releaseImage string, releaseImageMirror string,
+func NewGenerator(workDir string, cluster *common.Cluster, releaseImage string, releaseImageMirror string,
 	serviceCACert string, installInvoker string, s3Client s3wrapper.API, log logrus.FieldLogger, providerRegistry registry.ProviderRegistry,
-	installerReleaseImageOverride, clusterTLSCertOverrideDir string, storageCapacityLimit int64, manifestApi manifestsapi.ManifestsAPI, eventsHandler eventsapi.Handler) Generator {
+	installerReleaseImageOverride, clusterTLSCertOverrideDir string, manifestApi manifestsapi.ManifestsAPI, eventsHandler eventsapi.Handler, installerCache *installercache.Installers) Generator {
 	return &installerGenerator{
 		cluster:                       cluster,
 		log:                           log,
 		releaseImage:                  releaseImage,
 		releaseImageMirror:            releaseImageMirror,
 		workDir:                       workDir,
-		installerDir:                  installerDir,
 		serviceCACert:                 serviceCACert,
 		s3Client:                      s3Client,
 		enableMetal3Provisioning:      true,
@@ -127,7 +125,7 @@ func NewGenerator(workDir string, installerDir string, cluster *common.Cluster, 
 		providerRegistry:              providerRegistry,
 		installerReleaseImageOverride: installerReleaseImageOverride,
 		clusterTLSCertOverrideDir:     clusterTLSCertOverrideDir,
-		installerCache:                installercache.New(installerDir, storageCapacityLimit, eventsHandler, log),
+		installerCache:                installerCache,
 		manifestApi:                   manifestApi,
 	}
 }
--- a/internal/ignition/installmanifests_test.go
+++ b/internal/ignition/installmanifests_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openshift/assisted-service/internal/constants"
 	eventsapi "github.com/openshift/assisted-service/internal/events/api"
 	"github.com/openshift/assisted-service/internal/host/hostutil"
+	"github.com/openshift/assisted-service/internal/installercache"
 	manifestsapi "github.com/openshift/assisted-service/internal/manifests/api"
 	"github.com/openshift/assisted-service/internal/network"
 	"github.com/openshift/assisted-service/models"
@@ -78,23 +79,25 @@ var _ = Describe("Bootstrap Ignition Update", func() {
 	  }`
 
 	var (
-		err           error
-		examplePath   string
-		db            *gorm.DB
-		dbName        string
-		bmh           *bmh_v1alpha1.BareMetalHost
-		config        *config_32_types.Config
-		mockS3Client  *s3wrapper.MockAPI
-		workDir       string
-		cluster       *common.Cluster
-		ctrl          *gomock.Controller
-		manifestsAPI  *manifestsapi.MockManifestsAPI
-		eventsHandler eventsapi.Handler
+		err            error
+		examplePath    string
+		db             *gorm.DB
+		dbName         string
+		bmh            *bmh_v1alpha1.BareMetalHost
+		config         *config_32_types.Config
+		mockS3Client   *s3wrapper.MockAPI
+		workDir        string
+		cluster        *common.Cluster
+		ctrl           *gomock.Controller
+		manifestsAPI   *manifestsapi.MockManifestsAPI
+		eventsHandler  eventsapi.Handler
+		installerCache *installercache.Installers
 	)
 
 	BeforeEach(func() {
 		// setup temp workdir
 		workDir, err = os.MkdirTemp("", "bootstrap-ignition-update-test-")
+		installerCache = installercache.New(filepath.Join(workDir, "install-config-generate", "installercache"), 5, eventsHandler, logrus.New())
 		Expect(err).NotTo(HaveOccurred())
 		examplePath = filepath.Join(workDir, "example1.ign")
 		var err1 error
@@ -113,7 +116,7 @@ var _ = Describe("Bootstrap Ignition Update", func() {
 			},
 		}
 		db, dbName = common.PrepareTestDB()
-		g := NewGenerator(workDir, "", cluster, "", "", "", "", mockS3Client, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+		g := NewGenerator(workDir, cluster, "", "", "", "", mockS3Client, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 		Expect(g.updateBootstrap(context.Background(), examplePath)).To(Succeed())
 
@@ -241,16 +244,17 @@ SV4bRR9i0uf+xQ/oYRvugQ25Q7EahO5hJIWRf4aULbk36Zpw3++v2KFnF26zqwB6
 -----END CERTIFICATE-----`
 
 	var (
-		masterPath    string
-		workerPath    string
-		caCertPath    string
-		dbName        string
-		db            *gorm.DB
-		cluster       *common.Cluster
-		workDir       string
-		ctrl          *gomock.Controller
-		manifestsAPI  *manifestsapi.MockManifestsAPI
-		eventsHandler eventsapi.Handler
+		masterPath     string
+		workerPath     string
+		caCertPath     string
+		dbName         string
+		db             *gorm.DB
+		cluster        *common.Cluster
+		workDir        string
+		ctrl           *gomock.Controller
+		manifestsAPI   *manifestsapi.MockManifestsAPI
+		eventsHandler  eventsapi.Handler
+		installerCache *installercache.Installers
 	)
 
 	BeforeEach(func() {
@@ -274,6 +278,7 @@ SV4bRR9i0uf+xQ/oYRvugQ25Q7EahO5hJIWRf4aULbk36Zpw3++v2KFnF26zqwB6
 		ctrl = gomock.NewController(GinkgoT())
 		manifestsAPI = manifestsapi.NewMockManifestsAPI(ctrl)
 		eventsHandler = eventsapi.NewMockHandler(ctrl)
+		installerCache = installercache.New(filepath.Join(workDir, "install-config-generate", "installercache"), 5, eventsHandler, logrus.New())
 	})
 
 	AfterEach(func() {
@@ -283,7 +288,7 @@ SV4bRR9i0uf+xQ/oYRvugQ25Q7EahO5hJIWRf4aULbk36Zpw3++v2KFnF26zqwB6
 
 	Describe("update ignitions", func() {
 		It("with ca cert file", func() {
-			g := NewGenerator(workDir, "", cluster, "", "", caCertPath, "", nil, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+			g := NewGenerator(workDir, cluster, "", "", caCertPath, "", nil, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 			err := g.updateIgnitions()
 			Expect(err).NotTo(HaveOccurred())
@@ -305,7 +310,7 @@ SV4bRR9i0uf+xQ/oYRvugQ25Q7EahO5hJIWRf4aULbk36Zpw3++v2KFnF26zqwB6
 			Expect(file.Path).To(Equal(common.HostCACertPath))
 		})
 		It("with no ca cert file", func() {
-			g := NewGenerator(workDir, "", cluster, "", "", "", "", nil, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+			g := NewGenerator(workDir, cluster, "", "", "", "", nil, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 			err := g.updateIgnitions()
 			Expect(err).NotTo(HaveOccurred())
@@ -324,7 +329,7 @@ SV4bRR9i0uf+xQ/oYRvugQ25Q7EahO5hJIWRf4aULbk36Zpw3++v2KFnF26zqwB6
 		})
 		Context("DHCP generation", func() {
 			It("Definitions only", func() {
-				g := NewGenerator(workDir, "", cluster, "", "", "", "", nil, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+				g := NewGenerator(workDir, cluster, "", "", "", "", nil, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 				g.encodedDhcpFileContents = "data:,abc"
 				err := g.updateIgnitions()
@@ -342,7 +347,7 @@ SV4bRR9i0uf+xQ/oYRvugQ25Q7EahO5hJIWRf4aULbk36Zpw3++v2KFnF26zqwB6
 			})
 		})
 		It("Definitions+leases", func() {
-			g := NewGenerator(workDir, "", cluster, "", "", "", "", nil, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+			g := NewGenerator(workDir, cluster, "", "", "", "", nil, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 			g.encodedDhcpFileContents = "data:,abc"
 			cluster.ApiVipLease = "api"
@@ -429,14 +434,15 @@ var _ = Describe("createHostIgnitions", func() {
 		}`
 
 	var (
-		dbName        string
-		db            *gorm.DB
-		mockS3Client  *s3wrapper.MockAPI
-		cluster       *common.Cluster
-		ctrl          *gomock.Controller
-		workDir       string
-		manifestsAPI  *manifestsapi.MockManifestsAPI
-		eventsHandler eventsapi.Handler
+		dbName         string
+		db             *gorm.DB
+		mockS3Client   *s3wrapper.MockAPI
+		cluster        *common.Cluster
+		ctrl           *gomock.Controller
+		workDir        string
+		manifestsAPI   *manifestsapi.MockManifestsAPI
+		eventsHandler  eventsapi.Handler
+		installerCache *installercache.Installers
 	)
 
 	BeforeEach(func() {
@@ -457,6 +463,7 @@ var _ = Describe("createHostIgnitions", func() {
 		manifestsAPI = manifestsapi.NewMockManifestsAPI(ctrl)
 		eventsHandler = eventsapi.NewMockHandler(ctrl)
 		cluster = testCluster()
+		installerCache = installercache.New(filepath.Join(workDir, "install-config-generate", "installercache"), 5, eventsHandler, logrus.New())
 	})
 
 	AfterEach(func() {
@@ -492,7 +499,7 @@ var _ = Describe("createHostIgnitions", func() {
 				host.ID = &id
 			}
 
-			g := NewGenerator(workDir, "", cluster, "", "", "", "", nil, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+			g := NewGenerator(workDir, cluster, "", "", "", "", nil, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 			err := g.createHostIgnitions()
 			Expect(err).NotTo(HaveOccurred())
@@ -546,7 +553,7 @@ var _ = Describe("createHostIgnitions", func() {
 				host.ID = &id
 			}
 
-			g := NewGenerator(workDir, "", cluster, "", "", "", "", nil, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+			g := NewGenerator(workDir, cluster, "", "", "", "", nil, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 			err := g.createHostIgnitions()
 			Expect(err).NotTo(HaveOccurred())
@@ -590,7 +597,7 @@ var _ = Describe("createHostIgnitions", func() {
 				host.ID = &id
 			}
 
-			g := NewGenerator(workDir, "", cluster, "", "", "", "", nil, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+			g := NewGenerator(workDir, cluster, "", "", "", "", nil, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 			err := g.createHostIgnitions()
 			Expect(err).NotTo(HaveOccurred())
@@ -638,7 +645,7 @@ var _ = Describe("createHostIgnitions", func() {
 			host.ID = &id
 		}
 
-		g := NewGenerator(workDir, "", cluster, "", "", "", "", nil, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+		g := NewGenerator(workDir, cluster, "", "", "", "", nil, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 		g.nodeIpAllocations = make(map[strfmt.UUID]*network.NodeIpAllocation)
 		for i, h := range cluster.Hosts {
 			g.nodeIpAllocations[*h.ID] = &network.NodeIpAllocation{
@@ -679,7 +686,7 @@ var _ = Describe("createHostIgnitions", func() {
 			IgnitionConfigOverrides: `{"ignition": {"version": "3.2.0"}, "storage": {"files": [{"path": "/tmp/example", "contents": {"source": "data:text/plain;base64,aGVscGltdHJhcHBlZGluYXN3YWdnZXJzcGVj"}}]}}`,
 		}}
 
-		g := NewGenerator(workDir, "", cluster, "", "", "", "", nil, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+		g := NewGenerator(workDir, cluster, "", "", "", "", nil, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 		err := g.createHostIgnitions()
 		Expect(err).NotTo(HaveOccurred())
@@ -750,7 +757,7 @@ spec:
 				MachineConfigPoolName: "infra",
 			}}
 
-			g := NewGenerator(workDir, "", cluster, "", "", "", "", mockS3Client, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+			g := NewGenerator(workDir, cluster, "", "", "", "", mockS3Client, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 			mockS3Client.EXPECT().ListObjectsByPrefixWithMetadata(gomock.Any(), filepath.Join(clusterID.String(), constants.ManifestFolder, models.ManifestFolderOpenshift)).Return([]s3wrapper.ObjectInfo{{Path: "mcp.yaml"}}, nil).Times(1)
 			mockS3Client.EXPECT().ListObjectsByPrefixWithMetadata(gomock.Any(), filepath.Join(clusterID.String(), constants.ManifestFolder, models.ManifestFolderManifests)).Times(1)
 			mockS3Client.EXPECT().Download(gomock.Any(), gomock.Any()).Return(io.NopCloser(strings.NewReader(mcp)), int64(0), nil)
@@ -776,7 +783,7 @@ spec:
 				MachineConfigPoolName: "infra",
 			}}
 
-			g := NewGenerator(workDir, "", cluster, "", "", "", "", mockS3Client, logrus.New(), nil, "", "", 5, manifestsAPI, eventsHandler).(*installerGenerator)
+			g := NewGenerator(workDir, cluster, "", "", "", "", mockS3Client, logrus.New(), nil, "", "", manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 			mockS3Client.EXPECT().ListObjectsByPrefixWithMetadata(gomock.Any(), filepath.Join(clusterID.String(), constants.ManifestFolder, models.ManifestFolderOpenshift)).Return([]s3wrapper.ObjectInfo{{Path: "mcp.yaml"}}, nil).Times(1)
 			mockS3Client.EXPECT().ListObjectsByPrefixWithMetadata(gomock.Any(), filepath.Join(clusterID.String(), constants.ManifestFolder, models.ManifestFolderManifests)).Times(1)
 			mockS3Client.EXPECT().Download(gomock.Any(), gomock.Any()).Return(io.NopCloser(strings.NewReader(mc)), int64(0), nil)
@@ -1582,10 +1589,11 @@ var _ = Describe("Set kubelet node ip", func() {
 
 var _ = Describe("Bare metal host generation", func() {
 	var (
-		workDir       string
-		ctrl          *gomock.Controller
-		manifestsAPI  *manifestsapi.MockManifestsAPI
-		eventsHandler eventsapi.Handler
+		workDir        string
+		ctrl           *gomock.Controller
+		manifestsAPI   *manifestsapi.MockManifestsAPI
+		eventsHandler  eventsapi.Handler
+		installerCache *installercache.Installers
 	)
 
 	BeforeEach(func() {
@@ -1595,6 +1603,7 @@ var _ = Describe("Bare metal host generation", func() {
 		ctrl = gomock.NewController(GinkgoT())
 		manifestsAPI = manifestsapi.NewMockManifestsAPI(ctrl)
 		eventsHandler = eventsapi.NewMockHandler(ctrl)
+		installerCache = installercache.New(filepath.Join(workDir, "install-config-generate", "installercache"), 5, eventsHandler, logrus.New())
 	})
 
 	AfterEach(func() {
@@ -1607,7 +1616,6 @@ var _ = Describe("Bare metal host generation", func() {
 			// Create the generator:
 			generator := NewGenerator(
 				workDir,
-				"",
 				testCluster(),
 				"",
 				"",
@@ -1618,9 +1626,9 @@ var _ = Describe("Bare metal host generation", func() {
 				nil,
 				"",
 				"",
-				5,
 				manifestsAPI,
 				eventsHandler,
+				installerCache,
 			).(*installerGenerator)
 
 			// The default host inventory used by these tests has two NICs, each with
@@ -1680,14 +1688,15 @@ var _ = Describe("Bare metal host generation", func() {
 
 var _ = Describe("Import Cluster TLS Certs for ephemeral installer", func() {
 	var (
-		certDir       string
-		dbName        string
-		db            *gorm.DB
-		cluster       *common.Cluster
-		workDir       string
-		ctrl          *gomock.Controller
-		manifestsAPI  *manifestsapi.MockManifestsAPI
-		eventsHandler eventsapi.Handler
+		certDir        string
+		dbName         string
+		db             *gorm.DB
+		cluster        *common.Cluster
+		workDir        string
+		ctrl           *gomock.Controller
+		manifestsAPI   *manifestsapi.MockManifestsAPI
+		eventsHandler  eventsapi.Handler
+		installerCache *installercache.Installers
 	)
 
 	certFiles := []string{"test-cert.crt", "test-cert.key"}
@@ -1718,6 +1727,7 @@ var _ = Describe("Import Cluster TLS Certs for ephemeral installer", func() {
 		ctrl = gomock.NewController(GinkgoT())
 		manifestsAPI = manifestsapi.NewMockManifestsAPI(ctrl)
 		eventsHandler = eventsapi.NewMockHandler(ctrl)
+		installerCache = installercache.New(filepath.Join(workDir, "install-config-generate", "installercache"), 5, eventsHandler, logrus.New())
 	})
 
 	AfterEach(func() {
@@ -1726,7 +1736,7 @@ var _ = Describe("Import Cluster TLS Certs for ephemeral installer", func() {
 	})
 
 	It("copies the tls cert files", func() {
-		g := NewGenerator(workDir, "", cluster, "", "", "", "", nil, logrus.New(), nil, "", certDir, 5, manifestsAPI, eventsHandler).(*installerGenerator)
+		g := NewGenerator(workDir, cluster, "", "", "", "", nil, logrus.New(), nil, "", certDir, manifestsAPI, eventsHandler, installerCache).(*installerGenerator)
 
 		err := g.importClusterTLSCerts(context.Background())
 		Expect(err).NotTo(HaveOccurred())
--- a/internal/installercache/installercache.go
+++ b/internal/installercache/installercache.go
@@ -16,6 +16,7 @@ import (
 	"github.com/openshift/assisted-service/models"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/google/uuid"
 )
 
 var (
@@ -135,8 +136,7 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
 	// return a new hard link to the binary file
 	// the caller should delete the hard link when
 	// it finishes working with the file
-	link := filepath.Join(workdir, "ln_"+fmt.Sprint(time.Now().Unix())+
-		"_"+binary)
+	link := filepath.Join(workdir, fmt.Sprintf("ln_%d_%s_%s", time.Now().Unix(), uuid.NewString(), binary))
 	err = os.Link(path, link)
 	if err != nil {
 		return nil, errors.Wrap(err, fmt.Sprintf("Failed to create hard link to binary %s", path))
--- a/pkg/generator/generator.go
+++ b/pkg/generator/generator.go
@@ -3,11 +3,11 @@ package generator
 import (
 	"context"
 	"os"
-	"path/filepath"
 
 	"github.com/openshift/assisted-service/internal/common"
 	eventsapi "github.com/openshift/assisted-service/internal/events/api"
 	"github.com/openshift/assisted-service/internal/ignition"
+	"github.com/openshift/assisted-service/internal/installercache"
 	manifestsapi "github.com/openshift/assisted-service/internal/manifests/api"
 	"github.com/openshift/assisted-service/internal/provider/registry"
 	logutil "github.com/openshift/assisted-service/pkg/log"
@@ -39,18 +39,20 @@ type installGenerator struct {
 	providerRegistry registry.ProviderRegistry
 	manifestApi      manifestsapi.ManifestsAPI
 	eventsHandler    eventsapi.Handler
+	installerCache   *installercache.Installers
 }
 
 func New(log logrus.FieldLogger, s3Client s3wrapper.API, cfg Config, workDir string,
-	providerRegistry registry.ProviderRegistry, manifestApi manifestsapi.ManifestsAPI, eventsHandler eventsapi.Handler) *installGenerator {
+	providerRegistry registry.ProviderRegistry, manifestApi manifestsapi.ManifestsAPI, eventsHandler eventsapi.Handler, installGeneratorWorkdir string, installerCache *installercache.Installers) *installGenerator {
 	return &installGenerator{
 		Config:           cfg,
 		log:              log,
 		s3Client:         s3Client,
-		workDir:          filepath.Join(workDir, "install-config-generate"),
+		workDir:          installGeneratorWorkdir,
 		providerRegistry: providerRegistry,
 		manifestApi:      manifestApi,
 		eventsHandler:    eventsHandler,
+		installerCache:   installerCache,
 	}
 }
 
@@ -71,15 +73,13 @@ func (k *installGenerator) GenerateInstallConfig(ctx context.Context, cluster co
 		}
 	}()
 
-	installerCacheDir := filepath.Join(k.workDir, "installercache")
-
 	// runs openshift-install to generate ignition files, then modifies them as necessary
 	var generator ignition.Generator
 	if k.Config.DummyIgnition {
 		generator = ignition.NewDummyGenerator(clusterWorkDir, &cluster, k.s3Client, log)
 	} else {
-		generator = ignition.NewGenerator(clusterWorkDir, installerCacheDir, &cluster, releaseImage, k.Config.ReleaseImageMirror,
-			k.Config.ServiceCACertPath, k.Config.InstallInvoker, k.s3Client, log, k.providerRegistry, installerReleaseImageOverride, k.Config.ClusterTLSCertOverrideDir, k.InstallerCacheCapacity, k.manifestApi, k.eventsHandler)
+		generator = ignition.NewGenerator(clusterWorkDir, &cluster, releaseImage, k.Config.ReleaseImageMirror,
+			k.Config.ServiceCACertPath, k.Config.InstallInvoker, k.s3Client, log, k.providerRegistry, installerReleaseImageOverride, k.Config.ClusterTLSCertOverrideDir, k.manifestApi, k.eventsHandler, k.installerCache)
 	}
 	err = generator.Generate(ctx, cfg)
 	if err != nil {


carbonin:
@paul-maidment there's no reason to constantly retest failing optional tests unless you think the specific failure is relevant to your changes.

Looking at that test's history it seems to be failing much more frequently than it's passing so I wouldn't worry about it.

carbonin:
Still working on a full review, but one thing I'm noticing is that you have a _lot_ of log info messages that I'm not sure are going to be super interesting once this is all up and running.

I'd consider whether we will really need to see these every time and either remove some or change them to debug.

I just don't want to clutter up the review with a bunch of comments saying "consider changing this to debug or removing it"

internal/installercache/installercache.go
```diff
@@ -183,33 +291,34 @@ func (i *Installers) evict() {
...
-	//in ignition.go did not succeeded as expected
-	for idx := 0; idx < len(links); idx++ {
-		finfo := links[idx]
-		//Allow a grace period of 5 minutes from the link creation time
-		//to ensure the link is not being used.
-		grace := time.Now().Add(DeleteGracePeriod).Unix()
-		if finfo.info.ModTime().Unix() < grace {
-			os.Remove(finfo.path)
-		}
-	}
```
carbonin, line 204: Just FYI it's tough to review stuff like this where you both move and slightly change something.
Given that this PR is already pretty big and the changes you made in this case really a no-op I think it would have been better to just not change this at all.

Now that you've done it, it's fine, but just wanted to make you aware.

internal/installercache/installercache.go
```diff
@@ -183,33 +291,34 @@ func (i *Installers) evict() {
...
-		finfo, _ := files.Pop()
-		totalSize -= finfo.info.Size()
-		//remove the file
-		if err := i.evictFile(finfo.path); err != nil {
-			i.log.WithError(err).Errorf("failed to evict file %s", finfo.path)
+	// delete the oldest file if necessary
+	// prefer files without hardlinks first because
+	// 1: files without hardlinks are likely to be least recently used anyway
+	// 2: files without hardlinks will immediately free up storage
+	queues := i.splitQueueOnHardLinks(files)
```
carbonin, line 310: If we're using the actual space used on disk as our metric here why consider evicting files with hardlinks at all?
Removing one of those won't get us any space back.

The amount of extra engineering here is the reason I suggested we use number of releases rather than the actual size for this. This extra complexity is bound to make this fragile, but I understand if we don't want to change directions now.

internal/installercache/installercache_test.go
```diff
@@ -119,19 +145,162 @@ var _ = Describe("installer cache", func() {
 		}
 		Expect(l.Path).ShouldNot(BeEmpty())
 
-		time.Sleep(1 * time.Second)
+		time.Sleep(10 * time.Millisecond)
```
carbonin, line 148: Why is this sleep necessary? Can you add a comment explaining it if it really is needed?

internal/installercache/installercache_test.go
```diff
@@ -119,19 +145,162 @@ var _ = Describe("installer cache", func() {
...
+	type launchParams struct {
+		releaseID string
+		version   string
+		clusterID strfmt.UUID
+	}
+
+	// runParallelTest launches a batch of fetches from the installer cache in order to simulate multiple requests to the same node
+	// releases are automatically cleaned up as they are gathered
+	// returns the first error encountered or nil if no error encountered.
+	runParallelTest := func(params []launchParams) error {
```
carbonin, line 163: I'm not sure this kind of test belongs in unit-tests.
At best if/when it fails it will give inconsistent results.

I totally understand if you want to write these tests for yourself, but I'm not as sure about having them here.
Curious what @rccrdpccl thinks.

internal/installercache/installercache.go
```diff
@@ -183,33 +291,34 @@ func (i *Installers) evict() {
...
+	for _, q := range queues {
+		for i.shouldEvict(totalSize) {
```
carbonin, line 312: Because we're evaluating this each loop it feels very much like we should not need the loop in get and should never hit the capacity error, right?

Doesn't this effectively ensure we'll evict releases until we have enough space to download a new one? Or am I misunderstanding something?

carbonin (changes requested):

internal/ignition/installmanifests_test.go
```diff
@@ -78,29 +81,38 @@ var _ = Describe("Bootstrap Ignition Update", func() {
...
+		manifestsAPI   *manifestsapi.MockManifestsAPI
+		eventsHandler  *eventsapi.MockHandler
+		installerCache *installercache.Installers
 	)
 
 	BeforeEach(func() {
 		// setup temp workdir
 		workDir, err = os.MkdirTemp("", "bootstrap-ignition-update-test-")
 		Expect(err).NotTo(HaveOccurred())
+		Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
paul-maidment: Good catch!

internal/installercache/installercache.go
```diff
@@ -80,22 +96,54 @@ func (rl *Release) Cleanup(ctx context.Context) {
...
+		default:
+			release, err := i.get(releaseID, releaseIDMirror, pullSecret, ocRelease, ocpVersion, clusterID)
+			if err == nil {
+				return release, nil
+			}
+			_, isCapacityError := err.(*errorInsufficientCacheCapacity)
+			if !isCapacityError {
+				return nil, errors.Wrapf(err, "failed to get installer path for release %s", releaseID)
+			}
+			i.log.WithError(err).Errorf("insufficient installer cache capacity for release %s", releaseID)
```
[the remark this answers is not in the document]
paul-maidment: This is not the case.

The capacity check performed here is of the amount of free space while accounting for hard links that have not yet been removed.

An eviction is removing release files, not hard links (unless we count the questionable " hard link pruning" functionality.)

A release that we have deleted may still be in use by the caller, until that release is freed by calling `Cleanup` then the actual space will not be freed.

By the time we reached this line we have done the following

1: Checked the free space while accounting for hard links
2: Determined that we do not have enough space in which to store a release
3: Called evict and performed some eviction on the releases
4: Returned to the top of the loop and checked the free space while accounting for hard links again.
5: Determined that we do not have enough space in which to store a release
6: Detected that this is a 'blocking' issue and that unless we have the space, we cannot write
7: Determined that the eviction has already been called and should not be called again as it will most likely not have an impact. 

This is a "fail fast" operation to allow a retry to be attempted outside of the context of the Mutex.

internal/installercache/installercache.go
```diff
@@ -106,43 +154,108 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
...
+				break
+			}
+			i.log.Infof("checking for space to extract release %s", releaseID)
+			// Determine actual space usage, accounting for hardlinks.
+			var usedBytes uint64
+			usedBytes, _, err = i.diskStatsHelper.GetDiskUsage(i.config.CacheDir)
+			if err != nil {
+				if os.IsNotExist(err) {
+					// Installer cache directory will not exist prior to first extraction
+					i.log.WithError(err).Warnf("skipping capacity check as first extraction will trigger creation of directory")
```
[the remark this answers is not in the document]
paul-maidment: I was following the approach taken by the original code but I see your point.

I think we could explicitly create the directory.

internal/installercache/installercache.go
paul-maidment: I will bear this in mind, I'll try to keep things aligned so they 'diff' better in future.
Thanks for raising this.

internal/installercache/installercache.go
paul-maidment: I thought about this and I think there may be cases where we want to trigger eviction sooner.

Consider the scenario where we have capacity for a single release and there are several requests coming in.

In this scenario, the release will always have a hard link.

If we perform an eviction on a release that still has a hard link, this is OK as the actual 'deletion' only happens when all callers are finished with the release.

internal/installercache/installercache_test.go
paul-maidment: I'll add a comment, the general idea was to add a slight simulation of a release being held by the caller, but I didn't want this simulation to take a lot of time, just enough to perhaps force some waiting during the tests.

internal/installercache/installercache_test.go
paul-maidment: The intent here is to have these as smoke tests for the happy path and the general idea is that we are not expecting failure. All I expect in these cases is that any time there is a failure they will break the test so that we know there is a problem.

I don't see this as the comprehensive solution to testing this and expect that a lot more will be done in QE.
If we really want to push this further, we could do some tests in openshift/release

internal/installercache/installercache.go
paul-maidment: Already discussed above, I think something has been misunderstood here.

internal/installercache/installercache.go
```diff
@@ -106,43 +154,108 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
...
+		if !os.IsExist(err) {
+			return nil, errors.Wrap(err, fmt.Sprintf("failed to create hard link to binary %s", path))
+		}
+	}
+}
+
+func (i *Installers) shouldEvict(totalUsed int64) bool {
+	shouldEvict, _ := i.checkEvictionStatus(totalUsed)
+	return shouldEvict
+}
```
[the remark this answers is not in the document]
paul-maidment: The loop inside evict does not have to worry about the 'isBlocked' eviction status as it doesn't deal with this scenario. 

I wanted a simplified version of the `checkEvictionStatus` function that would just return a simple boolean for `shouldEvict`

I am conscious of repeating unnecessary code so this wraps the checkEvictionStatus

This is in the name of keeping the code in evict quite simple, especially as this decision is used to determine the behaviour of a loop.

internal/installercache/installercache.go
```diff
@@ -106,43 +154,108 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
...
+
+func (i *Installers) checkEvictionStatus(totalUsed int64) (shouldEvict bool, isBlocked bool) {
+	if i.config.MaxCapacity == 0 {
+		// The cache eviction is completely disabled.
+		return false, false
+	}
+	if i.config.MaxCapacity-totalUsed < i.config.MaxReleaseSize {
+		// We are badly blocked, not enough room for even one release more.
+		return true, true
```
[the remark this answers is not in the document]
paul-maidment: This is specifically for the case where we are not able to write another release due insufficient space.

The code that handles this case uses the 'second boolean' as 'isBlocked' meaning that no permission is given to write the release.

internal/installercache/installercache_test.go
paul-maidment: At best if/when it fails it will give inconsistent results. <-

How do you feel the results will be inconsistent, can you point out an example in this case?

internal/installercache/installercache_test.go
paul-maidment: You are right, I just tested this function again and we don't need this for the assertions we are performing here.

I think I was attempting to simulate a small amount of latency between fetching a release using `installers.Get(...)` and calling `Cleanup(...)` on the release. In practice however, this is not needed for this simple, non parallel test.

internal/installercache/installercache.go
paul-maidment: I think there could be limited some edge cases where pruning a release with a hard link could present a small advantage but certainly nothing too significant.

It's safe to include them but in this case, I tend to agree that we should just filter them out.

internal/installercache/installercache.go
```diff
@@ -80,22 +96,54 @@ func (rl *Release) Cleanup(ctx context.Context) {
 		"cached", rl.cached,
 		"extract_duration", rl.extractDuration,
 	)
+	if err := os.Remove(rl.Path); err != nil {
+		logrus.New().WithError(err).Errorf("Failed to delete release link %s", rl.Path)
```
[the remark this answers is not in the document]
paul-maidment: This is actually a condition from the original code. I think that in practice this scenario would only occur where the release has already somehow been cleaned up (maybe by an overly aggressive pruning operation?)

I notice that we do not check if this is os.IsNotExist(...), so I will add this filter to the error so we ignore those completely as they are likely caused by the race condition I describe above (we might still want to log when it happens though)

The question is, what to do if it's another type of error?

I'm tending towards a hard fail and failing the install in these cases. My reasoning below...

There are only a few cases that can cause this type error in the context of our solution, the release link is always a file and not a directory so we only need to cover file use cases.

1: Something else cleaned up the release link for us (erroneous premature pruning.) (already covered, we can just ignore.)
2: Some kind of permissions issue or other critical IO issue 
- I think this would only occur if something was wrong with the filesystem. In order to have a `release` to `Cleanup` a lot of things have to have gone right before this moment. If they are going wrong now then we have much bigger problems.

cmd/main.go
```diff
@@ -172,6 +174,18 @@ var Options struct {
 
 	// EnableXattrFallback is a boolean flag to enable en emulated fallback methoid of xattr on systems that do not support xattr.
 	EnableXattrFallback bool `envconfig:"ENABLE_XATTR_FALLBACK" default:"true"`
+
```
gamli75, line 177: please move all the environment variables to the installer cache package

gamli75:
generic remark - if you want to control the installer cache environment variables and change it per environment and during your testing - you need to add it to https://github.com/openshift/assisted-service/blob/master/openshift/template.yaml

pkg/generator/generator.go
```diff
@@ -21,11 +22,10 @@ type InstallConfigGenerator interface {
...
 type Config struct {
-	ServiceCACertPath      string `envconfig:"SERVICE_CA_CERT_PATH" default:""`
-	ReleaseImageMirror     string `envconfig:"OPENSHIFT_INSTALL_RELEASE_IMAGE_MIRROR" default:""`
-	DummyIgnition          bool   `envconfig:"DUMMY_IGNITION"`
-	InstallInvoker         string `envconfig:"INSTALL_INVOKER" default:"assisted-installer"`
-	InstallerCacheCapacity int64  `envconfig:"INSTALLER_CACHE_CAPACITY"`
+	ServiceCACertPath  string `envconfig:"SERVICE_CA_CERT_PATH" default:""`
+	ReleaseImageMirror string `envconfig:"OPENSHIFT_INSTALL_RELEASE_IMAGE_MIRROR" default:""`
+	DummyIgnition      bool   `envconfig:"DUMMY_IGNITION"`
+	InstallInvoker     string `envconfig:"INSTALL_INVOKER" default:"assisted-installer"`
```
gamli75, line 28: You removed INSTALLER_CACHE_CAPACITY from this file but it's still in use in https://github.com/openshift/assisted-service/blob/40b8a8264e6b2509af8773d2324b0b0592e0481a/openshift/template.yaml#L220

paul-maidment: Good catch, thank you!

internal/installercache/installercache.go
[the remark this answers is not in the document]
carbonin: No, I would very much not panic on purpose.
We should fail in a way that can be monitored and alerted on.

internal/installercache/installercache.go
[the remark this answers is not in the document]
carbonin: Why not this?
```
for shouldEvict, _ := i.checkEvictionStatus(totalSize); shouldEvict; {
  ...
}
```

docs/enhancements/installercache.md
```diff
@@ -0,0 +1,51 @@
+# Installer Cache for Openshift Releases
```
carbonin, line 1: If you're going to put something in enhancements I think it should really follow the format and likely have a separate PR.

Alternatively, if you just want to write a dev doc (which I think is what you're doing) we have a separate directory for those (docs/dev)

internal/ignition/installmanifests.go
```diff
@@ -182,6 +180,7 @@ func (g *installerGenerator) Generate(ctx context.Context, installConfig []byte)
 	//cleanup resources at the end
 	defer release.Cleanup(ctx)
 
+	// Critical section -- start
```
carbonin, line 183: The "critical section" of concurrent code is the part that cannot be run in multiple threads at once. That's not the case for this so I'm not sure what you're trying to indicate with these comments.

paul-maidment: Apologies, this is just something that should not have ended up in the PR as I refactor to respond to comments.
Please disregard it for now. Was not meant to be here.

docs/enhancements/installercache.md
paul-maidment: I'll put something in dev docs instead of this then.

internal/installercache/installercache.go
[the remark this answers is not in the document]
paul-maidment: A big question here is how to fail in a way that wont result in a deadlock?

We do have the link pruning code, so we could fail the request without panic and then allow the code in other threads to lock up until the link is pruned?

I guess what I am trying to say here is that 

1: The scenario where we get an error for a reason other than the 'non existence' of the file is so rare that it should be considered unexpected and possibly leading to undefined behaviour.

2: We either need to view this as a scenario that is either

i: Such a rare edge case that it's not worth major concern.

or 

ii: In spite of the rarity, the risk of this happening presents a huge problem that must be immediately addressed.

If this issue occurs in the way I describe, it WILL lead to deadlock due to the inability to clean up a hard link. In these scenarios I have doubts that even the 'hardllink pruning' would be of much help (as if a deletion failed here, it will most likely fail again for the same reasons).

Moreover, it would not self resolve, the installercache would simply lock up on the Mutex.

Clearing this blockage would also be a challenge as it is hard to see how other processes could be helped. 

One thing that would stand a better chance of helping though, is to initiate a panic as this would force a pod eviction and would restore the cache to an empty state, allowing future installs to proceed.

I agree that we should have some alerts or reporting on this scenario, that makes sense. 

I don't agree that we should not panic in this case.

internal/installercache/installercache.go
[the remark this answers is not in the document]
paul-maidment: > Why not this?
> 
> ```
> for shouldEvict, _ := i.checkEvictionStatus(totalSize); shouldEvict; {
>   ...
> }
> ```

Because that doesn't do what you _think_ it does...

Please see the following go playground...

https://go.dev/play/p/3Y858WLm_Y3

```
package main

import "fmt"

func shouldRun(number int) (bool, int) {
	fmt.Printf("Call to shouldRun for number = %d\n", number)
	return number < 100, number
}

func main() {
	n := 1
	for run, _ := shouldRun(n); run; {
		n++
	}

}
```

The function is only called once, on the start of the for loop, even though we update `n`, these updates are never picked up and the program eventually times out.

The fix is to write the for loop as follows...

```
for run, _ := shouldRun(n); run; run, _ = shouldRun(n) {
```

arguably less readable than the simpler form I propose.

internal/installercache/installercache.go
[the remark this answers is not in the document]
paul-maidment: For now, I simply bubble the error up so that it will cause the install for this particular instance to fail.

I think this is a reasonable way to handle things if we are not going to resort to something that might be able to auto recover from the issue.

internal/installercache/installercache_test.go
rccrdpccl: I agree this is not a unit test, but we already have many "unit-tests" which are not properly unit.
As long as it's not slowing down the total "unit-test" time, I think it could be a good test to have in case this code needs to be changed in the future (deadlock here would be horrible).

We could argue that we could split properly unit and more "integration" tests by tag, and execute them separately, but I think it'd be outside the scope of this PR.

Wouldn't ginkgo support parallel execution of table tests? That would make it way more readable than currently. If ginkgo does not support it, I think we should improve readability of these functions

internal/installercache/installercache_test.go
```diff
@@ -118,20 +144,161 @@ var _ = Describe("installer cache", func() {
...
+			}(param)
+		}
+		wg.Wait()
+		return firstError
+	}
+
+	getUsedBytesForDirectory := func(directory string) uint64 {
+		var totalBytes uint64
+		seenInodes := make(map[uint64]bool)
+		err := filepath.Walk(directory, func(path string, fileInfo os.FileInfo, err error) error {
```
rccrdpccl, line 194: Don't we already have a function that does this? Why are we duplicating it here?

internal/metrics/os_disk_stats_helper_test.go
```diff
@@ -7,13 +7,14 @@ import (
...
 	. "github.com/onsi/gomega"
+	"github.com/sirupsen/logrus"
 )
 
 var _ = Describe("OS disk stats helper", func() {
 
 	var (
 		tempDir         string
-		diskStatsHelper *OSDiskStatsHelper = NewOSDiskStatsHelper()
+		diskStatsHelper *OSDiskStatsHelper = NewOSDiskStatsHelper(logrus.New())
```
rccrdpccl, line 17: consider discarding output from the logger for the tests

internal/installercache/installercache_test.go
paul-maidment: It's best practice to not use a function from production code as this may be the function that has a bug in it and to use the same function in a test would sway the test.

So repetition of code here is acceptable IMO.

internal/installercache/installercache.go
```diff
@@ -232,3 +374,42 @@ func (i *Installers) evictFile(filePath string) error {
...
+	for {
+		if in.Len() == 0 {
+			break
+		}
+		fi, _ := in.Pop()
+		stat, ok := fi.info.Sys().(*syscall.Stat_t)
+		if !ok {
+			// If we do encounter an error while performing stat - do not filter that item
+			// it's not optimal, but we don't break anything.
+			i.log.Errorf("could not determine hardlink status for %s - item will not be filtered", fi.path)
```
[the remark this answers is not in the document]
paul-maidment: Yes, I suppose warning makes more sense.

internal/installercache/installercache.go
```diff
@@ -29,11 +34,58 @@ var (
...
+
+type Size int64
+
+type Config struct {
+	CacheDir string
+	// MaxCapacityGB is the capacity of the installer cache in GiB
+	MaxCapacity Size `envconfig:"INSTALLER_CACHE_CAPACITY" default:"0"`
+	// MaxReleaseSizeGB is the expected maximum size of a single release in GiB
+	MaxReleaseSize Size `envconfig:"INSTALLER_CACHE_MAX_RELEASE_SIZE" default:"2 GiB"`
+	// ReleaseFetchRetryIntervalMicroseconds is the number of microseconds that the cache should wait before retrying the fetch of a release if unable to do so for capacity reasons.
```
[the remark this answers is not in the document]
paul-maidment: I did refactor MaxReleaseSize to deal with human readable values so I suppose I could do it here. 
I felt it was maybe a bit overkill to do so, but if you feel strongly?

internal/installercache/installercache.go
```diff
@@ -80,22 +130,55 @@ func (rl *Release) Cleanup(ctx context.Context) {
 		"cached", rl.cached,
 		"extract_duration", rl.extractDuration,
 	)
+	if err := os.Remove(rl.Path); err != nil && !os.IsNotExist(err) {
+		logrus.New().WithError(err).Errorf("failed to delete release link %s", rl.Path)
```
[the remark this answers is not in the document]
paul-maidment: Fair point

internal/installercache/installercache.go
```diff
@@ -29,11 +34,58 @@ var (
...
+	eventsHandler   eventsapi.Handler
+	diskStatsHelper metrics.DiskStatsHelper
+	config          Config
+}
+
+type Size int64
+
+type Config struct {
+	CacheDir string
+	// MaxCapacityGB is the capacity of the installer cache in GiB
```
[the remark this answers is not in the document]
paul-maidment: I think the docs do mention this here https://github.com/openshift/assisted-service/pull/7205/files#diff-710b8e277940964203e1b84f3781c9e051cf2dca759f148f63b0bd6b36238b6fR13

Yeah, missed the label for that comment inconsistency, will fix.

Internally, everything is in bytes, but when being parsed from env var it will have the format "20 GiB"

internal/installercache/installercache.go
```diff
@@ -29,11 +34,58 @@ var (
...
+	MaxCapacity Size `envconfig:"INSTALLER_CACHE_CAPACITY" default:"0"`
+	// MaxReleaseSizeGB is the expected maximum size of a single release in GiB
+	MaxReleaseSize Size `envconfig:"INSTALLER_CACHE_MAX_RELEASE_SIZE" default:"2 GiB"`
+	// ReleaseFetchRetryIntervalMicroseconds is the number of microseconds that the cache should wait before retrying the fetch of a release if unable to do so for capacity reasons.
+	ReleaseFetchRetryIntervalMicroseconds uint `envconfig:"INSTALLER_CACHE_RELEASE_FETCH_RETRY_INTERVAL_MICROSECONDS" default:"30000000"`
+	// InstallerCacheEvictionThresholdPercent is the percentage of capacity at which the cache will start to evict releases.
```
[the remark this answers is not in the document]
paul-maidment: I think the idea behind the threshold is that we don't want to be too close to the capacity before we start evicting. There is no reason not to make this default to 95% for example.

I think when the cache runs very close to the mandatory limit, because we are enforcing strict rules around writes, stuff has to wait.

So trimming the cache before this is a last resort is the idea here.

internal/installercache/installercache.go
```diff
@@ -29,11 +34,58 @@ var (
...
+	EvictionThresholdPercent uint `envconfig:"INSTALLER_CACHE_EVICTION_THRESHOLD_PERCENT" default:"80"`
+}
+
+func (s *Size) Decode(value string) error {
```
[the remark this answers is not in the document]
paul-maidment: Thanks for the suggestion, the units API docs were not so clear and I did not spot the Base2Bytes function,

I will implement as follows...

```
func (s *Size) Decode(value string) error {
	bytes, err := units.ParseBase2Bytes(value)
	if err != nil {
		return err
	}
	*s = Size(bytes)
	return nil
}
```

internal/installercache/installercache.go
```diff
@@ -106,56 +189,115 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
...
+			usedBytes, _, err = i.diskStatsHelper.GetDiskUsage(i.config.CacheDir)
+			if err != nil {
+				if os.IsNotExist(err) {
+					// Installer cache directory will not exist prior to first extraction
+					i.log.WithError(err).Warnf("skipping capacity check as first extraction will trigger creation of directory")
+					break
+				}
+				return nil, errors.Wrapf(err, "could not determine disk usage information for cache dir %s", i.config.CacheDir)
+			}
+			shouldEvict, isBlocked := i.checkEvictionStatus(int64(usedBytes))
```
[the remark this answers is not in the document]
paul-maidment: Different school of thought really...

My take on it is that while we are in the `Get` function, we are still in the Mutex, in other words, we still have our 'place' in the queue.

My idea is 

- Check available space
- Try to evict cache
- Check available space again
- If we freed enough space, great....start extraction
- Otherwise exit

That second check gives one last chance before losing our place in the queue. 

I accept that the loop structure of this could make it harder to follow but I argue that it exists for two good reasons.

1: To implement the approach above
2: To avoid unnecessary code repetition.

If were to agree that the second space check could be skipped then I would agree with you, but I don't agree we should skip that second check...

internal/installercache/installercache.go
```diff
@@ -183,35 +325,35 @@ func (i *Installers) evict() {
...
-		finfo := links[idx]
-		//Allow a grace period of 5 minutes from the link creation time
-		//to ensure the link is not being used.
-		grace := time.Now().Add(DeleteGracePeriod).Unix()
-		if finfo.info.ModTime().Unix() < grace {
-			os.Remove(finfo.path)
-		}
-	}
+	// TODO: We might want to consider if we need to do this longer term, in theory every hardlink should be automatically freed.
+	// For now, moved to a function so that this can be tested in unit tests.
```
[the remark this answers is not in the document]
paul-maidment: I disagree, documentation of my motive for keeping this here is the intention. 
Otherwise you might think that I did not consider it...

internal/installercache/installercache.go
```diff
@@ -183,35 +325,35 @@ func (i *Installers) evict() {
...
+	i.pruneExpiredHardLinks(links, linkPruningGracePeriod)
 
-	//delete the oldest file if necessary
-	for totalSize >= int64(float64(i.storageCapacity)*CacheLimitThreshold) {
-		finfo, _ := files.Pop()
+	// delete the oldest file if necessary
+	// prefer files without hardlinks first because
+	// 1: files without hardlinks are likely to be least recently used anyway
+	// 2: files without hardlinks will immediately free up storage
```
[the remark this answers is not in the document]
paul-maidment: Yeah, I can agree with that

internal/metrics/os_disk_stats_helper_test.go
paul-maidment: What do you mean?

cmd/main.go
```diff
@@ -488,7 +491,12 @@ func main() {
 	failOnError(err, "failed to create valid bm config S3 endpoint URL from %s", Options.BMConfig.S3EndpointURL)
 	Options.BMConfig.S3EndpointURL = newUrl
 
-	generator := generator.New(log, objectHandler, Options.GeneratorConfig, Options.WorkDir, providerRegistry, manifestsApi, eventsHandler)
+	installGeneratorDirectoryConfig := generator.InstallGeneratorDirectoryConfig{WorkDir: Options.WorkDir}
+	Options.InstallerCacheConfig.CacheDir = filepath.Join(installGeneratorDirectoryConfig.GetWorkingDirectory(), "installercache")
+	installerCache, err := installercache.New(Options.InstallerCacheConfig, eventsHandler, diskStatsHelper, log)
+	failOnError(err, "failed to instantiate installercache")
+
+	generator := generator.New(installGeneratorDirectoryConfig, log, objectHandler, Options.GeneratorConfig, providerRegistry, manifestsApi, eventsHandler, installerCache)
```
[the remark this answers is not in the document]
paul-maidment: Honestly, I think it's overkill.

internal/installercache/installercache.go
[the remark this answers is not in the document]
paul-maidment: You are right, this would work better if I just get envconfig to parse to a time.Duration.

internal/installercache/installercache.go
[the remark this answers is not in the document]
carbonin: Ah okay, right, that initialization portion would only run once.

Okay can you put a comment here then saying why this function exists? Maybe put it down by it's use as well just to be super clear.

internal/installercache/installercache.go
```diff
@@ -183,35 +325,35 @@ func (i *Installers) evict() {
...
+	q := i.filterOutHardlinkedReleases(files)
+	for i.shouldEvict(totalSize) {
+		if q.Len() == 0 { // If we have cleaned out this queue then break the loop
```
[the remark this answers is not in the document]
paul-maidment: I replied to a comment from @carbonin about this loop, I think a similar thing applies to your suggestion

https://go.dev/play/p/ciDr1oev1Ou

Here is a small simulation where I paraphrase the concept you suggest

```
package main

import "fmt"

func shouldRun(number int) bool {
	fmt.Printf("Call to shouldRun for number = %d\n", number)
	return number < 100
}

func main() {
	q := 50
	n := 1
	for shouldRun(n); q > 0; {
		n++
		q--
	}

}

```

The output of this is 

```

File: internal/installercache/installercache.go (hunk 24 above)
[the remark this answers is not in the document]
Comment (carbonin) [reply internal/installercache/installercache.go:214]: I think my comment is coming from the same place as @rccrdpccl's this function to too complex.

Maybe it could be improved if `evict` itself told us (via return value) if it actually successfully removed enough releases. Maybe an error from `evict` indicates that it was not able to reclaim enough space (whether because the releases were in use or because of some error).

Then we wouldn't have to also check here, we'd just check `evict`'s return value.
This would also resolve the confusion I was having around why we're checking if we have enough space in both `evict` and in `get`.

File: internal/installercache/installercache_test.go (hunk 4 above)
Comment (paul-maidment) [reply internal/installercache/installercache_test.go:163]: > I agree this is not a unit test, but we already have many "unit-tests" which are not properly unit. As long as it's not slowing down the total "unit-test" time, I think it could be a good test to have in case this code needs to be changed in the future (deadlock here would be horrible).
> 
> We could argue that we could split properly unit and more "integration" tests by tag, and execute them separately, but I think it'd be outside the scope of this PR.
> 
> Wouldn't ginkgo support parallel execution of table tests? That would make it way more readable than currently. If ginkgo does not support it, I think we should improve readability of these functions

Why specifically is it not a unit test?

The unit under test is the `Get` function of the installer cache, what exactly are we integrated with?

I think it's a weak argument to suggest that the event or metrics mocks represent a significant integration. Writing to disk is a core responsibility of the unit under test so I don't consider the filesystem to be an integration either.

The Mutex is part of the 'Unit' we are testing and is the very thing being tested by the parallel code here.

I am willing to investigate the suggestion that we could make this easier to follow, so I will first look at the suggestion about Ginko parallel tests and later at what can be done to make things clearer.

**Comment by paul-maidment** [comment]
> @paul-maidment Any chance you can provide some description about the installerCache now with your changes (what configuration it should get (and why), what is its main purpose, how it operators, etc. ?
> 
> I know it should function as a cache for openshift/installer binaries, but not much more than that

Added something to the docs directory to describe the functionality

File: internal/installercache/installercache.go (hunk 24 above)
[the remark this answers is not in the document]
Comment (paul-maidment) [reply internal/installercache/installercache.go:194]: The issue is that the eviction that takes place in evict checks only the space we should evict, i.e. the space occupied by releases. 

Whereas the check that takes place in the first loop checks while accounting for hard links.

The primary purpose is to `guarantee` that we do not blow past the storage limit.

If we do not account for the space including hard links in this initial check then this opens up that possibility. 
I know we have spoke before about 'allowing the limit to be broken momentarily' but I don't think we should allow this at all.

I'd argue that if we want to properly maintain this limit then we need the check to be performed slightly differently in the two places. I get what you are saying about complexity but it's there for a good reason in my opinion.

**Comment by gamli75** [comment]
@paul-maidment let's hold this PR or change it to WIP until we finish the code review process - there is no need to run tests every few minutes.

File: internal/installercache/installercache.go (hunk 24 above)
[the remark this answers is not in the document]
Comment (rccrdpccl) [reply internal/installercache/installercache.go:194]: >
    Check available space
    Try to evict cache
    Check available space again
    If we freed enough space, great....start extraction
    Otherwise exit

Assuming `enoughSpaceForRelease()` is checking available space, and `evict()` tries to evict, the sample code is doing exactly what you are describing. But I'd argue it's way easier to follow, I see no benefits in leaving the code as it is, nor in knowing if the eviction is blocked or not (if there's not enough space, the eviction is blocked, no need to complicate the `shouldEvict` function)

File: internal/installercache/installercache.go (hunk 25 above)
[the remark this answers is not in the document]
Comment (rccrdpccl) [reply internal/installercache/installercache.go:312]: I meant the unit test part. Just write the unit test, no need to write about "moving it to a function", the next person going through here wouldn't even know this wasn't a function before. The more we can keep comments relevant the more chances we have that people read them

File: internal/installercache/installercache.go (hunk 22 above)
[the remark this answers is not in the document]
Comment (paul-maidment) [reply internal/installercache/installercache.go:51]: Because when you are at the threshold (dead on) you are not allowed to write anything, I would argue that this could slow things down a bit.

File: internal/metrics/os_disk_stats_helper_test.go (hunk 17 above)
Comment (rccrdpccl) [reply internal/metrics/os_disk_stats_helper_test.go:17]: For example https://github.com/openshift/assisted-service/blob/ecbb8a578309e35728c0bf10ec10227914230aa0/pkg/thread/example_test.go#L19

IIRC we should have a utility function that initializes the logger for tests

File: internal/installercache/installercache.go (hunk 24 above)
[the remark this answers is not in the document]
Comment (paul-maidment) [reply internal/installercache/installercache.go:194]: `if there's not enough space, the eviction is blocked` -> How do you imagine we are determining this?

File: cmd/main.go (hunk 27 above)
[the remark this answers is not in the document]
Comment (rccrdpccl) [reply cmd/main.go:499]: it's not about clarity of the added code, is about a 1000 lines file. However it's just a nit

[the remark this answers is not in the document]
Comment (paul-maidment) [reply cmd/main.go:499]: Given that we had a bug related to incorrect instantiation, I'd rather keep the instantiation in a place where I can predict it well. I think this is the right place.

File: internal/installercache/installercache_test.go (hunk 16 above)
Comment (paul-maidment) [reply internal/installercache/installercache_test.go:197]: > You have a test for that function that tests that. Repetition of this code adds zero value, 
I have been given negative reviews by @carbonin in the past for using production code in this way (even when it was covered by tests.) 

>infact forces us to know that this function is here and to modify it in the case (however unlikely) that this changes.
I'd argue the same for the fragility of changing a method in production code.

But let me know what is most idiomatic, I'm confused by the variable standards on this.

File: internal/installercache/installercache.go (hunk 24 above)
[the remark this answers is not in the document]
Comment (paul-maidment) [reply internal/installercache/installercache.go:194]: I still think that you are mistaken

We need to consider a few things

- We are running in a multi threaded environment where a new request to GET from the cache may be queued up behind this one.
- If you base eviction on release size + hardlinks and gauge success on your ability to purge releases that do not have hardlinks then you will over evict.
-- You don't know the true disk space availability until pending hardlinks have been removed.
-- This will lead to the evict routine believing there is less free space than we actually have
-- This will lead to releases without hardlinks being evicted
-- Not every release without a hardlink needs to be evicted
-- If enough threads run in the timespan it takes to free a hardlink, you could have a prematurely empty cache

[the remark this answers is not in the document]
Comment (paul-maidment) [reply internal/installercache/installercache.go:194]: > I think my comment is coming from the same place as @rccrdpccl's - this function to too complex.
> 
> Maybe it could be improved if `evict` itself told us (via return value) if it actually successfully removed enough releases. Maybe an error from `evict` indicates that it was not able to reclaim enough space (whether because the releases were in use or because of some error).

You can certainly do this but this makes an assumption that it is always correct to evict a release if it is in the queue and does not have a hardlink associated with it.

In order for this to be safe, the 'has enough space' calculation that we perform during the eviction needs to be based on the disk space used by the releases while excluding the hard links. 

If we account for the hard links then we will (from the perspective of eviction) under report free space. Which will cause over eviction to occur.

> 
> Then we wouldn't have to also check here, we'd just check `evict`'s return value. This would also resolve the confusion I was having around why we're checking if we have enough space in both `evict` and in `get`.

We need to ensure that there is enough space to write a release, we do not know about the status of the hardlinks, this means that we need to account for that space before making a write.

I don't think there is a good way around this.

File: internal/installercache/installercache.go (hunk 22 above)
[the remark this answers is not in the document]
Comment (rccrdpccl) [reply internal/installercache/installercache.go:51]: Wouldn't you always have this issue? Wouldn't it be best to keep it simple and risk to incur into it more often? I think the root cause of this is that eviction doesn't actually evict. if we want to fix this, I'd rather fix eviction

File: internal/installercache/installercache_test.go (hunk 16 above)
Comment (carbonin) [reply internal/installercache/installercache_test.go:197]: Yeah, I think my concern is more where there's some function that is actually part of what's being tested also being used to validate the output of the test.

That isn't really the case here because the function is an external (to the package) utility that is tested in isolation.

I don't care strongly about how you resolve this though. Leaving it here or using the utility makes no difference to me.

File: internal/installercache/installercache.go (hunk 25 above)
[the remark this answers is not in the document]
Comment (paul-maidment) [reply internal/installercache/installercache.go:294]: I agree. was initially confused as you highlighted both lines.

File: internal/installercache/installercache_test.go (hunk 16 above)
Comment (paul-maidment) [reply internal/installercache/installercache_test.go:197]: If OK, I will keep this as is for now, this may change if I refactor the tests to work with parallel table.

File: internal/installercache/installercache.go [hunk 29]
```diff
@@ -80,22 +111,87 @@ func (rl *Release) Cleanup(ctx context.Context) {
...
+	if err != nil && !os.IsNotExist(err) {
+		return 0, errors.Wrapf(err, "could not determine disk usage information for cache dir %s", i.config.CacheDir)
+	}
+	return usedBytes, nil
+}
+
+func (i *Installers) extractReleaseIfNeeded(path, releaseID, releaseIDMirror, pullSecret, ocpVersion string, ocRelease oc.Release, release *Release) error {
+	_, err := os.Stat(path)
+	if err == nil {
+		return nil // No error indicates that the release already exists
```
carbonin, line 172: This isn't really true because we return `nil` if we successfully fetch the release as well, right?

I think it's just as valid to just not have this comment.

internal/installercache/installercache.go
```diff
@@ -80,22 +111,87 @@ func (rl *Release) Cleanup(ctx context.Context) {
...
+	if i.shouldEvict(int64(usedBytes)) && !i.evict() {
+		return &errorInsufficientCacheCapacity{Message: fmt.Sprintf("insufficient capacity in %s to store release", i.config.CacheDir)}
+	}
+	extractStartTime := time.Now()
+	_, err = ocRelease.Extract(i.log, releaseID, releaseIDMirror, i.config.CacheDir, pullSecret, ocpVersion)
+	if err != nil {
+		return err
+	}
+	release.extractDuration = time.Since(extractStartTime).Seconds()
+	release.cached = false
```
carbonin, line 190: Generally I prefer returning extra variables over side effects like setting fields on pointer arguments.
This is definitely a personal preference thing, but what do you think about returning variables for `cached` and `extractDuration` where `err == nil` and `cached == false` would mean `extractDuration` should be valid?

internal/installercache/installercache.go
```diff
@@ -232,3 +327,42 @@ func (i *Installers) evictFile(filePath string) error {
...
+
+// pruneExpiredHardLinks removes any hardlinks that have been around for too long
+// the grace period is used to determine which links should be removed.
+func (i *Installers) pruneExpiredHardLinks(links []*fileInfo, gracePeriod time.Duration) {
+	for idx := 0; idx < len(links); idx++ {
+		finfo := links[idx]
+		graceTime := time.Now().Add(-1 * gracePeriod)
+		grace := graceTime.Unix()
+		if finfo.info.ModTime().Unix() < grace {
+			os.Remove(finfo.path)
```
carbonin, line 339: I know we were not checking this before but should we be checking this error?
Also if we think we really shouldn't have to ever run this should we maybe log if we do end up deleting something here?

internal/installercache/installercache.go
```diff
@@ -183,35 +282,31 @@ func (i *Installers) evict() {
...
-	}
+	// TODO: We might want to consider if we need to do this longer term, in theory every hardlink should be automatically freed.
+	i.pruneExpiredHardLinks(links, linkPruningGracePeriod)
 
-	//delete the oldest file if necessary
-	for totalSize >= int64(float64(i.storageCapacity)*CacheLimitThreshold) {
-		finfo, _ := files.Pop()
+	// delete the oldest file if necessary
+	evicted := false
+	q := i.filterOutHardlinkedReleases(files)
```
carbonin, line 298: Since we now only use the one list (files without hardlinks) can we do this filtering in the `visit` function rather than in this post-processing?

internal/installercache/installercache_test.go
```diff
@@ -53,60 +59,81 @@ var _ = Describe("release event", func() {
 			"cached", r.cached,
 			"extract_duration", r.extractDuration,
 		).Times(1)
-		r.Cleanup(ctx)
+		err = r.Cleanup(ctx)
+		Expect(err).To(BeNil())
```
carbonin, line 63: Nit, but I prefer `Expect(r.Cleanup(ctx)).To(Succeed())` for these cases.

internal/installercache/installercache_test.go
```diff
@@ -53,60 +59,81 @@ var _ = Describe("release event", func() {
...
+			ReleaseFetchRetryInterval: 1 * time.Microsecond,
+		}
+		manager, err = New(installerCacheConfig, eventsHandler, diskStatsHelper, logrus.New())
+		Expect(err).NotTo(HaveOccurred())
 		ctx = context.TODO()
 	})
 
 	AfterEach(func() {
-		os.RemoveAll(cacheDir)
+		//os.RemoveAll(cacheDir)
```
carbonin, line 102: If we don't need this then just remove it.

internal/installercache/installercache_test.go
```diff
@@ -53,60 +59,81 @@ var _ = Describe("release event", func() {
...
+				err = os.WriteFile(fname, []byte("abcde"), 0600)
 				return "", err
-			})
+			}).AnyTimes()
+	}
+
+	testGet := func(releaseID, version string, clusterID strfmt.UUID, expectCached bool) (string, string) {
+		workdir := filepath.Join(manager.config.CacheDir, "quay.io", "release-dev")
+		fname := filepath.Join(workdir, releaseID)
+		mockReleaseCalls(releaseID, version)
```
carbonin, line 135: Should this depend on `expectCached` as well?

internal/installercache/installercache_test.go
```diff
@@ -161,42 +331,42 @@ var _ = Describe("installer cache", func() {
...
+		By("verify that the oldest file was deleted")
 		_, err := os.Stat(r1)
-		Expect(os.IsNotExist(err)).To(BeFalse())
+		Expect(os.IsNotExist(err)).To(BeTrue())
 		_, err = os.Stat(r2)
 		Expect(os.IsNotExist(err)).To(BeFalse())
 		_, err = os.Stat(r3)
 		Expect(os.IsNotExist(err)).To(BeFalse())
 
+		By("verify that the links were purged")
```
carbonin, line 352: Isn't this just testing that you called `Cleanup` in `testGet`?

internal/installercache/installercache_test.go
```diff
@@ -161,42 +331,42 @@ var _ = Describe("installer cache", func() {
...
+		Expect(os.IsNotExist(err)).To(BeTrue())
+		_, err = os.Stat(l2)
+		Expect(os.IsNotExist(err)).To(BeTrue())
+		_, err = os.Stat(l3)
+		Expect(os.IsNotExist(err)).To(BeTrue())
+		// Now measure disk usage, we should be under the cache size
+		Expect(getUsedBytesForDirectory(manager.config.CacheDir)).To(BeNumerically("<=", manager.config.MaxCapacity))
 	})
 
 	It("extracts from the mirror", func() {
```
carbonin, line 364: Where is this test actually testing anything related to the mirror? This one and the following test look almost exactly the same (at least the verification bits do).
Can we make sure this is testing what it is supposed to be testing?

pkg/generator/generator.go
```diff
@@ -39,18 +39,29 @@ type installGenerator struct {
 	providerRegistry registry.ProviderRegistry
 	manifestApi      manifestsapi.ManifestsAPI
 	eventsHandler    eventsapi.Handler
+	installerCache   *installercache.Installers
 }
 
-func New(log logrus.FieldLogger, s3Client s3wrapper.API, cfg Config, workDir string,
-	providerRegistry registry.ProviderRegistry, manifestApi manifestsapi.ManifestsAPI, eventsHandler eventsapi.Handler) *installGenerator {
+type InstallGeneratorDirectoryConfig struct {
```
carbonin, line 45: Why create this struct? It doesn't seem to improve anything over just passing the workdir.

carbonin (commented):
Generally looking much better after the last round of changes.

internal/installercache/installercache.go
paul-maidment: I kinda meant "no error from this call"
But I'll remove it

internal/installercache/installercache.go
paul-maidment: I considered that, the only thing that worries me is that I do have a piece of code in this filtering method that takes some action if a stat fails. 

In initially preferred to avoid this just in case I introduce a problem, but I can take another look. 

I can see the benefit of cutting down on a few cycles

internal/installercache/installercache_test.go
paul-maidment: is better, one liner also

internal/installercache/installercache_test.go
paul-maidment: I guess it tests that cleanup actually cleans the links?
Code that was here before in any case, I tried not to interfere too much with existing tests

pkg/generator/generator.go
paul-maidment: The original implementation of installercache was instantiated relative to the working directory of the install generator.

I wanted to find a way to leave the responsibility for finding our which root directory would be used by the installer generator with the installer generator.

This had to be done without instantiating the generator as the installercache is a required component of the generator.

TLDR: This is the point

https://github.com/openshift/assisted-service/pull/7205/files#diff-9fde1b602271222d4351f8e635e3398b84f0af0cc0df19ce8b9f5095fefb5c1bR50-R52

internal/installercache/installercache_test.go
paul-maidment: You know what, you're right...
Again these methods were already in the tests and I did not look too deeply at them.

But we can basically delete one of them, all we are proving is that the release mirror parameter gets passed to the openshift extraction code. No value in testing both paths.

internal/installercache/installercache.go
paul-maidment: We can certainly log something but you should also know that I plan to record this in the metrics ticket.

internal/installercache/installercache_test.go
paul-maidment: I am not sure what you mean exactly? 

But the intent here is to be able to pass expectations about the cached status within the test so that assertions may be made.

It seems to be doing the job?

internal/installercache/installercache.go
paul-maidment: I'll switch it up so we avoid side effect and use return values, I can see that value in this

pkg/generator/generator.go
paul-maidment: I see we can use the GeneratorConfig instead, I have moved the things I need to that

internal/installercache/installercache.go
paul-maidment: Not any more, we made some changes.

Though I was simply referring to the meaning of err == nil on LN 171, perhaps the comment should have been there.

internal/installercache/installercache_test.go
carbonin: I guess I'm asking if we want to ensure we're not calling the release code if we're expecting the release to be cached.

internal/installercache/installercache_test.go
paul-maidment: I have given the functions a tidy up, I looked into Ginko and I couldn't find way to run parallel table tests.
I did clean our existing tests up though to make them more readable.

@rccrdpccl

internal/installercache/installercache_test.go
paul-maidment: I mean, TBH, calling `release.Cleanup` is not a huge issue, we need to remember that this is just cleaning up the hardlinks to simulate a normal flow of the system.

Even if something is cached, a hardlink is issued every time someone requests that release via `Get`

Clearing the hardlink may assist the eviction of that release in some cases (as there will be no hardlinks) but I don't think any of the cases in `testGet` are doing anything like this.

internal/installercache/installercache_test.go
```diff
@@ -118,20 +147,173 @@ var _ = Describe("installer cache", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		var wg sync.WaitGroup
+		var reportedError error
+		var errMutex sync.Mutex
+		for _, t := range tests {
+			wg.Add(1)
+			go func(t test) {
+				defer wg.Done()
+				release, err := runTest(t, manager)
+				handleError(&errMutex, err, &reportedError)
```
rccrdpccl, line 226: IIUC this would only report the first error. Writing the error on a channel would be more idiomatic and might help to both simplify and report errors correctly.

paul-maidment: I actually used a channel based approach before but did not like it for a number of reasons.

- It did not happen often but I found I would get the occasional test flake where all of the errors had not been collected (it was rare but I did not like this)
- I only need the first error in any case and it does not add value to have them all.
- The Mutex does a great job of making sure we can examine the error once the async jobs are done.

internal/installercache/installercache.go
```diff
@@ -8,19 +8,22 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
+	"github.com/alecthomas/units"
```
carbonin, line 14: Would it be better to use a more widely accepted library for this?
Maybe `k8s.io/apimachinery/pkg/api/resource`?

internal/installercache/installercache.go
```diff
@@ -177,41 +268,49 @@ func (i *Installers) evict() {
...
+		}
+		if stat.Nlink == 1 { // This means we have no hardlinks to the file.
+			includeFile = true
+		}
+
+		// if the file has no hardlinks then add it to the queue
+		if includeFile {
+			files.Add(&fileInfo{path, info})
+			totalSize += info.Size()
+		}
```
carbonin, line 285: This intermediate variable is not needed. Something like this would work

```suggestion
		stat, ok := info.Sys().(*syscall.Stat_t)
		if !ok || stat.Nlink == 1 {
		        if !ok {
  			      i.log.Warnf("could not determine hardlink status for %s - item will not be filtered", path)
  			}
			files.Add(&fileInfo{path, info})
			totalSize += info.Size()
		}
```

internal/installercache/installercache.go
```diff
@@ -177,41 +268,49 @@ func (i *Installers) evict() {
...
-		}
-	}
+	// TODO: We might want to consider if we need to do this longer term, in theory every hardlink should be automatically freed.
+	i.pruneExpiredHardLinks(links, linkPruningGracePeriod)
 
-	//delete the oldest file if necessary
-	for totalSize >= int64(float64(i.storageCapacity)*CacheLimitThreshold) {
+	// delete the oldest file if necessary
+	evicted := false
+	for i.shouldEvict(totalSize); i.shouldEvict(totalSize) && files.Len() > 0; {
```
carbonin, line 303: Pretty sure you don't need the three clause for loop here any more since you're not initializing a variable or incrementing anything.

```suggestion
	for i.shouldEvict(totalSize) && files.Len() > 0 {
```

internal/installercache/installercache_test.go
carbonin: I don't understand this comment.. This line has nothing to do with `Cleanup`

`mockReleaseCalls` is saying "I'm expecting this release to be extracted", but we shouldn't expect that if the release should be cached, right?

openshift/template.yaml
```diff
@@ -218,7 +218,7 @@ parameters:
 - name: DEPLOYMENT_TYPE
   value: "SaaS"
 - name: INSTALLER_CACHE_CAPACITY
-  value: "6442450944"
+  value: "32 GiB"
```
carbonin, line 221: I would expect this to use the same memory units as kubernetes resources.
ref: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

internal/installercache/installercache.go
carbonin: Ah okay, I didn't know this one was already in use. That's fine then.

openshift/template.yaml
carbonin: This can be ignored too if we're sticking with "github.com/alecthomas/units"

internal/installercache/installercache_test.go
paul-maidment: Oh sorry, I misunderstood your original statement, you mean to check we are not calling the 'release extraction' code. Got it! I'll look at this.

openshift/template.yaml
paul-maidment: I think we are sticking with github.com/alecthomas/units for now

internal/installercache/installercache_test.go
paul-maidment: The original reason was actually to test that the hardlink eviction was purging stuff. 

https://github.com/openshift/assisted-service/blob/master/internal/installercache/installercache_test.go#L71

So, with this in mind, I'll get rid of these verifications.

internal/installercache/installercache.go
```diff
@@ -177,41 +268,42 @@ func (i *Installers) evict() {
...
 		}
 
-		//save the other files based on their mod time
-		files.Add(&fileInfo{path, info})
-		totalSize += info.Size()
+		stat, ok := info.Sys().(*syscall.Stat_t)
+		if !ok || stat.Nlink == 1 {
+		    if !ok {
+  				i.log.Warnf("could not determine hardlink status for %s - item will not be filtered", path)
+  			}
```
carbonin, line 275: This indentation is weird, I imagine the linter will complain. You should probably run that locally before pushing.

carbonin (commented):
I think this is okay to test at this point.
You can remove draft now @paul-maidment

rccrdpccl (approved):

PR #7213, opened 2025-01-21
MGMT-19599: Update get bundles endpoint to return a list of bundles instead of list of bundles names

eranco74:
This is a followup PR following a review comment https://github.com/openshift/assisted-service/pull/7170#discussion_r1922569273

- Is this PR relying on CI for an e2e test run? yes
- Should this PR be tested in a specific environment? no

### DIFF
--- a/internal/operators/manager_test.go
+++ b/internal/operators/manager_test.go
@@ -828,10 +828,17 @@ var _ = Describe("Operators manager", func() {
 
 		It("ListBundle should return the list of available bundles", func() {
 			bundles := manager.ListBundles()
-			Expect(bundles).To(ConsistOf(
-				operatorscommon.BundleVirtualization,
-				operatorscommon.BundleOpenshiftai,
-			))
+			Expect(bundles).To(HaveLen(2))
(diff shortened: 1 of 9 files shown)


internal/operators/manager_test.go
```diff
@@ -821,10 +821,10 @@ var _ = Describe("Operators manager", func() {
 
 		It("ListBundle should return the list of available bundles", func() {
 			bundles := manager.ListBundles()
-			Expect(bundles).To(ConsistOf(
-				operatorscommon.BundleVirtualization,
-				operatorscommon.BundleOpenshiftai,
-			))
+			Expect(bundles).To(HaveLen(2))
+			for _, bundle := range bundles{
+				Expect(bundle.Name).To(BeElementOf(operatorscommon.BundleVirtualization, operatorscommon.BundleOpenshiftai))
```
paul-maidment, line 826: As we are no longer simply processing strings but instead processing bundles, is there anything else we should expect here, I know that the Bundle also have a list of operators for example?

PR #7206, opened 2025-01-20
NO-ISSUE: Recalculate operator dependencies before validations

jhernand:
Currently operator dependencies are only calculated when a cluster is created or updatated. But certain dependencies are dynamic, and may change when new hosts are added. For example, if a cluster has the OpenShift AI operator installed, it will also require the NVIDIA GPU operator only if there are hosts that have NVIDIA GPUs. To support those dynamic dependencies this patch modifies the cluster monitor so that it recalculates the operator dependencies before checking validations.

## List all the issues related to this PR

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [X] Automation (CI, tools, etc)
- [X] Cloud
- [X] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [X] Manual (Elaborate on how it was tested)
- [ ] No tests needed

Tested manually creating a cluster with the OpenShift AI operator enabled and modified so that it dynamically adds the NVIDIA GPU operator (that will be in a separate pull request). Initially the NVIDIA GPU operator isn't added to the list of operators, but when a node with a GPU is added the NVIDIA GPU operator is automatically enabled.

## Checklist

- [X] Title and description added to both, commit and PR.
- [X] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [X] This change does not require a documentation update (docstring, `docs`, README, etc)
- [X] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #7218, opened 2025-01-23
NO-ISSUE: Add NVIDIA GPU operator only if there are NVIDIA GPUs

jhernand:
Currently when the OpenShift AI operator is enabled the NVIDIA GPU is enabled by default, even if there are no NVIDIA GPUs in the hosts. This patch changes that so that the NVIDIA GPU operator will only be added when there is at least one NVIDIA GPU present.

This is a preparation to add support for other GPU operators, in particular the AMD GPU operator: we don't want to always enable all the GPU operators.

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/operators/nvidiagpu/nvidia_gpu_operator.go
+++ b/internal/operators/nvidiagpu/nvidia_gpu_operator.go
@@ -2,6 +2,8 @@ package nvidiagpu
 
 import (
 	"context"
+	"slices"
+	"strings"
 	"text/template"
 
 	"github.com/kelseyhightower/envconfig"
@@ -15,8 +17,8 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// nvidiaVendorID is the PCI vendor identifier of NVIDIA devices.
-const nvidiaVendorID = "10de"
+// VendorID is the PCI vendor identifier of NVIDIA devices.
+const VendorID = "10de"
 
 var Operator = models.MonitoredOperator{
 	Namespace:        "nvidia-gpu-operator",
@@ -130,12 +132,7 @@ func (o *operator) ValidateHost(ctx context.Context, cluster *common.Cluster, ho
 }
 
 func (o *operator) hasNvidiaGPU(inventory *models.Inventory) bool {
-	for _, gpu := range inventory.Gpus {
-		if gpu.VendorID == nvidiaVendorID {
-			return true
-		}
-	}
-	return false
+	return slices.ContainsFunc(inventory.Gpus, IsSupportedGpu)
 }
 
 func (o *operator) isSecureBootEnabled(inventory *models.Inventory) bool {
@@ -192,4 +189,8 @@ func (o *operator) GetFeatureSupportID() models.FeatureSupportLevelID {
 
 func (o *operator) GetBundleLabels() []string {
 	return []string(Operator.Bundles)
+}
+
+func IsSupportedGpu(gpu *models.Gpu) bool {
+	return strings.EqualFold(gpu.VendorID, VendorID)
 }
--- a/internal/operators/openshiftai/openshift_ai_operator_test.go
+++ b/internal/operators/openshiftai/openshift_ai_operator_test.go
@@ -2,16 +2,21 @@ package openshiftai
 
 import (
 	"context"
+	"encoding/json"
 	"os"
+	"strings"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	"github.com/openshift/assisted-service/internal/common"
 	"github.com/openshift/assisted-service/internal/operators/api"
+	"github.com/openshift/assisted-service/internal/operators/nvidiagpu"
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/assisted-service/pkg/conversions"
 )
+
+const virtIOVendorID = "1a4f"
 
 var _ = Describe("Operator", func() {
 	var (
@@ -131,7 +136,7 @@ var _ = Describe("Operator", func() {
 			"NVIDIA is supported by default",
 			nil,
 			&models.Gpu{
-				VendorID: "10de",
+				VendorID: nvidiagpu.VendorID,
 			},
 			true,
 		),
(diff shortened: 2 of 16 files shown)


jhernand:
/hold

@eifrach for this to work we need to recalculate operator dependencies when new hosts are added, as suggested in #7206, otherwise in the UI workflow the the NVIDIA GPU operator will never be added, because the steps to add operators happens before the step to add hosts.

internal/operators/openshiftai/openshift_ai_operator_test.go
```diff
@@ -171,5 +173,111 @@ var _ = Describe("Operator", func() {
 			},
 			true,
 		),
+		Entry(
+			"NVIDIA is supported even if not explicitly added",
+			map[string]string{
+				"OPENSHIFT_AI_SUPPORTED_GPUS": "1af4",
+			},
+			&models.Gpu{
+				VendorID: "10de",
```
paul-maidment, line 182: I think we want to make this a constant like `VENDOR_ID_NVIDIA` so that we can use it wherever it is needed instead of hardwiring it in various places.

internal/operators/nvidiagpu/nvidia_gpu_operator.go
```diff
@@ -193,3 +189,7 @@ func (o *operator) GetFeatureSupportID() models.FeatureSupportLevelID {
 func (o *operator) GetBundleLabels() []string {
 	return []string(Operator.Bundles)
 }
+
+func IsSupportedGpu(gpu *models.Gpu) bool {
+	return gpu.VendorID == nvidiaVendorID
```
paul-maidment, line 194: Do we support any NVidia GPU?
I know that consumer cards also have this vendor ID not just the data center versions.

internal/operators/openshiftai/openshift_ai_operator_test.go
```diff
@@ -171,5 +173,111 @@ var _ = Describe("Operator", func() {
 			},
 			true,
 		),
+		Entry(
+			"NVIDIA is supported even if not explicitly added",
+			map[string]string{
+				"OPENSHIFT_AI_SUPPORTED_GPUS": "1af4",
```
paul-maidment, line 179: Trying to understand why 1AF4 is here?

Looking up the vendor ID I can see that this is a Virtio device? (of which GPU can be one?)

https://devicehunt.com/search/type/pci/vendor/1AF4/device/any

paul-maidment: Also, make this a constant, same as the suggestion for the Nvidia one

paul-maidment: Do we need a test case for the scenario where the Nvidia GPU has not been explicitly added?

internal/operators/openshiftai/openshift_ai_operator_test.go
jhernand: I will make the `nvidiagpu.nvidiaVendorID` constant public and use it.

internal/operators/nvidiagpu/nvidia_gpu_operator.go
jhernand: The NVIDIA documentation only talks about data center GPUs: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/platform-support.html#supported-nvidia-data-center-gpus-and-systems . It doesn't say anything (positive or negative) about consumer cards.

We could improve this explicitly check the list of GPUs listed in that document. I am not sure if that is worth.

Anyhow, if we want to do that it should be in a different pull request. This pull request only moves the existing check to the `nvidiagpu` package. Please open a ticket if you believe this needs to be improved.

internal/operators/openshiftai/openshift_ai_operator_test.go
jhernand: As you discovered `14fa` is the vendor identifier of VirtIO devices. We use it only for testing purposes. Setting the `OPENSHIFT_AI_SUPPORTED_GPUS` environment variable to `14fa` developers and QE engineers can test the OpenShift AI operator feature, specially the validations, without requiring an actual NVIDIA GPU.

I will add the constant and the test case.

jhernand: Actually there is already a test case for the scenario where NVIDIA is not explicitly added to `OPENSHIFT_AI_SUPPORTED_GPUS`, look for `NVIDIA is supported even if not explicitly added` in this test.

internal/operators/nvidiagpu/nvidia_gpu_operator.go
paul-maidment: I mean, I tested the GPU operator a while back and it does seem to work with desktop GPUs to some extent, I had it running on an RTX3080ti.

I suppose it's a question of supported vs not supported though, so maybe there should be a list of supported devices.

paul-maidment:
If it becomes necessary to filter by device, this could be added later. I certainly don't think it's a common use case.

eifrach:
@jhernand my work is going to be delayed a bit. maybe you should merged this - I will rebase / refactor my work

jhernand:
> @jhernand my work is going to be delayed a bit. maybe you should merged this - I will rebase / refactor my work

@eifrach  is it OK if I reopen #7206 then?

eifrach:
Sure

jhernand:
> > @jhernand my work is going to be delayed a bit. maybe you should merged this - I will rebase / refactor my work
> 
> @eifrach is it OK if I reopen #7206 then?

@eifrach i wasn't able to reopen #7206 because I have changed the branch. I have instead opened a new pull request: #7227.

jhernand:
After discussing with the UX team it has been decided that the user will explicitly select the kind of GPUs that she wants to use. To do so the UI will present two different bundles "OpenShift AI with NVIDIA" and "OpenShift AI with AMD". So this is no longer needed.

PR #7220, opened 2025-01-23
MGMT-19102: Install to existing root when discovery boot device is persistent

carbonin:
This PR creates an API in the agent inventory and in the agent CR for storing the boot device type. The options for this type are ephemeral when an ISO or ipxe is being used for discovery, or persistent when discovery is being done from the same device we're installing to.

The install strategy will be chosen based on which kind of device is in use rather than choosing a single strategy for all installs using an environment variable.

The choice of strategy will still be communicated to the agent/installer by providing the coreos image to be used. 

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

Deployed assisted manually and installed a cluster to a host using the existing root filesystem and also using the ISO from the same assisted deployment.

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/v1beta1/agent_types.go
+++ b/api/v1beta1/agent_types.go
@@ -155,6 +155,7 @@ type HostDisk struct {
 type HostBoot struct {
 	CurrentBootMode string `json:"currentBootMode,omitempty"`
 	PxeInterface    string `json:"pxeInterface,omitempty"`
+	DeviceType      string `json:"deviceType,omitempty"`
 }
 
 type HostSystemVendor struct {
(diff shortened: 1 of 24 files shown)


carbonin:
hmmm maybe I need to break this up into the new inventory and then make the changes to the install to existing root conditionals to avoid breaking @rccrdpccl 's environment.

This would need to go in before the agent PR to collect the new inventory to ensure we have the correct APIs in the agent, but if this goes in and we're not collecting the inventory yet, the existing env var approach will break.

@rccrdpccl what do you think? Want me to split this out?

carbonin:
I'm just going to split this. It will take a bit longer to get in but it makes more sense to not break things.

PR #7222, opened 2025-01-23
MGMT-20060: Add OpenShift AI AMD operator bundle

jhernand:
This patch adds a new _Openshift AI (AMD)_ operator bundle that contains the OpenShift AI operator and the AMD GPU operator.

The patch also adds the KMM operator, which is a dependency of the AMD GPU operator.

https://issues.redhat.com/browse/MGMT-20060

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


jhernand:
/hold

Needs to be rebased after #7218 is merged.

jhernand:
/hold

Needs to be rebased after #7278 is merged.

internal/jq/jq_query.go
```diff
@@ -0,0 +1,206 @@
+package jq
```
rccrdpccl, line 1: Nice! I think this should be under `pkg`, so we could use it from other projects if needed. If we are concerned about timing I'm ok merging as is and we can extract it in a follow up PR.

internal/operators/openshiftai/openshift_ai_manifests_test.go
```diff
@@ -36,15 +69,138 @@ var _ = Describe("Manifest generation", func() {
...
+	It("Includes the data science cluster", func() {
+		// Convert the manifests into a list of objects:
+		_, manifests, err := operator.GenerateManifests(cluster)
+		Expect(err).ToNot(HaveOccurred())
+		objects := decodeManifests(manifests)
+
+		// Try to find the data science cluster:
+		var clusters []any
+		err = jqTool.Evaluate(
+			`.[] | select(.kind == "DataScienceCluster")`,
```
rccrdpccl, line 91: Although the jq tool is very nice, why would we need this to select based on kind? Wouldn't a simple for loop do? Am I missing something here?

jhernand: It can be a loop, but it isn't as simple as it may look, because we aren't importing (and don't want to import) the library that contains the "DataScienceCluster" type, so there would be multiple field presence and type checks, something like this:

```go
for _, object := range objects {
        objectMap, ok := object.(map[string]any)
        if ok {
                kind, ok := objectMap["kind"]
                if ok {
                        kindString, ok := kind.(string)
                        if ok {
                                if kindString == "DataScienceCluster" {
                                        clusters = append(clusters, cluster)
                                }
                        }
                }
        }
}
```

I think that the jq tool make that simpler to write and less error prone. It is also used below, for some slightly more complicated data extractions. I believe it is worth, but will remove it if you think it isn't acceptable.

internal/jq/jq_query.go
jhernand: Sure, I will move it.

jhernand: Done.

internal/operators/openshiftai/openshift_ai_manifests_test.go
rccrdpccl: The tool makes it much clearer, I agree with you I think it is worth it.

If I understand the use case, probably something like https://github.com/tidwall/gjson would be great, unfortunately as the manifest is in yaml it wouldn't work for us.

rccrdpccl:
@carbonin it seems the test is now triggering the error condition `no release image found for host`

PR #7227, opened 2025-01-27
NO-ISSUE: Recalculate operator dependencies before validations

jhernand:
Currently operator dependencies are only calculated when a cluster is created or updated. But certain dependencies are dynamic, and may change when new hosts are added. For example, if a cluster has the OpenShift AI operator installed, it will also require the NVIDIA GPU operator only if there are hosts that have NVIDIA GPUs. To support those dynamic dependencies this patch modifies the cluster monitor so that it recalculates the operator dependencies before checking validations.

- [X] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/cluster/refresh_status_preprocessor.go
+++ b/internal/cluster/refresh_status_preprocessor.go
@@ -2,15 +2,21 @@ package cluster
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"reflect"
 	"sort"
 	"strings"
 
+	"github.com/dustin/go-humanize/english"
 	"github.com/go-openapi/swag"
 	"github.com/openshift/assisted-service/internal/common"
+	eventsapi "github.com/openshift/assisted-service/internal/events/api"
 	"github.com/openshift/assisted-service/internal/host"
 	"github.com/openshift/assisted-service/internal/operators"
 	"github.com/openshift/assisted-service/internal/operators/api"
+	operatorcommon "github.com/openshift/assisted-service/internal/operators/common"
+	"github.com/openshift/assisted-service/internal/usage"
 	"github.com/openshift/assisted-service/models"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -30,23 +36,28 @@ type stringer interface {
 }
 
 type refreshPreprocessor struct {
-	log          logrus.FieldLogger
-	validations  []validation
-	conditions   []condition
-	operatorsAPI operators.API
+	log           logrus.FieldLogger
+	validations   []validation
+	conditions    []condition
+	operatorsAPI  operators.API
+	usageAPI      usage.API
+	eventsHandler eventsapi.Handler
 }
 
-func newRefreshPreprocessor(log logrus.FieldLogger, hostAPI host.API, operatorsAPI operators.API) *refreshPreprocessor {
+func newRefreshPreprocessor(log logrus.FieldLogger, hostAPI host.API, operatorsAPI operators.API, usageAPI usage.API,
+	eventsHandler eventsapi.Handler) *refreshPreprocessor {
 	v := clusterValidator{
 		log:     log,
 		hostAPI: hostAPI,
 	}
 
 	return &refreshPreprocessor{
-		log:          log,
-		validations:  newValidations(&v),
-		conditions:   newConditions(&v),
-		operatorsAPI: operatorsAPI,
+		log:           log,
+		validations:   newValidations(&v),
+		conditions:    newConditions(&v),
+		operatorsAPI:  operatorsAPI,
+		usageAPI:      usageAPI,
+		eventsHandler: eventsHandler,
 	}
 }
 
@@ -84,6 +95,17 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
 			Message: message,
 		})
 	}
+
+	// Before validating the operators we need to recalculate the dependencies because changes in the hosts may
+	// imply changes in the dependencies between operators. For example, if the OpenShift AI operator is enabled and
+	// a new host with an NVIDIA GPU has been discovered, then the NVIDIA GPU operator will need to be added as a
+	// dependency, and then we will need to validate that secure boot is disabled.
+	err = r.recalculateOperatorDependencies(ctx, c)
+	if err != nil {
+		err = errors.Wrapf(err, "failed to recalculate operator dependencies for cluster '%s'", c.clusterId)
+		return nil, nil, err
+	}
+
 	// Validate operators
 	results, err := r.operatorsAPI.ValidateCluster(ctx, c.cluster)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


internal/cluster/refresh_status_preprocessor.go
```diff
@@ -124,6 +141,103 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
...
+			addedOperators = append(addedOperators, currentOperator)
+		}
+	}
+	for _, previousOperator := range previousOperators {
+		if !operatorcommon.HasOperator(currentOperators, previousOperator.Name) {
+			deletedOperators = append(deletedOperators, previousOperator)
+		}
+	}
+	for _, addedOperator := range addedOperators {
+		err = c.db.Save(addedOperator).Error
```
jhernand, line 177: @rccrdpccl in the [previous](https://github.com/openshift/assisted-service/pull/7206#discussion_r1923262118) version of this pull request (now closed) @gamli75 was concerned 
that saving the cluster changes directly to DB will cause inconsistent values in the elastic. I added code below to also update the feature set. What else needs to be done to ensure consistency?

rccrdpccl: In order to keep elastic data consistent, we notify about important (cluster, hosts and infraenv, only fields we deem important as we want to keep event volume contained) state changes. Unfortunately, there isn't a centralized place for this, so the code used to do this is dispersed.
An example could be [updating hosts](https://github.com/openshift/assisted-service/pull/7227#discussion_r1930284119) in some circumstances, but in this case we should [update the cluster state](https://github.com/openshift/assisted-service/blob/398cf47615a80714a11c2bd722681fc8c70a7cc4/internal/cluster/cluster.go#L294).
Related: https://github.com/openshift/assisted-service/blob/398cf47615a80714a11c2bd722681fc8c70a7cc4/internal/stream/notification_stream.go

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -124,6 +141,103 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
 	return stateMachineInput, validationsOutput, nil
 }
 
+// recalculateOperatorDependencies calculates the operator dependencies and updates the database and the passed cluster
+// accordingly.
+func (r *refreshPreprocessor) recalculateOperatorDependencies(c *clusterPreprocessContext) error {
+	// Calculate and save the operators that have been added, updated or deleted:
+	previousOperators := c.cluster.MonitoredOperators
+	currentOperators, err := r.operatorsAPI.ResolveDependencies(c.cluster, c.cluster.MonitoredOperators)
```
paul-maidment, line 149: Just a quick clarifying question.

operatorsAPI.ResolveDependencies returns dependencies (more operators) for the operators being supplied in it's parameters?

internal/cluster/refresh_status_preprocessor.go
```diff
@@ -124,6 +141,103 @@ func (r *refreshPreprocessor) preprocess(ctx context.Context, c *clusterPreproce
...
+		err = c.db.Save(addedOperator).Error
+		if err != nil {
+			return errors.Wrapf(
+				err,
+				"failed to add operator '%s' to cluster '%s'",
+				addedOperator.Name, *c.cluster.ID,
+			)
+		}
+	}
+	for _, updatedOperator := range updatedOperators {
```
paul-maidment, line 186: I don't get the difference between an added and an updated operator in this code, it looks like exactly the same operations are performed on each?

specifically `c.db.Save(updatedOperator).Error` vs `c.db.Save(addedOperator).Error`

Why do we have this split if the behaviour is the same?

internal/cluster/refresh_status_preprocessor.go
jhernand: > `c.cluster.MonitoredOperators` are the current registered operators for the cluster?
> 

Yes, I think so.

> `operatorsAPI.ResolveDependencies` returns dependencies of the current cluster operators?
> 

I beleive that `ResolveDependencies` calculates the dependencies of the list of operators that is passed as a parameter. Then it merges that with the complete list of operators of the cluster, and returns the result. Note it returns everything: the operators explicitly added by the user as well as the ones automatically added to resolve the dependencies.

> I think the terms `previousoperators` and `currentoperators` are a little confusing
> 
> If I understand this code block correctly, we are doing the following
> 
> * Obtaining a list of the monitored operators for the cluster
> * Finding dependencies required for these monitored operators, this is again a list of MonitoredOperators
> * The resulting operators are then created, updated, deleted according to their presence in the `previous operators`
> 
> I wonder if `operators` and `operatorDependencies` (you refer to `currentOperators` as "operator dependencies" in the error message) or something along those lines would be better?
> 

The ideal names would be `lostOfOperatorsBeforeRecalculatingDependencies` and `listOfOperatorsAfterRecalculatingDependencies`, but that is too long. i can apply the renaming that you suggest, but take into account that "current..." is not only the dependencies; it is everythinng.

> In it's current form, it's quite hard to follow the intent of this code (but I think I got the basic gist?)

internal/cluster/refresh_status_preprocessor.go
jhernand: Operators are not only a name: they have also other properties that could change during the recalculation of dependencies. `updatedOperators` contains the list of operators that already existed before the recalculation and have changed (other than the name) after the recalculation. Currently we happen to do almost the same for both: we save them to the database with the `Save` method (although internally that `Save` method does an INSERT for added operators and an UPDATE for updated operators). I say "almost" because the error message generated if that failed is also different: `failed to add ...` vs `failed to update ...`.

When I wrote this initially I just wrote three independent loops because I didn't know at that point what I will be writing inside. The result ended up having only very minor differences. I think it makes sense, even if there was no difference. But merging them is also fine. If you insist a wee bit I will do that.

internal/cluster/refresh_status_preprocessor.go
jhernand: I am renaming the variables `operatorsBeforeResolve` and `operatorsAfterResolve`. Hope that clarifies it a bit.

internal/cluster/refresh_status_preprocessor.go
jhernand: Added code to send the notifications, please take another look @rccrdpccl.

jhernand:
After discussion with the UX team we have decided that the user will explicitly decide if she wants to use NVIDIA or AMD GPUs: she will select one of the "OpenShift AI with NVIDIA" or "OpenShift AI with AMD" operator bundles. That manes that this is no longer necessary, so I am closing it.

PR #7253, opened 2025-01-31
MGMT-14495: Stop removing the finalizers from BMH

MahnoorAsghar:
This code was added to handle OCPBUGS-7581, and it was meant to be removed after it was resolved. Its fix has long merged and was backported till OCP 4.12; therefore this code can now be removed.
Removing the BMH finalizer causes the BMH to instantly be deleted, which means that BMO is unable to do any necessary cleanup. In OCPBUGS-25927, BMO encounters an error while attempting to update the status of an already deleted BMH:
2024-11-13T07:28:35.215792213Z {"level":"info","ts":1731482915.2157853,"logger":"controllers.BareMetalHost","msg":"saving host status","baremetalhost":{"name":"gateway-1.workload.fc18.lab","namespace":"openshift-machine-api"},"provisioningState":"unmanaged","operational status":"discovered","provisioning state":"deleting"}
2024-11-13T07:28:35.220351574Z {"level":"error","ts":1731482915.2201939,"msg":"Reconciler error","controller":"baremetalhost","controllerGroup":"metal3.io","controllerKind":"BareMetalHost","BareMetalHost":{"name":"gateway-1.workload.fc18.lab","namespace":"openshift-machine-api"},"namespace":"openshift-machine-api","name":"gateway-1.workload.fc18.lab","reconcileID":"dbda2c02-2b4d-4a5e-ac98-531d4fc46263",
"error":"failed to save host status after \"unmanaged\": Operation cannot be fulfilled on baremetalhosts.metal3.io \"gateway-1.workload.fc18.lab\": StorageError: invalid object, Code: 4, Key: /kubernetes.io/metal3.io/baremetalhosts/openshift-machine-api/gateway-1.workload.fc18.lab, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: c207fd48-3dc4-455a-ae12-b4e44fda2285, UID in object meta: ","errorVerbose":"Operation cannot be fulfilled on baremetalhosts.metal3.io \"gateway-1.workload.fc18.lab\": StorageError: invalid object, Code: 4, Key: /kubernetes.io/metal3.io/baremetalhosts/openshift-machine-api/gateway-1.workload.fc18.lab, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: c207fd48-3dc4-455a-ae12-b4e44fda2285, UID in object meta: \nfailed to save host status after \"unmanaged\"\n

- Should this PR be tested by the reviewer? If possible, it would be helpful
- Is this PR relying on CI for an e2e test run? Yes
- Should this PR be tested in a specific environment? With baremetal-operator running and when assisted issues a delete request for a BareMetalHost.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [] None

## How was this code tested?
- [x] No tests needed

## Checklist

This PR should be backported till 4.12

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -348,13 +348,6 @@ func deleteBMHForMachine(ctx context.Context, spokeClient client.Client, machine
 		return errors.Wrapf(err, "failed to delete BMH %s", bmhNSName)
 	}
 
-	// TODO: remove this once OCPBUGS-7581 is fixed
-	patch := client.MergeFrom(bmh.DeepCopy())
-	bmh.ObjectMeta.Finalizers = nil
-	if err = client.IgnoreNotFound(spokeClient.Patch(ctx, bmh, patch)); err != nil {
-		return errors.Wrapf(err, "failed to remove BMH %s finalizers", bmhNSName)
-	}
-
 	return nil
 }
 


zaneb:
Let's open a new ticket (can be in the MGMT project) for this rather than use [OCPBUGS-45927](https://issues.redhat.com//browse/OCPBUGS-45927), since this is unrelated to the original issue in that ticket.

> This PR should be backported till 4.12

The release cycle for MCE is orthogonal to that of OCP (multiple versions of MCE can install multiple versions of OCP). The fix for [OCPBUGS-7581](https://issues.redhat.com/browse/OCPBUGS-7581) is present in OCP 4.14.0, 4.13.2, and 4.12.21, so this patch is safe in any version of MCE that does not support installing earlier versions of OCP than those.

carbonin:
I have a ticket for removing this https://issues.redhat.com/browse/MGMT-14495

I think it should be safe at this point to remove this.
MCE works based on y-streams so I think we'd want to ensure we're only removing this for versions that only support 4.13 and up.

MCE support matrix is N-2 - N+1 (N being the OCP relase the corresponding ACM version releases with).
I think this means we could go as far back as ACM 2.10 with this (removing the hack), but IIRC we have some support exceptions to allow 4.12 with 2.10 so if we want to backport we should probably only go back to 2.11.

@gamli75 what do you think?

carbonin:
>MCE works based on y-streams so I think we'd want to ensure we're only removing this for versions that only support 4.13 and up.

To be clear, for this I mean I wouldn't want a version of MCE that "supports 4.12" only actually work for some 4.12 spoke release versions unless someone has strong feelings to the contrary about this.

gamli75:
> I have a ticket for removing this https://issues.redhat.com/browse/MGMT-14495
> 
> I think it should be safe at this point to remove this. MCE works based on y-streams so I think we'd want to ensure we're only removing this for versions that only support 4.13 and up.
> 
> MCE support matrix is N-2 - N+1 (N being the OCP relase the corresponding ACM version releases with). I think this means we could go as far back as ACM 2.10 with this (removing the hack), but IIRC we have some support exceptions to allow 4.12 with 2.10 so if we want to backport we should probably only go back to 2.11.
> 
> @gamli75 what do you think?

@carbonin  What will happen if we keep this hack in ACM 2.12 and below and remove it only for ACM 2.13? If there is no harm - I think it will be better to leave it as is instead of struggling if the fix was merged and to which z-stream

carbonin:
Yeah, I think the ACM 2.12 support matrix should officially only be 4.15-4.18 so I think it's pretty unlikely that we'd see anyone installing early 4.12 z streams and also using 2.12 or up.

carbonin:
Ah just reread Zane's comment. Looks like the first `.0` z-stream the fix is in is actually 4.14. Let's definitely only backport to 2.12 then (if at all)

carbonin (approved):

PR #7254, opened 2025-01-31
MGMT-19844: Validate the internal ignition override

carbonin:
Users have seen some situations where they are unable to download the
discovery ISO because the ignition doesn't generate correctly on the
service side.

The only explanation seems to be that there is a malformed override and
the only override we're adding in all of these cases is the internal one
which also happens to not be validated.

Relates to https://issues.redhat.com/browse/MGMT-19844

- [ ] New Feature 
- [x] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4984,7 +4984,7 @@ func (b *bareMetalInventory) validateInfraEnvCreateParams(ctx context.Context, p
 		}
 	}
 
-	if err = b.validateInfraEnvIgnitionParams(ctx, params.InfraenvCreateParams.IgnitionConfigOverride); err != nil {
+	if err = b.validateInfraEnvIgnitionParams(ctx, params.InfraenvCreateParams.IgnitionConfigOverride, nil); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


CrystalChun (approved):
/lgtm

This is great! Maybe not for this PR, but in the future, do you think it's a good idea to try and merge the ignition override and the internal ignition override as part of the validation?

carbonin:
I'm not sure it's common enough to have both and I also don't know if merging them in the absence of the base ignition is super meaningful. I'm just hoping this will give us enough information to figure out what's going on in these cases then actually fix whatever that problem is.

gamli75:
I think it's worth backporting it to older versions - ACM 2.12, 2.11 and 2.10

carbonin:
Doing the cherry-picks manually because the bot did a terrible job of cherry-picking (the diff is just wrong).
`git cherry-pick` manually seems to handle it fine though :man_shrugging:

PR #7279, opened 2025-02-05
MGMT-19649: Monitored host count shouldn't be gauge

rivkyrizel:
The `service_assisted_installer_monitored_hosts` metric was previously implemented as a gauge, which could lead to misleading insights. Since the host monitoring loop runs every few seconds while Prometheus scrapes at a different interval, some monitored host data may be lost.

Similarly, `service_assisted_installer_monitored_clusters` faced the same issue.

To improve the accuracy of these metrics, they have been changed from a gauge to a counter, ensuring Prometheus records cumulative increases rather than snapshot values. This change provides a more reliable view of monitored hosts and clusters over time.

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

/cc @gamli75 @rccrdpccl

### DIFF
--- a/internal/metrics/metricsManager.go
+++ b/internal/metrics/metricsManager.go
@@ -117,8 +117,8 @@ type MetricsManager struct {
 	serviceLogicClusterValidationFailed                *prometheus.CounterVec
 	serviceLogicClusterValidationChanged               *prometheus.CounterVec
 	serviceLogicFilesystemUsagePercentage              *prometheus.GaugeVec
-	serviceLogicMonitoredHosts                         *prometheus.GaugeVec
-	serviceLogicMonitoredClusters                      *prometheus.GaugeVec
+	serviceLogicMonitoredHosts                         *prometheus.CounterVec
+	serviceLogicMonitoredClusters                      *prometheus.CounterVec
 	collectors                                         []prometheus.Collector
 }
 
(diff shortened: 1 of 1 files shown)


internal/metrics/metricsManager.go
```diff
@@ -476,11 +478,11 @@ func (m *MetricsManager) FileSystemUsage(usageInPercentage float64) {
 }
 
 func (m *MetricsManager) MonitoredHostsCount(monitoredHosts int64) {
```
rccrdpccl, line 480: I think ideally we should change approach: increase by one the counter each time we process an host. This way we'll have an idea closer to reality of how long it takes to monitor hosts
At the moment we are adding the [whole count of monitored hosts after processing](https://github.com/openshift/assisted-service/blob/2fbd268905ee504013b6130adb53ae96e4e012a4/internal/host/monitor.go#L268)
Similarly for clusters

I think we should also add an histogram, to have more details about single host processing, should we maybe tackle it in another PR/task? WDYT @adriengentil @gamli75

rivkyrizel: I'll add a task to introduce a histogram to capture how long each host takes to be monitored. 
I think increasing the counter for each processed host might be redundant since the histogram will already provide this level of granularity. WDYT?

PR #7281, opened 2025-02-06
MGMT-19840: Gather operational metrics from installercache

paul-maidment:
We implement the following metrics:
Record cache hits and misses

The purpose of this metric is to measure efficiency of the cache. This metric should be a counter. The total sum should be the equivalent to the sum of hits+misses (each request served). Errors should be logged instead.

 

assisted_installer_release_cache{hit=(true|false),release="\d.\d"} 

 
Labels

    hit: hit should either be true or false
    release: release is major and minor openshift version, so that cardinality is contained. if we can't retrive major/minor, we label it as `unknown`

Record cache evictions

This metric is a counter. The purpose of this metric is to count the number of eviction attempts. For each eviction attempt we should just count one with success=true label if the operation evicted at least one file, false if evicted 0 files). Individual file deletion and possible errors should be logged instead.

assisted_installer_release_cache_eviction{success=(true|false)}

Labels
    success: is true if the operation removed at least one file, false otherwise

This, combined with the event based metrics gathered in openshift#7156 should provide enough information to track the behaviour of the cache.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/installercache/installercache.go
+++ b/internal/installercache/installercache.go
@@ -36,6 +36,7 @@ type Installers struct {
 	eventsHandler   eventsapi.Handler
 	diskStatsHelper metrics.DiskStatsHelper
 	config          Config
+	metricsAPI      metrics.API
 }
 
 type Size int64
@@ -116,7 +117,7 @@ func (rl *Release) Cleanup(ctx context.Context) error {
 }
 
 // New constructs an installer cache with a given storage capacity
-func New(config Config, eventsHandler eventsapi.Handler, diskStatsHelper metrics.DiskStatsHelper, log logrus.FieldLogger) (*Installers, error) {
+func New(config Config, eventsHandler eventsapi.Handler, metricsAPI metrics.API, diskStatsHelper metrics.DiskStatsHelper, log logrus.FieldLogger) (*Installers, error) {
 	if config.MaxCapacity > 0 && config.MaxReleaseSize == 0 {
 		return nil, fmt.Errorf("config.MaxReleaseSize (%d bytes) must not be zero", config.MaxReleaseSize)
 	}
@@ -128,6 +129,7 @@ func New(config Config, eventsHandler eventsapi.Handler, diskStatsHelper metrics
 		eventsHandler:   eventsHandler,
 		diskStatsHelper: diskStatsHelper,
 		config:          config,
+		metricsAPI:      metricsAPI,
 	}, nil
 }
 
@@ -135,6 +137,11 @@ func New(config Config, eventsHandler eventsapi.Handler, diskStatsHelper metrics
 // the referenced release image. Tries the mirror release image first if it's set. It is safe for concurrent use. A cache of
 // binaries is maintained to reduce re-downloading of the same release.
 func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSecret string, ocRelease oc.Release, ocpVersion string, clusterID strfmt.UUID) (*Release, error) {
+	majorMinorVersion, err := ocRelease.GetMajorMinorVersion(i.log, releaseID, releaseIDMirror, pullSecret)
+	if err != nil {
+		i.log.Warnf("unable to get majorMinorVersion to record metric for %s falling back to full URI", releaseID)
+		majorMinorVersion = "unknown"
+	}
 	for {
 		select {
 		case <-ctx.Done():
@@ -142,6 +149,7 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
 		default:
 			release, err := i.get(releaseID, releaseIDMirror, pullSecret, ocRelease, ocpVersion, clusterID)
 			if err == nil {
+				i.metricsAPI.InstallerCacheGetReleaseCached(majorMinorVersion, release.cached)
 				return release, nil
 			}
 			_, isCapacityError := err.(*errorInsufficientCacheCapacity)
@@ -303,6 +311,7 @@ func (i *Installers) evict() bool {
 		}
 		evicted = true
 	}
+	i.metricsAPI.InstallerCacheReleaseEvicted(evicted)
 	return evicted
 }
 
@@ -334,8 +343,7 @@ func (i *Installers) pruneExpiredHardLinks(links []*fileInfo, gracePeriod time.D
 		grace := graceTime.Unix()
 		if finfo.info.ModTime().Unix() < grace {
 			i.log.Infof("attempting to prune hard link %s", finfo.path)
-			err := os.Remove(finfo.path)
-			if err != nil {
+			if err := os.Remove(finfo.path); err != nil {
 				i.log.WithError(err).Errorf("failed to prune hard link %s", finfo.path)
 			}
 		}
--- a/internal/metrics/metricsManager.go
+++ b/internal/metrics/metricsManager.go
@@ -3,6 +3,7 @@ package metrics
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/alecthomas/units"
@@ -39,6 +40,8 @@ const (
 	counterFilesystemUsagePercentage              = "assisted_installer_filesystem_usage_percentage"
 	histogramMonitoredHostsDurationMs             = "assisted_installer_monitored_hosts_duration_ms"
 	histogramMonitoredClustersDurationMs          = "assisted_installer_monitored_clusters_duration_ms"
+	counterInstallerReleaseCache                  = "assisted_installer_release_cache"
+	counterInstallerReleaseCacheEviction          = "assisted_installer_release_cache_eviction"
 )
 
 const (
@@ -61,6 +64,8 @@ const (
 	counterDescriptionFilesystemUsagePercentage              = "The percentage of the filesystem usage by the service"
 	histogramDescriptionMonitoredHostsDurationMs             = "Histogram/sum/count of monitored hosts duration (ms)"
 	histogramDescriptionMonitoredClustersDurationMs          = "Histogram/sum/count of monitored clusters duration (ms)"
+	counterDescriptionInstallerReleaseCache                  = "Counts the cache hit status for the labelled release"
+	counterDescriptionInstallerReleaseCacheEviction          = "Counts the number of times that at least one release was evicted"
 )
 
 const (
@@ -77,6 +82,9 @@ const (
 	imageLabel                 = "imageName"
 	hosts                      = "hosts"
 	clusters                   = "clusters"
+	labelCacheHit              = "hit"
+	labelReleaseID             = "releaseId"
+	labelSuccess               = "success"
 )
 
 type API interface {
@@ -94,6 +102,8 @@ type API interface {
 	FileSystemUsage(usageInPercentage float64)
 	MonitoredHostsDurationMs(monitoredHostsMillis float64)
 	MonitoredClustersDurationMs(monitoredClustersMillis float64)
+	InstallerCacheGetReleaseCached(releaseId string, cacheHit bool)
+	InstallerCacheReleaseEvicted(success bool)
 }
 
 type MetricsManager struct {
@@ -119,7 +129,10 @@ type MetricsManager struct {
 	serviceLogicFilesystemUsagePercentage              *prometheus.GaugeVec
 	serviceLogicMonitoredHostsDurationMs               *prometheus.HistogramVec
 	serviceLogicMonitoredClustersDurationMs            *prometheus.HistogramVec
-	collectors                                         []prometheus.Collector
+	serviceLogicInstallerReleaseCache                  *prometheus.CounterVec
+	serviceLogicInstallerReleaseEvicted                *prometheus.CounterVec
+
+	collectors []prometheus.Collector
 }
 
 var _ API = &MetricsManager{}
@@ -290,6 +303,22 @@ func NewMetricsManager(registry prometheus.Registerer, eventsHandler eventsapi.H
 			Help:      histogramDescriptionMonitoredClustersDurationMs,
 			Buckets:   []float64{10, 100, 200, 500, 1000, 10000, 30000},
 		}, []string{}),
+
+		serviceLogicInstallerReleaseCache: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: namespace,
+				Subsystem: subsystem,
+				Name:      counterInstallerReleaseCache,
+				Help:      counterDescriptionInstallerReleaseCache,
+			}, []string{labelReleaseID, labelCacheHit}),
+
+		serviceLogicInstallerReleaseEvicted: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: namespace,
+				Subsystem: subsystem,
+				Name:      counterInstallerReleaseCacheEviction,
+				Help:      counterDescriptionInstallerReleaseCacheEviction,
+			}, []string{labelSuccess}),
 	}
 
 	m.collectors = append(m.collectors, newDirectoryUsageCollector(metricsManagerConfig.DirectoryUsageMonitorConfig.Directories, diskStatsHelper, log))
@@ -314,6 +343,8 @@ func NewMetricsManager(registry prometheus.Registerer, eventsHandler eventsapi.H
 		m.serviceLogicFilesystemUsagePercentage,
 		m.serviceLogicMonitoredHostsDurationMs,
 		m.serviceLogicMonitoredClustersDurationMs,
+		m.serviceLogicInstallerReleaseCache,
+		m.serviceLogicInstallerReleaseEvicted,
 	)
 
 	for _, collector := range m.collectors {
@@ -487,4 +518,12 @@ func (m *MetricsManager) MonitoredClustersDurationMs(monitoredClustersMs float64
 
 func bytesToGib(bytes int64) int64 {
 	return bytes / int64(units.GiB)
+}
+
+func (m *MetricsManager) InstallerCacheGetReleaseCached(releaseId string, cacheHit bool) {
+	m.serviceLogicInstallerReleaseCache.WithLabelValues(releaseId, fmt.Sprintf("%t", cacheHit)).Inc()
+}
+
+func (m *MetricsManager) InstallerCacheReleaseEvicted(success bool) {
+	m.serviceLogicInstallerReleaseEvicted.WithLabelValues(fmt.Sprintf("%t", success)).Inc()
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -492,7 +492,7 @@ func main() {
 	Options.BMConfig.S3EndpointURL = newUrl
 
 	Options.InstallerCacheConfig.CacheDir = filepath.Join(Options.GeneratorConfig.GetWorkingDirectory(), "installercache")
-	installerCache, err := installercache.New(Options.InstallerCacheConfig, eventsHandler, diskStatsHelper, log)
+	installerCache, err := installercache.New(Options.InstallerCacheConfig, eventsHandler, metricsManager, diskStatsHelper, log)
 	failOnError(err, "failed to instantiate installercache")
 
 	generator := generator.New(log, objectHandler, Options.GeneratorConfig, providerRegistry, manifestsApi, eventsHandler, installerCache)
--- a/internal/ignition/installmanifests_test.go
+++ b/internal/ignition/installmanifests_test.go
@@ -94,6 +94,7 @@ var _ = Describe("Bootstrap Ignition Update", func() {
 		manifestsAPI   *manifestsapi.MockManifestsAPI
 		eventsHandler  *eventsapi.MockHandler
 		installerCache *installercache.Installers
+		metricsAPI     *metrics.MockAPI
 	)
 
 	BeforeEach(func() {
@@ -105,12 +106,13 @@ var _ = Describe("Bootstrap Ignition Update", func() {
 		err1 = os.WriteFile(examplePath, []byte(bootstrap1), 0600)
 		Expect(err1).NotTo(HaveOccurred())
 		ctrl = gomock.NewController(GinkgoT())
+		metricsAPI = metrics.NewMockAPI(ctrl)
 		installerCacheConfig := installercache.Config{
 			CacheDir:       filepath.Join(workDir, "some-dir", "installercache"),
 			MaxCapacity:    installercache.Size(5),
 			MaxReleaseSize: installercache.Size(5),
 		}
-		installerCache, err = installercache.New(installerCacheConfig, eventsHandler, metrics.NewOSDiskStatsHelper(logrus.New()), logrus.New())
+		installerCache, err = installercache.New(installerCacheConfig, eventsHandler, metricsAPI, metrics.NewOSDiskStatsHelper(logrus.New()), logrus.New())
 		Expect(err).NotTo(HaveOccurred())
 		mockS3Client = s3wrapper.NewMockAPI(ctrl)
 		manifestsAPI = manifestsapi.NewMockManifestsAPI(ctrl)
@@ -262,6 +264,7 @@ SV4bRR9i0uf+xQ/oYRvugQ25Q7EahO5hJIWRf4aULbk36Zpw3++v2KFnF26zqwB6
 		ctrl           *gomock.Controller
 		manifestsAPI   *manifestsapi.MockManifestsAPI
 		eventsHandler  eventsapi.Handler
+		metricsAPI     *metrics.MockAPI
 		installerCache *installercache.Installers
 	)
 
@@ -286,12 +289,13 @@ SV4bRR9i0uf+xQ/oYRvugQ25Q7EahO5hJIWRf4aULbk36Zpw3++v2KFnF26zqwB6
 		ctrl = gomock.NewController(GinkgoT())
 		manifestsAPI = manifestsapi.NewMockManifestsAPI(ctrl)
 		eventsHandler = eventsapi.NewMockHandler(ctrl)
+		metricsAPI = metrics.NewMockAPI(ctrl)
 		installerCacheConfig := installercache.Config{
 			CacheDir:       filepath.Join(workDir, "some-dir", "installercache"),
 			MaxCapacity:    installercache.Size(5),
 			MaxReleaseSize: installercache.Size(5),
 		}
-		installerCache, err = installercache.New(installerCacheConfig, eventsHandler, metrics.NewOSDiskStatsHelper(logrus.New()), logrus.New())
+		installerCache, err = installercache.New(installerCacheConfig, eventsHandler, metricsAPI, metrics.NewOSDiskStatsHelper(logrus.New()), logrus.New())
 		Expect(err).NotTo(HaveOccurred())
 	})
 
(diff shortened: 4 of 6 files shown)


internal/metrics/metricsManager.go
```diff
@@ -61,6 +69,14 @@ const (
...
 	counterDescriptionMonitoredClusters                      = "Number of clusters monitored by cluster monitor"
+	counterDescriptionInstallerCachePrunedHardlink           = "Counts the number of times the installercache pruned a hardlink for being too old"
+	counterDescriptionInstallerCacheGetReleaseOK             = "Counts the number of times that a release was fetched succesfully"
+	counterDescriptionInstallerCacheGetReleaseTimeout        = "Counts the number of times that a release timed out or had the context cancelled"
+	counterDescriptionInstallerCacheGetReleaseError          = "Counts the number of times that a release fetch resulted in error"
+	counterDescriptionInstallerCacheReleaseCached            = "Counts the number of times that a release was found in the cache"
+	counterDescriptionInstallerCacheReleaseExtracted         = "Counts the number of times that a release was extracted"
+	counterDescriptionInstallerCacheTryEviction              = "Counts the number of times that the eviction function was called"
+	counterDescriptionInstallerCacheReleaseEvicted           = "Counts the number of times that a release was evicted"
```
rccrdpccl, line 79: I think this could be better handled by adding relevant labels instead of multiple metrics.

`assisted_installer_cache_release_evicted{success="true|false"}` will cover evictions (failed and successfull)
`assisted_installer_cache_get_release{hit="true|false"}` will cover cache hits

Do we need to know if hardlink are pruned or not? Can it be inferred? What value does it add? 
Same for errors, do we want to add a metrics for it? What value would we add? If needed we could add another metrics for error and types of error.
About extraction, wouldn't a cache miss actually count as extraction? If we want to know the time taken we can always add an histogram for it

internal/metrics/metricsManager.go
```diff
@@ -77,6 +93,10 @@ const (
 	imageLabel                 = "imageName"
 	hosts                      = "hosts"
 	clusters                   = "clusters"
+	release                    = "release"
+	status                     = "status"
```
gamli75, line 97: Could you please tell me why you added those labels? I didn't see that you used them in the code

paul-maidment: You are right, we can remove these.

I was originally working on a different approach (actually something similar to what @rccrdpccl is suggesting now) but decided to make things simpler.

internal/metrics/metricsManager.go
paul-maidment: > Do we need to know if hardlink are pruned or not? ... What value does it add?
I have a strong hunch that we don't use the hardlink pruning in practice. 
I want to measure this stat to prove/disprove this hunch.

> Can it be inferred
As we have more than  one way of cleaning a hardlink, I don't think so. We could log something but that's really just a 'cheap' alternative to a metric.

> Same for errors, do we want to add a metrics for it? What value would we add? If needed we could add another metrics for error and types of error.

My thought on this is that it would be ideal to know the number of requests made for a release and be able to distinguish between "success", "failure" and "time out". I don't care as much about the specific errors, only their type (success, fail, time out) as this provides some indication about the health of the cache and percentage of requests that had a given outcome.

If you like, we could maybe take success/fail and put them together into a "not timed out" category, but I think this would be odd.

> wouldn't a cache miss actually count as extraction?
Maybe I am missing something here (pardon the pun) but I don't think so, this counter exists to count the number of extractions vs cached for given releases. These counts are intended to show the "effectiveness" of the cache (the more hits you have -- cached items) the more effective the cache. 

I think this is distinct from the extraction time which we record elsewhere and just because we have an extraction time does not negate the need to assess how effective the cache is.

rccrdpccl: > I did originally consider this, I felt that this was actually more complicated than this simpler approach that makes it easy to register a metric in the place where something of interest occurs and I don't think there are significant disadvantages honestly.

Is it really relevant the effort taken to implement this? I think it would be in both case low effort.
Querying this metric though it will be much harder if you have two different metrics, potentially with diverging labels over time.
Rule of thumb is to try keep one metric until cardinality is low, if cardinality becomes a problem, then we should split.

> I have a strong hunch that we don't use the hard link pruning in practice.
I want to measure this metric over a period of time to prove/disprove this hunch.
After some time a decision could be made to remove this metric (and potentially hard link pruning.)

Do we need a metric to confirm this? Wouldn't be enough with logs? Why a metric should be preferred to logs?

> A lot of "time out" counts would indicate problems with the cache that would need to be addressed. So I think this is quite an important metric.

Again, I think this would fit more into logs - unless this is something we really want to track and draw stats from it. I would say what we want

> A metric of the duration of an extraction is not the same and serves another purpose, to educate future decisions based on the processing time for an extraction.

In this PR this metric is a counter though, right?


Can we take a step back, and try to clearly define what we want to measure and why? It will make it easier to define how we want to record the metrics

paul-maidment: > In this PR this metric is a counter though, right?

Yes and I argue that this counter is necessary.

I also accept that we might want an _additional_ gauge metric to account for extraction time. Though we do already have this data in metric events.

> Can we take a step back, and try to clearly define what we want to measure and why? It will make it easier to define how we want to record the metrics

As I see it, we are interested in getting some information that will help us to tune cache behaviour moving forward, to assert that the cache is working correctly and to gather some metrics that may assist in future enhancement.

**Tuning cache behavior**
Hits/Misses (counts) will be used to determine the efficacy of the cache, how does the efficacy vary over time and with various settings etc? When using this metric, we are attempting to 'tune' the cache to favour hits.

**Proving the cache works**
We already have a collector for a gauge that shows disk space usage within the data directory over time. We are looking to see if this indicates that we are within the cache limit.
Logging timeouts would allow us to create alerting based on logs.

**Measuring for improved knowledge**
The amount of time taken to extract a release is gathered already by metrics added in another PR.
Extraction times provide useful information on how long it takes to extract a release (therefore how long it will take to `Get` a release)

@rccrdpccl does this cover it or is there anything else you would add?

rccrdpccl: >Proving the cache works
We already have a collector for a gauge that shows disk space usage within the data directory over time. We are looking to see if this indicates that we are within the cache limit.
Logging timeouts would allow us to create alerting based on logs.

Proving that something works it seems more the job of logs, maybe even at debug level (for successful prove, warning/error for errors that prevent the feature from working)

> Measuring for improved knowledge
The amount of time taken to extract a release is gathered already by metrics added in another PR.
Extraction times provide useful information on how long it takes to extract a release (therefore how long it will take to Get a release)

How would you measure for improved knowledge? What data would you need? What would you do with it? I think answering tho these questions will help us define what data for "improved knowledge" we want


> ** Knowing if something fails **
The biggest failure risk would be a deadlock of the Mutex that we use within the installercache. In cases where there was a problem in this area, would would see context timeouts start to occur.

> I presently propose a metric / counter for this but, provided we can create alerts based on log entries, would be happy to move this to logs.

I think an error metric would be interesting, as we cannot alert by logs, however it would not be straightforward. How would it work? Just counting some specific errors? Wouldn't that mislead us if other errors are happening? How can we detect deadlocks?

paul-maidment: > I think an error metric would be interesting, as we cannot alert by logs, however it would not be straightforward. How would it work? Just counting some specific errors? Wouldn't that mislead us if other errors are happening? How can we detect deadlocks?

As far as I know, we would be able to detect deadlocks when the "retry" on a `GET` timed out. This would indicate that nothing was able to free up space in order to write the release.

paul-maidment: > Can you provide a realistic example on how we would use this metric?

Referring to the metric for pruned hardlinks.

Given that we only want to determine if hardlinks are ever automatically pruned and this is not in any way critical, I agree we can remove this metric and fall back to using logs instead.

> What would be a realistic use case for this metric? What I can think is "monitoring extraction time, in case it would increase in some release and/or under some condition" - however I struggle how can we perform any action on this. What should we do if we find out the time increases by 100%?

The intent of this was simply 'cache hit = false' 

This has been addressed by removing this and doing everything we need to do within another metric.

internal/installercache/installercache.go
```diff
@@ -247,6 +256,7 @@ func (i *Installers) shouldEvict(totalUsed int64) (shouldEvict bool) {
 //
 // Locking must be done outside evict() to avoid contentions.
 func (i *Installers) evict() bool {
+	i.metricsAPI.InstallerCacheTryEviction()
```
rccrdpccl, line 259: Why do we want to register this metric differently than the other eviction metric? What is the idea behind it? Could you please explain a scenario where we'd need this info in this format (as opposed to not having this metric, compact it in the same metric, or possibly even switching metric type)?

internal/metrics/metricsManager.go
```diff
@@ -486,3 +537,24 @@ func (m *MetricsManager) MonitoredClusterCount(monitoredClusters int64) {
...
+func (m *MetricsManager) InstallerCacheGetReleaseOK(releaseId string, cacheHit bool) {
+	m.serviceLogicInstallerCacheGetRelease.WithLabelValues(labelValueOK, releaseId, fmt.Sprintf("%t", cacheHit)).Inc()
+}
+
+func (m *MetricsManager) InstallerCacheGetReleaseTimeout(releaseId string, cacheHit bool) {
+	m.serviceLogicInstallerCacheGetRelease.WithLabelValues(labelValueTimeout, releaseId, fmt.Sprintf("%t", cacheHit)).Inc()
+}
+
+func (m *MetricsManager) InstallerCacheGetReleaseError(releaseId string, cacheHit bool) {
+	m.serviceLogicInstallerCacheGetRelease.WithLabelValues(labelValueError, releaseId, fmt.Sprintf("%t", cacheHit)).Inc()
```
rccrdpccl, line 551: What kind of values would `releaseId` get?

internal/metrics/metricsManager.go
```diff
@@ -288,6 +312,30 @@ func NewMetricsManager(registry prometheus.Registerer, eventsHandler eventsapi.H
...
 			Help:      counterDescriptionMonitoredClusters,
 		}, []string{hosts}),
+
+		serviceLogicInstallerCacheGetRelease: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: namespace,
+				Subsystem: subsystem,
+				Name:      counterInstallerCacheGetRelease,
+				Help:      counterDescriptionInstallerCacheGetRelease,
+			}, []string{labelStatus, labelReleaseID, labelCacheHit}),
```
rccrdpccl, line 322: I think the label shouldn't be status, but rather error in case of error, empty otherwise

internal/installercache/installercache.go
paul-maidment: It is possible that you may attempt to evict and yet find no candidates. 

I wanted to ensure that we have a simple count of the number of times that eviction has been attempted.
This would be a good indicator of how often we end up with storage pressure.

internal/metrics/metricsManager.go
paul-maidment: Why do you think that?

internal/metrics/metricsManager.go
rccrdpccl: This could have huge cardinality, can we maybe simplify it to major-minor? Would that provide enough info?

internal/installercache/installercache.go
rccrdpccl: If there's storage pressure (for which IIRC we already have a metric for), we could check if there's any eviction going on (metric being worked on in this PR). If the eviction metric reports 0 evictions, something is fishy and we can check logs, where we could place a warning for the conditions you're worried about.

How would adding this metric help with this process?

internal/metrics/metricsManager.go
paul-maidment: The real purpose of this dimension is to report the outcome of the request and add it to the count actually

There are three status we are tracking

"ok" - indicating success in the request
"error" - indicating that an error (distinctly not timeout) occurred.
"timeout" - indicating that a timeout occurred and that perhaps there are issues with the Mutex

The sum of these counts should be equal to the total number of requests made. This is why we include them all, so that we can have the full context, especially if there are outliers.

I actually think we should keep it this way.

If you believe otherwise, please explain why?

internal/metrics/metricsManager.go
paul-maidment: Fair point. Yes, I will look into doing that. Yes I suppose that combined with logging an error we would provide enough of a clue to anyone monitoring the application. 

I will change this.

internal/installercache/installercache.go
paul-maidment: > If there's storage pressure (for which IIRC we already have a metric for)

Which metric is that?

paul-maidment: OK, let's remove this metric

internal/metrics/metricsManager.go
rccrdpccl: My suggestion is only about semantics, it would technically be the same (just replacing "status" label with "error", values would be `{"", "unknown", "timeout"}` or something like that)

rccrdpccl: As mentioned before, I still think reporting errors in this case doesn't add any actionable value as metric.
However, if we want to go down this route, I'd say that's exactly what we are trying to do: error reporting.
From a metric point of view, we're not interested in "return status" if it's ok or not. We just want to know when it worked if it was cached or not.

Do we want to add a separate metric just to report errors? It might have a clearer meaning, but how would that add value?

paul-maidment: >  doesn't add any actionable value as metric 

Why is this not actionable?

Am I mistaken to think that we could have an alert on the timeout status that would be actionable?
We already discussed logging as an option and that's a non starter because you can't trigger alerts based on this (according to what you said above)

I don't think cache status in the case of this error even comes into it. If you have a timeout, the only way this can happen is due to a failed extraction due to capacity issues.

So I think this is leaning more towards "a timeout error occurred" and we just ditch the other two status from the metric, do you agree?

internal/metrics/metricsManager.go
paul-maidment: I thought about an edge case here and wonder how we should handle (given cardinality concerns)

Obtaining a majorMinorVersion can fail

```
majorMinorVersion, err := ocRelease.GetMajorMinorVersion(i.log, releaseID, releaseIDMirror, pullSecret)
			if err != nil {
				i.log.Warnf("unable to get majorMinorVersion to record metric for %s falling back to full URI", releaseID)
				majorMinorVersion = releaseID
			}
```

In the snippet above, I go for what I feel is the most sensible option, to fall back to `releaseID` if this fails, what do you think?

rccrdpccl: I think in this case we can add `unknown` as release and further info can be found on the warning message log. This will keep cardinality low even if we are not able to extract major-minor version. WDYT?

paul-maidment: Yes, good idea

rccrdpccl: BTW I think extracting the version from the image could be costly, not sure it's worth it in this case. Is there any easy way we can have this value without this overhead?

paul-maidment: I don't think there are any guarantees about the image URI, it could be based on digest rather than tag for example, in these cases, we cannot infer the version from the URI.

My assumption is that there is no other reliable way to obtain this information, otherwise, why do we even need this function in the release? I can't see any solid alternatives in the code?

paul-maidment: > not sure it's worth it in this case

Another option is to remove the release label entirely.

Given that we are moving in a direction of only reporting timeout (based on the other comment we are still discussing) then it may be acceptable to simply say "there are timeouts - check it out" rather than labelling a version.

Further info could be gathered from logs anyway.

WDYT?

internal/metrics/metricsManager.go
paul-maidment: > > Why is this not actionable?
> 
> If we trigger an alert, what action should we take? I can only think of one: fix the code and redeploy. This would be the case with a lot of other errors throughout the code base. What guarantees a metric+alert for this specific case?
> 
> Regarding errors IMO we should strive to have generic alerts and easy traceability (with logs, traces, metrics) to the root cause. This seems very specific, if we had to do this for every potential cause of error, we'd have tons of error-related metrics.
> 
> What other side-effects having timeouts error would the app experience?

So you are suggesting we don't monitor and alert on this from the feature?

What exactly do you mean by "generic" in this case?

internal/metrics/metricsManager.go
rccrdpccl: That's a good point, we probably should use this value. Could this value be somehow recycled? Are we using it already up the chain? Are we saving it in the DB? Do we have it accessible in this scope? Can it be made accessible easily?

paul-maidment: >If we have only the full OpenshiftVersion value, we could extract major-minor from that with semver package

I think this may be the way to go, but beware that Openshift version is not technically the release version.

So we might be able to say "for this openshift version I had a problem with a release", that's about as truthful as we can be.

internal/metrics/metricsManager.go
rccrdpccl: > I am not clear on what is being suggested here, to not track these errors?
Is this not a site reliability issue at that point?

What I am suggesting is to make sure we have a path of action for each metric. In this case, SRE could not act on this metric, but it doesn't necessarily mean we do not want to track it.

I do not see why this error would need to be treated in a special way, but I might be wrong - that's what I am trying to find out.

> The inability to extract any release due to Mutex lockup we can assume.

What other symptoms would this cause? Deadlock in the main thread and app unresponsive?

> I think the alerting policy, we can defer until later as this can probably be done (or not done if you prefer) in Prometheus right? I don't think it's incorrect to make this possible by making sure the data is there to alert on.

The alert policy can be defined in a later stage, however we want to make sure we expose metrics that will be useful. I think exposing "just in case" it's not a good practice and will eventually just result in an overloaded prometheus.

paul-maidment: > What other symptoms would this cause? Deadlock in the main thread and app unresponsive?

As mentioned above 

"The inability to extract any release due to Mutex lockup we can assume. A severe enough condition to require immediate intervention."

More specifically this would result in near immediate retries for any requests that needed to download a release as these would be waiting for storage to be free.

Releases that have already been cached would be allowed to proceed.

But the service degradation would be quite nasty. 

> The alert policy can be defined in a later stage, however we want to make sure we expose metrics that will be useful. I think exposing "just in case" it's not a good practice and will eventually just result in an overloaded prometheus.

This is not "just in case" - we are fairly certain we want to alert when this metric goes 'non zero' in a timeframe. 
This metric can be used to infer with a high degree of reliability that there is an error with the Mutex.

My point here was less about registering something "just in case" and more a response to your comment about this metric being "for a specific error". I am making the point that this is not the case. 

We are recording this metric to help us detect a specific problem, sure... But we do not decide here whether this is a problem or not (Spoiler -- it is and we would definitely write an alert in this case). That will be done later.

How else would you suggest we proceed here....just ignore it?

internal/metrics/metricsManager.go
rccrdpccl: As discussed I think you're right here and we could use OC based extraction if we don't find the value up the chain already

internal/metrics/metricsManager.go
rccrdpccl: As discussed offline we can get a compromise solution and have a metric to expose error(s)

internal/metrics/metricsManager.go
paul-maidment: Nothing found further up the chain, so will mark this one as resolved.

internal/installercache/installercache.go
```diff
@@ -334,9 +351,9 @@ func (i *Installers) pruneExpiredHardLinks(links []*fileInfo, gracePeriod time.D
 		grace := graceTime.Unix()
 		if finfo.info.ModTime().Unix() < grace {
 			i.log.Infof("attempting to prune hard link %s", finfo.path)
-			err := os.Remove(finfo.path)
-			if err != nil {
+			if err := os.Remove(finfo.path); err != nil {
 				i.log.WithError(err).Errorf("failed to prune hard link %s", finfo.path)
+				continue
```
rccrdpccl, line 356: What is the purpose of this `continue`?

internal/installercache/installercache.go
```diff
@@ -138,14 +140,25 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
...
 		case <-ctx.Done():
-			return nil, ctx.Err()
+			err := ctx.Err()
+			if err == context.DeadlineExceeded {
+				i.metricsAPI.InstallerCacheGetReleaseTimeout(releaseID)
+			}
+			return nil, err
 		default:
-			release, err := i.get(releaseID, releaseIDMirror, pullSecret, ocRelease, ocpVersion, clusterID)
+			majorMinorVersion, err := ocRelease.GetMajorMinorVersion(i.log, releaseID, releaseIDMirror, pullSecret)
```
rccrdpccl, line 149: I think this should be done outside the for select

internal/installercache/installercache.go
```diff
@@ -138,14 +140,25 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
...
+			if err != nil {
+				i.log.Warnf("unable to get majorMinorVersion to record metric for %s falling back to full URI", releaseID)
+				majorMinorVersion = releaseID
```
rccrdpccl, line 152: as discussed releaseID should not be a value. We can set `unknown` or `undfined` or something if we cannot extract the value from the release (which should never happen anyway)

internal/installercache/installercache.go
```diff
@@ -138,14 +140,25 @@ func (i *Installers) Get(ctx context.Context, releaseID, releaseIDMirror, pullSe
...
+			}
+			release, err := i.get(releaseID, releaseIDMirror, pullSecret, ocRelease, ocpVersion, clusterID, majorMinorVersion)
 			if err == nil {
+				i.metricsAPI.InstallerCacheGetReleaseOK(majorMinorVersion)
```
rccrdpccl, line 156: We agreed on having 2 metrics:
* `assisted_installer_cache_get_release{hit="(true|false)"}`
*  `assisted_installer_cache_get_release_error{error="(timeout|whatever|other/unknown/undefined)"}`

I still think the error metric won't be of much (or any) use, but definitely we do not want to count all requests there. What I thought of this is hits+misses+errors should return the total number of requests to this function. 

Did I misunderstand anything?

internal/metrics/metricsManager.go
```diff
@@ -39,6 +40,10 @@ const (
 	counterFilesystemUsagePercentage              = "assisted_installer_filesystem_usage_percentage"
 	histogramMonitoredHostsDurationMs             = "assisted_installer_monitored_hosts_duration_ms"
 	histogramMonitoredClustersDurationMs          = "assisted_installer_monitored_clusters_duration_ms"
+	counterInstallerCacheGetRelease               = "assisted_installer_cache_get_release"
+	counterInstallerCacheReleaseCached            = "assisted_installer_cache_get_release_cached"
```
rccrdpccl, line 44: I think this is unused

internal/installercache/installercache.go
```diff
@@ -310,8 +325,10 @@ func (i *Installers) evictFile(filePath string) error {
 	i.log.Infof("evicting binary file %s due to storage pressure", filePath)
 	err := os.Remove(filePath)
 	if err != nil {
+		i.metricsAPI.InstallerCacheReleaseEvicted(false)
```
rccrdpccl, line 328: Shouldn't we measure this in the outer method `evict()`? https://github.com/openshift/assisted-service/pull/7281/files#diff-cf73de10b668273452bf6aeb594f0385d3f3ea5b1a851c0f577d2d8989e6cccdR320

If we really want to add number of files evicted we should change the metric type to bucketed

internal/metrics/metricsManager.go
paul-maidment: I don't think so...

internal/installercache/installercache.go
paul-maidment: I think you did, I thought we were at a consensus that we would record both

internal/installercache/installercache.go
paul-maidment: I don't see any point to doing that as nothing apart from this case requires this.
But since you asked, I'll move it...

internal/installercache/installercache.go
paul-maidment: The intention here is just to show evictions are occurring, maybe we can make do with logs, I'll remove this metric.

internal/metrics/metricsManager.go
paul-maidment: We need the one for cached as far as I can tell...

paul-maidment: I removed `counterInstallerCacheGetRelease` 

Though `counterInstallerCacheReleaseCached` is still used.

rccrdpccl:
@paul-maidment  I've updated the specs on the Jira issue, they are now much more detailed. Please refer to those.

internal/metrics/metricsManager.go
```diff
@@ -77,6 +83,12 @@ const (
 	imageLabel                 = "imageName"
 	hosts                      = "hosts"
 	clusters                   = "clusters"
+	labelStatus                = "status"
+	labelValueOK               = "ok"
+	labelValueError            = "error"
```
rccrdpccl, line 88: unused

internal/metrics/metricsManager.go
```diff
@@ -39,6 +40,9 @@ const (
 	counterFilesystemUsagePercentage              = "assisted_installer_filesystem_usage_percentage"
 	histogramMonitoredHostsDurationMs             = "assisted_installer_monitored_hosts_duration_ms"
 	histogramMonitoredClustersDurationMs          = "assisted_installer_monitored_clusters_duration_ms"
+	counterInstallerCacheReleaseCached            = "assisted_installer_cache_get_release_cached"
```
rccrdpccl, line 43: unused

internal/installercache/installercache.go
```diff
@@ -161,14 +168,16 @@ func (i *Installers) getDiskUsageIncludingHardlinks() (uint64, error) {
...
+func (i *Installers) extractReleaseIfNeeded(path, releaseID, releaseIDMirror, pullSecret, ocpVersion string, ocRelease oc.Release, majorMinorVersion string) (extractDuration float64, cached bool, err error) {
 	_, err = os.Stat(path)
 	if err == nil {
+		i.metricsAPI.InstallerCacheGetReleaseCached(majorMinorVersion, true)
 		return 0, true, nil // release was found in the cache
 	}
 	if !os.IsNotExist(err) {
 		return 0, false, err
 	}
+	i.metricsAPI.InstallerCacheGetReleaseCached(majorMinorVersion, false)
```
rccrdpccl, line 180: IIUC this could be done with one call in the main Get function:
```
			release, err := i.get(releaseID, releaseIDMirror, pullSecret, ocRelease, ocpVersion, clusterID, majorMinorVersion)
			if err == nil {
				return release, nil
			}
                        i.metricsAPI.InstallerCacheGetReleaseCached(majorMinorVersion, release.cached)
                        _, isCapacityError := err.(*errorInsufficientCacheCapacity)
			if !isCapacityError {
				return nil, errors.Wrapf(err, "failed to get installer path for release %s", releaseID)
			}
time.Sleep(i.config.ReleaseFetchRetryInterval)
```

This will avoid passing down the chain major/minor version and have one single point for this metric

PR #7303, opened 2025-02-11
ACM-17767: Don't remove spoke resources when cluster is being deleted

carbonin:
The intention in the agent controller was to only remove the spoke resources when the entire cluster was not being removed. This was previously done my check the cluster deployment, but the agent cluster install is what actually triggers the agent removal. This means that we really should be checking both resources to avoid a race when everything is removed at the same time.

Before this change it was possible for a node or two to be removed before the deletion timestamp was set on the cluster deployment. Now this will no longer happen as one of the resources will need to be marked as deleted before any agent gets removed.

Resolves https://issues.redhat.com/browse/ACM-17767

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

Manually tested by deploying a 6-node cluster. Before the change one of the nodes would be removed when deleting all the resources at once from the hub, after the change all the nodes remain in the spoke after removing everything from the hub.

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/controller/controllers/agent_controller.go
+++ b/internal/controller/controllers/agent_controller.go
@@ -32,6 +32,7 @@ import (
 	bmh_v1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	. "github.com/openshift/assisted-service/api/common"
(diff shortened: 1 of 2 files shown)


CrystalChun (approved):
/lgtm

PR #7307, opened 2025-02-13
MGMT-19887: Add Histogram Metric for Host Monitoring Time

rivkyrizel:
This PR updates the monitoring metrics, which were previously reported as gauge, to use histograms.

Instead of simply counting the number of monitored hosts and clusters, the new histograms will track the duration of each monitoring operation, allowing us to retain valuable timing data and improve performance insights.

Previously, using gauge led to misleading metrics—since the host monitoring loop runs every few seconds but Prometheus scrapes at different intervals, a significant number of monitored hosts could be lost, leading to inaccurate analysis. The same issue applied to monitored clusters.

By switching to histograms, we can now accurately measure how long each monitoring process takes and identify potential performance bottlenecks.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/metrics/metricsManager.go
+++ b/internal/metrics/metricsManager.go
@@ -37,8 +37,8 @@ const (
 	counterClusterValidationFailed                = "assisted_installer_cluster_validation_is_in_failed_status_on_cluster_deletion"
 	counterClusterValidationChanged               = "assisted_installer_cluster_validation_failed_after_success_before_installation"
 	counterFilesystemUsagePercentage              = "assisted_installer_filesystem_usage_percentage"
-	counterMonitoredHosts                         = "assisted_installer_monitored_hosts"
-	counterMonitoredClusters                      = "assisted_installer_monitored_clusters"
+	histogramMonitoredHostsDurationMs             = "assisted_installer_monitored_hosts_duration_ms"
+	histogramMonitoredClustersDurationMs          = "assisted_installer_monitored_clusters_duration_ms"
 )
 
 const (
@@ -59,8 +59,8 @@ const (
 	counterDescriptionClusterValidationFailed                = "Number of cluster validation errors"
 	counterDescriptionClusterValidationChanged               = "Number of cluster validations that already succeed but start to fail again"
 	counterDescriptionFilesystemUsagePercentage              = "The percentage of the filesystem usage by the service"
-	counterDescriptionMonitoredHosts                         = "Number of hosts monitored by host monitor"
-	counterDescriptionMonitoredClusters                      = "Number of clusters monitored by cluster monitor"
+	histogramDescriptionMonitoredHostsDurationMs             = "Histogram/sum/count of monitored hosts duration (ms)"
+	histogramDescriptionMonitoredClustersDurationMs          = "Histogram/sum/count of monitored clusters duration (ms)"
 )
 
 const (
@@ -92,8 +92,8 @@ type API interface {
 	DiskSyncDuration(syncDuration int64)
 	ImagePullStatus(imageName, resultStatus string, downloadRate float64)
 	FileSystemUsage(usageInPercentage float64)
-	MonitoredHostsCount(monitoredHosts int64)
-	MonitoredClusterCount(monitoredClusters int64)
+	MonitoredHostsDurationMs(monitoredHostsMillis float64)
+	MonitoredClustersDurationMs(monitoredClustersMillis float64)
 }
 
 type MetricsManager struct {
@@ -117,8 +117,8 @@ type MetricsManager struct {
 	serviceLogicClusterValidationFailed                *prometheus.CounterVec
 	serviceLogicClusterValidationChanged               *prometheus.CounterVec
 	serviceLogicFilesystemUsagePercentage              *prometheus.GaugeVec
-	serviceLogicMonitoredHosts                         *prometheus.GaugeVec
-	serviceLogicMonitoredClusters                      *prometheus.GaugeVec
+	serviceLogicMonitoredHostsDurationMs               *prometheus.HistogramVec
+	serviceLogicMonitoredClustersDurationMs            *prometheus.HistogramVec
 	collectors                                         []prometheus.Collector
 }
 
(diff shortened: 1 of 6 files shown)


internal/metrics/metricsManager.go
```diff
@@ -37,8 +37,8 @@ const (
 	counterClusterValidationFailed                = "assisted_installer_cluster_validation_is_in_failed_status_on_cluster_deletion"
 	counterClusterValidationChanged               = "assisted_installer_cluster_validation_failed_after_success_before_installation"
 	counterFilesystemUsagePercentage              = "assisted_installer_filesystem_usage_percentage"
-	counterMonitoredHosts                         = "assisted_installer_monitored_hosts"
-	counterMonitoredClusters                      = "assisted_installer_monitored_clusters"
+	counterMonitoredHostsMillis                   = "assisted_installer_monitored_hosts_milliseconds"
```
rccrdpccl, line 40: can we please check how would the final metric name look like? It's available at the /metrics endpoint

rivkyrizel: The final metric names in `/metrics` appear as `service_assisted_installer_monitored_hosts_milliseconds`.

rccrdpccl: IIRC wouldn't it be something like `service_assisted_installer_monitored_hosts_milliseconds_buckets`, `service_assisted_installer_monitored_hosts_milliseconds_count` and `service_assisted_installer_monitored_hosts_milliseconds_sum`?

internal/metrics/metricsManager.go
```diff
@@ -275,19 +275,21 @@ func NewMetricsManager(registry prometheus.Registerer, eventsHandler eventsapi.H
...
-		serviceLogicMonitoredClusters: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		serviceLogicMonitoredClustersMillis: prometheus.NewHistogramVec(prometheus.HistogramOpts{
 			Namespace: namespace,
 			Subsystem: subsystem,
-			Name:      counterMonitoredClusters,
-			Help:      counterDescriptionMonitoredClusters,
-		}, []string{hosts}),
+			Name:      counterMonitoredClustersMillis,
+			Help:      counterDescriptionMonitoredClustersMillis,
+			Buckets: []float64{0.1, 0.5, 1, 2, 5, 10, 20, 50, 100},
```
rccrdpccl, line 291: is this milliseconds? Assuming it is, do we care if a cluster is monitored in 0.1 or 0.5 ms? I think the low levels have too many buckets, but the high buckets have not enough: we wouldn't know if it took 100ms or 10s.

rivkyrizel: I wasn’t sure what the expected times should be. In my local tests with a few servers, most monitoring times were under a millisecond, so I set up the buckets this way. If we expect higher values in production, I can change them.

internal/metrics/metricsManager.go
```diff
@@ -59,8 +59,8 @@ const (
 	counterDescriptionClusterValidationFailed                = "Number of cluster validation errors"
 	counterDescriptionClusterValidationChanged               = "Number of cluster validations that already succeed but start to fail again"
 	counterDescriptionFilesystemUsagePercentage              = "The percentage of the filesystem usage by the service"
-	counterDescriptionMonitoredHosts                         = "Number of hosts monitored by host monitor"
-	counterDescriptionMonitoredClusters                      = "Number of clusters monitored by cluster monitor"
+	counterDescriptionMonitoredHostsMillis                   = "Histogram/sum/count of monitored time for hosts (ms)"
```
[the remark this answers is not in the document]
rivkyrizel: I noticed that all metrics are defined this way. Should I still change it?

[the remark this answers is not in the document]
rccrdpccl: yes please, let's change at least the ones we are modifying to have a more meaningful name. Feel free to change other metrics too in order to make them more meaningful, although that's optional as it's outside the scope of this PR.

internal/metrics/metricsManager.go
rccrdpccl: if all goes well I'd expect it to be really fast, but I wouldn't really care if it's done in 0.1 , 0.5, 1, 2 or 5ms

Basically we want to understand:
* are messages being processed as fast as it could be?
* if something is slow, are all clusters going slow? Is anything particularly slow?

if we monitor 2000 clusters, at 1ms per cluster it would take 2s, at 10ms 20s, and so on. As we have a total window of 5 minutes (configurable), we expect to finish all clusters within those 5 minutes, we definitely want to know if we potentially go above that.

I'd go for something like 1, 10, 100, 200, 500, 1000, 10000, 60000, 120000, 180000, 240000, 300000 - but this is just a guess. @rivkyrizel WDYT?

@adriengentil what buckets do you think would be useful?

adriengentil: I think we don't need more buckets after 1s, if it fills up, then it means we have an issue. And below 100ms,  I think I don't care much if it takes 1ms or 10ms. I would stick with 0, 100, 200, 500, 1000. WDYT?

internal/metrics/metricsManager.go
```diff
@@ -475,12 +477,12 @@ func (m *MetricsManager) FileSystemUsage(usageInPercentage float64) {
 	m.serviceLogicFilesystemUsagePercentage.WithLabelValues().Set(usageInPercentage)
 }
 
-func (m *MetricsManager) MonitoredHostsCount(monitoredHosts int64) {
-	m.serviceLogicMonitoredHosts.WithLabelValues(hosts).Set(float64(monitoredHosts))
+func (m *MetricsManager) MonitoredHostsDurationMs(monitoredHostsMs float64) {
+	m.serviceLogicMonitoredHostsDurationMs.WithLabelValues(hosts).Observe(monitoredHostsMs)
```
rccrdpccl, line 481: This will mark label hosts with value hosts.
```
assisted_installer_monitored_hosts_duration_ms{hosts="hosts"}
```
I think we can safely drop it

internal/metrics/metricsManager.go
```diff
@@ -475,12 +477,12 @@ func (m *MetricsManager) FileSystemUsage(usageInPercentage float64) {
...
 }
 
-func (m *MetricsManager) MonitoredClusterCount(monitoredClusters int64) {
-	m.serviceLogicMonitoredClusters.WithLabelValues(clusters).Set(float64(monitoredClusters))
+func (m *MetricsManager) MonitoredClustersDurationMs(monitoredClustersMs float64) {
+	m.serviceLogicMonitoredClustersDurationMs.WithLabelValues(clusters).Observe(monitoredClustersMs)
```
rccrdpccl, line 485: same as hosts, we can drop the label

rccrdpccl (commented):
Looks good, let's just remove the labels and we can merge it!

internal/metrics/metricsManager.go
```diff
@@ -475,12 +477,12 @@ func (m *MetricsManager) FileSystemUsage(usageInPercentage float64) {
 	m.serviceLogicFilesystemUsagePercentage.WithLabelValues().Set(usageInPercentage)
 }
 
-func (m *MetricsManager) MonitoredHostsCount(monitoredHosts int64) {
-	m.serviceLogicMonitoredHosts.WithLabelValues(hosts).Set(float64(monitoredHosts))
+func (m *MetricsManager) MonitoredHostsDurationMs(monitoredHostsMs float64) {
+	m.serviceLogicMonitoredHostsDurationMs.WithLabelValues().Observe(monitoredHostsMs)
```
rccrdpccl, line 481: `WithLabelValues` seems to follow a fluent interface pattern. I think it can be dropped altogether if called with no arguments.

PR #7310, opened 2025-02-13
MGMT-19928: Add dev doc for disk image install

carbonin:
Resolves https://issues.redhat.com/browse/MGMT-19928

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/dev/disk-image-install.md
+++ b/docs/dev/disk-image-install.md
@@ -1,0 +1,22 @@
+# Assisted Installation With a Disk Image
+
+You can use the Assisted Installer to install directly to the existing root filesystem of a disk image.
+This should be done for platforms that cannot take the discovery ISO as boot media, and also cannot use iPXE.
+
+## Available Disk Images
+
+You can download disk images for use with this process at https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/
+An example for installing OCP 4.17 with libvirt would be https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.17/4.17.2/rhcos-4.17.2-x86_64-qemu.x86_64.qcow2.gz
+
+## Discovery and Installation
+
+You must write the Assisted Installer discovery ignition to the disk on first boot in order to install in this way.
+You can do this using whatever metadata service is available for your [target platform](https://coreos.github.io/ignition/supported-platforms/#supported-platforms).
+
+You can download the discovery ignition using the assisted-service REST API.
+A presigned URL is published on the InfraEnv status (`.status.bootArtifacts.discoveryIgnitionURL`) if using the kubernetes APIs.
+Otherwise you can download it using the endpoint `/v2/infra-envs/{infra_env_id}/downloads/files` with the query parameter `file_name=discovery.ign`
+
+Once the Assisted Installer agent is running, discovery will proceed as normal outside of some disabled validations (specifically the disk speed check which writes directly to the target install device).
+When the host is installed the OS content from the selected release image will be installed to a new [ostree stateroot](https://ostreedev.github.io/ostree/deployment/) and the install ignition will be applied to it when it boots for the first time.
+During this first boot the initial stateroot and boot files are also cleaned up by the `cleanup-assisted-discovery-stateroot` service.


docs/dev/disk-image-install.md
```diff
@@ -0,0 +1,22 @@
...
+## Discovery and Installation
+
+You must write the Assisted Installer discovery ignition to the disk on first boot in order to install in this way.
+You can do this using whatever metadata service is available for your target platform.
+
+You can download the discovery ignition assisted-service REST API.
+A presigned URL is published on the InfraEnv status (`.status.bootArtifacts.discoveryIgnitionURL`) if using the kubernetes APIs.
+Otherwise you can download it using the endpoint `/v2/infra-envs/{infra_env_id}/downloads/files` with the query parameter `file_name=discovery.ign`
+
+Once the Assisted Installer agent is running, discovery will proceed as normal outside of some disabled validations (specifically the disk speed check which writes directly to the target install device).
```
CrystalChun, line 20: Sorry, I'm not familiar with this method. Do you need to run something after the ignition is written to disk?

docs/dev/disk-image-install.md
```diff
@@ -0,0 +1,22 @@
+# Assisted Installation With a Disk Image
+
+You can use the Assisted Installer to install to install directory to the existing root filesystem of a disk image.
```
CrystalChun, line 3:
```suggestion
You can use the Assisted Installer to install a directory to the existing root filesystem of a disk image.
```
I like the emphasis though 😆

rccrdpccl:
> @rccrdpccl @CrystalChun do you think we need anything more here? Can you take a look?

I think this gives a good overview, I don't believe any more detail is needed.

docs/dev/disk-image-install.md
carbonin: Run what? The validation? We don't run it. That's what I'm calling out here.

carbonin: On first boot of the host the discovery ignition is written to disk. This contains the service to run the agent.

When you start a host in this way the disk it is created with already contains rhcos and it will look to pull ignition over some metadata service that is specific to the platform. So this all happens before we "install" anything (just like with the ISO).

CrystalChun: Oh! Sorry, I didn't know it started like the live iso (on first boot [duh 🤦‍♀️] ). Thanks for the clarification!

docs/dev/disk-image-install.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Available Disk Images
+
+You can download disk images for use with this process at https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos
+An example for installing OCP 4.17 with libvirt would be https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.17/4.17.2/rhcos-4.17.2-x86_64-qemu.x86_64.qcow2.gz
+
+## Discovery and Installation
+
+You must write the Assisted Installer discovery ignition to the disk on first boot in order to install in this way.
+You can do this using whatever metadata service is available for your target platform.
```
CrystalChun, line 14: Because I'm not familiar with this 😅 but it would be helpful to have an example of a service or how to do this?

CrystalChun: But if it's common knowledge that doesn't need to be documented, then feel free to ignore my comment

carbonin: I don't know.

I was thinking of adding an example of how I was doing this for testing, but I didn't feel like that would really be helpful for anyone who wanted to do this on another platform.

I don't think it makes sense to go into everything you and Riccardo are doing now for CAPI around this either here so I left it pretty vague. I suppose I could link https://coreos.github.io/ignition/supported-platforms/#supported-platforms which talks a bit about how ignition is pulled for all those platforms?

carbonin: For example I was doing this:

>[QEMU](https://www.qemu.org/) (qemu) - Ignition will read its configuration from the ‘opt/com.coreos/config’ key on the QEMU Firmware Configuration Device (available in QEMU 2.4.0 and higher).

But I don't think how to do that is particularly helpful here.

CrystalChun: Ah I can see why you didn't document a concrete example. The link would definitely be helpful for anyone who wants to try this out on one of the platforms!

docs/dev/disk-image-install.md
```diff
@@ -0,0 +1,22 @@
+# Assisted Installation With a Disk Image
+
+You can use the Assisted Installer to install directly to the existing root filesystem of a disk image.
+This should be done for platforms that cannot take the discovery ISO as boot media, and also cannot use iPXE.
+
+## Available Disk Images
+
+You can download disk images for use with this process at https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos
+An example for installing OCP 4.17 with libvirt would be https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.17/4.17.2/rhcos-4.17.2-x86_64-qemu.x86_64.qcow2.gz
```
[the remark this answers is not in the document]
carbonin: I think it's out of scope.

Right now it's something @rccrdpccl is doing through the metal CAPI integration so there's many layers of abstraction we'd need to explain to reference how it is actually being used.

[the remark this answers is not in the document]
carbonin: If someone works out how to do a deploy this way for, say, Oracle or AWS, then that might warrant its own guide, but I think this is mostly just trying to explain the approach.

rccrdpccl:
/retest

flake

PR #7321, opened 2025-02-17
OCPEDGE-1311: add two node openshift deployment

qJkee:
Add new deployment type to the assisted service called two node with arbiter(TNA). Install config changes are made according to https://github.com/openshift/enhancements/pull/1674 proposal

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/installcfg/builder/builder.go
+++ b/internal/installcfg/builder/builder.go
@@ -112,6 +112,18 @@ func (i *installConfigBuilder) getBasicInstallConfig(cluster *common.Cluster) (*
 		SSHKey:     cluster.SSHPublicKey,
 	}
 
+	if cfg.ControlPlane.Replicas == 2 {
+		cfg.Arbiter = struct {
+			Hyperthreading string `json:"hyperthreading,omitempty"`
+			Name           string `json:"name"`
+			Replicas       int    `json:"replicas"`
+		}{
+			Hyperthreading: i.getHypethreadingConfiguration(cluster, "master"),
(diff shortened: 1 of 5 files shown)


carbonin:
Are we additionally expected to deploy the arbiter node? How is that being handled?

internal/installcfg/builder/builder.go
```diff
@@ -112,6 +112,18 @@ func (i *installConfigBuilder) getBasicInstallConfig(cluster *common.Cluster) (*
 		SSHKey:     cluster.SSHPublicKey,
 	}
 
+	if cfg.ControlPlane.Replicas == 2 {
```
carbonin, line 115: How will we differentiate between the arbiter case and the two node with fencing case once that comes through?

carbonin (changes requested):
After reading more of the enhancement it does seem like we'll need significantly more changes to support this.

We'll need to update host minimum requirements for the arbiter role, we'll need to support reading the new ignition from the installer and ensuring the host designated as the arbiter downloads it, and we'll probably also need some way to ensure this is gated to only openshift versions that support TNA

gamli75:
Closing this because we will handle it as part of https://issues.redhat.com/browse/MGMT-19980

PR #7325, opened 2025-02-17
MGMT-19939: Allow using the rootfs install method to add nodes

carbonin:
Previously the coreos image would only be added to the install command for clusters that were initally being installed, this was a mistake, this install method should also be available when adding additional nodes to a cluster

Resovles https://issues.redhat.com/browse/MGMT-19939

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

Manually tested by deploying an SNO and adding a new node using the "install to existing root" method.

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/host/hostcommands/install_cmd.go
+++ b/internal/host/hostcommands/install_cmd.go
@@ -139,29 +139,36 @@ func (i *installCmd) getFullInstallerCommand(ctx context.Context, cluster *commo
 	}
 	request.NotifyNumReboots = i.notifyNumReboots
 
-	// those flags are not used on day2 installation
-	if swag.StringValue(cluster.Kind) != models.ClusterKindAddHostsCluster {
-		releaseImage, err := i.versionsHandler.GetReleaseImage(ctx, cluster.OpenshiftVersion, cluster.CPUArchitecture, cluster.PullSecret)
+	cpuArch := cluster.CPUArchitecture
+	// cluster cpu arch is not set for imported clusters so use infraenv for this case
+	if cpuArch == "" {
+		cpuArch = infraEnv.CPUArchitecture
+	}
+
+	releaseImage, err := i.versionsHandler.GetReleaseImage(ctx, cluster.OpenshiftVersion, cpuArch, cluster.PullSecret)
+	if err != nil {
+		return "", err
+	}
+
+	if inventory != nil && inventory.Boot != nil && inventory.Boot.DeviceType == models.BootDeviceTypePersistent {
+		request.CoreosImage, err = i.ocRelease.GetCoreOSImage(i.log, *releaseImage.URL, i.instructionConfig.ReleaseImageMirror, cluster.PullSecret)
 		if err != nil {
 			return "", err
 		}
+		i.log.Infof("installing to disk with CoreOS image %s", request.CoreosImage)
+	}
 
+	// those flags are not used on day2 installation
+	if swag.StringValue(cluster.Kind) != models.ClusterKindAddHostsCluster {
 		request.McoImage, err = i.ocRelease.GetMCOImage(i.log, *releaseImage.URL, i.instructionConfig.ReleaseImageMirror, cluster.PullSecret)
 		if err != nil {
 			return "", err
 		}
 
-		if inventory != nil && inventory.Boot != nil && inventory.Boot.DeviceType == models.BootDeviceTypePersistent {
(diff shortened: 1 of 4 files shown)


carbonin:
I'll look into the subsystem failure. It seems to be around day2 install so maybe it's real.

internal/host/hostcommands/install_cmd.go
```diff
@@ -139,29 +139,30 @@ func (i *installCmd) getFullInstallerCommand(ctx context.Context, cluster *commo
 	}
 	request.NotifyNumReboots = i.notifyNumReboots
 
-	// those flags are not used on day2 installation
-	if swag.StringValue(cluster.Kind) != models.ClusterKindAddHostsCluster {
-		releaseImage, err := i.versionsHandler.GetReleaseImage(ctx, cluster.OpenshiftVersion, cluster.CPUArchitecture, cluster.PullSecret)
+	releaseImage, err := i.versionsHandler.GetReleaseImage(ctx, cluster.OpenshiftVersion, cluster.CPUArchitecture, cluster.PullSecret)
```
carbonin, line 142: So it seems like the subsystem test is failing because the cluster doesn't have a CPU arch here.

The log I found when running the test locally is this:
```
time="2025-02-19T14:56:11Z" level=warning msg="Failed to generate steps for command *hostcommands.installCmd" func="github.com/openshift/assisted-service/internal/host/hostcommands.(*InstructionManager).GetNextSteps" file="/assisted-service/internal/host/hostcommands/instruction_manager.go:180" error="validation failure list:\ncpu_architecture in body should be one of [x86_64 aarch64 arm64 ppc64le s390x multi]" go-id=13645 host_id=c8052627-52d3-4626-a5a7-902496ae4f90 infra_env_id=e742e826-30c3-450f-9f92-99b5f0a48d7c pkg=instructions request_id=b58547e8-d467-4364-918e-e502db71b569
```

I wonder if this is a problem? The test in question is creating the cluster using the [import cluster API](https://github.com/openshift/assisted-service/blob/78d44ecb4a04b452cb56fcb5b919a0bcc399a5cf/swagger.yaml#L5207-L5226) which doesn't seem to accept a CPU architecture. So should this be using the infraenv arch? I suppose that would make more sense in more cases, right?

What do you think @rccrdpccl?

carbonin: I'm not sure how this (using the infraenv arch) might affect installs on day1 that are using multi-arch release images.

I could imagine this failing if someone were to configure a single os image and a single release image (the multiarch one). We would then fail to find the release image matching the cluster version and infraenv arch.

carbonin: Thought about this some more...

We don't currently require finding a release image for installing day2 hosts at all and don't require a CPU arch for clusters that are imported. This means we can't reliably find a release image for imported clusters.

This isn't something I'm going to suggest changing in this PR, but will leave us with an edge case we can't deal with no matter what we do here. This is when a user (presumably on prem), has a cluster and infraenv whose version and arch combination do not have a corresponding release image. Normally (when using an ISO) this wouldn't be a problem for adding a node since all we do is boot using the ISO, but now it will be a problem when using the direct rootfs install.

We really should validate this in a way that the user can see. Unfortunately returning an error here is only logged (as we found out during the subsystem failure). I also don't want to change this behavior (ignoring errors from `GetSteps`) as it seems to have been done very intentionally. The only options then are to either validate access to a release image from the actual install request or just log here and fail to install mostly silently in this rather slim, but valid, edge case.

This validation really should only be done if we really do need the release image, which will mean inspecting the host inventory and looking up a release image during the request to install. This seems a bit heavy for this kind of check, but as this is a a real situation someone could end up in and it seems unlikely that we're going to solve this (by requiring cpu arch or release image selection in some other way when importing a cluster) I'm inclined to go this direction.

Will take a look at adding this tomorrow.

carbonin:
Moved to draft to stop retests as we know the issue is due to the changes here.

rccrdpccl:
@carbonin it seems the test is now triggering the error condition no release image found for host

carbonin:
/unhold

Subsystem should pass now

gamli75 (approved):

PR #7348, opened 2025-02-25
MGMT-20022: Handle missing ManagedCluster in non MCE environments

paul-maidment:
When running the infrastructure operator in a non MCE environment it is possible to encounter an error due to the missing CRD. This error is cryptic and hard to understand.

This PR clears that up by raising an error if the ManagedCluster CRD is not in the cluster.

Additionally, this has been added to deploy/operator so that this CRD is installed whenever that operator is installed using this script.

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist


cmd/operator/main.go
```diff
@@ -86,6 +88,17 @@ func init() {
 	utilruntime.Must(certtypes.AddToScheme(scheme))
 }
 
+func doesManagedClusterCRDExist(mgr manager.Manager) error {
```
danmanor, line 91: Lets distinguish between an actual error fetching the CRD to a `false` answer on whether it exists

cmd/operator/main.go
```diff
@@ -114,6 +127,13 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	log := logrus.New()
+	failOnError := func(err error, msg string, args ...interface{}) {
```
danmanor, line 131: We can probably avoid using function for only one use case and write the code directly

cmd/operator/main.go
paul-maidment: If you look at the implementation, you can see that in the error cases, this is where we would return 'false'.
So any error is the same as saying 'false', no need to change this.

```
func doesManagedClusterCRDExist(mgr manager.Manager) (bool, error) {
	gvk, err := apiutil.GVKForObject(&clusterv1.ManagedCluster{}, mgr.GetScheme())
	if err != nil {
		return false, err
	}
	if _, err = mgr.GetRESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version); err != nil {
		return false, err
	}
	return true, nil
}
```

danmanor:
/override ci/prow/edge-e2e-ai-operator-ztp-3masters ci/prow/edge-e2e-ai-operator-ztp-disconnected

hack/crds/hive/metal3.io_preprovisioningimages.yaml
carbonin, line 1: This is not a hive crd. This is specific to metal3/BMO

hack/crds/hive/metal3.io_baremetalhosts.yaml
carbonin, line 1: Same here, this is not hive

deploy/operator/setup_assisted_operator.sh
```diff
@@ -101,6 +101,9 @@ EOF
 }
 
 function install_from_catalog_source() {
+  # ensure that the necessary MCE CRDs are installed.
+  tee oc apply -f ${__root}/hack/crds/mce
```
carbonin, line 105: Why `tee`? Just for logs?

cmd/operator/main.go
```diff
@@ -114,6 +127,11 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	log := logrus.New()
+	err = validateManagedClusterCRDExist(mgr)
+	if err != nil {
```
carbonin, line 132: Nit, but this can be one line.

cmd/operator/main.go
```diff
@@ -114,6 +127,11 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	log := logrus.New()
```
carbonin, line 130: Why do we use `setupLog` at all here?
I'd prefer we either change the `setupLog` uses below this line to `log` lines or just leave this where it was and use `setupLog` for this like the surrounding calls.

cmd/operator/main.go
```diff
@@ -114,6 +127,10 @@ func main() {
 		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
+	if validateManagedClusterCRDExists(mgr) != nil {
```
carbonin, line 130: @rccrdpccl FYI this will be another thing you'll need to install for sylva deployments when not using MCE.

carbonin (approved):

gamli75:
/override ci/prow/edge-e2e-ai-operator-ztp-capi ci/prow/edge-e2e-ai-operator-disconnected-capi

danmanor:
@paul-maidment Can you please create a manual cherry-pick for `release-ocm-2.11` as well ? Some manual conflict resolution is needed there

PR #7352, opened 2025-02-26
MGMT-19929: Use sonic instead of standard json library

adriengentil:
Last investigations have shown that assisted spend a lot of time de-serializing json, and it can struggle at processing large inventories.

[sonic](https://github.com/bytedance/sonic) is a performant json library that can help the assisted at coping with these scalability issues. The advantage of this library is that is it an [almost](https://github.com/bytedance/sonic?tab=readme-ov-file#compatibility) drop-in library for the standard json library: to keep the maximum compatibility, I ensured that we used `ConfigStd` on marshaling operations because it makes the json strings comparable (keys will be sorted).

I did not update the `pkg/` folder to avoid public users of our library to be forced to use this library.

Memory is slightly higher (3 hosts with 3MB inventory):
image

CPU usage is divided by 2 (3 hosts with 3MB inventory):
image

p90 time to monitor a host is divided by 5 (3 hosts with 3MB inventory):
image
 
average time to monitor a host is divided by 3.5 (3 hosts with 3MB inventory):
image


adriengentil:
/cc @rccrdpccl 
let me know if you have suggestions! I would like to try on integration first.

rccrdpccl:
> let me know if you have suggestions! I would like to try on integration first.

it looks great! let's see how it performs in integration but LGTM

adriengentil:
Closing, while this change might be interesting, I think that https://issues.redhat.com/browse/MGMT-20454 whould be implemented first.

PR #7357, opened 2025-02-27
MGMT-20043: Modify MTU Validatin CMN to support multiple machine networks

linoyaslan:
With the recent addition of the external LB feature in CMN, enabling multiple machine networks, this PR adjusts the MTU validation to accommodate scenarios with multiple machine networks.

/cc @ori-amizur @eifrach @danmanor @AlonaKaplan 

- [ ] New Feature 
- [ ] Enhancement 
- [X] Bug fix
- [X] Tests
- [ ] Documentation
- [ ] CI/CD 

- [X] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/host/validator.go
```diff
@@ -346,19 +346,16 @@ func (v *validator) isMtuValidAllInterfaces(c *validationContext, connectivityRe
...
+			for _, machineNet := range c.cluster.MachineNetworks {
+				_, mNetwork, err := net.ParseCIDR(string(machineNet.Cidr))
 				if err != nil {
-					return ValidationError, err.Error()
+					return ValidationError, "Internal error - failed to parse machine network CIDR"
 				}
-				remoteIP := net.ParseIP(mtuReport.RemoteIPAddress)
-				if mNetwork.Contains(remoteIP) && sourceIP != "" {
+				// We are only interested in the remote IP within machine networks. If it is not part of any machine network, we don't need to consider it.
+				if mNetwork.Contains(remoteIP) {
```
AlonaKaplan, line 358: I think we should also verify the source IP belongs to any of the machine networks.

internal/host/validator.go
```diff
@@ -346,19 +346,24 @@ func (v *validator) isMtuValidAllInterfaces(c *validationContext, connectivityRe
...
-		if err != nil {
-			return ValidationError, "Internal error - failed to parse machine network CIDR"
-		}
-		for _, r := range connectivityReport.RemoteHosts {
-			for _, mtuReport := range r.MtuReport {
-				sourceIP, err := network.FindSourceIPInMachineNetwork(mtuReport.OutgoingNic, mNetwork, c.inventory.Interfaces)
+	for _, r := range connectivityReport.RemoteHosts {
+		for _, mtuReport := range r.MtuReport {
+			// We are only interested in the source IP within machine networks. If it is not part of any machine network, we don't need to consider it.
+			isSourceIPOk, err := network.IsSourceIPBelongsToAnyMachineNetwork(mtuReport.OutgoingNic, c.cluster.MachineNetworks, c.inventory.Interfaces)
```
danmanor, line 352: Maybe `isSourceIPOk` -> `isSourceIPBelongToAnyMachineNetwork` ? or any name describing the result value of this variable rather than `OK` ?

internal/host/validator.go
```diff
@@ -346,19 +346,24 @@ func (v *validator) isMtuValidAllInterfaces(c *validationContext, connectivityRe
...
+			if err != nil {
```
danmanor, line 353: Probably worth checking for error before the other logic no ?

internal/host/validator.go
```diff
@@ -346,19 +346,24 @@ func (v *validator) isMtuValidAllInterfaces(c *validationContext, connectivityRe
...
+				return ValidationError, err.Error()
+			}
+			if !isSourceIPOk {
```
danmanor, line 356: Maybe we can move this few lines above for early skip and avoid the opposite `if`

internal/host/validator.go
danmanor: I am not sure I understand this check. Why do we care if the interface belongs to any machine network ? is it because we want to ignore ISCSI interfaces ?

danmanor: The MTU check is layer 3 check right ? why do we care about machine networks at all ? if in `UMN` platform we check each interface against all other interfaces, I don't see why shouldn't we do it here as well

internal/host/validator.go
```diff
@@ -346,19 +346,24 @@ func (v *validator) isMtuValidAllInterfaces(c *validationContext, connectivityRe
 
 // isMtuValidInMachineNetwork - intended for CMN, focusing exclusively on the MTU reports for the machine network.
 func (v *validator) isMtuValidInMachineNetwork(c *validationContext, connectivityReport *models.ConnectivityReport) (ValidationStatus, string) {
```
danmanor, line 348: General note about this validation, I think we can improve its performance - I M worried about the time it will take to perform it on large clusters

internal/host/validations_test.go
```diff
@@ -2737,27 +2749,40 @@ var _ = Describe("Validations test", func() {
...
 				{MtuSuccessful: true, RemoteIPAddress: "1.2.4.5", OutgoingNic: "eth0"},
-			}, ValidationSuccess, true, false),
+			}, mNet, ValidationSuccess, true, false),
 			Entry("isMtuValidAllInterfaces (UMN) - Bad flow - all MTU reports are unsuccessful", []*models.MtuReport{
 				{MtuSuccessful: false, RemoteIPAddress: "1.2.3.5", OutgoingNic: "eth0"},
 				{MtuSuccessful: false, RemoteIPAddress: "1.2.4.5", OutgoingNic: "eth0"},
-			}, ValidationFailure, true, false),
-			Entry("Day 2 cluster - the validation should be skipped", []*models.MtuReport{}, ValidationSuccess, false, true),
+			}, mNet, ValidationFailure, true, false),
+			Entry("isMtuValidInMachineNetwork (CMN) - Happy flow - all reports where both sourceIP and remoteIP are within any of the machine networks are successful", []*models.MtuReport{
```
AlonaKaplan, line 2775: Please also test dual stack happy and failing flow

internal/host/validator.go
linoyaslan: > I am not sure I understand this check. Why do we care if the interface belongs to any machine network ?

Because in CMN, we only care about communication between interfaces within the machine networks. If the ping is sent from an interface outside the machine network, we don't want to consider it.

> why do we care about machine networks at all ?

In CMN, the machine networks are known in advance, so we only check interfaces within those networks. In UMN, since the machine networks are not known in advance, we check all interfaces.

internal/host/validator.go
linoyaslan: that's what I do. checking the error before if !isSourceIPOk

internal/host/validator.go
linoyaslan: I'm not sure how. I need the MTU report data, so I placed it right after the for loop begins.

internal/host/validator.go
danmanor: btw - does transitivity holds ? I mean if host1 MTU with host2 is valid and host2 MTO with host3 is valid, does it mean that host1 with host3 is valid as well ?

danmanor: I think we should think about ways to optimize this, and limit the runtime in case it takes too much time

internal/host/validator.go
linoyaslan: > , don't we care that the MTU between host1 and host3 is valid ?

Yes, we care, and that's exactly what this fix come to solve

internal/host/validator.go
linoyaslan: > btw - does transitivity holds ? I mean if host1 MTU with host2 is valid and host2 MTO with host3 is valid, does it mean that host1 with host3 is valid as well ?

should be

internal/host/validations_test.go
linoyaslan: @AlonaKaplan Just to clarify my understanding—why is this important here? I agree that checking IPv6 on the agent side is necessary (which I already did), but in this case, we aren't concerned with whether it's IPv6 or IPv4. What matters here is whether the report successful or not.

internal/host/validator.go
pastequo: jumping in the conversation, I don't see any super costly operation here. We can add a benchmark test to be sure about the situation.

(My point is: let's not pre-optimize when it's not necessary)

AlonaKaplan: > btw - does transitivity holds ? I mean if host1 MTU with host2 is valid and host2 MTO with host3 is valid, does it mean that host1 with host3 is valid as well ?

No, host1 doesn't communicate with host 3 via host 2. You may have host1<->switch1<->host2 host2<->switch2<->host3 and host1<->switch3<->host3

AlonaKaplan: We can try to improve the performance in a later iteration. @adriengentil based on you latest performance testing and profiling, do you think there may be a performance issue here?

internal/host/validator.go
```diff
@@ -346,19 +346,24 @@ func (v *validator) isMtuValidAllInterfaces(c *validationContext, connectivityRe
...
+			if !isSourceIPBelongsToAnyMachineNetwork {
+				continue
+			}
+			remoteIP := net.ParseIP(mtuReport.RemoteIPAddress)
+			for _, mNetwork := range parsedMachineNetworks {
+				// We are only interested in the remote IP within machine networks. If it is not part of any machine network, we don't need to consider it.
+				if mNetwork.Contains(remoteIP) {
 					if !mtuReport.MtuSuccessful {
 						return ValidationFailure, fmt.Sprintf("MTU is broken. Interface: %s, remote IP address: %s", mtuReport.OutgoingNic, mtuReport.RemoteIPAddress)
 					}
```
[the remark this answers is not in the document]
linoyaslan: or break

internal/host/validations_test.go
```diff
@@ -2737,27 +2749,56 @@ var _ = Describe("Validations test", func() {
...
-			Entry("isMtuValidAllInterfaces (UMN) - Bad flow - all MTU reports are unsuccessful", []*models.MtuReport{
-				{MtuSuccessful: false, RemoteIPAddress: "1.2.3.5", OutgoingNic: "eth0"},
+			}, mNet, ValidationSuccess, true, false),
+			Entry("isMtuValidAllInterfaces (UMN) - Bad flow - at least one unsucessfull report", []*models.MtuReport{
+				{MtuSuccessful: true, RemoteIPAddress: "1.2.3.5", OutgoingNic: "eth0"},
 				{MtuSuccessful: false, RemoteIPAddress: "1.2.4.5", OutgoingNic: "eth0"},
-			}, ValidationFailure, true, false),
-			Entry("Day 2 cluster - the validation should be skipped", []*models.MtuReport{}, ValidationSuccess, false, true),
+			}, mNet, ValidationFailure, true, false),
+			Entry("isMtuValidAllInterfaces (UMN) - Happy flow - Dual stack, all reports successful", []*models.MtuReport{
```
AlonaKaplan, line 2793: I think that the validation doesn't cover the following scenario -

eth0 has 
- ipv4 address in one of the machine network
- ipv6 address outside of the machine network

MTU check to remote IPv6 address that is inside the machine network failed.
The validation shouldn't fail, but I believe it will succeed.

internal/host/validator.go
```diff
@@ -346,19 +346,24 @@ func (v *validator) isMtuValidAllInterfaces(c *validationContext, connectivityRe
...
+			isSourceIPBelongsToAnyMachineNetwork, err := network.IsSourceIPBelongsToAnyMachineNetwork(mtuReport.OutgoingNic, parsedMachineNetworks, c.inventory.Interfaces)
+			if err != nil {
+				return ValidationError, err.Error()
+			}
+			if !isSourceIPBelongsToAnyMachineNetwork {
+				continue
+			}
+			remoteIP := net.ParseIP(mtuReport.RemoteIPAddress)
+			for _, mNetwork := range parsedMachineNetworks {
+				// We are only interested in the remote IP within machine networks. If it is not part of any machine network, we don't need to consider it.
```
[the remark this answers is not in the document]
AlonaKaplan: so during this iterations we should call `IsIPBelongToAnyMachineNetwork` per IP. We also should verify the IP is in the same IP family as the remote IP (refer to my comment on the tests).

danmanor (approved):

internal/network/utils.go
```diff
@@ -491,24 +491,51 @@ func IsLoadBalancerUserManaged(c *common.Cluster) bool {
...
+		}
+		for _, address := range addresses {
+			ip, _, err := net.ParseCIDR(address)
+			if err != nil {
+				return false, fmt.Errorf("internal error - failed to parse outgoing nic address CIDR: %w", err)
+			}
+
+			// If an address is found in any machine network, we are done
+			if IsIPBelongsToAnyMachineNetwork(ip, mNetworks) {
+				return true, nil
```
linoyaslan, line 514: @AlonaKaplan PTAL

PR #7362, opened 2025-03-03
ACM-18347: additional SSH keys in InfraEnv

zszabo-rh:
[ACM-18347](https://issues.redhat.com//browse/ACM-18347): additional SSH keys in InfraEnv

New set of public SSH keys can be injected to the target hosts by listing them under spec/additionalSshAuthorizedKeys in InfraEnv CR:

```
apiVersion: agent-install.openshift.io/v1beta1
kind: InfraEnv
spec:
  sshAuthorizedKey: 'primary_ssh_key'
  additionalSshAuthorizedKeys:
  - 'additional_key_1'
  - 'additional_key_2'
  - 'additional_key_3'
```

After getting validated, all these keys are included in the sshAuthorizedKeys array of ignition config, and be applied in the target host:

```
{
  "ignition": {
    ...
  },
  "passwd": {
    "users": [
      {
        ...
        "sshAuthorizedKeys": [
          "primary_ssh_key",
          "additional_key_1",
          "additional_key_2",
          "additional_key_3"
        ]

      }
    ]
  },
  ...
}
```
The primary purpose of the new set of keys is to allow a whole group of users to access the provisioned nodes instead of sharing a single single keypair. The additional keys are not transferred to ClusterDeployment.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/v1beta1/infraenv_types.go
+++ b/api/v1beta1/infraenv_types.go
@@ -62,6 +62,11 @@ type InfraEnvSpec struct {
 	// +optional
 	SSHAuthorizedKey string `json:"sshAuthorizedKey,omitempty"`
 
+	// AdditionalSSHAuthorizedKeys is a set of additional SSH public keys that can be used for
+	// debugging the installation.
+	// +optional
+	AdditionalSSHAuthorizedKeys []string `json:"additionalSSHAuthorizedKeys,omitempty"`
+
 	// PullSecretRef is the reference to the secret to use when pulling images.
 	PullSecretRef *corev1.LocalObjectReference `json:"pullSecretRef"`
 
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -7423,6 +7423,9 @@ definitions:
       ssh_authorized_key:
         type: string
         description: SSH public key for debugging the installation.
+      additional_ssh_authorized_keys:
+        type: string
+        description: A set of additional SSH public keys that can be used for debugging the installation.
       pull_secret_set:
         type: boolean
         description: True if the pull secret has been added to the cluster.
@@ -7538,6 +7541,10 @@ definitions:
         type: string
         description: SSH public key for debugging the installation.
         x-nullable: true
+      additional_ssh_authorized_keys:
+        type: string
+        description: A set of additional SSH public keys that can be used for debugging the installation.
+        x-nullable: true
       pull_secret:
         type: string
         description: The pull secret obtained from Red Hat OpenShift Cluster Manager at console.redhat.com/openshift/install/pull-secret.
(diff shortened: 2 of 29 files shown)


swagger.yaml
```diff
@@ -7423,6 +7423,9 @@ definitions:
       ssh_authorized_key:
         type: string
         description: SSH public key for debugging the installation.
+      additional_ssh_authorized_keys:
+        type: string
```
pastequo, line 7427: MAybe a naive question but why those are not arrays as well ?

api/v1beta1/infraenv_types.go
```diff
@@ -62,6 +62,11 @@ type InfraEnvSpec struct {
 	// +optional
 	SSHAuthorizedKey string `json:"sshAuthorizedKey,omitempty"`
 
+	// AdditionalSSHAuthorizedKeys is a set of additional SSH public keys that can be used for
+	// debugging the installation.
+	// +optional
+	AdditionalSSHAuthorizedKeys []string `json:"additionalSshAuthorizedKeys,omitempty"`
```
rccrdpccl, line 68: Shouldn't this be `additionalSSHAuthorizedKeys` instead (same in the json manifest)?

zszabo-rh: Mostly I was trying to follow the naming patterns in all modules. But can it be even json mashaled if the field starts with lower case letter? I think it would be omitted by encoding/json, as a non-exported field.

rccrdpccl: Yes, I meant just the json representation of it, as SSH it's an acronym IMO we should keep it capitalized. The Golang name is fine as is.

swagger.yaml
zszabo-rh: The keys here are already in concatenated format ('\n' as separator). The validator expects them like this, probably as a remainder from earlier times when multiple ssh keys were also supported in the primary field. 
I wanted this change as lightweight as possible, so didn't refactored the way validation works.

api/v1beta1/infraenv_types.go
zszabo-rh: ok I see
yes it looks a bit strange now, I admit ;)

rccrdpccl:
Probably the jira issue should be MGMT and not ACM, but apart from this the change looks good

/lgtm
/hold

holding to give a chance to somebody else to review, feel free to unhold when necessary

zszabo-rh:
According to the Jira discussion, we will do a different approach: additional SSH keys will be injected to the cluster using install config override (AgentClusterInstall annotation).

PR #7402, opened 2025-03-13
MGMT-19980: TNA clusters

giladravid16:
Enhancement to define the implementation of installing TNA clusters.
See [openshift's enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/arbiter-clusters.md) for more details about TNA clusters.

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/arbiter-clusters.md
+++ b/docs/enhancements/arbiter-clusters.md
@@ -1,0 +1,142 @@
+---
+title: arbiter-clusters
+authors:
+  - "@giladravid16"
+creation-date: 2025-04-08
+last-updated: 2025-04-08
+---
+
+# Support 2 Node + 1 Arbiter Node HA Cluster
+
+## Summary
+
+This enhancement describes adding support in assisted-service to install TNA
+(Two Nodes + Arbiter) clusters - OpenShift clusters with control planes that
+consists of at minimum 2 normal-sized nodes, and at least 1 node that can be
+less powerful than the recommended node size.
+Arbiter nodes will only be running critical components for maintaining
+HA (e.g etcd) to allow the arbiter node size to be as small and as low cost as
+possible within reason.
+
+## Motivation
+
+Customers at the edge are requiring a more economical solution for HA deployments
+at the edge. They can support running 2 node clusters for redundancy but would like
+the option to deploy a lower cost node as an arbiter to supply the 3 nodes for etcd quorum.
+
+### Goals
+
+- Allow for a TNA cluster to be installed.
+- Allow adding arbiter nodes as day2 nodes to TNA clusters only.
+- Track usage of Arbiter nodes in metrics.
+
+### Non-Goals
+
+- Moving from 2 + 1 to a conventional 3 node cluster.
+
+## Proposal
+
+This enhancement will include the following:
+- In the API we will:
+    - Add a new host_role for arbiter.
+    - Allow setting cluster's control_plane_count to 2.
+    - Add more values to the enums for cluster's hyperthreading and disk
+      encryptions in order to specify that they should be enabled on arbiters.
+    - Add a new feature_support_level_id for TNA.
+- We will update the clusters' validations to support TNA clusters.
+- We will update the clusters' transition handler to consider arbiter nodes.
+- We will add arbiter nodes' minimum requirements for clusters and per operator.
+- We will auto-assign the arbiter role to a host if the cluster's control_plane_count
+  is 2.
+
+### User Stories
+
+N/A - TNA clusters are a future feature in openshift, and the
+[openshift enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/arbiter-clusters.md#user-stories)
+has user stories that are not specific for assisted-service.
+
+### Implementation Details/Notes/Constraints
+
+The enums for cluster's hyperthreading and disk encryption will be:
+- 'none'
+- 'masters'
+- 'arbiters'
+- 'workers'
+- 'masters,arbiters'
+- 'masters,workers'
+- 'arbiters,workers'
+- 'masters,arbiters,workers'
+- 'all' which will be the same as 'masters,arbiters,workers'
+
+When generating the ignitions for a TNA cluster, a new base ignition named
+arbiter.ign will be created. If it exists we need to patch it and upload it
+to s3. The list of patches arbiter.ign needs:
+- Set CA cert if it was provided.
+- If all the hosts in the cluster have IPv6 address then we need to add a
+  NetworkManager configuration ensuring that IPv6 DHCP requests use consistent
+  client identification.
+
+Our validations will need to be updated as follows:
+- Clusters can set control_plane_count to be 2. Clusters that do that must
+  add at least one more host that will be assigned the arbiter role.
+- TNA clusters' platform must be baremetal.
+- TNA clusters' ocp version must be at least 4.19.
+
+In the cluster's transition handler, the function enoughMastersAndWorkers
+needs to be updated to also check for arbiters. This is needed to correctly
+decide if the cluster is in installing/finalizing.
+
+We need to add arbiter's hardware requirements in:
+- The HW_VALIDATOR_REQUIREMENTS environment variable (by default we'll take
+  the worker's requirements to avoid issues mid-development).
+- The HostTypeHardwareRequirementsWrapper structs used by the operators.
+
+We will auto-assign the arbiter role to a host if all the following are true:
+- The cluster's control_plane_count is 2.
+- The cluster doesn't have any host that is assigned the arbiter role.
+- The host has the minimum hardware requirements for arbiter nodes.
+- Either the host doesn't have the minimum hardware requirements for master
+  nodes, or the host's cluster has enough master nodes.
+
+Outside of assisted-service, we also need to do the following:
+- After we update the API in assisted-services, we need to update the vendor
+  directory in assisted-installer-agent.
+- We need to update the function waitForMinMasterNodes in assisted-installer
+  to set the min master nodes to be the minimum of 2 (the current value)
+  and number of control planes (read it from the install-config) minus 1.
+
+Until we have this feature fully implemented, we should have an environment
+variable to decide if we allow TNA clusters (default value is false).
+
+### Risks and Mitigations
+
+N/A
+
+## Design Details
+
+### Open Questions
+
+- We should disable this feature in the kube-api and when using ABI
+  until after this feature is officially added to them. We should
+  be in contact with them to know when this should be enabled.
+
+### UI Impact
+
+- In the host discovery we need to allow setting a host's role to be arbiter.
+- In the cluster details:
+    - We need to allow setting control_plane_count to 2 if the openshift version
+      is at least 4.19.
+    - We need a checkbox for setting arbiters' disk-encryption.
+
+### Test Plan
+
+- Add an e2e test that will create a TNA cluster.
+- Add unit tests and subsystem tests for arbiter nodes and TNA clusters.
+
+## Drawbacks
+
+N/A
+
+## Alternatives
+
+N/A


docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+- TNA clusters can have 2-5 control plane nodes.
+- TNA clusters' platform must be baremetal.
+- TNA clusters are expected to be TP stating from OCP 4.19.
+
+We need to add arbiter's hardware requirements in:
+- The HW_VALIDATOR_REQUIREMENTS environment variable (by default we'll take
+  the worker's requirements to avoid issues mid-development).
+- The HostTypeHardwareRequirementsWrapper structs used by the operators.
+
+We will auto-assign the arbiter role to a host if the following are true:
```
tsorya, line 72: Probably we need to assign it to the host with lowest HW in case all 3 matches HW requirements to be masters?

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+### Non-Goals
+
+- Moving from 2 + 1 to a conventional 3 node cluster.
+
+## Proposal
+
+This enhancement will include the following:
+- In the API we will:
+  - Add a new host_role for arbiter.
+  - Add a new high_availability_mode for TNA.
```
gamli75, line 40: the high_availability_mode is deprecated, and we will use control_plane_count instead. 
I'm not sure if 3 CP == 2 CP with 1 arbiter.
@danmanor WDYT?

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+  - Deprecate the existing parameters and create new ones for hyperthreading
+    and disk encryption (they will be maps of host_role to bool).
+  - Add a new feature_support_level_id for TNA.
+- We will update the clusters' validations and transitions for TNA clusters.
```
gamli75, line 44: can you add more details about the validations that will be affected by this change?

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+- As a solutions architect for cloud infrastructures, I want to offer low cost
+  OpenShift deployments on purpose built hardware for a 2 + 1 configuration.
+- As an OpenShift cluster admin I want non-critical applications deployed to my
+  2 + 1 arbiter node cluster to not be scheduled to run on the arbiter node.
+
+### Implementation Details/Notes/Constraints
+
+Our validations will need to be updated as follows:
+- A host can only be assigned the arbiter role if the cluster is TNA.
+- TNA clusters must have at least 1 arbiter node.
```
gamli75, line 62: Can we add a host with an arbiter role on day 2? If not, let's add it here

docs/enhancements/arbiter-clusters.md
giladravid16: My initial idea was to check for the lowest HW that pass the requirements, but then I saw that right now for masters we auto-select the first ones we find without comparing it to the other hosts (even if it's a total waste of resources) and I assume it's because it can be difficult to determine which host is smallest (we would need to compare [ClusterHostRequirementsDetails](https://github.com/openshift/assisted-service/blob/master/models/cluster_host_requirements_details.go#L18)).

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Test Plan
+
+- Add an e2e test that will create a TNA cluster.
+- Add unit tests and subsystem tests for arbiter nodes and TNA clusters.
+
+## Drawbacks
+
+A few drawbacks we have is that we will be creating a new variant of OpenShift
+that implements a new unique way of doing HA for kubernetes. This does mean
```
carbonin, line 124: This isn't a drawback of our implementation. This sounds like it's talking more about TNA generally.

This enhancement should focus on the changes in assisted installer rather than TNA as an OpenShift topology.

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+We need to add arbiter's hardware requirements in:
+- The HW_VALIDATOR_REQUIREMENTS environment variable (by default we'll take
+  the worker's requirements to avoid issues mid-development).
+- The HostTypeHardwareRequirementsWrapper structs used by the operators.
+
+We will auto-assign the arbiter role to a host if the following are true:
+- The host's cluster is TNA.
+- The host's cluster has no arbiter nodes.
+- The host has the minimum hardware requirements for arbiter nodes.
+
```
carbonin, line 76: If you're trying to be exhaustive here, we also should call out that there will be a new ignition file coming from the installer that we need to parse and apply our changes to. I don't see that referenced anywhere.

docs/enhancements/arbiter-clusters.md
giladravid16: That's why I thought of using high_availability_mode, in openshift-installer they set the ControlPlaneTopology to HighlyAvailableArbiter and InfrastructureTopology to HighlyAvailable for this reason.

docs/enhancements/arbiter-clusters.md
giladravid16: I checked and it is allowed to add arbiter nodes as day2 nodes, if the cluster was installed as TNA.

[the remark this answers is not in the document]
giladravid16: The openshift-installer does not set a max limit on the number of arbiter nodes, it just has to be at least 1.

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+  - Deprecate the existing parameters and create new ones for hyperthreading
+    and disk encryption (they will be maps of host_role to bool).
+  - Add a new feature_support_level_id for TNA.
+- We will update the clusters' validations and transitions for TNA clusters.
+- We will add arbiter nodes' minimum requirements for clusters and per operator.
+- We will auto-assign the arbiter role to a host.
+
+### User Stories
+
+- As a solutions architect for a retail organization, I want to deploy OpenShift
```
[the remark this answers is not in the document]
giladravid16: @gamli75 are there any user stories of our customers wanting a topology like this?

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+- As an OpenShift cluster admin I want non-critical applications deployed to my
+  2 + 1 arbiter node cluster to not be scheduled to run on the arbiter node.
+
+### Implementation Details/Notes/Constraints
+
+Our validations will need to be updated as follows:
+- A host can only be assigned the arbiter role if the cluster is TNA.
+- TNA clusters must have at least 1 arbiter node.
+- TNA clusters can have 2-5 control plane nodes.
+- TNA clusters' platform must be baremetal.
```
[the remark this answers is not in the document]
giladravid16: The openshift-installer only allows TNA clusters on baremetal for now, it might change in the future.

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+  OpenShift deployments on purpose built hardware for a 2 + 1 configuration.
+- As an OpenShift cluster admin I want non-critical applications deployed to my
+  2 + 1 arbiter node cluster to not be scheduled to run on the arbiter node.
+
+### Implementation Details/Notes/Constraints
+
+Our validations will need to be updated as follows:
+- A host can only be assigned the arbiter role if the cluster is TNA.
+- TNA clusters must have at least 1 arbiter node.
+- TNA clusters can have 2-5 control plane nodes.
```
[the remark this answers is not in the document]
giladravid16: The openshift-installer allows it, so I assume we should also allow it.

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,134 @@
...
+- Moving from 2 + 1 to a conventional 3 node cluster.
+
+## Proposal
+
+This enhancement will include the following:
+- In the API we will:
+  - Add a new host_role for arbiter.
+  - Add a new high_availability_mode for TNA.
+  - Deprecate the existing parameters and create new ones for hyperthreading
+    and disk encryption (they will be maps of host_role to bool).
```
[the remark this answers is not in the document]
giladravid16: Currently hyperthreading is one of ['masters', 'workers', 'none', 'all'], I think it's better to define a struct that has a boolean for each role then add more combinations with arbiter.
Disk encryption is similar with it's enable_on field, I would add a new field.

docs/enhancements/arbiter-clusters.md
gamli75: If we want to handle it in day2 - let's add it to the enhancement

docs/enhancements/arbiter-clusters.md
avishayt: As @gamli75 said, high_availability_mode is deprecated in the Assisted API.

docs/enhancements/arbiter-clusters.md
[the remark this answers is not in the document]
avishayt: From what I can tell, hyperthreading will only affect the REST API, so that's easy.  Disk encryption would break the k8s API, which we should avoid.

I recommend extending the possibilities to avoid breaking the API.  You can add:
'masters,workers'
'masters,arbiter'
'workers,arbiter'
'masters,workers,arbiter'

'all' would now mean 'masters,workers,arbiter'.
Using 'masters,workers,arbiter' would be future-proof in case we add another type later.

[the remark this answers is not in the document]
giladravid16: Why would it break the k8s API?

docs/enhancements/arbiter-clusters.md
giladravid16: I added arbiter_count instead

docs/enhancements/arbiter-clusters.md
giladravid16: I added it as a goal

docs/enhancements/arbiter-clusters.md
giladravid16: I added it

carbonin:
This looks fine to me

@tsorya @avishayt @gamli75 do you want to take another look?

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,144 @@
...
+### Open Questions
+
+N/A
+
+### UI Impact
+
+- In the host discovery we need to allow setting a host's role to be arbiter.
+- In the cluster details:
+  - we need to allow setting the arbiter_count, with zero as the default.
+  - We need to allow setting control_plane_count to 2, but then force setting
```
gamli75, line 129: @carbonin Will this be correct when we have the Two Node OpenShift with Fencing?

carbonin: Right, I think this is fine for now when we only have TNA, but I don't think there's anything better we can do until we also need to handle TNF.

carbonin: I suppose I take this back based on the conversation in the kickoff.

It might be better to just validate that we have a good combination rather than forcing arbiters to 1 since that's something we'd need to pull out once we have TNF.

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,144 @@
...
+### Non-Goals
+
+- Moving from 2 + 1 to a conventional 3 node cluster.
+
+## Proposal
+
+This enhancement will include the following:
+- In the API we will:
+  - Add a new host_role for arbiter.
+  - Add a new field for clusters named arbiter_count. The default value will
```
carbonin, line 41: Based on the kickoff it would be better to not require this up front.

We should be able to infer when an arbiter would be required and do autoassign and validations accordingly after discovery.

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,144 @@
...
+
+### User Stories
+
+N/A - TNA clusters a future feature in openshift, the
+[openshift enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/arbiter-clusters.md#user-stories)
+has user stories that are not specific for assisted-service.
+
+### Implementation Details/Notes/Constraints
+
+The enums for cluster's hyperthreading and disk encryption will be:
```
carbonin, line 61: During kickoff it was raised that we may want to evaluate the usage of the disk encryption feature, but I'm not sure that really affects what we want to handle in this feature.

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Risks and Mitigations
+
+N/A
+
+## Design Details
+
+### Open Questions
+
+N/A
```
carbonin, line 122: We should add ABI concerns here. We should either ensure it works or disable it explicitly.

carbonin: Same with kube-api for the first pass. We should disable this explicitly in the controllers if we're not going to support it on the first pass.

docs/enhancements/arbiter-clusters.md
```diff
@@ -0,0 +1,144 @@
...
+HA (e.g etcd) to allow the arbiter node size to be as small and as low cost as
+possible within reason.
+
+## Motivation
+
+Customers at the edge are requiring a more economical solution for HA deployments
+at the edge. They can support running 2 node clusters for redundancy but would like 
+the option to deploy a lower cost node as an arbiter to supply the 3 nodes for etcd quorum.
+
+### Goals
```
carbonin, line 27: Maybe add a goal for tracking this feature usage in metrics.

carbonin (approved):

PR #7429, opened 2025-03-19
MGMT-20134: Disable OLM Catalog for CAPI CI tests

CrystalChun:
Starting from ACM 2.13/MCE 2.8, Hypershift
blocks and prevents installation of hosts to a hosted cluster if the catalog sources are unreachable,
which will always happen in the disconnected case unless the catalog is mirrored.

To resolve this, the OLM catalog will now be disabled during the hosted cluster creation and the OLM catalog is now set to guest mode.

---
Note: this will need to be backported to ACM 2.13 branch

## List all the issues related to this PR

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [x] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [x] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [ ] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [ ] dev-scripts environment
- [ ] Reviewer's test appreciated
- [x] Waiting for CI to do a full test run
- [x] Manual (Elaborate on how it was tested): Tested on rehearsal CI
- [ ] No tests needed

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #7438, opened 2025-03-20
MGMT-18437: Consumer label on AgentClusterInstall to identify clusters used by CAPI providers

CrystalChun:
Defines the label `"agentclusterinstalls.agent-install.openshift.io/consumer"` in the `AgentClusterInstall` that can be set by the CAPI OpenshiftAssisted providers to indicate the cluster is used by these providers. 

The value of the label is added to the `tags` field of the cluster, which is sent with the cluster data during on-prem data collection

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

/cc @rccrdpccl

### DIFF
--- a/internal/controller/controllers/clusterdeployments_controller.go
+++ b/internal/controller/controllers/clusterdeployments_controller.go
@@ -1176,6 +1176,11 @@ func (r *ClusterDeploymentsReconciler) updateIfNeeded(
 		update = true
 	}
 
+	if label, exists := clusterInstall.GetLabels()[hiveext.ClusterConsumerLabel]; exists && label != cluster.Tags {
+		params.Tags = &label
+		update = true
+	}
+
 	if !update {
 		return cluster, nil
 	}
@@ -1462,6 +1467,10 @@ func CreateClusterParams(clusterDeployment *hivev1.ClusterDeployment, clusterIns
 
 	if len(olmOperators) != 0 {
 		clusterParams.OlmOperators = olmOperators
+	}
+
+	if label, exists := clusterInstall.GetLabels()[hiveext.ClusterConsumerLabel]; exists && label != "" {
+		clusterParams.Tags = &label
 	}
 
 	return clusterParams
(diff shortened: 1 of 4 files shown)


CrystalChun:
NOTE: once this merges, the AgentClusterInstall label is a constant which can be used once we pull the latest assisted changes into our CAPI providers repo

swagger.yaml
```diff
@@ -5021,6 +5021,9 @@ definitions:
         x-nullable: true
       load_balancer:
         $ref: '#/definitions/load_balancer'
+      consumer:
```
rccrdpccl, line 5024: Is it necessary to change the API? I think if we can avoid it it would be best, as I can't really think of any other scenario where this is needed.

rccrdpccl:
I think the overall approach is good, my only concern is exposing this field in the API.

swagger.yaml
CrystalChun: Hmm I guess we can avoid it by either:

- adding another optional parameter during create/update cluster 
- or creating a new function to just store this value 

What do you think?

gamli75: Why do we need to add a new parameter? Can we use the tags field for this?
https://github.com/openshift/assisted-service/blob/master/swagger.yaml#L5017 
the UI is using it and we have it in our data

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1464,6 +1469,10 @@ func CreateClusterParams(clusterDeployment *hivev1.ClusterDeployment, clusterIns
 		clusterParams.OlmOperators = olmOperators
 	}
 
+	if label := clusterInstall.GetLabels()[hiveext.ClusterConsumerLabel]; label != "" {
```
rccrdpccl, line 1472: nit: we should test for the key instead of relying on the 0 value, same in the update function

internal/controller/controllers/clusterdeployments_controller.go
```diff
@@ -1176,6 +1176,11 @@ func (r *ClusterDeploymentsReconciler) updateIfNeeded(
 		update = true
 	}
 
+	if label := clusterInstall.GetLabels()[hiveext.ClusterConsumerLabel]; label != cluster.Tags {
+		params.Tags = &label
```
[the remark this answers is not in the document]
CrystalChun: Yes!

internal/controller/controllers/clusterdeployments_controller.go
CrystalChun: Modified! How does it look?

CrystalChun:
This test will fail until https://github.com/openshift/assisted-service/pull/7429 is merged

rccrdpccl:
/retest

flaky konflux

rccrdpccl:
/retest

:unamused:

PR #7448, opened 2025-03-24
MGMT-20207: avoid adding system CA bundle to AdditionalTrustBundle

danielerez:
When generating the install config, it should not include the entire system CA bundle. I.e. when setting the AdditionalTrustBundle[1]. However, when using a mirror regsitry (e.g. with MirrorRegistryRef[2] in ASC), the installConfigBuilder is adding the content of tls-ca-bundle.pem[3]. This pem file is created[4] by ASC controller, and includes the full system CA bundle since the bundle[5] is injected into 'cluster-trusted-ca-bundle'.

Therefore, the suggestion solution is to create a separate pem file just for the certificates specified in MirrorRegistryRef. I.e. generate a new 'user-registry-ca-bundle.pem' file, that will be included as part of AdditionalTrustBundle. These certificates will propagate into the 'user-ca-bundle' CM during cluster installation.
This will ensure that 'user-ca-bundle' CM indeed includes only custom certificates mandatory for the user, instead of the system CA bundle.

Note: as backwards compatibility, for flows like ABI, we keep a fallback to the current behaviour (i.e. include 'tls-ca-bundle.pem' only when 'user-registry-ca-bundle.pem' doesn't exist).

[1] https://github.com/openshift/assisted-service/blob/40ab10db5e872e519ab0a97e82fc318423feeaba/internal/installcfg/builder/builder.go#L275
[2] https://github.com/openshift/assisted-service/blob/cb169a2d2c97bb3dccd06ad4b75f2937e01f78f4/vendor/github.com/openshift/assisted-service/api/v1beta1/agentserviceconfig_types.go#L82
[3] https://github.com/openshift/assisted-service/blob/a1c3229afee1f0f774b286283fb0d0098b9eac03/internal/common/common.go#L35
[4] https://github.com/openshift/assisted-service/blob/341f9860c455cccc42741f350024d05aa72755f8/internal/controller/controllers/agentserviceconfig_controller.go#L1848
[5] https://github.com/openshift/assisted-service/blob/341f9860c455cccc42741f350024d05aa72755f8/internal/controller/controllers/agentserviceconfig_controller.go#L1059

- [ ] New Feature 
- [ ] Enhancement 
- [x] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [ ] None

## How was this code tested?

- [x] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/pkg/mirrorregistries/generator.go
+++ b/pkg/mirrorregistries/generator.go
@@ -18,12 +18,14 @@ type ServiceMirrorRegistriesConfigBuilder interface {
 type mirrorRegistriesConfigBuilder struct {
 	MirrorRegistriesConfigPath      string
 	MirrorRegistriesCertificatePath string
+	SystemCertificateBundlePath     string
 }
 
 func New() ServiceMirrorRegistriesConfigBuilder {
 	return &mirrorRegistriesConfigBuilder{
 		MirrorRegistriesConfigPath:      common.MirrorRegistriesConfigPath,
 		MirrorRegistriesCertificatePath: common.MirrorRegistriesCertificatePath,
+		SystemCertificateBundlePath:     common.SystemCertificateBundlePath,
 	}
 }
 
@@ -59,7 +61,12 @@ func (m *mirrorRegistriesConfigBuilder) IsMirrorRegistriesConfigured() bool {
 // the mirror registries are not configured.
 // empty dir is due to the way we mao configmap in the assisted-service pod
 func (m *mirrorRegistriesConfigBuilder) GetMirrorCA() ([]byte, error) {
-	return os.ReadFile(m.MirrorRegistriesCertificatePath)
+	bytes, err := os.ReadFile(m.MirrorRegistriesCertificatePath)
+	if err != nil {
+		// fallback to tls-ca-bundle.pem (used by ABI)
+		return os.ReadFile(m.SystemCertificateBundlePath)
+	}
+	return bytes, nil
 }
 
 // GetMirrorRegistries returns error if the file is not present, which will also indicate that
--- a/pkg/mirrorregistries/mirror_registry_test.go
+++ b/pkg/mirrorregistries/mirror_registry_test.go
@@ -132,6 +132,56 @@ var _ = Describe("Generator tests", func() {
 		})
 	})
 
+	var _ = Describe("GetMirrorCA", func() {
+		var (
+			tempRegistryCA *os.File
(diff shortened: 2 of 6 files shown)


pkg/mirrorregistries/mirror_registry_test.go
```diff
@@ -129,6 +130,19 @@ var _ = Describe("Generator tests", func() {
 				res := m.IsMirrorRegistriesConfigured()
 				Expect(res).Should(Equal(true))
 			})
+
+			It("returns true when only the system CA exists (and registry.conf has mirrors)", func() {
```
carbonin, line 134: I don't understand this test. Why is there a conditional and why does `IsMirrorRegistriesConfigured` have anything to do with the system CA, shouldn't it be based on the user one?

carbonin (commented):
Just one comment, otherwise looks good.

pkg/mirrorregistries/mirror_registry_test.go
danielerez: Yeah, fair enough. So added a new set of tests for GetMirrorCA.

pkg/mirrorregistries/generator.go
```diff
@@ -59,7 +63,12 @@ func (m *mirrorRegistriesConfigBuilder) IsMirrorRegistriesConfigured() bool {
 // the mirror registries are not configured.
 // empty dir is due to the way we mao configmap in the assisted-service pod
 func (m *mirrorRegistriesConfigBuilder) GetMirrorCA() ([]byte, error) {
-	return os.ReadFile(m.MirrorRegistriesCertificatePath)
+	bytes, err := m.fileReader(m.MirrorRegistriesCertificatePath)
+	if err != nil {
+		// fallback to tls-ca-bundle.pem (used by ABI)
+		return m.fileReader(common.SystemCertificateBundlePath)
```
carbonin, line 69: Instead of injecting the reader function, can we set this path on the struct?
It will be less likely to confuse folks in the future that don't know why this `fileReader` exists

danielerez: Sure, good idea.

carbonin (approved):

danmanor:
@giladravid16 @pastequo  Do you have an idea why these 2 konflux jobs are not triggering for a long time ?

PR #7458, opened 2025-03-25
NO-JIRA: OKD: Use /tmp as socket directory when calling pg_ctl and postgres

Prashanth684:
OKD installations are failing when using agent and assisted service as described here: https://github.com/okd-project/okd/discussions/2068#discussioncomment-11733877 with:

```
FATAL:  could not create lock file "/var/run/postgresql/.s.PGSQL.5432.lock": No such file or directory"
```

Explicitly tell postgres to use /tmp as the unix socket directory to fix the issue.

### DIFF
--- a/hack/agent_installer/start_db.sh
+++ b/hack/agent_installer/start_db.sh
@@ -4,9 +4,9 @@ set -o nounset
 set -euo pipefail
 set -o errexit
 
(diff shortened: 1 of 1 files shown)


andfasano:
Let's verify the full agent suite

/test e2e-agent-sno-ipv6
/test e2e-agent-ha-dualstack
/test e2e-agent-4control-ipv4
/test e2e-agent-5control-ipv4

rwsu (commented):
/lgtm

PR #7391, opened 2025-03-10
AGENT-538: Add ability for ABI UI to generate kubeconfig

bfournie:
This adds the ability for ABI, via the OVE UI, to generate a kubeconfig when the cluster is in the ready state. Instead of adding a new endpoint to allow the UI to generate the kubeconfig, the enviroment variable - EPHEMERAL_INSTALLER_CLUSTER_TLS_CERTS_OVERRIDE_DIR - which is specific to ABI, is used. If no TLS certs have been added to the override it indicates that the certs, and the kubeconfig, should be generated. The kubeconfig can then be retrieved in the ready or preparing-for-installation states.

## List all the issues related to this PR

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

## What environments does this code impact?

- [ ] Automation (CI, tools, etc)
- [ ] Cloud
- [ ] Operator Managed Deployments
- [x] None

## How was this code tested?

- [ ] assisted-test-infra environment
- [x] dev-scripts environment
- [ ] Reviewer's test appreciated
- [ ] Waiting for CI to do a full test run
- [ ] Manual (Elaborate on how it was tested)
- [x] No tests needed

## Checklist

- [x] Title and description added to both, commit and PR.
- [x] Relevant issues have been associated (see [CONTRIBUTING] guide)
- [x] This change does not require a documentation update (docstring, `docs`, README, etc)
- [ ] Does this change include unit-tests (note that code changes require unit-tests)

## Reviewers Checklist

- Are the title and description (in both PR and commit) meaningful and clear?
- Is there a bug required (and linked) for this change?
- Should this PR be backported?

[Kubernetes community documentation]: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines
[CONTRIBUTING]: https://github.com/openshift/assisted-service/blob/master/CONTRIBUTING.md

PR #7472, opened 2025-03-27
AGENT-538: For ABI, also allow early retrieval of kubeadmin-password

bfournie:
In https://github.com/openshift/assisted-service/pull/7391 changes were made to allow the kubeconfig to be retrieved in the Ready state for ABI only, in order to be integrate with the OVE UI. This change should also be made for kubeadmin-password, to allow it be retrieved prior to installing the cluster.

Some related changes address additional comments in the PR above.

- [ ] New Feature 
- [x] Enhancement 
- [] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [x] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4067,8 +4067,7 @@ func (b *bareMetalInventory) checkFileForDownload(ctx context.Context, clusterID
 	case constants.KubeadminPassword:
 		fallthrough
 	case constants.KubeconfigNoIngress:
-		agentInstaller := b.installerInvoker == "agent-installer"
-		err = clusterPkg.CanDownloadKubeconfigFiles(cluster, fileName, agentInstaller)
+		err = clusterPkg.CanDownloadKubeconfigFiles(cluster, fileName, b.installerInvoker)
 	case constants.ManifestFolder:
 		// do nothing. manifests can be downloaded at any given cluster state
 	default:
--- a/internal/bminventory/inventory.go
+++ b/internal/bminventory/inventory.go
@@ -4064,9 +4064,11 @@ func (b *bareMetalInventory) checkFileForDownload(ctx context.Context, clusterID
 	switch fileName {
 	case constants.Kubeconfig:
 		err = clusterPkg.CanDownloadKubeconfig(cluster)
+	case constants.KubeadminPassword:
+		fallthrough
 	case constants.KubeconfigNoIngress:
-		agentInstaller := b.Config.ClusterTLSCertOverrideDir != ""
-		err = clusterPkg.CanDownloadKubeconfigNoIngress(cluster, agentInstaller)
+		agentInstaller := b.installerInvoker == "agent-installer"
+		err = clusterPkg.CanDownloadKubeconfigFiles(cluster, fileName, agentInstaller)
 	case constants.ManifestFolder:
 		// do nothing. manifests can be downloaded at any given cluster state
 	default:
@@ -6219,22 +6221,24 @@ func (b *bareMetalInventory) V2DownloadClusterCredentials(ctx context.Context, p
 	fileName := params.FileName
 	respBody, contentLength, err := b.V2DownloadClusterCredentialsInternal(ctx, params)
 
-	// Kubeconfig-noingress has been created during the installation, but it does not have the ingress CA.
-	// At the finalizing phase, we create the kubeconfig file and add the ingress CA.
-	// An ingress CA isn't required for normal login but for oauth login which isn't a common use case.
-	// Here we fallback to the kubeconfig-noingress for the kubeconfig filename.
-	if err != nil && fileName == constants.Kubeconfig {
-		fileName = constants.KubeconfigNoIngress
+	if err != nil {
+		// Kubeconfig-noingress has been created during the installation, but it does not have the ingress CA.
+		// At the finalizing phase, we create the kubeconfig file and add the ingress CA.
+		// An ingress CA isn't required for normal login but for oauth login which isn't a common use case.
+		// Here we fallback to the kubeconfig-noingress for the kubeconfig filename.
+		if fileName == constants.Kubeconfig {
+			fileName = constants.KubeconfigNoIngress
+			respBody, contentLength, err = b.v2DownloadClusterFilesInternal(ctx, fileName, params.ClusterID.String())
+		}
 
-		if b.Config.ClusterTLSCertOverrideDir != "" {
-			// Since only ABI uses ClusterTLSCertOverrideDir, if it is set then
-			// kubeconfig must be generated here for retrieval prior to cluster installation.
+		if err != nil && common.IsNotFoundError(err) && b.installerInvoker == "agent-installer" {
+			// For ABI, kubeconfig must be generated here for retrieval prior to cluster installation, only if file not found.
 			if agentErr := b.generateAgentInstallerKubeconfig(ctx, params); agentErr != nil {
 				b.log.WithError(agentErr).Errorf("failed to generate agent installer kubeconfig")
 				return common.GenerateErrorResponder(agentErr)
 			}
+			respBody, contentLength, err = b.v2DownloadClusterFilesInternal(ctx, fileName, params.ClusterID.String())
 		}
-		respBody, contentLength, err = b.v2DownloadClusterFilesInternal(ctx, fileName, params.ClusterID.String())
 	}
 
 	if err != nil {
--- a/pkg/generator/generator.go
+++ b/pkg/generator/generator.go
@@ -65,7 +65,14 @@ func New(log logrus.FieldLogger, s3Client s3wrapper.API, cfg Config,
 // GenerateInstallConfig creates install config and ignition files
 func (k *installGenerator) GenerateInstallConfig(ctx context.Context, cluster common.Cluster, cfg []byte, releaseImage, installerReleaseImageOverride string) error {
 	log := logutil.FromContext(ctx, k.log)
-	err := os.MkdirAll(k.workDir, 0o755)
+
+	entries, err := os.ReadDir(k.workDir)
+	if err == nil && len(entries) > 0 {
+		log.Infof("Work directory %s already exists, skipping generation of InstallConfig", k.workDir)
+		return nil
+	}
+
+	err = os.MkdirAll(k.workDir, 0o755)
 	if err != nil {
 		return err
 	}
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -703,8 +703,8 @@ func CanDownloadFiles(c *common.Cluster) (err error) {
 	return checkDownloadAllowed(c, "files", allowedStatuses)
 }
 
-func CanDownloadKubeconfigFiles(c *common.Cluster, file string, agentInstaller bool) (err error) {
-	if !agentInstaller {
+func CanDownloadKubeconfigFiles(c *common.Cluster, file, installerInvoker string) (err error) {
+	if installerInvoker != "agent-installer" {
 		return CanDownloadFiles(c)
 	}
 
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -666,7 +666,6 @@ func (m *Manager) ClusterMonitoring() {
 }
 
 func getDownloadFilesAllowedStatuses() []string {
-
 	return []string{
 		models.ClusterStatusInstalling,
 		models.ClusterStatusFinalizing,
@@ -678,19 +677,16 @@ func getDownloadFilesAllowedStatuses() []string {
 	}
 }
 
-func CanDownloadFiles(c *common.Cluster) (err error) {
+func checkDownloadAllowed(c *common.Cluster, file string, allowed []string) (err error) {
 	clusterStatus := swag.StringValue(c.Status)
-
-	allowedStatuses := getDownloadFilesAllowedStatuses()
-	if !funk.Contains(getDownloadFilesAllowedStatuses(), clusterStatus) {
-		err = errors.Errorf("cluster %s is in %s state, files can be downloaded only when status is one of: %s",
-			c.ID, clusterStatus, allowedStatuses)
+	if !funk.Contains(allowed, clusterStatus) {
+		err = errors.Errorf("cluster %s is in %s state, %s can be downloaded only when status is one of: %s",
+			c.ID, clusterStatus, file, allowed)
 	}
 	return err
 }
 
 func CanDownloadKubeconfig(c *common.Cluster) (err error) {
-	clusterStatus := swag.StringValue(c.Status)
 	allowedStatuses := []string{
 		models.ClusterStatusFinalizing,
 		models.ClusterStatusInstalled,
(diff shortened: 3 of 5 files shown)


internal/bminventory/inventory.go
```diff
@@ -6219,22 +6221,26 @@ func (b *bareMetalInventory) V2DownloadClusterCredentials(ctx context.Context, p
...
+		// At the finalizing phase, we create the kubeconfig file and add the ingress CA.
+		// An ingress CA isn't required for normal login but for oauth login which isn't a common use case.
+		// Here we fallback to the kubeconfig-noingress for the kubeconfig filename.
+		if fileName == constants.Kubeconfig {
+			fileName = constants.KubeconfigNoIngress
+		}
+		respBody, contentLength, err = b.v2DownloadClusterFilesInternal(ctx, fileName, params.ClusterID.String())
 
-		if b.Config.ClusterTLSCertOverrideDir != "" {
+		if err != nil && (b.installerInvoker == "agent-installer" && b.Config.ClusterTLSCertOverrideDir != "") {
```
andfasano, line 6234: Could be simplified to

```suggestion
		if err != nil && (b.installerInvoker == "agent-installer") {
```

internal/bminventory/inventory.go
```diff
@@ -6219,22 +6221,26 @@ func (b *bareMetalInventory) V2DownloadClusterCredentials(ctx context.Context, p
...
-		fileName = constants.KubeconfigNoIngress
+	if err != nil {
+		// Kubeconfig-noingress has been created during the installation, but it does not have the ingress CA.
+		// At the finalizing phase, we create the kubeconfig file and add the ingress CA.
+		// An ingress CA isn't required for normal login but for oauth login which isn't a common use case.
+		// Here we fallback to the kubeconfig-noingress for the kubeconfig filename.
+		if fileName == constants.Kubeconfig {
+			fileName = constants.KubeconfigNoIngress
+		}
+		respBody, contentLength, err = b.v2DownloadClusterFilesInternal(ctx, fileName, params.ClusterID.String())
```
andfasano, line 6232: I'm not clear about this one.

On the first call (so when kubeconfig/kubeadmin-password haven't been generated yet) then b.V2DownloadClusterCredentialsInternal at [L6220](https://github.com/openshift/assisted-service/pull/7472/files#diff-747a429a0e12307ef65d754c83b6898ce3858b621005e1e7af6e420da1f1657cL6220) will fail (a specific error could be useful to trap the particular condition btw)

The previous implementation, in case of ABI, generated those files by calling `generateAgentInstallerKubeconfig()` , and then re-invoked `b.v2DownloadClusterFilesInternal` to allow the current dowload operation to succeed.

After that, any subsequent call to `V2DownloadClusterCredentials` is expected to succeed - assuming that the env conditions are correct (== allowed states), so no need to have another explicit `v2DownloadClusterFilesInternal` call.

I think that the main issue here is that the new code, specifically: 
```
if agentErr := b.generateAgentInstallerKubeconfig(ctx, params); agentErr != nil {
  b.log.WithError(agentErr).Errorf("failed to generate agent installer kubeconfig")
  return common.GenerateErrorResponder(agentErr)
}
respBody, contentLength, err = b.v2DownloadClusterFilesInternal(ctx, fileName, params.ClusterID.String())
```
gets executed unconditionally in case of error (regardless of the error), and not only when the requested files are missing.

IINW the current [Download](https://github.com/openshift/assisted-service/blob/d00ddaef174e781057c44e98f7dcb108664a99d8/pkg/s3wrapper/filesystem.go#L227) method returns a `common.NotFound` error (embbeded in a [NewApiError](https://github.com/openshift/assisted-service/blob/d00ddaef174e781057c44e98f7dcb108664a99d8/internal/bminventory/inventory.go#L6272), so I think it would be sufficient trigger the files generation (and just another download call) only when the returned error is a `common.NotFound`

bfournie: Good point, will add an error check for file not found. However I do need a 2nd call to `v2DownloadClusterFilesInternal` when checking `KubeconfigNoIngress` after the  first check for `Kubeconfig` fails which is how it worked previously.

internal/bminventory/inventory.go
```diff
@@ -6219,22 +6221,29 @@ func (b *bareMetalInventory) V2DownloadClusterCredentials(ctx context.Context, p
...
+			// For ABI, kubeconfig must be generated here for retrieval prior to cluster installation, only if file not found.
+			if _, ok := err.(*common.ApiErrorResponse); ok {
+				notFoundErr := err.(*common.ApiErrorResponse).Err()
+				if _, ok := notFoundErr.(common.NotFound); ok {
+					if agentErr := b.generateAgentInstallerKubeconfig(ctx, params); agentErr != nil {
+						b.log.WithError(agentErr).Errorf("failed to generate agent installer kubeconfig")
+						return common.GenerateErrorResponder(agentErr)
+					}
+					respBody, contentLength, err = b.v2DownloadClusterFilesInternal(ctx, fileName, params.ClusterID.String())
+				}
```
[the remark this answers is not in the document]
bfournie: We still need the call to v2DownloadClusterFilesInternal() for `KubeconfigNoIngress` if `Kubeconfig fails` since the NoIngress may exist. This 2nd call won't be needed for kubeadmin-password

andfasano:
cc @danielerez

internal/ignition/installmanifests.go
```diff
@@ -1279,6 +1279,10 @@ func uploadToS3(ctx context.Context, workDir string, cluster *common.Cluster, s3
 		fullPath := filepath.Join(workDir, fileName)
 		key := filepath.Join(cluster.ID.String(), fileName)
 		err := s3Client.UploadFile(ctx, fullPath, key)
+		if skipUpload(ctx, key, s3Client, installInvoker) {
```
andfasano, line 1282: The problem discussed is that currently the [generateClusterInstallConfig](https://github.com/openshift/assisted-service/blob/03b3c1c1a50e9ded67ca49ca40d8cffd5c37f62b/internal/bminventory/inventory.go#L1812C30-L1812C58) is invoked multiple times (at least twice in the interactive workflow):
* On demand, when the user [downloads/checks](https://github.com/openshift/assisted-service/blob/03b3c1c1a50e9ded67ca49ca40d8cffd5c37f62b/internal/bminventory/inventory.go#L6832) the kubeconfig/kubeadmin-password file in the UI
* When the cluster is effectively [installed](https://github.com/openshift/assisted-service/blob/03b3c1c1a50e9ded67ca49ca40d8cffd5c37f62b/internal/bminventory/inventory.go#L1433), when the user clicks on the install button in the UI.

The problem observed is that the kubeadmin-password gets overwritten, invalidating the previous one that the user originally downloaded (while the new kubeconfig is still ok).

Since in the standard assisted flow the generation of these files is performed only once, I think it would be better to avoid exercising unnecessary code until strictly required, and cut off the execution flow as soon as possible (rather than at the end of the chain) when such files are already available: is there any specific reason to regenerate all these files again? If not the current cache should be good enough to proceed.

A possible simpler approach could be to add a `b.generator.FilesAlreadyAvailable` check at the beginning of [generateClusterInstallConfig](https://github.com/openshift/assisted-service/blob/03b3c1c1a50e9ded67ca49ca40d8cffd5c37f62b/internal/bminventory/inventory.go#L1812) (or any other better name) and return immediately, reusing the existing cache of generated files.
The name of the working dir is [unique](https://github.com/openshift/assisted-service/blob/d22ff98262bd808c2ddb24e212edef4cda0ac1e8/pkg/generator/generator.go#L72) for each distinct cluster, so it seems safe to avoid regenerating all the files every time.

The benefits would be to avoid too much agent-specific logic in the code (simpler maintenance) and unneeded code execution.

internal/bminventory/inventory.go
```diff
@@ -4064,9 +4064,11 @@ func (b *bareMetalInventory) checkFileForDownload(ctx context.Context, clusterID
 	switch fileName {
 	case constants.Kubeconfig:
 		err = clusterPkg.CanDownloadKubeconfig(cluster)
+	case constants.KubeadminPassword:
+		fallthrough
 	case constants.KubeconfigNoIngress:
-		agentInstaller := b.Config.ClusterTLSCertOverrideDir != ""
-		err = clusterPkg.CanDownloadKubeconfigNoIngress(cluster, agentInstaller)
+		agentInstaller := b.installerInvoker == "agent-installer"
+		err = clusterPkg.CanDownloadKubeconfigFiles(cluster, fileName, agentInstaller)
```
andfasano, line 4071:
```suggestion
		err = clusterPkg.CanDownloadKubeconfigFiles(cluster, fileName, b.installerInvoker)
```

nit: embedding the logic in `CanDownloadKubeconfigFiles` could save one line for every potential user of the method

internal/ignition/installmanifests.go
bfournie: I had originally tried the approach to not run [GenerateInstallConfig]( https://github.com/openshift/assisted-service/blob/master/pkg/generator/generator.go#L68)
if workDir had already been created, which is the same as your suggestion. However I found in that case an error that bootstrap.ign wasn't found (see below). It appeared that the invocation when the cluster was installed was necessary to create that file. I can debug this but am concerned that the approach you suggest will be more error prone because the timing is different as to when GenerateInstallConfig will generate the files. That's why I implemented with the focus only on the `kubadmin-password` file so as to not affect the generation of some files which can only be done when the cluster is installed.

```
Apr 01 21:31:02 master-0 service[3187]: time="2025-04-01T21:31:02Z" level=error msg="Failed downloading bootstrap ignition file" func="github.com/openshift/assisted-service/internal/bminventory.(*bareMetalInventory).storeOpenshiftClusterID" file="/src/internal/bminventory/inventory.go:1281" cluster_id=4d25df38-73f1-49af-ae24-014793be661a error="object 4d25df38-73f1-49af-ae24-014793be661a/bootstrap.ign was not found" go-id=637 pkg=Inventory request_id=987af082-a063-44bd-aec8-7ed5770dc8ec
Apr 01 21:31:02 master-0 service[3187]: time="2025-04-01T21:31:02Z" level=warning msg="Cluster installation initialization failed" func="github.com/openshift/assisted-service/internal/bminventory.(*bareMetalInventory).InstallClusterInternal.func3.1" file="/src/internal/bminventory/inventory.go:1426" cluster_id=4d25df38-73f1-49af-ae24-014793be661a error="object 4d25df38-73f1-49af-ae24-014793be661a/bootstrap.ign was not found" go-id=556 pkg=Inventory request_id=987af082-a063-44bd-aec8-7ed5770dc8ec
Apr 01 21:31:02 master-0 service[3187]: time="2025-04-01T21:31:02Z" level=warning msg="Failed to prepare installation of cluster 4d25df38-73f1-49af-ae24-014793be661a" func="github.com/openshift/assisted-service/internal/cluster.(*Manager).HandlePreInstallError" file="/src/internal/cluster/cluster.go:1009" cluster_id=4d25df38-73f1-49af-ae24-014793be661a error="object 4d25df38-73f1-49af-ae24-014793be661a/bootstrap.ign was not found" go-id=637 pkg=cluster-state request_id=
Apr 01 21:31:02 master-0 service[3187]: time="2025-04-01T21:31:02Z" level=info msg="Successfully handled pre-installation error, cluster 4d25df38-73f1-49af-ae24-014793be661a" func="github.com/openshift/assisted-service/internal/cluster.(*Manager).HandlePreInstallError" file="/src/internal/cluster/cluster.go:1019" cluster_id=4d25df38-73f1-49af-ae24-014793be661a go-id=637 pkg=cluster-state request_id=

```

bfournie: OK, after more testing it appears that the errors above were related to the test and not a result of a problem due to the skipping of the install-config generation on the 2nd iteration. So updated to not run GenerateInstallConfig if the workDir already exists. Installation succeeds and the kubeadmin-password is only uploaded once.

internal/bminventory/inventory.go
bfournie: Changed.

andfasano:
Thanks @bfournie tested locally the new patch and was working fine!

/lgtm

danielerez:
Hey @bfournie. Not sure if the subsystem failures related to the change. Have you maybe tried to run them locally?

pkg/generator/generator.go
```diff
@@ -65,7 +65,14 @@ func New(log logrus.FieldLogger, s3Client s3wrapper.API, cfg Config,
 // GenerateInstallConfig creates install config and ignition files
 func (k *installGenerator) GenerateInstallConfig(ctx context.Context, cluster common.Cluster, cfg []byte, releaseImage, installerReleaseImageOverride string) error {
 	log := logutil.FromContext(ctx, k.log)
-	err := os.MkdirAll(k.workDir, 0o755)
+
+	_, err := os.Stat(k.workDir)
```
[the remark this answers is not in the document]
bfournie: The check in this case is for no error `err == nil`, indicating that the directory exists.

bfournie:
> Hey @bfournie. Not sure if the subsystem failures related to the change. Have you maybe tried to run them locally?
I haven't been able to get subsystem tests to run locally yet.

The `subsystem- kubeapi-aws` looks likes its failing the same way, sporadically, in other PRs, so this one's probably not an issue. I'm not sure yet about ` edge-subsystem-aws` as I don't see other PRs with this same failure. I wonder if its related to the [workdir check](https://github.com/openshift/assisted-service/pull/7472/files#diff-9fde1b602271222d4351f8e635e3398b84f0af0cc0df19ce8b9f5095fefb5c1bR69) as it wasn't failing before that was introduced.

andfasano:
@danielerez any idea about the reason of the aws failures?

By looking at the [aws failure](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_assisted-service/7472/pull-ci-openshift-assisted-service-master-edge-subsystem-aws/1909042158615138304/artifacts/subsystem-aws/subsystem-deploy-and-run/build-log.txt) it looks like at least 37 tests are failing because the installation hang for some reason

pkg/generator/generator.go
```diff
@@ -65,7 +65,14 @@ func New(log logrus.FieldLogger, s3Client s3wrapper.API, cfg Config,
 // GenerateInstallConfig creates install config and ignition files
 func (k *installGenerator) GenerateInstallConfig(ctx context.Context, cluster common.Cluster, cfg []byte, releaseImage, installerReleaseImageOverride string) error {
 	log := logutil.FromContext(ctx, k.log)
-	err := os.MkdirAll(k.workDir, 0o755)
+
+	entries, err := os.ReadDir(k.workDir)
+	if err == nil && len(entries) > 0 {
+		log.Infof("Work directory %s already exists, skipping generation of InstallConfig", k.workDir)
+		return nil
+	}
```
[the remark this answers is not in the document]
bfournie: @CrystalChun OKay. Yes I will add an ABI specific check for this.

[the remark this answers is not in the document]
andfasano: Hi @CrystalChun, just for our understanding, under which code flow path it was generating the issue (maybe the controller)? Thanks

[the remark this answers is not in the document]
giladravid16: It's an issue after we generate the manifests for one cluster, and then want to install another cluster. Even ignoring the temp directory for each cluster, there's also a cache directory that keeps the openshift-install binary for each release.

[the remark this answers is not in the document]
andfasano: Thanks @giladravid16 for the details, I think I've found what is the root cause of the issue.

@bfournie the [k.workDir](https://github.com/openshift/assisted-service/blob/d6a37d8229ee6392e0305a6480fb832954783baa/pkg/generator/generator.go#L69) is just the binary installer cache, not the specific cluster one containing the cert files, as mentioned above, ie usually something like `/data/install-config-generate/installercache/...`
The effective folder containing the certs (and ignition files) is instead [clusterWorkDir](https://github.com/openshift/assisted-service/blob/d6a37d8229ee6392e0305a6480fb832954783baa/pkg/generator/generator.go#L79), appearing as `/data/56cf647e-1f0c-4128-aa75-68cf957aa786`, so to fix the problem it should be sufficient to perform the emptiness check against `clusterWorkDir` - rather than `k.workDir`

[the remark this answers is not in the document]
bfournie: To avoid any unintended consequences I just changed this to be agent-specific with https://github.com/openshift/assisted-service/pull/7546

PR #7492, opened 2025-04-01
MGMT-20228: Enhancement to allow cancelling day-2 host's installation

CrystalChun:

- [ ] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [x] Documentation
- [ ] CI/CD 

- [ ] Cloud
- [x] None

## How was this code tested?

- [ ] dev-scripts environment
- [x] No tests needed

## Checklist

## Reviewers Checklist

### DIFF
--- a/docs/enhancements/cancel-day-2-host-installation.md
+++ b/docs/enhancements/cancel-day-2-host-installation.md
@@ -1,0 +1,193 @@
+---
+title: cancel-day-2-host-installation
+authors:
+  - "@CrystalChun"
+creation-date: 2025-04-30
+last-updated: 2025-04-30
+---
+
+# Cancel Day-2 Host Installation 
+
+## Summary
+
+In Assisted-Service's Kubernetes API implementation, there is currently no way to cancel the
+installation of a day-2 host. This leads to scenarios where an Agent can become stuck for a long
+time if the installation fails and the timeout hasn't been reached, or if the user changes their 
+mind about installing a host before the the installation finishes. This proposal aims to allow
+cancellation of installation for a day-2 host. The main use-case is for Hypershift.
+
+## Motivation
+
+There would be times a user is installing a day-2 host and in the middle of installation, they need to
+change something, or the install becomes stuck because of an error. There are also instances where a
+user might need to provision and deprovision a host quickly as it may no longer be needed, such as when
+autoscaler is used. 
+
+When a host starts installation, there's no way to remove it from a cluster until the host installation
+times out and errors out or it finishes installation. These can take a very long time, and it can be a
+frustrating experiences for users.
+
+### Goals
+
+- Once a day-2 host starts installation, the installation can be cancelled
+- There's a straight-forward way to trigger cancellation
+- Host whose installations are cancelled can be reclaimed back to their InfraEnvs and be reused
+- Host cancellation should not affect other hosts in the cluster and should not affect the cluster itself
+- Host cancellation should happen within a reasonable amount of time in order to support dynamic scaling 
+
+### Non-Goals
+
+- Hosts that are installed in a day-1 cluster is out of scope of this enhancement
+- Cluster installation cancellation is out of scope
+- The way reclaim/unbind works should not be affected by these changes
+- This enhancement is focused solely on Assisted-Service offerings using Kubernetes API (SaaS should not be affected by this)
+- The host should still be responsive, in other words, any hosts that are unreachable are out scope of this proposal
+
+## Proposal
+
+Allow cancelling an installing day-2 host through Assisted-Service's Kubernetes API.
+
+### User Stories
+
+#### Story 1
+
+As a cluster admin, when I create a hosted cluster using Hypershift, I'd like to bind and install worker hosts into my cluster. 
+However, if I want to remove the host from the cluster before they finish installing, I should be able to do so and reuse the host for another cluster.
+
+#### Story 2
+
+As a cluster admin using Hypershift, I want to use the autoscaler.
+When the scaling happens and hosts are bound/unbound from the cluster, the hosts should bind and unbind 
+easily from the cluster.
+
+#### Story 3
+
+As a cluster admin, if I install a day-2 host to a cluster, and I see the host installation fails from
+inside the host. I should be able to cancel the installation once it starts and quickly reset the host.
+
+### Implementation Details/Notes/Constraints [optional]
+
+There is an already existing cancel host installation API which maps to the unbind and reclaim stages.
+This is currently only being used when a day-1 cluster installation is cancelled.
+
+#### Detection of Cancellation
+
+When an Agent's cluster deployment reference is removed and the host is in the middle of installation,
+the controller will consider that a trigger to cancel the installation.
+
+Currently, there is a webhook that prevents the cluster reference from being removed
+when a host is in the middle of installation. The proposal for this enhancement is to remove
+that webhook to unblock the removal of the cluster reference.
+
+#### Cancelling Installation
+
+The host should transition to a cancelled state. This will happen through the Agent controller.
+
+#### Cancellation Stages
+
+There are several stages a host can be in when the install is cancelled.
+These are the expected outcomes during each stage:
+
+| Cancellation Stage | Expected Post Cancellation |
+|--------------------|----------------------------|
+| Host is still running assisted installer agent | No reboot needed, should transition back to discovering |
+| Host is/has rebooted | Assisted should not approve CSRs and Node should not be created in the spoke cluster |
+| Host registered as Node to the spoke cluster | This should be considered post-installation and unbind should take care of this as usual. The Node CR should be removed and cleaned up from the spoke cluster|
+
+#### Resetting the Host
+
+After the installation is cancelled, the host will need to be moved back into discovering.
+There are two pathways for this depending on how the host was booted.
+
+##### Automatic "reset"
+
+If the host has a BMH associated with it, there is already an existing flow which will remove the
+detached annotation from the BMH and that will cause it to reprovision the host automatically with
+the discovery image. 
+
+Once the host has been rebooted with the discovery image, it should be "reset"
+and ready to be bound to another cluster.
+
+##### Manual "reset"
+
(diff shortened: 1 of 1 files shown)


docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,117 @@
...
+
+There is an already existing cancel host installation API which maps to the unbind and reclaim stages.
+This is currently only being used when a day-1 cluster installation is cancelled. 
+
+We can use the existing host installation cancellation API, and allow the install cancellation 
+to happen from the Kubernetes API by removing the existing Agent webhook that blocks removing
+the cluster refrence while the Agent is installing.
+
+In the Agent controller, detect if the Agent becomes unbound from cluster while installing.
+If it is unbound, then allow it to be cancelled.
```
carbonin, line 77: How will this work for hosts that are part of an infraenv that references a cluster (early-binding)?

Is that a case we want to handle? Maybe for those hosts deleting the agent could be the cancellation trigger (if that's not already happening)?

CrystalChun: Good point! So I see that there's nothing in the webhook preventing the Agent from being deleted

In the Agent controller, it just tries to remove the spoke resources ([src](https://github.com/openshift/assisted-service/blob/1dc66326d304c3c2bffe346de6e2ab278f0d0ac9/internal/controller/controllers/agent_controller.go#L501)) (which shouldn't be an issue)

And the Agent/host from the DB ([src](https://github.com/openshift/assisted-service/blob/master/internal/bminventory/inventory.go#L3459)) with nothing gating deletion.

I'm not sure cancellation will happen 🤔 (though I might've missed something) so this seems like another case to consider adding a cancellation when the Agent is being deleted, but is not fully installed.

docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,137 @@
...
+Once an Agent is deleted, it is simply removed from the database.
+
+This flow would need to be modified to detect if the Agent is in the middle of installation,
+and if it is, then send the cancel install host command to the Agent before completely allowing
+it to be removed. The Agent should move to the cancel installation stage until the host command
+returns successfully. If the command is not successful, then the Agent will move to an Error state.
+
+There may be some cases where a user might not care if the host is cancelled if they're deleting
+the Agent, so there should also be a hook that allows the host to be force deleted without waiting
+for the cancel command to complete successfully.
```
carbonin, line 97: What is the value of canceling the installation for a host if the host is going to be deleted in any case.
Do we do something on the host itself?

What happens today when you're installing a day2 host with early binding and delete the agent?
Does the host continue to install? Do we just not approve the CSR?

CrystalChun: That's a good point, if the user wanted to delete the host then it shouldn't be gated.

I tested out deleting an Agent when the host went into rebooting and saw that the host continued to try and install, but it failed to start kubelet. Didn't get to the point of approving CSRs, though. 

At any rate, I decided to limit the scope of this enhancement to just late-binding hosts to keep it simple

docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,177 @@
...
+- The host should still be responsive, in other words, any hosts that are unreachable are out scope of this proposal
+
+## Proposal
+
+Allow cancelling an installing day-2 host through Assisted-Service's Kubernetes API.
+
+### User Stories
+
+#### Story 1
+
```
jhernand, line 53: May be good to make it clear in the introduction, or even in the title, that this is for Hypershift only.

docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,177 @@
...
+Host installation cancellation is a completely new flow that could have unintended consequences.
+For example, it might not be able to be cancelled during a certain part of installation, or the
+host might end up in a state that's unreachable rendering it stuck.
+
+##### Mitigation
+
+The way to mitigate this is to fully explore and ensure all paths into and out of this state are covered.
+When testing out this feature, we should test cancelling the installation at different times and phases
+to ensure that cancellation is possible.
```
jhernand, line 121: I'd suggest to also provide a documented and supported manual procedure to return the host to the pool.

docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Test Plan
+
+An e2e test will need to be added or updated for day-2 host installation and cancellation.
+
+## Drawbacks
+
+There are a lot of risks when changing the host stages like this and when interrupting an installation as it can leave the host in a broken state.
+
+If the host becomes unreachable during installation and the user wants to cancel it, there may be no way of even bringing this host back in the first place and would require user intervention.
```
jhernand, line 143: Would be great if we can document the manual steps to do that.

docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,177 @@
...
+- Host cancellation should not affect other hosts in the cluster and should not affect the cluster itself
+- Host cancellation should happen within a reasonable amount of time in order to support dynamic scaling 
+
+### Non-Goals
+
+- Hosts that are installed in a day-1 cluster is out of scope of this enhancement
+- Cluster installation cancellation is out of scope
+- The way reclaim/unbind works should not be affected by these changes
+- This enhancement is focused solely on Assisted-Service offerings using Kubernetes API (SaaS should not be affected by this)
+- The host should still be responsive, in other words, any hosts that are unreachable are out scope of this proposal
```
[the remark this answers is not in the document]
CrystalChun: Good question! Yes, I think that would be considered responsive if a controller like BMO can still control the host. What do you think?

docs/enhancements/cancel-day-2-host-installation.md
CrystalChun: Will do, thank you!

docs/enhancements/cancel-day-2-host-installation.md
CrystalChun: Great suggestion, will add that too. Thank you!

docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,181 @@
...
+the host is unreachable and transition the host to a pending-user-action state.
+
+#### Post Cancellation
+
+##### Successful Cancellation
+
+If a host cancellation was successful, the host will be transitioned to unbinding then reclaiming.
+
+During reclaim, the host should be rebooted with the discovery ISO and its state should be reset to
+discovering once the host tries to register again. This already exists with the current reclaim flow
```
carbonin, line 99: Just a reminder that this will only happen for hosts that have a BMH or have been successfully installed.

If a host never joins the cluster after reboot and does not have a BMH the user will need to manually boot it into the discovery ISO.

I think that's fine (as there's nothing else we can do), but I just wanted to make a note here because it sounds like we're always expecting the host to go through the actual reclaim process.

CrystalChun: Thank you for mentioning this!! I actually didn't know this. Will add some language about this

docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,181 @@
...
+#### Story 3
+
+As a cluster admin, if I install a day-2 host to a cluster, and I see the host installation fails from
+inside the host. I should be able to cancel the installation once it starts and quickly reset the host.
+
+### Implementation Details/Notes/Constraints [optional]
+
+There is an already existing cancel host installation API which maps to the unbind and reclaim stages.
+This is currently only being used when a day-1 cluster installation is cancelled. There is also
+existing logic in the
```
carbonin, line 72: Should there be more words here?

docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,181 @@
...
+#### Cancelling Installation
+
+The host should transition to cancelled state. Once in cancelled state, the host can then
+be unbound from the cluster.
+
+#### Post Cancellation
+
+##### Successful Cancellation
+
+If a host cancellation was successful, the host will be transitioned to unbinding-pending-user-action.
```
carbonin, line 92: What does a successful cancellation mean? Does something need to be done on the host? 
Given the previous section says:
>Once in cancelled state, the host can then be unbound from the cluster.

What is the difference between cancelling and moving the host directly to unbinding-pending-user-action?

docs/enhancements/cancel-day-2-host-installation.md
```diff
@@ -0,0 +1,181 @@
...
+When testing out this feature, we should test cancelling the installation at different times and phases
+to ensure that cancellation is possible.
+
+Providing a supported documented procedure to bring the host back to discovery manually should help to mitigate this too.
+
+## Design Details [optional]
+
+### Open Questions
+
+1. Are there limitations when cancelling an in-progress OpenShift install?
```
carbonin, line 126: I wonder specifically what would happen if the host already rebooted into the node. Is there any time there that we would want to prevent this still (since we wouldn't be able to do anything except mark it as not installed even though it might be installed)?

docs/enhancements/cancel-day-2-host-installation.md
CrystalChun: Good question, modified the section to be automatic vs manual reset.

---
I initially thought cancel would send the `stop-installation` host command to the Agent, but it turns out not to be the case!

docs/enhancements/cancel-day-2-host-installation.md
CrystalChun: Oops this was left over, removed it.

docs/enhancements/cancel-day-2-host-installation.md
CrystalChun: Good question! This might need a doc as well if the host has already joined the cluster as a node.

carbonin (approved):

PR #7511, opened 2025-04-07
ACM-19143: remove obsolete image versions

zszabo-rh:
[ACM-19143](https://issues.redhat.com//browse/ACM-19143): default osImages versions should be taken from [supported-ocp-versions.json](https://github.com/stolostron/acm-hive-openshift-releases/blob/main/supported-ocp-versions.json)
Pls note that corresponding backplane version is 5 versions behind release-ocm version.

Main changes: 
- data/default_os_images.json was modified (+ "make generate")
- from now on subsystem tests use the default version instead of hardcoded values


subsystem/subsystem_suite_test.go
```diff
@@ -33,12 +34,9 @@ import (
 var log *logrus.Logger
 var wiremock *utils_test.WireMock
 var kubeClient k8sclient.Client
+var openshiftVersionLong string = "4.11.59"
 var openshiftVersion string = "4.11"
-var snoVersion string = "4.11"
 var multiarchOpenshiftVersion string = "4.11.0-multi"
```
danmanor, line 39: what are these still used  for ? 4.11 is not available anymore no ?

zszabo-rh: sure, the initial value could be anything, I intentionally kept 4.11 just to see if it really gets overridden by an existing image version.

subsystem/kubeapi/kubeapi_suite_test.go
```diff
@@ -32,7 +34,7 @@ import (
 var log *logrus.Logger
 var wiremock *utils_test.WireMock
 var kubeClient k8sclient.Client
-var VipAutoAllocOpenshiftVersion string = "4.14.0"
```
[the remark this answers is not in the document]
zszabo-rh: it's a deleted line :)

zszabo-rh:
Closed as currently there is no way to separate ACM and SaaS related changes in master.
Tailoring the supported image versions needs to be done for all ACM versions after they are branched down from master.

PR #7521, opened 2025-04-08
ACM-19143: align default osImages for 2.13 with supported-ocp-versions

zszabo-rh:
[ACM-19143](https://issues.redhat.com//browse/ACM-19143): default osImages versions should be taken from [supported-ocp-versions.json](https://github.com/stolostron/acm-hive-openshift-releases/blob/main/supported-ocp-versions.json)
Pls note that corresponding backplane version is 5 versions behind release-ocm version.

Commits:

Main changes: data/default_os_images.json was modified, all other changes were generated.
Necessary unit and subsystem test modifications.


zszabo-rh:
accidentaly rebased on master

PR #7523, opened 2025-04-09
MGMT-20119: Create feature gate to allow installing TNA using the API

giladravid16:
This is the first PR for adding support for TNA Clusters in assisted-service.
The purpose of this PR is to add the minimum required to install TNA clusters using the REST API.
The changes include:
- A new feature gate that when enabled it allows us to set a host's role to "arbiter", if it belongs to a cluster whose OCP version is 4.19 or higher (TNA Clusters are TP in 4.19).
- Allow the user to set a cluster's control plane count to 2 if it's OCP version is 4.19 or higher.
- If the cluster's control plane count is 2 then it must have at least 1 arbiter host.
- Generating ignitions for arbiter hosts 
- Updating the validations and transitions to check arbiter hosts as well.

To check this PR I used my build to install a few TNA clusters and verified that the clusters installed successfully.

Closes [MGMT-20119](https://issues.redhat.com//browse/MGMT-20119)

- [x] New Feature 
- [ ] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

## Checklist

## Reviewers Checklist


internal/cluster/validator.go
```diff
@@ -388,6 +388,12 @@ func (v *clusterValidator) SufficientMastersCount(c *clusterPreprocessContext) (
 		status = ValidationFailure
 	}
 
+	if numMasters < common.MinMasterHostsNeededForInstallationInHaMode &&
+		numMasters >= common.MinMasterHostsNeededForInstallationInHaArbiterMode &&
+		!common.IsClusterTopologyHighlyAvailableArbiter(c.cluster) {
```
carbonin, line 393: This is a strange condition. It's really saying if the number of masters is 2 and we don't have an arbiter then fail.

I feel like we could say that more directly. Maybe change `common.GetHostsByEachRole` to also support arbiter?

internal/ignition/installmanifests.go
```diff
@@ -751,15 +752,22 @@ func (g *installerGenerator) updateBootstrap(ctx context.Context, bootstrapPath
...
 
 			// The BMH files in the ignition are sorted according to hostname (please see the implementation in installcfg/installcfg.go).
 			// The masters and workers are also sorted by hostname.  This enables us to correlate correctly the host and the BMH file
-			if bmhIsMaster(bmh, masterHostnames, workerHostnames) {
+			role := getBmhRole(bmh, masterHostnames, arbiterHostnames, workerHostnames)
+			if role == models.HostRoleMaster {
 				if len(masters) == 0 {
 					return errors.Errorf("Not enough registered masters to match with BareMetalHosts")
 				}
 				host, masters = masters[0], masters[1:]
```
carbonin, line 765: It feels like this might be worth breaking out into a function at this point and reusing it for each role.

internal/ignition/installmanifests.go
```diff
@@ -978,6 +986,18 @@ func (g *installerGenerator) updateIgnitions() error {
 		}
 	}
 
+	// An ignition for arbiter nodes will only be generated for clusters that include arbiter nodes
+	isArbiterExist := common.IsClusterTopologyHighlyAvailableArbiter(g.cluster)
```
carbonin, line 990: nit, but maybe call this `arbiterExists` or `isArbiterCluster`. I think it just reads a bit better

internal/ignition/installmanifests.go
```diff
@@ -1213,13 +1237,20 @@ func (g *installerGenerator) writeHostFiles(hosts []*models.Host, baseFile strin
...
 func (g *installerGenerator) createHostIgnitions() error {
-	masters, workers := sortHosts(g.cluster.Hosts)
+	masters, arbiters, workers := sortHosts(g.cluster.Hosts)
 
 	err := g.writeHostFiles(masters, masterIgn, g.workDir)
 	if err != nil {
 		return errors.Wrapf(err, "error writing master host ignition files")
 	}
 
+	if funk.NotEmpty(arbiters) {
```
carbonin, line 1247: I wonder if this is even needed given that we could have no workers, right?

internal/bminventory/inventory.go
```diff
@@ -2023,19 +2024,32 @@ func validateHighAvailabilityWithControlPlaneCount(highAvailabilityMode string,
...
-		)
+	arbiterClustersNotSupported, err := common.BaseVersionLessThan(common.MinimumVersionForArbiterClusters, openshiftVersion)
+	if err != nil {
+		return err
+	}
+
+	if highAvailabilityMode == models.ClusterCreateParamsHighAvailabilityModeFull {
+		var minMasterHostsNeededForInstallation int64 = common.MinMasterHostsNeededForInstallationInHaMode
+		minVersion := common.MinimumVersionForNonStandardHAOCPControlPlane
+		if !arbiterClustersNotSupported {
```
carbonin, line 2035: I'd prefer this get reversed in some way so we don't have a double negative here. It's harder to reason about.

internal/bminventory/inventory.go
```diff
@@ -2023,19 +2024,32 @@ func validateHighAvailabilityWithControlPlaneCount(highAvailabilityMode string,
...
+			minMasterHostsNeededForInstallation = common.MinMasterHostsNeededForInstallationInHaArbiterMode
+			minVersion = common.MinimumVersionForArbiterClusters
```
carbonin, line 2037: The use of version in this function is very confusing.

It seems like we're trying to reason about different cases for different versions, but I never see the version in a condition anywhere. Is that right?

internal/bminventory/inventory.go
```diff
@@ -6405,12 +6419,33 @@ func (b *bareMetalInventory) V2UpdateHostInternal(ctx context.Context, params in
...
+			err = errors.Errorf("Cannot set role arbiter to host %s in infra-env %s, it must be bound to a cluster with openshift version %s or newer", host.ID, host.InfraEnvID, common.MinimumVersionForArbiterClusters)
+			if cluster != nil && cluster.OpenshiftVersion != "" {
+				arbiterClustersNotSupported, err2 := common.BaseVersionLessThan(common.MinimumVersionForArbiterClusters, cluster.OpenshiftVersion)
+				if err2 != nil {
+					return err2
+				}
+				if !arbiterClustersNotSupported {
+					err = nil
+				}
+			}
```
carbonin, line 6442: This logic feels a bit backwards. Why create the error first just to nil it out if we hit the success case?
Can you refactor this?

internal/bminventory/inventory.go
giladravid16: Changed it

internal/bminventory/inventory.go
giladravid16: Changed it

internal/bminventory/inventory.go
giladravid16: The variable minVersion is only for the error message (if the controlPlaneCount is not allowed). The variable minMasterHostsNeededForInstallation changes based on whether or not the version is at least 4.19 or not (either 2 or 3), and that is used to validate the controlPlaneCount.

internal/bminventory/inventory.go
```diff
@@ -6405,12 +6419,34 @@ func (b *bareMetalInventory) V2UpdateHostInternal(ctx context.Context, params in
...
+					err = errors.Errorf("Cannot set role arbiter to host %s in infra-env %s, it must be bound to a cluster with openshift version %s or newer", host.ID, host.InfraEnvID, common.MinimumVersionForArbiterClusters)
+				}
+			} else {
+				err = errors.Errorf("Cannot set role arbiter to host %s in infra-env %s, it must be bound to a cluster with openshift version %s or newer", host.ID, host.InfraEnvID, common.MinimumVersionForArbiterClusters)
+			}
+		}
+		if err != nil {
+			log.Error(err)
+			return common.NewApiError(http.StatusBadRequest, err)
+		}
```
carbonin, line 6448:
```suggestion
		if !b.TNAClustersSupport {
			err := errors.Errorf("TNA clusters support is disabled, cannot set role arbiter to host %s in infra-env %s", host.ID, host.InfraEnvID)
			log.Error(err)
			return common.NewApiError(http.StatusBadRequest, err)
		}
		
		if cluster == nil || cluster.OpenshiftVersion == nil {
        		  err := errors.Errorf("Cannot set role arbiter to host %s in infra-env %s, it must be bound to a cluster with openshift version %s or newer", host.ID, host.InfraEnvID, common.MinimumVersionForArbiterClusters)
  			log.Error(err)
			return common.NewApiError(http.StatusBadRequest, err)
		}

		arbiterClustersSupported, err := common.BaseVersionGreaterOrEqual(common.MinimumVersionForArbiterClusters, cluster.OpenshiftVersion)
		if err != nil {
			return err
		}
		
		if !arbiterClustersSupported {
			err := errors.Errorf("Cannot set role arbiter to host %s in infra-env %s, it must be bound to a cluster with openshift version %s or newer", host.ID, host.InfraEnvID, common.MinimumVersionForArbiterClusters)
  			log.Error(err)
			return common.NewApiError(http.StatusBadRequest, err)
		}
```

Not super urgent, but I generally prefer guards with early returns to a lot of "else" statements. I find something like this easier to read because you don't have to consider multiple conditions at the same time when thinking about why some code is running. Additionally you don't have to worry about error shadowing here.

Also if we want to deduplicate the logging I'd suggest logging the error from the caller rather than every error case in this function.

giladravid16: Updated it like you suggested

internal/bminventory/inventory.go
```diff
@@ -2023,19 +2024,32 @@ func validateHighAvailabilityWithControlPlaneCount(highAvailabilityMode string,
...
-			http.StatusBadRequest,
-			fmt.Errorf(
-				"there should be %d-%d dedicated control plane nodes for high availability mode %s in openshift version %s or newer",
-				common.MinMasterHostsNeededForInstallationInHaMode,
-				common.MaxMasterHostsNeededForInstallationInHaModeOfOCP418OrNewer,
-				highAvailabilityMode,
-				common.MinimumVersionForNonStandardHAOCPControlPlane,
-			),
-		)
+	arbiterClustersSupported, err := common.BaseVersionGreaterOrEqual(common.MinimumVersionForArbiterClusters, openshiftVersion)
```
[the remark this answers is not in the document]
giladravid16: Originally I thought it should but I'm not sure anymore, because I don't know at this point what cluster topology the user is going for (TNA / TNF / some other future 2 CP topology). I can add it if you think we should and later on we'll add ORs for the supported topologies.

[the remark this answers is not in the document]
carbonin: I mean if we only have 2 CP hosts and `TNAClustersSupport` is `false` then it doesn't matter, right? We should still fail this check.

internal/cluster/validator_test.go
```diff
@@ -1103,6 +1135,35 @@ var _ = Describe("SufficientMastersCount", func() {
...
+							{
+								Role: models.HostRoleMaster,
+							},
+						},
+					}},
+			}
+
+			status, message := validator.SufficientMastersCount(preprocessContext)
+			Expect(status).To(Equal(ValidationFailure))
+			Expect(message).To(Equal("The cluster must have exactly 2 dedicated control plane nodes. Add or remove hosts, or change their roles configurations to meet the requirement."))
```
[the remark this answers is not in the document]
carbonin: I think we should ignore TNF in this case for now. Ideally we should update the message to state the actual options (2CP nodes with arbiter or 3 CP nodes).

If I had to guess I'd say we're going to know of a cluster is TNF because it will likely look like a provider integration or some additional fields/flag at cluster creation.

internal/bminventory/inventory.go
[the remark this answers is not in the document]
giladravid16: updated it

PR #7525, opened 2025-04-09
MGMT-20124: Introduce DependencyOnly property on Operator

pastequo:
This property should be reserved for operator that can not be installed as standalone operator.

This change is a prerequisite to be able to remove nvidia-gpu, amd-gpu or kmm operator if, after choosing openshiftai, we found that no host has amd/nvidia gpu.

Alongside with the initiative to have a single openshiftAI bundle, we need to be able to add dependent operators (namely nvidia-gpu or amd-gpu) based on the host discovery result
Which also imply to be able to remove those dependency-operator if they are not needed (in the first palce or if the only host with nvidia GPU has been removed for example)

This PR proposes to add a new property on Operator that describes if the operator can only exist as a dependency of another operator. So that another PR can remove those operator if they are not considered as needed anymore

The property is intended to be used like this: https://github.com/openshift/assisted-service/pull/7528

- [ ] New Feature 
- [x] Enhancement 
- [ ] Bug fix
- [ ] Tests
- [ ] Documentation
- [ ] CI/CD 

- [x] Cloud
- [ ] None

## How was this code tested?

- [ ] dev-scripts environment
- [ ] No tests needed

Local cluster alongside with all other changes related to have a single openshiftAI bundle

## Checklist

## Reviewers Checklist

### DIFF
--- a/api/vendor/github.com/openshift/assisted-service/models/monitored_operator.go
+++ b/api/vendor/github.com/openshift/assisted-service/models/monitored_operator.go
@@ -27,6 +27,9 @@ type MonitoredOperator struct {
 	// Format: uuid
 	ClusterID strfmt.UUID `json:"cluster_id,omitempty" gorm:"primaryKey"`
 
+	// Whether the operator can't be installed without being required by another operator.
+	DependencyOnly bool `json:"dependency_only,omitempty"`
+
 	// Unique name of the operator.
 	Name string `json:"name,omitempty" gorm:"primaryKey"`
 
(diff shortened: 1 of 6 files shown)


pastequo:
I intend to make this property of the operator itself. I'm aware of that limitation (see description in https://github.com/openshift/assisted-service/pull/7528) I first intended to keep allowing to install nvidia-gpu as a standalone, but after many discussion it was considered as not-supported.

The main issue is that we are not keeping what was the original request of the user. And even if we keep it, the current UX design was to select all operators from a bundle and let the backend remove the non-relevant ones. So I can't figure out when nvidia-gpu should be removed or kept. If the logic changes and becomes "when a user choose a bundle, the minimal set of operators are selected" I might be able to make the distinction. But if a user click on "previous" and "next", we have to keep the information of what the user selected

Another way of describing the problem is that we don't have the notion of bundle in the cluster creation request. If it was a thing, we could say the list of operator are the ones the users want in any case, the ones from the bundle fields are "situational" 

@jhernand WDYT ?

cc @AlonaKaplan

jhernand:
/lgtm

If the decision to not support the NVIDIA GPU operator standalone is made, then I have nothing else to argue.

AlonaKaplan:
/hold
let's discuss the nvidia gpu decision offline first.

